PR #1820, opened 2020-09-10
Make post process uneditable in frame settings for probes

FrancescoC-unity:
Given that post process is not available in probes, with this change it is not possible to override the frame setting of it for reflection cameras. 

This is a fix for https://fogbugz.unity3d.com/f/cases/1271825/

image

**What did I test:**  Verified that the field is not editable for probes but still is for normal cameras


TomasKiniulis (approved):
Looks good to me. I think that verifying normal cameras can still be overridable is sufficient enough.

sebastienlagarde (approved):

sebastienlagarde:
@TomasKiniulis have you check that it was working in default settings as well? (i.e switch from camera to reflection probe and real time and be back.

TomasKiniulis:
> @TomasKiniulis have you check that it was working in default settings as well? (i.e switch from camera to reflection probe and real time and be back.

Ah yes sorry about that. Checked right now, works properly. Camera Post Process default frame settings are still saved when switching between modes
Wg1Smwcuam
Validating that Reflection Probe Post Process frame settings are indeed uneditable now
ZkHdLp7qBR

sebastienlagarde:
thanks!

PR #1823, opened 2020-09-10
Override LOD settings with the one from HDRP

adrien-de-tocqueville:
Fix for https://fogbugz.unity3d.com/f/cases/1211957/
and https://fogbugz.unity3d.com/f/cases/1264734/

HDRP overrides the lod settings from QualitySettings just when rendering, then put back to their initial values.
So when the LODGroup component inspector is draw, it uses some meaningless values.

But for rendering, the lod settings can be changed by the framesettings, so I use the one from the HDRP default frame settings, which are the one used by the editor camera.

I tested that the value displayed in the LODGroup  inspector is correct
and that the LOD transitions in the viewport match the UI


JMargevics (approved):
I0vtvVBpaP

Checked in both HDRP Default settings and camera Custom Frame Settings, Inspector warning message and Scene UI now show correct LOD Bias values ✅

JulienIgnace-Unity:
The fix is included in a bigger LoD related PR: https://github.com/Unity-Technologies/Graphics/pull/2072

PR #1834, opened 2020-09-11
Allow for profiling stats to be displayed as an average over one second 

FrancescoC-unity:
Quality of life: https://fogbugz.unity3d.com/f/cases/1276046/

This change allows to have an option to display profiling stats as averaged over one second rather than immediate values.

This helps in having a more stable value displayed if there is a lot of fluctuation and should smooth out outliers a bit. Now there are more fancy outliers handling we could do, but not sure if we need it for now. 

Samplers


com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
```diff
@@ -733,16 +755,49 @@ void DisableProfilingRecordersRT()
             m_RecordedSamplersRT.Clear();
         }
 
+        float GetSamplerTiming(DebugProfilingType type, ProfilingSampler sampler)
+        {
+            if (data.averageProfilerTimingsOverASecond)
+            {
+                // Find the right accumulated dictionary and add it there if not existing yet.
```
FrancescoC-unity, line 762: Wrong comment  :) Will correct.

sebastienlagarde (approved):

PR #1839, opened 2020-09-11
Improve MSAA color resolve when very different samples are merged together

FrancescoC-unity:
The Fast ToneMap/Invert Tonemap use Max in place of luminance which can introduce bad bias in the computation. This change uses luminance for the reinhard tonemap operator. 

Also, this cleans up the shader a bit so that there isn't as much copy paste as there was before. 

Fix for https://fogbugz.unity3d.com/f/cases/1260556 

**Before** 

image

**After**
image

**IMPORTANT!!** This changes the reference screenshots for MSAA, I am running yamato and will update the PR with new images as soon as it is done.

**What did I test**: The repro project, tried realtime and baked probes.


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

RemyUnity (approved):
Quickly double checked locally, fix confirmed.

PR #1842, opened 2020-09-11
Fix compile issue on vulkan when a min16float4 is generated and shouldn't

FrancescoC-unity:
This is a tentative fix for https://fogbugz.unity3d.com/f/cases/1267920/ , I say tentative as I wasn't able to repro, but the fix would be this. 

**For QA:** It would be great if you could find a way to repro the issue and then see if this fixes it as I was unable to.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
```diff
@@ -48,7 +48,6 @@ struct HDShadowData
         public Vector4      zBufferParam;
         public Vector4      shadowMapSize;
 
-        [SurfaceDataAttributes(precision = FieldPrecision.Real)]
```
sebastienlagarde, line 51: adding @jessebarker as I guess he may be the one who have done this change.

jessebarker: I don't think I added it (hard to tell - git says that line was last edited by you :-) ).  The only issue I can think of is whether both C# and the generated shader struct understand the data as being the same type/layout.  Shouldn't be a problem to take "Real" precision away.

jessebarker (approved):
LGTM

TomasKiniulis (commented):
@FrancescoC-unity I repro the issue, and it's still present with your fix.

Following Gytis steps seems to work for me. What I did was:
1. Open HDRP_Tests project with -force-vulkan -force-xr command line arguments
2. Open 4013_TrasnparentMotionVector scene
3. Mark HDRP_Test_Camera as XR Compatible
4. Run the scene through test runner
5. Unhandled log message: '[Error] Shader error in 'Shader Graphs/SG_Fabric_TransparentWithMotionVector': '' : compilation terminated at line 1103 (on vulkan) shows up

Note that the error shows up only once running the test scene and project has to be reopened and test rerun for shader error to appear again

FrancescoC-unity:
> @FrancescoC-unity I repro the issue, and it's still present with your fix.
> 
> Following Gytis steps seems to work for me. What I did was:
> 
> 1. Open HDRP_Tests project with -force-vulkan -force-xr command line arguments
> 2. Open 4013_TrasnparentMotionVector scene
> 3. Mark HDRP_Test_Camera as XR Compatible
> 4. Run the scene through test runner
> 5. Unhandled log message: '[Error] Shader error in 'Shader Graphs/SG_Fabric_TransparentWithMotionVector': '' : compilation terminated at line 1103 (on vulkan) shows up
> 
> Note that the error shows up only once running the test scene and project has to be reopened and test rerun for shader error to appear again

That's what I did and failed to repro :( In any case I reopend the bug and bounced back to aleksandr that pointed as this as a possible fix.

TomasKiniulis (changes requested):
Requesting changes, due to compile issue still reproducible

sebastienlagarde (approved):

PR #1845, opened 2020-09-11
Fixed a collision in the interaction of RTR and RTGI.

anisunity:
While investigating an other bug, i found out that if a scene was set up with both RTR and RTGI. Enabling or disabling RTGI had an effect on the result of RTR (which it shouldn't). The issue doesn't repro with SSGI.
I fixed the issue and added a test that doesn't pass if the issue is not fixed.

I also found a bit of inconsistencies in the binding of constant buffers for Rtr and RTGI, cleaned that a bit.


iM0ve:
@anisunity  To clarify, all these changes regarding RTGI and constant buffer change are covered by automation and no manual testing for this is required?

anisunity:
I think automation should be enough yes

iM0ve (approved):
A new scene was added to automation and regressions should be covered by existing automation scenes. No user-facing UI. Should be good to go.

sebastienlagarde (approved):

PR #1847, opened 2020-09-11
Force probe to render again if first time was during async shader compilation

adrien-de-tocqueville:
Fix for https://fogbugz.unity3d.com/f/cases/1274519/

When reflection probes are set to render on 'On Enable' and they are enabled during an asnyc shader compilation, objects will be rendered with the cyan shader, so the probe needs to be rendered again when shader compilation is done.

### Testing status

Tested the repro steps of the fogbugz.

Note: That's not easy to test given that the reproduction scene is Amalienborg Palace, but I think you can just clear the Shader Cache instead of removing the whole Library folder, that saves a few minutes.


TomasKiniulis (approved):
For me it seems like the issue of probes getting stuck is only on 8.1.0, if I clear ShaderCache and reopen the project on HDRP/staging I don't get the issue anymore, but updating project directly from 8.1.0 to HDRP/staging does keep the issue, while this PR rerenders the probes and the result is correct. So the fix addresses the issue.

sebastienlagarde (approved):

PR #1849, opened 2020-09-11
Refresh environment library field in Lookdev

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1262990/ 

The UI element showing the library name was not refreshed upon refreshing the library display + the library display is not refreshed on repaint, meaning that it wouldn't necessarily reflect changes done via code, now it should. 

**What did I test**: the repro case in various flavor. Directly as stated, after changing the library from lookdev and after the lookdev was open already.


iM0ve (changes requested):
**Tested:**
The repro steps for the fb

**Issues:**
Change seems to break lookdev on my end. I checked same steps on HDRP/staging and there were no errors

Capture

lookdev

FrancescoC-unity:
> **Tested:**
> The repro steps for the fb
> 
> **Issues:**
> Change seems to break lookdev on my end. I checked same steps on HDRP/staging and there were no errors
> 
> Capture
> 
> lookdev

Very odd :/ Will delete library and try on fresh project.

FrancescoC-unity:
@iM0ve  Should be fixed it now

iM0ve (approved):
Seems to be working properly now

lookdev2

com.unity.render-pipelines.core/Editor/LookDev/DisplayWindow.cs
```diff
@@ -533,7 +533,11 @@ void IViewDisplayer.SetTexture(ViewCompositionIndex index, Texture texture)
             }
         }
 
-        void IViewDisplayer.Repaint() => Repaint();
+        void IViewDisplayer.Repaint()
+        {
+            RefreshLibraryDisplay();
```
RSlysz, line 538: Repaint already call this. It seams not needed.

RSlysz (changes requested):
Remove duplicate call to RefreshLibraryDisplay and it will be good

sebastienlagarde (approved):

com.unity.render-pipelines.core/Editor/LookDev/DisplayWindow.cs
FrancescoC-unity: Done!

sebastienlagarde:
issue adressed. merging

PR #1851, opened 2020-09-11
Fix for cross-pipeline work

elizabeth-legros:
Needed to make sure that unrecognized node types are always considered "invalid" by the target. Added a filterable attribute NeverAllowedByTarget and made targets mark that as not allowed by default.
Also, slight conceptual change in node validation - theres a bit of a difference between a node being invalid in a given target context and a node reporting a validation error. A validation error is just saying that the node is setup wrong, whereas a target saying a node is invalid is saying that this node is wrong AND wont generate code because its wrong. That means a node can not say that it is "valid" or not (because valid is used to determine code generation) but it can add validation errors to itself that get reported. Naming might need to be a bit more explicit about this distinction internally.


sebastienlagarde (approved):
miss a changelog

sebastienlagarde:
Any update on this PR ?
Poking  @phi-lira  for validation

elizabeth-legros:
> Any update on this PR ?
> Poking @phi-lira for validation
Should be ready to go, just poking for reviews

cdxntchou (approved):

phi-lira (approved):

ghost (approved):
Tested a bit, looks good! Made sure different combinations of targets and HDRP being in / not being in the project gave the correct results.

PR #1858, opened 2020-09-14
Seans/performance/urp/update build commands

seanstolberg-unity:
Yaml updates to support building and running URP perf tests in BA.


.yamato/ruamel/jobs/projects/commands/win.py
```diff
@@ -109,13 +109,19 @@ def install_unity_config(project_folder):
     cmds = [
         f'choco source add -n Unity -s https://artifactory.prd.it.unity3d.com/artifactory/api/nuget/unity-choco-local',
         f'choco install unity-config',
+		f'cd {TEST_PROJECTS_DIR}/{project_folder} && unity-config project set registry candidates --project-path .',
```
sophiaaar, line 112: we already set the registry to candidates in the main job, so we don't need this line

sophiaaar: (i can fix this in my branch though, so no worries)

sophiaaar (approved):

PR #1109, opened 2020-07-02
Removes the allocation threshold in the graphic test framework

alelievr:
### Purpose of this PR
Removes the allocation threshold in the graphic test framework. As the allocation has been fixed in trunk: https://fogbugz.unity3d.com/f/cases/1206364/ we can remove the threshold on our side.

BUT there was another allocation hidden in the graphic test framework, that this PR fixes: https://github.cds.internal.unity3d.com/unity/com.unity.testframework.graphics/pull/46
Therefore we should wait for this PR to be merged, the graphic package to be released, and update our manifest before merging this PR.

---
### Testing status
Ran locally HDRP_Tests, green.

https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix%252Fremove-allocation-threshold/.yamato%252Fall-hdrp.yml%2523All_HDRP_fast-trunk/3264748/job

PR #1862, opened 2020-09-14
Removed the GC alloc treshold in the graphic tests

alelievr:
New iteration of https://github.com/Unity-Technologies/Graphics/pull/1109


JulienIgnace-Unity (approved):

sebastienlagarde:
With latest HDRP/staging, we have 2 GC alloc happening (guess is it 2 news one that we haven't catch because of the current threshold of 2.

sebastienlagarde:
will be done in another PR

PR #1866, opened 2020-09-14
Added Simple mode to Earth Preset for PBR Sky

adrien-de-tocqueville:
### Purpose of this PR

Simplifies the PBR Sky UI by adding a simple mode to the earth preset, that hides parameters not frequently used by artists.

image

Also fixed baking that used the 'Air Maximum Altitude' parameter in earth mode instead of the default value (as this parameter is hidden).

---
### Testing status

- [x] Built a player
- [x] Checked new UI names with UX convention
- [x] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [x] C# and shader warnings (supress shader cache to see them)
- [x] Checked new resources path for the reloader (in developer mode, you have a button at end of resources that check the paths)
- Other: 

Tested that disabled parameters are not used by the computations.
Tested migration of profiles with the added parameter.

https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fpbr-sky-ui

---
### Comments to reviewers


EvgeniiG:
I am not convinced this is a good idea since it adds yet another not particularly obvious option.
All parameters have sensible defaults. Artists don't need to touch them if they don't want to.

EvgeniiG:
To make this section visible, set **Type** to **Earth (Advanced)** or **Custom Planet**.

| **Property**                  | **Description**                                              |
| ----------------------------- | ------------------------------------------------------------ |
| **Space Rotation**            | The orientation of space.                                    |
| **Space Emission Texture**    | Specifies a Texture that represents the emissive areas of space. |
| **Space Emission Multiplier** | A multiplier that HDRP applies to the **Space Emission Texture**. |


So you can't even create a starry sky in the Basic mode?

sebastienlagarde:
> I am not convinced this is a good idea since it adds yet another not particularly obvious option.
> All parameters have sensible defaults. Artists don't need to touch them if they don't want to.

The request come from our artist currently when we have done quick meeting about what should be improve in HDRP; 
they wanted to have simplify UX to avoid to touch parameter that shouldn't be touch :)

adrien-de-tocqueville:
After discussion with artist, i hidden the settings they never used, because even if they have good default values, the UI is just overwhelming.

> So you can't even create a starry sky in the Basic mode?

That's one of the things they don't use, partly because it needs a cubemap with a pretty high resolution to look good. Procedural  stars would be appreciated there.

com.unity.render-pipelines.high-definition/Documentation~/Override-Physically-Based-Sky.md
```diff
@@ -33,11 +33,16 @@ To change how much the atmosphere attenuates light, you can change the density o
 
 [!include[](snippets/Volume-Override-Enable-Properties.md)]
 
+### Model
+
+| **Property**                   | **Description**                                         |
+| ------------------------------ | ------------------------------------------------------- |
+| **Type**                       | Indicates a preset HDRP uses to simplify the Inspector. Selecting Earth will only show properties suitable to simulate Earth. |
```
JordanL8, line 40: Indicates a preset HDRP uses to simplify the Inspector. If you select **Earth (Simple)** or **Earth (Advanced)**, the Inspector only shows properties suitable to simulate Earth.

JordanL8 (changes requested):
Looks good, just one suggestion though.

EvgeniiG:
> After discussion with artist, i hidden the settings they never used, because even if they have good default values, the UI is just overwhelming.
> 
> > So you can't even create a starry sky in the Basic mode?
> 
> That's one of the things they don't use, partly because it needs a cubemap with a pretty high resolution to look good. Procedural stars would be appreciated there.

That's no different from any (HDRI) sky cubemap.

You can render to texture with a custom shader, but it is obviously going to be more expensive.

I will not do anything that is built-in.

EvgeniiG:
> > I am not convinced this is a good idea since it adds yet another not particularly obvious option.
> > All parameters have sensible defaults. Artists don't need to touch them if they don't want to.
> 
> The request come from our artists currently when we have done quick meeting about what should be improve in HDRP;
> they wanted to have simplify UX to avoid to touch parameter that shouldn't be touch :)

???

Just... Don't touch them? Where's the logic here? It's not like they don't fit on the screen.

The "Advanced Settings" concept we have in Unity is not helpful since it impedes discoverability, and I saw many people (including myself) struggle to find the parameters or hidden buttons while I was still at the office.

sebastienlagarde:
> Just... Don't touch them? Where's the logic here? It's not like they don't fit on the screen.

We have a UX team now, and our artists participate actively in discussion. It is not a personal opinion.

sebastienlagarde:
> > After discussion with artist, i hidden the settings they never used, because even if they have good default values, the UI is just overwhelming.
> > > So you can't even create a starry sky in the Basic mode?
> > 
> > 
> > That's one of the things they don't use, partly because it needs a cubemap with a pretty high resolution to look good. Procedural stars would be appreciated there.
> 
> That's no different from any (HDRI) sky cubemap.
> 
> You can render to texture with a custom shader, but it is obviously going to be more expensive.
> 
> I will not add anything that is built-in.

> That's no different from any (HDRI) sky cubemap.
yes, and we don't see start in night sky either.

> You can render to texture with a custom shader, but it is obviously going to be more expensive.
doesn't solve the resolution issue.

> I will not add anything that is built-in.
We will discuss with the artists. There seems to be a need to be able to get night sky with stars. We will evaluate the priority and put it or not in the roadmap if required. The team will handle the roadmap.

EvgeniiG (approved):

sebastienlagarde (approved):

TomasKiniulis (approved):
What I've tested:
- Updating 7.4.1 HDRP project with PBR sky ticket - PBR sky gets correctly updated to "Earth(Advanced)"
- Updating 7.4.1 HDRP project with PBR sky unticked- PBR sky gets correctly updated to "Custom Preset"
- Checked if variables are shared between presets

Nice change! Sea Level kind of stands out not taking the value from "Planet Center Position", but I guess making it shared would create even more confusion since X and Z values would not be shared between presets anyway

PR #1873, opened 2020-09-14
Post Processing disabling in Universal Render Pipeline asset

lukaschod:
# **Please read;**

### Checklist for PR maker
- [ ]  Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).

---
### Purpose of this PR
Moved PostProcessData into Universal Render Pipeline asset and made it optional. Builds will no longer include post processing shaders and textures when it will be disabled.

This includes:
- Added a toggle in the Universal Render Pipeline asset to disable post processing.
- Added a warning to the Camera Inspector when PostProcessing is enabled on a camera but the include toggle is turned off.
- Unity automatically strips post processing data as it is controlled by reference.
- Supports old code path for projects that has modified PostProcessData in Renderer.

After discussing @phi-lira and @ellioman we decided that it falls into Quality initiative:
- It fixes bugs that are coming related to it. This is something frequently request by users (@ellioman found this too https://gist.github.com/grapefrukt/c6b90a9d427d6176f58f8b981ebc4779) and from time to time preceived as bug https://fogbugz.unity3d.com/f/cases/1269524/.
- It is low risk.
- Quite trivial fix.

## This PR is based on https://github.com/Unity-Technologies/Graphics/pull/752.

---
### Testing status

- [x] Opened test project + Run graphic tests locally
- [x] Built a player
- [x] Checked new UI names with UX convention
- [x] Tested if post process data is stripped when post processing disabled
- [x] Tested new UI changes, if they act correctly
- [ ] Test post processing option disabling with ForwardRenderer and Renderer2D
- [ ] Currently post processing stripping works by excluding PostProcessData asset from referencing in Universal Render Pipeline Asset (In UI it is coverted to bool, where null is false and true is anything else). We need to make sure it is good workflow for user and we do not get in strange states
- [ ] Check all UI changes (There is images below). See if from user perspective it is logic and easy to use
- [ ] Test if it does not break old ForwardRenderer and Renderer2D post processing. Especially in cases where they had custom PostProcessData assigned

**Automated Tests**: It would require integration testing which is currently not present in this repository.

**Yamato**: https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics/tree/univeral%252Fdisable-post-processing-v2

For testing the feature I recommend using project in case https://fogbugz.unity3d.com/f/cases/1269524/.

**Did some testing with build report**
When post processing enabled:
image

When post processing disabled:
image

We roughly save 5mbs of memory with it.

---
### UI/UX

**Forward Renderer now looks like this**
image

**Option in Universal Render Pipeline asset**
Disabled
image
Enabled
image

**If project uses modified PostProcessData in Renderer**
Renderer
image
URP asset
image

**In camera when post processing is enabled, but disabled in renderer**
image


Verasl (changes requested):
I don't see anything handling the two post processed AA options, FXAA and SMAA, are they still selectable and also what is our stance here, strip them out still?

phi-lira:
@martint-unity and @TheoWong-pixel for UX.

TheoWong-pixel:
I think we should grey it out to reinforce to users that they need to enable PP through the renderer Asset. the way our UI is laid out its not immediately obvious that AA is reliant on PP being enabled. I would advise that i this state the tooltip for AA also mentions that it is not available because PP is not enabled in the renderer asset.

lukaschod:
I have updated UI

Ecksters:
@lukaschod Thanks for doing this man, this is a much-needed fix.

Verasl (approved):
New changes look good to me

com.unity.render-pipelines.universal/Runtime/ForwardRendererData.cs
```diff
@@ -62,7 +62,6 @@ public sealed class ShaderResources
             public Shader materialErrorPS;
         }
 
-        [Reload("Runtime/Data/PostProcessData.asset")]
         public PostProcessData postProcessData = null;
```
lukaschod, line 65: I want to add Obsolete attribute here, but it is a bit tricky. We still want to access it internally, so we could support old code path. It means we can not reference name postProcessData. Otherwise we will get warrnings.

Only solution I could think up is and it is not clean:
- `public PostProcessData postProcessData = null;` =>`[FormerlySerializedAsAttribute("postProcessData ")] public PostProcessData m_PostProcessData = null;`
- add `[Obsolete("")] public PostProcessData postProcessData { get =>m_PostProcessData; set { m_PostProcessData  = value;}}`
- add `internal PostProcessData _postProcessData { get =>m_PostProcessData; set { m_PostProcessData  = value;}}`
And this way we would arhieve all needed things:
- Warn users not to use postProcessData.
- Use _postProcessData internally to support old code path.
- Ensure that does not break serialization.
- Keep correct syntax in public eye at least.

WDYT? TBH I do not know if changing public field to property is considered breaking change or not? I think there is some cases were you could have different results.

TheoWong-pixel (approved):

ellioman (approved):

com.unity.render-pipelines.universal/Runtime/ForwardRendererData.cs
ellioman: Pretty scared that changing a field to a property will be considered a breaking change, which is actually okay in a major version bump like this one but not for the backports :)

You can put the obsolete tag on it and then where we use it internally use the `#pragma warning disable 618` and `#pragma warning restore 618`. That's fine as we're using it for old code paths but we do not want users to do that.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -885,6 +903,15 @@ static void UpgradeAsset(UniversalRenderPipelineAsset asset)
 
                 asset.k_AssetPreviousVersion = 5;
             }
+
+            if (asset.k_AssetPreviousVersion < 6)
```
martint-unity, line 907: Why an empty upgrade?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -101,8 +101,8 @@ public partial class UniversalRenderPipelineAsset : RenderPipelineAsset, ISerial
         ScriptableRenderer[] m_Renderers = new ScriptableRenderer[1];
 
         // Default values set when a new UniversalRenderPipeline asset is created
-        [SerializeField] int k_AssetVersion = 5;
-        [SerializeField] int k_AssetPreviousVersion = 5;
+        [SerializeField] int k_AssetVersion = 7;
+        [SerializeField] int k_AssetPreviousVersion = 7;
```
martint-unity, line 105: Why 7 and not 6?

com.unity.render-pipelines.universal/Runtime/ForwardRendererData.cs
martint-unity: #pragma warning disable 618 // Obsolete warning 
Maybe is the right thing to do

rainsing (approved):

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
lukaschod: Before version 7 it would trigger this call:
            if (k_AssetPreviousVersion != k_AssetVersion)
            {
                EditorApplication.delayCall += () => UpgradeAsset(this);
            }

However now that I added version 7 it is actually not needed.
Going to remove it

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
lukaschod: The latest version now is 7 as I understand when we create the asset we want the latest one. Now why it was not 6 before, somebody forgot to bump it (So we were actually always upgrading new asset from 5 to 6.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -56,6 +56,8 @@ internal class Styles
             public static GUIContent supportsSoftShadows = EditorGUIUtility.TrTextContent("Soft Shadows", "If enabled pipeline will perform shadow filtering. Otherwise all lights that cast shadows will fallback to perform a single shadow sample.");
 
             // Post-processing
+            public static GUIContent postProcessLabel = new GUIContent("Post Process Data", "The asset containing references to shaders and Textures that the Renderer uses for post-processing.");
```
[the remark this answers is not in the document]
lukaschod: Fixed in latest commit

com.unity.render-pipelines.universal/Runtime/ForwardRendererData.cs
lukaschod: Deprecated it in latest commit and add #pragma warning disable 618. Thanks for suggestion it is much better 👍

martint-unity (approved):
lgtm

erikabar (changes requested):
Issues are described in the following document: https://docs.google.com/document/d/1pmcuExn7o9o2D0_mdWq8fYXWfdVnUXCVx0iQtmigPR8/edit?usp=sharing

erikabar (approved):
all issues were fixed https://docs.google.com/document/d/1pmcuExn7o9o2D0_mdWq8fYXWfdVnUXCVx0iQtmigPR8/edit#heading=h.lpmwclu94k8

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -887,6 +905,12 @@ static void UpgradeAsset(UniversalRenderPipelineAsset asset)
 
                 asset.k_AssetPreviousVersion = 5;
             }
+
+            if (asset.k_AssetPreviousVersion < 7)
+            {
+                asset.postProcessData = PostProcessData.GetDefaultPostProcessData();
+                asset.k_AssetPreviousVersion = 7;
+            }
```
Verasl, line 913: After having a good look at this, it seems there is some work to be done for the upgrader, for now the only tweak I would suggest is to add `EditorUtility.SetDirty(asset);` after these two upgraders to make sure we are dirtying users asset files, otherwise they may continue to upgrade their assets without saving.
Then it's green from me and we need to look at making this asset upgrader more along the lines of the material upgrader in a later PR.

Verasl (changes requested):
This PR has exposed issues with the upgrader, we can add one line here at least to solve some of the pain, and the rest can be done in an additional PR to make the asset upgrader nicer.

Verasl (approved):
LGTM

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -94,6 +94,8 @@ private static class Profiling
 
         public ForwardRenderer(ForwardRendererData data) : base(data)
         {
+            UniversalRenderPipelineAsset urpAsset = GraphicsSettings.renderPipelineAsset as UniversalRenderPipelineAsset;
```
phi-lira, line 97: Shouldn't it be GraphicsSettings.currentPipelineAsset? i'm actually confused about the difference between this and renderPipelineAsset

phi-lira (approved):

phi-lira:
I think you should add an entry to the changelog to mention that now it's possible to strip post-processing. 
 A mention in the documentation explaining how to strip post-processing and a paragraph on the What's new page.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
lukaschod: I took it from line 140, not sure

com.unity.render-pipelines.universal/Documentation~/universalrp-asset.md
```diff
@@ -95,6 +95,8 @@ This section allows you to fine-tune global post-processing settings.
 
 | Property         | Description                                                  |
 | ---------------- | ------------------------------------------------------------ |
+| __Post Processing__ | Check this box to enable post processing in your scene. This includes stripping post processing shaders and textures. <br/>**Note:** It will not strip assets if they are referenced somewhere else in scene or even in other URP Asset. |
```
oleks-k, line 98: "in your scene" - remove.
"This includes stripping post processing shaders and textures." - "If you clear this check box, Unity excludes post-processing shaders and textures from the build."
"**Note:** It will not strip assets if they are referenced somewhere else in scene or even in other URP Asset." - "**Note:** Unity does not exclude assets related to post-processing in one of the following cases:
* Other assets in the build refer to the assets related to post-processing.
* A different URP asset has the Post-processing property enabled."

oleks-k (approved):
Approved with my changes.

TheOneVries:
So when will this appear in the URP asset?

Hitomilras:
There is still no possibility to disable post processing in Unity 2020 LTS and last URP. Please, dot it, it's very important for some projects.

lukaschod:
> There is still no possibility to disable post processing in Unity 2020 LTS and last URP. Please, dot it, it's very important for some projects.

So it should be landed and working. Could you eleborate what is not working are you missing option (BTW option is in renderer)?

PR #1877, opened 2020-09-14
[ShaderGraph] [bugfix 1182895] Clean up asset dependency declarations, use GUIDs instead of Paths

cdxntchou:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1182895/

Some background:  https://docs.google.com/presentation/d/1yH5VyURYhj5ADZ8pkRIMBo9v7efDaAA1Zsuwxo4O9wQ/edit#slide=id.g988aef9a67_0_189

The issue is that asset dependencies are recorded by path, which does not work if the dependency file has never existed in the project.  This PR changes dependencies to be recorded by GUID instead, which will work even if the file has never existed in the project.

This PR also cleans up the asset dependency system, converting everything to use an "AssetCollection" class that records GUIDs and intended use flags, making it much clearer what each dependency is used for.

This also lets us remove some unnecessary re-import cases.
For example, we can now differentiate between artifact dependencies and ExportPackage dependencies, which allows us to optimize re-imports for CustomFunctionNode include file dependencies inside SubGraphs.   (previously changing the included file would cause the entire subgraph chain to reimport; but now only the final ShaderGraph will reimport).

There are some minor optimizations included, mainly from using GUIDs instead of strings everywhere, and not having to copy lists around because we're using a single unified AssetCollection structure.

---
### Testing status

Manually tested in ShaderGraph (windows editor):

Tested copying ShaderGraph using SubGraph from another project (not copying the SubGraph) -- ShaderGraph fails with missing SubGraph.

Then verified that copying the SubGraph causes the ShaderGraph to reimport and fix itself.

Validated manually (via debug output) that the correct dependencies are declared for a graph containing:
One and two-level SubGraph dependencies, with Custom Function nodes at each level.
Assigned Diffusion Profile (via node and via slot).

Verified that SubGraph import dependencies correctly invalidate and reimport ShaderGraphs and SubGraphs.
Verified that CustomFunctionNode include file dependencies only invalidate the final ShaderGraph.
Verified that template and Target/Sub-Target C# file dependencies correctly invalidate only the ShaderGraphs that use them.

Verified that circular dependencies are still detected correctly (A->B->A) and (A->B->C->A).

**Yamato**:
https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252F1182895

---
### Comments to reviewers
Looks big, but all of the target/subtarget files are just changing them to use GUIDs instead of GUID strings.

Basically it's just introducing the AssetCollection class, and changing everything to use that instead of directly passing around Lists of strings.  A good entry point is to look at AssetCollection.cs, then take a look at IHasDependencies.cs and the classes that implement it:  SubGraphNode.cs, CustomFunctionNode.cs, and DiffusionProfileInputMaterialSlot.cs.

Then the meat of the changes is in the importer classes:  ShaderGraphImporter.cs and ShaderSubGraphImporter.cs, where it uses the gathered AssetCollections to declare the dependencies properly.


elizabeth-legros (approved):

com.unity.shadergraph/Editor/Generation/Collections/AssetCollection.cs
```diff
@@ -0,0 +1,47 @@
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace UnityEditor.ShaderGraph
+{
+    // this class is used to track asset dependencies in shadergraphs and subgraphs
+    // that is: it tracks files that the shadergraph or subgraph must access to generate the shader
+    // this data is used to automatically re-import the shadergraphs or subgraphs when any of the tracked files change
+    public class AssetCollection
```
alelievr, line 10: Is it expected for this class to be public? If so, you need to add the documentation

alelievr (approved):

com.unity.shadergraph/Editor/Generation/Collections/AssetCollection.cs
cdxntchou: good point, I can probably make it internal for now -- it will be made public when we make Targets public.

pbbastian (approved):

cdxntchou:
Katana all green:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252F1182895/.yamato%252F_abv.yml%2523all_project_ci_trunk/3510631/job/pipeline

GrantLamb-Unity (approved):
Testing that was listed as completed within the PR description looks to be sufficient coverage for this type of change. I'm going to approve this PR. 

Excellent work!

PR #1884, opened 2020-09-15
Remove BLEND keywords and replace with constant branch 

FrancescoC-unity:
Mostly all in the title.  This should reduce build times when a mix of blend modes are used in the project. 

There were some gotchas with VFX and shadergraph, but all are sorted in this PR. 
Tests pass locally, but launching yamato anyway to be sure.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2020.1-to-2020.2.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2020.1-to-2020.2.md
@@ -183,6 +183,8 @@ From 10.x, HDRP includes a new optimization for [Planar Reflection Probes](Plana
 For example, the call in the Lit shader has been updated to:
 `float4 preLD = SampleEnv(lightLoopContext, lightData.envIndex, R, PerceptualRoughnessToMipmapLevel(preLightData.iblPerceptualRoughness) * lightData.roughReflections, lightData.rangeCompressionFactorCompensation, posInput.positionNDC);`
 
+## Raytracing
+
 From Unity 2020.2, the Raytracing Node in shader graph now apply the raytraced path (previously low path) to all raytraced effects but path tracing.
 
 ## Custom pass API


alelievr (approved):
LGTM

com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl
```diff
@@ -76,23 +80,29 @@ float4 EvaluateAtmosphericScattering(PositionInputs posInput, float3 V, float4 i
     float3 volColor, volOpacity;
     EvaluateAtmosphericScattering(posInput, V, volColor, volOpacity); // Premultiplied alpha
 
-    #if defined(_BLENDMODE_ALPHA)
+    if (_BlendMode == BLENDMODE_ALPHA)
```
sebastienlagarde, line 83: we shoud simplify those branch

result.rgb = result.rgb * (1.0 - volOpacity) + (_BlendMode != BLENDMODE_ADDITIVE) ? volColor * result.a : 0.0;

FrancescoC-unity: I did think about it, but did not to mantain the lengthy comments there and the actual cost is fairly small. 

I guess I can reword the comments, we'd lose a bit of clarity, but should still be ok.

sebastienlagarde: you can just comment all the current code and put an optimized version at the end, it is fined :)

FrancescoC-unity: To be clear though, both version compile to exactly the same amount of ops/cycles as it recognizes the two branches are the same. 

The scalar branch and the ternary operator both translate to a v_cmp_eq_f32 anyway.

PaulDemeulenaere (approved):
We discussed about this : https://github.com/Unity-Technologies/Graphics/pull/1884/files#diff-0f3209dd5b092701c9a479cd175f56c9R434

It seems fine, no real impact on the VFX where it's still a #define 

Specific to HDRP.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitTessellation.shader
```diff
@@ -353,6 +351,8 @@ Shader "HDRP/LitTessellation"
     // LitShading.hlsl implements the light loop API.
     // LitData.hlsl is included here, LitShading.hlsl is included below for shading passes only.
 
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl"
```
sebastienlagarde, line 354: note for myself: upgrade guide

com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl
```diff
@@ -33,27 +33,31 @@ float4 ApplyBlendMode(float3 diffuseLighting, float3 specularLighting, float opa
     // When doing off screen rendering (low-res transparency or after post process pass) and rendering opaque objects, opacity needs to be set to zero for proper compositing.
     if (_OffScreenRendering != 0)
         opacity = 0;
-#endif
+
+    return float4(diffuseLighting + specularLighting, opacity);
```
sebastienlagarde, line 37: hmm... I don't understand how the previous code was working?

sebastienlagarde: ok got it :)

com.unity.render-pipelines.high-definition/Editor/VFXGraph/Outputs/VFXAbstractParticleHDRPLitOutput.cs
```diff
@@ -431,13 +431,13 @@ public override void OnEnable()
                 switch (blendMode)
                 {
                     case BlendMode.Alpha:
-                        forwardDefines.WriteLine("#define _BLENDMODE_ALPHA");
+                        forwardDefines.WriteLine("#define _BlendMode 0");
```
sebastienlagarde, line 434: hmm... wondering if we can define based on 

#define BLENDMODE_ALPHA (0)
#define BLENDMODE_PREMULTIPLY (4)
#define BLENDMODE_ADDITIVE (1)

here

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.shader
```diff
@@ -120,7 +120,8 @@ Shader "HDRP/AxF"
 
 
 //      [ToggleUI] _EnableFogOnTransparent("Enable Fog", Float) = 1.0
-//      [ToggleUI] _EnableBlendModePreserveSpecularLighting("Enable Blend Mode Preserve Specular Lighting", Float) = 1.0
+        // We never set this as it is always 1, but its definition is needed when performing blending in Material.hlsl
+        [HideInInspector][ToggleUI] _EnableBlendModePreserveSpecularLighting("Enable Blend Mode Preserve Specular Lighting", Float) = 1.0
```
sebastienlagarde, line 124: it is not true (at least for now as _BLENDMODE_PRESERVE_SPECULAR_LIGHTING is still a keyword)

com.unity.render-pipelines.high-definition/Editor/VFXGraph/Outputs/VFXAbstractParticleHDRPLitOutput.cs
FrancescoC-unity: No won't work as we don't have the include in hlsl all the time for vfx last I checked. I can see if I can make sure that is included, but there must have been a reason.

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.shader
FrancescoC-unity: I am referencing the property not keyword here, beforehand the property was not used, here I instead just define it (the keyword is still used as it should)

FrancescoC-unity: In any case it is actually not true anymore (is not needed anymore after some subsequent changes I made) so I will revert this bit tomorrow.

com.unity.render-pipelines.high-definition/Editor/VFXGraph/Outputs/VFXAbstractParticleHDRPLitOutput.cs
FrancescoC-unity: That said, while def looks better, it is still hardcoded in a string, so won't save our ass in case of changes.

PaulDemeulenaere: FYI, this defines lands here
Unlit :
image

Lit : 
image

Indeed, there isn't any include before these define, we can potential feed these `#define BLENDMODE_ALPHA (0)` within the same function, before the switch, not sure it's worth it.
Isn't there a enum which fits these value ? I could be a generated string.
```
#define BLENDMODE_ALPHA (0)
#define BLENDMODE_PREMULTIPLY (4)
#define BLENDMODE_ADDITIVE (1)
```

@FrancescoC-unity Can you confirm these defines are only needed for lit output ? The output which are including :
- Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl
- Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
- Packages/com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinUtilities.hlsl

FrancescoC-unity: There is an enum on C# that generate an hlsl file with the defines, that hlsl file is included in Material/Material.hlsl
 , **however** we cannot do our define after the include of Material.hlsl, otherwise it won't compile. So we need an additional include for that specific file. 

Whether is worth it or not I guess I'll leave the decision to you guys, feels a bit overkill to me, but indeed it'd make it a tad more readable. 

This would be needed for whatever includes Material.hlsl, so yes, I assume only lit output.

sebastienlagarde: no, it is required for both lit and unlit transparent.
ok nevermind for now, this is something that can't change anyway to not break the universe, let's just add a comment on which enum it map for now.

FrancescoC-unity: No, unlit vfx doesn't use the blend mode as in HDRP, so does not need this as it does not include Material.hlsl

FrancescoC-unity: This was verified btw running the test, if it was needed for Unlit too the test wouldn't have compiled

PaulDemeulenaere: > no, it is required for both lit and unlit transparent.

> No, unlit vfx doesn't use the blend mode as in HDRP,

I confirm @FrancescoC-unity point, only needed for Lit HDR. For instance, it's needed using internal EvaluateAtmosphericScattering
For the fog evaluation, we are manually applying falloff using other defines : https://github.cds.internal.unity3d.com/unity/vfx-graphics/blob/vfx/staging/com.unity.render-pipelines.high-definition/Runtime/VFXGraph/Shaders/VFXCommon.hlsl#L154

If there is a compilation error in case of mistake, that's fine !

sebastienlagarde (approved):

sebastienlagarde (approved):

PR #1885, opened 2020-09-15
Move diffusion profile default settings

alelievr:
### Purpose of this PR
Move the diffusion profile settings list from the HDRP asset to the HDRP Default settings in the Project Settings Window.
image

Note that there is no migration code in this PR, it's only a visual change, the actual list of diffusion profile is still stored in the HDRP asset.

The only thing that changes is that now we're supposed to use `HDRenderPipeline.defaultAsset.diffusionProfileSettingsList` to access the list of diffusion profiles instead of the current HDRP asset.

---
### Design notes

We decided to not move the default diffusion profiles to the default volume system because it required a complex migration step that would have broken a lot more projects and as we are in a stabilization phase, we definitely don't want this kind of changes to land now.
That's why we decided to do the "soft" version: only move the UI instead of the data. Note that these data in the default HDRP asset will be moved to a separate asset in the future with the work of Jennifer.

---
### Testing status

I Tested in the HDRP_Tests project that all the diffusion profiles in the project are still rendering correctly (test 1218)
Also checked that the list in the test project is displayed correctly:
image

**Does it needs more testing?**
Maybe a more complex test with different diffusion profiles per quality levels in different HDRP assets? Not sure it's required because this is not supposed to work and needs to be handled manually (cf upgrade guide).


JMargevics:
Hey! 
YtVNy2eFts
Plus button doesn't really work for me.

alelievr:
I pushed a fix for the diffusion profile list

+ I improved HDRP default settings window performance by ~x5

Before:
image
After:
image

sebastienlagarde (approved):

sebastienlagarde:
@alelievr Merge needed + added lewis for doc check

JMargevics (approved):
List is working now! 

- Adding profiles via Fix Now button works correctly ✅
- Adding and removing items in the list works correctly now ✅


Not much else to check, all good.

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2020.1-to-2020.2.md
```diff
@@ -185,3 +185,12 @@ protected override void Execute(CustomPassContext ctx) { ... }
 
 Previously, to convert a 2D flipbook texture to the 3D format Density Mask Textures require, you needed to use the __Density Volume Texture Tool__ in the __Window > Rendering__ menu.
 From Unity 2020.2, you can now do this conversion directly through the __Texture Importer__. For information on how to use the importer to convert the flipbook texture, see the [Density Volume documentation](Density-Volume.md).
+
+## Diffusion Profiles
+
+The diffusion profile list have been moved from the HDRP asset inspector UI to the **Default HDRP Settings** (in the project settings window).
```
sebastienlagarde, line 191: ceci ets un test

PR #1886, opened 2020-09-15
Fixed compilation Non-invocable member 'HDWizard.ToolbarRadio.radioLength' cannot be used like a method [skip ci] 

alelievr:
### Purpose of this PR
Fixed compilation error:
image
```
com.unity.render-pipelines.high-definition\Editor\Wizard\HDWizard.Window.cs(384,36): error CS1955: Non-invocable member 'HDWizard.ToolbarRadio.radioLength' cannot be used like a method.
```

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
@@ -381,7 +381,7 @@ namespace UnityEditor.Rendering.HighDefinition
             var toolbar = new ToolbarRadio();
             toolbar.AddRadios(tabs);
             //make sure when we open the same project on different platforms the saved active tab is not out of range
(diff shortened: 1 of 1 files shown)


RSlysz (commented):
🟢 
Now, why this issue have come up in the first place? It seams to be old code that worked fine

RSlysz (approved):

alelievr:
The code was changed here few hours ago: https://github.com/Unity-Technologies/Graphics/pull/1609/files#diff-16b4be0020518de614ff58b1517a0718R384

sebastienlagarde (approved):

sebastienlagarde:
> 🟢
> Now, why this issue have come up in the first place? It seams to be old code that worked fine

The problem is more that Yamato was not launch, and I haven't noticed that.

RSlysz:
Yes but the former PR say that Opened test project + Run graphic tests locally have been tested. This should have throw the compile error...

esdasuka:
it is my bad, i pparently have some uncommitted changes in the branch locally. should have been more careful, but still I thought the yamato jobs would kick in anytime there is a merge.

PR #1887, opened 2020-09-15
[ShaderGraph] [bugfix 1257015] previews should have a fallback shader

cdxntchou:
### Purpose of this PR
Fixes bug https://fogbugz.unity3d.com/f/cases/1257015/

When the active RP is not supported by the current set of Targets on a ShaderGraph, the preview will now show up with the fallback shader.

The original change here was made by Matt back in March, (f58e278 - Fix preview generation on assembly reload)

I undid the change, and tested assembly reloads -- previews seem to work properly now, so I think the preview refactor has fixed the issues this was originally working around.

---
### Testing status

Tested manually in editor (windows) with repro steps.
Tested assembly reloads, with both HDRP and URP only shaders, and with combined HDRP+URP shaders.
Tested swapping active render pipeline, to see that preview updates correctly.

---
### Comments to reviewers


esmelusina (approved):

LandonTownsendUnity:
Preview doesn't show when the current render pipeline is builtin:
image
Do we want to fix this?

LandonTownsendUnity (approved):
Result with builtin is expected. Approving.

PR #1889, opened 2020-09-15
Sg/context undo block

esmelusina:
### Purpose of this PR
Address this bug and a few adjacent; https://fogbugz.unity3d.com/f/cases/1257020/

Soft-locking is caused by a null exception in the undo-redo. In many cases this is from the logical and visual hierarchy of block nodes being inconsistent. The other part is from how context menus are built when a stack node is floating. This fix should snap the block node back to its owning stack node instead of attempting to build a bad context menu.

backport: #1931

### Testing status
Manual testing- drag a block node outside of the stack node container and right click to build a context menu. Then repeatedly undo. Graph should return to a non-broken state.


LandonTownsendUnity:
I'm getting this compile error
`E:\Unity Repositories\Graphics\com.unity.render-pipelines.high-definition\Editor\Wizard\HDWizard.Window.cs(384,36): error CS1955: Non-invocable member 'HDWizard.ToolbarRadio.radioLength' cannot be used like a method.`

I tried on both a20 and b3 and it happens on both. It doesn't happen on master.

cdxntchou (approved):

ghost (approved):
Tested right click, middle click, and escape key. All seems to work fine now.

marctem (approved):

PR #1891, opened 2020-09-15
[10.x.x] Adding Rendering Layer Mask to the Render Objects renderer feature

ellioman:
# Purpose of this PR
[Users have asked](https://forum.unity.com/threads/renderobjects-rendere-feature-is-missing-support-for-rendering-layer-mask.867391/#post-6291881) to be able to use the **[Rendering Layer Mask](https://docs.unity3d.com/ScriptReference/Renderer-renderingLayerMask.html)** in the **Render Object** renderer feature.

This PR exposes the rendering layer mask and which gets used when calling `context.DrawRenderers()`. URP Reserves 20 layers for internal usage and users are able to use the remaining 12.

image

# TODO
- Update the documentation with information about the Render Objects feature (Likely done in a separate PR)

# QA
### What has been tested?
- Tested rendering various objects with different rendering layer masks with the RenderObjects feature, using the 126_RendererFeatureActive test scene
### What needs testing?
- Make sure the feature works with multiple types of renderers and renderer features
- Make sure it works with Camera Stacking and multiple cameras

# Automated Tests
Test scene 126 has been updated like so:
- All the renderers now only render to layer mask 25
- Two of the RenderObjects features (one enabled and one disabled) only render objects in layer 25
- A new enabled RenderObjects feature was added, that renders every layer except layer 25.

The test passes on my machine but is currently disabled because of instabilities in our framework. I will look into that in a followup PR so we can enable it.

# Yamato

### DIFF
--- a/TestProjects/UniversalGraphicsTest/Assets/Scenes/126_RendererFeatureActive.unity
+++ b/TestProjects/UniversalGraphicsTest/Assets/Scenes/126_RendererFeatureActive.unity
@@ -266,6 +266,7 @@ MonoBehaviour:
   m_StopNaN: 0
   m_Dithering: 0
   m_ClearDepth: 1
+  m_AllowXRRendering: 1
   m_RequiresDepthTexture: 0
   m_RequiresColorTexture: 0
   m_Version: 2
@@ -345,6 +346,7 @@ Light:
   m_UseColorTemperature: 0
   m_BoundingSphereOverride: {x: 0, y: 0, z: 0, w: 0}
   m_UseBoundingSphereOverride: 0
+  m_UseViewFrustumForShadowCasterCull: 1
   m_ShadowRadius: 0
   m_ShadowAngle: 10
 --- !u!4 &1014564737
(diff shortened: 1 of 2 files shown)


phi-lira (approved):

phi-lira:
LGTM. If we are adding to the RenderObjects features, shouldn't we also add to the renderer opaque and transparent masks to match?

ellioman:
> LGTM. If we are adding to the RenderObjects features, shouldn't we also add to the renderer opaque and transparent masks to match?

Good point! ==> #1954

kamyker:
Tbh I'm surprised that nothing happened to original Layer Masks in srp that should be separated from physics system. https://forum.unity.com/threads/separate-physics-and-rendering-layers.505580/

pbbastian:
@kamyker Can you elaborate on this? The rendering layer mask is an SRP thing that is separate from the physics system.

kamyker:
> @kamyker Can you elaborate on this? The rendering layer mask is an SRP thing that is separate from the physics system.
These layers: img
We were told something is going to be done about it in 2018 but still nothing happened:
https://forum.unity.com/threads/separate-physics-and-rendering-layers.505580/

laurentopia:
> I will see if I can raise this a bit. Will need some discussion with the team. The main point is that there's a good possibility that URP will need some of these layers and we want to avoid breaking projects at all costs.

Working in URP v8.2 I see that all the rendering layers are available, Camera culling can be set per rendering layer which I was in the process of using for culling out non important objects.  Will v10 or v11 break project if I make use of layers 0-19?

Verasl (changes requested):
Needs more discussion on implementation.

ellioman:
Currently work is being done to introduce light layers to URP that has similar mechanism as HDRP, where you decide between using culling mask or rendering layers. That work makes this PR redundant so I'm closing it.

prodigga:
I'm a little confused as to why this light layers functionality cancels out this pull request. So we would put mesh renderers into *Light* layers instead of rendering Layers, and specify the light layers we wish to render in Render Objects renderer feature with material/stencil overrides etc? (Our end goal might have nothing to do with lighting) So will this Render Objects feature get support for light layers instead?

Kasper-LuGus:
Similarly concerned. From the name it sounds like these Light Layers are intimately tied to lighting. That feels as though we're solving the issue of bitmasks being used for a variety of unrelated purposes with a solution that immediately recreates the problem again? Are we seeing this wrong?

kaychang-unity:
The new light layers (using same mechanism as HDRP) use 8 bits of the rendering layers to indicate which lights influence which meshes. In HDRP, an additional 8 bits of the rendering layers are used for decals, and currently 16 bits are not used (reserved for future use). An important point is that the UI and mechanism will be the same as HDRP so that it will be easier to move between the 2 SRPs.

prodigga:
Great though this doesn't answer the concerns. So the new lighting layer functionality will operate within the 20 Rendering Layers reserved for URP. For the remaining 12, we can use it for our own effects (again, potentially nothing to do with lighting.). Render Objects render feature is a great tool to help facilitate that. For example, outline effects like here:

https://learn.unity.com/tutorial/custom-render-passes-with-urp

In this tutorial, the classic/old layer mask system is used.

We want to use the Rendering Layer instead.. In this way we can have a Rendering Layer we designate as "outline effect", another as "occlusion effect", etc. Then a mesh renderer can be part of one or multiple effects. With Layer Masks, this is not possible, since Mesh Renderer has to be part of only one Layer.

Hopefully this explains our confusion. Doesn't sound like Light Layer functionality has anything to do with this PR and it does not explain its closure. Maybe it was not clear what some of us in the community wanted to achieve with this feature, hopeful this explanation makes it clear. Also, see this thread for more information and feedback:

https://forum.unity.com/threads/renderobjects-rendere-feature-is-missing-support-for-rendering-layer-mask.867391

kaychang-unity:
> We want to use the Rendering Layer instead.. In this way we can have a Rendering Layer we designate as "outline effect", another as "occlusion effect", etc. Then a mesh renderer can be part of one or multiple effects. With Layer Masks, this is not possible, since Mesh Renderer has to be part of only one Layer.
I think you'll be able to implement this using the new interface (light layers and rendering layers as exposed in HDRP).

I think the reason why this PR is closed is because it overlaps with the light layers feature, which also uses and exposes rendering layers but with HDRP interface.

prodigga:
Thanks for trying to explain this and get it resolved! Though, after all this talk _I am still not sure_ if this means I'll be able to render objects on Rendering Layers of my choosing using the Render Objects render feature... 

Is this functionality being scrapped completely? Or will it be re-introduced in another PR in the future..? Or is it that the new Light Layer changes make this functionality impossible/infeasible? I am still trying to figure out why those changes make this PR _"redundant"_? Though I don't use HDRP so maybe I am missing something painfully obvious. I am a URP customer and have never touched HDRP. [We develop mobile games. ](https://play.google.com/store/apps/details?id=com.pandaarcade.picotanks)

@ellioman are you able to please comment on this so we aren't all guessing why this was closed? Tagging you directly as I imagine you are no longer subscribed to this thread since its closure.

prodigga:
Any update on this? Please?

kaychang-unity:
I opened a draft PR for URP light layers: https://github.com/Unity-Technologies/Graphics/pull/3677
Still early.

prodigga:
Yes, great, and an exciting feature for sure. but could someone clear up what any of this have to do with this PR? Why does the addition of light layers cause this PR to become redundant?

Kasper-LuGus:
Just want to add some quick context since prodigga and myself are both quite insistent on this. Just in case we give off that impression, this is not a small personal pet peeve. All this is part of the much bigger topic of freeing ourselves from the [one-purpose-fits-all gameobject layer](https://forum.unity.com/threads/separate-physics-and-rendering-layers.505580/). This affects many projects, regularly. I think there's barely been a project at [my company](https://www.lugus-studios.be/) where that hasn't been a problem or consideration. "Ugh, layers" is a running joke at this point.

There's no easy, engine-wide way to fix this. IMO, the best bet is asking dev teams of feature sets (like SRP/URP/HDRP) to take their architectural responsibility and keep their layers/bitmasks reasonably self-contained. Re-using some existing one might not create problems in an isolated test case, but in a realistic project it soon clashes with other feature sets that went down the same path.

The Light Layer sounds like a step in the right direction. If you want to have light affect different objects based on a bitmask, then we should have a bitmask specifically for that. But given the context above, it should be obvious why the assertion that the linked PR invalidates this one instantly raises some red flags. Surely, the argument cannot be: "We have chosen the clean solution of creating a separate bitmask for purpose X. Incidentally, it will also be used for unrelated purpose Y"?

phi-lira:
@prodigga @Kasper-LuGus Thanks for sharing your feedback and concerns. I've talked with a few devs and let me clarify a few things. The light layers PR will not replace the need for supporting filtering objects in the Renderer and RenderObjects feature. It does use the same `rendering layer` exposed in the mesh renderer to filter objects being draw with `DrawRenderers` thus why the original confusion.

This PR is far from ready to land. Culling/Filtering of renderers is not unrelated to light culling, on the contrary, they are very relatable. Even today with the current implementation to filter objects using the old layer system has some problems in it, f.ex, how to filter shadow casters and how to handle filtering and depth prepass with RenderObjects. 

Depth prepass might affect several things related to lighting (shadows might be resolved in screen space, ambient occlusion, decals, depth information could be used for a forward+ renderer https://portal.productboard.com/unity/1-unity-engine-graphics/c/261-forward-renderer-support)

We need to properly consider how we are going to expose rendering layers for filtering objects. Given the amount of things that we are working on for Unity 2021.2 and I will say that this is not something we can prioritize for that version.

kamyker:
> This PR is far from ready to land. Culling/Filtering of renderers is not unrelated to light culling, on the contrary, they are very relatable.
But physics layers are completely unrelated to light or renderers.

prodigga:
Hey @phi-lira , thanks for replying!

So from what I understand this PR will be closed and this feature will be shelved until your team figures out how expose Rendering Layer filtering [(changes to FilteringSettings incoming?)](https://docs.unity3d.com/ScriptReference/Rendering.FilteringSettings.html) in a way that works well with lighting/shadows/etc. Is this conclusion correct?

XanNava:
Has there been any progress on the FilteringSettings so this could be re looked at?

PR #1897, opened 2020-09-16
[ON HOLD] Remove planar and triplanar mapping shader keyword (use property instead)

FrancescoC-unity:
As in the title, all tests are green when run locally.  Do we need manual validation anyway on this? 

This removes 5 more keywords from the Lit of shaders.


FrancescoC-unity:
There are some potential concerns that I did not consider, so will need to look at the disassembly, in the meantime let's put on hold.

sebastienlagarde:
clsoing for now

PR #1901, opened 2020-09-16
Screen Space Ambient Occlusion for Shadow Matte

anisunity:
On user request, we added the support of SSAO for shadow matte Unit Shader Graph.

image

Testing status
The HDRP_Tests were ran locally and are green.
Added a test to cover this use case.
Made sure the normal was not exported in the the case shadow matte is not checked.
Made sure the AO is not read in case shadow matte is not checked.


iM0ve (approved):
Testing looks sufficient, an automated test added. 
I also double checked it on Mac:
* Before PR objects with shadow matte unlit graph could not be affected by AO
* After PR they are affected by AO
* If Shadow Matte is unchecked in shadergraph, AO no longer has any effect on the gameobject

sebastienlagarde (approved):

sebastienlagarde:
Note: I have updated all the screenshots but vulkan ones don't work :(

PR #1903, opened 2020-09-16
[ShaderGraph] [bugfix 1241602] Sample2DLOD node will compile on GLES2 devices, but always returns opaque black

cdxntchou:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1241602/

SampleTexture2DLOD is not supported on some devices (and will fail shader compilation).
This change causes the node to return opaque black as the replacement output on devices that do not support the instruction.

---
### Testing status

Tested that the node still works in ShaderGraph and in scene (PC DX11 editor).

image

Setting my graphics device to OpenGLES2, it correctly falls back to opaque black:

image

---
### Comments to reviewers


alindmanUnity (approved):
UX is good, black is the fallback for other invalid nodes as well

ghost (approved):
Just wanted to make sure that the node still works correctly in GLES3. Otherwise, developer testing is sufficient.

cdxntchou:
Yamato all green except VFX HDRP Win DX11 which is also failing on master

PR #1905, opened 2020-09-16
[ShaderGraph] [bugfix 1257028] Get rid of target bitmask shadow state

cdxntchou:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1257028/

The core of the issue was that active Targets were stored in two locations: as a list of active targets (serialized) and as a bitmask that flagged the active targets in the list of valid targets.  Since the bitmask was not serialized, it does not update itself automatically with undo/redo.  In general having state stored in two locations and having to keep them in sync everywhere is fragile, so I changed it so we only store the list of active targets.

Merging this with a change Alex was working on to replace the bitmask UI with a reorderable list UI, we got rid of the bitmasks altogether, which also removes the limit of 32 Targets.

This change also stops the list from being reorderable; the active Target list is always sorted in alphabetical order, so there's no reason to give the user the impression that they can reorder it.  Originally the reordering was introduced to be able to place VFX Target first, as the VFX Target required it.   However, that requirement is no longer necessary, all of our existing Targets do not care about order, and we force-sort the active Target list on deserialization.  After discussion with the team, we decided it should not be user reorderable, so I removed the UI ability to do so, and also refactored a bit of the Active Target code to help ensure it stays in sorted order.  We determined that if future Targets needed ordering, we could create a source-code-level interface to control ordering, which would be better than a UI control.

This PR also makes a few improvements to the handling of unknown Targets.  Their name is now displayed as "ClassName (Unknown)", and a nice explanation text is added to their UI.  When an unknown Target is made inactive, we no longer immediately remove it from the potential Target list, so the user can add it back if they want.  Since we save the Target settings for potential Targets, if the user removes and adds back an Unknown Target, all of that Targets serialized settings will be unchanged.

image

image

This change also ensures that removed Target data is retained (at least until the ShaderGraph window is closed or there is an Assembly Reload).  So whenever a Target is removed, the users can add it back and retain their old settings.  Checked that this works with Undo/Redo as well.

---
### Testing status

Added two automated tests to check expected Target invariants.

Ran repro case, tested undo/redo/assembly reload with modifying target list.

Tested adding and removing SRP packages, to see that targets become marked as unknown properly, and their data is retained if they are removed and re-added, undo/redo, even while unknown.

Tested mixes of add/remove/undo/redo, to make sure all combinations retain Target settings.

---
### Comments to reviewers


com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -1937,21 +2060,26 @@ void UpgradeFromBlockMap(Dictionary<BlockFieldDescriptor, int> blockMap)
...
-                    // On ShaderGraph side we dont know what Targets exist so make no assumptions
-                    if(masterNode != null)
+                    // Try to upgrade all potential targets from master node
+                    if (masterNode != null)
                     {
-                        foreach(var target in m_ValidTargets)
+                        foreach (var potentialTarget in m_AllPotentialTargets)
                         {
-                            if(!(target is ILegacyTarget legacyTarget))
+                            if (potentialTarget.IsUnknown())
```
elizabeth-legros, line 2068: are these equivalent? It feels like the ! legacy target catches more cases

elizabeth-legros (approved):

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
cdxntchou: The corresponding !legacy target is below -- github's diff is not lining up the comparison correctly.

Basically we should never get an unknown target in m_AllPotentialTargets at this point ,as we're coming off of a deserialization, m_AllPotentialTargets should just be the reflection-initialized set of known Targets.   I even assert that is the case below...   we could just copy that assert up here as well and drop this check.

cdxntchou:
Katana all green except for known failure in master on Build Universal Win DX11

ghost (approved):
Tested undo/redo, adding/removing targets, and opening a shadergraph with all targets with HDRP or URP missing from the project

PR #1909, opened 2020-09-17
Fix compile errors of SSGI on PS4

FrancescoC-unity:
Cannot call .Load on RWTextureArrays on ps4 apparently.


anisunity (changes requested):
The second texture was supposed to be a read-only resource. i'll change that

anisunity:
pushed the second part of the fix

anisunity (approved):

sebastienlagarde (approved):

PR #1911, opened 2020-09-17
Default contact shadow LIGHT quality setting set to true on High

FrancescoC-unity:
"Fix" for https://fogbugz.unity3d.com/f/cases/1276693/ 

However I **strongly** believe this setting should be deleted; having three choices to set one checkbox feels wrong and moreover this specific case is very confusing as we have a **volume** quality setting named the same.
But this is going to be part of the larger scalability setting discussion (Already mentioned to Julien).

**What have I tested** tested creating a new asset and see how the checkbox for contact shadows for High is set to true.


iM0ve (approved):
I agree with what Francesco said about the setting being unnecessary.

Not retesting manually since all the PR does is changes the default value for contact shadow checkbox.
Also, no code changed, just new code. Should be no regressions.

sebastienlagarde (approved):

PR #1912, opened 2020-09-17
[HDRP][Compositor] Fix camera stacking for AOVs in the graphics compositor

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1273223/

Camera stacking in the graphics compositor was not working with AOVs because the debug view RTs were always cleared.
This PR does not clear the debug view RT when the camera is stacked on top of another one.

9800_Compositor test scene when selecting Normals in the AOV output of the first layer: 
Before (the two first sub-layers  of the first layer are missing):
image

After (the sub-layers of the first layer are visible and disaplay the normal AOV):
image

---
### Testing status
- Compositor test is green
- Tested in 9800_Compositor test scene, by selecting Normal AOV in the first layer, got the expected result. Also tested by selecting different AOVs in the second layer (the sphere), everything was working as expected.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipelineResources/HDRenderPipelineResources.asset
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipelineResources/HDRenderPipelineResources.asset
@@ -85,6 +85,8 @@ MonoBehaviour:
       type: 3}
     copyDepthBufferPS: {fileID: 4800000, guid: 42dfcc8fe803ece4096c58630689982f, type: 3}
     blitPS: {fileID: 4800000, guid: 370f7a9cc4e362d488af024d371091e8, type: 3}
+    blitColorAndDepthPS: {fileID: 4800000, guid: c6e57f5bdbd2a284a86a3097c03884c8,
+      type: 3}
     downsampleDepthPS: {fileID: 4800000, guid: 67d6171b0acc6554aad48c845ec7e67f, type: 3}
     upsampleTransparentPS: {fileID: 4800000, guid: 2ad7ce40f0dbaf64dadef1f58d8524d3,
       type: 3}


remi-chapelain (approved):
Confirmed this fix testing with our Full Test Pass project Compositor scene ✔️

JulienIgnace-Unity (changes requested):
This needs to be checked with Render Graph enabled.
See CreateColorBuffer in HDRenderpipeline.RenderGpaph.cs where the clear is setup for the color buffer.

pmavridis:
> This needs to be checked with Render Graph enabled.
> See CreateColorBuffer in HDRenderpipeline.RenderGpaph.cs where the clear is setup for the color buffer.

Indeed the RenderGraph implementation is missing. I'm still working on this and I will update the PR when it's done.

pmavridis:
I just pushed the RenderGraph implementation and also did the following testing (since QA is away):
- compositor test is green with both RenderGraph ON and OFF
- AOV stacking works with both RenderGraph ON and OFF (as shown in the screenshot at the PR)
- I also fixed some other minor issues that we did not catch before (toggling visibility of AOV layers was not working).

As I mention in the code comments, this is not the most optimal implementation (we are doing an extra blit when stacking is active in debug view) but is perhaps the most pragmatic one. Ideally we could avoid the blit by re-using the debug buffers of the previous camera in the stack, but this would require significant design changes in the compositor (and it is not clear how to switch to Tex2DArrays from Tex2D and hide this from user-facing composition graphs).

sebastienlagarde:
will merge this one. But we should add a more compositor test (and with this case of this PR) to have better coverage. Can be done in another PR

sebastienlagarde (approved):

PR #1914, opened 2020-09-17
Fix a null ref in the blackboard window

adrien-de-tocqueville:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1277989/

Fixed an issue where setting a name at creation for a property in the blackboard would trigger an error.

---
### Testing status

Tested with the repro steps given in the fb


alelievr (approved):

marctem (commented):
Thanks for confirming the crash is resolved. Did you also ensure that the newly-changed name appears in the Graph Inspector? That's what the `m_inspectorUpdateTrigger()` handles, so we need to ensure that _is_ getting called eventually.

adrien-de-tocqueville:
Well since it is at creation I don't think it really needs to update anything. In any case, the name seems to appear correctly everywhere

ghost (approved):
Confirmed that changed name appears in graph inspector.

sebastienlagarde (approved):

PR #1916, opened 2020-09-17
Fix double addition to sorted volume lists 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1249536/ 

Upon registering it is possible (especially if user tries to register manually) to perform a double addition to the sorted light lists. 

If added multiple times it caused upon deletion of the volume, only one copy to be removed and having another one still there.

This change adds stuff to the list only if it wasn't already there, which regardless seems like a good thing to do anyway. 

**What did I test** the repro case, and run hdrp tests


JulienIgnace-Unity (approved):

iM0ve (approved):
Testing should be sufficient for the change

sebastienlagarde (approved):

PR #1921, opened 2020-09-17
Added the possibility to replace or overwrite already existing local package when installing a new one.

JulienIgnace-Unity:
### Purpose of this PR
Currently, there is a workflow issue with the local configuration package:
- Install HDRP version X
- Install local config package
- Upgrade HDRP to version Y with package manager => config package is now the "stock" Y version too.
- Use Wizard to install local config package again => current content of the local package is kept as is.

While this is the expected behavior (as we want to not overwrite locally modified files), the fact that current content is silently kept is confusing for users and can also lead to a version discrepancy between the local config package and the main HDRP package (possibly leading to crashes).
To solve that, if  the local package is already present (but not currently used after an upgrade), we've added a dialog box asking users whether they want to keep the content or not, also warning them of the consequences of either choice.

This is a partial fix for issue https://fogbugz.unity3d.com/f/cases/1269654/
Partial because this one involved DXR in particular. As we are not required anymore to have a local config package for DXR anymore in latest package versions, we are not going to fix it explicitely.

Testing:
This one is difficult to test because it requires full package workflow which cannot be attained with a local repository.
To emulate that I did the following using my local dev repository:
- Install local package via wizard
- Change manifest to go back to the one from the repository.
- Comment out code in the wizard that prevents to reinstall it (because it's based on info given by packman which we don't have).
- Try both Replace/Keep to check that the content is either replaced or kept as is.

This might be tested later when doing the full package release if we want to.


JordanL8 (approved):

TomasKiniulis (approved):
No PR specific issues found. Tested with tarball packages in editor, the HDRP Wizard the popup "Keep", "Replace" option is working as expected when clicking "Install Configuration Editable Packages"

RSlysz (approved):

sebastienlagarde (approved):

PR #1922, opened 2020-09-17
[11.x.x] Soft Particle bugfixes. Cases 1162556, 1256953& 1226288 + 1268079

ellioman:
# Purpose of this PR
This PR fixes these three issues:

1. [[LWRP] UNLIT PARTICLES SHADER IS NOT RENDERED WHEN SOFT PARTICLES ARE ENABLED ON BUILT APPLICATION](https://issuetracker.unity3d.com/issues/lwrp-unlit-particles-shader-is-not-rendered-when-soft-particles-are-enabled-on-built-application)
2. [URP TRANSPARENT PARTICLE SYSTEM IS NOT VISIBLE / DISAPPEARS WHEN 'DEPTH TEXTURE' IS OFF AND 'SOFT PARTICLES' ARE ENABLED](https://issuetracker.unity3d.com/issues/urp-transparent-particle-system-is-not-visible-slash-disappears-when-depth-texture-is-off-and-soft-particles-are-enabled)
3. [[URP] DEPTH TEXTURE SAMPLE NODE DOES NOT USE CORRECT TEXTURE IN SOME FRAMES](https://issuetracker.unity3d.com/issues/urp-depth-texture-sample-node-does-not-use-correct-texture-in-some-frames)
4. [[URP] OBJECTS THAT ARE USING SOFT PARTICLES ARE RENDERED OPAQUE WHEN OPENGL IS USED](https://issuetracker.unity3d.com/issues/urp-objects-that-are-using-soft-particles-are-rendered-opaque-when-opengl-is-used)
 

# How?
### Issue 1, 2 & 3
The solution is to assign a black or white texture as the depth texture, depending on whether we're on a platform with reverse-z or not, if the frame doesn't have a prepass or copydepth pass.

We now also display a warning message in the material inspector when soft particles are enabled but depth texture is not in the asset.

### Issue 4
The fix is to use a different `LinearEyeDepth()` method inside `Particles.hlsl`. Instead of using the one that uses `_ZBufferParams` it now uses the one that requires world position and view matrix, which is recommended if we've already computed the world space position.

This ended up being a larger fix as I needed to get access to the world position and instead of simply adding that parameter to the SoftParticles function, I decided to refactor some of the code, create a new struct (`ParticleParams`) with the common variables used in our particle shaders and create new functions using that struct. That allows us to modify the functions implementations in the future without having to change the API hopefully again. 

As all of the particle shaders used the same attributes and varyings, except the editor ones, I decided to move those in to a separate hlsl file to reduce duplicate code.

# QA
### What has been tested
- Tested the three projects sent with the bug reports on Metal and OpenGL in editor
- Tested them with Lit, SimpleLit and Unlit particle shaders
- Compiled the particle shaders and made sure we don't have any errors or warnings
- Ran URP Graphics Tests locally

### What needs testing
- Soft Particles with and without depth textures
- Different Graphics API's (GLES, Vulkan, etc) on builds as I've only tested in editor

# Yamato
master: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fsoft-particle-fixes
2020.2: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fsoft-particle-fixes/.yamato%252F_abv.yml%2523all_project_ci_CUSTOM-REVISION/3977446/job

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
@@ -430,8 +430,8 @@ namespace UnityEngine.Rendering.Universal
                 EnqueuePass(m_CopyDepthPass);
             }
 
-            // Set the depth texture to the far Z if we do not do a depth prepass or copy detph
-            if (!requiresDepthPrepass && !requiresDepthCopyPass)
+            // For Base Cameras: Set the depth texture to the far Z if we do not have a depth prepass or copy depth
+            if (cameraData.renderType == CameraRenderType.Base && !requiresDepthPrepass && !requiresDepthCopyPass)
             {
                 Shader.SetGlobalTexture(m_DepthTexture.id, SystemInfo.usesReversedZBuffer ? Texture2D.blackTexture : Texture2D.whiteTexture);
             }


com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -115,7 +115,7 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
                 cmd.SetGlobalVector(ShaderPropertyId.scaleBiasRt, scaleBiasRt);
 
                 cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, m_CopyDepthMaterial);
-
+                cmd.SetGlobalVector(ShaderPropertyId.availableTexturesParams, new Vector4(0f, 0f, 0f, 1f));
```
ellioman, line 118: This is probably not a good enough solution if/when we want to add other texture types to this parameter, such as the opaque texture. Could we add the data to RenderingData? Other ideas welcome.

com.unity.render-pipelines.universal/ShaderLibrary/Particles.hlsl
```diff
@@ -8,6 +8,37 @@
 #include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/DeclareOpaqueTexture.hlsl"
 #include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/ParticlesInstancing.hlsl"
 
+struct ParticleParams
+{
+    float4 positionWS;
+    float4 particleColor;
```
phi-lira, line 14: is this vertex color? if so, maybe name it like that for better clarity.

com.unity.render-pipelines.universal/ShaderLibrary/Particles.hlsl
```diff
@@ -8,6 +8,37 @@
...
+    float4 projectedPosition;
+    half4 color;
```
phi-lira, line 16: baseColor for better clarity?

com.unity.render-pipelines.universal/Shaders/Particles/ParticlesInput.hlsl
```diff
@@ -0,0 +1,75 @@
+#ifndef UNIVERSAL_PARTICLES_INPUT_INCLUDED
+#define UNIVERSAL_PARTICLES_INPUT_INCLUDED
+
+#if PARTICLES_EDITOR
```
phi-lira, line 4: Why is it that we need different attributes/varyings for editor?

phi-lira (approved):
LGTM. We should ask for cross pipeline team / HDRP on what is the design regarding checking if a texture is available in the shader.

phi-lira:
When testing, please compile shaders in editor to see if there are no error / warnings in any platform.

com.unity.render-pipelines.universal/ShaderLibrary/Particles.hlsl
ellioman: Good point! Renamed to vertexColor.

com.unity.render-pipelines.universal/ShaderLibrary/Particles.hlsl
ellioman: Yup, renamed to baseColor.

com.unity.render-pipelines.universal/Shaders/Particles/ParticlesInput.hlsl
ellioman: I don't know. We need to put that in our task list to look at that.

richardkettlewell: This smells very strange :)

If i'm reading it correctly, you are saying lighting is totally disabled/broken in the Editor (because you disregard the normal + tangent inputs)

I don't think you want to ship this as-is unless it's already in this broken state in master.

ellioman: I could fix this in this PR as well but this how Editor Particles are currently on Master. I'm just moving the code into one separate file.

https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Shaders/Particles/ParticlesEditorPass.hlsl

richardkettlewell: No this file you linked isn't for rendering of particles in the editor, it's for rendering the editor *specific* passes, namely the selection outline and the picking buffer version.

richardkettlewell: It seems like I should have added a comment in this file explaining this :)
(I added it when I added instanced particle rendering support, which needs custom passes for picking+selection outlines)

ellioman: Ahh!! My bad! 
Wondering what's the best solution here. Would we be fine with moving the define into the relevant passes and keep the attributes/varyings in the same file?

richardkettlewell: Perhaps add a define to the editor passes, then use it to exclude normal + tangents from the inputs?
You can find the passes by looking at any pass that uses the *vertParticleEditor* vertex program.

I was also thinking you could just wrap it in #if !UNITY_PASS_META but i think we use META passes for other stuff that may need those inputs. But im not sure about that. I don't know exactly what qualifies as a meta pass.

richardkettlewell (approved):

erikabar (approved):
Verified bug reports (1162556, 1256953, 1226288, 1268079)
Verified that a message is thrown when Soft Particles are enabled but Depth Texture is disabled
smoke tested Particles FTP project on Mac, Android (Vulkan, OpenGLES3, OpenGLES2) and iOS (Metal)

phi-lira (approved):

phi-lira:
merging to stage branch

PR #1923, opened 2020-09-17
Fixed Cutoff not working properly with ray tracing shaders default and SG (case 1261292).

anisunity:
There was a mismatch between the shadow map and the ray traced shadow in the management of alpha cutoff values. This PR fixes that and makes sure it all works the same.
https://fogbugz.unity3d.com/f/cases/1261292/

image

Testing status
The DXR SSS GI test is failing locally for me even if i didn't touch anything that impacted it. Not sure what is going there.
Tested extesively all the configuration SG and non SG. Added a test that should cover it.


alelievr (approved):

remi-chapelain (approved):
Looks good to me, both shadows are now coherent. 
About the DXR SSS test, it's also failing on my side, but it's also failing on staging so it's probably not related to this PR in particular

sebastienlagarde (approved):

PR #1925, opened 2020-09-17
Fixed an exception thrown when closing the lookdev because there is no active SRP anymore.

JulienIgnace-Unity:
### Purpose of this PR
Fixed an exception thrown when closing the lookdev because there is no active SRP anymore.

https://fogbugz.unity3d.com/f/cases/1262169/

---
### Testing status

Tested the repro provided in the fogbugz case:
Open lookdev, dock it, remove the current HDRP asset from the graphics settings. Notice that no exception is thrown anymore.
Also tested undocked just in case.


JMargevics (approved):
If you remove HDRP asset with docked LookDev window it now just throws "LookDev is not supported by this Scriptable Render Pipeline: No SRP in use" and closes the window. A good and correct behaviour, no other errors. ✅

RSlysz (approved):

sebastienlagarde (approved):

PR #1928, opened 2020-09-17
SSR Improvement

skhiat:
### SSR Improvement

The improvement:
* Uses the VNDF Importance Sampling
* Temporal accumulation controllable with the new parameter 'Accumulation Amount' (0 No accumulation, 1 Accumulation uses multiple frames). More accumulation generate more ghosting
image

Better match:
New SSR:
image

Pathtracer (Bounce == 2):
image

### Checklist for PR maker
[Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).

---
### Purpose of this PR
SSR will give us a better match with Pathtracer.
* Blurriness (blur radius)
* With the adjustment of "Accumulation Amount" will be "override" by the code depending of the motion vector (of the original point and the hit point).

---
### Testing status

On the scene add a moving object for example a cylinder and move it, note the motion vector for objects are only available in the GameView:
`    public float speed = 5.0f;

    // Update is called once per frame
    void Update()
    {
        transform.Rotate(Vector3.right * speed * Time.deltaTime);
    }`

For rough surface, we will have more noise even with large accumulation, on region of this cylinder.
Ref:
https://drive.google.com/file/d/1zyrvVWU-bzVqYOyOERe7OMvnZXtjbxzI/view?usp=sharing

Parameters to consider during a test:
Object thickness, which allow the ray to pass behind thin object:
image
Correct value: scene dependents

Allow more accumulation with multiple frame:
image

---
### Comments to reviewers

### DIFF
--- a/TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXOperatorUtilityTests.cs
+++ b/TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXOperatorUtilityTests.cs
@@ -637,7 +637,119 @@ namespace UnityEditor.VFX.Test
             var operation = (UnityEngine.VFX.VFXExpressionOperation)Enum.Parse(typeof(UnityEngine.VFX.VFXExpressionOperation), expressionName);
             var referenceExpression = VFXBuiltInExpression.Find(operation);
             Assert.IsTrue(VFXDynamicBuiltInParameter.s_BuiltInInfo.Values.Any(o => o.expression == referenceExpression));
+        }
 
+        public struct ApplyAddressingModeTestCase
+        {
+            public ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode _mode, uint _count)
+            {
+                mode = _mode;
+                count = _count;
+                expectedSequence = new uint[Mathf.Max(50, 7 * (int)_count)];
+
+                //Naive implementation for reference
+                if (mode == VFXOperatorUtility.SequentialAddressingMode.Clamp)
+                {
+                    for (uint i = 0; i < expectedSequence.Length; ++i)
+                    {
+                        expectedSequence[i] = i < count ? i : count - 1;
+                    }
+                }
+                else if (mode == VFXOperatorUtility.SequentialAddressingMode.Wrap)
+                {
+                    uint current = 0u;
+                    for (uint i = 0; i < expectedSequence.Length; ++i)
+                    {
+                        expectedSequence[i] = current;
+                        current++;
+                        if (current >= count)
+                            current = 0u;
+                    }
+                }
+                else if (mode == VFXOperatorUtility.SequentialAddressingMode.Mirror)
+                {
+                    uint current = 0u;
+                    bool increment = true;
+                    for (uint i = 0; i < expectedSequence.Length; ++i)
+                    {
+                        expectedSequence[i] = current;
+                        if (increment)
+                        {
+                            current++;
+                            if (current >= count)
+                            {
+                                increment = false;
+                                current = count > 2u ? count - 2u : 0u;
+                            }
+                        }
+                        else
+                        {
+                            if (current == 0u)
+                            {
+                                increment = true;
+                                current = count == 1u ? 0u : 1u;
+                            }
+                            else
+                            {
+                                current--;
+                            }
+                        }
+                    }
+                }
+            }
+
+            public VFXOperatorUtility.SequentialAddressingMode mode;
+            public uint count;
+            public uint[] expectedSequence;
+
+            public override string ToString()
+            {
+                return string.Format("{0}_{1}_{2}", mode.ToString(), count, expectedSequence.Length);
+            }
+        }
+
+
+        static readonly ApplyAddressingModeTestCase[] ApplyAddressingModeTestCase_ValueSource =
+        {
+            //The 0 case is always undefined
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Wrap, 1u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Wrap, 4u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Clamp, 1u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Clamp, 4u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 1u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 2u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 3u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 4u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 7u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 8u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 9u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 13u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 15u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 27u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 32u),
+            new ApplyAddressingModeTestCase(VFXOperatorUtility.SequentialAddressingMode.Mirror, 33u),
+        };
+
+        [Test]
+        public void CheckExpectedSequence_ApplyAddressingMode([ValueSource("ApplyAddressingModeTestCase_ValueSource")] ApplyAddressingModeTestCase addressingMode)
+        {
+            var computedSequence = new uint[addressingMode.expectedSequence.Length];
+            for (uint index = 0u; index < computedSequence.Length; ++index)
+            {
+                var indexExpr = VFXValue.Constant(index);
+                var countExpr = VFXValue.Constant(addressingMode.count);
+                var computed = VFXOperatorUtility.ApplyAddressingMode(indexExpr, countExpr, addressingMode.mode);
+
+                var context = new VFXExpression.Context(VFXExpressionContextOption.CPUEvaluation);
+                var result = context.Compile(computed);
+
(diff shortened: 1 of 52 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
```diff
@@ -59,20 +70,300 @@ TEXTURE2D_X(_DepthTexture);
...
+
+    int globalSampleIndex = _FrameCount;
+    float2 Xi;
+    Xi.x = GetBNDSequenceSample(positionSS, globalSampleIndex, 0);
+    Xi.y = GetBNDSequenceSample(positionSS, globalSampleIndex, 1);
+
+    // Pick which subpixel we will be launching our effects from
+    float subPixelSample = GetBNDSequenceSample(positionSS, globalSampleIndex, 3);
+    int subPixel = clamp((int)(subPixelSample * 4.0), 0, 3);
+    uint2 shift = HalfResIndexToCoordinateShift[subPixel];
```
anisunity, line 193: This is code in RayTracingReflection.compute is meant to handle BSDF data info fetch in half res mode, in the current state SSR doesn't have half res mode. This is a copy paste mistake i guess.

anisunity:
I don't see code that handle the transparent SSR case, it will break.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
skhiat: We switched to an algorithm that does temporal accumulation this shift help for quality. I am open to discuss that.

anisunity: This usage of the code, if on purpose, increases the variance and biases the signal even more, I am sure of that.

skhiat: That help on the edges & very close hit (corners: Wall vs Floor/Ceilling)

eturquin:
- There is no Fresnel absorption here (F0 is always 1, hence F is too), so all occurrences of a Fresnel-related variable/computation should be removed.
- Not quite sure about your pdf computation in the GGX evaluation (which should probably be taken out from GetSampleInfo(...) for more clarity), the Jacobian seems to be missing.
- The copy-pasting of all GGX routines is a little crude, could use some refactor (some functions like SampleGGXVisibleNormal(...) or  Lambda_GGX(...) are virtually identical to the originals. We can discuss where to put them.

skhiat:
> * There is no Fresnel absorption here (F0 is always 1, hence F is too), so all occurrences of a Fresnel-related variable/computation should be removed.
cf. Last PR.

> * Not quite sure about your pdf computation in the GGX evaluation (which should probably be taken out from GetSampleInfo(...) for more clarity), the Jacobian seems to be missing.
 That was done to fit with the previous implementation to have the same behavior (open to discussion)

> * The copy-pasting of all GGX routines is a little crude, could use some refactor (some functions like SampleGGXVisibleNormal(...) or  Lambda_GGX(...) are virtually identical to the originals. We can discuss where to put them.
True

remi-chapelain (commented):
Adding a [test result document](https://confluence.unity3d.com/display/HDRP/SSR+Improvement) here for tracking
Copy pasting the conclusion here for now :

The new algorithm provides a new approach that gives visually better results in most cases. 

The old algorithm might still be a better solution in some edge case, it's good that we still have the choice. 

Performance have improved during testing but is still worse compared to the legacy Approximation one

Comparing Approx and PBR Accumulation, results are only better on rough (non smooth) surface using PBR Accumulation.

On rough surface without any texturing and a low accumulation factor, the image stay noisy. However, with texturing the result can be satisfactory and meet expectations and still be better than what Approximation algorithm does. (see comparisons)

With a moving object, it's really hard to find balance and find a correct accumulation factor that creates satisfactory results without ghosting, it depends a lot of the speed of the objects.

sebastienlagarde:
@remi-chapelain can you ensure that receive SSR flag is working correctly? i.e if one want to remove moving object from SSR, it should be able to do it. thanks

sebastienlagarde:
note: I have retarget the PR to master

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/VertMesh.hlsl
```diff
@@ -1,3 +1,6 @@
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs.hlsl"
```
sebastienlagarde, line 1: Why do you need those hearders here? it is not expected that we put  such header in VertMesh.hlsl

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -4766,13 +4764,19 @@ RenderSSRParameters PrepareSSRParameters(HDCamera hdCamera, in HDUtils.PackedMip
             cb._ColorPyramidUvScaleAndLimitPrevFrame = HDUtils.ComputeViewportScaleAndLimit(hdCamera.historyRTHandleProperties.previousViewportSize, hdCamera.historyRTHandleProperties.previousRenderTargetSize);
             cb._SsrColorPyramidMaxMip = hdCamera.colorPyramidHistoryMipCount - 1;
             cb._SsrDepthPyramidMaxMip = depthPyramid.mipLevelCount - 1;
+            if (hdCamera.isFirstFrame || hdCamera.cameraFrameCount <= 2)
+                cb._SsrAccumulationAmount = 1.0f;
+            else
+                cb._SsrAccumulationAmount = Mathf.Pow(2, Mathf.Lerp(-0.0f, -7.0f, volumeSettings.accumulationFactor.value));
```
sebastienlagarde, line 4770: typo, there is a -0.0 here

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
```diff
@@ -457,19 +465,56 @@ class RenderSSRPassData
...
-                    //passData.hitPointsTexture = builder.WriteTexture(renderGraph.CreateTexture(new TextureDesc(Vector2.one, true, true)
-                    //    { colorFormat = GraphicsFormat.ARGBFloat, clearBuffer = true, clearColor = Color.clear, enableRandomWrite = true, name = "SSR_Debug_Texture" }));
+                    { colorFormat = GraphicsFormat.R16G16_UNorm, clearBuffer = true, clearColor = Color.clear, enableRandomWrite = true, name = transparent ? "SSR_Hit_Point_Texture_Trans" : "SSR_Hit_Point_Texture" });
+
+                    if (ssrVolumeSettings.usedAlgorithm.value == ScreenSpaceReflectionAlgorithm.PBRAccumulation)
+                    {
+                        passData.ssrAccum = builder.WriteTexture(ssrAccum);
+                        passData.ssrAccumPrev = builder.WriteTexture(ssrAccumPrev);
+                        passData.lightingTexture = builder.CreateTransientTexture(new TextureDesc(Vector2.one, true, true)
+                        { colorFormat = GraphicsFormat.R16G16B16A16_SFloat, clearBuffer = true, clearColor = Color.clear, enableRandomWrite = true, name = "SSR_Lighting_Texture" });
```
sebastienlagarde, line 483: this must be of color buffer format (i.e 101010f or 16f depends on the option of hdrp asset)

sebastienlagarde: I see it is not related to your PR, I will take care of it separately

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -4824,17 +4839,61 @@ RenderSSRParameters PrepareSSRParameters(HDCamera hdCamera, in HDUtils.PackedMip
...
+                    cmd.DispatchCompute(cs, parameters.accumulateKernel, HDUtils.DivRoundUp(parameters.width, 8), HDUtils.DivRoundUp(parameters.height, 8), parameters.viewCount);
+
+                    if (parameters.transparentSSR)
+                        CoreUtils.SetKeyword(cmd, "DEPTH_SOURCE_NOT_FROM_MIP_CHAIN", false);
+                }
+            }
+
+            if (!usePBRAlgo)
+            {
+                CoreUtils.SetKeyword(cmd, "SSR_APPROX", false);
```
sebastienlagarde, line 4896: I don't understand why you do this at the end. You must no do this but instead do:

 CoreUtils.SetKeyword(cmd, "SSR_APPROX", usePBRAlgo);

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -4827,17 +4862,61 @@ RenderSSRParameters PrepareSSRParameters(HDCamera hdCamera, in HDUtils.PackedMip
...
                 cmd.SetComputeTextureParam(cs, parameters.reprojectionKernel, HDShaderIDs._SsrClearCoatMaskTexture, clearCoatMask);
                 cmd.SetComputeTextureParam(cs, parameters.reprojectionKernel, HDShaderIDs._CameraMotionVectorsTexture, motionVectorsBuffer);
-                cmd.SetComputeTextureParam(cs, parameters.reprojectionKernel, HDShaderIDs._NormalBufferTexture, normalBuffer);
 
                 ConstantBuffer.Push(cmd, parameters.cb, cs, HDShaderIDs._ShaderVariablesScreenSpaceReflection);
 
                 cmd.DispatchCompute(cs, parameters.reprojectionKernel, HDUtils.DivRoundUp(parameters.width, 8), HDUtils.DivRoundUp(parameters.height, 8), parameters.viewCount);
+
+                if (parameters.transparentSSR)
+                    CoreUtils.SetKeyword(cmd, "DEPTH_SOURCE_NOT_FROM_MIP_CHAIN", false);
```
sebastienlagarde, line 4886: you don't need to reset keyword at the end (unless you reuse the shader, you should setup

CoreUtils.SetKeyword(cmd, "DEPTH_SOURCE_NOT_FROM_MIP_CHAIN", parameters.transparentSSR); at the beginning

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
sebastienlagarde: oh unless it is for the transparent case?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
sebastienlagarde: oh unless it is for the transparent case?

PR #1929, opened 2020-09-17
Fixed UI alignment of framesettings

adrien-de-tocqueville:
Fix for https://fogbugz.unity3d.com/f/cases/1260513/

The background of framesettings headers in HDRP Default Settings was not correclty aligned.

before:
before

after:
after

That's maybe not the cleanest way to fix this but didn't find any other way

### Test Status

The other places in which it is drawn have not changed (Project Settings window and inspectors)


RSlysz (approved):
This is drawn at several location.
I see hardcoded spacing here. Please try the following location:
- in Camera's Inspector
- in HDRP Default settings

RSlysz (commented):
edit review: wrongly approved

RSlysz (changes requested):
Check the two displayed cases

RSlysz (approved):
Checked with Adrien. Both location have been tested

sebastienlagarde (approved):

PR #1931, opened 2020-09-17
9.x.x/sg/context undo block

esmelusina:
Purpose of this PR

Backport of #1889. Address this bug and a few adjacent; https://fogbugz.unity3d.com/f/cases/1257020/

Soft-locking is caused by a null exception in the undo-redo. In many cases this is from the logical and visual hierarchy of block nodes being inconsistent. The other part is from how context menus are built when a stack node is floating. This fix should snap the block node back to its owning stack node instead of attempting to build a bad context menu.

Testing status
Manual testing- drag a block node outside of the stack node container and right click to build a context menu. Then repeatedly undo. Graph should return to a non-broken state.


LandonTownsendUnity (approved):
Confirmed fix for right click, middle click, and escape key.

marctem (approved):

PR #1932, opened 2020-09-17
Fix property references to not use copied guid refs

elizabeth-legros:
The old copy paste system relied on copied properties copying guids as well. Use the copied ShaderInput instead of the original in the CopyPasteGraph.

https://fogbugz.unity3d.com/f/cases/1274306/

Tried copy and pasting just nodes/property nodes, then just shader inputs, then both. 
copying just nodes inlined found properties (should this instead make a new property anyways?)
copying just (blackboard) inputs copied the inputs over
copy both inputs and nodes brought nodes over and hooked up properties correctly


alindmanUnity (approved):

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -1207,7 +1207,7 @@ internal static void InsertCopyPasteGraph(this MaterialGraphView graphView, Copy
                         foreach (var node in dependentPropertyNodes)
                         {
                             node.owner = graphView.graph;
-                            node.property = property;
+                            node.property = (AbstractShaderProperty)copiedInput;
```
marctem, line 1210: Nice catch on these, looks like that `copiedInput` has been unused since it was written last March. =/

I'd prefer to do a little cleanup and get rid of the top-level `copiedInput` entirely, just do `var copiedProperty = DuplicateShaderInputs(...) as ASP;` inside this switch (same for keywords below).

marctem (changes requested):

ghost (approved):
Tested:
Pasting a property with existing reference name but different type (inlines as expected)
Pasting a property with existing reference name and same type (links up as expected)
Pasting several combinations of properties that do or don't link up to properties on the graph they're being pasted on (all works as expected)

marctem (approved):

ghost (commented):
Tested the following for a single graph:
Copy and paste a single property - as expected, duplicates the property
Copy and paste a property and its node - as expected, duplicates the property and the new node is tied to the new property
Copy and paste just the node - duplicates the node but not the property

From graph to another graph
Copy and paste a property that doesn't exist: works
Copy and paste a property and node that don't exist on this branch: copies them over
Copy and paste a property that exists on this graph: pastes as a new property
Copy and paste a node whose property exists on this graph: pastes the node and hooks it up to the existing property
Copy and paste a node whose property doesn't exist on this graph: inlines the node
Copy and paste a property and node whose property exists already: adds the property as a new property  and attaches the node to it (not sure if this is expected but seems right to me)

Also tested:
Subgraphs
Pasting lots of properties and nodes at once

One last issue found (will detail repro steps in next comment)

ghost (approved):
My bad, on second look the other issue wasn't a bug, just me misreading the (1) indicating that it was a one dimensional vector, and thinking it was connected to the duplicate variable with a (1) at the end of its name.

Approving.

PR #1934, opened 2020-09-17
Add decal angle fading + clean tech debt in decal C#

sebastienlagarde:
### Purpose of this PR

This PR add the support of angle based fading for decal when using the Decal Layer feature (as we have access to the vertex normal in this case).

Angle based fading looks like this:
visuel_decal_slide_working

The PR add a test 7512 to check it.

Currently the angle fading is  not linear (as we play with angle) to reduce the ALU cost in shader. In practice it shouldn't be a problem as such usage is mainly to avoid uv stretching on hard surfaces.

This feature only work for projector.

What I tested:
I tested with cluster decal (regualr decal with affect transparent on a transparent material), shader graph decal and regular decal, tested that the feature is correctly enabled/disable based on Decal Layer enabling and non default value.

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalSystem.cs
```diff
@@ -458,13 +458,11 @@ private BoundingSphere GetDecalProjectBoundingSphere(Matrix4x4 decalToWorld)
                 return res;
             }
 
-            public void UpdateCachedData(Matrix4x4 localToWorld, Quaternion rotation, Matrix4x4 sizeOffset, float drawDistance,
-                float fadeScale, Vector4 uvScaleBias, bool affectsTransparency, DecalHandle handle,
-                int layerMask, ulong sceneLayerMask, float fadeFactor, DecalLayerEnum decalLayerMask)
+            public void UpdateCachedData(DecalHandle handle, DecalProjector.CachedDecalData data)
```
JulienIgnace-Unity, line 461: You should pass data as an "in" parameter to avoid unnecessary copies

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalProjector.cs
```diff
@@ -224,6 +262,45 @@ internal DecalSystem.DecalHandle Handle
...
+            public float endAngleFade;
+            public Vector4 uvScaleBias;
+            public bool affectsTransparency;
+            public int layerMask;
+            public ulong sceneLayerMask;
+            public float fadeFactor;
+            public DecalLayerEnum decalLayerMask;
+        }
+
+        internal CachedDecalData GetCachedDecalData()
```
JulienIgnace-Unity, line 283: Is there any way we can actually cache that and not reconstruct every time we need it?  (might not be, not super familiar with the rest of the code)

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalSystem.cs
```diff
@@ -479,34 +477,39 @@ private BoundingSphere GetDecalProjectBoundingSphere(Matrix4x4 decalToWorld)
...
-                UpdateCachedData(transform.localToWorldMatrix, transform.rotation, sizeOffset, drawDistance,
-                    fadeScale, uvScaleBias, affectsTransparency, handle,
-                    layerMask, sceneLayerMask, fadeFactor, decalLayerMask);
-            }
-
-            public DecalHandle AddDecal(Matrix4x4 localToWorld, Quaternion rotation, Matrix4x4 sizeOffset, float drawDistance,
-                float fadeScale, Vector4 uvScaleBias, bool affectsTransparency, int materialID,
-                int layerMask, ulong sceneLayerMask, float fadeFactor, DecalLayerEnum decalLayerMask)
+            // Update memory allocation and assign decal handle, then update cached data
+            public DecalHandle AddDecal(int materialID, DecalProjector.CachedDecalData data)
```
JulienIgnace-Unity, line 512: same here, data should be an "in" parameter

JulienIgnace-Unity (changes requested):
As discussed over slackerd, there's an additional cleanup that can be done on the "current" cached data in the system.

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalProjector.cs
sebastienlagarde: I am not sure, guess when we call this function it usually mean that there is a change (in particular the rotation matrix), but I can investigate)

JulienIgnace-Unity (approved):

sebastienlagarde:
merge for 10.1

PR #1936, opened 2020-09-18
Fix procedural sky with orthographic camera

EvgeniiG:
### Purpose of this PR
Fix case https://fogbugz.unity3d.com/f/cases/resolve/1278013/

---
### Testing status

- [x] Opened test project + Run graphic tests locally
- [ ] Built a player
- Other: Manually tested the procedural sky in a simple scenario.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-orthographic-sky/.yamato%252F_abv.yml%2523all_project_ci_trunk

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -77,6 +77,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix for lookdev toggling renderers that are set to non editable or are hidden in the inspector.
 - Fixed issue with mipmap debug mode not properly resetting full screen mode (and viceversa). 
 - Added unsupported message when using tile debug mode with MSAA.
+- Fix procedural sky with orthographic camera (case 1278013).
 
 ### Changed
 - Preparation pass for RTSSShadows to be supported by render graph.


JulienIgnace-Unity (approved):

TomasKiniulis (commented):
@EvgeniiG it fixes the case submitted issue, but Orthographic projection just displays the Sky in a solid color now. Which is different from what it used to be or what is now in Legacy and URP. Could the fix be done while still keeping the old way of sky displayed?

Attaching comparison results:
**PR Orthographic**
image
**PR Perspective**
image
**8.1.0 Orthographic**
image
**8.1.0 Perspective**
image
**Legacy Orthographic**
image

EvgeniiG:
> @EvgeniiG it fixes the case submitted issue, but Orthographic projection just displays the Sky in a solid color now. Which is different from what it used to be or what is now in Legacy and URP. Could the fix be done while still keeping the old way of sky displayed?

The new way of rendering the sky is correct.
Recall what orthographic projection is - you have the same single view direction for all pixels of the screen. The skybox is very far away, so all pixels end up looking at the same skybox texel.

TomasKiniulis:
> > @EvgeniiG it fixes the case submitted issue, but Orthographic projection just displays the Sky in a solid color now. Which is different from what it used to be or what is now in Legacy and URP. Could the fix be done while still keeping the old way of sky displayed?
> 
> The new way of rendering the sky is correct.
> 
> Recall what orthographic projection is - you have the same single view direction for all pixels of the screen. The skybox is very far away, so the difference in pixel positions does not matter, and all pixels end up looking at the same skybox texel.

Ah right, wish it was the same between all pipelines, but that makes sense

TomasKiniulis (approved):

JulienIgnace-Unity:
Even though it might be mathematically correct, as it stands, it makes the sky in ortho more or less unusable anyway. I think it would be worth exploring other possibilities and see if we can get a usable result.

sebastienlagarde:
Close as we used the fixed deliver in this PR instead: https://github.com/Unity-Technologies/Graphics/pull/2955

PR #1937, opened 2020-09-18
Fixed an issue where entering playmode would close the LookDev window.

JulienIgnace-Unity:
### Purpose of this PR
Fixed an issue where entering playmode would close the LookDev window.

This PR is based on https://github.com/Unity-Technologies/Graphics/pull/1925

Fixes this issue: https://fogbugz.unity3d.com/f/cases/1263108/

---
### Testing status

After opening the lookdev window (with a valid env library and an object loaded to observe proper rendering)
- With undocked window, tried entering/exiting playmode => Window still open
- Same with docked
- Tried opening the lookdev from various environment library assets inspector
- Also tried removing the HDRP asset to check that LookDev would close properly with the right error message.

@Unity-Technologies/gfx-qa-hdrp my testing seems good enough but I'd like a double check since the lookdev has been known to have weird behavior like these before.


TomasKiniulis (commented):
Still getting the following when exiting playmode with LookDev open:
```
NullReferenceException: Object reference not set to an instance of an object
UnityEditor.Rendering.LookDev.Stage.OnBeginRendering (UnityEngine.Rendering.LookDev.IDataProvider dataProvider) (at C:/Users/Tomas/Documents/Github_Projects/Graphics2/com.unity.render-pipelines.core/Editor/LookDev/Stage.cs:247)
```
Can't find any particular patter of this error appearing, but it seems to be relatively common switching between docked and undocked window and entering back to playmode

com.unity.render-pipelines.core/Editor/LookDev/DisplayWindow.cs
```diff
@@ -625,6 +625,19 @@ IStyle GetEnvironmentContenairDraggerStyle()
...
+            // Note: After an editor reload, we might get a null OnUpdateRequestedInternal and null SRP for a couple of frames, hence the check.
+            if (!LookDev.supported && OnUpdateRequestedInternal != null)
+            {
+                // Print an error and close the Lookdev window (to avoid spamming the console)
+                Debug.LogError($"LookDev is not supported by this Scriptable Render Pipeline: "
+                    + (RenderPipelineManager.currentPipeline == null ? "No SRP in use" : RenderPipelineManager.currentPipeline.ToString()));
+                LookDev.Close();
+            }
+        }
+
```
RSlysz, line 640: Isn't this change part of another PR ?

RSlysz (approved):
lgtm

com.unity.render-pipelines.core/Editor/LookDev/DisplayWindow.cs
JulienIgnace-Unity: Yep, it's mentioned in the PR description :) We need to merge them in order.

sebastienlagarde (approved):

PR #1938, opened 2020-09-18
Fix for lens distortion breaking rendering

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1273814/ (HDRP part) 

Lens distortion was considered active even with 0 value, in any case this method of checking if active is equivalent (and more reliable I'd say) and fixes the issue. 

I will open a fix for URP too soon. 

**What I tested / notes for QA**: repro project (though if you could test again would be nice) and verified lens distortion still works (run the tests and tried in a scene various parameters). Note that the repro project is not public, you might need to ask permission to access it (or contact me and I can zip it and send over)


iM0ve:
I think this PR is safe to merge if Distortion automatic test is not breaking. However FrancescoC wanted to perform one more test, we can merge whenever he's ready.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #1941, opened 2020-09-18
Various fixes found while investigating (case 1278465).

anisunity:
Multiple issues were found while investigating the issue. 

The first issues being that because of an optimization (only one sample and no temporal denoising when the surface is perfectly smooth of ray traced reflections), the optimization only makes sense when we only have one bounce.

The tangent and bitangent are not to be used if the surface is not anisotropic, a workaround was found for that.

https://fogbugz.unity3d.com/f/cases/1278465/

- Added a new overridable function that defines the local frame to use for a given material.
- Fixed profiling score for quality RTGI.
- Fixed the denoising and multi-sample not being used for smooth multibounce RTReflections.
- Changed which local frame is used for multi-bounce RTReflections.

Testing status
Ran the DXR tests locally and they are green except SSS RTGI that always fails locally for me.
Tested the scene that was reported in the fogbug

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
+++ b/com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
@@ -106,7 +106,7 @@ namespace UnityEngine.Rendering
         {
             if (cmd != null)
 #if UNITY_USE_RECORDER
-                if (sampler != null)
+                if (sampler != null && sampler.isValid)
                     cmd.BeginSample(sampler);
                 else
                     cmd.BeginSample(name);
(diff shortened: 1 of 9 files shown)


remi-chapelain (approved):
On my side I tested in different mode (quality and perf) and various setup that the issue does not repros, I managed to creates some really minor one but I believe this is expected in those cases. 
had a look at the profiling score to be sure there was no other issue with it.
Tested with material with and without anisotropy to test the new behavior.

Also I added a test un DXR_Tests_Project to be able to prevent a regression (_106_ReflectionsMultiBounceRough_)
**Before the PR**
image
**After the PR**
106_ReflectionsMultiBounceRough

JulienIgnace-Unity (approved):

PR #1942, opened 2020-09-18
Fixed error when changing planar probe resolution

adrien-de-tocqueville:
Fix for https://fogbugz.unity3d.com/f/cases/1278236
The nullref happened when initializing an unused variable so it's an easy fix.

### Testing Status

Doing the repro steps don't throw errors anymore.
Tested changing planar probe resolution and that the preview image correctly updates.


remi-chapelain (approved):
Trusting @adrien-de-tocqueville on this one, but I still wonder how and when this regression was introduced

adrien-de-tocqueville:
I guess it's because the planar probes had no preview image before

remi-chapelain:
Yes, I took a look at the changes and the codes afterwards 👍

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #1944, opened 2020-09-18
Hiding the advanced wheel when ray traced ambient occlusion is enabled on the ambient occlusion component (QOL 1277999).

anisunity:
https://fogbugz.unity3d.com/f/cases/1277999/
The wheel now disappears when ray tracing is checked.

Testing Status:
Tried the combination of enabling override and value on the volume component.

aoAdvanced


remi-chapelain (approved):
Since this is a one-liner, I trust @anisunity on this one and won't do further testing ! 
In slack, we discussed to remove one and for all the bilateral upsample from the advanced parameters (thus getting rid of the issue there) but after discussing with @FrancescoC-unity, it makes some sense to leave it in the advanced section, so approving ✔️

sebastienlagarde (approved):

PR #1946, opened 2020-09-18
Fix master node bug in upgrade

elizabeth-legros:
Master nodes had a bug where the input fragment normal space could mismatch from the master node settings, which would then get copied over to the target. Because of how we now generate some code, this would lead to shader compilation errors with the mismatch. Added code to upgrade path for shader graphs that explicitly checks for this case and fixes on mismatched spaces.

https://fogbugz.unity3d.com/f/cases/1255011/


elizabeth-legros:
Want to verify that this cant happen with targets. If so, Then will need to make a new bug to catch it in 8.x and 7.x and fix the actual bug. If it can happen, Ill fix here and mark for backport back to 7.x

alindmanUnity (approved):

PR #1898, opened 2020-09-16
camera prefab null reference fix

martint-unity:
---
### Purpose of this PR
Fix for this case https://fogbugz.unity3d.com/f/cases/1244430/
Removing camera component in prefab, caused a null reference.

PR #1953, opened 2020-09-21
fix case 1244430 removing camera comp from prefab

martint-unity:
### Purpose of this PR
Added same fix for HDRP as for URP
https://github.com/Unity-Technologies/Graphics/pull/1898

Should also probably be backported.

Fixes this issue
https://fogbugz.unity3d.com/f/cases/1244430/

When removing a camera component from a prefab it gives warnings and errors in console.


RSlysz (approved):
lgtm

madsnyholm (approved):

sebastienlagarde (approved):

PR #1954, opened 2020-09-21
[10.x.x] Adding Rendering Layer Mask to the Forward Renderer and relevant passes

ellioman:
# Purpose of this PR
As a followup of #1891, I'm adding Rendering Layer Masks, for opaque and transparent objects, to the Forward Renderer. 

# Dependencies
#1891 needs to be merged before this one.

# TODO
- Create a test scene in the URP Graphics Project for this feature
- Update the documentation with information about the Render Objects feature (Likely done in a separate PR)

# QA
### What has been tested?
- Tested rendering various objects with different rendering layer masks
### What needs testing?
- Make sure the masks works with multiple types of renderers and renderer features
- Make sure it works with Camera Stacking and multiple cameras

# Automated Tests
A test scene will be added when code reviewers have approved the code changes.

# Yamato


pschraut:
> The main point is that there's a good possibility that URP will need some of these layers and we want to avoid breaking projects at all costs.

Is there any reason you don't use a 64 bits wide data type for the Rendering Layer Mask? Then you could reserve 32 bits for URP and the remaining 32 bits are given to your users.

I believe it's better change it to 64 bits now, than waiting a few years when URP has been widely adopted and would potentially break a lot of projects. Don't do the same mistake that was done with the physics layer mask.

sebastienlagarde:
Hey, just a though Be careful about you will use rendering layer mask.
If you aim to use it for Lightlayers  (like HDRP) or in future for decal layer (like HDRP), then you can't really have a concept of opaque rendering layer mask and transparent rendering layer mask. Of course all depends on the usage. poking @jenniferd-unity for awareness.

Verasl (changes requested):
Needs more discussion on implementation.

ellioman:
Currently work is being done to introduce light layers to URP that has similar mechanism as HDRP, where you decide between using culling mask or rendering layers. That work makes this PR redundant so I'm closing it.

PR #1955, opened 2020-09-22
[10.x.x] Fixed XR depth share in URP forward

thomas-zeng:
# **Please read;**

### Checklist for PR maker
- [ ]  Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).

---
### Purpose of this PR
Fixed XR depth sharing not working properly in URP forward.
- Added depth sharing handling for case where forward renderer renders to intermediate buffers
- Added XR support to CopyDepthPass. 

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers
- CopyDepthPass does not work for MSAA depth if bindMS is false. I have set bindMS to true under some condition for XR case. I think non-XR case should also consider set it to true and do MSAA handling from SRP land. 
- Yflip handling in copyDepthPass is a bit complicate due to source texture's yflip state could change based on previous render pass. The complexity is mainly introduced by a case where backbuffer color surface and intermediate depth surface are used together. I have ensured the case above won't happen in XR by setting createDepth and createColor to the same value. We should consider to set yflip state on a per surface basis in URP or ensure renderTexture and backbuffer won't be used altogether.
See https://docs.unity3d.com/Manual/SL-PlatformDifferences.html where it describes the yflip handling when rendering to RenderTexture.


phi-lira:
Does that mean XR might have two copy depth passes now? one regular copy depth after rendering opaques and one after everything (XrCopyDepth)?

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -396,7 +394,9 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             }
 
 #if ENABLE_VR && ENABLE_XR_MODULE
-            if (cameraData.xr.hasValidOcclusionMesh)
+            // URP uses writes to depth buffer for the XR occlusion mesh pass.
+            // When copy depth is required, we turn off occlusion pass so that depth buffer contains valid data
```
fabien-unity, line 398: Why is this an issue? can you explain a bit more ?

thomas-zeng: Added more comments in 586eabb

The issue is that occlusion mesh pass stores near clip plane value to the ooclusion mesh area in depth buffer.
When depth sharing is enabled, the depth buffer is used in timewarp for reprojection purpose and the near clip plane value in occlusion mesh region causes issue described in https://fogbugz.unity3d.com/f/cases/1242142/

thomas-zeng:
> Does that mean XR might have two copy depth passes now? one regular copy depth after rendering opaques and one after everything (XrCopyDepth)?

That's correct.  XR uses two copy depth passes in URP. One idea I am thinking about is to combine DepthAttachment and DepthTexture together. This could save 1 copy after rendering opaques.

When depth sharing is enabled in XR, XRSDK expects eye texture depth surface to contain valid data.
Before this PR, URP only meets the expectation in direct-render-to-eye case.
After this PR, all cases should be handled.
We do a second blit because current URP draw to DepthAttachment or DepthTexture when depth texture is enabled. And final blit only blits color surface.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
thomas-zeng: As we discusssed, reverted the code at: bd360b6
Forwarded the bug to WMR provider to be handled in provider layer.

phi-lira (approved):

fabien-unity (approved):

PR #1961, opened 2020-09-22
Distortion optimizations

anisunity:
- Added a rough distortion frame setting and and info box on distortion materials (On the Frame settings, disabled by default)
- Added tests to cover the new feature.
Did the rendergraph support.

Tested locally, all the test were green except an unrelated one.
I added one new tests to cover the feature.

Made sure that the rough distortion was unchked by default. 
Made sure that the info message only displayed when distortion was checked.
Made sure when checked the distortion only did the offset part.
Made sure when unchked, the other color pyramid was not built


com.unity.render-pipelines.high-definition/Runtime/Material/GGXConvolution/IBLFilterGGX.cs
```diff
@@ -209,6 +209,15 @@ void CheckIntermediateTexturesSize(int texWidth, int texHeight)
             }
         }
 
+        void OutputSmoothPlanarReflection(CommandBuffer cmd, RenderTexture sourceColor, RenderTexture target)
+        {
+            int currentTexWidth = sourceColor.width;
+            int currentTexHeight = sourceColor.height;
+
+            // The first color level can be copied straight away in the mip chain. Unfortunately due to a format incompatibility, we have to go through a compute shader copy.
```
JulienIgnace-Unity, line 217: Is this comment relevant? seems to me we don't use compute here.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -1248,13 +1250,17 @@ class RenderDistortionPassData
                 builder.SetRenderFunc(
                 (RenderDistortionPassData data, RenderGraphContext context) =>
                 {
+                    if (data.smoothDistortion)
+                        HDUtils.BlitCameraTexture(context.cmd, m_CameraColorBuffer, data.sourceColorBuffer);
```
JulienIgnace-Unity, line 1254: This seems to be a mistake. Looks like it uses the regular path camera color buffer.
(And this will also GCAlloc here). I suppose it was meant to be passed along with the intermediate buffer in the pass data.

JulienIgnace-Unity (changes requested):
There seem to be an implementation mistake in the render graph path.

com.unity.render-pipelines.high-definition/Runtime/Material/GGXConvolution/IBLFilterGGX.cs
anisunity: yup, removed

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
anisunity: fixed

anisunity:
Fixed both  @JulienIgnace-Unity

JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -1133,19 +1149,14 @@ class GenerateColorPyramidData
 
         void GenerateColorPyramid(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle inputColor, TextureHandle output, bool isPreRefraction)
         {
-            // Here we cannot rely on automatic pass culling if the result is not read
-            // because the output texture is imported from outside of render graph (as it is persistent)
-            // and in this case the pass is considered as having side effect and cannot be culled.
-            if (isPreRefraction)
+            if (!isPreRefraction)
```
JulienIgnace-Unity, line 1152: This logic here is becoming too complicated IMO.
We have logic outside the function (cf above RenderDistorsion) and also insde the function.
In term of RenderGraph logic, GenerateColorPyramid should do just what it says it does and all logic calling it or not should be outside. I think this would make things much clearer here (both on caller side, and inside the function)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -198,13 +198,29 @@ class TempPassData { };
...
+                            enableRandomWrite = true,
+                            useMipMap = true,
+                            autoGenerateMips = false,
+                            name = "DistortionColorBufferMipChain"
+                        });
+                    GenerateColorPyramid(m_RenderGraph, hdCamera, colorBuffer, distortionColorPyramid, false);
+                }
+                else
+                {
+                    distortionColorPyramid = currentColorPyramid;
```
JulienIgnace-Unity, line 217: Just to make sure I get it. Here we use the previous pyramid and just read mip 0 for smooth disto right?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -198,13 +198,29 @@ class TempPassData { };
...
+                        {
+                            colorFormat = GetColorBufferFormat(),
+                            enableRandomWrite = true,
+                            useMipMap = true,
+                            autoGenerateMips = false,
+                            name = "DistortionColorBufferMipChain"
+                        });
+                    GenerateColorPyramid(m_RenderGraph, hdCamera, colorBuffer, distortionColorPyramid, false);
+                }
+                else
```
JulienIgnace-Unity, line 215: Also, I think we don't ever use currentColorPyramid after this code so we could just skip the else statement and store result of first branch in currentColorPyramid if needed by rough disto.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -198,13 +198,29 @@ class TempPassData { };
                 }
 
                 // This final Gaussian pyramid can be reused by SSR, so disable it only if there is no distortion
-                if (hdCamera.frameSettings.IsEnabled(FrameSettingsField.Distortion) || hdCamera.IsSSREnabled())
-                    GenerateColorPyramid(m_RenderGraph, hdCamera, colorBuffer, currentColorPyramid, false);
+                TextureHandle distortionColorPyramid;
+                if (hdCamera.frameSettings.IsEnabled(FrameSettingsField.Distortion) && hdCamera.frameSettings.IsEnabled(FrameSettingsField.RoughDistortion))
+                {
+                    distortionColorPyramid = m_RenderGraph.CreateTexture(
```
JulienIgnace-Unity, line 204: One thing to note here.
Before those changes, I think we'd just use a double buffered history pyramid from the hdCamera.
Here we create a new one. Since we have no other (I think) textures with mips in the pipeline, this will actually allocate more memory than before. Can't we just reuse one of the history buffers?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
anisunity: pretty much that is what we should be doing, that or reading directly the color buffer.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
anisunity: done

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
anisunity: discussed, not relevant in this case

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
anisunity: fixed

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/Distortion/ApplyDistortion.shader
```diff
@@ -12,6 +12,8 @@ Shader "Hidden/HDRP/ApplyDistortion"
         #pragma only_renderers d3d11 playstation xboxone vulkan metal switch
         #pragma editor_sync_compilation
 
+        #pragma multi_compile _ SMOOTH_DISTORTION
```
sebastienlagarde, line 15: change for a dynamic branch , no need for more variant here given the cost

JulienIgnace-Unity: I'd argue that this is a pretty simple shader here. Compilation cost is probably meaningless and maintenance complexity is very low so might as well have the most performant code (even if it's just a tiny bit).

sebastienlagarde: I don't think the perf improvement will be visible here, but I think the
 CoreUtils.SetKeyword(context.cmd, "SMOOTH_DISTORTION", data.smoothDistortion);
could easily pollute the global space, so better so keep everything simple for such cases.

sebastienlagarde: side note: use saturate (free) and not clamp (min / max ALU) for clamp(distortionBlur, 0.0, 1.0);

sebastienlagarde: and if you want to avoid branch at all you can do

float mip = (_ColorPyramidLodCount - 1) * clamp(distortionBlur, 0.0, 1.0);

=> float mip = (_ColorPyramidLodCount - 1) * saturate(distortionBlur) * RoughDistortion;
with RoughDistortion == 0 if smooth
it cost just an extra mul

anisunity: All done, relaunching the tests locally and pushing

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2020.1-to-2020.2.md
```diff
@@ -49,6 +49,8 @@ For project migrating from old 9.x.x-preview package. There is a change in the o
 
 From 10.x, the debug lens attenuation has been removed, however the lens attenuation can now be set in the HDRP Default setting Panel as either modelling a perfect lens or an imperfect one. 
 
+From 10.x, the [Screen Space Reflection](Override-Screen-Space-Reflection.md) effect always use the color pyramid generate after the Before Refraction transparent pass. Thus the color buffer only includes transparent GameObjects that use the **BeforeRefraction** [Rendering Pass](Surface-Type.md). Previously the content was dependending on activation of Distortion effect.
```
JordanL8, line 52: From 10.x, the [Screen Space Reflection](Override-Screen-Space-Reflection.md) effect always uses the color pyramid HDRP generates after the Before Refraction transparent pass. This means the color buffer only includes transparent GameObjects that use the **BeforeRefraction** [Rendering Pass](Surface-Type.md). Previously the content depended on whether the Distortion effect was active.

com.unity.render-pipelines.high-definition/Documentation~/whats-new-10-0.md
```diff
@@ -205,6 +205,14 @@ HDRP, being a high-end modern renderer, contains a lot of compute shader passes.
 
 Planar reflection probe filtering is a process that combines the result of planar reflection and surfaces smoothness. Up until this version, the implementation for planar reflection probe filtering did not always produce results of fantastic quality. This version of HDRP includes a new implementation that is closer to being physically-based and improves on the image quality significantly.
 
+### Screen space reflection
+
+[Screen Space Reflection](Override-Screen-Space-Reflection.md) effect always use the color pyramid generate after the Before Refraction transparent pass. Thus the color buffer only includes transparent GameObjects that use the **BeforeRefraction** [Rendering Pass](Surface-Type.md).
```
JordanL8, line 210: [Screen Space Reflection](Override-Screen-Space-Reflection.md) now uses the color pyramid HDRP generates after the Before Refraction transparent pass. This means the color buffer only includes transparent GameObjects that use the **BeforeRefraction** [Rendering Pass](Surface-Type.md).

com.unity.render-pipelines.high-definition/Documentation~/whats-new-10-0.md
```diff
@@ -205,6 +205,14 @@ HDRP, being a high-end modern renderer, contains a lot of compute shader passes.
...
+
+### Distortion
+
+Distortion effect now support a rough distortion effect which is disabled by default. Disabling Rough Distortion allow to save the cost of generating a color pyramid for the effect and instead rely on a copy of the screen.
```
JordanL8, line 214: The distortion effect now supports a rough distortion which is disabled by default. Disabling Rough Distortion saves resources as the effect does not generate a color pyramid and instead uses a copy of the screen.

JordanL8 (changes requested):
Reworded a few bits

PR #1968, opened 2020-09-23
[7.x.x] Fixing nightly issues

ellioman:
# Purpose of this PR
Fixing nightly issues. It does so by making the ColorGradingLutPass.Cleanup() internal and removing the IDisposable and the dispose function from Renderer Features, which got added in a Material Leak fix. 

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/7.x.x%252Ffixing-nightly-issues


ellioman:
The failures on Metal are for Python versions not related to this PR.

PR #1971, opened 2020-09-23
Fixed the Ray-Tracing related Debug Display not working in render graph mode.

anisunity:
In render graph mode, some of the Full Screen Debug /Debug Displays were not supported. This PR makes sure that all modes work the same way they do for non render graph.

Testing Status
Ran the DXR tests locally, they are all green.
Made sure that every menu option displayed the right signal when the target feature was on or off.
Made sure that there were no exception thrown when asking to display the light cluster while not being computed.


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
merging for package 10.1

remi-chapelain:
Double check all the debug mode on Small Office Raytracing project ✔️

PR #1972, opened 2020-09-23
[HDRP][Compositor] Fix outlines in transitions between post-processed and plain regions 

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1278775/

We were getting some undesired outlines when applying selectively color post-process  filters to WorldSpace UI. 

**Why?**
This problem is caused because the transition between the post-processed and plain areas is controlled by the alpha channel. So we have a green version of the image/text (the post-processed one), and a blue one (the non post-processed), and when blending between them, due to the low value of the alpha at the edges, we get the blue outlines.

**How it was solved?**
This is a general problem, indicating that sometimes a smooth transition is desired between the post-processed and plain regions and sometimes a more steep transition is better. The solution I have implemented is to allow the user to control how steep the transition will be, as shown in this gif:
case_1278775
Note: The blue outline disappears as the alpha range is adjusted and the max value approaches 0. 

**Code changes**
- The uber post processing shader now includes an extra FMA when alpha is enabled (cc @FrancescoC-unity )
- Extra code to read the FMA (alpha scale/bias) values from the UI and pass them to the shader
(this includes custom UI code for the min/max slider)

---
### Testing status
- I run the compositor and the DoF-alpha unit tests manually
- I tested with the scene in the bug ticket (also shown in the GiF)

Note: such composition works better when using a second layer (graph composition) as shown in the GIF instead of camera stacking. But still the new control works with and improves camera stacking too.


sebastienlagarde (approved):
Note let's discuss about the othere PR implying alpha for dof and MB

sebastienlagarde:
merge for 10.1

remi-chapelain:
New remapping slider seems to be working fine on my side ✔️

PR #1985, opened 2020-09-24
[HDRP][LookDev] Ensure stylesheet are loaded before applying them on Domain Reload

jenniferd-unity:
---
### Purpose of this PR
When switching from Dark to Light Editor skin, when the LookDev is opened, its layout is completely wrong.
After investigation, this is caused by the stylesheet not being properly loaded before it is applied.
Related case: [case 1278802](https://fogbugz.unity3d.com/f/cases/1278802)

---
### Testing status
Tested manually in the Editor on Windows by switching between Dark and Light mode while the LookDev window was opened. Tested with both Docked and Undocked window mode. 
Tested the upgrade package scenario.
case1278802


RSlysz (commented):
This will not be enough:
OnEnable seams to be sometimes called before AssetDataBase is ready. 
See https://unity.slack.com/archives/C6Y79CZM0/p1599121599123800

jenniferd-unity:
I updated the file and tested the upgrade scenario too on top of the rest

sebastienlagarde (approved):

sebastienlagarde:
Merging for 10.1 package. @RSlysz  please inform us if there is any concern after the last update. Thanks

RSlysz (commented):
Seams better 💚

TomasKiniulis:
Thanks for the nice description! I've checked it on 10.1 package, works great!

PR #1986, opened 2020-09-24
Decal cross fade support

FrancescoC-unity:
Fix for : https://fogbugz.unity3d.com/f/cases/1259395/

The issue with decals looking like they double applied is ultimately due to the fact that they did not support cross fading, hence they applied twice (once per LOD). 

**What have I tested**: the repro case (**NOTE** you need to enable cross fade on the shader graph for the decal in the repro)  and similar thing with a non shader graph decal material.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl
@@ -2,6 +2,7 @@
 #define UNITY_SHADER_VARIABLES_FUNCTIONS_INCLUDED
 
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl"
+#include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Packing.hlsl"
 
 // Helper function for Rendering Layers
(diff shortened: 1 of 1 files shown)


alelievr (approved):

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/Decal.shader
```diff
@@ -78,6 +78,8 @@ Shader "HDRP/Decal"
     #pragma shader_feature_local _MATERIAL_AFFECTS_MASKMAP
 
     #pragma multi_compile_instancing
+    // enable dithering LOD crossfade
+    #pragma multi_compile _ LOD_FADE_CROSSFADE
```
sebastienlagarde, line 82: If I understand the issue correctly this should only be apply with decal mesh right?
or it works with decal projector? if it apply only to decal mesh, then the 
#pragma multi_compile _ LOD_FADE_CROSSFADE
must be move in DecalMeshForwardEmissive and DBufferMesh pass (same for shader graph)

FrancescoC-unity: Done

sebastienlagarde:
merging for 10.1 package

TomasKiniulis:
Repro case testing status looks sufficient. No additional qa testing required

PR #1988, opened 2020-09-24
Fix Custom Post Process affecting preview cameras

alelievr:
### Purpose of this PR
Fix custom post-processing effect affecting previews: https://fogbugz.unity3d.com/f/cases/1277673/

Before:
image
After:
image

The fix is a one-liner that disables custom post processes for all "Preview" camera types.

---
### Testing status
I validated the fix using the ShaderGraph preview as well as a material preview in the inspector.

I didn't test on earlier versions of HDRP but I think the issue exists since we added the custom post processes


FrancescoC-unity (approved):

iM0ve (approved):
The fix looks simple enough. I think the testing is sufficient for the change.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #1989, opened 2020-09-24
Fix box light disappearing when range is below 1 and no range attenuation

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1273188 

The issue is that by setting distance squared to 1 to "disable" attenuation, everything is valid unless the range is less than 1. In those cases we always assume that the light is outside of its range leading to no lighting. 

This fixes the issue, note that even though we still compute distSq because the attenuation scale/bias are set accordingly, there will still be no attenuation. 

**What did I test**: repro project, tried moving the light so that lit object is farther than range (verified there is no light) and closer (verified there is light). Modified range to values smaller than 1 and verify it now all works as expected.

### DIFF
--- a/.yamato/_abv.yml
+++ b/.yamato/_abv.yml
@@ -66,8 +66,6 @@ all_project_ci_nightly_trunk:
     dependencies:
       -  path: .yamato/_abv.yml#all_project_ci_trunk
          rerun: on-new-revision
-      -  path: .yamato/_templates.yml#all_template_ci_trunk
-         rerun: on-new-revision
       -  path: .yamato/_abv.yml#smoke_test_playmode_trunk
          rerun: on-new-revision
       -  path: .yamato/_abv.yml#smoke_test_editmode_trunk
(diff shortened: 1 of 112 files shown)


sebastienlagarde (approved):

sebastienlagarde:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-box-range-when-no-attenuation/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk

sebastienlagarde:
@FrancescoC-unity : this fix broke 2 test: https://yamato-artifactviewer.prd.cds.internal.unity3d.com/463fc3c4-b03e-42ee-9952-4c95c7973764%2Flogs%2FTestProjects%2FHDRP_Tests%2Ftest-results/TestReportV1.html

FrancescoC-unity:
> @FrancescoC-unity : this fix broke 2 test: https://yamato-artifactviewer.prd.cds.internal.unity3d.com/463fc3c4-b03e-42ee-9952-4c95c7973764%2Flogs%2FTestProjects%2FHDRP_Tests%2Ftest-results/TestReportV1.html

I am gonna look into it, but my guess is that that is how they were supposed to look to start with.

FrancescoC-unity:
Fixed the issue, was a typo (y instead of z)

sebastienlagarde:
Merging on the trust of auto test (as they catched issue). Would be nice to still have a bit of QA.

iM0ve:
Looks good. Checked that bug is fixed and it still works when Range Attenuation is checked

box

PR #1990, opened 2020-09-24
[2d] FallOff Texture to Resources.

venkify:
To minimize startup time, we load FallOffTexture from resources, instead of generating on fly.
Note: 
Texture settings sRGB (as its a Lookup) set to Clamp/Bilinear and format set to Uncompressed RGBA8.

# **Please read;**

### Checklist for PR maker

---
### Purpose of this PR

---
### Testing status

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Optimized 2D Renderer performance on mobile GPUs by reducing the number of render target switches.
 - Optimized 2D Renderer performance by rendering the normal buffer at the same lower resolution as the light buffers.
 - 2D Renderer: The per Blend Style render texture scale setting was replaced by a global scale setting for all Blend Styles.
+- Optimized 2D Renderer performance by using a tiny light texture for layer/blend style pairs for which no light is rendered.
 - FallOff Lookup Texture is now part of 2D RenderData.
 
 ### Fixed
--- a/com.unity.render-pipelines.universal/Runtime/2D/Light2DCullResult.cs
+++ b/com.unity.render-pipelines.universal/Runtime/2D/Light2DCullResult.cs
@@ -12,6 +12,7 @@ namespace UnityEngine.Experimental.Rendering.Universal
         public int totalNormalMapUsage;
         public int totalVolumetricUsage;
         public uint blendStylesUsed;
+        public uint blendStylesWithLights;
     }
 
     internal interface ILight2DCullResult
@@ -54,6 +55,8 @@ namespace UnityEngine.Experimental.Rendering.Universal
                     returnStats.totalVolumetricUsage++;
 
                 returnStats.blendStylesUsed |= (uint)(1 << light.blendStyleIndex);
+                if (light.lightType != Light2D.LightType.Global)
+                    returnStats.blendStylesWithLights |= (uint)(1 << light.blendStyleIndex);
             }
 
             return returnStats;
--- a/com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
+++ b/com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
@@ -374,7 +374,12 @@ namespace UnityEngine.Experimental.Rendering.Universal
                 if (!Light2DManager.GetGlobalColor(layerToRender, i, out var clearColor))
                     clearColor = Color.black;
 
-                var identifier =  layerBatch.GetRTId(cmd, rtDesc, i);
+                var anyLights = (layerBatch.lightStats.blendStylesWithLights & (uint)(1 << i)) != 0;
+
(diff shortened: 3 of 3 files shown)


darrenquek (approved):
Did a Performance and Smoke test comparison of URP Released 8.1.0 vs this PR using the Lost crypt project.
Overall there is a slight improvement of performance in the FPS in the project. 
Especially for Low end android.

**LOST CRYPT URP 8.1.0, Unity Editor Version: 2021.1.0a2.727**
On MAC, Unity Editor Playmode: AVG 79FPS
On Windows, Unity Editor Playmode: AVG 210FPS
Mac Build: AVG 60FPS
Android Build(Google Nexus 6): AVG 6FPS

**LOST CRYPT URP PR 1990, Unity Editor Version: 2021.1.0a2.727**
On MAC, Unity Editor Playmode: AVG 81FPS
On Windows, Unity Editor Playmode: AVG 225FPS
Mac Build: AVG 60FPS
Android Build(Google Nexus 6): AVG 12FPS

UnityPaul:
how do we generate the texture?

UnityPaul (changes requested):
not sure how the texture is generated... do we need code to generate it here... thoughts?

venkify:
> not sure how the texture is generated... do we need code to generate it here... thoughts?

Code to generate the texture is in the same file already but commented out ..

UnityPaul (approved):

rainsing (changes requested):
Can we reference the texture in 2D Renderer Data instead of using Resources? That's more consistent with how we reference such resources in URP.

UnityPaul:
> Can we reference the texture in 2D Renderer Data instead of using Resources? That's more consistent with how we reference such resources in URP.

this is actually a really good point... however, what I want is the texture to be builtin to the package. 
1. is that still `Resources` folder? 
2. Even if it is referenced from the Renderer Data, it wouldn't be visible to user. Right?

venkify:
> > Can we reference the texture in 2D Renderer Data instead of using Resources? That's more consistent with how we reference such resources in URP.
> 
> this is actually a really good point... however, what I want is the texture to be builtin to the package.
> 
> 1. is that still `Resources` folder?
> 2. Even if it is referenced from the Renderer Data, it wouldn't be visible to user. Right?

Fixed in https://github.com/Unity-Technologies/Graphics/pull/1990/commits/1ee28dc663cf218e45eac4cd448c330e2681be94

venkify:
> Can we reference the texture in 2D Renderer Data instead of using Resources? That's more consistent with how we reference such resources in URP.

Fixed in https://github.com/Unity-Technologies/Graphics/pull/1990/commits/1ee28dc663cf218e45eac4cd448c330e2681be94

rainsing (approved):

com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/Light2DLookupTexture.cs
```diff
@@ -14,10 +14,14 @@ public static Texture GetLightLookupTexture()
             return s_PointLightLookupTexture;
         }
 
-        public static Texture GetFalloffLookupTexture()
+        public static Texture GetFalloffLookupTexture(Renderer2DData data)
```
UnityPaul, line 17: shouldn't this be the implementation of the getter of `data.fallOffLookup`?

UnityPaul (changes requested):

com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/Light2DLookupTexture.cs
venkify: s_FalloffLookupTexture = data.fallOffLookup;

And like you mentioned in the last PR, if it fails to load the texture, the fallback to create Texture2D is here.

venkify: If you don't need the fallback, no need to access data.fallOffLookup via GetFalloffLookupTexture(). I can remove this function and access it directly wherever its been used. Let me know.

UnityPaul: yes, this pattern is really weird... let's make it so that it is more common...  and we also get to remove yet another static... which is good.

UnityPaul (changes requested):

com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/Light2DLookupTexture.cs
venkify: Got it. Will do it..

venkify: Fixed in https://github.com/Unity-Technologies/Graphics/pull/1990/commits/a49ac4e7e04c1266bf38fccbc3dbd5e750941e7d

UnityPaul (approved):

unity-cchu (approved):

PR #1991, opened 2020-09-24
Better management of max ray length for ray tracing effects (case 1279849).

anisunity:
Added a new volume component parameter to control the max ray length of directional lights
Removed the max clamp value for the RTR, RTAO and RTGI's ray length.
https://fogbugz.unity3d.com/f/cases/1279849/

On user's demand, i changed the way the max ray length was managed for RTR, RTAO and RTGI. It is now clamped to 1mm, but has no maximal clamp value.
I also added a parameter to control the max ray length for directional shadows as it is the only shadow that does not have a range.

image
image
image
image

Testing status
The DXR tests were ran locally and are green.
I made sure that the minimal clamp value was working on all 3 RT effects.
I made sure that the max directional ray length affect both color and monochrome shadows.


sebastienlagarde (approved):

sebastienlagarde:
merging for 10.1 package

remi-chapelain:
Tested on Small Office raytracing to make sure that everything is working properly for the 4 changes. 
Found one minor non blocking issue on the perf mode of RTGI
MaxRayLength is displayed properly but the same value is also displayed under with the name RayLength, this should not be there.

8c71c1e07b57a7cb76baaff1dce6e3b6

PR #1993, opened 2020-09-24
[8.x.x] Fixed a null ref exception when baking probes.

JulienIgnace-Unity:
### Purpose of this PR
Fixed a null ref exception when baking probes.

Fixes https://fogbugz.unity3d.com/f/cases/1277619/

---
### Testing status
Tested by using locally the 8.x.x/release branch and doing the repro steps from the fogbugz case.

---
### Comments to reviewers
Make sure to test on the 8.x.x/release branch instead of 7 based version as explained in the case.


iM0ve (approved):
Fix seems simple enough I think it is safe to merge.

We had a pair test with Julien. The issue is only the error message, the reflection probe bakes even with the error. Also might be project specific since I tried to reproduce on an empty project on mac HDRP 7.5.1 using same steps and It did not reproduce. While it reproduced on 7.5.2(branch) for that project

PR #1996, opened 2020-09-24
Fixed the usage of GUIEnable for volume components (case 1280018).

anisunity:
https://fogbugz.unity3d.com/f/cases/1280018/
I went through all the volume component that had the problem reported in the fogbug and fixed them and tested them. here is the list.
AmbientOcclusion
Bloom
DepthOfField
ScreenSpaceReflection
GlobalIllumination
IndirectLightingController
ProbeSettingsUI
Bloom
ContactShadow
ChromaticAbberation
MotionBlur


alelievr (approved):

JMargevics (approved):
image
Verified from my side. Disabled overrides now work properly ✅

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #2000, opened 2020-09-24
Fixed the diffusion profile becoming invalid when hitting the reset button (case 1269462).

anisunity:
When hitting the Reset button on the UI, the hash of the diffusion profile became invalid and the default values didn't make any sense. Now when hitting the button, the default values (when creating a new diffusion profile) are set and the hash is properly computed.
https://fogbugz.unity3d.com/f/cases/1269462/

Testing status
Did the steps to repro and the issue doesn't happen anymore.
Checked the hash using the debug menu and it stays the right value after the reset.


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -83,6 +83,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed SSGI compilation issues on PS4.
 - Fixed "Screen position out of view frustum" error when camera is on exactly the planar reflection probe plane.
 - Workaround issue that caused objects using eye shader to not be rendered on xbox.
+- Fixed the usage of GUIEnable for volume components (case 1280018).
```
JulienIgnace-Unity, line 86: Leftover from another PR?

anisunity: yeah seems like it

anisunity:
@JulienIgnace-Unity removed

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
repro step of Anis should be enough , merging.

PR #2002, opened 2020-09-24
Fixed save popup appearing twice

adrien-de-tocqueville:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1264774/

The call to `RequestCloseAndRelaunchWithCurrentArguments` is already asking wheter or not to save a scene
I also removed the call to `GUIUtility.ExitGUI()` because it is throwing errors in the console

---
### Testing status

Tested that the combinations behaved as expected:
- Answering save or don't save with automatic restarting
- Answering cancel and restarting manually after


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
Simple enough that we don't require QA.

PR #2003, opened 2020-09-24
Fixed wizard DXR setup for non-compatible systems

alelievr:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1263928/

I installed my old GTX 960 in my system to repro this issue, I added a check to SystemInfo.supportRaytracing in the wizard so even when everything is set up correctly, there is still an error:
Raytracing-error

---
### Testing status
I clicked on the fix button  / fix all button and it displays the correct error message.


JMargevics (approved):
Since we know that SystemInfo.supportsRayTracing is correctly detecting Linux and MacOS, no additional testing is required from our side. Warning message is clear. Approving as is ✅

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #2004, opened 2020-09-24
[ShaderGraph] [bugfix 1280106] AddSlot should not reorder slots

cdxntchou:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1280106/
https://fogbugz.unity3d.com/f/cases/1282855/

When calling AddSlot(), the code used to remove existing slots with the same ID, then add the new slot.  This had the side effect of moving the slot id to the end of the list.

In cases where slots are dynamically updated, like in CustomFunctionNode, this causes any modified slot to pop to the end of the list.

This PR changes the way AddSlot works, by attempting to replace the first matching slot with the same ID, and only appending if there is no matching slot.  It will still remove all other matching slots.

This shouldn't impact any other usages of AddSlot(), as none of them rely on the "always moves to the end of the list" behavior.

---
### Testing status
Tested against repro case.  Tested that modifying CFN slot type or name does not reorder the slots.

Tested that reordering the slots still works, and that displayed slot order matches the order in the List.  

Tested that reordering slots still correctly remaps connected edges.

Tested that the order, type, and names still serialize correctly.

Tested that modifications to order, type, and names work with undo/redo.

---
### Comments to reviewers


elizabeth-legros (approved):

com.unity.shadergraph/Editor/Data/Nodes/AbstractMaterialNode.cs
```diff
@@ -761,16 +761,36 @@ public MaterialSlot AddSlot(MaterialSlot slot, bool findPrevInstance = true)
...
+                    else
+                    {
+                        // this is a duplicate slot id... remove it
+                        m_Slots.RemoveAt(x);
+                        x--;
+                    }
+                }
+            }
+            if (!insertedSlot)
+                m_Slots.Add(slot);
```
marctem, line 793: This seems a lot cleaner and C#-ish to me
```suggestion
            var foundIndex = m_Slots.FindIndex(x => x.value.id == slot.id);
            if(foundIndex < 0)
            {
                foundIndex = m_Slots.Count;
            } else {
                m_Slots.RemoveAll(x => x.value.id == slot.id);
            }
            m_Slots.Insert(foundIndex, slot);
            
```

marctem (changes requested):

com.unity.shadergraph/Editor/Data/Nodes/AbstractMaterialNode.cs
cdxntchou: Not sure if it's cleaner, just more C# style, and slower..  :)

cdxntchou: Hmm..  if there was a RemoveAllFromRange() it would be expressable in C# in the near optimal form...

cdxntchou: How about this?
```

            int firstIndex = m_Slots.FindIndex(s => s.value.id == slot.id);
            if (firstIndex >= 0)
            {
                m_Slots[firstIndex] = slot;
                m_Slots.RemoveAllFromRange(s => s.value.id == slot.id, firstIndex + 1, m_Slots.Count - (firstIndex + 1));
            }
            else
                m_Slots.Add(slot);


// with RemoveAllFromRange function:

        public static int RemoveAllFromRange<T>(this List<T> list, Predicate<T> match, int startIndex, int count)
        {
            // TODO: to match behavior of RemoveRange we should throw exceptions when startIndex/Count are bad:

            int readIndex = startIndex;
            int writeIndex = startIndex;
            while (readIndex < list.Count)
            {
                T element = list[readIndex];
                readIndex++;

                bool remove = match(element);
                if (!remove)
                {
                    list[writeIndex] = element;
                    writeIndex++;
                }
            }

            // once we're done, we can remove the entries at the end
            int numberRemoved = list.Count - writeIndex;
            if (numberRemoved > 0)
            {
                list.RemoveRange(writeIndex, numberRemoved);
            }

            return numberRemoved;
        }




```

marctem: Makes sense, let's put that in a ListUtilities file and ship it.

cdxntchou:
Katana all green except for "Build Universal on Win_DX11_Player on version trunk " which is also failing on master.

marctem (approved):

ghost (approved):
Also ensured that it fixes https://fogbugz.unity3d.com/f/cases/1282855/

PR #22, opened 2020-04-06
[10.x.x] Pure URP XRSDK

thomas-zeng:
### Checklist for PR maker
- [x] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [ ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR.
- [x] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
This PR reworked XR in URP.

---
### Testing status

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in developer mode, you have a button at end of resources that check the paths)
- Other: 

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

Any test projects to go with this to help reviewers?

---
### Comments to reviewers
What dev has tested:
- URP template in XR on Win10 with d3d11 is working for MockHMD and Oculus with either multi-pass or single-pass instanced
- URP template in XR on the Quest with GLES3 + Multiview is working
- URP template in XR on the Quest with GLES3 + Multipass has some bugs with lighting
- URP template in XR on the Quest with Vulkan is working without FFR&MSAA

Useful links:
- SRP Status Confluence Page:
https://confluence.unity3d.com/display/XG/SRP+Status

PR #2006, opened 2020-09-24
[7.x.x][case 1251402] Fixed an issue where gamma correction was being applied twice for Google VR on iOS devices when FXAA was enabled

todds-unity:
### Purpose of this PR
Fixed an issue where gamma correction was being applied twice for Google VR on iOS devices when FXAA was enabled. [case 1251402](https://issuetracker.unity3d.com/issues/metal-urp-ios-google-vr-ios-white-filter-appears-when-using-fxaa-plus-post-processing-plus-single-pass-stereo-rendering-mode)

### Testing status
Manually tested the change on local mobile devices.

### DIFF
--- a/.yamato/_abv.yml
+++ b/.yamato/_abv.yml
@@ -92,8 +92,6 @@ all_project_ci_nightly_2019.4:
     dependencies:
       -  path: .yamato/_abv.yml#all_project_ci_2019.4
          rerun: on-new-revision
-      -  path: .yamato/_templates.yml#all_template_ci_2019.4
-         rerun: on-new-revision
       -  path: .yamato/_abv.yml#smoke_test_Standalone_2019.4
          rerun: on-new-revision
       -  path: .yamato/_abv.yml#smoke_test_playmode_2019.4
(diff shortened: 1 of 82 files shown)


thomas-zeng:
Note: the linear colorspace check is already landed in master in https://github.com/Unity-Technologies/Graphics/pull/22

thomas-zeng (approved):
LGTM. Thanks!

DiannaUnity (approved):
Version 2019.4.8f1 (60781d942082)
Wed, 05 Aug 2020 14:08:36 GMT
Branch: 2019.4/staging

Before fix; the sky and ground is lighter, skybox looks more grey
image



After fix; the scene becomes darker and matches the scene viewer
image

Confirmed that the fix works on iPhone XS.

phi-lira:
@todds-unity where is the PR to master? we need to land it on master first then backport to other versions.

phi-lira:
Updated on this PR?

eh-unity (approved):
Looks like this is entirely contained in XR-side (under defines). LGTM.

PR #2008, opened 2020-09-25
hotfix for regression caused by bugfix for issues with floating block…

esmelusina:
### Purpose of this PR
Fix for regression caused by #1889.

The original fix provides some protections from generating context menus while a block node is floating. These protections were applied across a larger case than originally intended.  Context Menus across many objects were failing to build as a result.

### Testing status

Manually testing, ensure that context menus build properly across MaterialGraphView objects.


elizabeth-legros (approved):

ghost (approved):

marctem (approved):

marctem:
SG CI passed, reamining failures are also in master.

PR #2012, opened 2020-09-25
Fixed not showing "no cameras rendering" message when overlay cameras…

Nzollner:
### Purpose of this PR
This PR fixes the missing "no cameras rendering" message in "game view" when only overlay cameras are active.
https://fogbugz.unity3d.com/f/cases/1276168/
---
### Testing status
I have tested the feature manually. I have also tested the UniversalGraphicsTest locally but without passing all the tests but the tests also won't pass on master. This will be fixed before being changed to a PR.
---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
@@ -144,7 +144,7 @@ namespace UnityEngine.Rendering.Universal
 #if UNITY_EDITOR
                 UniversalAdditionalCameraData cameraAdditionalData;
                 camera.TryGetComponent<UniversalAdditionalCameraData>(out cameraAdditionalData);
-                if (cameraAdditionalData.renderType == CameraRenderType.Overlay && camera.targetDisplay != -1)
+                if (cameraAdditionalData?.renderType == CameraRenderType.Overlay && camera.targetDisplay != -1)
                     camera.targetDisplay = -1;
 #endif
                 if (IsGameCamera(camera))


Nzollner:
This seems to be the same bug: https://fogbugz.unity3d.com/f/cases/1266533/

Issue with my solution is when changing from overlay back to base then the display value is reset to display 1 for this fix.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -661,6 +661,8 @@ void DrawCameraType()
             EditorGUILayout.PropertyField(m_AdditionalCameraDataCameraTypeProp, Styles.cameraType);
             if (EditorGUI.EndChangeCheck())
             {
+                camera.targetDisplay = (((CameraRenderType)m_AdditionalCameraDataCameraTypeProp.intValue) == CameraRenderType.Base)
```
lukaschod, line 664: Yea I was thinking about same solution. However this will fail if user will change target display with script.

Nzollner: Ohh I didn't think of that case.
Then we would need to override the function for URP specifically since the additional camera data is unique for URP.

lukaschod: Yea, but users can still change target display in camera. https://docs.unity3d.com/ScriptReference/Camera-targetDisplay.html

Nzollner: I added a statement in universalRenderPipeline setting each camera overlay to -1 for editor only.
Are the penalty to high or okay when it only runs in the editor?

PR #2014, opened 2020-09-25
[10.x.x] Update 049 reference image on iOS

tomzig16:
### Purpose of this PR
This PR updates reference image for test 049 for iOS as it was taken with higher quality settings. 
This PR also disables test 35 as it causes iOS to crash. Bug is reported in [fogbugz case 1280307](https://fogbugz.unity3d.com/f/cases/1280307/) and being investigated.

---

### Testing status
Yamato tests
Yamato branch: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/urp%252Fios%252Fcrash-investigation

---
### Comments to reviewers
Due to changed structure of filters asset file, this PR is the first one to update it. Tested locally and on Yamato, seems like it filters out without any issues.

### DIFF
--- a/TestProjects/UniversalGraphicsTest/Assets/Test/TestFilters/TestCaseFilters.asset
+++ b/TestProjects/UniversalGraphicsTest/Assets/Test/TestFilters/TestCaseFilters.asset
@@ -538,4 +538,4 @@ MonoBehaviour:
     GraphicsDevice: 16
     XrSdk: None
     StereoModes: 0
-    Reason: iOS crashes (scene 035)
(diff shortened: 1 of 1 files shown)


ellioman (approved):
Looks good. I would though include the bug ID when disabling tests in the test filter. That way we can easily check the status.

tomzig16:
> 
> 
> Looks good. I would though include the big ID when disabling tests in the test filter. That way we can easily check the status.

Totally agree! Addressed in 4136db4

PR #2017, opened 2020-09-25
Fixed an obsolete API usage

JulienIgnace-Unity:
### Purpose of this PR
Fixed an obsolete API usage for ConstantBuffers

---
### Testing status
Automatic testing should be enough for a trivial change like this.

---
### Comments to reviewers
We need to wait for current issues with trunk to be solved to be able to test this against latest trunk and not an older pinned revision.


sophiaaar:
requires a newer unity version, we're working on it (cc @kivistikliis )

sebastienlagarde (approved):

sophiaaar:
The trunk version has been updated now, so this should be fine to merge

PR #2023, opened 2020-09-25
Fix hair shader compilation issue with Debug Display mode

sebastienlagarde:
Fix hair shader compilation issue with Debug Display mode


anisunity (approved):
haven't tested it locally, looks fine to me from the code

PR #2026, opened 2020-09-25
Fix rendergraph failing on consoles

FrancescoC-unity:
On consoles in the tests: 

 - Main camera does require cameraFilteringBuffer buffer as it is using MSAA
- Planar does not use MSAA and on console it does not require cameraFilteringBuffer 

As a consequnce planar try to use an handle from another camera that is invalid for the rendergraph execution of the planar.


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

JulienIgnace-Unity:
Merged in https://github.com/Unity-Technologies/Graphics/pull/1783

PR #2027, opened 2020-09-25
Added support for box and pyramid point lights in path tracing

eturquin:
Addresses https://fogbugz.unity3d.com/f/cases/1278506/

The change is quite minimal, and well localized in terms of code. User-wise:
- Before, pyramid and box modes were ignored in path tracing, resulting in an omni-directional point light
- Now, they are taken into account and should match their rasterized counterparts


sebastienlagarde (approved):

sebastienlagarde:
We need to add a test for box projector and pyrmaid (could be done in next PR)

sebastienlagarde:
mergeing for 10.1 pacakge

remi-chapelain:
Tested IES + cookie, outer angles, inner angle, aspect ratio and size X/Y, everything match ✔️

PR #2029, opened 2020-09-25
[10.2][Shader Graph] Small UI Bugfixes

alindmanUnity:
### Purpose of this PR

Fixes two UI bugs:
1. Exposure Node has clipped text. Changed the drop down label to 'Type" to avoid clipping issues. [1265057]
2. Mesh Deformation nodes have no Category color. We cannot add a new category color due to UX constraints, so I have instead moved the nodes to live in the Utility category. [1227081]

---
### Testing status

image
Nodes added to new Shader Graph file, visual changes confirmed. 

---
### Comments to reviewers

@JussiKnuuttila please confirm that moving the mesh deformation node category is acceptable.


JussiKnuuttila:
I don't have any specific experience on mesh deformation, but I don't think changing a the color category could cause problems.

@svens-unity @HedvigAxelsson

svens-unity:
Utility could work. It is a bit tricky to put them in a category. I suppose both of them could be considered input too. The compute deformation node is similar to VertexPosition but with more magic from the deformation system. It is the deformed vertex position. Linear blend skinning does the same from a functional point of view but takes input... 
All in all, I definitely think the nodes should be in the same place.

alindmanUnity:
> Utility could work. It is a bit tricky to put them in a category. I suppose both of them could be considered input too. The compute deformation node is similar to VertexPosition but with more magic from the deformation system. It is the deformed vertex position. Linear blend skinning does the same from a functional point of view but takes input...
> All in all, I definitely think the nodes should be in the same place.

Yeah, I'm not married to either Input or Utility since ultimately they're your nodes. We can put them in either category, we just can't create a new category for them because of the UX limitations. 

If you want to move them to Input let me know and I'll do that, otherwise just slap a green dot on this one and we can leave them in Utility. either way, they will definitely live in the same place.

svens-unity:
I understand. Let's do Input/Mesh Deformations or Input/Geometry. I think that would be more fitting. Thanks for handling this.

sebastienlagarde (approved):

ghost (approved):
Seems good, should you add a docs reviewer? Not sure about the effects of this PR on the docs.

alindmanUnity:
@sharlenet I've made updates to the Table of Contents and Input Node pages to reflect that changes for the mesh deformation nodes, as well as adding some notes in  Graph Targets.md and the Upgrade Guide for cross pipeline support. Could you please review and make any needed changes?

svens-unity (approved):
Thank you

sharlenet (approved):

marctem (approved):

PR #2035, opened 2020-09-28
Reduce bloom flickering

adrien-de-tocqueville:
### Purpose of this PR

The bloom effect has flickering problems when there is small and bright objects visible, like the sun from the pbr sky.
The PR fix texture sample coordinates in the prefilter pass, which didn't use the correct pixel size when sampling from the input texture.
It also adds a "high quality prefiltering" advanced option wich helps in certain scenarios.

Note that there are still flickering issues on the sun when using taa

---
### Testing status

Tested with the sun in the pbr sky because it is mainly what is causing problems. There is less flickering now


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/BloomUpsample.compute
```diff
@@ -42,8 +42,5 @@ void KMain(uint3 dispatchThreadId : SV_DispatchThreadID)
 
     float3 output = lerp(highRes, lowRes, Scatter);
 
-    // Guard bands
-    output *= all(dispatchThreadId.xy < uint2(_TexelSize.xy));
```
FrancescoC-unity, line 46: This is gone with a recent PR that fixed a bug, please remove from this PR to avoid conflicts.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/BloomPrefilter.compute
```diff
@@ -37,11 +55,11 @@ void KMain(uint3 dispatchThreadId : SV_DispatchThreadID)
         float2(-0.4, -0.9)
     };
 
-    float3 color0 = SAMPLE_TEXTURE2D_X_LOD(_InputTexture, sampler_LinearClamp, ClampAndScaleUVForBilinear(_TexelSize.zw * kTaps[0] + uv, _TexelSize.zw), 0.0).xyz;
-    float3 color1 = SAMPLE_TEXTURE2D_X_LOD(_InputTexture, sampler_LinearClamp, ClampAndScaleUVForBilinear(_TexelSize.zw * kTaps[1] + uv, _TexelSize.zw), 0.0).xyz;
-    float3 color2 = SAMPLE_TEXTURE2D_X_LOD(_InputTexture, sampler_LinearClamp, ClampAndScaleUVForBilinear(_TexelSize.zw * kTaps[2] + uv, _TexelSize.zw), 0.0).xyz;
-    float3 color3 = SAMPLE_TEXTURE2D_X_LOD(_InputTexture, sampler_LinearClamp, ClampAndScaleUVForBilinear(_TexelSize.zw * kTaps[3] + uv, _TexelSize.zw), 0.0).xyz;
-    float3 color4 = SAMPLE_TEXTURE2D_X_LOD(_InputTexture, sampler_LinearClamp, ClampAndScaleUVForBilinear(_TexelSize.zw * kTaps[4] + uv, _TexelSize.zw), 0.0).xyz;
+    float3 color0 = SAMPLE_TEXTURE2D_X_LOD(_InputTexture, sampler_LinearClamp, ClampAndScaleUVForBilinear(_ScreenSize.zw * kTaps[0] + uv, _ScreenSize.zw), 0.0).xyz;
```
FrancescoC-unity, line 58: Have you tested this change by rescaling the window? 

The ScreenSize is the actual widht/height (so with the RTHandleScale baked in), but ClampAndScaleUVForBilinear scale with that RTHandleScale, so I am a bit unsure about this. 

Also, why the change at all? _TexelSize.zw is the size of the first mip.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/BloomPrefilter.compute
```diff
@@ -7,17 +7,34 @@
...
 #pragma kernel KMain
 
+#pragma multi_compile LOW_QUALITY HIGH_QUALITY
+
 TEXTURE2D_X(_InputTexture);
 
 RW_TEXTURE2D_X(float3, _OutputTexture);
 
 SAMPLER(sampler_LinearClamp);
+SAMPLER(sampler_BilinearClamp);
```
FrancescoC-unity, line 17: linear sampler is bilinear 

Also, we should use the pre-declared samplers (i.e. s_point_clamp_sampler)

com.unity.render-pipelines.high-definition/Documentation~/Post-Processing-Bloom.md
```diff
@@ -40,6 +40,7 @@ Bloom includes [more options](More-Options.md) that you must manually expose.
 | **Property**               | **Description**                                              |
 | -------------------------- | ------------------------------------------------------------ |
 | **Resolution**             | Use the drop-down to set the resolution at which HDRP processes the Bloom effect. If you target consoles that use a very high resolution (for example, 4k), select **Quarter,** because it is less resource-intensive.<br />&#8226; **Quarter**: Uses quarter the screen resolution.<br />&#8226; **Half**: Uses half the screen resolution. This property only appears when you enable [more options](More-Options.md). |
+| **High Quality Prefiltering** | Enable the checkbox to make HDRP use 13 bilinear samples insted of 4 linear samples during the prefiltering pass. This increases the resource intensity of the Bloom effect, but results in less flickering by small and bright objects like the sun.<br />This property only appears when you enable [more options](More-Options.md). |
```
FrancescoC-unity, line 43: As said below, linear == bilinear in this context

FrancescoC-unity (changes requested):

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/BloomPrefilter.compute
FrancescoC-unity: **IF** this is what you wanted to do 

Then you should use 

ClampAndScaleUVForBilinear(float2 UV) 

as it embeds _ScreenSize.zw as texel size already.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/Bloom.cs
```diff
@@ -114,6 +134,10 @@ public bool highQualityFiltering
         [SerializeField, FormerlySerializedAs("resolution")]
         private BloomResolutionParameter m_Resolution = new BloomResolutionParameter(BloomResolution.Half);
 
+        [Tooltip("When enabled, bloom uses multiple bilinear samples for the prefiltering pass.")]
+        [SerializeField]
+        private BoolParameter m_HighQualityPrefiltering = new BoolParameter(true);
```
FrancescoC-unity, line 139: We usually default to whatever is the medium preset, so this should be false (or true in the medium preset, but in that case we need to profile)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/BloomPrefilter.compute
adrien-de-tocqueville: That actually seems correct , but thanks for the check, I didn't do it. So I updated the function call to use the overload

FrancescoC-unity (approved):

sebastienlagarde (approved):

PR #2038, opened 2020-09-28
Added support for exposure for the case of planar reflections.

anisunity:
https://fogbugz.unity3d.com/f/cases/1234529/
Previously, the exposure support for the planar probes was not complete. In case the user srt the exposure weight attenuation, the result was inconsistent with what is seen directly through the camera.
With this PR, the result is equivalent for both shader graph and .shader versions.
Now when enabling the exposure mode frame setting on planar reflections, the results takes into account the exposure mode properly.
The behavior does not concern reflections probes (dynamic or static).

image

Testing status:
- Not all the HDRP tests were not green before the PR.
- Locally all the Planar probe related tests were green before and after the PR
- Added a new tests that covers the new behavior (with or without the frame setting).
- Varied the exposure (when 0, it is the same behavior, with or without the fix). When the exposure is higher the result is only the same with the fix.

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/2220_PlanarProbeExposure.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/2220_PlanarProbeExposure.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:1391f70df9667286e147c41130332093eaa5ca7d3ca502d15b893eb4f1125b60
+size 11391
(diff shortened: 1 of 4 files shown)


JMargevics:
Hey! Checked enabling Exposure Control and played with Emissive weight on a shader. Also checked Inspector preview and everything seem to look correct so far. One question: when you enable Exposure Control scene planar reflection preview turns completely black. I believe this expected because unlike Inspector preview we don't have the exposure slider there? 

BcTZUwv0dR

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2050,6 +2050,13 @@ ref _cullingResults
 
                         hdCamera.parentCamera = parentCamera; // Used to inherit the properties of the view
 
+                        if (visibleProbe.type == ProbeSettings.ProbeType.PlanarProbe && hdCamera.frameSettings.IsEnabled(FrameSettingsField.ExposureControl))
+                        {
+                            RTHandle exposureTexture = GetExposureTexture(hdCamera);
+                            visibleProbe.RequestProbeExposureValue(exposureTexture);
```
anisunity, line 2056: Add comment

anisunity:
Merged master into this branch and relaunched all the yamato HDRP CI Trunk tests

PR #2040, opened 2020-09-28
Shadergraph/fix/vfx properties

iTris666:
---
### Purpose of this PR

Fixing https://fogbugz.unity3d.com/f/cases/1278469/

While it look like vfx doesn't update it is caused by the shadergraphvfxasset giving the first property set since last domain reload instead of the latest one.

This appends because the shadergraphvfxasset object on which the ScripterImporter set the new properties and sets m_Data as null is a copy then it is serialized and deserialized into the persistent shadergraphvfxasset that still has the old m_Data, and since it is not null it is not deserialized. Adding m_Data = null on OnAfterDesirialize ensure m_Data for the right shadergraphvfxasset is set to null after import.

---
### Testing status

Tested that now, on deletion or add of a property and save of shader graph, the vfx is updated.

---
### Comments to reviewers


elizabeth-legros (commented):
Should we get rid of the default initialization and explicitly call `EnsureProperties` in the OnAfterDeserialize?

iTris666:
> Should we get rid of the default initialization and explicitly call `EnsureProperties` in the OnAfterDeserialize?

I think we should also call EnsureProperties inside SetProperties, and remove other calls. I'll make the changes

elizabeth-legros (approved):

PR #2042, opened 2020-09-28
Add support of input system v2

alelievr:
### Purpose of this PR
Add support of Input System V2 where it's possible.

Please see this document to know which part/bindings are affected by these changes: https://docs.google.com/spreadsheets/d/1fhyg9aEsTVky_s04cNZjCfnV5mqI1xFz-OIhWLQBOyo/edit#gid=0

---
### Testing status
There is no automated test for input related features, so I tested everything manually with a gamepad, touch simulator and mouse/keyboard.

I tested all the controls in the document above with Input System Package enabled in the project settings. 
image
For testing, please do not use "Both" mode because in some cases, the old input system is preferred instead of the new one.

Note that I didn't tested the debug menu in build.
Some additional testing may be required regarding all the input configurations: old, old + new (Both), none (new but without the package), new.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Camera/FreeCamera.cs
+++ b/com.unity.render-pipelines.core/Runtime/Camera/FreeCamera.cs
@@ -124,7 +124,6 @@ namespace UnityEngine.Rendering
             inputRotateAxisY = 0.0f;
             leftShiftBoost = false;
             fire1 = false;
-            Vector2 delta;
 
 #if USE_INPUT_SYSTEM
             var lookDelta = lookAction.ReadValue<Vector2>();


JulienIgnace-Unity (approved):

com.unity.visualeffectgraph/Runtime/Utilities/EventBinding/Implementation/VFXMouseEventBinder.cs
```diff
@@ -32,41 +37,132 @@ protected override void SetEventAttribute(object[] parameters)
...
+#if USE_INPUT_SYSTEM
+            return Pointer.current.position.ReadValue();
+#else
+            return Input.mousePosition;
+#endif
+        }
+
         private void OnMouseDown()
         {
+            Debug.Log("OnMouseDown");
```
PaulDemeulenaere, line 135: Remove these log before merge 😆

com.unity.visualeffectgraph/Runtime/Utilities/PropertyBinding/Implementation/VFXInputAxisBinder.cs
```diff
@@ -22,6 +22,7 @@ public override bool IsValid(VisualEffect component)
 
         public override void UpdateBinding(VisualEffect component)
         {
+#if ENABLE_LEGACY_INPUT_MANAGER
```
PaulDemeulenaere, line 25: How does this behave when com.unity.input *isn't* installed ?
Is ENABLE_LEGACY_INPUT_MANAGER=1 automatically defined ?
Otherwise, it should be 
#if ENABLE_LEGACY_INPUT_MANAGER || !ENABLE_INPUT_SYSTEM_PACKAGE

PaulDemeulenaere (changes requested):
Some minor issue to address.

com.unity.visualeffectgraph/Runtime/Utilities/PropertyBinding/Implementation/VFXInputAxisBinder.cs
alelievr: This defined is supposed to be managed by Unity: https://docs.unity3d.com/Manual/PlatformDependentCompilation.html
So it should work disable the define if the module is not installed

sebastienlagarde (approved):

sebastienlagarde:
Yamato is green. We need this PR in 10.1 package, it is critical and we need it for the templates. Merging. QA will do a pass on it while testing 10.1 package.

remi-chapelain:
checked for smoke testing : 10.2.0
Tested in new template build with the old, new and both system that input works correctly with keyboard + gamepad ✔️ 
Also tested debug menu in build with the old, new and both input system ✔️ 
Tried to test the binding with mouse buttons on VFX graph, works fine in editor but ended up finding an [issue](https://fogbugz.unity3d.com/f/cases/1290625/) in player, will try again when this is fixed.

PR #2043, opened 2020-09-28
[ShaderGraph] [bugfix 1254483] Disallow VirtualTexture as SubGraph output

cdxntchou:
### Purpose of this PR
Fix for bug https://fogbugz.unity3d.com/f/cases/edit/1254483/

If you add a VirtualTexture as a SubGraph output, it generates cryptic errors.
This is not a supported usage of VirtualTextures, so this PR explicitly removes the ability to add a VirtualTexture as a SubGraph output, at the UI level.

VirtualTexture shows up disabled in the list of possible types:
image

---
### Testing status
Tested adding VirtualTexture to a SubGraph -- the choice is no longer selectable.
Existing VirtualTexture outputs will not be modified, but will continue to spew the cryptic error message.

Tested CFN slot UI, which also uses the same class.  Virtual Texture is not disabled as a choice in that context.

---
### Comments to reviewers


esmelusina (commented):
Looks fine.

esmelusina (approved):

cdxntchou:
Katana all green except for "Build Universal Win DX11" which is also failing on master.

ghost (approved):
Looks like it works pretty well.

PR #2052, opened 2020-09-29
[URP] Fix temporary RTs having mips for no reason

akeidonas:
### Purpose of this PR
fix temporary RTs/PostProcessing inheriting mipmap settings from main RT

---
### Testing status
tested on UniversalRenderingExamples' Cockpit scene 

---
### Comments to reviewers


pbbastian (changes requested):
Code looks good! Change log has lots of unrelated changes though, can you remove those?

pbbastian (approved):

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -668,13 +668,17 @@ void CreateCameraRenderTarget(ScriptableRenderContext context, ref RenderTexture
                 {
                     bool useDepthRenderBuffer = m_ActiveCameraDepthAttachment == RenderTargetHandle.CameraTarget;
                     var colorDescriptor = descriptor;
+                    colorDescriptor.useMipMap = false;
```
phi-lira, line 671: I wonder if we should change this on the descriptor creation so it propagates to other passes as well. (what about the _CameraOpaqueTexture in CopyColorPass?) https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs#L439-L458

akeidonas: > I wonder if we should change this on the descriptor creation so it propagates to other passes as well.
Mips are enabled on Render texture inside Editor and then it's copied around. This is case by case mip flag removal.
>what about the _CameraOpaqueTexture in CopyColorPass?
this fix was recommended by apple and I'm still familiarizing myself with URP, so I guess we missed it.

akeidonas: but wait, "what" about _CameraOpaqueTexture - shouldn't it be a precise reflection of active Camera target (as in with all mips and everything)?

phi-lira (approved):

PR #2055, opened 2020-09-30
SSAO support for deferred renderer.

kaychang-unity:
---
### Purpose of this PR
This adds support for SSAO with deferred renderer.

---
### Testing status
UniversalGraphicsTests test scenes 142, 143 are passing.

---
### Comments to reviewers
Many changes made to make this work. Some important points here:
- SSAO code was changed to work with world-space normals instead of view-with-inverted-z-space normals (deferred renderer output world space normals and requires depthNormal prepass to also output world space normals
- I still haven't fixed `SSAO.hlsl` `PickSamplePoint()` to work with world space normals. It generates a random vector, but the distribution may not be uniform around a sphere?
- `_CameraNormalsTexture` format was changed to R8G8B8A8_snorm without octahedron encoding (now optional). The result SSAO was very close from original effect but _not identical_ and I increased "Average Correctness Threshold" from 0.001 to 0.0015.
- in deferred mode,  `_CameraNormalsTexture` is not created, and instead the gbuffer texture 3rd slot is used. In that case, octahedron encoding for normals is controlled by `_GBUFFER_NORMALS_OCT` shader keyword.
- A new DepthNormalsOnly pass was added for forward-only materials. This is needed in order to only render forward-only in depthNormal prepass when deferred renderer is enabled (bakedLit and ComplexLit materials).
- Added RenderPassEvent `BeforeRenderingGbuffer`, `AfterRenderingGbuffer`, `BeforeRenderingDeferred` and `AfterRenderingDeferred`
- I added a small hack in the code to make depth prepass work when priming the final depth buffer (`_CameraDepthAttachment`) instead of the depth texture copy (`_CameraDepthTexture`). The bug is that several locations in the code create `RenderTargetIdentifier` with depthSlice=0 (default value, but incorrect for XR support), and other locations with depthSlice=-1 (non-default value, but correct value for XR support). The proper fix is to change all code that create `RenderTargetIdenfiers` to use a constructor that explicitely set depthSlice=-1 but I think it should be made in a separate PR because it can introduce new bugs.
- Refactored gbuffer layout to store occlusion channel (occlusion from baked lighting).

### Render passes when deferred mode is selected:
1) render depthNormal prepass, but only for forward-only materials. Use real depth and normal attachments (not texture copies)
2) Render gbuffer on standard materials, which continue filling up the depth and normal buffers bound in 1)
3) Generate SSAO texture from complete depth and normal buffers
3) Apply SSAO on lighting buffer, which at this stage contains baked lighting from gbuffer pass
4) Deferred lighting pass use SSAO texture
5) Render forward-only materials, which also sample the SSAO texture

Known failed test:
- UniversalGraphicsTest 145_SoftParticlesStackWithPP scene is failing with deferred renderer. This is a known issue. MSAA post-effect that runs between the 2 camera stack clears the depth buffer due to a limitation with Unity CommandBuffer that forces clearing stencil and depth components together, while the code only intents to clear stencil. This does not create a bug with Forward renderer because the copy of the depth buffer is performed only once before MSAA post-effect. In Deferred renderer, a copy of the depth buffer is additionally performed during the second camera stack, which invalidates the contents of the depth buffer copy made during the first camera stack. The real fix is to make MSAA clear stencil only, which is not easily implementable with Unity CommandBuffer interface.


kaychang-unity:
Currently, when deferred is setup in the forward renderer, we must also set the SSAO renderer feature manually to source=gbuffer
image

thomas-zeng (approved):
Looks good for XR part code changes.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -306,6 +315,13 @@ inline void InitializeBRDFDataDirect(half3 diffuse, half3 specular, half reflect
 #endif
 }
 
+// Legacy: do not call, will not correctly initialize albedo property.
+inline void InitializeBRDFDataDirect(half3 diffuse, half3 specular, half reflectivity, half oneMinusReflectivity, half smoothness, inout half alpha, out BRDFData outBRDFData)
+{
+    InitializeBRDFDataDirect(half3(0.0, 0.0, 0.0), diffuse, specular, reflectivity, oneMinusReflectivity, smoothness, alpha, outBRDFData);
+}
```
ellioman, line 322: Could we move this function to `Deprecated.hlsl`?

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
```diff
@@ -209,13 +218,18 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
...
-        AmbientOcclusionFactor aoFactor = GetScreenSpaceAmbientOcclusion(screen_uv);
-        unityLight.color *= aoFactor.directAmbientOcclusion;
+            AmbientOcclusionFactor aoFactor = GetScreenSpaceAmbientOcclusion(screen_uv);
+            unityLight.color *= aoFactor.directAmbientOcclusion;
             #if defined(_DIRECTIONAL) && defined(_DEFERRED_FIRST_LIGHT)
-            alpha = aoFactor.indirectAmbientOcclusion;
+            half occlusion = min(surfaceDataOcclusion, aoFactor.indirectAmbientOcclusion);
+            // What we want is really to apply the mininum occlusion between the baked occlusion from surfaceDataOcclusion and real-time occlusion from SSAO.
+            // But we already applied the baked occlusion during gbuffer pass, so we have to cancel it out here.
+            occlusion /= surfaceDataOcclusion;
```
ellioman, line 234: Thought: Would changing this to `occlusion *= rcp(surfaceDataOcclusion)` be faster than a division or make no difference at all?

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
```diff
@@ -531,6 +550,7 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
             HLSLPROGRAM
 
             #pragma multi_compile _SSAO_ONLY
```
ellioman, line 565: Where does this keyword get enabled/disabled?

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -434,20 +439,13 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
                 {
                     if (this.actualRenderingMode == RenderingMode.Deferred)
```
ellioman, line 440: For me it would feel a lot cleaner to have a comment that this is only needed for the Forward path and change the condition to `if (this.actualRenderingMode != RenderingMode.Deferred)` and skip the else.

ellioman (approved):
Looks really good from my point of view. I'd like @phi-lira to do a pass as well as these are a lot of changes and I might miss something. One thing that concerns me how we view Shader API changes in SSAO.hlsl. Do we consider that API to be public and need to add them to Deprecated.hlsl?

I added some minor comments. We should then include QA from URP and XR to test this thoroughly.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
kaychang-unity: Actually I can't. This function relies on `struct BRDFData` to be declared, which is at the top of `Lighting.hlsl`.
However, our include order is:
ShaderLibrary\Core.hlsl -> ShaderLibrary/Deprecated.hlsl -> ShaderLibrary\Lighting.hlsl

So I cannot move `InitializeBRDFDataDirect()` in `Deprecated.hlsl`.
Unless we refactor the code and split globals/struct declarations in a different header from functions?

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
kaychang-unity: The keyword never need to be enabled because it is the only option (there is no '_' option). It is equivalent to '#define _SSAO_ONLY', but the later does not work, I suppose because `_SSAO_ONLY` is tested in the `HLSLINCLUDE` block, which get added before the `#define _SSAO_ONLY` is actually parsed...

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
ellioman: Ahh I see. That's a very good idea!
I won't let you worry about that work though. I'm noting this down and will talk to the URP team about this as I've been meaning to clean up our shader code for a while and this would fit into that.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
kaychang-unity: Updated the code.

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
kaychang-unity: Ok, done!

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -15,9 +15,14 @@ SAMPLER(sampler_BaseMap);
...
 // Params
-float4 _BlurOffset;
 float4 _SSAOParams;
 float4 _SourceSize;
+float4 _ProjectionParams2;
+float4x4 _CameraViewProjections[2]; // This is different from UNITY_MATRIX_VP (platform-agnostic projection matrix is used). Handle both non-XR and XR modes.
+float4 _CameraViewTopLeftCorner[2];
+float4 _CameraViewXExtent[2];
+float4 _CameraViewYExtent[2];
+float4 _CameraViewZExtent[2];
```
manuele-bonanno, line 26: do you need 32 bit float precision for these uniforms and the ReconstructViewPos calculations? On mobile we are already doing too much fp32 bit math. Using half precision would be recommended (both for the uniforms + all the local variables)

manuele-bonanno: (this will be addressed as part of a future optimisation PR)

manuele-bonanno (approved):
Looks good, optimisation issues will be addressed in a following PR

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
kaychang-unity: I added a comment about potentially moving these uniforms to half.
Since the whole SSAO code does not use half at the moment, it would be a bigger task to optimize SSAO for half types.

Jonasmortensen (approved):
Looks good from tech-art perspective! 👍

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -56,45 +56,65 @@ public enum ScriptableRenderPassInput
         /// </summary>
         AfterRenderingPrePasses = 200,
 
+        /// <summary>
+        /// Executes a <c>ScriptableRenderPass</c> before rendering gbuffer pass.
+        /// </summary>
+        BeforeRenderingGbuffer = 250,
```
kaychang-unity, line 62: @phi-lira modifying the enum values break serialized enum values in existing project (ex: UniversalRenderingExamples). Should I squeeze the 4 new enum values between the range 200-250 instead?

cinight (approved):
All good from my side. Test doc: https://docs.google.com/document/d/19rramzmJDkAPpJIguv3D68XZAi6ycyI6ZbNkOKTP7XA/edit

Erika is also testing this so added her as reviewer.

ernestasKupciunas (approved):
Looks good to me. I don't have any more active bug cases.
Testing doc: https://docs.google.com/document/d/1QvjMmwLUTYHXYae3MQNxD5mEnhd3BArhwQmHfwcGtCI/edit#

ryanhy-unity (approved):
I checked XR SSAO forward rendering with Thomas's changes (https://github.com/Unity-Technologies/Graphics/pull/2663/files#submit-review) to fix terrain for xr too and the eyes were aligned properly and the SSAO looked proper.

Checked D3D11 (Mock HMD) and Oculus Quest GLES3 with Multiview.

erikabar (approved):
All issues found during testing were fixed, testing doc: https://docs.google.com/document/d/1JAGpzDJrcN_Azc_i0kXEg21vBYdz6OrIMjwC3bnuYD0/edit?usp=sharing

DennisDeRykeUnity:
For my intial round of testing I set Player settings > Other settings > Script compilation > Scripting Define Symbols > ENABLE_RENDERING_PATH_UI. 
Then I re-started the Unity Editor.  
Next I set Forward Renderer asset > Lighting > Renderer Path > Deferred.
I verified that the ForwardRenderer asset is the first item in the UniversalRPAsset > General > Renderer List.

Using the following config I viewed the following scenes in Play mode and saw no anomalies, judging from the non-XR reference images for these test scenes and my previous experiences testing SSAO for XR with URP. 

- Unity (trunk ed1e382e88f9)
- URP (universal/ssao-deferred-support ff1f691)
- D3D11
- Single Pass Instancing

140_SSAO_DepthOnly_Projection
142_SSAO_DepthNormal_Projection
144_SSAO_RenderToBackBuffer

Orthographic rendering is not relevant for XR so I did not test the Ortho scenes. More testing tomorrow...

DennisDeRykeUnity (approved):
Today I built and ran Windows Standalone Players for these two test using the following configuration, and found no problems.  I am approving this PR based on my tests and the observations of the other reviewers. 

@kaychang-unity  However, a lot of this PR's Yamato pipelines are red.  Have you compared these against SRP master to verify that your code changes are not causing any of these test failures?

142_SSAO_DepthNormal_Projection (Multi Pass Stereo)
144_SSAO_RenderToBackBuffer (Single Pass Instancing


- Unity (trunk ed1e382e88f9)
- URP (universal/ssao-deferred-support ff1f691)
- D3D11
- Quest Link

kaychang-unity:
@DennisDeRykeUnity 144_SSAO_RenderToBackBuffer is indeed not correct on iPhone Metal, I will have a look tomorrow.
 What is the error about 142_SSAO_DepthNormal_Projection ?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -61,6 +61,26 @@ public enum ScriptableRenderPassInput
         /// </summary>
         AfterRenderingPrePasses = 200,
 
+        /// <summary>
+        /// Executes a <c>ScriptableRenderPass</c> before rendering gbuffer pass.
+        /// </summary>
+        BeforeRenderingGbuffer = 210,
```
phi-lira, line 67: What happens if a render pass is schedule in BeforeRenderingGBuffer but executed in Forward mode?
Do we execute or skip it?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -61,6 +61,26 @@ public enum ScriptableRenderPassInput
...
+
+        /// <summary>
+        /// Executes a <c>ScriptableRenderPass</c> after rendering gbuffer pass.
+        /// </summary>
+        AfterRenderingGbuffer = 220,
+
+        /// <summary>
+        /// Executes a <c>ScriptableRenderPass</c> before rendering deferred shading pass.
+        /// </summary>
+        BeforeRenderingDeferred = 230,
```
phi-lira, line 77: Maybe more specific
```suggestion
        BeforeRenderingDeferredLights = 230,
```

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
```diff
@@ -184,10 +206,48 @@ public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderin
...
+#if ENABLE_VR && ENABLE_XR_MODULE
+                int eyeCount = renderingData.cameraData.xr.enabled && renderingData.cameraData.xr.singlePassEnabled ? 2 : 1;
+#else
+                int eyeCount = 1;
+#endif
+                for (int eyeIndex = 0; eyeIndex < eyeCount; eyeIndex++)
+                {
+                    Matrix4x4 view = renderingData.cameraData.GetViewMatrix(eyeIndex);
+                    Matrix4x4 proj = renderingData.cameraData.GetProjectionMatrix(eyeIndex);
+                    m_CameraViewProjections[eyeIndex] = proj * view;
```
phi-lira, line 218: Why do we upload specific matrices here instead of using the camera matrices, `unity_CameraProjection` et al?        
https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl#L28-L32

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
```diff
@@ -270,6 +331,18 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+                    // If true, SSAO pass is inserted after opaque pass and is expected to modulate lighting result now.
+                    if (m_CurrentSettings.AfterOpaque)
+                    {
+                        // This implicitely also bind depth attachment. Explicitely binding m_Renderer.cameraDepthTarget does not work.
+                        cmd.SetRenderTarget(
+                            m_Renderer.cameraColorTarget,
+                            RenderBufferLoadAction.Load,
+                            RenderBufferStoreAction.Store
+                        );
+                        cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, (int)ShaderPasses.AfterOpaque);
```
phi-lira, line 344: Will this work in XR? @DennisDeRykeUnity @kaychang-unity please check XR path was tested with AfterOpaque

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -15,9 +15,14 @@ SAMPLER(sampler_BaseMap);
 SAMPLER(sampler_ScreenSpaceOcclusionTexture);
 
 // Params
-float4 _BlurOffset;
 float4 _SSAOParams;
 float4 _SourceSize;
+float4 _ProjectionParams2;
+float4x4 _CameraViewProjections[2]; // This is different from UNITY_MATRIX_VP (platform-agnostic projection matrix is used). Handle both non-XR and XR modes.
```
phi-lira, line 21: Yes, but how is it different  from unity_CameraProjection * unity_WorldToCamera?

phi-lira (approved):

phi-lira:
LGTM. Minor comments can be addressed in follow up PR if needed.

phi-lira:
Checking yamato...

phi-lira:
Possible failures introduced by this PR: 
 - SSAO test failures in iPhone: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fssao-deferred-support/.yamato%252Funiversal-iphone-metal.yml%2523Universal_iPhone_Metal_Standalone_il2cpp_Linear_trunk/4617153/job/pipeline
  iPhone tests are green on master: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Fall-universal.yml%2523PR_Universal_trunk/4617895/job/pipeline

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -303,8 +304,8 @@ internal GraphicsFormat GetGBufferFormat(int index)
         {
             if (index == GBufferAlbedoIndex) // sRGB albedo, materialFlags
                 return QualitySettings.activeColorSpace == ColorSpace.Linear ? GraphicsFormat.R8G8B8A8_SRGB : GraphicsFormat.R8G8B8A8_UNorm;
-            else if (index == GBufferSpecularMetallicIndex) // sRGB specular, [unused]
-                return QualitySettings.activeColorSpace == ColorSpace.Linear ? GraphicsFormat.R8G8B8A8_SRGB : GraphicsFormat.R8G8B8A8_UNorm;    
+            else if (index == GBufferSpecularMetallicIndex) // sRGB specular, reflectivity
+                return GraphicsFormat.R8G8B8A8_UNorm;
```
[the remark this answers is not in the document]
kaychang-unity: It's not working well when blending terrain layers (blending is done by group of 4 layers). The specular/reflectivity are better "added" in linear space and it generate less discrepancy with forward renderer output.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
kaychang-unity: I think we would execute it.
Gbuffer pass does not contain logic to clear the camera color attachment, it is implicitly done in the infamous `ScriptableRenderer.SetRenderPassAttachments()` method. So I expect the render pass to execute correctly, then gbuffer pass and deferred ligthing pass write on top of it without clearing camera color attachment.
Normally forward passes should be added with `BeforeRenderingOpaques`/`AfterRenderingOpaques`.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
kaychang-unity: We need to combined view proj matrix. Unity/XR team does not provide `unity_CameraViewProjection`, so I have to compute it manually. I talked to XR team about adding that but they don't think it is either high priority or really needed. I want to avoid calculating view proj matrix in shader code as the calculation runs per pixel, so I ended up having to write the extra scaffolding code to provide this matrix manually (as well as `unity_eyeIndex` mechanism, which works in both non XR and XR environment).

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
kaychang-unity: I believe @DennisDeRykeUnity tested it already.

The api `cmd.SetRenderTarget()` is not great when dealing with a RenderTargetIdentifier that point to combined color+depth surfarce. I don't think the behaviour is well defined.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
kaychang-unity: You could do this multiplication in shader code, but I would not recommend it.
I spent years optimizing post processing effects, and moving that computation out of the shader would be the first thing we do on day 1 of optimization task. It shouldn't be so much work to provide `unity_CameraViewProjection` instead.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
DennisDeRykeUnity: @phi-lira You asked me to "check XR path was tested with AfterOpaque" but I do not know what this means.  I tested Quest in Link mode (basically make the Quest act like a Rift, tethered to my PC via USB) using test scenes 142 and 144 (modified slightly to enable XRSDK+Oculus and pose driver for camera tracking).  Would I need to skim through the shader in those scenes and look for an AfterOpaque pass?

DennisDeRykeUnity:
> @DennisDeRykeUnity 144_SSAO_RenderToBackBuffer is indeed not correct on iPhone Metal, I will have a look tomorrow.
> What is the error about 142_SSAO_DepthNormal_Projection ?

I think my comment above was not clear.  I was trying to list the test scenes I used, not saying that they had errors.  I'll edit my previous comment to make it clearer.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
kaychang-unity: @DennisDeRykeUnity Ah I got it! "AfterOpaque" is a mode for SSAO where it is applied after opaque geometry lighting instead of before (the default). If you check the SSAO renderer feature UI, there is a checkbox "After Opaque" that will move SSAO application after opaque. The AO will be slightly different and "darker", but it is expected (because less physically correct).

DennisDeRykeUnity:
I am redacting my approval of this PR because entering Play mode with scene  142_SSAO_DepthNormal_Projection on Quest Link (i.e. Quest connected to PC via USB cable to act like a Rift HMD) using After Opaque SSAO  hangs the Unity Editor using the following settings + config + scene.   

* Settings
Player settings > Other settings > Script compilation > Scripting Define Symbols > ENABLE_RENDERING_PATH_UI.
Forward Renderer asset > Lighting > Renderer Path > Deferred.
I verified that the ForwardRenderer asset (which I set to Deferred) is the first item in the UniversalRPAsset > General > Renderer List.
For each Renderer in 140_SSAO > Renderers I enabled: 
 New Screen Space Ambient Occlusion (Screen Space Ambient Occlusion) > After Opaque. 

* Config
- Unity (trunk ed1e382e88f9)
- URP (universal/ssao-deferred-support 1f610e3)
- D3D11
- Single Pass Instancing
- Play mode 

* Scene
142_SSAO_DepthNormal_Projection

The HMD goes entirely black except for the Chaperone UI, and I am unable to close the Editor normally (had to End Task in Task Manager). 

I cannot find a stack trace in the Editor log.  

This hang does not occur using MockHMD; I have only seen it happen using the Oculus provider.  @kaychang-unity If you do not have access to a Rift or Quest HMD perhaps you can get help from XR Graphics devs.  Or send me debugging instructions.

kaychang-unity:
@DennisDeRykeUnity I don't have a Right or Quest HMD. Will need XR Graphics devs help for this...

kaychang-unity:
@phi-lira I think the test failure on iPhone might have been bad timing with another PR that normalize normals per pixels. That in turned affect bilateral blur for SSAO as it relies on comparing normals. It's looking better locally after I merged master again but I cannot run yamato tests anymore for iPhone because it fails to build (same error on master branch). More bad timing. I suppose we will have to wait until that gets fixed first.

kaychang-unity:
@phi-lira iphone metal tests are now green using custom trunk revision:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fssao-deferred-support/.yamato%252Funiversal-iphone-metal.yml%2523Universal_iPhone_Metal_Standalone_il2cpp_Linear_CUSTOM-REVISION/4638615/job

DennisDeRykeUnity (approved):
After discussing with Thomas Z. I repeated my After Opaque tests from yestaerday except I used the latest trunk, Unity (trunk 31233d44259e).  Now I no longer experience the Editor hang, and scene 142_SSAO_DepthNormal_Projection looks correct in Oculus Link in Play mode using Single Pass Instancing.  

I approve this PR.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
kaychang-unity: Renamed.

PR #2059, opened 2020-09-30
Implementing scalabity settings for planar resolutions

anisunity:
Added the scalabiity settings on the asset and the planar probe for the resolution parameter.
image
image

Testing status
Tested that the values were actually matching the one used for rendering.
Tested that the values displayed in the mono behavior matched the ones in the asset.
Tested that the High value was clamped to the atlas resolution.
HDRP tests are green at the exception of an IES test, not sure how it is related.


anisunity:
I still have to handle the migration

alelievr (approved):

anisunity:
image
I changed the text from PlanarReflectionResolutionXXXX to ResolutionXXXX as it was taking too mcuh space.

I also supported migration.

I think this PR should be tested by QA

sebastienlagarde (approved):

sebastienlagarde:
For QA: I test Anis PR. There is probleme with multi selection which will be fixed in a separate PR

PR #2060, opened 2020-09-30
Misc Render Graph Improvements

JulienIgnace-Unity:
### Purpose of this PR
Collection of small improvements for render graph.
- Split Debug Overlay in several passes to avoid unncessary resource lifetime prolongation
- Fixed decal overlay
- Added proper name to the backbuffer resource (for future graph viewer)
- Fixed debugger display for resource handles.
- Added meta data to resource handles to better check handles validity.
- Fixed a leak
- Improved spacing of resource log
- Renamed variables in the pool code.
- Refactored Debug Overlay into a class to better suit Render Graph needs
- Implemented Probe Volume debug overlay with render graph.

---
### Testing status
Did not change any workflow so relied on automated testing.


com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphObjectPool.cs
```diff
@@ -88,14 +88,14 @@ internal void ReleaseAllTempAlloc()
         // Regular pooling API. Only internal use for now
         internal T Get<T>() where T : new()
         {
-            var toto = SharedObjectPool<T>.sharedPool;
-            return toto.Get();
+            var pool = SharedObjectPool<T>.sharedPool;
```
sebastienlagarde, line 91: lol :)

sebastienlagarde (approved):

PR #1806, opened 2020-09-09
Improvements on the volumetrics user experience

anisunity:
After profiling and getting feedback from @iM0ve @AlixMi Pierre Yves (I don't know what is his github tag).  Here is a bunch of user-facing improvements plus some bug fixes that should help with controling the performance, the in-editor quality and making the setup more intuitive.

Fixed a Nan issue in the volumetric lighting filtering.
Fixed the second pass in the denoising being done on the wrong axis.
Removing the reprojection frame setting.
Making the volumetric scene view reprojection work.
Reduce the cost of the volumetric lighting gaussian denoising.
Adding the option to have only the directional light contributing to the volumetrics.
Adding a new configuration mode for screen resolution and volume slices.
Changing the maximal values for the volumetric resolution and slices.
Adding a more intuitive way of defining the denoising process of the volumetrics.
Changing what is in the advanced mode for volumetrics
Renaming user-facing parameter names.
Displaying an info message when the user triggers anistropy

### Testing status
Tests on the new template that i didn't introduce any regression.
Ran the HDRP tests locally, some were not green, even if not related to the changes.

PR #2062, opened 2020-09-30
Hdrp - Fix volumetric fog presets

anisunity:
This PR is done on top of the volumetric user experience PR https://github.com/Unity-Technologies/Graphics/pull/1806
It adds quality presets of the control mode, budget and depthratio.
I changed the volumes in the HDRP tests to put them back to custom in order to not break the tests.

Testing status
HDRP tests are green locally.
checked that the right values where passed when changing from one preset to an other and to the custom mode (on the template)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Sky/AtmosphericScattering/FogEditor.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Sky/AtmosphericScattering/FogEditor.cs
@@ -96,23 +96,26 @@ namespace UnityEditor.Rendering.HighDefinition
             }
 
             PropertyField(m_ColorMode);
-            EditorGUI.indentLevel++;
-            if (!m_ColorMode.value.hasMultipleDifferentValues && (FogColorMode)m_ColorMode.value.intValue == FogColorMode.ConstantColor)
-            {
-                PropertyField(m_Color);
-            }
-            else
-            {
-                PropertyField(m_Tint);
 
-                if (isInAdvancedMode)
+            using (new HDEditorUtils.IndentScope())
+            {
+                if (!m_ColorMode.value.hasMultipleDifferentValues &&
+                    (FogColorMode) m_ColorMode.value.intValue == FogColorMode.ConstantColor)
                 {
-                    PropertyField(m_MipFogNear);
-                    PropertyField(m_MipFogFar);
-                    PropertyField(m_MipFogMaxMip);
+                    PropertyField(m_Color);
+                }
+                else
+                {
+                    PropertyField(m_Tint);
+
+                    if (isInAdvancedMode)
+                    {
+                        PropertyField(m_MipFogNear);
+                        PropertyField(m_MipFogFar);
+                        PropertyField(m_MipFogMaxMip);
+                    }
                 }
             }
-            EditorGUI.indentLevel--;
 
             bool volumetricLightingAvailable = false;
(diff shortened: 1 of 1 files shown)


iM0ve (changes requested):
Its unclear what the feature does exactly from UX, see below issues:

1) Based on indentation, Quality should control 5 items. But only 3 are grayed out. Should quality setting affect Anisotropy? If not we should change identation

Identation

2) When changing quality, for example med to low. None of the settings are refreshed, resulting in uncertainty of what is actually changed. I think this will be reworked by UX team, because its not only for fog. But still mentioning since its an issue, maybe its worth waiting for the rework before making the quality presets?

Zft9sy9ldS

3) To find out what do the Quality features do exactly, usually one would go to the Lighting Quality section and check. But regarding Fog, there is no such setting. Again resulting in uncertainty of what values are affected and how. Unclear why quality levels are not found in the expected location.

QualitySettings

anisunity:
- The identation is fixed.
- The preview of the values works for almost no volume overrides (if none?), so it will not be working for this one.
- I added the display of the values on the asset 
image

iM0ve (changes requested):
My previous concerns were addressed. Found one more small issue. It is possible to set higher values than 1 in the quality settings, breaking fog if done so. We should limit to 0 - 1

NAjfYOQpro

anisunity:
@iM0ve can you please check the changes, i just pushed the fix for the review you made.

iM0ve (approved):
Issues fixed. A great addition, however, I dont think it should be hidden in advanced.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalLightingQualitySettings.cs
```diff
@@ -228,7 +241,12 @@ internal GlobalLightingQualitySettings()
         public int[] RTRDenoiserRadius = new int[s_QualitySettingCount];
 
         // TODO: Volumetric fog quality
-
+        /// <summary>Controls which control mode should be used to define the volumetric fog parameters.</summary>
+        public FogControl[] Fog_ControlMode = new FogControl[s_QualitySettingCount];
+        /// <summary>Controls the budget of the volumetric fog effect.</summary>
+        public float[] Fog_Budget = new float[s_QualitySettingCount];
+        /// <summary>Controls how the budget is shared between screen resolution and depth.</summary>
+        public float[] Fog_DepthRatio = new float[s_QualitySettingCount];
```
johnpars, line 249: Hey, so it seems we currently only added quality presets for the **Balance** mode. (**Fog Budget, DepthRatio**), however the **Manual** mode still has none. Is it intended? I think we will need to add them, to comply with the Quality UI pattern, otherwise it will be unclear to users if they change quality preset and nothing happens.

anisunity: Yes it is intended, the manual mode is not used for the preset as it is considered "Advanced" so having default values for it doesnt make sense to me.

johnpars:
Updated the editor to support some of the new quality UX functionality:

FogQuality

johnpars:
Also, there was a small general issue with Quality UX override state that has been fixed here: #2382

sebastienlagarde (approved):

PR #2070, opened 2020-10-01
[HDRP][Compositor] Fix issue with screen space UI not drawing

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1279272/

Screen space UI elements were not drawn in the graphics compositor.

**Why it was not working?**
It was not working because the compositor was internally (always) cloning the layer camera's, to make sure there are no problems when sharing the same source camera in two layers. However in the 2D UI elements, you have to specify a target camera, and after the clone, it was not valid anymore.

**How it was solved?**
I solved the issue by making sure that the layer is drawn with the originally specified camera (and not a cloned one),  unless the camera is shared with another layer. In this case (shared camera), the UI will still not work by design because of the cloning. The user has to use a unique camera per layer to get UI elements. 

---
### Testing status
- The compositor test is green
- Tested by adding UI text in the foreground and background layers of the repro project that is attached in the bug ticket, and the UI text is now visible:
image


com.unity.render-pipelines.high-definition/Runtime/Compositor/CompositionLayer.cs
```diff
@@ -210,39 +215,47 @@ public void Init(string layerID = "")
...
-                    hideFlags = HideFlags.HideInInspector | HideFlags.HideInHierarchy | HideFlags.HideAndDontSave
-                };
-                m_LayerCamera = newCameraGameObject.AddComponent<Camera>();
-                newCameraGameObject.AddComponent<HDAdditionalCameraData>();
-                CopyInternalCameraData();
-
-                m_LayerCamera.name = "Compositor" + layerID;
-                m_LayerCamera.gameObject.hideFlags = HideFlags.HideInInspector | HideFlags.HideInHierarchy | HideFlags.HideAndDontSave;
-                if(m_LayerCamera.tag == "MainCamera")
+                if (!compositor.IsThisCameraShared(m_Camera))
```
pmavridis, line 223: The only change here is the 3-line if statement that makes the clone only if the camera was shared. But the indentation of the code has changed, so github marked everything as changed! Not sure if it was possible to create a cleaner diff somehow.

sebastienlagarde: ok no worry

sebastienlagarde (approved):

sebastienlagarde:
Note: this will still need a QA pass and we should add test for this. Merging for 10.1 package.

remi-chapelain:
Had a look on another project to make sure that this is fixed since we had some issue with UI recently ! ✔️ 
Maybe it will be worth reporting this in the documentation in case someone else run into this issue ? @JordanL8 

> "If the camera is shared with another layer. In this case, the UI will still not work by design because of the cloning. The user has to use a unique camera per layer to get UI elements."

JordanL8:
Is that the same as this: https://github.com/Unity-Technologies/Graphics/pull/2151#discussion_r502400104 ?

remi-chapelain:
Yep, indeed ! Everything is fine then :)
Thanks for taking the time to find it !

pmavridis:
Note that in the other PR I have made the requirements for UI drawing a bit more strict: 
the layer should also not override any camera state (otherwise we clone the camera internally). But it is included in the documentation, so lets hope users will see it.

PR #2072, opened 2020-10-01
Fixed LOD mesh stripping and LOD Group UI

JulienIgnace-Unity:
### Purpose of this PR
This PR fixes sevevral LoD related issues.
- Fixed the LoD UI using the wrong LoD bias value (it now uses the default frame settings which is consistent with the scene view)
- LoD meshes are now correctly stripped from the build depending on the maximum LoD value contained in all HDRP assets used in the build.
- Updated the documentation with a section about LoD Management where it explains briefly the differences with builtin and stating that QualitySettings lod APIs are not supported anymore.

---
### Testing status
Did a test setup with a LOD GRoup with 4 LoDs. Each one was a different mesh asset (but not a cube/sphere otherwise they would not be stripped).
Added 3 quality levels to the project. 
- Low with specific HDRP asset containing Max lod of 2.
- Medium with no specific HDRP asset (used the default one from Graphics instead with max LoD of 1)
- High with specific HDRP asset containing max lod of 0.

In the subsequent tests I used the com.unity.build-report-inspector package (added with this url in packman) to check the stripping.
Go to window => "Open Last Build Report" after a build. Go to source assets and sort them by type to see the list of model assets contained in the build.
Did a few tests:
- Include all three quality levels: Expected result was that no mesh was stripped (because High max lod is 0).
- Remove high: Notice that lod 0 mesh is stripped.
- Remove High and Medium: Notice that lod 0 and 1 are stripped.
- Used a camera with a max lod of 1 while only having the Low settings available: Notice that lod 2 is used instead of 1 as expected.

@Unity-Technologies/gfx-qa-hdrp I did the above tests but LoD being a bit complex, I'd rather have someone else check as well that nothing was broken or that there is no remaining corner cases.


com.unity.render-pipelines.high-definition/Documentation~/Feature-Comparison.md
```diff
@@ -210,6 +210,13 @@ The tables that follow provide an overview of the **Features** that the High Def
 | Hair                        | Not supported                                 | Yes                                                          |
 | Fabric                      | Not supported                                 | Yes                                                          |
 
+## LOD Management
+In the builtin render pipeline, LOD is managed from the QualitySettings. Each quality setting can define a LOD Bias and a Maximum LOD value. As such, they are global to this quality setting and cannot be changed per camera. In HDRP, things are different. As described [here](HDRP-Asset.md) and [here](Frame-Settings.md), HDRP introduces the notion of scalability settings. This allows user to change the LOD settings per camera by using either predetermined values contained in the HDRP asset of the current quality level or overridden values.
```
JordanL8, line 214: In the Built-in Render Pipeline, you manage levels of detail (LOD) from the QualitySettings. Each quality setting defines a LOD Bias and a Maximum LOD value. As such, they are global to the quality setting and you cannot change them on a per camera basis. In HDRP, there are scalability settings that allow you to change the LOD settings per camera by using either predetermined values contained in the HDRP Asset of the current quality level or overridden values. For more information, see [HDRP Asset](HDRP-Asset.md) and [Frame Settings](Frame-Settings.md.

com.unity.render-pipelines.high-definition/Documentation~/Feature-Comparison.md
```diff
@@ -210,6 +210,13 @@ The tables that follow provide an overview of the **Features** that the High Def
...
+
+This has two consequences:
```
JordanL8, line 216: Managing LOD in this way has two consequences:

com.unity.render-pipelines.high-definition/Documentation~/Feature-Comparison.md
```diff
@@ -210,6 +210,13 @@ The tables that follow provide an overview of the **Features** that the High Def
...
+- Default LOD settings for a quality level are now stored in the HDRP asset instead of the Quality Settings.
```
JordanL8, line 217: HDRP asset > HDRP Asset

com.unity.render-pipelines.high-definition/Documentation~/Feature-Comparison.md
```diff
@@ -210,6 +210,13 @@ The tables that follow provide an overview of the **Features** that the High Def
...
+- Builtin APIs such as QualitySettings.lodBias or QualitySettings.maximumLODLevel are not supported anymore. Instead, users need to change them through the camera frame settings. Using them will have no effect at all.
```
JordanL8, line 218: - Built-in APIs such as QualitySettings.lodBias or QualitySettings.maximumLODLevel no longer work. Instead, you need to change them through the camera Frame Settings. If you use the Built-in APIs, they have no effect at all.

JordanL8 (changes requested):
Added doc suggestions

JordanL8 (approved):

com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessBuild.cs
```diff
@@ -38,16 +55,35 @@ public void OnPreprocessBuild(BuildReport report)
...
+            // Update all quality levels with the right max lod so that meshes can be stripped.
+            // We don't take lod bias into account because it can be overridden per camera.
+            int qualityLevelCount = QualitySettings.names.Length;
+            for (int i = 0; i < qualityLevelCount; ++i)
+            {
+                QualitySettings.SetQualityLevel(i, false);
+                var renderPipeline = QualitySettings.renderPipeline as HDRenderPipelineAsset;
+                if (renderPipeline != null)
+                {
+                    QualitySettings.maximumLODLevel = GetMinimumMaxLoDValue(renderPipeline);
```
sebastienlagarde, line 80: Don't we need to revert this value somewhere?

sebastienlagarde (approved):

sebastienlagarde:
merged for 10.1

com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessBuild.cs
JulienIgnace-Unity: Thought about this but actually the Quality Settings are hidden to the users anyway and now we state clearly in the doc that they are not supported so using this API would not yield any result (since we override it for each camera afterward).
Also, now when we build a player we override all the various quality level parameters anyway for mesh stripping.

sebastienlagarde: I know it is not an issue currently, but with pipeline interoperability, if you switch back to urp after having build HDRP, settings can be a mess. but let's not deal with it for now (but poke jenny about it :) )

JulienIgnace-Unity: Yeah, exact same problem as the skybox material. As long as we don't have proper APIs for that we are doomed to have the same issues everywhere.

TomasKiniulis:
Seems to be working as expected
- Made sure meshes are not missing running player on different quality levels
- Made sure that the right meshes are stripped when there are multiple quality levels set
- Made sure that the right meshes are stripped when Maximum LOD levels are set in camera frame setting overrides

PR #2074, opened 2020-10-01
Fix "ToSubgraph" with multiple outputs and allow passthrough

elizabeth-legros:
Had to fix how we create the subgraph when using the contextual "ToSubgraph" option in the MaterialGraphEditWindow. Depends on fixes to property copying from https://github.com/Unity-Technologies/Graphics/pull/1932/

for https://fogbugz.unity3d.com/f/cases/1256975/

redid repro steps and could not get bug to occur


cdxntchou (approved):

ghost (approved):
Ran the following mindmap on subgraphs
Subgraph smoke test
Only found 1 bug and that bug also reproduced on Master.
Approving.

ghost (changes requested):
Actually, "convert to subgraph" no longer detecting recursions is a regression, now that I checked. This should be fixed.

ghost (approved):
Tested again to make sure I'm not wasting your time, and it does detect recursion with convert to subgraph. Must have made a mistake. Reapproving

marctem (approved):

PR #2075, opened 2020-10-01
Universal/deferred fixed

kaychang-unity:
---
### Purpose of this PR
Fixed various bugs with deferred rendering modes.
Some of the fixes are in the universal/rename-standard-renderer branch, but since it is unclear when the former branch can be merged, I duplicated these fixes in the current branch.
- Fixed MSAA incorrectly enabled in some situations when deferred rendering is enabled
- Fixed dynamic resolution
- incorrect shader pass selection in some platforms (due to pass stripping)
- lighting buffer output alpha=1.0 instead of 0.0 when lighting buffer is a non-HDR format
- 014_CameraMulti_MiniMap error threshold adjusted for deferred
- Mixed lighting shadowmask support (043_Lighting_Mixed_ShadowMask)
- Mixed lighting subtractive fix (049_Lighting_Mixed_Subtractive)

---
### Testing status
UniversalGraphicsTests are passing locally, ignoring some error threshold delta due to local development machine specs.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
+++ b/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
@@ -458,7 +458,7 @@ namespace UnityEngine.Rendering.Universal.Internal
             m_StencilDeferredMaterial.SetInt(ShaderConstants._ClearStencilReadMask, (int)StencilUsage.MaterialMask);
             m_StencilDeferredMaterial.SetInt(ShaderConstants._ClearStencilWriteMask, (int)StencilUsage.MaterialMask);
 
-            // Compute some platform limits.
+            // Compute some platform limits (for deferred tiling).
             m_MaxDepthRangePerBatch = (DeferredConfig.UseCBufferForDepthRange ? DeferredConfig.kPreferredCBufferSize : DeferredConfig.kPreferredStructuredBufferSize) / sizeof(uint);
             m_MaxTilesPerBatch = (DeferredConfig.UseCBufferForTileList ? DeferredConfig.kPreferredCBufferSize : DeferredConfig.kPreferredStructuredBufferSize) / System.Runtime.InteropServices.Marshal.SizeOf(typeof(TileData));
             m_MaxPunctualLightPerBatch = (DeferredConfig.UseCBufferForLightData ? DeferredConfig.kPreferredCBufferSize : DeferredConfig.kPreferredStructuredBufferSize) / System.Runtime.InteropServices.Marshal.SizeOf(typeof(PunctualLightData));
@@ -1720,12 +1720,11 @@ namespace UnityEngine.Rendering.Universal.Internal
             Vector4 lightPos, lightColor, lightAttenuation, lightSpotDir, lightOcclusionChannel;
             UniversalRenderPipeline.InitializeLightConstants_Common(visibleLights, index, out lightPos, out lightColor, out lightAttenuation, out lightSpotDir, out lightOcclusionChannel);
 
-            punctualLightBuffer[storeIndex * 6 + 0] = new uint4(FloatToUInt(lightPos.x), FloatToUInt(lightPos.y), FloatToUInt(lightPos.z), FloatToUInt(visibleLights[index].range * visibleLights[index].range));
-            punctualLightBuffer[storeIndex * 6 + 1] = new uint4(FloatToUInt(lightColor.x), FloatToUInt(lightColor.y), FloatToUInt(lightColor.z), 0);
-            punctualLightBuffer[storeIndex * 6 + 2] = new uint4(FloatToUInt(lightAttenuation.x), FloatToUInt(lightAttenuation.y), FloatToUInt(lightAttenuation.z), FloatToUInt(lightAttenuation.w));
-            punctualLightBuffer[storeIndex * 6 + 3] = new uint4(FloatToUInt(lightSpotDir.x), FloatToUInt(lightSpotDir.y), FloatToUInt(lightSpotDir.z), (uint)lightFlags);
-            punctualLightBuffer[storeIndex * 6 + 4] = new uint4(FloatToUInt(lightOcclusionChannel.x), FloatToUInt(lightOcclusionChannel.y), FloatToUInt(lightOcclusionChannel.z), FloatToUInt(lightOcclusionChannel.w));
-            punctualLightBuffer[storeIndex * 6 + 5] = new uint4(0, 0, 0, 0);
+            punctualLightBuffer[storeIndex * 5 + 0] = new uint4(FloatToUInt(lightPos.x), FloatToUInt(lightPos.y), FloatToUInt(lightPos.z), FloatToUInt(visibleLights[index].range * visibleLights[index].range));
+            punctualLightBuffer[storeIndex * 5 + 1] = new uint4(FloatToUInt(lightColor.x), FloatToUInt(lightColor.y), FloatToUInt(lightColor.z), 0);
+            punctualLightBuffer[storeIndex * 5 + 2] = new uint4(FloatToUInt(lightAttenuation.x), FloatToUInt(lightAttenuation.y), FloatToUInt(lightAttenuation.z), FloatToUInt(lightAttenuation.w));
+            punctualLightBuffer[storeIndex * 5 + 3] = new uint4(FloatToUInt(lightSpotDir.x), FloatToUInt(lightSpotDir.y), FloatToUInt(lightSpotDir.z), (uint)lightFlags);
+            punctualLightBuffer[storeIndex * 5 + 4] = new uint4(FloatToUInt(lightOcclusionChannel.x), FloatToUInt(lightOcclusionChannel.y), FloatToUInt(lightOcclusionChannel.z), FloatToUInt(lightOcclusionChannel.w));
         }
 
         void StoreTileData(ref NativeArray<uint4> tileList, int storeIndex, uint tileID, uint listBitMask, ushort relLightOffset, ushort lightCount)
--- a/com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
+++ b/com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
@@ -34,22 +34,20 @@
 #define USE_CBUFFER_FOR_LIGHTLIST 0
 #endif
 
+// This structure is used in StructuredBuffer.
+// TODO move some of the properties to half storage (color, attenuation, spotDirection, flag to 16bits, occlusionProbeInfo)
 struct PunctualLightData
 {
     float3 posWS;
-    float radius2;           // squared radius
-    half4 color;
-    half4 attenuation;       // .xy are used by DistanceAttenuation - .zw are used by AngleAttenuation (for SpotLights)
-    half3 spotDirection;     // spotLights support
-    int flags;               // Light flags (enum kLightFlags and LightFlag in C# code)
-    half4 occlusionProbeInfo;
-    int shadowLightIndex;
-    int spare0;
-    int spare1;
-    int spare2;
+    float radius2;              // squared radius
+    float4 color;
+    float4 attenuation;         // .xy are used by DistanceAttenuation - .zw are used by AngleAttenuation (for SpotLights)
+    float3 spotDirection;       // spotLights support
+    int flags;                  // Light flags (enum kLightFlags and LightFlag in C# code)
+    float4 occlusionProbeInfo;
 };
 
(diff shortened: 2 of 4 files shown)


com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -454,18 +499,20 @@ internal void SetupLights(ScriptableRenderContext context, ref RenderingData ren
 
             DeferredShaderData.instance.ResetBuffers();
 
-            this.RenderWidth = renderingData.cameraData.cameraTargetDescriptor.width;
-            this.RenderHeight = renderingData.cameraData.cameraTargetDescriptor.height;
+            Camera camera = renderingData.cameraData.camera;
+            // Support for dynamic resolution.
+            this.RenderWidth = camera.allowDynamicResolution ? Mathf.CeilToInt(ScalableBufferManager.widthScaleFactor * camera.pixelWidth) : camera.pixelWidth;
```
phi-lira, line 504: Should we apply the render scale from the pipeline asset here? I think the targetDescriptor.width differs from camera.pixelWidth due to that.

kaychang-unity: It doesn't work for scene 127_DynamicScaling_DOF. This scene sets the dynamic resolution directly via `ScalableBufferManager` (lowest C# interface for dynamic resolution), so bypasses `UniversalRenderPipelineAsset.renderScale`.
`cameraData.renderScale` and `targetDescriptor.width` derive their values from `UniversalRenderPipelineAsset.renderScale`.

kaychang-unity:
Note that current Yamato tests do not have deferred rendering mode enabled, so only forward renderer path is effectively tested.
To enable deferred renderer, add`ENABLE_RENDERING_PATH_UI` as a new scripting symbol (expose rendering mode UI menu), then set forwardRenderer to deferred rendering mode.

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
```diff
@@ -87,9 +128,11 @@ FragmentOutput SurfaceDataToGbuffer(SurfaceData surfaceData, InputData inputData
     output.GBuffer0 = half4(surfaceData.albedo.rgb, PackMaterialFlags(materialFlags));   // albedo          albedo          albedo          materialFlags   (sRGB rendertarget)
     output.GBuffer1 = half4(surfaceData.specular.rgb, 0);                                // specular        specular        specular        [unused]        (sRGB rendertarget)
     output.GBuffer2 = half4(packedNormalWS, packedSmoothness);                           // encoded-normal  encoded-normal  encoded-normal  packed-smoothness
-    output.GBuffer3 = half4(globalIllumination, 0);                                      // GI              GI              GI              [not_available] (lighting buffer)
-    #if USE_SHADOWMASK
+    output.GBuffer3 = half4(globalIllumination, 1);                                      // GI              GI              GI              [optional: see OutputAlpha()] (lighting buffer)
+    #if OUTPUT_SHADOWMASK == 1
     output.GBuffer4 = unity_ProbesOcclusion;
```
lukaschod, line 133: There is no need for this logic. inputData.shadowMask will have unity_ProbesOcclusion value if subtractive lighting is used. Basically inputData.shadowMask has baked shadows now

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
```diff
@@ -186,11 +186,9 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
             {
                 #if defined(_MAIN_LIGHT_SHADOWS)
                     float4 shadowCoord = TransformWorldToShadowCoord(posWS.xyz);
-                    unityLight.shadowAttenuation = MainLightRealtimeShadow(shadowCoord);
-                    unityLight.shadowAttenuation = ApplyShadowFade(unityLight.shadowAttenuation, posWS.xyz);
+                    unityLight.shadowAttenuation = MainLightShadow(shadowCoord, posWS.xyz, shadowMask, _MainLightOcclusionProbes);
```
lukaschod, line 189: Is there a difference between _LightOcclusionProbInfo and _MainLightOcclusionProbes?

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -1020,8 +1080,7 @@ internal void ExecuteDeferredPass(ScriptableRenderContext context, ref Rendering
                 // This does 2 things:
                 // - baked geometry are skipped (do not receive dynamic lighting)
                 // - non-baked geometry (== non-static geometry) use shadowMask/occlusionProbes to emulate baked shadows influences.
-                if (renderingData.lightData.supportsMixedLighting && this.MixedLightingSetup == MixedLightingSetup.Subtractive)
-                    cmd.EnableShaderKeyword(ShaderKeywordStrings._DEFERRED_SUBTRACTIVE_LIGHTING);
+                CoreUtils.SetKeyword(cmd, ShaderKeywordStrings._DEFERRED_MIXED_LIGHTING, this.MixedLightingSetup != MixedLightingSetup.None);
```
lukaschod, line 1083: Use UseShadowMask property here

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
lukaschod: Where do you set _MainLightOcclusionProbes? I can not find it in https://github.com/Unity-Technologies/Graphics/blob/c6b2a6fa7d598650b4d61c63aef424bcbedb4f05/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
kaychang-unity: Done.

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
```diff
@@ -164,9 +188,11 @@ FragmentOutput BRDFDataToGbuffer(BRDFData brdfData, InputData inputData, half sm
     output.GBuffer0 = half4(brdfData.diffuse.rgb, PackMaterialFlags(materialFlags)); // diffuse         diffuse         diffuse         materialFlags   (sRGB rendertarget)
     output.GBuffer1 = half4(specular, brdfData.reflectivity);                        // specular        specular        specular        reflectivity    (sRGB rendertarget)
     output.GBuffer2 = half4(packedNormalWS, packedSmoothness);                       // encoded-normal  encoded-normal  encoded-normal  smoothness
-    output.GBuffer3 = half4(globalIllumination, 0);                                  // GI              GI              GI              [not_available] (lighting buffer)
-    #if USE_SHADOWMASK
+    output.GBuffer3 = half4(globalIllumination, 1);                                  // GI              GI              GI              [optional: see OutputAlpha()] (lighting buffer)
+    #if OUTPUT_SHADOWMASK == 1
     output.GBuffer4 = unity_ProbesOcclusion;
+    #elif OUTPUT_SHADOWMASK == 2
+    output.GBuffer4 = inputData.shadowMask;
```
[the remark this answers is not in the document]
kaychang-unity: Done.

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
kaychang-unity: Done.

lukaschod (approved):
Noicee

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
```diff
@@ -31,7 +36,7 @@ struct FragmentOutput
     half4 GBuffer1 : SV_Target1;
     half4 GBuffer2 : SV_Target2;
     half4 GBuffer3 : SV_Target3; // Camera color attachment
-    #if USE_SHADOWMASK
+    #if OUTPUT_SHADOWMASK
     half4 GBuffer4 : SV_Target4;
```
phi-lira, line 40: Do we do any validation for platforms that don't support more than 4MRT?

com.unity.render-pipelines.universal/Shaders/Utils/TileDeferred.shader
```diff
@@ -329,9 +339,14 @@ Shader "Hidden/Universal Render Pipeline/TileDeferred"
...
             #pragma target 4.5
 
+            #pragma multi_compile_fragment _ _DEFERRED_ADDITIONAL_LIGHT_SHADOWS
+            #pragma multi_compile_fragment _ _SHADOWS_SOFT
+            #pragma multi_compile_fragment _ LIGHTMAP_SHADOW_MIXING
+            #pragma multi_compile_fragment _ SHADOWS_SHADOWMASK
+
             #pragma multi_compile_fragment _LIT
             #pragma multi_compile_fragment _ _GBUFFER_NORMALS_OCT
-            #pragma multi_compile_fragment _ _DEFERRED_SUBTRACTIVE_LIGHTING
```
phi-lira, line 334: This could be a breaking change for custom shaders

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
```diff
@@ -44,6 +44,9 @@ struct PunctualLightData
     int flags;               // Light flags (enum kLightFlags and LightFlag in C# code)
     half4 occlusionProbeInfo;
     int shadowLightIndex;
+    int spare0;
```
phi-lira, line 47: Do you know if we can define packing rules for this? Otherwise we will have to rename this vars if we give meaning to them later and that could be a considered a breaking change.

phi-lira (approved):
LGTM. Minor comments.

phi-lira:
Adding  GFX Foundation QA to check if we are not regressing any behaviour.

cinight (approved):
lgtm

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
kaychang-unity: Yes. it will automatically goes back to forward renderer if the platform does not support the required number of RTs. QA already raised and tested it.
On a side note, the shadowMask add a 4th RT. RenderPass will add another RT in order to store depth-as-color. This means up to 6 RTs will be needed on mobiles.

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
kaychang-unity: I have a new version of this file without the `spare0` member. I didn't submit the new code because I was waiting for this PR to get through since it already passed all Yamato tests.
Generally, I don't think URP has an API to declare the fields of the struct with correct packing/alignment, has it? Such API would also need to know how to handle half types as storage.

com.unity.render-pipelines.universal/Shaders/Utils/TileDeferred.shader
kaychang-unity: I renamed `_DEFERRED_SUBTRACTIVE_LIGHTING` into `_DEFERRED_MIXED_LIGHTING` This #define is only used by the deferred shaders (stendilDeferred.shader and TileDeferred.shader), and since there has been no official release yet, it should not break anybody.

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
kaychang-unity: I pushed a new version without the `spare0` field.
I promoted back some fields from half to float, as half in structured buffer is considered 16bits (which is not the case in cbuffer on non-metal platforms). Eventually, we need an API to declare shared structures between C# code and shader code that deal with automatic alignment and half storage types.
All these changes will not affect stencil deferred. I locally tested UniversalGraphicsTest with forward and deferred renderer and all tests are passing as usual.

phi-lira:
VFX Graph failure, I'm not sure if caused by this PR, but master branch is green: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Fvfx_urp-win-dx11.yml%2523VFX_URP_Win_DX11_playmode_mono_apiNet2_Linear_trunk/3913591/job

phi-lira:
VFX Graph Failure is known instability.

PR #2076, opened 2020-10-02
Fix rendergraph on consoles when motion vectors are enabled

FrancescoC-unity:
My guess is that the issue was always there however most of the time I reproed locally was showing as few small black spots around and probably passing the threshold on the test. 

The issue was that when motion vectors are enabled, the rendergraph version does not correctly set the multi compile, as a consequence the motion vectors are never resolved. 

The reason why it is passing on PC is because on PC resources are black by default, hence neutral motion vectors, however on consoles they contain garbage.  Because of the random content, it fails differently at different times. 

Also fixed the format of the AO history. 

Locally all looks fixed now, I am about to kick the yamato to verify there too.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
@@ -891,7 +891,7 @@ namespace UnityEngine.Rendering.HighDefinition
 
             public RTHandle Allocator(string id, int frameIndex, RTHandleSystem rtHandleSystem)
             {
-                return rtHandleSystem.Alloc(Vector2.one * scaleFactor, TextureXR.slices, filterMode: FilterMode.Point, colorFormat: GraphicsFormat.R32_SFloat, dimension: TextureXR.dimension, useDynamicScale: true, enableRandomWrite: true, name: string.Format("{0}_AO Packed history_{1}", id, frameIndex));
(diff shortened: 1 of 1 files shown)


JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -891,7 +891,7 @@ struct AmbientOcclusionAllocator
 
             public RTHandle Allocator(string id, int frameIndex, RTHandleSystem rtHandleSystem)
             {
-                return rtHandleSystem.Alloc(Vector2.one * scaleFactor, TextureXR.slices, filterMode: FilterMode.Point, colorFormat: GraphicsFormat.R32_UInt, dimension: TextureXR.dimension, useDynamicScale: true, enableRandomWrite: true, name: string.Format("{0}_AO Packed history_{1}", id, frameIndex));
+                return rtHandleSystem.Alloc(Vector2.one * scaleFactor, TextureXR.slices, filterMode: FilterMode.Point, colorFormat: GraphicsFormat.R32_SFloat, dimension: TextureXR.dimension, useDynamicScale: true, enableRandomWrite: true, name: string.Format("{0}_AO Packed history_{1}", id, frameIndex));
```
sebastienlagarde, line 894: Haven't we already discuss this exact opposite change in another PR in the past?

FrancescoC-unity: You are right, my bad here.

The reason why it is confusing is that history is uint and main buffer is float.

sebastienlagarde (approved):

PR #2079, opened 2020-10-02
Meshes with decal material not added to RT acceleration structure anymore

eturquin:
This addresses the following bug report: https://fogbugz.unity3d.com/f/cases/1264765/

Small change for ray and path tracing:
* Before: meshes assigned with a decal material would be added in the acceleration structure, so be intersected by rays, but then appear black because the material is not handled
* After: meshes assigned with a decal material are not added anymore in the acceleration structure, so not visible to rays


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
```diff
@@ -221,8 +221,8 @@ void RaytracingManagerCleanupNonRenderGraphResources()
                     // Grab the material for the current sub-mesh
                     Material currentMaterial = materialArray[meshIdx];
 
-                    // The material is transparent if either it has the requested keyword or is in the transparent queue range
-                    if (currentMaterial != null)
+                    // Make sure that the material is both non-null and non-decal
+                    if (currentMaterial != null && !DecalSystem.IsHDRenderPipelineDecal(currentMaterial))
```
sebastienlagarde, line 225: the correct functoin to use is IsDecalMaterial, this function IsHDRenderPipelineDecal will not work with decal shader graph

sebastienlagarde (approved):

remi-chapelain:
Tested on the template ✔️

PR #2081, opened 2020-10-02
Fix alpha output by forward opaque materials

FrancescoC-unity:
Forward materials output the alpha sampled from the alpha channel of the base color texture even if opaque. This lead to errors depending on the import settings of the texture. 

We instead hard code a fully opaque output when opaque. 

Note: the issue with motion blur reported on the forum I can't repro, which makes me think was fixed in later packages.
Also, fog still outputs what looks like a color in the alpha channel and this is expected at the moment, it will need bigger changes to restore the alpha.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -117,6 +117,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed issue with box light not visible if range is below one and range attenuation is off.
 - Fixed alpha not having TAA applied to it.
 - Fixed issue with bloom showing a thin black line after rescaling window. 
+- Fix issue with alpha output in forward.
 
 ### Changed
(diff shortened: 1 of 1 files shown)


pmavridis (approved):
Indeed, motion blur + alpha was fixed a few months ago by me.

I'm a bit afraid that this PR might break a project that was forward-only and (for example) was using a texture with alpha on a quad + some blending (to create something like a custom decal). But I don't have a better suggestion.

FrancescoC-unity:
> Indeed, motion blur + alpha was fixed a few months ago by me.
> 
> I'm a bit afraid that this PR might break a project that was forward-only and (for example) was using a texture with alpha on a quad + some blending (to create something like a custom decal). But I don't have a better suggestion.

That would still work, this replace happens _just before_ output not during the shader processing.  If they need to output alpha, then they shouldn't use opaque anyway.

sebastienlagarde:
closing, integrated into this PR: https://github.com/Unity-Technologies/Graphics/pull/2083

PR #2083, opened 2020-10-02
[HDRP][Compositor] Fix issues with bloom, alpha and HDR layers in the compositor

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1272621/

This bug revealed a few issues in the compositor:
- The alpha value was using to guide the application of bloom. However bloom should spread over the entire RT, even at regions with zero alpha. This is the case now.
- The composition layers were created as UNORM resources, so HDR values were clamped. I have changed that to SFloat. 

This PR also includes:
- Some updates on the alpha output from @FrancescoC-unity (with a minor change to fix the alpha to 1 only when alpha testing is NOT performed, as discussed offline. The TAA fix was fine but we had to update the reference image)
- An updated compositor test scene to test UI rendering and AOV stacking (+ updated reference image)

---
### Testing status
After some minor modifications to the repro project in the bug ticket, applying bloom on the cubes now gives the same result:

image

The modifications to make this work are the following:
- Dithering should be disabled in the background camera (because it clamps the values to [0,1] )
- When using camera stacking, as with the repro project, bloom should only be applied on the last layer of the stack (here the foreground). For the background layer I have created a new volume with only an exposure override (set to the same exposure as foreground) but not bloom.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -140,7 +140,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed backface selection on some shader not ignore correctly.
 - Disable quad overdraw on ps4.
 - Fixed error when resizing the graphics compositor's output and when re-adding a compositor in the scene
-- Fixed issue with screen space UI not drawing on the graphics compositor (case 1279272).
 - Fixed issues with bloom, alpha and HDR layers in the compositor (case 1272621).
 - Fixed alpha not having TAA applied to it.
 - Fix issue with alpha output in forward.


Soaryn:
Having looked over the report image, I noticed that it came from my report. I'm hoping I misunderstood something with the potential fix here, but I wanted to ensure a use case was achievable:

Having 3 layers, 
- A: Post Processed Background as a default back drop.
- B: No post processing for external capture sources that renders over the backdrop.
- C: Post Processed overlay with bloom that renders on top of all of the aforementioned sources.

The backdrop is purely there to be in case the capture source is not present, but can still render high end graphic such as a VFX animated wall. The Capture sources are coming in via a capture card and read into a Render Texture which can then be used anywhere in Layer B (currently as quads to allow reshaping) without color modifications such as tone mapping or emitting its own bloom values. The last layer is to allow a dynamic overlay with bloom and other post processes to overlay onto the capture. It is important to have bloom from layer C to show on Layer A or B, but I want to ensure that Layer B itself does not create a bloom.

 I don't mind elaborating more, but I'd prefer to have a vocal conversation with someone regarding this. The compositor thus far seems to be a nice broadcasting production solution, but some things are slightly issued. I am currently using it in fact, but restricted by issues that have already been fixed such as the screen flickering internally, but not fixed publicly.

pmavridis:
> The last layer is to allow a dynamic overlay with bloom and other post processes to overlay onto the capture. It is important to have bloom from layer C to show on Layer A or B, but I want to ensure that Layer B itself does not create a bloom.

First of all, thanks for using the Graphics Compositor and for all the feedback. We really appreciate that.

Yes, you will be able to do what you want (after my last update). The alpha channel is used to control which areas of the image create bloom/glow. So on your last layer (Layer C in your example) make sure the "clear alpha" checkbox is selected. 

Here is a screenshot on how this looks with the cubes:
image
In this example, only the last layer has bloom enabled. 
- On the left side, when the "clear alpha" option is not selected, the alpha channel includes the regions that were drawn from previous layers, so both cubes create alpha. 
- On the right side, the alpha is cleared before the last layer is drawn, therefore only the last cube/layer is creating glow. 

So I think this should cover your use case. If you have any other concerns, please let use know.

remi-chapelain:
Check with a fresh project to understand what was happening ✔️

Soaryn:
> > The last layer is to allow a dynamic overlay with bloom and other post processes to overlay onto the capture. It is important to have bloom from layer C to show on Layer A or B, but I want to ensure that Layer B itself does not create a bloom.
> 
> First of all, thanks for using the Graphics Compositor and for all the feedback. We really appreciate that.
> 
> Yes, you will be able to do what you want (after my last update). The alpha channel is used to control which areas of the image create bloom/glow. So on your last layer (Layer C in your example) make sure the "clear alpha" checkbox is selected.
> 
> Here is a screenshot on how this looks with the cubes:
> image
> In this example, only the foreground layer has bloom enabled.
> 
> * On the left side, when the "clear alpha" option is not selected, the alpha channel includes the regions that were drawn from previous layers, so both cubes create glow because they are in the alpha channel.
> * On the right side, the alpha is cleared before the last layer is drawn, therefore only the last cube/layer is creating glow.
> 
> So I think this should cover your use case. If you have any other concerns, please let us know.

So now that 10.1 is available via the package manager I was able to sit down and try it out. Back to the use case at hand how would I achieve what I asked earlier? Because if I wipe everything out by clearing the alpha on the last layer, the Background layer is not able to bloom. The use case is that both the Foreground AND the background would have the same post processing FX, but the capturable layer (the layer between) should not have fx. Because the layer is in between the 2 Post processed layers clearing the alpha of the previous layers seems like not the correct approach.

In my attempt this is what I was able to achieve and this is not quite what was desired:
image

pmavridis:
> So now that 10.1 is available via the package manager I was able to sit down and try it out. Back to the use case at hand how would I achieve what I asked earlier? 

You have many options. Which one is better depends a lot on your content. In particular:
- If the capturable layer has depth (ie you display the video on a quad or other surface with depth), you can just move it on the top of your render schedule. The stacking of the sub-layers is depth-aware, so the order in the list is only important for transparency/translucency effects, otherwise the output will be the same even if you re-order the layers.. 
In this case you have to clear the alpha on the background layer (which will be in the middle of the list) and then enable bloom on the Foreground layer (without alpha clear), which will be the last rendered.
If your video has alpha/transparency (through chroma keying for example), and you expect to see the background behind your video, then with this approach you will be limited to alpha clipping. (You cannot render the video as transparent surface because the background will be rendered later)

- If the order of layers is important (for transparency for example), then perhaps the best option is to 
a) all layers keep (not clear) the alpha 
b) the sub-layer in-between with the capturable should output zero alpha (either for the entire screen, or for specific regions you don't want bloom and other post-processing to be applied). You can override the alpha to zero by adding an Alpha Mask "input filter" in the sub-layer, and then selecting a black texture. If you want more flexibility you can probably achieve the same thing in other ways (by drawing "invisible" objects with a material that writes alpha but not color or using a custom pass)

- Of course another option is to enable bloom in both background and the foreground (all layers clear the alpha). This is my least favorite option, because you will pay the cost of bloom two times, and any objects in the capture/foreground layer will be drawn on top of the bloom from the background layer, so you will have discontinuities. Generally I would try to make the second option work.

- Finally, the option that gives you the most flexibility (depending again on the content) might be to add three composition layers in the compositor (by adding two more Texture2D input properties in the composition graph - in this case the Compositor Window will automatically show the extra layers and you control the mixing with the Shader Graph). This way you have full control on the way the layers are mixed/blended together and you could even implement the blending modes that you know from PhotoShop, use a black-and-white version of a layer as alpha etc...

PR #2089, opened 2020-10-05
Changed default property value to be True for HDR

martint-unity:
---
### Purpose of this PR
When creating a new pipeline asset we now set the HDR property to true and not false.


phi-lira (approved):

phi-lira:
This should not affect ABV but waiting for it to pass anyway.

phi-lira:
Failure on `test all packages` is due to HDRP, not related to this PR.

PR #2101, opened 2020-10-06
Add option to use "Fast" versions of conversions between SRGB and Linear

skerys:
---
### Purpose of this PR
Adding an option to use "Fast" approximation functions for SRGB to Linear conversions and vice versa. Mostly aimed at low-end mobile devices where the expensive conversion functions were causing a significant performance hit (especially in color grading with a user LUT, where they are done even when using Linear color space in the Project settings).

---
**Manual Tests**: 
Tested locally on DX11, Metal, and Vulkan (Android)

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/mobile%252Ffast-srgb-linear-conversions

---

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
@@ -684,7 +684,6 @@ namespace UnityEngine.Rendering.Universal
             set { m_ColorGradingLutSize = Mathf.Clamp(value, k_MinLutSize, k_MaxLutSize); }
         }
 
-
         /// <summary>
         /// Returns true if fast approximation functions are used when converting between the sRGB and Linear color spaces, false otherwise.
         /// </summary>
--- a/.yamato/_formatting.yml
+++ b/.yamato/_formatting.yml
@@ -18,3 +18,5 @@ formatting:
       -  hg clone -u stable http://hg-mirror-slo.hq.unity3d.com/unity-extra/unity-meta ~/unity-meta
       -  perl ~/unity-meta/Tools/Format/format.pl --hgroot $(pwd) --dry-run .
     timeout: 1
+    triggers:
+        expression: pull_request.target eq "master"
--- a/.yamato/ruamel/jobs/formatting/formatting.py
+++ b/.yamato/ruamel/jobs/formatting/formatting.py
@@ -24,4 +24,5 @@ class Formatting_Job():
         job.set_agent(metafile['formatting_agent'])
         job.add_commands(commands)
         job.set_timeout(metafile['timeout'])
+        job.set_trigger_on_expression(f'pull_request.target eq "{metafile["target_branch"]}"')
         return job


skerys:
@Unity-Technologies/mobile-qa Please test visual correctness for Bloom, Bokeh Depth of Field and Dithering on Gamma color space

Also test visual correctness of the Color Lookup Post Processing effect on both Linear and Gamma

Chman:
General feedback:

Instead of having `#if`/`#else` everywhere and duplicated code (which is error-prone), can we just wrap it in a function in `Common.hlsl`? It would clean up a lot of code, reduce the amount of noise and simplify everything.

Chman (approved):

aleksandrasdzikia (approved):
Tested visual correctness for Bloom, Bokeh Depth of Field and Dithering on Gamma colour space as well as Colour Lookup Post Processing effect on both Linear and Gamma. The project used: universal rendering examples, 3d skybox scene was used with post processing. Built on Android (Vulkan) and iOS (Metal) - no effect on visuals.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -683,6 +684,12 @@ public int colorGradingLutSize
             set { m_ColorGradingLutSize = Mathf.Clamp(value, k_MinLutSize, k_MaxLutSize); }
         }
 
+        public bool useFastSRGBLinearConversion
```
phi-lira, line 687: All public API needs API docs, or consider making it internal/private if public is not intended.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -285,6 +285,7 @@ public struct PostProcessingData
     {
         public ColorGradingMode gradingMode;
         public int lutSize;
+        public bool useFastSRGBLinearConversion;
```
phi-lira, line 288: All new public API must be documented or make it internal.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -305,6 +306,7 @@ public static class ShaderKeywordStrings
         public static readonly string DepthMsaa8 = "_DEPTH_MSAA_8";
 
         public static readonly string LinearToSRGBConversion = "_LINEAR_TO_SRGB_CONVERSION";
+        public static readonly string UseFastSRGBLinearConversion = "_USE_FAST_SRGB_LINEAR_CONVERSION";
```
phi-lira, line 309: either make it internal or document it.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
```diff
@@ -9,6 +9,7 @@ Shader "Hidden/Universal Render Pipeline/UberPost"
         #pragma multi_compile_local_fragment _ _FILM_GRAIN
         #pragma multi_compile_local_fragment _ _DITHERING
         #pragma multi_compile_local_fragment _ _LINEAR_TO_SRGB_CONVERSION
+        #pragma multi_compile _ _USE_FAST_SRGB_LINEAR_CONVERSION
```
phi-lira, line 12: UberPost has already many variants, a new multi-compile here will introduce a noticeable increase in build times. Can we avoid it?

com.unity.render-pipelines.universal/Shaders/PostProcessing/Bloom.shader
```diff
@@ -3,6 +3,7 @@ Shader "Hidden/Universal Render Pipeline/Bloom"
     HLSLINCLUDE
         #pragma exclude_renderers gles
         #pragma multi_compile_local _ _USE_RGBM
+        #pragma multi_compile _ _USE_FAST_SRGB_LINEAR_CONVERSION
```
phi-lira, line 6: if this variant is only used in fragment you can add `_fragment` to multi_compile

com.unity.render-pipelines.universal/Shaders/PostProcessing/BokehDepthOfField.shader
```diff
@@ -2,6 +2,7 @@ Shader "Hidden/Universal Render Pipeline/BokehDepthOfField"
 {
     HLSLINCLUDE
         #pragma exclude_renderers gles
+        #pragma multi_compile _ _USE_FAST_SRGB_LINEAR_CONVERSION
```
phi-lira, line 5: if this variant is only used in fragment you can add _fragment to multi_compile

phi-lira (changes requested):

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -58,6 +58,7 @@ internal class Styles
             // Post-processing
             public static GUIContent colorGradingMode = EditorGUIUtility.TrTextContent("Grading Mode", "Defines how color grading will be applied. Operators will react differently depending on the mode.");
             public static GUIContent colorGradingLutSize = EditorGUIUtility.TrTextContent("LUT size", "Sets the size of the internal and external color grading lookup textures (LUTs).");
+            public static GUIContent useFastSRGBLinearConversion = EditorGUIUtility.TrTextContent("Use Fast SRGB/Linear conversions", "Use faster, but less accurate approximation functions when converting to SRGB from Linear color space and vice versa.");
```
oleks-k, line 61: Concerning the name of the property ("Use Fast SRGB/Linear conversions"): 
I'm not sure what's the current capitalization rule for UI elements. Please ask @TheoWong-pixel. It should either be "Use fast sRGB/Linear conversions" or "Use Fast sRGB/Linear Conversions".
sRGB is spelled with the lowercase "s".

The tooltip:
"Use faster, but less accurate approximation functions when converting between the sRGB and Linear color spaces."

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
skerys: Made the property not changeable during runtime and added a shader preprocessor rule to avoid an increase in build times.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
```diff
@@ -9,6 +9,7 @@ Shader "Hidden/Universal Render Pipeline/UberPost"
         #pragma multi_compile_local_fragment _ _FILM_GRAIN
         #pragma multi_compile_local_fragment _ _DITHERING
         #pragma multi_compile_local_fragment _ _LINEAR_TO_SRGB_CONVERSION
+        #pragma multi_compile_fragment _ _USE_FAST_SRGB_LINEAR_CONVERSION
```
phi-lira, line 12: minor: can the above be local keyword?

phi-lira (approved):

PR #2108, opened 2020-10-06
[Shader Graph] Fix for visual issue where node input/output areas were not hidden on initialization

marctem:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1262050/, actually happened with any node/block that only has ports on one side, such as input nodes. Would be wrong whenever graph reloaded (first open, undo/redo, new node creation), would fix itself when nodes redrew - attach/disconnect edges mostly.

### Testing status
Verified open, undo, redo, creation all look correct now for block nodes and regular nodes.


ghost (approved):
Fixes the bug.

cdxntchou (approved):

alindmanUnity (approved):

PR #2110, opened 2020-10-07
Fog volumetric scattering for path tracing

eturquin:
Added fog volumetric scattering support for path tracing.

* Before: toggling volumetric in the fog settings would not do anything in path tracing, only absorption was handled.
* After: toggling volumetric adds isotropic single scattering from fog, for all lights supported by the path tracer (all point lights, rect area lights, directional lights). Like raster mode, it also change the absorption color to black, which should solve bug 1261158 (https://fogbugz.unity3d.com/f/cases/1261158/).

Note: with a minor change to the sampler used in RT and PT, test references may have to be updated.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -20,6 +20,7 @@ The version number for this package has increased due to a version update of a r
 - Added the status check of default camera frame settings in the DXR wizard.
 - Added frame setting for Virtual Texturing. 
 - Added a fade distance for light influencing volumetric lighting.
+- Adding an "Include For Ray Tracing" toggle on lights to allow the user to exclude them when ray tracing is enabled in the frame settings of a camera.
 - Added fog volumetric scattering support for path tracing.
 
 ### Fixed
@@ -44,6 +45,7 @@ The version number for this package has increased due to a version update of a r
 - Fixed issue where Default Volume Profile Asset change in project settings was not added to the undo stack (case 1285268).
 - Fixed undo after enabling compositor.
 - Fixed the ray tracing shadow UI being displayed while it shouldn't (case 1286391).
+- Fixed issues with physically-based DoF, improved speed and robustness 
 
 ### Changed
 - Combined occlusion meshes into one to reduce draw calls and state changes with XR single-pass.
--- a/TestProjects/HDRP_DXR_Tests/Assets/Scenes/601_LightCluster.unity
+++ b/TestProjects/HDRP_DXR_Tests/Assets/Scenes/601_LightCluster.unity
@@ -123,6 +123,221 @@ NavMeshSettings:
     debug:
       m_Flags: 0
   m_NavMeshData: {fileID: 0}
+--- !u!1 &68665949
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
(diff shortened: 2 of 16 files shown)


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Changelog
+﻿hangelog
```
sebastienlagarde, line 1: typo

com.unity.render-pipelines.high-definition/Runtime/RenderPipelineResources/Foliage Diffusion Profile.asset
```diff
@@ -12,15 +12,13 @@ MonoBehaviour:
...
+  m_Version: 1
   profile:
-    name: Foliage
     scatteringDistance: {r: 0.7568628, g: 0.7019608, b: 0.24313727, a: 1}
     transmissionTint: {r: 1, g: 1, b: 1, a: 1}
     texturingMode: 0
     transmissionMode: 1
     thicknessRemap: {x: 0, y: 0.2873168}
-    worldScale: 1
+    worldScale: 0.01
```
sebastienlagarde, line 22: why this change? guess it is a mistake

eturquin: Yes, should not have been commited.

com.unity.render-pipelines.high-definition/CHANGELOG.md
eturquin: Yup.

eturquin:
All seems good now, I'm re-running all tests to generate new refs, to account for the minor alterations tied to the sampling sequence changes.

eturquin:
Ref images have been updated.

remi-chapelain (commented):
Implementation is great overall, noticed a few issue that 
It should also be mentionned somewhere what is not supported by this implementation, non exhaustive list : 
- Anisotropy not supported yet
- Shadow dimmer in Volumetric options in lights is ignored
- Density volume are not supported yet
- Multi scattering (volumetric indirect) is possible but not implemented yet because it can be very noisy and not implemented in raster so probably not expected

Few issues : 
**1. We get a lot of noise with directionnal light when max fog distance is high** 
image

**2. Albedo color seems to only affect what's in the range of the light in Path tracer** (only a spotlight here)
whereas in raster, albedo color affects the whole fog volume

195b298b5444ba1a0e28d95a472d14af
 

**3. Very hard to have a proper converged image with a high fog attenuation distance**
There's probably an issue with how the Fog Attenuation Distance is handled with the path tracer
- A low value darkens the sky 
- A high value render a proper sky but very noisy volumetric

22a7756e8a539b76181bbfd05b2e5022 (1)

b97a6ab307ad963d587fa9923c82fcbe

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingLight.hlsl
```diff
@@ -468,4 +487,225 @@ void EvaluateLights(LightList lightList,
     }
 }
 
+// Functions used by volumetric sampling
+
+bool SolvePoly2(float a, float b, float c, out float x1, out float x2)
+{
```
sebastienlagarde, line 493: We should move those function in a common file (math.hlsl?)

sebastienlagarde (approved):

sebastienlagarde:
@eturquin , all the point of Remi need to be mention in the documentation of pathtracer. 
I will merge this PR, but please add a description in what's new in 10, and in pathtrace doc during the doc week thanks

PR #2114, opened 2020-10-07
Shadowmask: Added a section that refers to the Unity Manual.

oleks-k:
Shadowmask: Added a section that refers to the Unity Manual, because Lighting Modes works the same as in Built-in.


lukaschod (approved):
Looks good. Good job!

phi-lira (approved):

PR #2118, opened 2020-10-07
[HDRP][Compositor] Fix errors when resizing the compositor's output 

pmavridis:
### Purpose of this PR
@TomasKiniulis found some regressions in the graphics compositor that were introduced by one of my recent PR's
- Resizing the compositor window print's errors in the console 
- Removing and then re-adding the compositor also results in error. 

Both issues are a regression from this PR: https://github.com/Unity-Technologies/Graphics/pull/2070 because the compositor is no longer cloning all cameras, so we must be a bit more careful with the management of the resources/RT's.

---
### Testing status
- Compositor test is green
- Resize the compositor output does not create problems anymore
- Removing / re-adding the compositor works as expected.


TomasKiniulis (approved):
Nice! Works as expected now. 

Tried enabling/disabling removing compositor scene while new hdrp asset was added during before disabling 
Tried resizing with current RP asset, new RP asset created and assigned

sebastienlagarde (approved):

PR #2119, opened 2020-10-07
[ShaderGraph] [bugfix 1283425] Handle invalid SubGraph assets during recursion check

cdxntchou:
### Purpose of this PR
Fix for regression https://fogbugz.unity3d.com/f/cases/1283425/

Handle null SubGraphAsset instead of throwing null access exception.

We assume any subgraphs we can't access do not contain recursions; erring on the side of letting the user save their files.

---
### Testing status

Tested on my local repro, not entirely sure how the repro occurred, but could repro it locally 100%.

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
+++ b/com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
@@ -139,7 +139,15 @@ namespace UnityEditor.ShaderGraph
             {
                 foreach (var sgNode in subGraphNodes)
                 {
-                    if ((sgNode.asset.assetGuid == overwriteGUID) || sgNode.asset.descendents.Contains(overwriteGUID))
+                    var asset = sgNode?.asset;
+                    if (asset == null)
+                    {
+                        // cannot read the asset; might be recursive but we can't tell... should we return "maybe"?
+                        // I think to be minimally intrusive to the user we can assume "No" in this case,
+                        // even though this may miss recursions in extraordinary cases.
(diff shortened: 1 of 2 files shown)


com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
```diff
@@ -139,7 +139,15 @@ internal static bool CheckForRecursiveDependencyOnPendingSave(string saveFilePat
             {
                 foreach (var sgNode in subGraphNodes)
                 {
-                    if ((sgNode.asset.assetGuid == overwriteGUID) || sgNode.asset.descendents.Contains(overwriteGUID))
+                    var asset = sgNode?.asset;
+                    if (asset == null)
```
marctem, line 143: From a maintenance perspective, at first glance I'd think this `if(asset == null) { // big comment } else { code }` block is a bug. I think putting the comment above `if(asset == null) return false;` is clearer in the intent.

marctem (changes requested):

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
cdxntchou: how about a continue; ?

marctem (approved):

ghost (approved):
It's a simple fix so I'm taking Chris's word for it that it's fixed.

cdxntchou:
Just waiting on Katana results...

cdxntchou:
Katana all green

PR #1413, opened 2020-07-24
Fixing color input to gamma correct

elizabeth-legros:
Needed a way to deprecate the Color Property and Color Node as painlessly as possible. Made system to handle deprecation/upgrading, as well as explicit override to allow previous version usage. 

Color property deprecation - exposed HDR color properties are passed through without modification to shader. Need to be corrected. Color picker assumes linear space, so assume HDR color properties are in linear space, and conditionally convert to gamma based on project.
Color node deprecation - Inline colors were always assumed to be in gamma space, and conditionally were converted to linear space. HDR colors are already in linear space, so fixed conditional space conversion. 

Modified input nodes test in test project to test (HDR and Default) inline color node, regular color property, and deprecated color property

PR #2120, opened 2020-10-07
[9.x.x][ShaderGraph] Backport of color deprecation and versioning system

elizabeth-legros:
Backport of https://github.com/Unity-Technologies/Graphics/pull/1413
Ran graphics tests in ShaderGraph test project locally


cdxntchou (approved):

ghost (commented):
Things tested:
1. Confirmed that Property -> Inline and back is the same in Universal for both Linear and Gamma space
2. Confirmed that Property -> Inline and back is the same in HDRP in Linear space
3. Confirmed ability to create deprecated node or property with proper option enabled in preferences
4. Confirmed node updating works for both inline and property with "show deprecated" on
5. Confirmed "hover for info" tooltip works
6. Confirmed node updating works for both inline and property with "show deprecated" off
7. Updated node from before PR and confirmed color property and inline become deprecated and can be upgraded

Repro steps for bug
1. Gamma space, universal RP, new lit shadergraph, show deprecated nodes on. Make sure you have a scene with HDR color enabled, post processing enabled, and a volume with bloom and intensity set to 1
2. Create a deprecated color node with HDR and intensity above 0, click it and upgrade it in the graph inspector
3. Plug it into emission
4. Save the shadergraph, put it on material on a sphere in your scene
5. Convert the color property node to inline
6. Look at the change in bloom level on object in scene

Alternative to step 5: instead of converting it to inline, duplicate it and plug the duplicated property into emission

This bug reproduces on Master which means it was a problem with the original PR. Will await further instruction on what to do about it.

ghost (approved):
Pretty sure the bug that's happening has nothing to do with this PR. In practice the result of the bug is an HDR color that has a specific color in the preview and in the shadergraph but is black in the material. I can't figure out how to consistently reproduce this bug but I did reproduce it once on master. I don't think it's caused by this PR so I'm going to approve.

marctem (approved):

PR #2122, opened 2020-10-08
Improve Light Unit Slider UX

johnpars:
### Purpose of this PR
This PR introduces various fixes, additions & improvements to the light unit UX as requested by artist and UX team, namely:

#### 1) Replaced Icons
Icons
Icons for Fluorescent, Shade, Blue Sky were added to the temperature slider.

#### 2) Preset Context Menu
ContextMenu
Clicking on the icon will present a list of presets to the user. Clicking on them will place the value in a recommended preset value in a value range, which is set by artist (in reference to [this table](https://docs.google.com/spreadsheets/d/1SfY3Sr1n24NP1FIYXpQ21uj-KI386H06atIyVb547yU/edit?usp=sharing)).

#### 3) Undo/Redo
Undo
Replicating the breaking case as pointed out by @iM0ve- it should now be fixed.

#### 4) Fixed alignment of 6500 Kelvin
Before:
6500Broken

After:
6500Fixed
As the temperature tooltip states, 6500 Kelvin must be equivalent to the white point. Prior to this PR, this was not reflected in the new slider UI, but now it will be true.

#### 5) Correction to spot reflector behavior
Before:
Reflector

After:
NewReflectorBehavior
Prior to this PR, spot light + reflector actually broke the nature of radiance units (like Candela), since they could be changed based on the shape of the spot light. After long discussion with artist + engineer, this is the acceptable way forward to correct the issue, with plans in the future for further improvement on the spot light case. Feel free to follow the conversation [here](https://unity.slack.com/archives/C01A9B2JFNK/p1600313335084000) and [here](https://unity.slack.com/archives/C01A9B2JFNK/p1602079290050800). 

#### 6) The spacing between icon and slider was removed, so that it is aligned with the rest of the inspector.

---
### Testing status

Tested these changes to the editor locally on two of my machines. 
- Undo/Redo
- Ensure matching light unit behavior in HDRP 8.2
- Have not tested making a build as this does not effect the runtime. 

---
### Comments to reviewers
- To fix 4 required modelling an exponential curve that fits an f(0.44) value to 6500; this must happen so that the value matches the white in the temperature gradient. It uses pre-computed coefficients I derived offline and these are used to fit the curve. There is more info in the code comment, but want to confirm that this is alright. From a user perspective, it yields the desired result with no issue (as far as I could test).

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/LightUnit/LightUnitSlider.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/LightUnit/LightUnitSlider.cs
@@ -146,9 +146,7 @@ namespace UnityEditor.Rendering.HighDefinition
         void DoIcon(Rect rect, GUIContent icon, SerializedProperty value, float range)
         {
             // Draw the context menu feedback before the icon
-            var menuRect = rect;
-            menuRect.width += 5;
-            GUI.Box(menuRect, GUIContent.none, SliderStyles.k_IconButton);
+            GUI.Box(rect, GUIContent.none, SliderStyles.k_IconButton);
 
             var oldColor = GUI.color;
             GUI.color = Color.clear;


sebastienlagarde (approved):

sebastienlagarde:
@iM0ve I will merge this one for 10.1 package as it fix a wrong behavior so it is important, but still provide your QA pass feedback here. THanks

iM0ve:
Testing seems sufficient, but I will check on the next 10.1 package revision

PR #2124, opened 2020-10-08
Disable shader debug symbols

alelievr:
### Purpose of this PR
Disable shader debug symbols + reduce build size

---
### Testing status
I only check that it compiles.


jenniferd-unity (approved):
I tested on a DX11 build. Reduced size of the CustomPassUtils.shader from 1.39MB to 11kB.

sebastienlagarde (approved):

sebastienlagarde:
yeah those stuff save a lot, good catch :)

PR #2125, opened 2020-10-08
[10.x.x] Filter list of events in RenderObjects pass to only show valid events.

phi-lira:
### Purpose of this PR
 Fix [case 1278763](https://issuetracker.unity3d.com/issues/urp-disabling-srp-batcher-breaks-shadow-rendering-when-custom-renderer-feature-is-used)

 Calling DrawRenderers before BeforeRenderingPrepasses will not work as at this point the pipeline always expect monoscopic rendering and the camera matrices are not setup.

Untitled

### Testing status
 Tested all events in UniversalRenderingExamples. There are two outstanding issues we need to follow up with a discussion:
 - When rendering after post-processing, the depth from post-processing is killed. Opened [case 1283623](https://fogbugz.unity3d.com/f/cases/1283623/) 
 - When rendering after everything, that still happens before final post. This seems wrong as this is essentially the same as AfterneRenderingPostProcessing. AfterEverything should render after color grading, post AA, dither and sRGB convertion. Opened [case 1283631](https://fogbugz.unity3d.com/f/cases/1283631/)

---
### Comments to reviewers
 Added code to filter all invalid events in the RenderObjects feature UI and also validation to make sure the runtime data isn't set to invalid events.

 We could call SetupCameraProperties at the beginning of frame and then restore camera matrices / constants after before rendering depth pre-pass. However, even so the RenderObjects pass should not be recommended to use before rendering prepasses imo and that change would require deep testing as it might affect XR and cause other unknown issues (things that might be handled implicit in SetupCameraProperties that we might forget to restore).

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/RendererFeatures/RenderObjectsPassFeatureEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/RendererFeatures/RenderObjectsPassFeatureEditor.cs
@@ -75,20 +75,20 @@ namespace UnityEditor.Experimental.Rendering.Universal
 
         private List<SerializedObject> m_properties = new List<SerializedObject>();
 
-        static Func<int, bool> filterRenderPassEvent = x =>
+        static bool FilterRenderPassEvent(int evt) =>
             // Return all events higher or equal than before rendering prepasses
-            x >= (int) RenderPassEvent.BeforeRenderingPrepasses &&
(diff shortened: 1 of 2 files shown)


pbbastian (changes requested):
Looks good, I have a few suggestions :)

pbbastian (approved):

PR #1616, opened 2020-08-17
Universal/point light shadows

Wilfrid-Unity:
### Purpose of this PR

This pull request adds support for real-time Point Light shadows in Universal RP, as well as some fixes for Universal RP real-time shadows.
It brings Unity's Universal RP closer to parity with the Built-In RP.

**Main changes**

Point Light shadows are implemented by extending pre-existing class _UnityEngine.Rendering.Universal.Internal.AdditionalLightsShadowCasterPass_, that was previously used only for Spot Lights shadows.

In Unity code the term "shadow slice" is used to refer to shadow map textures generated for a single shadow casting light.
While a Spot Lights only require 1 shadow slice, for Point Lights 6 shadow slices are needed (one for each face of a cube around the light). 
Therefore the main changes in this pull request are related to the indexing of those additional shadow slices.

**Other additions from this pull request**

- When applying Punctual Lights Shadow Normal Bias, the light direction used is taken separately for each different shadow caster vertex position. Previously, the same light direction was used for all vertices, which caused the shape of shadows to change when a point light was rotated.

- Punctual light shadow setting "Soft Shadow" or "Hard Shadow" can now be set independently for each light. Previously, assigning "Soft Shadows" to one punctual light in the scene forced all shadow-casting punctual lights in the scene to use the "Soft Shadows" setting.

- The number of shadow casters rendered when generating spot light shadow maps is reduced. Previously, culling settings caused excess casters to be rendered, with an impact on perfomance.

**Not addressed by this pull request**

- Per-light shadow resolution override is not implemented by this pull request. It will be added separately.

---
### Testing status

**Manual Tests**:
I manually verified that punctual light shadows work in several scenarios combining multiple types of lights and shadow settings.
I manually verified that point light shadows were correctly rendered when using the Deferred Renderer.
I manually verified that point light shadows were correctly rendered when using Structured Buffers instead of Uniform Buffers, to pass shadow information to the GPU during the Lighting pass.

**Automated Tests**: 
This pull request adds test "145_ManyAdditionalLightShadowTypes" to suite UniversalGraphicsTests:

This scene contains 4 punctual lights of various types and shadow settings. Its purpose is to detect regressions related to the indexing of punctual light shadows.
I manually checked that changing shadow type for lights in the scene, caused the test to detect a failure.

**Yamato**: 

Automated test results are available here (no regression introduced to Universal Graphics and VFX tests):
[...]

PR #2126, opened 2020-10-08
Universal/custom shadow resolution for additional lights

Wilfrid-Unity:
### Purpose of this PR
This pull request allows Universal RP additional (i.e punctual) lights to specify custom shadow resolutions. 

In current URP version, all additional (i.e punctual) shadowed lights use shadow maps of the same resolution (scaled to fit in the common shadow atlas).
Custom shadow resolutions make it possible to give more importance to some lights in the scene, ensuring their shadows will look as good as possible.

image

### U.I and assets
The interface provided by this pull request to control shadow resolution ; is similar to the one used in HDRP.

Different shadow resolution tiers can be defined in the Universal RP asset file.

image

Each light in the scene can specify one of the different resolution tiers, or a custom resolution.

image

### Assumptions and specs
- The default resolution tier values are defined [here](https://github.com/Unity-Technologies/Graphics/commit/6908ed3a306cac9b1e801db5d199675d320864b3#diff-cda577ab7c8553c11fda276e58e9d2a1R139) (currently 256/512/1024).
- The default custom resolution (for lights not using the pre-set tiers) is defined [here](https://github.com/Unity-Technologies/Graphics/commit/6908ed3a306cac9b1e801db5d199675d320864b3#diff-4aee3cc3d0396c8714e34b3a94b28781R33) (currently 128).
- All shadow map resolutions are rounded to power-of-two values.
- If the shadow atlas (common to all additional lights) is too small to contain all the shadow maps needed in a frame, they will be scaled down using a common ratio, until they all fit (logic implemented [here ("AtlasLayout" in AdditionalLightsShadowCasterPass.cs) ](https://github.com/Unity-Technologies/Graphics/commit/6908ed3a306cac9b1e801db5d199675d320864b3#diff-f1c7a626e87573f22063a54aebbca346R322)).

### Testing status
This pull request add test "146-AdditionalLightShadowCustomResolution" to the UniversalGraphicsTests suite, whose purpose is to check different shadow resolutions can be used.

image

### Comments to reviewers
I started this work on top of branch [universal/point-light-shadows](https://github.com/Unity-Technologies/Graphics/tree/universal/point-light-shadows).
Point Light Shadows branch has not yet been merged to trunk, pull request #1616 is still waiting for some approvals now.

To reduce the amount of changes to review, I created this draft pull request that currently only contains the commits related to custom shadow resolutions. The plan is to target it to trunk later (after #1616 has been merged).
I would appreciate reviews around the [changes in UniversalRenderPipelineAsset.cs](https://github.com/Unity-Technologies/Graphics/commit/6908ed3a306cac9b1e801db5d199675d320864b3#diff-cda577ab7c8553c11fda276e58e9d2a1R139) and the [code that updates UniversalAdditionalLightData](https://github.com/Unity-Technologies/Graphics/commit/6908ed3a306cac9b1e801db5d199675d320864b3#diff-7d3d58606b094ff94105f1b5837d8244R338).

### DIFF
--- a/TestProjects/UniversalGraphicsTest/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/146-AdditionalLightShadowCustomResolution.png
+++ b/TestProjects/UniversalGraphicsTest/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/146-AdditionalLightShadowCustomResolution.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:18934f843e99d05f6a7e9c53a399764e3b4ebf88c39d3490660cb451bb00a87e
-size 89177
+oid sha256:ede8664bf70f457df19f659a83b05272dd643caadb9870d1bdddb1b476161fbd
+size 84889


Verasl:
> All shadow map resolutions are rounded to power-of-two values.

This to me seems like we should just provide a drop down with preset values, so an enum rather than a int field that automatically gets rounded.

image
When there is not enough room for shadows(shown with warning 'Too many additional punctual lights..') there is some odd behaiviour with point light shadows, they seem to have faces getting fully shadowed.

Verasl (changes requested):
I would suggest changing the custom shadow resolution to a dropdown(eg, 128,256,512,1024,2048, 4096) since we are rounding it to nearest anyway, also we have a max additonal shadow atlas value which means there is no point letting users put in anything higher than 4k

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [11.0.x] - 202x-xx-xx
 ### Added
 - Added real-time Point Light Shadows.
+- Added support for custom additional (=punctual) light shadow resolutions.
```
martint-unity, line 10: Is this a typo? (=

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -563,6 +567,35 @@ public int additionalLightsShadowmapResolution
...
+
+        public int additionalLightsShadowResolutionTierMedium
+        {
+            get { return (int)m_AdditionalLightsShadowResolutionTierMedium; }
+        }
+
+        public int additionalLightsShadowResolutionTierHigh
+        {
+            get { return (int)m_AdditionalLightsShadowResolutionTierHigh; }
+        }
```
martint-unity, line 583: Since these are public they should probably be documented?

martint-unity (commented):
Left some feedback. I also agree with Andre. But that would diverge us from HDRP. So maybe take a small meeting with Theo and Andre and me?

Wilfrid-Unity:
> image
> When there is not enough room for shadows(shown with warning 'Too many additional punctual lights..') there is some odd behaiviour with point light shadows, they seem to have faces getting fully shadowed.

Thanks @Verasl for reporting this behavior.
I believe this is the same that the one in the repro project that @simon-engelbrecht-soerensen had attached to internal ticket https://fogbugz.unity3d.com/f/cases/1289619

I added code to prevent this from happening in revision c281bb0, it is included in the latest revision of the branch.

com.unity.render-pipelines.universal/CHANGELOG.md
Wilfrid-Unity: Thank for pointing out that it is not clear ; I meant that additional are punctual lights.
I updated the note to make it more clear in db8cef6.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Wilfrid-Unity: Thanks for the suggestion, I added that in 06f2e77.

Wilfrid-Unity:
> This to me seems like we should just provide a drop down with preset values, so an enum rather than a int field that automatically gets rounded.

> I would suggest changing the custom shadow resolution to a dropdown(eg, 128,256,512,1024,2048, 4096) since we are rounding it to nearest anyway, also we have a max additonal shadow atlas value which means there is no point letting users put in anything higher than 4k.

> I also agree with Andre. But that would diverge us from HDRP.

Thanks @Verasl and @martint-unity for this suggestion, that makes a lot of sense when considering Universal RP restrictions.


In this branch, I re-used the U.I implemented in HDRP in order to provide a unified experience between both Render Pipelines.



When opening this pull request, I consulted @TheoWong-pixel who had commented the section about shadow resolution UI in the [Tech Design Doc](https://docs.google.com/document/d/1uUoeB9B0HyR4ixI-GJcQD7XXG9bHEJV5bXvb9-h3WV8/edit?disco=AAAAJ84qgjc).
My understanding is that HDRP Punctual Lights Shadow Resolution Tiers U.I is currently not ideal: there is a plan to clean it up for Unity Next release ; and this will be the opportunity to harmonize it with Universal RP Additional Lights Shadow Resolution Tiers U.I.
Therefore Theo suggested going on with the current interface for now, but making sure that the behavior about rounding is explicited in the tooltips (which I did implement [here](https://github.com/Unity-Technologies/Graphics/blob/6908ed3a306cac9b1e801db5d199675d320864b3/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs#L50) and [here](https://github.com/Unity-Technologies/Graphics/blob/6908ed3a306cac9b1e801db5d199675d320864b3/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineLightEditor.cs#L36)).


Does this approach look OK to you?

com.unity.render-pipelines.universal/CHANGELOG.md
martint-unity: the = sign is it supposed to be there?

martint-unity: Saw the fix after posted. Thanks looks goo now. You can resolve this

martint-unity (approved):
lgtm

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -337,6 +353,44 @@ void DrawLightingSettings()
...
+        void DrawShadowResolutionTierSettings()
+        {
+            // UI code adapted from HDRP U.I logic implemented in com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/SerializedScalableSetting.cs )
+
+            var rect = GUILayoutUtility.GetRect(0, float.Epsilon, EditorGUIUtility.singleLineHeight, EditorGUIUtility.singleLineHeight);
+            var contentRect = EditorGUI.PrefixLabel(rect, Styles.additionalLightsShadowResolutionTiers);
+
+            EditorGUI.BeginChangeCheck();
+
+            const int k_ShadowResolutionTiersCount = 3;
```
lukaschod, line 365: I think this will trigger error in formatter, IIRC we only use prefix k_ const (My suggestion to change name)?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -565,6 +569,44 @@ public int additionalLightsShadowmapResolution
...
+            if (additionalLightsShadowResolutionTier <= UniversalAdditionalLightData.AdditionalLightsShadowResolutionTierLow /* 0 */)
+                return additionalLightsShadowResolutionTierLow;
+
+            if (additionalLightsShadowResolutionTier == UniversalAdditionalLightData.AdditionalLightsShadowResolutionTierMedium /* 1 */)
+                return additionalLightsShadowResolutionTierMedium;
+
+            if (additionalLightsShadowResolutionTier >= UniversalAdditionalLightData.AdditionalLightsShadowResolutionTierHigh /* 2 */)
+                return additionalLightsShadowResolutionTierHigh;
+
+            return additionalLightsShadowResolutionTierMedium;
```
lukaschod, line 607: Why not just:
return Mathf.Clamp(additionalLightsShadowResolutionTier, AdditionalLightsShadowResolutionTierLow , AdditionalLightsShadowResolutionTierHigh )

Or we expect values to change?

Verasl (approved):
LGTM
We need to make sure we track the needed UX work in the future

lukaschod:
Found small UI issue, when changing between no shadow and hard shadow - during transsition effect resolution property has wrong spacing.
ui-issue

lukaschod (approved):
Found small issues (Addded as comments), but nothing serious to stop from merging.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Wilfrid-Unity: Expected input to this function is 0, 1 or 2.
Output of the function, is the resolution defined in the RenderPipeline Asset. For example: 256, 512, 1024.
This is why I could not use Mathf.Clamp(input) here.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
Wilfrid-Unity: Thanks, it seems that the formatter job did not have problem with it:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fcustom-shadow-resolution/.yamato%252F_formatting.yml%2523formatting/4550981/job/(log:Execution)

I think I also found other occurrences of the same naming pattern in the project (UniversalRenderPipelineAsset.k_ShadowCascadeMinCount).

simon-engelbrecht-soerensen (approved):
Great work! Approved.

PR #2127, opened 2020-10-08
Fix wrong error thrown on DXR wizard fixup

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1263049/ 

Turns out SystemInfo.supportsRayTracing returns false for every device but D3D12, however the error message is confusing implying that the issue is not on the device type but is an hardware/OS issue. 

Upon fixing all, the API is set to D3D11 before the restart, hence it throws an error even though there is no real issue. 

I modified the error and what gets triggered at what stage in a way that we don't mention wrong hw unless we are sure that D3D12 is already set.


sebastienlagarde (approved):

TomasKiniulis (approved):
Cool that works!

Did a check when Template is set to "Fix All" and Restarted on DX11 and when project is already set to DX12, works as expected

PR #2129, opened 2020-10-08
Universal/update changelog

phi-lira:
### Purpose of this PR
 Added missing changelog entries. Code for those landed in 10.0, so  added to that section to avoid confusion.


com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -123,6 +123,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Use SystemInfo to check if multiview is supported instead of being platform hardcoded
 - Default attachment setup behaviour for ScriptableRenderPasses that execute before rendering opaques is now set use current the active render target setup. This improves performance in some situations.
 - Combine XR occlusion meshes into one when using single-pass (multiview or instancing) to reduce draw calls and state changes.
+- Universal built-in shaders now use local material keywords. This reduced the amount of global keywords used by the pipeline resulting in more space for global user keywords.
```
oleks-k, line 126: Is this edition correct?
"URP pre-built shaders now use local Material keywords instead of global keywords. This increases the amount of available global user-defined Material keywords."

phi-lira: fixed

PR #2133, opened 2020-10-08
Decal selection pass

adrien-de-tocqueville:
### Purpose of this PR

Fix https://fogbugz.unity3d.com/f/cases/1217722/
The HDRP and SG decal materials had no scene selection pass so picking didn't work in the editor

Unlike what's stated in the bug report, hdrp decal picking didn't work for me either, so I fixed it for both materials.
Actually it appears they could only be selected when clicking on a backface, but now it works

---
### Testing status

Tested decal picking on the repro scene with HDRP and SG materials


sebastienlagarde:
Please sure to test with both decal projector and mesh decal.  + Need to check with Yamato as changing the order of passes is always tricky in decal.

sebastienlagarde:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fdecal-selection-pass/.yamato%252Fall-hdrp.yml%2523All_HDRP_2020.2/3756743/job

adrien-de-tocqueville:
Decal projector still don't work but we agreed it's not necessary. Yamato seems ok with the changes
I am not sure how to test the stripper but looking at the code it looks like it should work ^^

sebastienlagarde (commented):
I need to think more about this, but I have hard time to figure out how it works , as code is separate between projector and mesh (and we used shaderpass to identify it), but the selection pass can't make this distinction.

adrien-de-tocqueville:
Replaced by https://github.com/Unity-Technologies/Graphics/pull/2234

PR #2134, opened 2020-10-08
[ShaderGraph]Fix for required PositionNode upgrade

elizabeth-legros:
PositionNode should upgrade automatically since its behavior is incorrect in previous versions and cant be maintained. Have OnAfterDeserialize call ChangeVersion if the version is less than 1.
Opened and reimported graphs to verify the PositionNode was automatically updated


cdxntchou (approved):

ghost (approved):

PaulDemeulenaere:
N.B. : If this code is backported, we should not forgot this fix PR : https://github.com/Unity-Technologies/Graphics/pull/2177

PR #2136, opened 2020-10-08
Custom pass scene visibility

alelievr:
### Purpose of this PR
This PR is a bit related to this issue: https://fogbugz.unity3d.com/f/cases/1277644/ but doesn't fix the case as it's by design.

It fixes the scene visibility buttons in the hierarchy for the custom pass, that was previously ignored

---
### Testing status
Manually tested with a custom pass using a Draw Renderer to render a cube in the scene: (note that the cube is using a hidden layer in the editor, otherwise we'd just see the original cube)
SceneVis
 
Also tested that only the scene view is affected by the visibility changes.


FrancescoC-unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
Merged for 10.1 package.

TomasKiniulis:
Testing status looks sufficient. No additional QA needed

PR #2137, opened 2020-10-08
Add Documentation Button to Volume Component Toolbar

johnpars:
### Purpose of this PR
This PR adds a help button to the volume component inspector that links a user to the relevant documentation. It was requested by artist [here](https://unity.slack.com/archives/C6Y79CZM0/p1601858275117000).

For programmers, this basically means that you can now use the `HelpURL` attribute to create the button for a volume component.

image

Usage example:
BlueBookHero

---
### Testing status
- Tested locally on my two machines
- Have not yet verified that this **should not** at all impact anything related to URP's volumes.

---
### Comments to reviewers
 - Adding this required an extension to the Volume Component Editor framework, which means that the contents of this PR are also relevant to stakeholder for Universal. The PR adds a new Volume Component attribute that users can decorate a component with, or alternatively provides an overridable documentation setting in the editor. 

- It is important to note that this new support for documentation is opt-in, unless you specify a documentation in either the override or the decorator, there should be absolutely no impact to the UI. This PR implements the new behavior to all the relevant volume [documentation](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.0/manual/) that I could find for HDRP 10.0.

- 10.1 documentation has not been published yet, so all of the buttons will link to "missing" web-page.


JordanL8:
This is great! For consolidating the raytraced/rasterized docs, I think we should keep the raytraced effects docs but just move the property reference out of them into the rasterized version. That way we can link to the rasterized page for the entire property reference and still keep the raytraced pages for specific details about the raytraced effect.

JordanL8 (approved):

com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
```diff
@@ -30,6 +30,25 @@ public VolumeComponentMenu(string menu)
         }
     }
 
+    /// <summary>
+    /// This attribute allows you to add a button to the editor that links to the documentation.
+    /// Useful for <see cref="VolumeComponent"/> that do not implement a custom <see cref="VolumeComponentEditor"/>.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public sealed class VolumeComponentDocumentation : Attribute
```
RSlysz, line 38: Is the [HelpURL] not usable on VolumeComponent ?
I think it is usable on non GameObject. (I don't remember having to modify it for the refactor and I use it on extension).
Though you need to grab it yourself in this case, but as you do here with [VolumeComponentDocumentation].

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -120,6 +120,11 @@ internal set
             }
         }
 
+        /// <summary>
+        /// Override this property to show a link to documentation on the editor toolbar.
+        /// </summary>
+        public virtual string documentationURL => null;
```
RSlysz, line 126: We only need this for Editor. 
Plus, do we need it outside? I think Protected is enough for this.
Furthermore, here we store one occurence of the documentation link per volume component object. We really do not need to charge the memory with this as the url should be static per volume component.

RSlysz (changes requested):
I think we can do better on the memory and API side.

- [HelpURL] should be enough, no need to add a new container for this. 
- the documentationURL field could store the url grabbed in the attribute. But in this case, it only make sens if the stored url is static to remove the burden of grabbing the attribute at each instance. Plus this should be an editor only API.
- we must avoid to search attribute at each OnGUI for performences.

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
johnpars: Ah yes, so the reason this was done initially was to support the case of multiple documentations in the case of Raytracing enabled or disabled. For example, AO currently has a different set URL for raytraced / rasterized modes (and this is the case for any volume component with Raytracing support). 

However after discussion with Seb, we will instead fix this on the documentation side and consolidate everything together.

So it should be safe to remove this override completely, we can stick with the attribute (since there should only ever be one URL per component).

com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
johnpars: Good point, swapped in `HelpURL` and removed the redundant `VolumeComponentDocumentation`.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -448,8 +448,8 @@ public static bool DrawSubHeaderFoldout(GUIContent title, bool state, bool isBox
         /// <param name="hasMoreOptions">Delegate saying if we have MoreOptions</param>
         /// <param name="toggleMoreOptions">Callback called when the MoreOptions is toggled</param>
         /// <returns>return the state of the foldout header</returns>
-        public static bool DrawHeaderToggle(string title, SerializedProperty group, SerializedProperty activeField, Action<Vector2> contextAction = null, Func<bool> hasMoreOptions = null, Action toggleMoreOptions = null)
-            => DrawHeaderToggle(EditorGUIUtility.TrTextContent(title), group, activeField, contextAction, hasMoreOptions, toggleMoreOptions);
+        public static bool DrawHeaderToggle(string title, SerializedProperty group, SerializedProperty activeField, string documentationURL, Action<Vector2> contextAction = null, Func<bool> hasMoreOptions = null, Action toggleMoreOptions = null)
+            => DrawHeaderToggle(EditorGUIUtility.TrTextContent(title), group, activeField, documentationURL, contextAction, hasMoreOptions, toggleMoreOptions);
```
martint-unity, line 452: You are changing the signature of this method and will break everyone that is using this. 
Just load URP and it breaks.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -459,7 +459,7 @@ public static bool DrawHeaderToggle(string title, SerializedProperty group, Seri
         /// <param name="hasMoreOptions">Delegate saying if we have MoreOptions</param>
         /// <param name="toggleMoreOptions">Callback called when the MoreOptions is toggled</param>
         /// <returns>return the state of the foldout header</returns>
-        public static bool DrawHeaderToggle(GUIContent title, SerializedProperty group, SerializedProperty activeField, Action<Vector2> contextAction = null, Func<bool> hasMoreOptions = null, Action toggleMoreOptions = null)
+        public static bool DrawHeaderToggle(GUIContent title, SerializedProperty group, SerializedProperty activeField, string documentationURL, Action<Vector2> contextAction = null, Func<bool> hasMoreOptions = null, Action toggleMoreOptions = null)
```
martint-unity, line 462: You are changing the signature of this method and will break everyone that is using this. 
Just load URP and it breaks.

martint-unity (changes requested):
Need to not break other packages and users that are using this method.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
johnpars: Oof, not sure why I wedged that right into the middle of the signature- fixed now.

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
johnpars: So actually, brought back the `documentationURL` to pre-process the attribute like you mentioned (saving the overhead on OnGUI).

sebastienlagarde:
Feedback have been taken into account, we will merge this one - the doc work will be done next week in a separate PR

PR #2141, opened 2020-10-09
Fix screen corruption on Xbox when in rendergraph mode

FrancescoC-unity:
This was an hard one to track :-) 

Very long story very short, the issue was due to the fact that a memory stomp happened on the TAA history by another unidentified target.
The reason why this happened with Motion Blur on is that the creation of a transient texture with the same color format of the scene color/history, the texture pool would end up in a state in which reuse was incorrectly triggered even though it was **not** the transient texture doing the stomping itself. 

Adding the fastmem condition to the hash used to determine reuse seems to fix the issue.


sebastienlagarde (approved):

TomasKiniulis (approved):
Build Xbox player on template, went through RenderGraph Clear Render Targets, Disable Pass Culling, Immediate mode debug modes. No rendering issues occur anymore.

PR #2142, opened 2020-10-09
Fix SSGI when enabled in rendergraph

FrancescoC-unity:
SSGI was calling m_SharedRTManager in the prepare parameters function which is also used in RG mode. However m_SharedRTManager  is not valid when in rendergraph. 

This is fixed by passing the depth mip info from outside so that we can pass it properly in both cases.


sebastienlagarde (approved):

sebastienlagarde:
Removing QA, merging in 10.1, it will be check in the package.

PR #2143, opened 2020-10-09
Fix full screen debug with double sided materials

adrien-de-tocqueville:
### Purpose of this PR

fix the compilation error by swapping the order of system generated inputs un fragment shader
NOTE: a define for xbox will need to be put in nda package

### Testing status

tested on template


sebastienlagarde (approved):

sebastienlagarde:
I remove QA, this will be tested when testing template on platform for 10.1

PR #2174, opened 2020-10-12
Fix volumetric fog nan

alelievr:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1283322/

Note that this fix was extracted from PR #1806: https://github.com/Unity-Technologies/Graphics/pull/1806/files?file-filters%5B%5D=.compute&file-filters%5B%5D=.cs#diff-1f847d87918c2da5530a9e6cd81a82cfR88

To give some context, there was an issue with the volumetric fog filter code sampling outside of the VBuffer area for the current camera. This resulted in sampling garbage values from the rendering of a previous camera, because the buffer is an RTHandle shared between all the cameras. Note that it only happens if you have two cameras of different resolutions, which is often the case with probes.

---
### Testing status
I validated the issue in the Amalienborg palace scene, note that you don't need to update to repro the issue, you can just change the center reflection probe to realtime and have both scene and game view opened at the same time.

Before;
image

After:
image

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
+++ b/com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
@@ -106,7 +106,7 @@ namespace UnityEngine.Rendering
         {
             if (cmd != null)
 #if UNITY_USE_RECORDER
-                if (sampler != null)
+                if (sampler != null && sampler.isValid)
                     cmd.BeginSample(sampler);
                 else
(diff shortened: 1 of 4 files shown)


sebastienlagarde (approved):

remi-chapelain (approved):
Issue is clear and testing seems sufficient regarding what has changed. Approving !

anisunity:
Was pair tested with @alelievr i approve

anisunity (approved):

PR #2176, opened 2020-10-12
8xx Take latest ruamel script from master

kivistikliis:
---
### Purpose of this PR
Because master uses editor pinning but the release branches dont, the yamato script is different between master and the rest, which complicates backporting yamato PRs. With #2112 landing on master, we can keep the script matching between the branches.

This PR takes the latest yamato changes (such that ymls stay content-wise unchanged for 8xx) in order to have the scripts between branches in sync.

Changes coming with this PR:
- reruncount=2 flag
- smoke tests in their own project metafiles
- optional editor pinning
- linux split build
- editmode disabled for Metal
- project logs included in artifacts
---
### Testing status
ABV https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/8.x.x%252Fyamato%252Fupdate-ruamel/.yamato%252F_abv.yml%2523all_project_ci_fast-2020.1/3778361/job

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue with a render texture failing assertion when chosing an invalid format. [case 1271035](https://issuetracker.unity3d.com/issues/the-error-occurs-when-a-render-texture-which-has-a-certain-color-format-is-applied-to-the-cameras-output-target)
 - Fixed an issue that caused the unity_CameraToWorld matrix to have z flipped values. [case 1257518](https://issuetracker.unity3d.com/issues/parameter-unity-cameratoworld-dot-13-23-33-is-inverted-when-using-universal-rp-7-dot-4-1-and-newer)
 - Fixed an issue where transparent meshes were rendered opaque when using custom render passes [case 1262887](https://issuetracker.unity3d.com/issues/urp-transparent-meshes-are-rendered-as-opaques-when-using-lit-shader-with-custom-render-pass)
+- Fixed Missing camera cannot be removed after scene is saved by removing the Missing camera label. [case 1252255](https://issuetracker.unity3d.com/issues/universal-rp-missing-camera-cannot-be-removed-from-camera-stack-after-scene-is-saved)
+- Fixed MissingReferenceException when removing Missing camera from camera stack by removing Missing camera label. [case 1252263](https://issuetracker.unity3d.com/issues/universal-rp-missingreferenceexception-errors-when-removing-missing-camera-from-stack)
 
 ## [8.2.0] - 2020-07-08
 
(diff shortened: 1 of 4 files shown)


sophiaaar (approved):

theo-pnv:
Can you add the backport for this one too please @kivistikliis ? https://github.com/Unity-Technologies/Graphics/pull/1980
It's only for 8.x.x, no need for backporting on 7.x.x as the line doesn't exist there

.yamato/config/_editor.metafile
```diff
@@ -6,4 +8,47 @@ platforms:
...
   - name: Linux
   - name: iPhone
-agent: cds_ops_ubuntu_small
\ No newline at end of file
+editor_priming_agent: cds_ops_ubuntu_small
+
+### EDITOR PINNING
+editor_pin_agent: package_ci_ubuntu_small
+
+# remove the following once editor pinning is done. these values will be picked up from __shared
```
theo-pnv, line 16: I know we have these on master, but should we? (have these lines) ie shouldn't we remove these lines both here and on master?
Isn't the editor pinning finished?

kivistikliis:
good one, backported #1980 now!

.yamato/ruamel/jobs/abv/abv_all_project_ci_nightly.py
```diff
@@ -4,40 +4,41 @@
 
 class ABV_AllProjectCiNightlyJob():
     
-    def __init__(self, editor, projects, test_platforms, nightly_config, target_branch):
-        if editor["version"] not in nightly_config["allowed_editors"]:
-            raise Exception(f'Tried to construct nightly with PR trigger for version {editor["version"]}')
-        self.job_id = abv_job_id_all_project_ci_nightly(editor["version"])
-        self.yml = self.get_job_definition(editor, projects, test_platforms, nightly_config.get("extra_dependencies",[]), target_branch).get_yml()
+    def __init__(self, editor, test_platforms, nightly_config, target_branch):
```
theo-pnv, line 7: I think `test_platforms` can be safely removed here, it was used in the `for` block in which we appended the smoke tests to the nigthly, but it's not used anymore.

.yamato/ruamel/build.py
```diff
@@ -87,8 +93,7 @@ def get_metafile(metafile_name, unfold_agents_root_keys=[], unfold_test_platform
     yml_dump_files(create_projectcontext_ymls(package_metafile))
 
     # create abv
-    print(f'Running: ABV')
```
theo-pnv, line 90: I know the print statement got removed on master (my mistake when moving the smoke tests to a specific project file), but I think it is actually intended to stay

theo-pnv (changes requested):

.yamato/config/_editor.metafile
kivistikliis: True, you have a good point. I left them there in case bigger editor pinning changes come up, so its quick to use these again, but indeed they don't add much value at this point

.yamato/ruamel/jobs/abv/abv_all_project_ci_nightly.py
kivistikliis: Good catch! Removing it

theo-pnv (approved):

PR #2182, opened 2020-10-12
[ShaderGraph] [bugfix 1283782] Fix for previews when selecting "None" as your texture property texture

cdxntchou:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1283782/

The core of the issue is that we cannot remove a Texture (or any property really) from a MaterialPropertyBlock, and we also cannot set the Texture to null (error).  So effectively we cannot go back to the unassigned Texture state that would allow us to automatically pick up the default texture values.

As a workaround, we now pass the defaults into the preview property for texture2Ds, so we can properly assign the texture in the MaterialPropertyBlock with the corresponding value.

Talking with Yao, it should be possible to actually make public RemoveProperty() functions on MaterialPropertyBlock, which would greatly simplify all of this, but that would be a feature in trunk, so not until 21.1 at the earliest.

---
### Testing status
Tested that all four default types show up properly in previews, and in the main preview, and match the color in scene.
image

Tested that assigning default types works with undo/redo/assembly reload, and that the previews update correctly.
Tested that assigning and unassigning (set to none) textures still works properly, with undo/redo/assembly reload and preview update.

Tested that non-2D texture types (2DArray, 3D, cubemap) still work as expected.

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -54,6 +54,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue where ShaderGraph would not prompt the user to save unsaved changes after an assembly reload
 - Fixed an issue with Position Node not automatically upgrading
 - Fixed an issue where failing SubGraphs would block saving graph files using them (recursion check would throw exceptions) [1283425]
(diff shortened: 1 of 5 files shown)


elizabeth-legros (approved):

ghost (approved):
Developer manual testing is sufficient.

cdxntchou:
Yamato green except VFX_HDRP Win DX11, which is also failing on master

PR #2185, opened 2020-10-12
Fix for Node previews on edge deletion

elizabeth-legros:
When a node is created already attached to an edge, the preview manager does not correctly gather default values for the connected edges. This fix makes the preview manager also regenerate properties on the input node of a deleted edge. 
fix for https://fogbugz.unity3d.com/f/cases/1284033/
Tried to repro bug and no longer could.


cdxntchou (approved):

LandonTownsendUnity:
We're only backporting to 9.x when required by an enterprise customer (according to Marc T)

PR #2186, opened 2020-10-12
[ShaderGraph] [bugfix 1206438] Detect and handle subgraph deletion while subgraph node is in use

cdxntchou:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1206438/

When editing a ShaderGraph file containign a SubGraph node, if the SubGraph asset is deleted, we now detect that and update the display of the SubGraph node to indicate the missing file error.

Putting the SubGraph back will also update the display to remove the error and return everything to normal.

Now the subgraph node preview will always be available, regardless if the SubGraph asset exists or not, it will simply be black if the asset does not exist.  This also fixes the issue where the preview expansion UI appears to freeze.

---
### Testing status
Tested deleting and putting back a subgraph file, correctly updates the display in both cases.
Tested undo/redo after doing the above.

Tested deleting and putting back a different version of the subgraph file to see that it properly adds new subgraph slots.

Tested doing so with preview expanded or collapsed.  State is correctly maintained, and is always interactive.

---
### Comments to reviewers


com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -157,7 +157,7 @@ public SubGraphAsset asset
 
         public override bool hasPreview
         {
-            get { return asset != null; }
+            get { return true; }
```
cdxntchou, line 160: preview displays black if there is no subgraph asset.   Forcing hasPreview to false leaves the view in a bad state (and it is weird in the UI if it auto-collapses the preview when the subgraph is missing).

com.unity.shadergraph/Editor/Importers/ShaderGraphAssetPostProcessor.cs
```diff
@@ -88,7 +88,7 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
             }
 
             // moved or imported subgraphs or HLSL files should notify open shadergraphs that they need to handle them
-            var changedFiles = movedAssets.Union(importedAssets)
+            var changedFiles = movedAssets.Concat(importedAssets).Concat(deletedAssets)
```
cdxntchou, line 91: Union is the same thing as Concat.Distinct, and we already have a Distinct below.

marctem (approved):
Looks Good.

ghost (approved):
Things tested:
Renaming subgraphs
Moving the subgraph into another folder
Deleting the subgraph's meta file (doesn't lose connection)
Deleting the subgraph's meta file and then renaming it (loses connection)
Manually inserting a Json syntax error into the shadergraph file (causes compile error but nothing worse)
Updating the number of slots in the subgraph and making sure the graph containing it still updates.
On top of the things the dev tested, this seems like enough.

cdxntchou:
Yamato all green

PR #2187, opened 2020-10-12
[ShaderGraph] [bugfix 1284029] TexelSize is now SRP batched properly

cdxntchou:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1235070/

Makes Texture2D shader properties properly declare their TexelSize constants in the right cbuffer for SRP batching.
To do so, have to switch the Texture2D shader property to using the compound/multi property pathways for declaring sub-properties.

---
### Testing status
Tested with a shader using TexelSize, verified was not SRP batchable before this PR, and is now SRP batchable.
Before:
image

After:
image

Can also see in a RenderDoc capture that the drawcalls are now issued as DrawIndexedInstanced calls.

Verified TexelSize is correctly corresponding to the texture pixel size:
image

image

Verified that the shadergraph importer version bump will cause all shaders to reimport (fixing their SRP batching behavior)

Tested with blackboard Texture2D properties, and with an inline Texture2dAsset node, in SubGraph and in ShaderGraphs.

---
### Comments to reviewers


elizabeth-legros (approved):

cdxntchou:
Yamato green except for VFX HDRP WinDX11 (also failing in matching trunk)

ghost (approved):

PR #2191, opened 2020-10-13
Fix flickering in static environment lighting (SkyManager) when scene and game view are side-by-side with a camera with 0 culling mask. 

pmavridis:
### Purpose of this PR
-  Fixed flickering in static environment lighting (SkyManager) when having scene and game view side-by-side with a camera with 0 culling mask and PBR sky (and static lighting is not None). 
- Fixed a null reference in compositor undo/redo callback. 

**Repro steps outside of the Compositor (for the flickering)**
The flickering issue was first found in Compositor by @TomasKiniulis, but it is a more general issue with the SkyManager in HDRP. Repro steps:
- Set the scene and game view side-by-side
- Set the Sky environment to PBR sky
- Set the static lighting of the scene to the same volume profile as the PBR sky in step 2.
- Set the culling mask of the main camera (the one that renders the gameview) to Nothing
(Note: when creating a new outdoors HDRP scene you automatically get steps 1,2 and 3 out-of-the-box)

Outcome: the game view flickers when moving the mouse (on Editor repaint)

**What was wrong?**
In the setup above, the scene view and the game view have effectively a different (static) lighting setup. In the scene manager we were only keeping one static lighting **update** context around and this was causing cache collisions in every frame in the cached sky **rendering** contexts (m_CachedSkyContexts) because both game view and scene view were writing on the same slot. With the delayed async read-back of the convolved probe, this was causing the static probe data to be out-of-sync as we flip-flop between cameras.

**How it is solved?**
When the ref count of a sky context is zero, we still keep it around for a frame, in case it is "reclaimed" by another view

---
### Testing status
- The flickering does not happen anymore with the repro steps above.
- Tests are green locally (tested DX11)


pmavridis:
As a side-note, the first fix that I have tried and was working was this:
image
to request a static lighting update after every cache collision.  

This was working fine (and it a single line !!!) but of course it was forcing a wait on async readback on every frame that static lighting was updated (a user drags a slider) or when we have the views side-by-side. So the real solution is to avoid the cache collision in the first place.

JulienIgnace-Unity (changes requested):
Unfortunately this fix is not what we want to do :/
By design the static lighting should never depend on the camera because we don't want it to depend on anything "current" for baking otherwise users would get inconsistent results all the time. So having multiple static lighting depending on the camera is not the right solution.
The problem here is a bit subtle though since indeed, the PBR sky depends on the current directional light... which can in practice change per camera based on culling result as you've seen.
This is actually a bigger problem than just culling as it could also depend on additive scene loading for example...
The only solution I see is to have an explicit sunlight for the static sky (which was discussed at some point) but I think there were issues because the lighting asset containing the parameters cannot reference objects from the scene (because it's an asset).
We need to look deeper into what can be done but this fix can't land as is.

pmavridis:
> We need to look deeper into what can be done but this fix can't land as is.

As you mention, the per-camera dependency of the static lighting is already there, it was introduced now. But yeah, I was also not very happy with this solution, and it is not clear right now how it could be fixed in a better way. So I will close this PR for now and re-open when we have a clear plan.

The compositor was already updated to not easily trigger this use-case (zero culling masks were used as the default for new layers) but still the problem might affect a few users, so we might want to solve this issue sooner rather than later.

pmavridis:
Re-opening as Draft with a different potential solution to the flickering issue: When the ref count of a sky context is zero, we still keep it around for a frame, in case it is "reclaimed" by another view. This is enough to fix the flickering in my tests, but still a work-around of other issues that Julien mentions.

TomasKiniulis (approved):
Not running into any flickering issues anymore. Tried repro with compositor and new creating new sceen with split screen, Main Camera Culling set to Nothing and switching RP assets.

PR #2198, opened 2020-10-13
Fix render queue migration to 10.x

alelievr:
### Purpose of this PR
Fixed a migration issue with the rendering queue in ShaderGraph when upgrading to 10.x;

Reference: https://unity.slack.com/archives/CC68SH5SB/p1602583588007500?thread_ts=1602287119.003300&cid=CC68SH5SB
Context:  The `RenderQueueType` enum doesn't match in 8.x and 10.x so we need an upgrade path to migrate correct values. Affected values are every render queue type above `PreRefraction`

8.x:
```
        public enum RenderQueueType
        {
            Background,
            // Opaque
            Opaque,
            AfterPostProcessOpaque,
            RaytracingOpaque,
            // Transparent
            PreRefraction,
            Transparent,
            LowTransparent,
            AfterPostprocessTransparent,
            RaytracingTransparent,
            Overlay,
            Unknown
        }
```
10.x:
```
        public enum RenderQueueType
        {
            Background,
            // Opaque
            Opaque,
            AfterPostProcessOpaque,
            // Transparent
            PreRefraction,
            Transparent,
            LowTransparent,
            AfterPostprocessTransparent,
            Overlay,
            Unknown
        }
```
---
### Testing status
I imported a ShaderGraph with Transparent After Post Process rendering pass created in HDRP 8.x to a project with HDRP 10.x, before this fix the SG editor was crashing when opening the surface options in the graph inspector. With this fix the issue is gone and the rendering pass value is correct (Transparent After Post Process)

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -159,6 +159,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed profiling scope for quality RTGI.
 - Fixed the denoising and multi-sample not being used for smooth multibounce RTReflections.
 - Fixed after post process rendering pass options not showing for unlit ShaderGraphs.
+- Fixed a migration issue with the rendering queue in ShaderGraph when upgrading to 10.x;
 
 ### Changed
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

sebastienlagarde:
no need for more QA, have been validated by external users of HDRP.

PR #2199, opened 2020-10-13
Fix fullscreen debug pass with XR DX12

adrien-de-tocqueville:
---
### Purpose of this PR

Fixed compilation errors on dx12 with XR mode.
Reorders the input varyings to pixel stage to put system generated semantics as last

---
### Testing status

Reenabled the two fullscreen debug pass tests
Added material with double sided in tests
Unit tests cover the following materials:
 - all hdrp 
 - all sg
 - double sided
 - POM displacement
 - transparents


sebastienlagarde (approved):

adrien-de-tocqueville:
I already launched the tests on the first commit, so it should be good (all green)

PR #2200, opened 2020-10-13
Universal/bugfix/double sided normal mode

Jonasmortensen:
### Purpose of this PR
Included URP lit shaders expose the ability to render the back faces of surfaces. This PR adds the ability to control how the normals of the back faces are calculated. Like HDRP, URP know have settings for double sided normal mode. It can be set to either flipped, mirrored or none which are all explained in the docs. The settings become visible when setting `Render Face` to `Both` or `Back`:
image

This before/after example image shows the result of the feature. The scene has a blue directional light pointing up under the quads and a red directional light pointing down from above:
image
---
### Testing status
I have been manually testing how the normals interact with the shadow bias settings and have observed some possible usability issues as it becomes more difficult to find suitable values for these settings. I decided to leave as is since this might be solved by moving the bias to the receiver side.

I've run the URP graphics tests on macOS Catalina.

I've added a test that renders all combinations of normal modes, shaders and normal maps similar to the example image given above. (148_DoubleSidedNormals)

---
### Comments to reviewers
To test the change in editor, create a material using either the Lit, Complex Lit, Simple Lit, Particles Lit or Particles Simple Lit shaders. Assign materials to models with visible back-faces and experiment with the normal modes. The feature also impacts the look of normal maps. 

The PR adds a graphics test using many materials which causes the files changed count to skyrocket. In reality the changes in the shader files, forward passes and gbuffer passes of mentioned shaders are the ones that require code review along with base shader GUI.

### DIFF
--- a/com.unity.render-pipelines.universal/Shaders/LitGBufferPass.hlsl
+++ b/com.unity.render-pipelines.universal/Shaders/LitGBufferPass.hlsl
@@ -69,6 +69,7 @@ void InitializeInputData(Varyings input, half3 normalTS, half3 doubleSidedConsta
         inputData.positionWS = input.positionWS;
     #endif
 
+    half3 viewDirWS = GetWorldSpaceNormalizeViewDir(input.positionWS);
     #if defined(_NORMALMAP) || defined(_DETAIL)
         float sgn = input.tangentWS.w;      // should be either +1 or -1
         float3 bitangent = sgn * cross(input.normalWS.xyz, input.tangentWS.xyz);
@@ -92,6 +93,7 @@ void InitializeInputData(Varyings input, half3 normalTS, half3 doubleSidedConsta
     #endif
 
     inputData.normalWS = NormalizeNormalPerPixel(inputData.normalWS);
+    inputData.viewDirectionWS = viewDirWS;
 
     #if defined(REQUIRES_VERTEX_SHADOW_COORD_INTERPOLATOR)
         inputData.shadowCoord = input.shadowCoord;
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/Documentation~/normal-modes.md
```diff
@@ -0,0 +1,8 @@
+# Normal Modes
+The lit shaders of URP all have configurable normal modes for back faces. This setting determines how the normals of the back face are calculated and can be set to one of three different modes:
+
+ Normal Mode                  | Description                                                  |
+| --------------------------- | ------------------------------------------------------------ |
+| Flip                        | This mode flips the normal direction on back faces. For surfaces seeming convex on the front face, this will create a concave impression on the back face and vice versa.
+| Mirror                      | This mode mirros the normal direction on back faces. This retains the perceived surface curvature and makes the front face indistinguishable from the back face.
```
erikabar, line 7: Typo in Mirror normal mode description: This mode mirros the normal direction on back faces. Should be: This mode **mirrors** the normal direction on back faces.

com.unity.render-pipelines.universal/Documentation~/normal-modes.md
```diff
@@ -0,0 +1,8 @@
...
+| None                        | This mode does not modify the the normal direction and keeps it the same as the front face.
```
erikabar, line 8: "the" is written twice, should be: This mode does not modify the normal direction and keeps it the same as the front face.

com.unity.render-pipelines.universal/Documentation~/normal-modes.md
Jonasmortensen: Updated

com.unity.render-pipelines.universal/Documentation~/normal-modes.md
Jonasmortensen: Updated

Jonasmortensen:
Closing for now because of the work being postponed

PR #2201, opened 2020-10-13
Fix a useless repaint & fix in editor

adrien-de-tocqueville:
### Purpose of this PR

Double perceived performance of HDRP in edit mode
https://docs.google.com/document/d/1xFaK-pFI1XHdngCkXg4HK16x7QJieGy32OMTJf7RpeI/edit

Most fixes will actually be done on this PR:
https://ono.unity3d.com/unity/unity/pull-request/114059/_/graphics/repaint-dirty-fixes

Another change needs to be made but it needs the API added in the PR, so we have to wait for this one

---
### Testing status

Fix a problem that will appear when the c++ PR is merged (currently there is so much repaint that it's not possible to see it)
No more repaints caused by the exposure texture changes (note that there are still lots of repaints anyway so this doesn't make a big difference, but it will)


sebastienlagarde (approved):

adrien-de-tocqueville:
follow up of #1888

PR #2202, opened 2020-10-13
Fix the exception where we can't use json in AfterDeserialize.

iTris666:
### Purpose of this PR

I have fixed the shader graph problem a bit too fast : FromJsonSerializeOverride, used in EnsureProperties can't be called in the deserialization thread, therefore it must be removed from AfterDeserialize and called lazily.

---
### Testing status
Tested that the exception on domain reload ( for instance when pressing play) is no longer thrown, but that the properties are fine when used by the vfx graph.

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Generation/ShaderGraphVfxAsset.cs
+++ b/com.unity.shadergraph/Editor/Generation/ShaderGraphVfxAsset.cs
@@ -120,6 +120,7 @@ namespace UnityEditor.ShaderGraph.Internal
         {
             get
             {
+                EnsureProperties();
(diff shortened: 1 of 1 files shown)


elizabeth-legros (approved):

VladNeykov (approved):
The inputs are disconnected in the VFX Samples project (bonfire, magic book), but it looks like a project-specific issue as they often broke in the past.

Otherwise, tried with new effects: creating, upgrading, duplicating, re-importing, moving between projects, editing in editor, editing in play mode, renaming names and reference names, adding/removing properties, HDRP and URP, with the VFX Window open or closed, and a standalone build: didn't find any issues.

PR #2207, opened 2020-10-14
Removing the asset database save call

martint-unity:
---
### Purpose of this PR

Removing the call to asset database. So we are not saving every time a property has changed.


phi-lira (approved):

ellioman (approved):

ellioman (approved):

ellioman:
Cancelled all Yamato jobs, so they appear to have failed here, because there is no test for this so Yamato won't detect anything. @erikabar will do a pass on this PR instead.

erikabar (approved):
Tested by tweaking all settings in URP asset and Forward Renderer, added new Renderer Features and tweaked all settings there, then saved a project, reopened it and checked if all settings were saved correctly.

PR #2211, opened 2020-10-14
[10.x.x] Clear of the render pass list happens at end of frame now

phi-lira:
### Purpose of this PR
 Previously the pipeline cleared the render pass list of a renderer just before start rendering. This was not ideal as users might wanted to inject render passes by C# scripting and the pipeline would clear those.
 This PR moves clearing of render pass list to end of frame so we don't clear the list of render passes users might have injected in that frame, allowing them to inject render passes in before camera rendering API.

---
### Testing status
 

---


phi-lira:
Close. Next iteration on: https://github.com/Unity-Technologies/Graphics/pull/2719

PR #2219, opened 2020-10-14
Added the support of eye shader for ray tracing.

anisunity:
This PR adds support of ray tracing effects for the eye shader. Now the additional sub-shaders required are properly generated.

Testings status
- Added an eye SG for the reflection tests (so eye works with RT reflections) and it looks similar to what is seen in the direct view
- Changed the Shadow test now that Eye SG is supported
- Ran the DXR tests locally they are green.


sebastienlagarde (approved):

sebastienlagarde:
@remi-chapelain I will expect that added test are enough.

remi-chapelain:
It's better than nothing, but I will try to add a more detailled test (like the 902_materials_variants) to catch regressions possible regressions better when I'll do a pass on it next week.

PR #2220, opened 2020-10-14
Fix Refraction

alelievr:
### Purpose of this PR

- Expose the refraction mode to the material inspector from SG. Note that the refraction mode is only displayed in the material when it's enabled in the shader graph:
image

- Added a warning when using a blend mode different than Alpha and refraction is enabled:
image

- Updated documentation accordingly:
image

- Also fixed this issue:
Issue: In the ShaderGraph settings, we can set the rendering pass value to pre-refraction **with** refraction enabled, which should not be possible.
This PR fixes this issue and hides the pre-refraction option in the rendering pass field when refraction is enabled in the shader.

- Note that this PR also include the code from https://github.com/Unity-Technologies/Graphics/pull/2198 that was reverted from staging

Behavior changes in the refraction material UI: now the pre-refraction is always visible in the rendering pass field and a warning is displayed if refraction is enabled with pre-refraction rendering pass instead of the refraction block being hidden.

image

---
### Testing status
Tested that pre-refraction is hidden when we enabled refraction in both the material and ShaderGraph.
Renderingpass

image

I also ran DXR + HDRP_Tests, all green


sebastienlagarde (approved):

JulienIgnace-Unity (approved):

TomasKiniulis (commented):
Checking with Antoine, there's an issue with Blending Mode still getting displayed in material(and material being affected by it) when material is created with Refraction Model set None. After shadergraph Refraction Model is changed to anything from "none", materials should hide Blending Mode option, but that does not happen

alelievr:
Fixed blend mode issue
BlnendMode

TomasKiniulis (approved):
Awesome! Not running into any issues anymore. Love it that there's a warning now in Shadergraph too, when blending mode is not alpha
image

Tested material creation before refraction is set and after refraction is set, changing refraction type ant reimporting the material keeps the changes unchanged

sebastienlagarde:
PR retargeted to master

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDSubShaderUtilities.cs
```diff
@@ -233,7 +233,8 @@ public static System.Collections.Generic.List<HDRenderQueue.RenderQueueType> Get
             }
             else
             {
-                result.Add(HDRenderQueue.RenderQueueType.PreRefraction);
+                if (!hasRefraction)
```
sebastienlagarde, line 236: thinking about that, what is happening if we are on prerefraction rendering pass and we select a refraction mode . In shader code it will work (refraction is just disabled), but in the UI. cc @TomasKiniulis

sebastienlagarde: I guess it is the same problem than for blend mode, we will need to always display the value but have a warning to say that it is not working. Thought?

alelievr: When I tested it was automatically setting the rendering pass to Default instead of Pre-Refraction

alelievr: I added the warning in both ShaderGraph UI and material inspector.

As a side effect, the refraction UI is now always visible even if we select pre-refraction rendering pass

image

com.unity.render-pipelines.high-definition/Documentation~/Lit-Shader.md
```diff
@@ -69,6 +69,10 @@ To create a new Lit Material, navigate to your Project's Asset window, right-cli
 
 Unity exposes this section if you select **Transparent** from the **Surface Type** drop-down. For information on the properties in this section, see the [Surface Type documentation](Surface-Type.md#TransparencyInputs).
 
+:warning: When you enabled the **Refraction**, make sure to use the **Blend Mode** **Alpha**, otherwise the effect will not work as expected. A similar warning will be displayed in the material inspector if you enabled refraction with a **Blend Mode** different than **Alpha**.
+
+You can also note that Refraction is not supported when using the rendering pass **Pre-Refraction**, a warning about it will show up in the material inspector and shader graph inspector in case your setup is incorrect.
```
JordanL8, line 74: Also, be aware that HDRP does not support Refraction when using the **Pre-Refraction** render pass. If you enable **Refraction** and use the **Pre-Refraction** render pass, a warning displays in the material and Shader Graph Inspector.

JordanL8 (changes requested):
Rewrote a couple of bits

com.unity.render-pipelines.high-definition/Documentation~/Lit-Shader.md
```diff
@@ -69,6 +69,10 @@ To create a new Lit Material, navigate to your Project's Asset window, right-cli
 
 Unity exposes this section if you select **Transparent** from the **Surface Type** drop-down. For information on the properties in this section, see the [Surface Type documentation](Surface-Type.md#TransparencyInputs).
 
+:warning: When you enabled the **Refraction**, make sure to use the **Blend Mode** **Alpha**, otherwise the effect will not work as expected. A similar warning will be displayed in the material inspector if you enabled refraction with a **Blend Mode** different than **Alpha**.
```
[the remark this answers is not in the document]
alelievr: > No emojis ❗ :rage1:

noooo! (btw, do you know why? we have them in the inspector, why not in the doc?)

PR #2226, opened 2020-10-15
[ShaderGraph] Fix property declaration bugs, simplify system

cdxntchou:
### Purpose of this PR
Fixes these bugs:
https://fogbugz.unity3d.com/f/cases/1256374/

Non-exposed VT properties will version themselves to Exposed.
The "Expose" checkbox is now disabled, so the user cannot re-expose them.
Note that VT properties do not allow the user to override their declaration, as they must be "UnityPerMaterial".
image

https://fogbugz.unity3d.com/f/cases/1285701/

HDRP target-injected properties are now flagged so they are removed if not needed by the Shader.
Verified by looking at the generated shader code, now those properties are no longer declared in HLSL:
image

https://fogbugz.unity3d.com/f/cases/1285700/

Existing properties will version themselves so their behavior is unchanged.  This means "non-exposed hybrid" properties get versioned to "override to Global" to maintain the old behavior.
image

Exposed hybrid properties get versioned to "override to HybidPerInstance"
image

Non-hybrid properties keep their existing behavior.  Exposed ==> UnityPerMaterial,    Non-exposed ==> Global.
Since this is the default, no override is necessary here.
image

And from now on, if the user sets their Property to "override HybridPerInstance", they can expose or unexpose the property without affecting the hybrid declaration.  I.e. this now does the right thing, declaring a hybrid property that doesn't get added to the Material:
image

https://fogbugz.unity3d.com/f/cases/1255951/

Matrices cannot be Exposed, which meant previously they were always treated Global, even if flagged as hybrid!

Old "hybrid" matrices are turned into "Override to Global" to match old behavior (they were never hybrid!)
User can change the override if they really DO want them to be hybrid.
image

Old non-hybrid matrices are also turned into "Override to Global" to match old behavior.

image

NEW matrix properties now default to "UnityPerMaterial" by default, to match the default state of the other float/vector/bool properties.
image

You can now flag Matrices to _actually_ be hybrid by using the override:
image

https://fogbugz.unity3d.com/f/cases/1235070/
Already fixed by https://github.com/Unity-Technologies/Graphics/pull/2187
But we change the code in this PR, so I re-validated that it still works correctly.

---
### Testing status

Created a shadergraph using every possible OLD property type with all possible combinations of Exposed and Hybrid flags.  Compared the generated shader code of the OLD system, with the same shadergraph versioned with this PR.  The only differences in the code were caused by intentional bug fixes to address the bugs above.  Detected differences:

1) hidden VT properties are now not hidden -- instead of global, they are declared in UnityPerMaterial.
     This is ok, because the hidden VT properties did not ever work, there should be no existing valid content.

2) HDRP target-injected variables removed.  Double checked with Seb that this is the correct list of variables to remove.  Ran full HDRP tests to verify.

3) Texture and Sampler declarations are separated to their own lines, instead of being a single line.

4) Virtual Texture fallback Texture declarations (used when VirtualTexturing is disabled) no longer declare their TexelSize properties, as they are unused by the VT sampling node.

That's it!

Also ran ShaderGraph and HDRP CI tests locally.
Ran HDRP and Universal Hybrid tests locally.
Ran VFX tests locally, verified matches exactly master results

Tested VFX Graph integration, checking that properties are connected.

---
### Comments to reviewers
For those outside ShaderGraph, just check the files relevant to your area (unless you're interested in the changes)


GraphineCharles (approved):
- Ok for SVT always being exposed and per material.

alelievr (approved):
LGTM! just had a small question about our stencil properties:
I see that they are not declared anymore in the shader, they only exist in the Properties section, and most of these properties are used like this:
```            Stencil
            {
                WriteMask [_StencilWriteMaskGBuffer]
                Ref [_StencilRefGBuffer]
                Comp Always
                Pass Replace
            }
```
Does this work in ShaderLab?

elizabeth-legros (approved):

cdxntchou:
>             {
>                 WriteMask [_StencilWriteMaskGBuffer]
>                 Ref [_StencilRefGBuffer]
>                 Comp Always
>                 Pass Replace
>             }
> Does this work in ShaderLab?

Yes, it's kind of an undocumented feature (at least I can't find the public docs on it anywhere, but you can see lots of references to using it in examples across the web).  Basically you can use Material properties to control the render and blend states in shaderlab.  HDRP is using them to allow the Material inspector to control those states, in order to allow switching between opaque and transparent rendering at the Material level..

FrancescoC-unity:
Ping now that this PR is open https://github.com/Unity-Technologies/Graphics/pull/2307 

Leaving a comment as if my PR gets merged before this, this PR will need changing, otherwise if this PR gets merged first I'll need a notification to modify my PR :-)

cdxntchou:
Yamato completed all green except for one job that was killed for being too old.

cdxntchou:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252Fproperty-declarations/.yamato%252F_abv.yml%2523all_project_ci_2020.2/3880578/job/pipeline

sharlenet (approved):

cdxntchou:
ABV completed on latest code:  https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252Fproperty-declarations/.yamato%252F_abv.yml%2523all_project_ci_2020.2/3912469/job/pipeline

Only failures are:

Universal on OSX_Metal_playmode_mono_apiNet2_Linear on version 2020.2  -- repros exactly on matching master
VFX_URP on Win_DX11_playmode_mono_apiNet2_Linear on version 2020.2 -- repros exactly on matching master
Test all packages -- found issue with gpuInstanced being removed from public API -- that should be fixed now.  will re-run these tests.

cdxntchou:
Verified the Package Test ABVs now only contains the issues fixed by #2312

PR #2231, opened 2020-10-15
Fix faulty area light tiling

FrancescoC-unity:
I think this is the fix, given that the center is shifted in front of z by half a range. 

Fix for  https://fogbugz.unity3d.com/f/cases/1283644/ 

I tested the repro case and moved around, I also verified we don't over allow area lights by looking at the tile debug view.


remi-chapelain (approved):
Auto approving since this looks like a trivial fix and @FrancescoC-unity is confident that fixes the issue !

sebastienlagarde (approved):

EvgeniiG (changes requested):
The existing code is correct. If there is indeed a problem, it is elsewhere.

sebastienlagarde:
See PR: https://github.com/Unity-Technologies/Graphics/pull/2281

EvgeniiG:
Fixed here: https://github.com/Unity-Technologies/Graphics/pull/2281

PR #2234, opened 2020-10-15
Fix backface and displacement picking for HDRP

adrien-de-tocqueville:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1282977/
and https://fogbugz.unity3d.com/f/cases/1217722/

Add a ScenePicking Pass to most shaders to have a custom picking
This allows to have picking depend on the material culling mode and to be able to correclty pick objects with vertex displacement.
This is done on HDRP Lit, LayeredLit, AxF, and Decal shaders and ShaderGraphs

[OUTDATED: This PR doesn't change anything for LitTessellation and LayeredLitTessellation so they are still 'broken']
Fix for https://fogbugz.unity3d.com/f/cases/1285370/
and https://fogbugz.unity3d.com/f/cases/1285367/
Tessellation picking now works correclty

The Unlit shader as well as URP shaders will be fixed by this PR:
https://ono.unity3d.com/unity/unity/pull-request/114181/_/editor/fix-picking-raster-state

Decal projectors cannot be picked because they are not taken into account by the picking pass.

To sum up, every object picking should be correct, with or without vertex displacement/tessellation

Materials for test scenes are upgraded in #2349

---
### Testing status
Tested picking of front and back faces of all aforementionned materials with alpha transparency and vertex displacement.
Tested the repro scene given in the second bug case

Lit Tessellation show selected outline on backfaces
2234_tess_culling

Lit Tessellation picking with vertex displacement
2234_tess_picking


TomasKiniulis (approved):
Checked all SG and HDRP shaders, backfaces are correctly ignored now(except for unlit since it's handled by another pr and tessellation).
RHGGilXWu4

For tessellation shaders I've filed a ticket here so it wouldn't get lost in the future: https://fogbugz.unity3d.com/f/cases/1285367/
Like Adrien mentions selection also ignores tessellation displacement and should be taken into account

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/PickingSpaceTransforms.hlsl
```diff
@@ -0,0 +1,108 @@
...
+#define CreateTangentToWorld CreateTangentToWorld_Picking
+#define TransformTangentToWorld TransformTangentToWorld_Picking
+#define TransformWorldToTangent TransformWorldToTangent_Picking
+#define TransformTangentToObject TransformTangentToObject_Picking
+#define TransformObjectToTangent TransformObjectToTangent_Picking
+
+
+float4x4 inverse(float4x4 m) {
+    float n11 = m[0][0], n12 = m[1][0], n13 = m[2][0], n14 = m[3][0];
+    float n21 = m[0][1], n22 = m[1][1], n23 = m[2][1], n24 = m[3][1];
```
adrien-de-tocqueville, line 64: Didn't find another solution to correctly compute the UNITY_MATRIX_I_M, since it is not set by the c++

sebastienlagarde: Lets move this function in common.hlsl (but we will do it in another PR

sebastienlagarde (approved):

PR #2235, opened 2020-10-15
Fixed default alpha input when using 11_11_10 color format in non-dev builds.

JulienIgnace-Unity:
### Purpose of this PR
Fixed an issue when rendering into texture where alpha would not default to 1.0 when using 11_11_10 color buffer in non-dev builds.

Fixes: https://fogbugz.unity3d.com/f/cases/1285290/

---
### Testing status
Tested with the repro project.


pmavridis (approved):

FrancescoC-unity (approved):

remi-chapelain (approved):
Solution seems to be straightforward, I'll trust the developer on this one

sebastienlagarde (approved):

PR #2238, opened 2020-10-15
Fix issues with reordering and hiding graphics compositor layers

pmavridis:
### Purpose of this PR
- Re-ordering composition layers now keeps the selected layer visible in the compositor window (QoL improvement)
- Fix error after re-ordering layers: https://fogbugz.unity3d.com/f/cases/1285282
- Fix editor freeze after re-ordering layers with filters: https://fogbugz.unity3d.com/f/cases/1283886
- Fix issue when disabling the first compositor layer: https://fogbugz.unity3d.com/f/cases/1283903

I could not repro the "null reference" in the last bug ticket, but there was still a bad UX behavior: when hiding the sub-layer the culling mask was set to Nothing, but when re-enabling it remained to Nothing, so the user lost his settings. I have fixed this by using  a clone camera.

I have grouped these fixes to one PR to reduce testing/merging overhead (and they are all related).

---
### Testing status
- Compositor test is green
- The errors reported in the three bug tickets do not repro anymore.
- Re-ordering layer "feels nicer" because the user selection is not reset anymore.

### DIFF
--- a/.yamato/_abv.yml
+++ b/.yamato/_abv.yml
@@ -95,8 +95,6 @@ all_project_ci_2020.2:
          rerun: on-new-revision
       -  path: .yamato/all-vfx_urp.yml#All_VFX_URP_2020.2
          rerun: on-new-revision
-    triggers:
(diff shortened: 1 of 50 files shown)


TomasKiniulis (approved):
Ran through the repros. Works as expected

sebastienlagarde (approved):

PR #493, opened 2020-05-15
Fix artifacts in volumetric lighting near opaque geometry

EvgeniiG:
# Purpose of this PR

Fix https://fogbugz.unity3d.com/f/cases/1231885/

# Testing status
## Manual Tests
Tested on the graphics tests scenes. More manual testing is needed.

## Automated Tests
Not yet.

## Links
**Yamato**: (Select your branch) https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics
More manual testing first.

# Comments to reviewers
So, there are two solutions.
One is to account for the geometry along the camera ray. If the geometry is within the voxel, I effectively make the voxel shorter.
Another is view shadow bias (disabled by default). I implemented view shadow bias by repurposing our normal shadow bias code. It does not appear to be particularly effective, so I disabled it for now. It can be re-enabled in code.

Here's an example where normal bias would help (although it's probably too expensive):

image

PR #2239, opened 2020-10-15
Fix atmospheric scattering precision

adrien-de-tocqueville:
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1283957
There was some precision issues when the fog opacity was near 1
The bug doesn't seem related to the density volume, unlike what's stated in the bug, but when the camera is very close to the ground

---
### Testing status

tested the repro steps of the bug report (it is related to the camera position on the y axis and not the density volume though)
and tried different extreme value to check there is no more banding


iM0ve (approved):
I could not reproduce the issue described in FB on 10.1
But I consulted with Antoine about the risk of code change and checked HDRP template fog, that the PR does not have any impact. I think its safe to merge

Before:
image

After:
image

adrien-de-tocqueville:
Note: the failed tests are due to PR #493, but it looks like the difference introduced then wasn't big enough to cause a failure. Now the small precision changes make it go over the threshold

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/OpaqueAtmosphericScattering.shader
```diff
@@ -118,7 +118,7 @@ Shader "Hidden/HDRP/OpaqueAtmosphericScattering"
         Pass
         {
             Cull Off    ZWrite Off
-            Blend One OneMinusSrcAlpha // Premultiplied alpha
+            Blend One SrcAlpha // Premultiplied alpha
```
EvgeniiG, line 121: Forgot to change comment

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/OpaqueAtmosphericScattering.shader
```diff
@@ -131,7 +131,7 @@ Shader "Hidden/HDRP/OpaqueAtmosphericScattering"
...
```
EvgeniiG, line 134: And here

PR #2252, opened 2020-10-16
[10.X.X] Fix Universal graphics test 130_UnityMatrixIVP on OpenGL ES 3

rolandk-unity:
### Purpose of this PR

This PR fixes and enables the URP graphics test 130_UnityMatrixIVP.

On OpenGL, values sampled from camera depth textures are in the range of [0, 1], while a clip space position's z value is in the range [-1, 1]. The test 130_UnityMatrixIVP did not take this into account, causing it to fail on OpenGL platforms. This issue is fixed by this PR.

### Testing status

Yamato tests completed on GLES3, Vulkan, Metal.

### Notes to the reviewers

Currently, a few Yamato tests are failing, which is in line with the SRP nightlies failing at the moment too. The changes in this PR are very local to just one shader used in just one test, so I am very confident that it is unrelated to any failures outside this test.


phi-lira (approved):

phi-lira:
Yamato failures unrelated to this PR.

PR #2253, opened 2020-10-16
[10.x.x] Hide warnings about mismatched particle normals

ellioman:
# Purpose of this PR
We always include vertex normals by default when building particle data, so users can switch between lit + unlit shaders without any extra config changes. This causes the standard particle gui to complain, but we should hide this error.

---
# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fparticles-ignore-mismatched-normals


richardkettlewell (approved):
Thanks, that was fast!

ellioman:
> Thanks, that was fast!

Well, you had already done all the work :)

phi-lira (approved):

phi-lira:
Closing, merged in : https://github.com/Unity-Technologies/Graphics/pull/2390

PR #2255, opened 2020-10-16
Baked reflection probes now correctly use static sky for ambient lighting.

JulienIgnace-Unity:
### Purpose of this PR
Baked reflection probes now correctly use static sky for ambient lighting.
Before the fix, when using dynamic ambient mode, reflection probes would use the current lighting sky for ambient. This was not consistent with other baked lighting (light probes and lightmaps).
Beside due to a frame delay in computing dynamic ambient, the first face of the probe would not have ambient lighting at all.
This is also fixed by using static lighting instead.
Fixes: https://fogbugz.unity3d.com/f/cases/1277334/

---
### Testing status
Tested with the repro scene and ran HDRP graphics tests.


iM0ve (approved):
I verified too using the steps. Also checked that baking with static still works and that baking without any sky as ambiance also works as intended.

0PsXTiHkG1

sebastienlagarde (approved):

PR #2256, opened 2020-10-16
Added a visualization tool for render graph

JulienIgnace-Unity:
### Purpose of this PR
Added a visualization tool for render graph. For now, this tool is aimed mostly at internal programmers to help them understand issues they might have when writing code for HDRP with Render Graph.
This viewer display several information.
- At the top we can the main tool bar to select which graph should be capture, the capture button itself, some visualization filters and the meaning of the various colors used in the tool.
- Then comes the main part: The list of passes in execution order can be seen horizontally and the list of used resources can be seen vertically (sorted by type, textures then compute buffers). In the space inbetween, we can see the "lifetime" of a resources. It starts at its creation (when a pass first writes to it) until its release (when a pass last reads it).
- When hovering on a pass, resources used in this pass will be highlighted. Red for a resource write, and green for a read.
- When hovering on a resource lifetime, the passes where it's used will be highlighted (same color as above).
- When a resource/pass is both read and write it will be shown as red.
- Filters at the top can be used to hide resource types, imported resources or even culled passes.

image

In order to achieve this, some more changes had to be made:
- Added export of debug data from the currently running graph.
- Made resource handle checking more strict

---
### Testing status
Tested it manually while developping it. Among other things:
- Tested all the filters.
- Tested capturing again the graph after changing frame settings to see the difference being reflected in the viewer
- Made sure that resource/pass read/writes are displayed correctly
- Made sure rsource lifetimes were correctly aligned with the passes on top.
- Resized the window in many ways to make sure the viewer would react properly.

---
### Comments to reviewers
@Unity-Technologies/gfx-qa-hdrp this tool is really aimed at internal programmers for now and is still experimental (as Render Graph is) so I'm not sure we should invest too much QA time in this. If someone wants to take a look at it, just send me a message and I will show how it works.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/RenderGraph/RenderGraphViewer.cs
+++ b/com.unity.render-pipelines.core/Editor/RenderGraph/RenderGraphViewer.cs
@@ -13,8 +13,8 @@ public class RenderGraphViewer : EditorWindow
         public static readonly GUIContent title = EditorGUIUtility.TrTextContent("Render Graph Viewer");
     }
 
-    public const float kRenderPassWidth = 20.0f;
-    public const float kResourceHeight = 15.0f;
+    const float kRenderPassWidth = 20.0f;
+    const float kResourceHeight = 15.0f;
 
     class CellElement : VisualElement
     {
--- a/com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraph.cs
+++ b/com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraph.cs
@@ -101,7 +101,7 @@ namespace UnityEngine.Experimental.Rendering.RenderGraphModule
     /// <param name="renderGraphContext">Global Render Graph context.</param>
     public delegate void RenderFunc<PassData>(PassData data, RenderGraphContext renderGraphContext) where PassData : class, new();
 
-    public class RenderGraphDebugData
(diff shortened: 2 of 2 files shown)


iM0ve (changes requested):
Hello, please add some visual information and perform at least a minimal amount of testing even for an internal tool

RSlysz (changes requested):
Well I reviewed befor all polution of yamato CI file change. I will not read it again...

the colors should be const (or static readonly) as read, write... will not change.

It will be simpler if you isolate the cells into a dedicated VisualElenent for both logic separation and potential upgrade (and it do not need to have all the things a button does, see snippet of code I sent you on slack). Same for column headers.

remi-chapelain:
Took a look while reviewing another render graph visualization tool PR
Aside from some [minor UI clipping issue](https://fogbugz.unity3d.com/f/cases/1289905/), it is fairly pleasant to look at and we understand quickly what it's about and the dependencies between the passes and the buffers/textures. Nice job !

remi-chapelain:
The only thing I would say is that it feels like something wrong (like your hover did not work) when you hover over a pass that uses no resources because nothing happens. 

Also, if clicking on the pass little square do nothing, is it possible to disable clicking (like on the resources lifetime) ?

PR #2257, opened 2020-10-16
Updated ray-traced shadows doc to include information about double sided materials

JordanL8:
### Purpose of this PR
Adds a note to tell users the Mesh Renderer's Two Sided option is not supported for ray-traced shadows and to use the Material's Double-Sided option instead.


remi-chapelain (approved):
Lgtm ✔️

PR #2258, opened 2020-10-16
Allow cubemap volume parameter to accept RenderTextures [2021.1]

alelievr:
### Purpose of this PR

This PR changes the behavior of the cubemap volume parameter, allowing to provide a render texture or custom render texture instead of a cubemap asset.

The field now has the same behavior as a texture field in the material inspector: it shows every texture asset in the project in the object picker
image
But only the textures of dimension cubemap are assigned to the field, while others leave the field to null.

Note that only two of our volumes uses the cubemap parameter for now: `HDRI Sky` and `Physically Based Sky`

image

Note that a side effect of this work is that a render texture can dynamically change its dimension without breaking the reference, this means that you can assign a cubemap CRT to the HDRI sky and then change the CRT to texture 2D leaving a texture 2D in a cubemap field. From what I tested it hasn't caused any further issue/errors.
Cubemap Sky

---
### Testing status
I tested with a custom render texture of cubemap format that the texture is accepted in all the cubemap fields, as well as Cubemap assets.

I also tested the upgrade: no asset seems to be lost due to the volume parameter type change.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fimprovement%252Fcubemap-volume-parameter/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/4292641/job


Chman (changes requested):
Looks good but it would be nice to output a warning in the console if the texture chosen is invalid, at least so that users know what's going on and why.

alelievr:
Would a warning at the picking level be enough? It won't catch the render texture dimension change though 😕

Chman:
Should be enough, as long as the user has a way to know why his texture didn't go in the slot at some point. No need to keep track of the texture dimensions IMO, they just need to know "why" the texture doesn't fit.

JulienIgnace-Unity (approved):

sebastienlagarde:
Sadly this change can't make it to 10.x :( as it break API

alelievr:
Yes, indeed

alelievr:
I added the error message:
image

Chman (approved):

TomasKiniulis (approved):
Thanks for the detailed description. Looks good!

- Checked HDRI Sky PBR Sky input fields on Game Object Volume
- Checked HDRI Sky PBR Sky input fields work as expected on HDRP Default Settings
- Made sure HDRI Sky updates RT texture change
oPEboSlJdq
- Made sure PBR Sky updates RT texture changes
PiLD2lmKoy
- checked that no issues occur when texture input fields is set to None, or Non-Cube texture(expected error is thrown to change type)
- changing RT type during runtime to non cube correctly throws an error to change back to cube type
- no issues occur when adjusting rt texture size

sebastienlagarde (approved):

PR #2262, opened 2020-10-16
[10.2][Shader Graph] Fix Small Bugs

alindmanUnity:
---
### Purpose of this PR

- Fixed an issue where the Main Preview window rendered too large on small monitors during first open. [1254392]
- Fixed an issue where Block nodes using Color slots would not be automatically removed from the Master Stack. [1259794]
- Fixed an issue where the Create Node menu would not close when pressing the Escape key. [1263667]

---
### Testing status

Cleared all cache and libraries and observed first time open behavior for shader graphs. I got this result on my machine: 
FirstTimeOpen_Fixed

Opened the create node menu and performed actions to confirm escape key now closes the window. 
EscapeMenu_Fixed

Added and removed targets to a Shader Graph to observe that all blocks now clear when unmodified. 
ColorBlock_Fixed

---
### Comments to reviewers
@LandonTownsendUnity I mostly am worried about the first time open Main Preview bug, I'm not sure if there are going to be weird machine differences.


ghost (approved):
First time open preview behavior seems to be as expected.
Also, tested removing targets when there are multiple targets to make sure that behavior is still good. Also tested removing all targets with something plugged into base color.

elizabeth-legros (approved):

alindmanUnity:
Small low risk SG only changes, tested and SG tests are green. No halo risk to either pipeline

PR #2264, opened 2020-10-16
Fixed a warning happening when putting the range of lights to 0 (case 1284682)

anisunity:
https://fogbugz.unity3d.com/f/cases/1284682/
When the range of a light is dragged to 0. Info messages are spammed in the console. The change clamps the value between 0.001 and +Inf.

Testing status.
Create a spot light dragged the value to 0 and written 0 in the texfield. Tested the fix on staging and 2020.2


martint-unity:
This might fix the issue for HDRP only. What about URP?
Fix should probably go into core where the drawing of this handle is happening.

sebastienlagarde (approved):

sebastienlagarde:
> This might fix the issue for HDRP only. What about URP?
> Fix should probably go into core where the drawing of this handle is happening.

hmm... Looks like Martin is correct that it isn't the right fixe. But at the same time there is no concept of Core for this.

So few thing:
1) check if it happen in builtin. Because given the bug, it must happen in builtin as well. 
If it is the case, then we need to do the min() on the C# side of light component

2) if it doesn't appear in builtin, there there is a very weird behavior somewhere as I will not expect that for range we do something different? In this case depends on what's happen then we may accept this current workaround and URP will need to do one as well. (There is no shared code for light in SRP).

anisunity:
the issue doesnt happen on the builtin pipeline on 2020.2b9
spot0

sebastienlagarde:
@martint-unity , so after check it appear that the issue don't appear in builtin. So it is not related to core component. As there is no core code handlight light it mean that each pipeline must do the check. HDRP will do this one, I will reopen the fogbugz and assign to URP for fix there.

PR #2267, opened 2020-10-16
Fix View Direction normalization

elizabeth-legros:
Used deprecation system to make new version of ViewDirection node that conditionally calls normalize in URP on the view direction.


alindmanUnity:
closing in favor of #2752

PR #2269, opened 2020-10-19
[11.x.x] Fixed issue with UNITY_REVERSED_Z and UNITY_UV_STARTS_AT_TOP and OGL

phi-lira:
### Purpose of this PR
 `UNITY_REVERSED_Z` and `UNITY_UV_STARTS_AT_TOP` are defined for platforms that don't support it.
 This has the side effect that when checking on shader `#if defined(UNITY_REVERSED_Z)` it will always be true.
 This behaviour is also different from built-in pipeline and confusing for users since we have doc pages with the usage of `#if defined(UNITY_REVERSED_Z)`. https://docs.unity3d.com/Manual/SL-PlatformDifferences.html

 This PR removes the definition of those macros for platforms that don't support it. This should not break existing shaders with the usage of `#if UNITY_REVERSED_Z` as that evaluates to false when the macro is not defined.

 Slack converstation: https://unity.slack.com/archives/C20E70CAC/p1602760055081400

---
### Testing status
 Tested by checking manually behaviour is the same when removing the macro.
 Any other failure should be catched by automation on GLES.

---


sebastienlagarde (approved):

phi-lira:
Removing backport label. We don't need to backport this one.

PR #2274, opened 2020-10-19
Fixed warnings when new input system is enabled

alelievr:
### Purpose of this PR
Fix warnings in the template when the new input system is enabled

```
warning CS0649: Field 'FreeCamera.fireAction' is never assigned to, and will always have its default value null
...
```
---
### Testing status
I opened the template project and checked that there are no compilation warnings.
I also check that it still compiles with the old input system and both at the same time.


JulienIgnace-Unity (approved):

PaulDemeulenaere (approved):
Yes, thanks, did the same fix at https://github.com/Unity-Technologies/Graphics/commit/aa73fcf5c50fa5ed5d237618ffd419712b55c99d
LGTM

PR #2275, opened 2020-10-19
Update default directional light intensity in HDRP

JulienIgnace-Unity:
### Purpose of this PR
This PR updates the directional light default intensity to 100k to better fit the PBR nature of HDRP.
In the documentation we state that 100k is the right value when using an HDRI sky and 130k for a PBR sky. We felt that 100k was a better default value because it fit the most common and simple case of using HDRI sky.
We also updated outdoor scene template directional lights to 130k instead of 110 as it is what is required by the PBR sky used in those scenes.

---
### Testing status
Created a new dir light and a new outdoor scene to check that values were correct.


iM0ve (approved):
Whats tested:
* Updated the new HDRP template
* Added new dir light to check new intensity
* Tested undo
* Opened a new Indoor scene to verify if there is no drastic visual change due to increase from 110k lux to 130k

sebastienlagarde (approved):

PR #2276, opened 2020-10-19
Add metallic remapping for lit and decals

adrien-de-tocqueville:
### Purpose of this PR

When using a mask map, metallic was used as scale factor. Now there is a remapping slider like for smoothness and AO.
Added documentation for Metallic and Metallic remapping

---
### Testing status
Tested that shaders behavior is consistent with previous behavior in the default case (ie. that migration works fine)
Tested that the remapping works


JordanL8 (approved):

TomasKiniulis (approved):
Looks good!
Can validate that remapping works when trying upgrading from 7.3.1, values are saved.
Metallic remapping slider remaps mask map as expected
Removing and readding back mask map keeps the option, and values are correctly saved. 

Project update from 7.5.1
jXAFoDWG1K

Mask Map metallic remap tweaking
AUm6LZ3uNF

adrien-de-tocqueville:
I added a material upgrade step to keep the metallic value when modifying the remapping max value, so that the behavior is consistent with smoothness and ambient occlusion remapping

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/Decal.shader
```diff
@@ -23,13 +23,14 @@ Shader "HDRP/Decal"
         _EmissiveExposureWeight("Emissive Pre Exposure", Range(0.0, 1.0)) = 1.0
 
         // Remapping
+        [HideInInspector] _MetallicRemapMin("_MetallicRemapMin", Range(0.0, 1.0)) = 0.0
```
sebastienlagarde, line 26: it wasn't possible to do the min max here?

com.unity.render-pipelines.high-definition/Documentation~/Decal-Shader.md
```diff
@@ -31,7 +31,7 @@ These properties allow you to set the inputs that affect the behavior of the dec
...
 | **Smoothness**                  | Use the slider to set the strength of the smoothness of the decal. |
-| **Metallic Scale**              | Use the slider to set the strength of the metallic effect of the decal. This targets the values of the Metallic map in the Red channel of the **Mask Map**. Choose a value from 0 and 1 where 0 means no effect and 1 means full effect.<br />This property only appears when you enable the **Metal and Ambient Occlusion properties** checkbox in your [HDRP Asset](HDRP-Asset.md#Decals). |
+| **Metallic Remapping**          | Remaps the metallic values from the **Mask Map** to the range you specify. Rather than [clamping](https://docs.unity3d.com/ScriptReference/Mathf.Clamp.html) values to the new range, it condenses the original range down to the new range uniformly.<br />This property only appears when you enable the **Metal and Ambient Occlusion properties** checkbox in your [HDRP Asset](HDRP-Asset.html#Decals). |
```
sebastienlagarde, line 34: Please, also add this in what's new (that's we have range remapping for metal)  and in upgrade guide (that we have added a new property and rename the one in decal)

sebastienlagarde (approved):
Very great PR!

PR #2279, opened 2020-10-19
Skip passes when everything is disabled in frame settings

FrancescoC-unity:
Also partial fix for https://fogbugz.unity3d.com/f/cases/1281049/ (I say partial because some draw calls are necessary regardless)  

This PR disables lots of passes mostly when opaque and transparent objects are turned off, having thanks to rendergraph very little resources allocated and essentially no draw calls happening 

**I did the change only for the rendergraph path assuming this will be the path to focus on going forward and it is what really culls stuff out,  I can do the change for legacy too**

Before: 

Unity_aKZw4fEeAY

After: 

Unity_w3fPu4FTfw

Also PR changes few little things about rendergraph default resources initialization and small fixes.


JulienIgnace-Unity (changes requested):
lgtm but can we look into consolidating frame settings for opaque/transparent dependent passes as we discussed (there is still the // Merge the two settings TODO_FCC comment about that :) )

FrancescoC-unity:
> lgtm but can we look into consolidating frame settings for opaque/transparent dependent passes as we discussed (there is still the // Merge the two settings TODO_FCC comment about that :) )

So in some cases I cannot do that as they are used singularly in other situation (e.g. object motion vector can be relevant for transparent motion vectors). The only one that I saw that was mergeable was the SSS one, but I'll do another quick pass later today and remove the TODO_FCC :D

FrancescoC-unity:
Done

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.RenderGraph.cs
```diff
@@ -36,7 +36,7 @@ internal static ShadowResult ReadShadowResult(in ShadowResult shadowResult, Rend
         internal void RenderShadows(RenderGraph renderGraph, in ShaderVariablesGlobal globalCB, HDCamera hdCamera, CullingResults cullResults, ref ShadowResult result)
         {
             // Avoid to do any commands if there is no shadow to draw
-            if (m_ShadowRequestCount != 0)
+            if (m_ShadowRequestCount != 0 && hdCamera.frameSettings.IsEnabled(FrameSettingsField.OpaqueObjects))
```
sebastienlagarde, line 39: I am not sure I understand this one, we can render transparent objects inside shadow, we cast shadow from them

FrancescoC-unity: Missing an additional check, that's why :)

JulienIgnace-Unity (approved):

remi-chapelain:
Tested the feature while smoke testing 10.2.0, everything seems to be working as expected. 
Found t[wo minor UI clipping issues](https://fogbugz.unity3d.com/f/cases/1289905/) while testing.

PR #2284, opened 2020-10-20
Restrict sun source to only directional lights

nigeljw-unity:
# Purpose of this PR
Currently, if a light is set to the sun source and it is not a directional light, it will still be set as the main light index. Only directional lights are supported as main lights, so this breaks rendering.

The sun source should be restricted to only directional lights and display warning message if user changes the sun light type to something other than .

https://fogbugz.unity3d.com/f/cases/1278768

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fsun-source-light-type

---
### Testing status
Built a manual test in latest Unity + URP to verify the functionality and fix. Also verified that the fix works with the provided "URPFeatures" test sample given in the bug report.
---
### Comments to reviewers


nigeljw-unity:
image

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -861,14 +861,18 @@ static int GetMainLightIndex(UniversalRenderPipelineAsset settings, NativeArray<
...
-                if (currVisibleLight.lightType == LightType.Directional && currLight.intensity > brightestLightIntensity)
+                if (currVisibleLight.lightType == LightType.Directional)
                 {
-                    brightestLightIntensity = currLight.intensity;
-                    brightestDirectionalLightIndex = i;
+                    // Sun source needs be a directional light
+                    if (currLight == sunLight)
+                        return i;
+
+                    // In case no shadow light is present we will return the brightest directional light
```
martint-unity, line 870: nitpicking here I think. :)
Is this comment correct? I guess it means to say no sun light?

martint-unity (approved):
Just a small comment on my part.

TheoWong-pixel (approved):

phi-lira (approved):

phi-lira:
Closing, merged in : #2390

PR #2286, opened 2020-10-20
Add VT frame settings

FrancescoC-unity:
Pretty much all in the title. 

For now this only skips resolve and clears, but eventually we should use it to drive a sort of multi compile as well.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -132,6 +132,11 @@ public enum FrameSettingsField
         /// <summary>When enabled, HDRP renders custom passes contained in CustomPassVolume components.</summary>
         [FrameSettingsField(0, autoName: CustomPass, customOrderInGroup: 11, tooltip: "When enabled, HDRP renders custom passes contained in CustomPassVolume components.")]
         CustomPass = 6,
+#if ENABLE_VIRTUALTEXTURES
```
sebastienlagarde, line 135: This framesettings must not but #ifdef, it should always be present (but greyd if VT is not enabled), otherwise serialization will fuck us.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -174,8 +174,11 @@ class TempPassData { };
 
                 m_PostProcessSystem.DoUserAfterOpaqueAndSky(m_RenderGraph, hdCamera, colorBuffer, prepassOutput.resolvedDepthBuffer, prepassOutput.resolvedNormalBuffer);
 
-                // No need for old stencil values here since from transparent on different features are tagged
-                ClearStencilBuffer(m_RenderGraph, colorBuffer, prepassOutput.depthBuffer);
+                if (hdCamera.frameSettings.IsEnabled(FrameSettingsField.OpaqueObjects)) // If we don't have opaque objects there is no need to clear.
```
JulienIgnace-Unity, line 177: I don't think this is supposed to be in this PR

FrancescoC-unity: Yup, bad switching of branches! Fixing it

JulienIgnace-Unity (changes requested):
Leftover from another PR seems to be in this one.

JulienIgnace-Unity (approved):

PR #2289, opened 2020-10-20
Now Receive SSR also controls Receive SSGI on opaque objects.

anisunity:
We found out that in deferred mode, the emissive was lost on the pixels where the SSGI was valid or RTGI was computed. While the real solution for this is to have a separate forward emissive pass, we decided to still have a work around for the package release.

Now on opaque object (shader or SG), the name of receive ssr and tooltip were changed to this.
image

Here is an image that shows how the materials behave in deferred (Receive SSGI Off vs On)
image

Here is an image that shows how the materials behave in forward(Receive SSGI Off vs On)
image

Also one of the path tracing scenes was using the result of an other test as the texture of an area light, had to change that as I changed the result of the other test, now it is using a normal texture.

Testing status
Updated the RTGI scene to have an emissive object with and without SSGI.
The DXR tests were green locally.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinGIUtilities.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinGIUtilities.hlsl
@@ -123,15 +123,15 @@ void SampleBakedGI(
     bakeDiffuseLighting = float3(0, 0, 0);
     backBakeDiffuseLighting = float3(0, 0, 0);
 
-    // Check if we are RTGI in which case we don't want to read GI at all (We rely fully on the raytrace effect)
-    // The check need to be here to work with both regular shader and shader graph
-    // Note: with Probe volume it will prevent to add the UNINITIALIZED_GI tag and
-    // the ProbeVolume will not be evaluate in the lightloop which is the desired behavior
-    // Also this code only needs to be executed in the rasterization pipeline, otherwise it will lead to udnefined behaviors in ray tracing
+    // Check if we are RTGI in which case we don't want to read GI at all (We rely fully on the raytrace effect)	
+    // The check need to be here to work with both regular shader and shader graph	
+    // Note: with Probe volume it will prevent to add the UNINITIALIZED_GI tag and	
+    // the ProbeVolume will not be evaluate in the lightloop which is the desired behavior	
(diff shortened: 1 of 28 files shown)


com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinGIUtilities.hlsl
```diff
@@ -123,16 +123,6 @@ void SampleBakedGI(
     bakeDiffuseLighting = float3(0, 0, 0);
     backBakeDiffuseLighting = float3(0, 0, 0);
 
-    // Check if we are RTGI in which case we don't want to read GI at all (We rely fully on the raytrace effect)
-    // The check need to be here to work with both regular shader and shader graph
-    // Note: with Probe volume it will prevent to add the UNINITIALIZED_GI tag and
-    // the ProbeVolume will not be evaluate in the lightloop which is the desired behavior
-    // Also this code only needs to be executed in the rasterization pipeline, otherwise it will lead to udnefined behaviors in ray tracing
-#if !defined(_SURFACE_TYPE_TRANSPARENT) && (SHADERPASS != SHADERPASS_RAYTRACING_INDIRECT) && (SHADERPASS != SHADERPASS_RAYTRACING_GBUFFER)
```
sebastienlagarde, line 131: Why do you remove this check? it prevent to sample lightmaps

anisunity: If we have RTGI, the value is overriden anyway by the lerp in the lightloop. We don't need the divergence of the code here, or maybe I am missing something?

JulienIgnace-Unity (approved):

remi-chapelain (approved):
Ideally we should have two separate flags but, I understand the constraints and it's still better than having the aforementioned issue. ✔️ 
Tested with a new project, also tested in forward and with transparent objects still have the proper SSR Transparent flag

sebastienlagarde:
> Ideally we should have two separate flags but, I understand the constraints and it's still better than having the aforementioned issue. ✔️
> Tested with a new project, also tested in forward and with transparent objects still have the proper SSR Transparent flag

Yup, stencil bit are extremely expensive, thus why.

sebastienlagarde (approved):

PR #2292, opened 2020-10-20
Expose HDRP shader tag IDs and per object config

alelievr:
### Purpose of this PR

When writing custom passes, users that want to render objects usually have to hardcode the shader passes and per-object data in their code. With this PR, they can use the properties we have in HDRP to make sure their work won't break when we change a render config or pass name.

---
### Testing status
I copy-pasted the DrawRenderersCustomPass.cs into a project and checked that it compiles. the DrawRenderers custom pass was indeed using internal pass names and render configs and some users were complaining about it.


JulienIgnace-Unity (approved):
We have to make sure we really want that :)
Once it's public, it's forever (tm)

Otherwise, looks good to me.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
```diff
@@ -15,8 +15,10 @@ namespace UnityEngine.Rendering.HighDefinition
     /// </summary>
     public class HDUtils
     {
-        internal const PerObjectData k_RendererConfigurationBakedLighting = PerObjectData.LightProbe | PerObjectData.Lightmaps | PerObjectData.LightProbeProxyVolume;
-        internal const PerObjectData k_RendererConfigurationBakedLightingWithShadowMask = k_RendererConfigurationBakedLighting | PerObjectData.OcclusionProbe | PerObjectData.OcclusionProbeProxyVolume | PerObjectData.ShadowMask;
+        /// <summary>Render configuration for baked static lighting, this value can be used in a RendererListDesc call to render Lit objects.</summary>
+        public const PerObjectData k_RendererConfigurationBakedLighting = PerObjectData.LightProbe | PerObjectData.Lightmaps | PerObjectData.LightProbeProxyVolume;
```
JulienIgnace-Unity, line 19: Change this to accessors as discussed on slack

JulienIgnace-Unity (changes requested):
Change to accessors for Per Object Data

JulienIgnace-Unity (approved):

PR #2293, opened 2020-10-20
[11.x.x] Fixed per camera MSAA regression 

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR
- Fixed per camera MSAA regression introduced by XR change. case https://fogbugz.unity3d.com/f/cases/1282963/ 

---
### Testing status

---
### Comments to reviewers
- Fixed xrRendering is true when no XR display is available.
- If no XR display provider is installed, when creating the new camera, the camera target eye will now set to None by default.

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
@@ -8,9 +8,6 @@ using UnityEngine;
 using UnityEngine.Rendering;
 using UnityEngine.Rendering.Universal;
 using Object = UnityEngine.Object;
-#if ENABLE_VR && ENABLE_XR_MODULE
-using UnityEngine.XR;
-#endif
 
 namespace UnityEditor.Rendering.Universal
 {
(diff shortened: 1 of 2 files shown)


phi-lira (approved):

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -193,8 +196,15 @@ public new void OnEnable()
...
             {
                 var additionData = (cameraTarget as Component).gameObject.GetComponent<UniversalAdditionalCameraData>();
-                if(additionData == null)
+                if (additionData == null)
+                {
                     additionData = (cameraTarget as Component).gameObject.AddComponent<UniversalAdditionalCameraData>();
+#if ENABLE_VR && ENABLE_XR_MODULE
+                    List<XRDisplaySubsystemDescriptor> displayDescriptors = new List<XRDisplaySubsystemDescriptor>();
+                    SubsystemManager.GetSubsystemDescriptors(displayDescriptors);
+                    additionData.allowXRRendering = displayDescriptors.Count > 0;
```
pbbastian, line 205: I don't think we should be changing serialized component data in `OnEnable`. We shouldn't have to change this value based on the presence of XR. It should be enough to have the checks during rendering.

pbbastian (changes requested):

martint-unity (approved):

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
thomas-zeng: For the fogbugz case, the runtime check is good enough. 
The code here improved the default value for camera's target eye. It will defaults new camera's target eye field to None when it gets created in editor and no XR provider is installed.

I will revert the change here and put it into another PR. We can review it separately.

thomas-zeng: Addressed in 
1d87eaa

pbbastian (approved):

fabien-unity (approved):

phi-lira:
merging to stage branch

PR #2298, opened 2020-10-21
Material upgrade cut out issue

martint-unity:
---
### Purpose of this PR
fix case [1235516](https://fogbugz.unity3d.com/f/cases/1235516/) upgrade material cutout issue
if Alpha test is on, alpha clip should be on

Tested the test project from the fogbugz case.


phi-lira (approved):

Verasl (approved):
LGTM

ellioman:
The Yamato issues are not related to this PR.
- Test packages is complaining because of Unity version as this ran against a custom revision plus removal of Deferred shaders.
- OSX Metal fails because of a known lost connection issue.
- HDRP issues not related to this PR.

PR #2300, opened 2020-10-21
Fix undo/redo compositor

adrien-de-tocqueville:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1285351/

Added undo redo support for the compositor enable button.
Also fix a type in a fonction name and fix a warning on undo

---
### Testing status
Tested combinations of undo/redo after compositor enable/disable
compositor_undo


pmavridis (approved):

remi-chapelain (approved):
Testing seems sufficient, approving ✔️

pmavridis:
Out of curiosity, does the undo also remove the default compositor resources that we add to the project  (in the Assets/Compositor folder) when enabling the compositor? Even if these resources are not removed, I think we are still ok with this PR. It is just one shader graph and .asset file that are unused.

adrien-de-tocqueville:
No, I hesitated to do it, but since the "Remove Compositor from scene" button doesn't delete them, i didn't

pmavridis:
> No, I hesitated to do it, but since the "Remove Compositor from scene" button doesn't delete them, i didn't

Thanks for the clarification. Generally I think it is better to keep the resources after removing the compositor: If someone has spend some hours making and configuring a composition graph for his project, I guess he does not want this work to disappear.

PR #2304, opened 2020-10-21
Adding an "Include For Ray Tracing" toggle on lights

anisunity:
Adding an "Include For Ray Tracing" toggle on lights to allow the usr to exclude them when ray tracing is enabled in the frame settings of a camera. The option can be found in the advanced mode of the emissive section of the light component. The light is excluded in all the light structures (Tile, Cluster and RT Cluster).

image

Testing status
Ran the DXR tests locally, they are green.
Added ray tracing excluded light to the tests (that tests ray tracing and rasterization).


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -976,6 +976,8 @@ static void DrawEmissionAdvancedContent(SerializedHDLight serialized, Editor own
                 --EditorGUI.indentLevel;
             }
 
+            EditorGUILayout.PropertyField(serialized.includeForRayTracing, s_Styles.includeLightForRayTracing);
```
sebastienlagarde, line 979: I think we should show this field only when Raytracing is enabled. What do you think?

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.Skin.cs
```diff
@@ -44,6 +44,7 @@ sealed class Styles
             public readonly string cookieNonPOT = "HDRP does not support non power of two cookie textures.";
             public readonly string cookieTooSmall = "Min texture size for cookies is 2x2 pixels.";
             public readonly string cookieBaking = "Light Baking for cookies disabled on the Project Settings.";
+            public readonly GUIContent includeLightForRayTracing = new GUIContent("Include For RayTracing", "When checked, the light will affect the scene for a camera that has the Ray-Tracing frame setting enabled.");
```
sebastienlagarde, line 47: >"Include For RayTracing"
@JordanL8  We are trying to find a label to say that a light is enabled when performing a rendering with a Camera that have a Framesettings "Raytracing" enable.  What do you think is the best wording for this? is Influce for Raytracing good? thanks

anisunity: this was discussed with lewis before the PR

JordanL8: Yeah I thought Include For Raytracing made sense. For the tooltip, how about:

When enabled, the light affects the scene for cameras that have the Ray-Tracing frame setting enabled.

JordanL8 (changes requested):
Left feedback for the tooltip. Also, can you add this property to the [light documentation](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.high-definition/Documentation~/Light-Component.md)?

The description can be:
Indicates whether this light affects the scene for cameras with the **Ray-Tracing** frame setting enabled.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #2307, opened 2020-10-21
Remove the _BLENDMODE_PRESERVE_SPECULAR_LIGHTING keyword

FrancescoC-unity:
As the title say, that property is removed and instead we use the shader property _EnableBlendModePreserveSpecularLighting . 

Tests locally are fine, but I already kicked yamato here as it will have a wider coverage and in this case won't hurt. 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fremove-blend-preserve-specular-keyword/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/3905997/job


alelievr (approved):

FrancescoC-unity:
Closing as I am changing branch.

PR #2308, opened 2020-10-21
Fix SampleGradientNode colorspace

elizabeth-legros:
SampleGradientNode assumed input gradient was in gamma space, but the color picker uses Linear space HDR values. Changed colorspace conversion to reflect this.
https://fogbugz.unity3d.com/f/cases/1286164/


alindmanUnity (approved):

LandonTownsendUnity:
The gradient preview UI is still wrong, any way to fix that?
image

LandonTownsendUnity:
> The gradient preview UI is still wrong, any way to fix that?
> image

OK, this was actually due to an out of date branch.

However, it looks like the problem is still there for the inline gradient node:
image

cdxntchou (changes requested):
Verified that Gradient Property and Gradient Slot show matching colors across gradient widget, gradient popup editor, shadergraph previews, and in game.

Works when Color mode is set to linear, but when color mode is set to gamma, the gradient popup editor and gradient widget show incorrect color values.  That's in trunk code though, so not fixable from this PR.  The output color values in preview and scene are correct though, so at least the part under our control here is correct.

One issue to fix:  Gradient Nodes are still using the gamma space gradient editor:
image

ghost (approved):
Gradient node inspector fixed, approving.

cdxntchou (approved):

alindmanUnity:
Failing job on OSX unrelated, and fixed by this pr: https://github.com/Unity-Technologies/Graphics/pull/2962

PR #2310, opened 2020-10-21
9.x.x/universal/fix gles2

kaychang-unity:
---
### Purpose of this PR
Fixed shader compilation bugs with gles2.

---
### Testing status
I built Android GLES2 player and verified all shaders now build.

---
### Comments to reviewers
I haven't updated the changelog because it seems it should be SRP revision past 9.0.0. Which one should it be?

### DIFF
--- a/com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
+++ b/com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
@@ -219,6 +219,8 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
             }
 
             HLSLPROGRAM
+            #pragma exclude_renderers gles
+            #pragma target 4.5
 
             #pragma multi_compile _ _SPOT
 
@@ -253,6 +255,8 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
             }
 
             HLSLPROGRAM
+            #pragma exclude_renderers gles
+            #pragma target 4.5
 
             #pragma multi_compile _POINT _SPOT
             #pragma multi_compile_fragment _LIT
@@ -292,6 +296,8 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
             }
 
             HLSLPROGRAM
+            #pragma exclude_renderers gles
+            #pragma target 4.5
 
             #pragma multi_compile _POINT _SPOT
             #pragma multi_compile_fragment _SIMPLELIT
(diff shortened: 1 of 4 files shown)


kaychang-unity:
ABV fast-2020.1 is broken, but normal ABV 2020.1 is green: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/9.x.x%252Funiversal%252Ffix-gles2/.yamato%252F_abv.yml%2523all_project_ci_2020.1/3945114/job

Ready to merge.

com.unity.render-pipelines.universal/Shaders/Utils/TileDepthInfo.shader
```diff
@@ -35,6 +35,277 @@ Shader "Hidden/Universal Render Pipeline/TileDepthInfo"
...
+        half listDepthFactorB = f16tof32(listDepthRange >> 16); // b = 32.0 / (listMaxDepth - listMinDepth)
+
+        uint geoBitmask = 0;
+
+        #if !USE_GATHER // Base reference version.
+            DOWNSAMPLING_LOOP for (int j = 0; j < downsamplingSize.y; ++j)
+            DOWNSAMPLING_LOOP for (int i = 0; i < downsamplingSize.x; ++i)
+            {
+                int2 tx = sourceCorner + int2(i, j);
+                tx.y = _DepthTexSize.y - tx.y; // TODO fixme: Depth buffer is y-inverted
```
eh-unity, line 123: does this depend on API? gl uv(0,0) == left-bottom

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
```diff
@@ -219,6 +219,8 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
             }
 
             HLSLPROGRAM
+            #pragma exclude_renderers gles
```
eh-unity, line 222: Should these actually be  #pragma exclude_renderers d3d11_9x gles gles3 glcore

eh-unity (approved):
Looks pretty good to me.

com.unity.render-pipelines.universal/Shaders/Utils/TileDepthInfo.shader
kaychang-unity: It could, although for the purpose of downsizing the depth buffer, it should not matter. GL is also not officially supported for deferred renderer. I think the comment above is referring to URP y-flipping the rendering, which causes issues later on as the logic for numbering tile is (0,0) for top-left corner.

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
kaychang-unity: Yes indeed!
I will make this fix in the master branch if that's ok, because the correct fix may require to declare fallback shaders. Deferred render is actually not enabled in 9.x.x, so this PR is just fixing compiling errors.

PR #2312, opened 2020-10-21
[ShaderGraph] Fix for Public API breakage

cdxntchou:
### Checklist for PR maker
- [ ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.

---
### Purpose of this PR
Fixing the PackageValidationTest errors, caused by public types being removed by a #if when the VFX package is not present.

The issue was introduced originally by https://github.com/Unity-Technologies/Graphics/pull/1340 -- I think to handle some test naming collisions for VFXTarget, but the #if was extended to the ShaderGraphVFXAsset for some reason.  @PaulDemeulenaere Can you verify that it should be ok to keep the ShaderGraphVFXAsset declaration?

---
### Testing status
Describe what manual/automated tests were performed for this PR

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #1340, opened 2020-07-20
[VFX/SG] Workflow improvement with SG

PaulDemeulenaere:
### Purpose of this PR
**1.** Avoid listing of `VFXTarget` (and so, Create VFX Graph), if visualeffectgraph package isn't present.
_vfx_exclude_when_no_package

**2.** Rework shaderImporter approach when we are mixing shader & VFX.
_mix_vfx_sg

**3.** Allow preview when there isn't only vfx target

*Before*
_before_preview

*After*
_after_preview

---
### Testing status
Tested locally.

I updated data from URP test to cover these different cases : 
- ReferenceSampleColor & SampleColor are sharing the same shader.
- ReferenceSampleDepth & SampleDepth are still independant but generates both target.
- ReferenceSampleGI is URP only, SampleGI is VFX only (as before).

*Before*
image

*After*
image

PR #2313, opened 2020-10-21
[ShaderGraph] [backport] Fix for Public API breakage

cdxntchou:
### Purpose of this PR

Backport of https://github.com/Unity-Technologies/Graphics/pull/2312 to 10.x

Fixing the PackageValidationTest errors, caused by public types being removed by a #if when the VFX package is not present.

The issue was introduced originally by #1340 -- I think to handle some test naming collisions for VFXTarget, but the #if was extended to the ShaderGraphVFXAsset for some reason. @PaulDemeulenaere Can you verify that it should be ok to keep the ShaderGraphVFXAsset declaration?

---
### Testing status
Tested shadergraph CI still passes, quick smoke test of shadergraph VFX path -- can use VFX Target with VFX package present.
Tested ShaderGraph still compiles with URP and no VFX package present, and VFX Target does not show up.
Ran Package Validation locally, passes (with warnings).

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Generation/ShaderGraphVfxAsset.cs
+++ b/com.unity.shadergraph/Editor/Generation/ShaderGraphVfxAsset.cs
@@ -1,4 +1,3 @@
-#if VFX_GRAPH_10_0_0_OR_NEWER
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -195,4 +194,3 @@ namespace UnityEditor.ShaderGraph.Internal
         }
     }
 }
-#endif


elizabeth-legros (approved):

cdxntchou:
Merging to unblock nightly build.   ABV still in progress, but seems green / match previous ABV so far.

PaulDemeulenaere:
Thanks for the fix !
ShaderGraphVfxAsset has been removed because VFXTarget is also excluded, this class wasn't needed anymore.
I think keeping ShaderGraphVfxAsset hasn't any side effect.
Checked locally, it behave as expected. :
**With URP only**
_universal

**With URP & VFX**
_universal_with_vfx

PaulDemeulenaere (commented):
🟢

PR #2316, opened 2020-10-22
Hdrp/update decal handles

RSlysz:
## Purpose of this PR
- Fixed issue where pivot of DecalProjector was not aligned anymore on Transform position when manipulating the size of the projector from the Inspector (on Z axis). (Was working well with handles, issue is only when manipulating from Inspector window).
- Removed the green coloration of the UV gizmo in the scene view (and only appear in edition mode).
- Highlight the projection plane with bolder edge.
- Separated the Depth in the inspector from width and height for more intuitive manipulation (formerly was X Y Z).

---
### Testing status
- Manually testing the pivot is always aligned with transform on Z.
- Manually tested tooltips for sube label on the size line.
- Manually tested that projection plane highlighted edge follow the projection volume in the gizmo when manipulating it.
- Manually tested that label Width and Height can be dragged to change the value.

**Not Tested:** if Decal still appears only when projection volume intersect a surface. (This logic have not changed so I am pretty confident on this)

---
### Comments to reviewers
Without edition mode:
image

With edition mode (both look the same):
image


JordanL8 (approved):

iM0ve (approved):
I think the testing made is sufficient for this change. Approving

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
```diff
@@ -302,17 +312,34 @@ static void DrawGizmosSelected(DecalProjector decalProjector, GizmoType gizmoTyp
...
-                using (new Handles.DrawingScope(face, Matrix4x4.TRS(decalProjector.transform.position - decalProjector.transform.rotation * (decalProjector.size * 0.5f + decalProjector.offset.z * Vector3.back), decalProjector.transform.rotation, Vector3.one)))
+                //draw UV and bolder edges
+                using (new Handles.DrawingScope(Matrix4x4.TRS(decalProjector.transform.position - decalProjector.transform.rotation * (decalProjector.size * 0.5f + decalProjector.offset.z * Vector3.back), decalProjector.transform.rotation, Vector3.one)))
                 {
-                    Handles.DrawSolidRectangleWithOutline(new Rect(start, size), face, Color.white);
+                    if (inEditMode)
+                    {
+                        Vector2 size = new Vector2(
+                            (decalProjector.uvScale.x > 100000 || decalProjector.uvScale.x < -100000 ? 0f : 1f / decalProjector.uvScale.x) * decalProjector.size.x,
+                            (decalProjector.uvScale.x > 100000 || decalProjector.uvScale.x < -100000 ? 0f : 1f / decalProjector.uvScale.y) * decalProjector.size.y
```
alelievr, line 322: Isn't it supposed to be `decalProjector.uvScale.y` here instead of `decalProjector.uvScale.x` ?

alelievr: Also, can `uvScale` be 0?

RSlysz: no but the representation or it which I comput here tend to 0

RSlysz:
Fix above solve issue where:
You attempt to use shortcut in the SceneView. But the Decal is not selected

sebastienlagarde (approved):

PR #2317, opened 2020-10-22
Fix bleeding of a view into another caused by TAA

FrancescoC-unity:
Fix for: https://fogbugz.unity3d.com/f/cases/1283616/ 

It was the usual case of bad clamping of UVs when using RTHandleScale 

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-TAA-bleeding-with-two-views/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/3919408/job

**What I tested**: repro case, and also tried rescaling the windows and then try the repro again to make.


FrancescoC-unity:
NVM found a local issue, moving back to draft :/

JulienIgnace-Unity (approved):

remi-chapelain (approved):
Tried to repro on a brand new template. Repro-ed and fixed with the PR. 
Couldn't manage to make the bleeding appear again with the fix. ✔️

PR #2318, opened 2020-10-22
[WIP] Adding test for Shader API Breakages

ellioman:
# Purpose of this PR
Currently we have no automatic detection for Shader API breakages that can occur in URP Shaders. This PR aims to improve that by adding a new test that compares the current version to the latest we have stored.

## How it works:
- Whenever we release a package we manually create a JSON file containing all the Shader API from our ShaderLibrary and a PR with that.
- In the new test, we:
-- find the JSON File that's closest to the one that's currently in the project
-- search for each function and the variation of it (variations = same function name but different parameters)
-- check if the function is still there or has been moved to deprecated.hlsl
-- check if all the parameters are still valid, adding a in, inout, out and some type changes are considered failures

## Current issues to solve
- The test doesn't seem to run on Windows platforms
- What type changes should be considered a breaking change? int <=> float/real/half?  float => half? etc…

## Example of a passing test
image

## Example of a failing test
image

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fshader-api-testing


Verasl:
This should definitnely be backported/remade for 7.x.x and maybe 10.x.x too.

ellioman:
> This should definitely be backported/remade for 7.x.x and maybe 10.x.x too.

This should work with all version of URP. All we need to do is create the JSON file for the packages we want to test against.
So for example if we want to test the next 7.x.x release, we need to go back to the last 7.5 release, create the Shader API JSON file and get that in. Then it should work.

I can do that in the backport PR's. :)

phi-lira:
Closing this PR as it has been staled for a while. I think this is nice work to land, we need to schedule it. Please reopen when planned to work on it.

PR #2319, opened 2020-10-22
Fix nan when decal affect normals

adrien-de-tocqueville:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1282691/
When a mesh has normal equal to zero, the normalize creates nan.

---
### Testing status
Tested on the repro scene.
Warning: nans seem to only appears on nvidia gpus


JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Thanks! Checked repro case. Works as expected now with my Nvidia gpu

sebastienlagarde:
We often have problem with normal at 0.  Anis have added a debug mode dedicated to this. I don't think we should hide the problem with a safeNormalize, as if a normal is 0 everything will be fucked anyway. Best to fixe the root cause. For now I will revert this change and we should try to have a way to tell users that there is a problem and our debug mode is one way but how to tell that to use it?

PR #2320, opened 2020-10-22
Avoid errors if user define null params in volume component

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1279917/ 

The pattern that is mentioned in the bug is a bug in itself, but indeed we don't do much to help the user finding it and we instead just block everything and spam errors. 

This change add some guards to avoid the error and adds a warning to point out which volume component has a faulty parameter (Cannot really print parameter name though sadly). 

**What did I test**: the repro project :)


JulienIgnace-Unity (approved):

JMargevics (approved):
No errors with uncommented line 70 anymore. ✅

sebastienlagarde (approved):

Chman:
The fix itself is fine but I see a lot of other unrelated fixes in this PR. Did a merge go wrong or...?

sebastienlagarde:
@Chman  PR updated, can you review, thanks

Chman (approved):

PR #2322, opened 2020-10-22
Reset RT Handle size every now and then in editor if set to a big resolution and then reverted back

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1286360/

This is a bit iffy since it will trigger a reset even when nothing is wrong and user just operates on mixed resolution, one of which being above 1080p. 

However, since we do it only in editor it shouldn't be a big problem and at least solve an issue of absolute chugging (due to swap in and out of memory) if a big resolution is set by mistake. 

**What did I test**: I looked at taskmanager while doing the repro steps and notice that before this PR if set to a huge dimension (8k by 8k) memory would stay super high when going back to 1080p, which this change, after a bit memory gets back to the expected usage.


sebastienlagarde (approved):

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Sweet, checked bug report repro with split and not split game/scene views, no performance lag occurs when setting back to regular resolution after a while, ResetReferenceSize is correctly not called if exceptionally high resolution stays set

PR #2325, opened 2020-10-22
Fix POM node not working with subgraphs

alelievr:
### Purpose of this PR
Fixed https://fogbugz.unity3d.com/f/cases/1221317/

---
### Testing status
I tested the POM node in a subgraph, as well as in a normal ShaderGraph. I verified that we can still put two POM nodes in a single graph.
I checked that the POM code we use in the HDRP lit shader still works.


cdxntchou (approved):
Looks good!

sebastienlagarde (approved):

alindmanUnity (approved):

sebastienlagarde:
Yamato is green (current failure are the regular master failure)

PR #2328, opened 2020-10-22
[10.x.x] [ShaderGraph] Backport of Fix property declaration bugs, simplify system

cdxntchou:
### Purpose of this PR
Backport of Original PR:  https://github.com/Unity-Technologies/Graphics/pull/2226

Fixes these bugs:
https://fogbugz.unity3d.com/f/cases/1256374/

Non-exposed VT properties will version themselves to Exposed.
The "Expose" checkbox is now disabled, so the user cannot re-expose them.
Note that VT properties do not allow the user to override their declaration, as they must be "UnityPerMaterial".
image

https://fogbugz.unity3d.com/f/cases/1285701/

HDRP target-injected properties are now flagged so they are removed if not needed by the Shader.
Verified by looking at the generated shader code, now those properties are no longer declared in HLSL:
image

https://fogbugz.unity3d.com/f/cases/1285700/

Existing properties will version themselves so their behavior is unchanged.  This means "non-exposed hybrid" properties get versioned to "override to Global" to maintain the old behavior.
image

Exposed hybrid properties get versioned to "override to HybidPerInstance"
image

Non-hybrid properties keep their existing behavior.  Exposed ==> UnityPerMaterial,    Non-exposed ==> Global.
Since this is the default, no override is necessary here.
image

And from now on, if the user sets their Property to "override HybridPerInstance", they can expose or unexpose the property without affecting the hybrid declaration.  I.e. this now does the right thing, declaring a hybrid property that doesn't get added to the Material:
image

https://fogbugz.unity3d.com/f/cases/1255951/

Matrices cannot be Exposed, which meant previously they were always treated Global, even if flagged as hybrid!

Old "hybrid" matrices are turned into "Override to Global" to match old behavior (they were never hybrid!)
User can change the override if they really DO want them to be hybrid.
image

Old non-hybrid matrices are also turned into "Override to Global" to match old behavior.

image

NEW matrix properties now default to "UnityPerMaterial" by default, to match the default state of the other float/vector/bool properties.
image

You can now flag Matrices to _actually_ be hybrid by using the override:
image

https://fogbugz.unity3d.com/f/cases/1235070/
Already fixed by https://github.com/Unity-Technologies/Graphics/pull/2187
But we change the code in this PR, so I re-validated that it still works correctly.

---
### Testing status
Created a shadergraph using every possible OLD property type with all possible combinations of Exposed and Hybrid flags.  Compared the generated shader code of the OLD system, with the same shadergraph versioned with this PR.  The only differences in the code were caused by intentional bug fixes to address the bugs above.  Detected differences:

1) hidden VT properties are now not hidden -- instead of global, they are declared in UnityPerMaterial.
     This is ok, because the hidden VT properties did not ever work, there should be no existing valid content.

2) HDRP target-injected variables removed.  Double checked with Seb that this is the correct list of variables to remove.  Ran full HDRP tests to verify.

3) Texture and Sampler declarations are separated to their own lines, instead of being a single line.

4) Virtual Texture fallback Texture declarations (used when VirtualTexturing is disabled) no longer declare their TexelSize properties, as they are unused by the VT sampling node.

That's it!

Also ran ShaderGraph and HDRP CI tests locally.
Ran HDRP and Universal Hybrid tests locally.
Ran VFX tests locally, verified matches exactly master results

Tested VFX Graph integration, checking that properties are connected.

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Graphs/AbstractShaderProperty.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/AbstractShaderProperty.cs
@@ -15,6 +15,13 @@ namespace UnityEditor.ShaderGraph.Internal
         [SerializeField]
         Precision m_Precision = Precision.Inherit;
 
+        [Obsolete("AbstractShaderProperty.gpuInstanced is no longer used")]
+        public bool gpuInstanced
+        {
+            get { return false; }
+            set { }
+        }
+
         // NOTE: this does not tell you the HLSLDeclaration of the entire property...
         // instead, it tells you what the DEFAULT HLSL Declaration would be, IF the property makes use of the default
         // to check ACTUAL HLSL Declaration types, enumerate the HLSL Properties and check their HLSLDeclarations...
@@ -207,9 +214,6 @@ namespace UnityEditor.ShaderGraph.Internal
                 "TEXTURE2D_ARRAY",
                 "SAMPLER",
         };
-
-        [Obsolete("gpuInstanced is no longer used")]
-        public bool gpuInstanced { get { return false; } set { } }
 
         public string GetValueTypeString()
         {


com.unity.shadergraph/Documentation~/Property-Types.md
```diff
@@ -8,13 +8,15 @@ Each property has an associated **Data Type**. See [Data Types](Data-Types.md) f
...
-In addition to values specific to their [Data Types](Data-Types.md), all properties have the following parameters.
+In addition to values specific to their [Data Types](Data-Types.md), most properties have the following common parameters.
 
 | Name        | Type  | Description |
 |:------------ |:---|:---|
 | Display Name | String | The display name of the property |
 | Exposed | Boolean | If true this property will be exposed on the material inspector |
 | Reference Name | String | The internal name used for the property inside the shader |
+| Override Property Declaration | Boolean | An advanced option to enable explicit control of the shader declaration for this property |
+| Shader Declaration | Enumeration | Controls the shader declaration of this property |
```
marctem, line 19: This should get its own section explaining the various options.

cdxntchou: Yeah, Sharlene wants to brush up this entire page.  Likely can do it as part of that.

cdxntchou:
ABV complete:   https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/10.x.x%252Fsg%252Ffix%252Fproperty-declarations/.yamato%252F_abv.yml%2523all_project_ci_2020.2/3932268/job/pipeline

Corresponding 10.x.x/release ABV:  https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/10.x.x%252Frelease/.yamato%252F_abv.yml%2523all_project_ci_2020.2/3931812/job/pipeline

Failing issues: 
    Universal_Hybrid on OSX_Metal_playmode_mono_apiNet2_Linear on version 2020.2 -- same on both
    Build Universal on Win_DX11_mono_apiNet2_Linear_Player on version 2020.2  -- re-run succeeded, same on both
    HDRP on Win_DX12_playmode_XR_mono_apiNet2_Linear on version 2020.2 -- both seem to have an issue where it executes for a very long time..
    HDRP on OSX_Metal_playmode_mono_apiNet2_Linear on version 2020.2  -- re-runs failed on both
    Universal on Win_DX11_Standalone_mono_apiNet2_Linear on version 2020.2 -- re-run succeeded, same on both
    Universal on OSX_Metal_playmode_mono_apiNet2_Linear on version 2020.2  -- re-run succeded same on both
    VFX_HDRP on Win_DX11_Standalone_mono_apiNet2_Linear on version 2020.2 -- both failed in same way
    VFX_HDRP on Win_DX11_playmode_mono_apiNet2_Linear on version 2020.2-- both failed in same way
    Test all packages [project context] Win 2020.2  -- both failed in same way
    Test minimum editor version - all packages [project context] Win -- both failed in same way
    Test all packages [project context] OSX 2020.2 -- both failed in same way
    Test minimum editor version - all packages [project context] OSX -- both failed in same way
    VFX_URP on Win_DX11_playmode_mono_apiNet2_Linear on version 2020.2  -- both failed in same way

LandonTownsendUnity (approved):
Did a basic smoke test (created a shadergraph, created some properties, changed the "override property declaration" options, saved the shadergraph), no problems.

PR #2295, opened 2020-10-21
Improvements to the DXR Wizard

anisunity:
On user request, we decided to add new checks and additional informaiton in the DXR wizard. Now the DXR wizard displays a warning message for unchecked feature in the HDRP asset and and an info message for the corresponding an unchecked default frame setting. The wizard also displays the name of the target asset that may require modification.

image

Testing status
Clicked on all the asset features and default camera frame settings and made sure the right messages popped when they are checked/unchecked.

PR #2329, opened 2020-10-22
Standardize naming for Transparent receive SSR

FrancescoC-unity:
"Fix" for this QoL case https://fogbugz.unity3d.com/f/cases/1287008/

Essentially now everything says "Transparent Receive SSR" and the option on the material always say only Receive SSR (but set the right flag depending whether is transparent or not)


remi-chapelain (changes requested):
I think it's much more clear that way. 
Although while checking I found two small things : 

1. We should either write "Transparent**s** receive SSR" or "Transparent receive**s** SSR". On most cases we are talking about all the transparents so, I'll suggest making it plural
2. I noticed a copy/pasted typo on the wizard here "th**r**ough" instead of "though", maybe we can correct this also :)
wizard_through

anisunity:
this PR is highly conflicting with this one  https://github.com/Unity-Technologies/Graphics/pull/2295

remi-chapelain (approved):

JulienIgnace-Unity (approved):

JulienIgnace-Unity:
Will merge this PR first and then https://github.com/Unity-Technologies/Graphics/pull/2295 since it's still missing some approvals

PR #2330, opened 2020-10-22
Added default volume profile asset change to the undo stack (case 1285268).

eturquin:
Changes, user-wise:

* Before: when changing the default volume profile asset in the project settings, and trying to undo this change, this would not work (because this change was not registered on the undo stack)

* After: when changing the default volume profile asset in the project settings, and trying to undo this change, this will now work as expected and the change will be reverted

For more details, see the corresponding bug report:
https://fogbugz.unity3d.com/f/cases/1285268/


iM0ve (approved):
I think the testing is sufficient for the fix and should be okay to merge

JulienIgnace-Unity (approved):

PR #2337, opened 2020-10-23
Fix volument component creation via script

adrien-de-tocqueville:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1284935/

Creating a volume component by script corrupts the volume profile.

---
### Testing status

I tested the repro script of the case, the profile is not corrupted anymore


adrien-de-tocqueville:
#2369

PR #2338, opened 2020-10-23
Hdrp/standardize transparent receive ssr

JulienIgnace-Unity:
Redid PR after wrong merge: https://github.com/Unity-Technologies/Graphics/pull/2329

"Fix" for this QoL case https://fogbugz.unity3d.com/f/cases/1287008/

Essentially now everything says "Transparent Receive SSR" and the option on the material always say only Receive SSR (but set the right flag depending whether is transparent or not)


com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
```diff
@@ -150,11 +150,11 @@ public ConfigStyle(string label, string error, string button = resolve, MessageT
                 label: "Reflection (Default Camera Frame Setting)",
                 error: "Screen Space Reflection is disabled in the default Camera Frame Settings. This means Cameras that use these Frame Settings do not render ray-traced reflections. To enable this feature, go to Project Settings > HDRP Default Settings > Frame Settings > Default Frame Settings For Camera > Lighting and enable Screen Space Reflections", messageType: MessageType.Info);
             public static readonly ConfigStyle dxrTransparentReflections = new ConfigStyle(
-                label: "Transparent Reflection (Asset)",
-                error: "Transparent Screen Space Reflection is disabled in the current HDRP Asset which means you cannot enable ray-traced reflections for transparent GameObjects from Volume components. To enable this feature, open your HDRP Asset, go to Lighting > Reflections, and enable Transparent Screen Space Reflections", messageType: MessageType.Warning);
+                label: "Transparent receive SSR (Asset)",
```
FrancescoC-unity, line 153: Transparent**s** here and below

JordanL8 (approved):

FrancescoC-unity (approved):

sebastienlagarde:
>the option on the material always say only Receive SSR (but set the right flag depending whether is transparent or not)
I think it is a bad idea and it is currently something I have change in the past. 
You can't have the same label here as it confused users when changing from opaque to transparent surface type. Because then this value change and users will not know why due to the fact it have same name, need to fix in another PR

PR #1645, opened 2020-08-19
[Shader Graph] Fix issue with enum keyword display names

elizabeth-legros:
Having enum keyword display names that contained [! @ $ % ^ & * - = / ? \  < > ; : ' " [ ] { } ` ~] would error. Changed our name sanitation to take in a custom regex pattern match, and made enum display names pattern match more strictly
Could not reproduce bug after fix
https://fogbugz.unity3d.com/f/cases/1270168/

PR #2352, opened 2020-10-23
[10.x.x][Shader Graph] Fix issue with enum keyword display names

alindmanUnity:
Backport #1645 to 10.x branch 

Manually tested repro steps from bug on compatible versions, confirmed backport still fixes issues as intended


ghost (changes requested):
What happened to allowing () and spaces?

ghost (approved):
Works good now, tested !@#$%^&*()_+{}|:"<>?-=[]\;',./ as well as space and only the (), #, and . and space worked, which don't cause a problem.

marctem (approved):

PR #2353, opened 2020-10-23
[HDRP] Physically-based DoF speed and robustness improvements + bugfixes 

pmavridis:
There are some known issues:
- With Render Graph the first frame flashes black. I already fixed a few DoF related things in this PR, but apparently there are still other issues not related to DoF. 
- I can always nit-pick good examples, but still there are edge-cases with artifacts that I'm fighting to fix.

### Purpose of this PR
This PR improves the robustness and speed of the Physically-based DoF method. In particular:

- Users can now get "more" defocus blur than before, especially when using longer focal lengths with low f-number in the physical camera. In such cases, the difference when turning on and off the physically-based option can be huge, as you see on the image:
image
(Left: physically-based OFF, Right: physically-based ON) 

- Quality is now reasonable even with low sample counts (we are now using mip maps). This gives a big performance boost for a given quality. The example above was produced with the sample slider set to 4 (which was giving 100fps at 1080p for that scene). 

- The Physically-based DoF now exports proper alpha.

- Near blur has been improved (there was a problem with the blending that was creating discontinuities) 

---
### Testing status
Tested on various camera views in the new HDRP template and AmalienborgPalace.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -43,6 +43,7 @@ The version number for this package has increased due to a version update of a r
 - Fixed issue where Default Volume Profile Asset change in project settings was not added to the undo stack (case 1285268).
 - Fixed undo after enabling compositor.
 - Fixed the ray tracing shadow UI being displayed while it shouldn't (case 1286391).
+- Fixed issues with physically-based DoF, improved speed and robustness 
 
 ### Changed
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

TomasKiniulis:
Running into two issues on 10.2:
I seem to get sharp flickering edges even with TAA enabled when Scene and Game view is split
image

Some noticeable tiling artifacts occur when entering fullscreen or exiting out
image

Mentioned these in 10.2 smoke test doc for now.

PR #2363, opened 2020-10-25
Fix ambiguous call to DrawHeaderToggle in Core

sebastienlagarde:
An ambiguous call appear in Core that need to be fixed to fix URP Yamato

### DIFF
--- a/com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
+++ b/com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
@@ -458,10 +458,9 @@ namespace UnityEditor.Rendering
         /// <param name="contextAction">The context action</param>
         /// <param name="hasMoreOptions">Delegate saying if we have MoreOptions</param>
         /// <param name="toggleMoreOptions">Callback called when the MoreOptions is toggled</param>
-        /// <param name="documentationURL">Documentation URL</param>
         /// <returns>return the state of the foldout header</returns>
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -458,10 +458,9 @@ public static bool DrawHeaderToggle(string title, SerializedProperty group, Seri
         /// <param name="contextAction">The context action</param>
         /// <param name="hasMoreOptions">Delegate saying if we have MoreOptions</param>
         /// <param name="toggleMoreOptions">Callback called when the MoreOptions is toggled</param>
-        /// <param name="documentationURL">Documentation URL</param>
         /// <returns>return the state of the foldout header</returns>
-        public static bool DrawHeaderToggle(string title, SerializedProperty group, SerializedProperty activeField, Action<Vector2> contextAction = null, Func<bool> hasMoreOptions = null, Action toggleMoreOptions = null, string documentationURL = null)
-            => DrawHeaderToggle(EditorGUIUtility.TrTextContent(title), group, activeField, contextAction, hasMoreOptions, toggleMoreOptions, documentationURL);
+        public static bool DrawHeaderToggle(GUIContent title, SerializedProperty group, SerializedProperty activeField, Action<Vector2> contextAction = null, Func<bool> hasMoreOptions = null, Action toggleMoreOptions = null)
+         => DrawHeaderToggle(title, group, activeField, contextAction, hasMoreOptions, toggleMoreOptions, null);
```
johnpars, line 463: So if I understand correct, this change of the function signature introduced a hard crash for volume component and it has since been patched here: #2367

PR #2377, opened 2020-10-26
[11.x.x] Early camera color target assignment

pbbastian:
### Purpose of this PR
This PR fixes a regression causing passthrough camera to not render. This is because the render target is fetched during AddRenderPasses, but the wrong target is setup during that call due to recent changes.

In the long run, we do not want the camera color target to be accessed outside of the ScriptableRenderPass callbacks, but to avoid breaking code we're allowing it for now. 

Currently, if there are any renderer features we create a camera color target no matter what. That means that we can assign the camera color target early in case it is needed during AddRenderPasses. 

---
### Testing status
Manually verified using the [3DCharacterUI](https://github.com/Unity-Technologies/UniversalRenderingExamples/tree/236090b6a9501ef2be2f35531bcfa519783e376a/Assets/_CompletedDemos/3DCharacterUI) from [UniversalRenderingExamples#236090b6a9501ef2be2f35531bcfa519783e376a](https://github.com/Unity-Technologies/UniversalRenderingExamples/tree/236090b6a9501ef2be2f35531bcfa519783e376a)
1. Open the 3DCharacterUI scene
2. Enter play mode
3. Click on one of the characters in the scene to spawn the character UI
4. Notice that the background is correctly blurred with the fix


com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -307,6 +326,7 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             // But if we only require it for post processing or the scene camera then we do it after rendering transparent objects
             m_CopyDepthPass.renderPassEvent = (!requiresDepthTexture && (applyPostProcessing || isSceneViewCamera)) ? RenderPassEvent.AfterRenderingTransparents : RenderPassEvent.AfterRenderingOpaques;
             bool createColorTexture = RequiresIntermediateColorTexture(ref cameraData);
```
phi-lira, line 328: Move the line `createColorTexture = (rendererFeatures.Count != 0);` to above code and then this one:

```suggestion
            createColorTexture |= RequiresIntermediateColorTexture(ref cameraData);
```

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Version Updated
```
phi-lira, line 9: This needs to be removed now.

pbbastian: Why do we even add that? Seems like a recipe for merge conflicts :/

phi-lira: It's added because we release packages in lockstep, and previously it happened that we released one package and the other package contained no changes at all, which was confusing to users. (we make a quick patch for URP or HDRP but not changes to Core).

 This will be solved with changelog tooling.

phi-lira (approved):

phi-lira (approved):

PR #2383, opened 2020-10-26
[11.x.x] Removed tile shaders are currently unused.

kaychang-unity:
---
### Purpose of this PR
Removed deferred tile related shaders that do not compile on some older GPUs (missing fallback shaders).

---
### Testing status
Loaded UniversalGraphicsTest from clean project, built Windows player with deferred renderer.

---
### Comments to reviewers
This PR is needed as a fix for SRP package 10.1.


kaychang-unity:
Backport for 10.x.x: https://github.com/Unity-Technologies/Graphics/pull/2385

phi-lira (approved):

phi-lira:
Adding gfx-foundation QA. Check that when we upgrade a previous asset with these shaders we don't have any error.

PR #2386, opened 2020-10-26
[11.x.x] Fixing NullReferenceException with Lift Gamma Gain & upgrading URP Package

ellioman:
# Purpose of this PR
Fixes issue [#1283588](https://issuetracker.unity3d.com/issues/argumentnullexception-is-thrown-when-upgrading-urp-package-and-volume-with-lift-gamma-gain-is-focused-in-inspector)

The solution is to simply create the material in OnGUI instead of in the constructor, like is done it other places.

# What has been tested?
Upgraded the project between 8.2.0 and a version containing the changes in this PR.

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fbugfix-1283588

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/Overrides/ShadowsMidtonesHighlightsEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/Overrides/ShadowsMidtonesHighlightsEditor.cs
@@ -16,6 +16,7 @@ namespace UnityEditor.Rendering.Universal
         SerializedDataParameter m_HighlightsStart;
         SerializedDataParameter m_HighlightsEnd;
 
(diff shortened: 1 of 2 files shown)


erikabar (approved):
Verified case 1283588
Also checked all the rest Post Processing Volume overrides, upgrading from 8.2 to this branch in 2021.1 and upgrading from 2020.1 + 8.2 to 2021.1 + this branch

phi-lira (approved):

phi-lira:
Are any of those shaders used at runtime?

martint-unity (approved):
LGTM

phi-lira:
merging to stage branch

PR #2387, opened 2020-10-27
 Axf raytracing support [21.1]

slunity:
---
### Purpose of this PR
Add support for raytracing for AxF materials (includes performance and quality mode, and recursive raytraced rendering for specific objects)

---
### Testing status
Ran tests that contain AxF materials locally, updated those where needed.
Currently running yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/axf-raytracing
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/axf-raytracing/.yamato%252F_abv.yml%2523all_project_ci_trunk/3974812/job

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/axf-raytracing/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/3981836/job

---
### Comments to reviewers

-Recursive rendering (object-specific), quality raytraced reflection, 1 bounce and multibounce, performance reflection (with gbuffer lit proxy) have all been implemented and tested. 
-Caveat: No path for transparent has been coded though
-Texture filtering has been added, as it is useful when reflections are clear, or for recursive rendering (which itself is useful for high quality reflections while still not tanking performance and enabling scene-wide quality reflections), where the primary ray is also raytraced and not rasterized.

-The UI has 3 parameters added: 
* A filtering amount slider for the aforementioned filtering feature
* A slider to adjust thetaH (half angle - NdotH) for indirect lighting for the carpaint model for the BRDF color table. This is useful since in split sum scenario (including indirect raytraced), this can't be calculated, so it was arbitrarily set to zero. Carpaints have a color shift table that also depend on thetaH. Some reflections have a hue that is a bit off because of that, but this slider allows to adjust that on a per material basis, effectively selecting a pixel column in the brdfcolor table where thetaD will vary.
* A slider to adjust thetaH (half angle - NdotH) for indirect lighting for the carpaint model for flakes. This is useful to adjust independently of the above, because it can control how easy it is to see flakes depending on the material.

Some pictures of the features described above and various results:

0a - axf rasterized primary filtering
0b - axf raytraced primary (recursive rendering) no filtering
0c - axf raytraced primary (recursive rendering) cone filtering
1 axf raytraced reflection color hue thetaH
2a axf carpaint performance mode (gbuffer proxy for 1 bounce raytraced reflections)
2b axf carpaint performance mode (gbuffer proxy for 1 bounce raytraced reflections) B
3a axf carpaint raytraced reflection quality mode
3b axf carpaint quality mode multibounce
4a axf mirror (alu) axf mirroring axf - recursive mode still perf
4b axf mirror (alu) performance raytraced reflection axf reflecting other axf (fixed bug in 2046 a)


com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.hlsl
```diff
@@ -86,89 +129,127 @@
...
+// From Walter 2007 eq. 40
+// Expects incoming pointing AWAY from the surface
+// eta = IOR_above / IOR_below
+// rayIntensity returns 0 in case of total internal reflection
+//
+// Walter et al. formula seems to have a typo in it: the b term below
+// needs to have eta^2 instead of eta.
+// Note also that our sign(c) term here effectively makes the refractive
+// surface dual sided.
+float3  Refract(float3 incoming, float3 normal, float eta, out float rayIntensity)
```
sebastienlagarde, line 174: Lets move this functoin in commonlighting.hlsl or in bsdf?

sebastienlagarde: @eturquin don't we have those function already somewhere?

sebastienlagarde: Will be required in the future anyway by other part of the code

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.hlsl
```diff
@@ -86,89 +129,127 @@
...
-{
-    return bsdfData.geomNormalWS;
+    const float exitBiasIfTIR = NdotVMinCosSpread; // so our exit direction isn't completely grazing
+
+    float k = eta * c - sign(c) * sqrt(saturate(b)) + (noTIR ? 0: exitBiasIfTIR);
+    float3 R = k * normal - eta * incoming;
+
+    float3 criticalDir = (float3)0;
+    incomingSaturated = incoming;
+    if (noTIR == false)
```
sebastienlagarde, line 234: will be better to say (if hasTIR)  :)

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.hlsl
```diff
@@ -184,10 +265,13 @@ float GetPreIntegratedFGDCookTorranceSampleMutiplier()
     // The other term is mostly a tweak to enable a desired match eg VRED
 }
 
-#define DIFFUSE_INDIRECT_FUDGE_FACTOR (1.0)
-#define ENVIRONMENT_LD_FUDGE_FACTOR (1.0)
-#define LTC_L_FUDGE_FACTOR (1.0)
-#define SSR_L_FUDGE_FACTOR (1.0)
+// Safe version preventing NaNs when IOR = 1
+real F_FresnelDieletricSafe(real IOR, real u)
```
sebastienlagarde, line 269: We already have a fresnel like this in bsdf.hlsl

sebastienlagarde (commented):
@slunity . Please this PR should be about adding Raytracing support, but it is full of plenty other change. You must learn to separte change in PR so they can be properly review...

slunity:
> @slunity . Please this PR should be about adding Raytracing support, but it is full of plenty other change. You must learn to separte change in PR so they can be properly review...

They are not real changes, they are moves/re-orders that were required so some already existing functionality could be called in two different contexts in AxFRaytracing.hlsl using SurfaceData and BsdfData.

sebastienlagarde (commented):
@slunity . Please this PR should be about adding Raytracing support, but it is full of plenty other change. You must learn to separte change in PR so they can be properly review...

Outside of few comment and the fact that there is a mix of various stuff, it seems good. 
Just factor some function and moves them. I don't know if @eturquin comment are already address or not?

slunity:
> @slunity . Please this PR should be about adding Raytracing support, but it is full of plenty other change. You must learn to separte change in PR so they can be properly review...
> 
> Outside of few comment and the fact that there is a mix of various stuff, it seems good.
> Just factor some function and moves them. I don't know if @eturquin comment are already address or not?

I'm in the process of doing the update like merging latest master

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Added a new API to bake HDRP probes from C# (case 1276360)
+- Added support for raytracing for AxF material
+- Added area light moment shadows for AxF material
```
sebastienlagarde, line 12: it is not moment shadow but just exponential. Let's just say add area shadow support

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.hlsl
eturquin: We could use HLSL's refract()? (returns null vector if no refraction)

slunity:
Note: I've gone through the reformating process. The yamato format job has failures, although none related to files this PR is touching. I've noticed that locally, the unity-meta format checker script shows I have 106 files needing reformating, so it seems merging master has re-introduced non conforming files. 

I could check that eg in master, SkyManager.cs had indeed such format problems:

image

I don't know why these are not showing up in the yamato job though.

sebastienlagarde (approved):
@slunity We will merge this PR as it is but please do a follow up PR to move/share all the refraction function you have added.

Also not a fan of the new parameter you have added. As it is simply impossible to an artists to understand what it mean. You really need to involve artists when adding new param.  As it is AxF I let it pass as it is suppose to be automatic, but there is no way an artists have any clue about what the property is diong with such a name...

PR #2397, opened 2020-10-27
Fixed broken ray tracing light cluster full screen debug.

anisunity:
The feature was broken after the camera relative PR was merged. This PR fixes the full screen debug mode.

Testing status
Tested that the result is as expected while changing the cluster size and max num lights.

Needs to be tested before the merge.


remi-chapelain (approved):
Tested with the new template, seems to fix the issue. ✔️

PR #2403, opened 2020-10-27
[ShaderGraph] [Backport 10.x.x] [bugfix 1287904] Fix for loading a ShaderGraph containing HDRP specific stuff w/o HDRP package

cdxntchou:
### Purpose of this PR
Backport of PR:  https://github.com/Unity-Technologies/Graphics/pull/2433

Fix for:
https://fogbugz.unity3d.com/f/cases/1287904/
https://fogbugz.unity3d.com/f/cases/1288475/

(note: the git branch name does not match the bugs it is fixing)

The Diffusion Profile nodes, properties and slots are defined in HDRP package.  When the package is not present, ShaderGraph will load them as "UnknownType" objects.  However, if the code doesn't expect UnknownType objects in certain places, it can crash.  This PR adds specialized "UnknownMaterialSlotType" and "UnknownShaderPropertyType" to handle these cases (nodes and targets are already handled), which act as placeholders that have minimal functionality and can keep the UI display working, with added warnings to notify the user that these objects are unknown.

This PR also fixes a related issue with UniversalTarget, where the subtargets are not parented to their Target properly when UnknownTargets are present.  This is because the parenting code runs in Setup(), which is part of the GraphView UI initialization, but some code is expecting it to be parented properly before that is run.  I added the parenting code to OnAfterMultiDeserialize() as well, so it happens just after loading.  There should be no issue with running the code twice.

Also fixed is a bug where Unknown nodes act weirdly and can make your graph unloadable if copied or duplicated.  This PR addresses the major issues by disallowing copying of Unknown nodes and properties.

Also removes MaterialSlot.priority, which was a serialized value, but is not used by any code anywhere..

---
### Testing status

Automated Testing:
Added some Editor tests to check that this HDRP specific nodes and properties will continue to load in URP.
Verified that the tests failed prior to this PR, and now succeed.

Manual Testing:
Created several ShaderGraphs in 10.1, using both HDRP and URP Targets, and HDRP-specific nodes and properties (DiffusionProfile property, DiffusionProfile node, HDSceneColor node, Emission node, Exposure node).

Tested opening those graphs in a project using this PR with HDRP only, and URP only.  Previous to this PR the files would fail to load, but they now load with no errors, and display warnings on the appropriate nodes and properties:

image

image

Tested the generated shader, it is still operational, but falls back to default values for unknown node connections.

Tested that saving the file resulted in no changes to the file on disk, other than the removal of the MaterialSlot priority (an unused value that I stripped out).  All of the unknown data was properly preserved and could be loaded back in the original URP and HDRP 10.1 project.

Tested copy/pasting unknown nodes and properties within same graph, between graphs, and between different instances of Unity running (one without HDRP one with HDRP).  Also tested duplicating (Ctrl+D) those nodes and properties.

Tested copy/pasting combinations of known and unknown nodes -- known nodes are pasted, unknown nodes (and any edges to them) are not pasted.

Tested undo/redo deleting of unknown nodes and properties.

Tested assembly reload of graphs with unknown nodes and properties.

Known existing bug:   Unknown nodes cannot be repositioned properly -- any time they are deserialized they will revert to their original position.  This cannot really be fixed without separating the UI position data out of the serialized node data (something we want to do with the new GTF node changes).

ABV: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/10.x.x%252Fsg%252Ffix%252F1284036/.yamato%252F_abv.yml%2523all_project_ci_2020.2/4006384/job/pipeline

10.x.x ABV:  https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/10.x.x%252Frelease/.yamato%252F_abv.yml%2523all_project_ci_2020.2/4005440/job/pipeline

Build Universal on Win_DX11_mono_apiNet2_Linear_Player on version 2020.2 -- same failure on both
HDRP on Win_DX11_playmode_NonRenderGraph_mono_apiNet2_Linear on version 2020.2 -- same failure on both
HDRP on Win_Vulkan_playmode_XR_mono_apiNet2_Linear on version 2020.2 -- same failure on both
HDRP_DXR on Win_DX12_playmode_NonRenderGraph_mono_apiNet2_Linear on version 2020.2 -- same failure on both
Universal on OSX_Metal_playmode_mono_apiNet2_Linear on version 2020.2 -- both disconnected
VFX_URP on Win_DX11_playmode_mono_apiNet2_Linear on version 2020.2 -- succeeded on re-run (non-deterministic)

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Serialization/MultiJsonInternal.cs
+++ b/com.unity.shadergraph/Editor/Serialization/MultiJsonInternal.cs
@@ -294,19 +294,19 @@ namespace UnityEditor.ShaderGraph.Serialization
             class SerializerHelper
             {
                 [SerializeField]
-                public string m_DisplayName;
+                public string m_DisplayName = null;
 
                 [SerializeField]
                 public SlotType m_SlotType = SlotType.Input;
 
                 [SerializeField]
-                public bool m_Hidden;
+                public bool m_Hidden = false;
 
                 [SerializeField]
-                public string m_ShaderOutputName;
+                public string m_ShaderOutputName = null;
 
                 [SerializeField]
-                public ShaderStageCapability m_StageCapability;
+                public ShaderStageCapability m_StageCapability = ShaderStageCapability.All;
             }
 
             public string jsonData;


elizabeth-legros (approved):

LandonTownsendUnity (approved):
Things I tried:
1. Diffusion Profile property, and in graph, and connected to graph, and inline connected to graph, and emission node connected to graph (still opens)
2. Attempting to copy and paste emission node by doing it while another allowed node is selected (doesn't work)
3. Attempting to copy and paste emission node by putting it in a group and copying the group (doesn't work)
4. Attempting to convert the emission node to subgraph (didn't work)
5. Converting the diffusion profile node to inline in attempt to create a bad state (didn't work)
6. Converting the diffusion profile inline node to property (option is greyed out)
7. Duplicating or copy pasting diffusion profile property (doesn't work)
8. After doing this, adding HDRP back to the project and opening the shadergraph to see if there are any problems (none that I could see)
9. Copying an emission node with HDRP in project, removing HDRP from project, and then pasting it (doesn't work)

Feels pretty foolproof to me. Approving.

marctem (approved):
Looks good to me, thanks for getting this sorted so quickly!

pbbastian (approved):

PR #2407, opened 2020-10-27
[Hdrp] Fix issue with linking FoV to physical camera

pmavridis:
### Purpose of this PR
There is an issue with linking the physical camera FoV with the focal length, discussed in this internal slack thread:
https://unity.slack.com/archives/GHD5LADU7/p1603794331473300
(see also the video from Pierre regarding the issue)

### Testing status
- when changing the FoV the  focal length of the physical camera does not change any more, unless the "link fov to phyiscal camera" option is selected.


sebastienlagarde:
Merging for 10.2 package

PR #2410, opened 2020-10-28
Fix SSR warning & Transparent discrepancy with Non-RenderGraph

skhiat:
### Purpose of this PR
* Fix Vulkan/Metal warning about 'weight' potentially not initialized on ScreenSpaceReflection.compute
* Fix SSR_Transparent test when we use Non-RenderGraph Intensity on the right sphere (fixed):
image

---
Test for review:
Recompile ScreenSpaceReflection.com (after removing the cache), on Metal & LinuxVulkan.
Launch 2701_TransparentSSR with Non-RenderGraph

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Override-Screen-Space-Reflection.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Override-Screen-Space-Reflection.md
@@ -27,7 +27,7 @@ HDRP uses the [Volume](Volumes.md) framework to calculate SSR, so to enable and 
 | **Property**                  | **Description**                                              |
 | ----------------------------- | ------------------------------------------------------------ |
 | **Enable**                    | Indicates whether HDRP processes SSR for Cameras in the influence of this effect's Volume . |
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

JMargevics (changes requested):
Please fix the description: 

- Remove default text
- Add what PR is about (even if the title is descriptive enough)
- Add what's tested and if needed a section "For reviewers" with any specifics

Also, please don't forget about changelog. 

Many thanks!

sebastienlagarde (approved):

sebastienlagarde:
For QA: I need to merge this PR as it fix Yamato. Our Yamato is currently red for non render graph and we need to fix it. Please do a follow up with Soufiane about it.

JMargevics (commented):
Okay tried deleting cache and local library folder, disabling Render Graph from Debug menu and opening 2701_TransparentSSR scene on both Win Vulkan and Metal. In both cases no warnings or other visual regressions.

PR #2011, opened 2020-09-25
[11.x.x] Updating reference images for scene 142 SSAO DepthNormal

ellioman:
# Purpose of this PR
Recently when we merged SpeedTree GI, some of the reference images were not updated. This PR fixes that.

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fupdate-incorrect-reference-images

PR #2416, opened 2020-10-28
[10.2.0] Backporting SSAO reference images 

ellioman:
# Purpose of this PR
Backporting #2011 

# Yamato


phi-lira (approved):

phi-lira:
Running just a single test as this was already tested in against trunk and the SRP package is the same.

PR #2419, opened 2020-10-28
[11.x.x] Fixed baked lit shader compiler error on hybrid renderer.

phi-lira:
### Purpose of this PR
 Fix shader compiler error in Baked Lit shader when using Hybrid Renderer.


ellioman (approved):

cinight (approved):
verified bug fix. all good!

PR #2422, opened 2020-10-28
Re-enabled async compute for contact shadows

JulienIgnace-Unity:
### Purpose of this PR
Re-enabled async compute for contact shadows

---
### Testing status
Tested by building a player with DX12 and checking that passes were indeed async with render doc.


JMargevics (approved):
Tests done by Julien cover enough, and no additional testing from QA is needed. Autoapproved ✅👍

sebastienlagarde (approved):

PR #2425, opened 2020-10-28
Smoother decal angle fade

adrien-de-tocqueville:
### Purpose of this PR

1. Change the two angle fade sliders to a single min max slider.
image

2. Fix https://fogbugz.unity3d.com/f/cases/1288165
Double sided was not taken into account by the depth prepass

3. Change the decal projector angle fading equation to look smoother with wide ranges starting a low values
It uses actually less instructions than before

First images are 0° - 180° range, second is 0° - 90°
For start angles above 90° the results are similar
Before            |  After
:-------------------------:|:-------------------------:
image |  image
image | image

---
### Testing status

Tested the repro package for the fogbugz
Tested various angles on a sphere for the angle fading

### DIFF
--- a/com.unity.render-pipelines.core/Editor/LookDev/Compositor.cs
+++ b/com.unity.render-pipelines.core/Editor/LookDev/Compositor.cs
@@ -130,7 +130,6 @@ namespace UnityEditor.Rendering.LookDev
             }
         }
 
-        IDataProvider m_DataProvider;
         IViewDisplayer m_Displayer;
         Context m_Contexts;
         RenderTextureCache m_RenderTextures = new RenderTextureCache();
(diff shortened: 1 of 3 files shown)


sebastienlagarde (approved):

sebastienlagarde:
Something is incorrect in the PR, there is part of the decal that stay show: 

image

adrien-de-tocqueville:
It was like that on the previous screenshot i didn't modify the test. It is the projector of the bottom row that is too high on y axis and affects the meshes of the top row, but it's not a problem

sebastienlagarde:
> It was like that on the previous screenshot i didn't modify the test. It is the projector of the bottom row that is too high on y axis and affects the meshes of the top row, but it's not a problem

good to know! :), I should have made a better test

sebastienlagarde:
@RemyUnity checked that the slider and double sided worked as expected on my side, will merge for 10.2 package

PR #400, opened 2020-05-10
[11.x] URP Shadergraph Fog Node should return fog intensity as density output

Looooong:
**Guide** : https://github.com/Unity-Technologies/Graphics/blob/pr-template-correction/.github/pr-read.png.md

**Display Addon** : https://userstyles.org/styles/182991/unity-graphics-pr-readme

# Purpose of this PR

> Why is this PR needed, what hard problem is it solving/fixing?

In the URP, the Fog Node in Shadergraph outputs `density` using the result of `ComputeFogFactor(z)`. However, the fog factor cannot be used mix fog color with fragment color. Instead, it must be passed to `ComputeFogIntensity(fogFactor)` to compute fog intensity, then this value is used to mix fog color with fragment color (this is how `MixFogColor(fragColor, fogColor, fogFactor)` is implemented). Therefore, instead of using fog factor, I suppose that Fog Node in Shadergraph should return fog intensity as `density` output. This will allow developer/designer to mix fog color using Lerp Node.

# Testing status

## Links
**Yamato**: (Select your branch) https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics
> Any test projects or documents to go with this to help reviewers?

# Comments to reviewers
> Notes for the reviewers you have assigned.

PR #2428, opened 2020-10-28
Universal/fragment fog

eh-unity:
### Checklist for PR maker

---
### Purpose of this PR
* Fix bad looking fog with large triangles by moving fog computation to fragment shader. 
* It's now based on viewspace instead of clipspace. Originally to make it more uniform and not require remapping, also view space pos was already available in some shaders.
But afterwards the clip space scaling was fixed to be similar on all platforms. 
* This just forces the fragment fog enabled which avoids variants, but it's easy enough to turn into option if necessary.

Tested some performance on my Mac by rendering 4k fullscreen foggy cube.
Virtually no difference in speed for Lit or Unlit shader.
For Lit couple microseconds diff in 1ms of work, vertex shader 10 less instructions, pixel shader 7 more instructions.
For Unlit (3x faster than Lit btw.) the diff was low tens of microseconds, but the fragment version was repeatedly faster. That's < 5% diff though and the noise seems to be tens of microseconds too, so within noise margin (fs fog was consistently a bit faster though).

---
### Testing status
Added a test case for fog.
Ran graphics tests locally.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Ffragment-fog

---
### Comments to reviewers
This PR contains the fix in this PR https://github.com/Unity-Technologies/Graphics/pull/400 for shadergraph.

### DIFF
--- a/TestProjects/UniversalGraphicsTest_Foundation/Assets/Test/TestFilters/TestCaseFilters.asset
+++ b/TestProjects/UniversalGraphicsTest_Foundation/Assets/Test/TestFilters/TestCaseFilters.asset
@@ -11,7 +11,7 @@ MonoBehaviour:
   m_EditorHideFlags: 0
   m_Script: {fileID: 11500000, guid: 5f6aa9f32113aec4a8bded44c1febe5c, type: 3}
   m_Name: TestCaseFilters
-  m_EditorClassIdentifier:
+  m_EditorClassIdentifier: 
   filters:
   - FilteredScene: {fileID: 0}
     FilteredScenes:
@@ -19,7 +19,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 19
     GraphicsDevice: 21
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1305591/
   - FilteredScene: {fileID: 0}
@@ -37,7 +37,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 19
     GraphicsDevice: 21
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1305591/
   - FilteredScene: {fileID: 0}
@@ -82,7 +82,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 31
     GraphicsDevice: 13
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1293775/
   - FilteredScene: {fileID: 0}
@@ -100,7 +100,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 31
     GraphicsDevice: 13
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1293775/
   - FilteredScene: {fileID: 0}
@@ -109,7 +109,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 31
     GraphicsDevice: 13
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1293775/
   - FilteredScene: {fileID: 0}
@@ -118,7 +118,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 31
     GraphicsDevice: 13
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1293775/
   - FilteredScene: {fileID: 0}
@@ -127,7 +127,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 31
     GraphicsDevice: 13
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1293775/
   - FilteredScene: {fileID: 0}
@@ -145,7 +145,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 31
     GraphicsDevice: 13
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1293775/
   - FilteredScene: {fileID: 0}
@@ -163,7 +163,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 31
     GraphicsDevice: 13
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1293775/
   - FilteredScene: {fileID: 0}
@@ -190,7 +190,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 31
     GraphicsDevice: 13
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1293775/
   - FilteredScene: {fileID: 0}
@@ -226,7 +226,7 @@ MonoBehaviour:
     ColorSpace: -1
     BuildPlatform: 24
     GraphicsDevice: 21
-    XrSdk:
+    XrSdk: 
     StereoModes: 0
     Reason: https://fogbugz.unity3d.com/f/cases/1300373
   - FilteredScene: {fileID: 0}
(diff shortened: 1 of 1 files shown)


Verasl (changes requested):
Before
image
After
image

Looking good, but I must say we should do the shadergraph targets here(can notice the road has shadergraph and still doing per vertex fog) at the same time, for example the pass templates for Lit is here and needs converteing/updating https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/PBRForwardPass.hlsl

ellioman:
I'm adding GFX QA as a reviewer as we'll need some performance testing as well once things are approved by other reviewers.

eh-unity:
> Looking good, but I must say we should do the shadergraph targets here(can notice the road has shadergraph and still doing per vertex fog) at the same time, for example the pass templates for Lit is here and needs converteing/updating https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/PBRForwardPass.hlsl

There's a PR to fix  the shadergraph which has broken fog node.
https://github.com/Unity-Technologies/Graphics/pull/400

eh-unity:
> Looking good, but I must say we should do the shadergraph targets here(can notice the road has shadergraph and still doing per vertex fog) at the same time, for example the pass templates for Lit is here and needs converteing/updating 
Fog should be fixed for shadergraph now. There's a bug in Fog node for Unlit, so that won't be fixed until PR https://github.com/Unity-Technologies/Graphics/pull/400 is merged.

eh-unity:
Universal only change,
Timeout and problems with multicamera test output
https://yamato-artifactviewer.prd.cds.internal.unity3d.com/eb045f1c-e75e-4dfc-ae2b-675cbceccade%2FTest_report%2FUtrTestReport1%2FTestProjects%2FUniversalGraphicsTest%2Ftest-results/TestReportV1.html

eh-unity:
Universal CI seems to be green, except a time out in win player build.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -39,20 +39,20 @@
 
 
 #if UNITY_REVERSED_Z
-    #if SHADER_API_OPENGL || SHADER_API_GLES || SHADER_API_GLES3
-        //GL with reversed z => z clip range is [near, -far] -> should remap in theory but dont do it in practice to save some perf (range is close enough)
-        #define UNITY_Z_0_FAR_FROM_CLIPSPACE(coord) max(-(coord), 0)
+    #if SHADER_API_GLCORE || SHADER_API_GLES || SHADER_API_GLES3
```
phi-lira, line 42: This path is dead code isn't it? OpenGL doesn't support reversed Z.

com.unity.render-pipelines.universal/Shaders/BakedLit.shader
```diff
@@ -69,6 +69,9 @@ Shader "Universal Render Pipeline/Baked Lit"
             #include "Packages/com.unity.render-pipelines.universal/Shaders/BakedLitInput.hlsl"
             #include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl"
 
+            // Force enable fog fragment shader evaluation
+            #define _FOG_FRAGMENT 1
```
phi-lira, line 73: If _FOG_FRAGMENT is already defined in shader library why do we need to redefine here?

com.unity.render-pipelines.universal/Shaders/Unlit.shader
```diff
@@ -62,9 +65,9 @@ Shader "Universal Render Pipeline/Unlit"
 
             struct Varyings
             {
-                float2 uv        : TEXCOORD0;
-                float fogCoord  : TEXCOORD1;
-                float4 vertex : SV_POSITION;
+                float4 vertex  : SV_POSITION;
+                float2 uv      : TEXCOORD0;
+                float fogCoord : TEXCOORD1;
```
phi-lira, line 70: In this case both fogCoord and uv are floats. Shouldn't we pack them in same interpolator? Or does the compiler do this automagically?

com.unity.render-pipelines.universal/Shaders/Unlit.shader
```diff
@@ -195,9 +217,9 @@ Shader "Universal Render Pipeline/Unlit"
...
```
phi-lira, line 222: Same interpolator question here.

phi-lira (approved):
LGTM. Minor comments.

com.unity.render-pipelines.universal/Shaders/Unlit.shader
nigeljw-unity: These should definitely get packed by the compiler for the load, but I suspect these would end up in separate barycentric ops, as they would be in separate slots as different varyings. There are other compiler optimizations where all bary ops can be optimized out of the frag shader but that one is a bit more convoluted. I don't remember there being a packing opt for the bary ops on adreno, but other platforms may do this implicitly for free as long as they are aligned properly as Erik has done. Since everything is a scalar architecture these days, not sure how much benefit we would see, but I still think it is an interesting idea and worth trying (as it may reduce reg pressure). Definitely worth a look at the generated IR, but this might be an opt you only see in the machine code.

eh-unity:
Universal CI passes, except for the Build win dx11 standalone which seems to have a timeout problem.

com.unity.render-pipelines.universal/Shaders/Unlit.shader
nigeljw-unity: So apparently, on GCN, the reduction in export is where we will see the benefit if we are export bound (and not ALU bound). This will be less true for RDNA, as it is less likely to be export bound, since exports are done in 4 dwords in batches of 4 (color and position have their own separate paths so not included in this on RDNA). Though as I mentioned, we won't see a reduction in ALU ops for the bary interpolations, but exports if pressured will cost more than an ALU op. There may be some implicit packing of exports, but this is less clear (though I will have a solid response on this at some point). While in the unlit shader, this will matter less, but in general, I think Felipe's suggestion is a really good general practice where possible.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
eh-unity: Currently as far as I know. But was there so I left it there, as REVERSE_Z doesn't exactly mean OpenGL (even if that is the case in practice.)
Do you want remove it?

com.unity.render-pipelines.universal/Shaders/BakedLit.shader
eh-unity: We don't if it doesn't include the library (might not be the case, but perhaps it then just should).
Most likely it's just leftovers when I had them per shader. I'll double check it.

com.unity.render-pipelines.universal/Shaders/Unlit.shader
eh-unity: FXC implies that those are merged, uv and fogCoord get the same register numbers. i.e. v1.xy = uv and v1.z = fogCoord.
It looks like at least on my machine those are appear to be packed on lower level too.

eh-unity:
ABV passed, except for some build timeouts for some mobile platforms and dx11.
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Ffragment-fog/.yamato%252F_abv.yml%2523all_project_ci_trunk/4375478/job

I triggered another Universal run because of the define changes, just to make sure.

ellioman:
Removing myself as a reviewer as we already have Felipe, Nigel, Andre and QA.

alindmanUnity:
@eh-unity Can we actually roll the fix for shader graph into this PR? The pr you mentioned was from a customer, which will take much longer to get up to correct state for merging. If we take it into this PR it will be much easier for us to fix and test and merge

eh-unity:
@alindmanUnity The PR https://github.com/Unity-Technologies/Graphics/pull/400 or its equivalent is rolled into this PR. 
It's in the ShaderGraphFunctions.hlsl.

Verasl (approved):
LGTM

nigeljw-unity (approved):
Looks good to me! +1/2

I still need more exposure to URP/SRP/trunk to feel comfortable giving larger changes a proper +1.

erikabar (approved):
tested frame time on Android devices, results are pretty similar to master branch: https://docs.google.com/spreadsheets/d/1BwQxOJXhzk4X6GeSvPqUlbuzKSFYtH_N_oIpxKsvtbg/edit#gid=1958641881&range=A1:D1

com.unity.shadergraph/Documentation~/Fog-Node.md
```diff
@@ -19,7 +19,7 @@ The High Definition Render Pipeline does **not** support this Node.
 |:------------ |:-------------|:-----|:---|:---|
 | Position      | Output | Vector 3 | Position (object space) | Mesh vertex/fragment's position |
 | Color      | Output | Vector 4 | None | Fog color |
-| Density       | Output | Float    | None | Fog density at the vertex or fragment's clip space depth |
+| Density       | Output | Float    | None | Fog density based on depth. 0 to 1, where 0 is no fog and 1 is full fog. |
```
elizabeth-legros, line 22: Small wording change request - "Returns a value between 0 and 1, where 0 is no fog and 1 is full fog"

hdb-unity (approved):
Checked that changes are made to new test projects (_Terrain and _Foundation).
Also, discussed latest failed CI tests to make sure they will be addressed - please don't merge till they are fixed.

Approving that changes are made to correct test project(s).
Note: Did not test. Did not run tests locally.

erikabar (approved):
approving this again from gfx-qa-urp side

eh-unity:
CustomRevision (known broken trunk), seems to be all green.

PR #2424, opened 2020-10-28
Hdrp/docs/10.2 shaders and refactor

JordanL8:
### Purpose of this PR
Adds new shader docs, eye utility shader graph node docs, and refactors some shader sections to use snippets.

PR #2413, opened 2020-10-28
Changed the names of some of the parameters for the Eye Utils SG Nodes.

anisunity:
While doing the documentation, i realized that some nodes had some attributes wrongly names. This PR corrects those mistakes.
I also pushed the updated sample shader graphs.

Testing status
Ran the DXR tests locally, all green.
RAn the eye SG test locally, green.

PR #2431, opened 2020-10-28
Add Eye Graph Preset to HDRP Shader List

johnpars:
### Purpose of this PR
This PR copies the Eye ShaderGraph preset in the HDRP Test Projects, and places it into HDRP Resources.
This makes it available to users as a quick starting point for creating eye materials.
Note that we have accompanying documentation for this on the way, in #2424

image

image

---
### Testing status
Locally, I have created a new material, switched the shader to **HDRP/Eye** in the shader list, and quickly authored an eye to test for parity with eye materials in the original Graphics Tests.

---
### Comments to reviewers
- Currently, there is no Eye sample in the HDRP Material Sample. Do we want one for 10.2? 
- Would it make sense to update the graphics tests to point to this graph in HDRP Resources?
- Slightly depends on a small change in #2413, which renames some util nodes inputs. I have used the eye graph from that PR branch, so that when this merges it will update automatically (tested locally).

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipelineResources/ShaderGraph/Eye.shadergraph
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipelineResources/ShaderGraph/Eye.shadergraph
@@ -1,0 +1,8681 @@
+{
+    "m_SGVersion": 2,
+    "m_Type": "UnityEditor.ShaderGraph.GraphData",
+    "m_ObjectId": "60ca904df98c4c1f9f421c2faee403b1",
+    "m_Properties": [
+        {
+            "m_Id": "e90c81090c004570bf74081a26ad8d32"
+        },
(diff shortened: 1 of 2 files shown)


remi-chapelain (approved):
Approving the PR since the graph is known and used in some DXR and HDRP Test ! ✔️ 
**YES**, I think it would be a great addition to our material sample to have this and/or multiple version of this as I already stated when the HDRP test have been made ! :) 
The same way we have our material sample for hair / decals / fabric, we certainly need one with Eye especially how hard it is to setup it and have a proper result ! And all the sample in the HDRP/DXR test looks great and can be integrated a wide range of use. 

As for linking the HDRP test with this one, My two cents, is I don't think it's a good idea since any slight modification on this graph would make the test fail and need to rebuild the capture. I think the test graph and the sample graph should remain different.

PR #2230, opened 2020-10-15
Mesh Baking

venkify:
# **Please read;**
## **PR Workflow for the Graphics repository:**
* **All PRs must be opened as draft initially**
* Reviewers can be added while the PR is still in draft
* The PR can be marked as “Ready for Review” once the reviewers have confirmed that **no more changes are needed**
* Tests will start automatically after the PR is marked as “Ready for Review”
* **Do not use [skip ci]** - this can break some of our tooling
* Read the [Graphics repository & Yamato FAQ](http://go/graphics-yamato-faq).

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR

What?

Cache the Light geometry data and save it in Editor time. This is optional and can be enabled/disabled. Users may not need this cache if they are going to Animate ShapeLights

Why? 

Geometry for Shape Lights (Freeform/Parametric) are generated in Editor time as well as Runtime. This can cost CPU cycles during loadtime on the Player if there are a lot of Lights in a scene as the generation happens on Awake. This can simply be prevented by saving Geometry Data in the Editor time and use it in the Player.

---
### Testing status

Added relevant tests in Tests\Runtime\Light2DTests.cs

---
### Comments to reviewers

This PR also fixes a bug where any Path changes in Freeform Light will not regenerate the Mesh.

PR #1721, opened 2020-08-31
Fix for lighting on empty background

unity-cchu:
### Checklist for PR maker
- [x] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.

---
### Purpose of this PR
This is a fix for an issue where transparency with normal maps on an empty background resulted in incorrect lighting values.

---
### Testing status

**Manual Tests**: What did you do?
- [X] Opened test project + Run graphic tests locally
- [X] Checked against Lost Crypt project

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics

PR #1707, opened 2020-08-31
[2d] Smooth Falloff for 2D Shape Lights.

venkify:
Please note :
  . All concave/convex polygons handled properly for both Shape and Offsets.

Future Todo:
  . Handle Offset overlaps and 

# **Please read;**
## **PR Workflow for the Graphics repository:**
* **All PRs must be opened as draft initially**
* Reviewers can be added while the PR is still in draft
* The PR can be marked as “Ready for Review” once the reviewers have confirmed that **no more changes are needed**
* Tests will start automatically after the PR is marked as “Ready for Review”
* **Do not use [skip ci]** - this can break some of our tooling
* Read the [Graphics repository & Yamato FAQ](http://go/graphics-yamato-faq).

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [ ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ]  Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Why is this PR needed, what hard problem is it solving/fixing?

---
### Testing status

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in developer mode, you have a button at end of resources that check the paths)
- Other: 

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics

Any test projects to go with this to help reviewers?

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #2434, opened 2020-10-29
[URP 2D] 2D staging to master

rainsing:
### Purpose of this PR
This PR is an aggregation of several PRs affecting the 2D Renderer:
#1798 #2435 #1990 #2230 #2520 #2495 #1707 #2487 #2524 #1808 #1721 #2568 #2599 #2593 #2630 #2632

---
### Testing status
Each PR is individually tested manually by QA.

---
### Comments to reviewers
Documentation changes will land as a separate PR.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
+++ b/com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
@@ -482,7 +482,7 @@ namespace UnityEngine.Experimental.Rendering.Universal
                 else
                     cmd.SetRenderTarget(pass.rendererData.normalsRenderTarget.Identifier(), RenderBufferLoadAction.DontCare, RenderBufferStoreAction.Store);
 
-                cmd.ClearRenderTarget(false, true, k_NormalClearColor);
+                cmd.ClearRenderTarget(true, true, k_NormalClearColor);
 
                 context.ExecuteCommandBuffer(cmd);
                 cmd.Clear();


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -279,6 +279,9 @@ internal static class ShaderPropertyId
         public static readonly int sourceTex = Shader.PropertyToID("_SourceTex");
         public static readonly int scaleBias = Shader.PropertyToID("_ScaleBias");
         public static readonly int scaleBiasRt = Shader.PropertyToID("_ScaleBiasRt");
+
+        // Required for 2D Unlit Shadergraph master node as it doesn't currently support hidden properties.
+        public static readonly int rendererColor = Shader.PropertyToID("_RendererColor");
```
phi-lira, line 284: Minor: Maybe this name is too generic for a global constant, might cause conflicts with custom shader properties from users?

phi-lira (approved):
LGTM. Minor comment about shader property name.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
rainsing: This name is used by the engine so we can't really change it here.

unity-cchu (approved):

phi-lira (approved):
LGTM on UniversalRP changes. I didn't look in detail at 2D Renderer files though.

ellioman:
Only HDRP fails on ABV which is unrelated to this PR.

PR #2439, opened 2020-10-29
[Deferred] GBuffer Normals format check

jonuuukas:
# **Please read;**

### Checklist for PR maker

---
### Purpose of this PR
Not all android devices support RGBA8_SNorm, so adding a check for it to fix case 1288042

---
### Testing status
Manually tested on repro project on Google Pixel (which doesn't support it)
---
### Comments to reviewers


kaychang-unity:
Thanks. the problem is we do use the whole [-1; 1] range to store normal vectors, so switching to R8G8B8A8_UNorm will cause graphical artefacts if the shader code is not fixed. If you look at the PackNormal/PackSmoothness/UnpackNormal/UnpackSmoothness functions:
https://github.com/Unity-Technologies/Graphics/blob/d066ddc120aa1d6b9ed46c388623cc2b6b19eca8/com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl#L55
Since we only know at runtime if the given Android platform supports R8G8B8A8_SNorm, you would need to add a shader variant for it.

I did consider always using R8G8B8A8_UNorm in all cases to make things simpler, but R8G8B8A8_SNorm offers 2 advantages:
- reduce artefact with terrain layers: the 5th layer of the terrain will blend gbuffer parameters, and the normals look more correct if the normals are stored in R8G8B8A8_SNorm. You can check this in UniversalGraphicsTests scene 038_Shaders_TerrainBlend and look at how specular is reflected on the terrain in forward vs deferred.
- Slightly cheaper to load/store normals. It does save some time on mobile platforms and Switch.

jonuuukas:
@kaychang-unity - updated the PR as we spoke - forcing the option now for those platforms instead of just switching the format

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -267,6 +267,10 @@ internal enum GBufferHandles
         internal int GBufferShadowMask { get { return UseShadowMask ? GBufferLightingIndex + 1 : -1; } }
         internal int GBufferSliceCount { get { return 4 + (UseRenderPass ? 1 : 0) + (UseShadowMask ? 1 : 0); } }
 
+        internal GraphicsFormat GBufferDefaultNormalsFormat = RenderingUtils.SupportsGraphicsFormat(GraphicsFormat.R8G8B8A8_SNorm, FormatUsage.Render)
```
kaychang-unity, line 270: This line has become unnecessary.

kaychang-unity (approved):

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
jonuuukas: thanks for noticing :) fixed in dffbcd2

ellioman (approved):

ellioman:
The Yamato Failures were all known issues not related to this PR plus the property being changed here does not have any testing.

kaychang-unity:
On a side note, I recently found the same issue with SSAO after it has been modified to support deferred renderer.. We are now using GraphicsFormat.R8G8B8A8_SNorm to store the normal buffer, but it is not supported on some Adreno GPUs. Because SSAO code can run on GLES2, I cannot use accurateGBufferNormal option as bit operatations are not available in the shaders.
Instead, I am using GraphicsFormat.R16G16B16A16_Float as a fallback in that case.

PR #2442, opened 2020-10-29
Removed the possibility to have MSAA (through the frame settings) when ray tracing is active.

anisunity:
https://fogbugz.unity3d.com/f/cases/1288459/
MSAA and ray tracing are incompatible in HDRP for various reasons, and this sometimes leads to crashes. The idea is it make MSAA not possible to enable when ray tracing is enabled on the asset. That is done through the frame setting, forced to false and grayed out.

Testing status
Tried all the combinations MSAA on/Off on the asset, Ray Tracing On/Off on the asset, RAy Tracing on/off on the frame settings. Worked as expected.


remi-chapelain (approved):
Tested "upgrade" where project was already supporting MSAA and repro step from the bug. 
Could not repro ✔️

sebastienlagarde (approved):

PR #2445, opened 2020-10-29
Fix shadows in lookdev when exiting playmode

FrancescoC-unity:
This was a tricky one :) Fix for the issue " Shadow Tint tints entire object" in the google doc. 

So the bug mentioned the full screen being tinted by shadow tint, what it actually is is that the shadows are not rendered in lookdev when exiting playmode. 

The reason is because Lookdev uses SingleShadow debug mode and then composite the result, but for reasons the asset used to set the shadow pass did not match the one where the debug mode was set on, so the pass that should've produced shadows produced instead a full rendering. 

The reason why this happened is because the data provider (i.e. the pipeline) used was cached and when exiting playmode there was a mismatch between the cached one and the one that is current.  The cached one can be removed.

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-shadows-in-lookdev/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk

**What did I test**: entering playmode notice all is fine, exiting playmode and notice all is now fine. Also tried that lookdev is still fine on domain reload and upon a package update process.


sebastienlagarde (approved):

sebastienlagarde:
I removed QA from this one, the change is vert small and localized and corevage by Francesco is enough. However let Remi if he have any remark. Merging the fix for 10.2 package

PR #2451, opened 2020-10-29
Properly reject history when enabling seond denoiser for RTGI.

anisunity:
Part of the 10.2 effort. When enabling the second denoiser, some ghosting occured because of bad history rejection from the previous frames. This PR fixes the issues.

giseconddenoiser3

Testing status
Ran the dxr tests locally, everything is green.
Enabled the denoiser, disabled it, moved re-enabled it, the issues doesn't happen anymore.


remi-chapelain (approved):
LGTM ✔️

sebastienlagarde (approved):

PR #2453, opened 2020-10-29
Fixed the cookie upgrade message

alelievr:
### Purpose of this PR
Fixed the cookie upgrade message

This PR shouldn't affect any behavior, as it only changes an upgrade warning to a debug log and show it every time we do the upgrade.


skhiat:
Good for me

sebastienlagarde (approved):

PR #2457, opened 2020-10-29
Fixed exceptions occuring when selecting mulitple decal projectors without materials assigned (case 1283659).

anisunity:
Testing status
Reproduced the steps described in the fogbugz, the issues doesn't happen anymore after the fix.


sebastienlagarde (approved):

sebastienlagarde:
no need for QA for this one, merging.

PR #2458, opened 2020-10-29
Fix LookDev error message

adrien-de-tocqueville:
### Purpose of this PR

Issue from the 10.2 Smoke test document
When reloading scripts with no scene view open (and nothing that renders a camera), RenderPipelineManager.currentPipeline is null, so lookdev fails. Changed the error message

---
### Testing status

Tested domain reload and changing editor theme with no camera rendering, and tested that removing the SRP still gives the same error message.


sebastienlagarde (approved):

sebastienlagarde:
No need for QA for this one, the fix is simple enough

PR #2463, opened 2020-10-29
[HDRP][Compositor] Fix temporary freeze when selecting AOV output

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1288744/

If you enable AOV output in the compositor in a complex scene, the Editor will freeze for awhile to compile the AOV/debug shaders.

With this PR the AOV shader compilation is now async, letting the user interact with the editor/scene while the shaders are compiling. A warning is also displayed in the compositor during the compilation to inform the user that the output might not be correct until the shaders have been compiled.

Note that we enable the async compile only for the compositor Editor/UI. 
The standard AOV API remains synchronous, to ensure people get always the correct results when they use it, even for the first frame.   

Note: I have also fixed a small issue with the text in the AOV infobox being clipped.

---
### Testing status
- Compositor test is green (DX11, tested locally) 
- In the new HDRP template, enabling the compositor and switching to an AOV output does not block the editor anymore.


sebastienlagarde (approved):

sebastienlagarde:
Merged for 10.2 package. The fix make sense. I let QA to a check afterward on this PR.

remi-chapelain:
Tested in master, seems good to me ✔️

PR #2475, opened 2020-10-30
[HDRP] Fix alpha channel in the stop NaNs and Motion Blur shaders.

pmavridis:
### Purpose of this PR
The Stop NaNs post-process pass was never updated for alpha channel support. If it was enabled, the alpha was destroyed and as a result,  any other pass that was properly supporting alpha (like motion blur) was also showing weird artifacts. 

The motion blur kernel was also missing a normalization term for the slow-path tiles.

---
### Testing status
- Tested on the default HDRP template, set the motion blur intensity to 10 and move the camera in scene view. 
Previously with stop NaNs ON for the scene-view we were getting artifacts (because the input alpha channel in mb was corrupted), now everything looks fine.
- Checked for any shader warnings by deleting the Library folder.
- Motion blur / motion vector tests are green (DX11, tested locally).

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -6,9 +6,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [11.0.0] - 2020-10-21
 
-### Fixed
-- Fixed XR shadows culling
-
 Version Updated
 The version number for this package has increased due to a version update of a related graphics package.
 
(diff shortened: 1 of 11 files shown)


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -75,6 +75,7 @@ The version number for this package has increased due to a version update of a r
 - Fixed temporary Editor freeze when selecting AOV output in graphics compositor (case 1288744).
 - Fixed normal flip with double sided materials.
 - Fix shadow resolution settings level in the light explorer.
+- Fix alpha channel in the stop NaNs shader.
```
pmavridis, line 78: I put it on 10.2, if it is too late let me know.

FrancescoC-unity (approved):

TomasKiniulis:
Testing status appears to be sufficient. No QA pass is needed.

PR #2477, opened 2020-10-30
Fixed undo of duplicate environments in the look dev environment library.

JulienIgnace-Unity:
### Purpose of this PR
Fixed undo of duplicate environments in the look dev environment library.
This was caused by duplication of environments in the library not being recorded properly for undoing.

This PR fixes https://fogbugz.unity3d.com/f/cases/1287257/

---
### Testing status
Tested the trivial repro case from the fogbugz case to validate the fix.
Did not run automated testing as Look Dev is not tested there.


RSlysz (approved):
lgtm

PR #2483, opened 2020-10-30
[11.x] Avoid the system 'AppendLine' during generation and serialization

jessebarker:
The system call uses the local environment newline which can cause
    problems when sharing shadergraph files between Windows and other
    systems.  This makes the newline explicitly "\n".
---
### Purpose of this PR
Fix for:
[https://fogbugz.unity3d.com/f/cases/1286430/](url)

New ShaderGraph shaders created on Windows had CR/LF line endings for certain scopes.  This could cause file difference detection to engage on an unmodified shader file when shared between users on different platforms.  This PR fixes that.

---
### Testing status
I created new shadergraph shaders on a Mac, and Chris could load them on Windows and show no differences after loading.
Chris was also able to create a new shader on Windows with no "\n" (CR) characters.

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Version Updated
 The version number for this package has increased due to a version update of a related graphics package.
 
+## [10.3.0] - 2020-11-03
+
+### Added
+
+### Changed
+
+### Fixed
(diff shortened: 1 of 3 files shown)


cdxntchou (approved):

cdxntchou:
Need to modify the shadergraph changelog

ghost (approved):
I think the developer manual testing is sufficient for this fix.

jessebarker:
I've compared the Yamato results for this PR against the ones for the nearest ancestor to my branch, and the results appear similar, so I don't believe I've introduced any new issues.

alindmanUnity:
yamato status matches master, merging

PR #2485, opened 2020-11-02
[ShaderGraph] Optimized the Camera's direction node.

sienaiwun:
Purpose of this PR:
Optimized the Camera's direction node. Camera direction can be gotten directly from the UNITY_MATRIX_V matrix. Two matrix operations can be saved compared to the previous version.

Testing status:
Tested with the sample URP and HDRP projects. Both work well.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this package are documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [11.0.x] - 2020-11-2
+### Changed
+- Optimized the Camera's direction node.
+
 ## [11.0.0] - 2020-10-21
 
 Version Updated
--- a/com.unity.shadergraph/Editor/Data/Nodes/Input/Scene/CameraNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Input/Scene/CameraNode.cs
@@ -49,7 +49,7 @@ namespace UnityEditor.ShaderGraph
             switch (slotId)
             {
                 case OutputSlot1Id:
-                    return "-1 * mul((float3x3)UNITY_MATRIX_M, transpose(mul(UNITY_MATRIX_I_M, UNITY_MATRIX_I_V)) [2].xyz)";
+                    return "-1 * UNITY_MATRIX_V[2].xyz";
                 case OutputSlot2Id:
                     return "unity_OrthoParams.w";
                 case OutputSlot3Id:


sienaiwun:
I was wondering why the Labeler/label action fails, and how can I resolve it?

cdxntchou (approved):
The code looks correct at first glance.

@LandonTownsendUnity  -- just double check that it is actually returning the same value as previously.

DennisDeRykeUnity:
@thomas-zeng 
Per documentation, the Shader Graph Camera node’s Direction method returns the camera’s forward vector direction.  
1) This would be the same vector for both XR and non-XR cases, wouldn’t it?  
2) For XR, do you suspect situations where the matrix math changed by this PR might result in a different answer than the current code?  

@LandonTownsendUnity 
If you have any existing test shaders which use the Camera node, could you share those with me?  If not then I'll try to cobble one together myself for some sanity tests on Quest (or Quest Link).

LandonTownsendUnity:
Today I'm going to make a test shader to test this PR vs the original algorithm. If there is even a slight change this shader should detect it.

LandonTownsendUnity:
[CameraDirParityCheck.zip](https://github.com/Unity-Technologies/Graphics/files/5519635/CameraDirParityCheck.zip)
This shader should appear as white or another very bright color if there is any disparity between the original equation for the camera direction and the one suggested by this PR. @DennisDeRykeUnity Try using this in XR scenarios in both HDRP and URP to see if it every shows up as anything besides solid black. I confirmed it works normally for both perspective and orthographic in both HDRP and URP but it's worth double checking in XR.

LandonTownsendUnity (approved):
Seems good to me but wait for XR QA. Linked shader in comments that should help with testing.

thomas-zeng:
Hey @DennisDeRykeUnity 
1. This would be the same vector for both XR and non-XR cases, wouldn’t it?
  -- In XR, forward vector is per eye. So they could potentially be different. 
2. For XR, do you suspect situations where the matrix math changed by this PR might result in a different answer than the current code?
  -- We should verify in XR, camera direction node behaves the same before/after this change.

PR changes looks good to me.

thomas-zeng (approved):

phi-lira (approved):

DennisDeRykeUnity:
My apologies for the delays.  I was busy with other tasks and vacation.  I intend to perform XR testing this week.

alindmanUnity:
@DennisDeRykeUnity were you able to preform XR testing on this PR?

DennisDeRykeUnity:
@alindmanUnity No.  I'll test it briefly tomorrow on URP and HDRP with CameraDirParityCheck shader.  I'll test perspective camera only since orthographic is not applicable for XR.

DennisDeRykeUnity:
First I tried testing this with the SRP packages from the PR 2485 dev branch (sienaiwun:sgcameradirection cd4bc9e) but this resulted in compiler errors with several recent versions of Unity.  So I obtained the SRP packages from Graphics master ce3633b and implemented one-line code change from PR 2485 locally in this file: 
com.unity.shadergraph/Editor/Data/Nodes/Input/Scene/CameraNode.cs

Next I started with the default empty 3D scene, added some cubes and capsules, and assigned them all a new material that uses Landon's Camera Direction Parity Check shader.  Using the configurations below, if I do not apply head tracking  (the camera remains stationary) then all the objects are black as expected.  

- Unity 2021.2.0a1 (trunk 31233d44259e)
- URP (Graphics master ce3633b + one-line code change from PR 2485)
- Simple test scene with Landon's CameraDirParityCheck shader applied
- Quest Link
- Multi Pass Stereo and Single Pass Instancing
- Playmode 

If head tracking is enabled, then the objects blink between an assortment of solid colors due to even subtle movements of the camera.  However, without XR enabled I can rotate the camera in the Editor in Play mode and see similar color blinking, so I suspect the shader only returns black if the camera rotation is the default zero values (0, 0, 0).  @LandonTownsendUnity Am I correct that your shader is only expected to return black when no rotation has been applied to the camera?

DennisDeRykeUnity (approved):
Landon realized that the constant float input for his shader's Multiply node was excessive.  Setting it to 10000 aleviates the color blinking, and objects are all black as expected even with XR head tracking enabled using these configurations: 

- Unity 2021.2.0a1 (trunk 31233d44259e)
- URP and HDRP (Graphics master ce3633b + one-line code change from PR 2485)
- Simple test scene with Landon's CameraDirParityCheck shader applied
- Quest Link
- Multi Pass Stereo and Single Pass Instancing
- Standalone player 


Also I saw no anomalies using these configurations: 

- Unity 2021.2.0a1 (trunk 31233d44259e)
- URP and HDRP (Graphics master ce3633b + one-line code change from PR 2485)
- URP  and HDRP* Template scenes
- Quest Link
- Single Pass Instancing
- Play mode

* For this PR I used the old "workshop" HDRP scene since I am familiar with it. 

I approve this PR on behalf of Unity XR QA.

sienaiwun:
Hi guys, is this issue still open?

esmelusina:
I went ahead and removed Alex from the PR. Please go ahead with this!

PR #2488, opened 2020-11-02
Clamp distortion UV properly so we don't sample outside permitted area

FrancescoC-unity:
As the title says, the issue we saw on VFX is because we were ending up sampling outside the permitted area. 

Even before Anis' change though there was an issue as we could still end up sampling outside the valid RTHandle area with lower mips sampled, so this take care of that issue too and should fix issues seen in the wild previously and some issues in the Runtime test scene. 

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-distortion-issue/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/4028939/job/pipeline 

I will need a screenshot update in the private repo when this is merged.


anisunity (approved):
looks good, thanks for the fix

sebastienlagarde (approved):

PR #2491, opened 2020-11-02
Removed unused blit input

rolandk-unity:
### Purpose of this PR

This PR fixes the blit operation in the`ColorGradingLutPass`, which generates the color grading LUT. Previously, this blit operation used the temporary render target used for the LUT as an input. This input, however, is never read by the shader, and the RT is also uninitialized at this point. This PR removes the unused input from the blit.

---
### Testing status

Tested on iOS, Android GLES 3, Android Vulkan, Windows

---
### Comments to reviewers

Note that some of the Yamato tests failed due to a current issue on trunk. I re-ran all failing tests with a pre-issue trunk changeset.

Note that setting the unused input previously resulted in an "load" load action being used for the LUT generation rendering pass on iOS. After removing the input, the more optimal "don't" care load action is used. So this change does actually save a nice amount of bandwidth every frame on iOS.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Passes/ColorGradingLutPass.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Passes/ColorGradingLutPass.cs
@@ -167,7 +167,7 @@ namespace UnityEngine.Rendering.Universal.Internal
                 renderingData.cameraData.xr.StopSinglePass(cmd);
 
                 // Render the lut
-                Blit(cmd, m_InternalLut.id, m_InternalLut.id, material);
+                cmd.Blit(null, m_InternalLut.id, material);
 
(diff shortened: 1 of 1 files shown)


phi-lira (approved):

phi-lira:
LGTM. Is this the only place we have this Blit pattern?
 Please check why ABV is red and this should be good to go. 

 (Ping me on slack if you need help)

unrk:
> LGTM. Is this the only place we have this Blit pattern?
I found a similar pattern in the screenspace shadow pass. As this feature is currently not on `master`, I reached out to @Seongdae, who is currently working on the screenspace shadows.

> Please check why ABV is red and this should be good to go.
Checking now.

PR #2492, opened 2020-11-02
Renamed files to be compliant with trunk convention.

JulienIgnace-Unity:
### Purpose of this PR
Renamed files to be compliant with trunk convention.
This is needed to get SRP to core unity.

---
### Testing status
Ran local tests and tested debug menu in a player.
Will run the formatting test once it's available in master.


theo-pnv (commented):
Thanks!
I'm approving on behalf of SDETs but leaving the full review to @sebastien-unity, he will bring more value.
There are still some HDRP related files in need of a [reformat in this job](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Frename-files/.yamato%252F_formatting.yml%2523formatting/4063838/job/(log:Execution)). After discussing with Julien, here's what we agreed on:
- Subgraph files: @sebastien-unity will do another PR
- HD Template files: Pierre will handle them in another PR

Note: After the renaming, we will have to reformat the code inside the files :)

sebastienlagarde (approved):

PR #2507, opened 2020-11-03
[VFX] Remove Exact Fixed Time option

PaulDemeulenaere:
### Purpose of this PR
We detected an unexpected behavior with Exact Fixed Time Step option on the VisualEffectGraph : 
_repro_exact_time
See case : https://fogbugz.unity3d.com/f/cases/1289829/

The actual fix require a C++ change, which can't be deliver before 10.2, so, we are temporarily removing this option in VisualEffect inspector.

The inspector will also reset the `VFXUpdateMode.ExactFixedTimeStep` logging the sanitize has been done.

---
### Testing status
**Manual Only**

*Before*
image

*After*
image

---
### Comments to reviewers
This feature was exposed to user in 2020.1/2020.2, so : 
- 8.x.x on 2020.x : reachable through enum
image
It will be update with this PR https://github.com/Unity-Technologies/Graphics/pull/1131
- 10.x.x on 2020.2 : reachable through toggle

### DIFF
--- a/TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXSpawnerTest.cs
+++ b/TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXSpawnerTest.cs
@@ -341,6 +341,7 @@ namespace UnityEditor.VFX.Test
 
             graph.GetResource().updateMode = (VFXUpdateMode)timeMode.vfxUpdateMode;
             AssetDatabase.ImportAsset(AssetDatabase.GetAssetPath(graph));
+            Assert.AreEqual(graph.GetResource().updateMode, (VFXUpdateMode)timeMode.vfxUpdateMode);
 
             var previousCaptureFrameRate = Time.captureFramerate;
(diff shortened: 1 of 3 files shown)


PaulDemeulenaere:
@julienf-unity As discussed on slack, use a sanitize instead of doing this in inspector at https://github.com/Unity-Technologies/Graphics/pull/2507/commits/9ea36db8b0ad21c8c6601a5b83092c3667d3e1a8
 image

Here an execution after loading the repro project ⬇️ 
_repro_exact_time_sanitize

Actually, I thought I had to update editor test but we can still test spawner behavior because the sanitize is only called on the first import (or opening the VisualEffectGraph window), added a explicit check of update mode after reimporting asset.

julienf-unity (approved):

PaulDemeulenaere:
```
Total tests: 96
Failed test count: 1
Successful tests count: 90
Not run tests count: 5
```
Expected result due to the unexpected message `message: Converting invalid MinMaxAABB`

PaulDemeulenaere:
🟢 (kind of), ready to be merged !

PR #2508, opened 2020-11-03
hdr toggle issues in scene view

martint-unity:
---
### Purpose of this PR
Fixing an issue where scene view camera ignored the HDR toggle from the Universal Pipeline Asset.
https://fogbugz.unity3d.com/f/cases/1284369/

fix_hdr_toggle


phi-lira (approved):

erikabar (approved):
Verified issue 1284369, checked all combinations of HDR settings:
- Main Camera Off + URP Asset true 
- Main Camera "Use pipeline settings" + URP Asset true
- Main Camera "Use pipeline settings" + URP Asset false.

All combinations result in Scene View Camera having same HDR settings as the Main Camera.

martint-unity:
Removing the tag to 10, since it is working in 10

PR #2510, opened 2020-11-03
[11.x.x] Fix issue causing materials to upgrade multiple times

pbbastian:
### Purpose of this PR
Fix case 1287931 where materials were getting their _BaseColor upgraded to (1, 1, 1, 1). The reason is that the material upgrader was broken previously, and would soft-crash when it hit a material with a sub-object where the script is missing. In the repro project, this was the HDRP material version. That in turn caused all materials from that point on to not get a URP material version assigned. That was fixed some time ago, and so the upgrader now thinks those materials are at version 0, causing it to apply all upgrades to it, even though it was actually much newer than that. So we have a bunch of materials in the wild with missing version numbers. There is no accurate way of knowing their version.

In this PR, I change the default to be the current project version instead of version 0. This is a guess, but I think it's a very reasonable guess. The only case it does not work in is if you're adding a material to your project that wrongly does not have a version, where the "true" version does not match the project material version. Otherwise, since the material is in your project it will have been used with the current project material version. It does mean that version 0 materials will no longer upgrade correctly, and thus require manual fix-up. Version 0 is from before LWRP went out of preview, and thus we think this is an acceptable loss.

Either way, this is a conservative approach that is making us upgrade less rather than more. Stuff that wasn't upgraded enough can be manually fixed, while upgrading too much could cause user data to be overwritten.

Material files with missing versions can still be manually touched up by copying the version YAML into the file, and setting it to the appropriate version. If it's a major issue for many people, we have the option of automating this from a contextual menu, but for now it is more important to land this fix ASAP.

I also introduce a new material version that doesn't actually do anything property-wise, but forces all materials in the project to get an asset version.

---
### Testing status
Verified the fix locally using the repro steps from fogbugz.


phi-lira (approved):

erikabar (approved):
Verified case 1287931, also tested UniversalRenderingExamples project. No issues found.

phi-lira:
Merging. Backport PR is passing on 2020.2, failures are known issues in master branch.

PR #2514, opened 2020-11-03
[Universal|Bugfix] Camera inspector requests current URP asset

Verasl:
### Purpose of this PR
Currently the Camera inspector grabs the URP asset from graphics settings, this can be overriden by a pipeline asset in quality settings if set, this PR changes the code so that it looks for the current asset as per defined by UniversalRenderPipeline.asset

---
### Testing status
Simple test in Karting template

---
### Comments to reviewers


erikabar (approved):
Fixes the Renderers issue when different URP assets are set in Graphics and Quality settings.

martint-unity (approved):

phi-lira (approved):
LGTM. There seems to be also same GraphicsSettings.renderPipelineAsset in LightEditor.

phi-lira (approved):

Verasl:
> LGTM. There seems to be also same GraphicsSettings.renderPipelineAsset in LightEditor.

Changed it there too, low risk change as it is editor only and will help provide correct information.

Also spotted it being used for the shader debug level, this is also low risk as it is simply for editor console print out when building, but will also stop confusion of where the settign is being used.

phi-lira (approved):

PR #2518, opened 2020-11-04
[HDRP] Fix issues with GenerateMaxZ when opaque object or decal frame settings is disabled

FrancescoC-unity:
The issue was twofold: 

1) The depth texture was relying on it being set from a previous pass which is clearly not reliable
2) The pass was still run even when we had no valid depth buffer (when opaque is disabled) 

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-issue-with-generate-max-z-and-disable-opaque/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/4079344/job


remi-chapelain (approved):
Tested procedurally (with and without every raytraced effects, in perf and quality mode) that no further errors are thrown when enabling / disabling any frame setting field.

sebastienlagarde (approved):

PR #2519, opened 2020-11-04
[Yamato] Use underscores for editor tracks in ymls

kivistikliis:
---
### Purpose of this PR
This PR affects nothing for master, because we use trunk, but it affects other tracks and thus should be merged in.

Replace . with _ in latest_editors_metafile (eg use 2020_2 instead of 2020.2), because if ymls call `{{editor_versions.2020.2_staging.windows.revision}}` then the . in 2020.2 is assumed to refer to a child-property in the dictionary and thus ymls fail to reach the revision property. See 714e2498e6cac0b75efe9951d162c931b8fc5995 for how it affects 2020.2 

---
### Testing status
No testing is needed because nothing is affected for master, and other tracks are tested on their corresponding branches


sophiaaar (approved):

theo-pnv:
Was backported to 7.x.x/release with https://github.com/Unity-Technologies/Graphics/pull/2543

PR #2523, opened 2020-11-04
Fixed exceptions when unchecking "Big Tile Prepass" on the frame settings with render-graph.

anisunity:
Fixing an issue that was detected during the smoketest 10.2. Happening when unchecking the bigtile frame setting on the hdrp template. This PR fixes that issue.

Testing status
Openned the HDRP new template checked and unchecked the frame setting, the crashn happened before, but not after the fix.


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

remi-chapelain (approved):
Fixed ✔️

PR #2525, opened 2020-11-04
Fix button Bake with Reflection Probe in custom mode

sebastienlagarde:
### Purpose of this PR
Fixed an assert of unknow format when clicking on Bake button while having setup custom mode for Reflection Probe.

The issue have been introduce after refactor of texture format for reflection probe PR. It was saving custom texture as 1111110f format instead of fp16.

I have tested in a very simple scene with sphere and sky that bake and custom bake produce the correct result.

---
### Testing status

---
### Comments to reviewers


JulienIgnace-Unity (approved):

sebastienlagarde:
Merging for 10.2, nothing particular here.

PR #2526, opened 2020-11-04
Fixed an issue where Universal Render Pipeline with disabled antiAliasing was overwriting QualitySettings.asset

lukaschod:
# **Please read;**

### Checklist for PR maker

---
### Purpose of this PR
Fixed an issue where Universal Render Pipeline with disabled antiAliasing was overwriting QualitySettings.asset on frequent cases (https://fogbugz.unity3d.com/f/cases/1219159/).

---
### Testing status
- [x] Tested that issue no longer reproduces in case.
- [x] Checked the code if XR needs fixing. Apperently they are handling it in native side 
```
void XRDisplaySubsystem::SetMSAALevel(int level)
{
    level = (level > 0) ? level : 1;
    m_TextureManager.SetMSAALevel(level);
}
```
- [x] Test if QualitySettings.antiAliasing is correctly update with URP.

---
### Comments to reviewers
I think we should discuss a bit more about antiAliasing value inconsistency between builtin and SRP. I have feeling it might bite us in the future again.


phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -144,8 +144,12 @@ public UniversalRenderPipeline(UniversalRenderPipelineAsset asset)
...
 
+            // In QualitySettings.antiAliasing disabled state uses value 0, where in URP 1
+            int qualitySettingsMsaaSampleCount = QualitySettings.antiAliasing > 0 ? QualitySettings.antiAliasing : 1;
+            bool msaaSampleCountNeedsUpdate = qualitySettingsMsaaSampleCount != asset.msaaSampleCount;
+
             // Let engine know we have MSAA on for cases where we support MSAA backbuffer
-            if (QualitySettings.antiAliasing != asset.msaaSampleCount)
+            if (msaaSampleCountNeedsUpdate)
             {
                 QualitySettings.antiAliasing = asset.msaaSampleCount;
```
phi-lira, line 154: don't we need to convert the data when setting antialiasing as well?

phi-lira: @thomas-zeng I see @lukaschod already checked SetMSAALevel but pinging you here for awareness

ellioman (approved):

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
lukaschod: conversion happens inside native side

ellioman:
The failures on Yamato are known and not caused by this PR.

PR #2528, opened 2020-11-04
HDRP - add  Default Frame settings under “Light Loop Debug” tab

sebastienlagarde:
all in title


sebastienlagarde:
No need for more review :)

remi-chapelain:
Yes, tooltips are always better that no tooltip. 
However, it could benefit some field to have a little more information, especially for those obscures, not often used features. 
For example, what does FTPL means ? What does big tile do in a sentence for light visibility ?

For example, I think the tooltip for asynchronous execution is fairly clear even if you don't know what is it beforehand
image

PR #2530, opened 2020-11-04
Refactor XR automated tests with MockHMD

fabien-unity:
### Checklist for PR maker

---
### Purpose of this PR
- refactor XR automated tests in order to rely on MockHMD
- unlock the ability to run existing tests on standalone player builds
- edit Yamato config files

---
### Testing status
Tested locally with UniversalGraphicsTest on Win10 + Google Pixel 1 :
- Playmode + Playmode_XR
- Standalone and Standalone_XR win64 d3d11 mono
- Standalone and Standalone_XR android gles3 mono

Yamato URP run 🟢 with trunk : https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/tests%252Fxr%252Fuse-mockhmd/.yamato%252Fall-universal.yml%2523All_Universal_trunk/4227949/job

---
### Comments to reviewers
The manifest file have been updated to use the latest version of `com.unity.testframework.graphics` (required for this PR to work).

### DIFF
--- a/.yamato/universal-android-opengles3.yml
+++ b/.yamato/universal-android-opengles3.yml
@@ -96,7 +96,7 @@ Universal_Android_OpenGLES3_Standalone_XR_il2cpp_Linear_trunk:
       -  path: .yamato/universal-android-opengles3.yml#Build_Universal_Android_OpenGLES3_Standalone_XR_il2cpp_Linear_trunk
          rerun: on-new-revision
 Build_Universal_Android_OpenGLES3_Standalone_XR_il2cpp_Linear_trunk:
-    name: Build Universal on Android_OpenGLES3_il2cpp_Linear_Player on version trunk
+    name: Build Universal on Android_OpenGLES3_il2cpp_Linear_Standalone_XR_build_Player on version trunk
     agent:
         type: Unity::VM
         image: mobile/android-execution-r19:stable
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.universal/Runtime/XR/XRPass.cs
```diff
@@ -406,6 +406,12 @@ internal void EndCamera(CommandBuffer cmd, CameraData cameraData)
 
         internal void RenderOcclusionMesh(CommandBuffer cmd)
         {
+
+        #if DEVELOPMENT_BUILD || UNITY_EDITOR
+            if (XRGraphicsAutomatedTests.enabled && XRGraphicsAutomatedTests.running)
```
thomas-zeng, line 411: Is there a good reason why we disable occlusion mesh when automation is in use?

fabien-unity: By default, the occlusion mesh from MockHMD will show up because we're using the full viewport 1:1 during tests.
The alternative would be to add a right-eye only mirror view mode without occlusion mesh to MockHMD somehow.

thomas-zeng (approved):

com.unity.render-pipelines.universal/Runtime/XR/XRPass.cs
thomas-zeng: Makes sense. Thanks!

sophiaaar (approved):

fabien-unity:
I'm switching back to draft mode because Yamato is so red at the moment (unrelated to this PR I think but can't confirm). I'll spend more time to extract the test code from URP to be able to share with HDRP.

TestProjects/UniversalGraphicsTest/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -307,4 +307,8 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/Scenes/146_MSAA_RenderToBackbuffer.unity
     guid: aee9f891d65e44daf811d30053402efd
-  m_configObjects: {}
```
phi-lira, line 310: Does this needs to be pushed?

phi-lira (approved):

fabien-unity:
I'm closing this PR in favor of https://github.com/Unity-Technologies/Graphics/pull/2812, which has a cleaner implementation by moving the XR specific test code outside of the universal test project and into an internal package so that it can be reused on the other test projects (HDRP_Tests, etc).

PR #2533, opened 2020-11-05
Fixed exceptions related to the debug display settings when changing the default frame settings.

anisunity:
The 100% repro that i found was opening the thest 2501_LightLayers and changing the default frame settings (like shown in the gif).

shadowExceptions2

Testing status
The exceptions do not happen anymore after the fix.


remi-chapelain (approved):
Had this issue a lot while smoke testing but couldn't find a proper repro. 
Since I don't have a repro, I'm trusting the dev on this one ! ✔️

sebastienlagarde (approved):

PR #2539, opened 2020-11-05
Fixed a case where overlay camera with output texture caused base camera not to render to screen

lukaschod:
### Checklist for PR maker

---
### Purpose of this PR
Fixed a case where overlay camera with output texture caused base camera not to render to screen. [case 1283225](https://issuetracker.unity3d.com/issues/game-view-renders-a-black-view-when-having-an-overlay-camera-which-had-output-texture-assigned-in-the-camera-stack)

---
### Testing status
- [x] Tested the case 1283225.

---
### Comments to reviewers
Personally I am not super happy with the fix, as it somewhat prevent this issue.

However there still code paths that will not be fixed:
- Cameras that already had overlay and output texture.
- If output texture is changed via scripting.

**Now a bit background why it is not trivial to fix in logic side:**

The reason nothing is rendered (black), because rendering actual takes output texture of overlay camera and not base camera. Even though in C# we have logic to take base camera target always, it does not work when we have base camera targeting back buffer. in SRP back buffer uses id -1, which is back buffer in most cases. However if we call SetupCameraProperties in native side it actually treats -1 as camera target and we do call it in URP for EVERY camera.

SetupCameraProperties calls
void ScriptableRenderContext::ExtractCameraProperties(const Camera* camera)
{
    m_ExtractedCameraProperties.m_CameraTargetValid = true;
    m_ExtractedCameraProperties.m_CameraCurrentTarget = camera->GetCurrentTargetTexture();
}

When cmd executed
           case kScriptRenderCommand_ExecuteCommandBuffer:
            {
                RenderingCommandBuffer* commandBuffer = m_CommandBuffers[m_Commands[i].perTypeIndex];
                RenderNodeQueue tempEmptyQueue(kMemTempJobAlloc);
                commandBuffer->ExecuteCommandBuffer(defaultContext, tempEmptyQueue,
                    RenderingCommandBuffer::kFlagsKeepState, &m_tempRTs,
                    kGPUQueueComputeOnGraphics,
                    (m_ExtractedCameraProperties.m_CameraTargetValid ? m_ExtractedCameraProperties.m_CameraCurrentTarget : NULL));
                tempEmptyQueue.SyncDependentJobs();
                break;
            }

During cmd prepare
    // The functionality here completely differs between SRP and the integrated render loops.
    // Besides CurrentActive, every other BuiltinRenderTextureType is managed
    // by the CameraStack, which does not exist in SRP.  Even though the Camera might be configured with its
    // own target texture, we don't have access to the Camera (and its state)
    // from a CommandBuffer.  Intermediate render texture management is currently left
    // up to SRP itself.
    // Currently, SRP is using BuiltinRenderTextureType.CameraTarget as the
    // de-facto 'back-buffer'.  For VR, we need to point CameraTarget to the
    // VR device eye textures.  When VR is active, we bind that texture to the Camera
    // CurrentTargetTexture, which we pass in as the override CameraTarget here.
    if (overrideCamRT != NULL)
        state->m_BuiltinRT[kBuiltinRTCameraTarget] = overrideCamRT;

Where `overrideCamRT `is `m_ExtractedCameraProperties.m_CameraCurrentTarget`

In order to change this logic, we would have to change native side where HDRP also depends. Which can lead to more issues when we had before...


com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -483,12 +483,19 @@ public override void OnInspectorGUI()
...
             {
+                // Settings only relevant to base cameras
                 DrawOutputSettings(rpAsset);
                 DrawStackSettings();
             }
+            else if (camType == CameraRenderType.Overlay)
+            {
+                // ScriptableRenderContext.SetupCameraProperties still depends on camera target texture
+                // In order for overlay camera not to override base camera target texture we null it here
+                if (settings.targetTexture.objectReferenceValue != null)
```
phi-lira, line 496: This is probably ok for upgrade reasons. However we should make sure the change gets applied when the user change from base -> overlay in UI and that we properly handle Undo in that case, restoring the camera target texture.

phi-lira (changes requested):

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
lukaschod: Fixed in latest. Nice catch!

phi-lira (approved):

phi-lira:
ABV was passing on previous changeset. I've fixed merge conflict with changelog only.

PR #2546, opened 2020-11-05
Added new render graph API for shared resources

JulienIgnace-Unity:
### Purpose of this PR
Currently, the only way to share some resources across different render graph executions (and different frames) is to import resources into the render graph. This means that users need to manage the life time of those resources themselves. This can be a complicated process if they have many different shared resources like that that may or may not be used at the same time. This kind of pattern is useful for example when you need to keep static shadow maps from one frame to another, or to keep a depth buffer available when doing camera stacking/composition etc...

To solve that, we introduce a new API:

public TextureHandle CreateSharedTexture(in TextureDesc desc)
public void ReleaseSharedTexture(TextureHandle texture)

Those allow users to create an explicitely shared texture without having to manage its lifetime.
In practice, users create a shared texture once outside of any render graph code and reuse it from one frame to another.
The actual graphics resource will only be allocated the first time the resource is used and will automatically be deallocated if not used anymore. This makes sharing resources much easier.

To achieve this new API some refactoring was needed:
- Refactor internal code to be more shared wrt resource creation/release (there was a lot of duplication)
- Renamed/Moved code to make the separation between RenderGraphResources and actual GraphicsResources cleaner.
- Moved code related to specific resource types (texture, compute buffer, renderer lists) to their own files

This PR also implements this feature for Shadow Map Atlases in HDRP.
Now all cached shadow atlases use this new SharedTexture API. Dynamic atlases remain unchanged.

---
### Testing status
This PR changes some core initialization code of the shadow system. Since static shadows are hard to test automatically, we need to do a thorough pass on them.
What I tested locally:
- All automatic tests pass
- Tested simple repro using shadow spot lights set to OnEnable. Tried moving lights and/or objets to validate the behavior.
- Tested mixed shadow workflow with 2021.1: Static Shadow Caster On/Off with OnEnable shadow light.

Would be good to do a thorough testing pass with QA since I probably missed a lot of use cases (having someone who tested the original static shadow feature would be perfect). Might also want to test without Render Graph.


FrancescoC-unity (approved):

TomasKiniulis (changes requested):
Thanks for the detailed description @JulienIgnace-Unity!

Running into a couple of issues so far:
- Shadows are not displayed correctly anymore when going back after going out of draw distance. Does not occurs only with "Always draw dynamic disabled"
- ~~Disabling Rendergraph breaks rendering in project settings breaks rendering~~
- PR reintroduces the following exception after a some time: 
`Render Graph Execution error at pass Partial Depth Prepass (Deferred - Decal + AlphaTest) (1)UnityEngine.GUIUtility:ProcessEvent (int,intptr,bool&)`

More details on coverage here: https://favro.com/organization/c564ede4ed3337f7b17986b6/8ddd69383a55181bd8da3bda?card=Uni-179889

TomasKiniulis (approved):
Checked editor and player build. Works well now.

- Shadows are displaying correctly when going back in from out of shadow draw distance
- Left my pc running with test project open overnight, and `Render Graph Execution error at pass Partial Depth Prepass` does not occur anymore

sebastienlagarde (approved):

PR #2547, opened 2020-11-05
Fixed an inconsistent behavior between RT Area shadows and Shadow maps when shading a transmission fragment.

anisunity:
https://fogbugz.unity3d.com/f/cases/1289675/
Before this PR, when doing the transmission we were reading the screen space shadows in the case of area lights. This PR changes it to read the shadows maps (and that is coherent with non-ray tracing), but that is still wrong. We need to implement proper transmission for area lights in both cases.
I also noted that the behavior of point and spot rt shadows had an issue at the edge. Something that i need to investigate later.

image

Testing status
The transmission now properly works with a ray traced area shadow (not black anymore, see screenshot).
The DXR tests were ran locally and are green.

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/ScriptableRendererDataEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/ScriptableRendererDataEditor.cs
@@ -47,7 +47,6 @@ namespace UnityEditor.Rendering.Universal
             m_FalseBool = editorObj.FindProperty(nameof(falseBool));
             UpdateEditorList();
         }
-
         private void OnDisable()
         {
             ClearEditorsList();


remi-chapelain (commented):
I understand the issue and how the fix is trying to workaround it. 
However, after pair-testing it, it seems that there's something weird going on with the ray-traced "specular shadow". When the object is not 100% diffuse (smoothness > 0), you can see the specular shadow edge (limit between the raytraced in the front and raster shadow in the back) moving according to the camera view. However, even when the object is 100% diffuse, you can still see the edge of this shadow moving depending on the camera view especially when the camera pass in front of the angle. 
GIF is with the denoiser off as is more easy to see what's the problem (Issue is still visible with but more subtle)

8d459c3e66aef537fe7708ce4b2976fe

Also, as noted in the PR, there's still the same (more subtle) issue with punctual lights

anisunity:
Fixed additional issues and pair tested

remi-chapelain (approved):
New fix has been pair tested and a lot of edges cases does not repro 🥳

The bug mentions more subtle issues with punctual lights. I'll investigate further to see if that's necessary to split the fogbugz in two in the future.

sebastienlagarde (approved):

anisunity:
Merged master and launched the yamato tests

sebastienlagarde:
Yamato was already green no need for relaunch

PR #2550, opened 2020-11-05
[ShaderGraph] Fix 'Reset Reference' on a Blackboard property throwing a NullReferenceException

Nightmask3:

### Checklist for PR maker

---
### Purpose of this PR

Fix for bug: https://fogbugz.unity3d.com/f/cases/1281831/

Also fixed an additional edge case around this functionality where a blackboard property could have its reference name changed, the graph saved and closed and then opened again, but when user right clicks on property there would not be any option to reset the reference. 

---
### Testing status

- Changing the reference name after accessing options and right-click menu through both Blackboard and the Property Node on the graph, and then resetting the reference
- Changing reference name, saving, closing and reopening graph and then trying to reset reference name, both through Blackboard and Property Node
- Added automated test in PropertyTests.cs to access blackboard property options, change the reference name and trigger the reference name reset through code and verify that the reference name matches the default reference name at end of test.

---
### Comments to reviewers
@marctem I added the backport labels for all the packages we're currently still supporting as I think it's a fairly annoying bug to have to deal with and basically breaks reference name usage entirely, but will defer to your opinion on what needs backporting to.


TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ShaderGraphUITestHelpers.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                var assemblyName = assembly.GetName().ToString();
+                if (assemblyName.Contains("UnityEngine.UIElementsModule"))
+                {
+                    uiElementAssembly = assembly;
+                }
+            }
+            Type uiElementsUtilityType = uiElementAssembly?.GetType("UnityEngine.UIElements.UIElementsUtility");
+
+            MethodInfo createEventMethodInfo = uiElementsUtilityType?.GetMethod("CreateEvent",
```
marctem, line 31: Also, cache the MethodInfo in a class member the first time you grab it so you only need to reflect once and can then just invoke it for subsequent events.

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ShaderGraphUITestHelpers.cs
```diff
@@ -0,0 +1,124 @@
...
+                var element = container.hierarchy[i];
+                TestAllElements(element, test);
+            }
+        }
+
+        public static void SendMouseEventToVisualElement(
+            VisualElement elementToNotify,
+            EventType eventType,
+            MouseButton mouseButton = MouseButton.LeftMouse,
+            Vector2 eventPositionOffset = new Vector2())
```
marctem, line 102: Nitpicky but to avoid unnecessary garbage, you can use Vector2.Zero as the default param value.

com.unity.shadergraph/Editor/Drawing/Blackboard/BlackboardProvider.cs
```diff
@@ -208,7 +208,7 @@ void AddPropertyItems(GenericMenu gm)
             shaderInputTypes.Sort((s1, s2) => {
                 var info1 = Attribute.GetCustomAttribute(s1, typeof(BlackboardInputInfo)) as BlackboardInputInfo;
                 var info2 = Attribute.GetCustomAttribute(s2, typeof(BlackboardInputInfo)) as BlackboardInputInfo;
-            
+
```
marctem, line 211: Revert this file, only whitespace changes are unnecessary

marctem (changes requested):

LandonTownsendUnity:
"reset reference" doesn't exist in 8.x or 7.x so we don't need to backport this to those versions

LandonTownsendUnity (approved):
1. Tried resetting reference name on every property type and making sure it worked
2. Tried saving, closing, reopening, and resetting reference name on every property type to make sure that worked.

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/PropertyTests.cs
```diff
@@ -1,10 +1,15 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
```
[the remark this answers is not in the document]
Nightmask3: Ah good call, these weren't necessary, will remove!

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ShaderGraphUITestHelpers.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                var assemblyName = assembly.GetName().ToString();
+                if (assemblyName.Contains("UnityEngine.UIElementsModule"))
+                {
+                    uiElementAssembly = assembly;
+                }
+            }
+            Type uiElementsUtilityType = uiElementAssembly?.GetType("UnityEngine.UIElements.UIElementsUtility");
```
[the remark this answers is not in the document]
Nightmask3: Thanks for reminding me of TypeCache, forgot that it was the preferred way, but in this particular case I don't think that will work because the type being derived from is an internal type, `IUIElementsUtility`, as is the derived type I'm actually referencing `UIElementsUtility`. Throws a compile error about the internal class access. :(

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ShaderGraphUITestHelpers.cs
Nightmask3: Will do! Good call 👍

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ShaderGraphUITestHelpers.cs
Nightmask3: Sounds good!

com.unity.shadergraph/Editor/Drawing/Blackboard/BlackboardProvider.cs
Nightmask3: Will do 👍

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ShaderGraphUITestHelpers.cs
Nightmask3: Something interesting was that I was unable to use Vector2.zero as a default parameter here, due to the compiler error:
"Default parameter must be a compile time constant". I was unaware that was a language requirement.

Replaced it with the `default` keyword which still instantiates a new instance, but the other option of explicitly stating `new Vector2(0.0f, 0.0f)` failed with the same compiler error as above.

marctem (approved):

marctem (approved):

elizabeth-legros (approved):

Nightmask3:
Current Master status:

image

Current PR status:

image

PR #2556, opened 2020-11-06
[HOLD] Specular color on HDRP/Lit below 2% can be used to suppress specular lighting completely

FrancescoC-unity:
The specular color can now be used to completely kill specular. Values from 2% to 0% slowly kill fresnel killing the remaining specular that would be there otherwise. 

Before with black specular color

image

Now with black specular color 

image

Here is a gif to show transition 

specColor

**What did I test**: 

Can be seen in the gif :) 

Yamato: (Just kicked) https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fkill-specular/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/4112834/job


TomasKiniulis (approved):
I think that dev testing is sufficient for this and I expect leftover issues would be revealed by automation. If Yamato is green, looks good to me.

FrancescoC-unity:
I am seeing several yamato failures, will investigate shortly.

sebastienlagarde (approved):

LaserYGD:
Hi, we need this, and currently it seems the only way to suppress specular is to disable affect specular from the directional light. 
And possible alternatives aren't viable in deferred mode.
Any hope/update on this pull?

BOXOPHOBIC:
Hi, the transmission/sss shaders don't have a specular color. Will this feature be available for those shader types? We really need this for foliage shaders. Could this be added to the smoothness instead? 0 smoothness = 0 spec highlight. Thanks!

sebastienlagarde:
recreated in private repo

Dawie3565:
Hi yes we need this 
also affects Unity's new HD cloud systems in that we can not shut off on skin or any foliage shaders 

Where is it ?

PR #2559, opened 2020-11-06
Fixed custom pass null reference exception in draw renderers custom pass

alelievr:
### Purpose of this PR
Fix this case: https://fogbugz.unity3d.com/f/cases/1289959/

Note that the repro of the bug is not correct, you need to use the draw renderers custom pass and not the fullscreen one to repro the issue (as written in the stack trace)

---
### Testing status
I create a Draw Renderers custom pass and removed HDRP from the graphics settings, no error in the console with this fix.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fcustom-pass-nullref-editor/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/4115673/job


sebastienlagarde (approved):

TomasKiniulis (approved):
Looks good to me. Code changes look minor and Antoine already validated that the issue does not occur.

PR #2570, opened 2020-11-09
[HDRP] Fix issue when changing FoV with the physical camera fold-out closed

pmavridis:
### Purpose of this PR
When the "link FOV to physical camera" checkbox is selected but the physical camera fold-out is closed, moving the slider is not working at all (slider is "frozen").
 
I can also repro this on HDRP-v8.2.0, so we had this issue for a long time. 
Some other issues with the linking were fixed recently ( https://github.com/Unity-Technologies/Graphics/pull/2407 ) but we have missed this one!

---
### Testing status
- Changing the FoV value by moving the slider works as expected, even when the physical camera fold-out is closed.
- I have tried various combinations of changing FoV and/or focal length with/without the "link FOV to physical camera" checkbox selected and physical camera fold-out closed or open. Seems to work fine.


remi-chapelain (approved):
Issue seems to be fixed indeed ✔️ 
I also tried with multiple camera on the scene ✔️

sebastienlagarde (approved):

PR #2572, opened 2020-11-09
[Performance][URP] Add iOS tests

sophiaaar:
### Purpose of this PR
* Add iOS runtime and build tests
* Fix Android build tests
* Change Boat Attack branch to master

---
### Testing status
* Ran tests on yamato (https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/performance%252Fios)
* Checked performance RunId in BigQuery to ensure the correct data is being reported
* Ran non-perf editmode tests to check that they are unaffected by the changes to their yml

---
### Comments to reviewers

### DIFF
--- a/.yamato/config/universal_perf_boatattack.metafile
+++ b/.yamato/config/universal_perf_boatattack.metafile
@@ -8,6 +8,7 @@ test_platforms:
     utr_flags:
       - [all]: --report-performance-data
       - [all]: --performance-project-id=BoatAttack
+      - [iPhone_Metal]: --timeout=2400
     utr_flags_build:
       - [all]: --extra-editor-arg="-executemethod" --extra-editor-arg="Editor.Setup"
       - [all]: --extra-editor-arg="-addscenetobuild=Assets/scenes/Testing/benchmark_island-static.unity"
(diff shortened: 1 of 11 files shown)


.yamato/config/__shared.metafile
```diff
@@ -138,6 +138,8 @@ test_platforms:
       - [all]: --scripting-backend=<SCRIPTING_BACKEND> 
       - [all]: --platform=editmode
       - [all]: --extra-editor-arg="-colorspace=<COLORSPACE>"
+      - [Android_OpenGLES3, Android_Vulkan]: --platform=Android
```
kivistikliis, line 141: I was wondering, if this is performance specific stuff, not general editmode (since in general we dont do editmode for those anyways), should it be in performance metafile or shared? I don't have much preference in this, so feel free to leave as is! just thought to bring it to discussion :)

kivistikliis (approved):
looks good to me!

.yamato/config/__shared.metafile
sophiaaar: i was trying to keep it consistent with standalone and playmode, and ios/android only run performance editmode tests, so this part doesnt affect the other jobs :)

sophiaaar:
cancelled un-needed jobs as windows was untouched in the last commit anyway

PR #2574, opened 2020-11-09
[HDRP] Fix issue with previous frame exposure texture being not null, but containing garbage

FrancescoC-unity:
The first frame we bind exposure to the empty texture to avoid using uninitalized history. Second frame both history textures are allocated, however because previous frame we did not use the history system, they will now contain garbage.

This fixes issues that were reported on some platforms that caused random black squares appearing on screen (due to volumetric using previous frame exposure along the current frame exposure). 

The fix perform a small clear to a valid exposure texture upon allocating the textures, so that they are never invalid. This is essentially free as it is a 1x1 texture. 

**What did I test**
This has been validated as a valid fix but the person that reported the original issue. 
I also checked that the template still looks the same. 

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-exposure-history-being-garbage-first-frame/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk

### DIFF
--- a/.yamato/config/universal.metafile
+++ b/.yamato/config/universal.metafile
@@ -14,7 +14,7 @@ test_platforms:
   - type: playmode
     name: playmode_XR
     utr_flags:
-      - [all]: --extra-editor-arg="-xr-tests" 
+      - [all]: --extra-editor-arg="-xr-tests"
 platforms:
   - name: Win
     apis:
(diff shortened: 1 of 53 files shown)


sebastienlagarde (approved):

sebastienlagarde:
https://github.com/Unity-Technologies/Graphics/pull/2621/

PR #2575, opened 2020-11-09
Consolidated ray tracing documentation to make blue book links point to reference documentaiton

JordanL8:
### Purpose of this PR
Moves the ray-traced effect reference documentation to the override where it actually exists and provided suitable links from the original page. Now when a user presses the blue book expecting to see information about ray-traced effects, they will see them

Are there any more to do? The rest didn't seem to be on overrides so I left them as they were.


sebastienlagarde (approved):

sebastienlagarde:
Great, yes you have catch all of them. thanks!

PR #2578, opened 2020-11-10
[HDRP] Physical DoF: fix tiling artifacts, switch to blue noise sampling, add anamorphism support

pmavridis:
### Purpose of this PR 
- Add anamorphism support (as requested by internal team)
- Fix/improve dithering artifacts by switching to blue noise sampling
- Fix tiling artifacts (removed min/max pyramid sampling that was causing this).
- Big shader clean-up (remove unused code)

Note on speed:
Blue noise sampling made things a bit slower than before. Still, the speed is between the  "medium" and "high" preset when tested on the new HDRP template scene. This is true as long as you don't use more than 9 samples in the settings, which with TAA looks very good. Most scenes look fine even with 3 samples + TAA (which internally translates to 3 rings).

---
### Testing status
- Deleted shader cache to check for any new compiler warning
- Tested in 6 different views of the new HDRP template scene, I could not get any tiling artifacts anymore.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFGather.compute
+++ b/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFGather.compute
@@ -125,7 +125,6 @@ void KMain(uint3 dispatchThreadId : SV_DispatchThreadID)
     int numSamples = maxRadius > 0 ? NumRings : 0;
 
     // Select the appropriate mip to sample based on the amount of samples. Lower sample counts will be faster at the cost of "leaking"
-    // Note: We have generated only 4 additional mip levels, so don't go any deeper than that
     float lod = min(MaxColorMip, log2(2 * PI * maxRadius * rcp(numSamples)));
 
     // Gather the DoF samples


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added missing documentation for unsupported SG RT nodes and light's include for raytracing attrbute.
 - Added documentation for LODs not being supported by ray tracing.
 - Added more options to control how the component of motion vectors coming from the camera transform will affect the motion blur with new clamping modes.
+- Added anamorphism support for phsyical DoF, switched to blue noise sampling and fixed tiling artifacts.
```
pmavridis, line 33: I don't see a tag for 10.2 yet, so I hope we can take this in :) Otherwise will move to next release.

sebastienlagarde (approved):

sebastienlagarde:
For QA: to test in the final package 10.2. thanks

TomasKiniulis:
Looks good!
- Validated that no tilling occurs when switching layouts, exiting entering fullscreen
- Validated that Anamorphism works with Physical DoF.
CTAyNMFgvP
- Dithering is improved and is not popping without AA and visible flickering does not occur with TAA
No TAA with PR
NoTAA_improvedDither
No TAA before PR
NoTAA_previousDither

PR #2582, opened 2020-11-10
[HOLD] Start of moving debug modes from Shaderpass forward to full screen debug pass 

FrancescoC-unity:
This is the start of a work moving debug modes to full screen debugs outside of the forward pass. It is a work in progress work, so not we cannot yet remove parts from the forward pass, but it will be when possible. 

This PR should essentially moves the material debug modes (excluding validator) to Full Screen Debug pass. 

**What did I test** 
I tried several material debug modes and verified it is equivalent to what was there before.


sebastienlagarde (approved):

FrancescoC-unity:
Custom yamato on 2020.2, 2 tests failing are expected as not supported on 2020.2 (cached shadows) and one that seems very unrelated only on vulkan (decal stress test) 

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fmove-some-debug-mode-to-debug-pass/.yamato%252Fall-hdrp.yml%2523All_HDRP_CUSTOM-REVISION/4255036/job/dependencies

FrancescoC-unity:
Apparently there is something missing on the VFX template, I am gonna fix that up first thing Monday.

FrancescoC-unity:
Right, the issue is that the pass was not added to VFX originally. What this mean is that also the Vertex/Quad Density debug views are currently broken in master. 

I contacted @PaulDemeulenaere  and actually adding the pass would require a big-ish change and VFX would like to offer some help as it could lead to a refactor they want.

PR #2589, opened 2020-11-11
[11.x] Move the graphics test reference images

jessebarker:
Move the graphics test reference images to the directories in which the test framework will look for them

    WindowsEditor/DX11 was already correct.
    No actual image changes, only location.
    Updates on a follow-up pass as we need those images to be generated by the bokken hosts.

### Purpose of this PR
The only platform on which we are running ShaderGraph graphics tests (playmode) is WindowsEditor on Direct3D 11.   All other testing is editmode.  In fact, if you request a run of the playmode tests on other platforms (e.g., WindowsEditor on Vulkan), it will fail due to "No reference image provided" because the images are in the wrong directory.  This PR puts all of the reference images for the Shader Graph test project into the correct subdirectory (notice the "None" in the file path).

---
### Testing status
Run locally on OSXEditor/Metal
Will be launching Yamato jobs for other configs shortly.
I don't expect the results to all match, as there are no changes to the images here, but the tests themselves should run successfully.

---
### Comments to reviewers
Note the 'None' in the file path.  This should also allow us to have results for multiple device configs in the future (if, for example, different GPUs generate different results that are allowable under the platform/API config).


jessebarker:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Fupdate-test-reference-image-locations/.yamato%252Fshadergraph-win-vulkan.yml%2523ShaderGraph_Win_Vulkan_playmode_mono_Linear_trunk/4163222/job
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Fupdate-test-reference-image-locations/.yamato%252Fshadergraph-linux-vulkan.yml%2523ShaderGraph_Linux_Vulkan_playmode_mono_Linear_trunk/4163221/job

jessebarker:
So they're failing because the images don't match, and not because the images aren't there.

elizabeth-legros (approved):

jessebarker:
The CI run is stalled on two of the sub-pipelines (elapsed time almost 20 hours).  However, we're not running these tests as part of CI (I kicked off the run to prove that I didn't touch anything that affects our current CI loop).  The two failed runs above were manually started to show that the test runner can now find the files and that there are differences (that's the next pass).  Hopefully this is mergeable soon so we can start evaluating differences to see what's correct and what needs tweaking.

marctem (approved):

PR #2592, opened 2020-11-12
HLSL CodeGen Improvements

fredericv-unity3d:
### Purpose of this PR
When the HLSL CodeGen C# parser parses a file that contains C# 7 feature, the codegen stop silently.

This PR fixes this behaviour and improve codegen performance slightly.

Explanation:
* The CodeGen uses NRefactory parser to parse all C# file to find the location of C# type definition to uses.
* NRefactory does not support C# 7

Solution:
* We don't need to parse the C# code to find the location of type definition, we already have an attribute on types to search.
* I updated the `GenerateHLSL` attribute constructor to get the source code file path by the compiler. (By using the [ CallerFilePath](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callerfilepathattribute?view=net-5.0) attribute).
* The NRefactory dependency and the parser are removed.

---
### Testing status
* Manual testing:
1. Deleted some generated files
1. Added C# 7 features the associated C# file containing types with `GenerateHLSL` attribute
1. Triggered the shader code generation
1. Observed that the shader code was appropriately generated for the deleted files
1. Observed that other generated files are the same as before

---

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/Fog.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/Fog.cs
@@ -234,7 +234,6 @@ namespace UnityEngine.Rendering.HighDefinition
     /// <summary>
     /// Fog Color Mode.
     /// </summary>
-    [GenerateHLSL]
     public enum FogColorMode
     {
         /// <summary>Fog is a constant color.</summary>
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Decal/Decal.cs.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Decal/Decal.cs.hlsl
@@ -9,17 +9,6 @@
 //
 #define DBUFFERMATERIAL_COUNT (4)
 
-// Generated from UnityEngine.Rendering.HighDefinition.Decal+DecalSurfaceData
-// PackingRules = Exact
-struct DecalSurfaceData
-{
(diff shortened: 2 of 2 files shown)


remi-chapelain (commented):
Thank you for adding what you've tested !
However, could you add what you think need further testing for QA if it's needed ?

fredericv-unity3d:
@remi-chapelain 
> 
> 
> Thank you for adding what you've tested !
> However, could you add what you think need further testing for QA if it's needed ?

Testing on your machine that the shaders are not broken when regenerating would be nice.

sebastienlagarde (approved):

TomasKiniulis (approved):
Seems to be good. I did not observe any broken or odd rendering shaders
What I've tested:
- Opening template from master and opening back on PR
- Opening template with clean library
- Upgrading Amelienborg project to PR.

sebastienlagarde:
Validated locally that is works correctly, 
Launched one standalone and one edit mode Yamato to be sure it compile:

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fhlsl_gen_7/.yamato%252Fhdrp-win-dx11.yml%2523HDRP_Win_DX11_editmode_mono_Linear_trunk

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fhlsl_gen_7/.yamato%252Fhdrp-win-dx11.yml%2523HDRP_Win_DX11_editmode_mono_Linear_trunk/4456988/job

PR #2597, opened 2020-11-12
Added a warning when trying to bake with static lighting being in an invalid state.

JulienIgnace-Unity:
### Purpose of this PR
Added a warning when trying to bake with static lighting being in an invalid state.

It is possible that HDRP hasn't rendered any frame when clicking the bake lighting button.
This can happen when baked lighting debug are used for example and no other window with HDRP is visible.
This will result in the static lighting cubemap not being up to date with what the user put in the Environment Lighting panel.
We detect this and issue a warning if applicable.

---
### Testing status
Tested the repro steps:
- Have only scene view displaying.
- Bake lighting with a valid static sky.
- Switch to debug baked lightmaps in the scene view.
- Switch static sky to None and bake.
- Switch back to a valid sky and bake again => lighting is not updated => This is where the warning should be issued.

Also made sure that the warning did not appear with a valid sky.


remi-chapelain (approved):
This one as already been discussed while pair-testing [this repro](https://fogbugz.unity3d.com/f/cases/1277157/) with @JulienIgnace-Unity
Issue was found using 7.x.x but not problematic enough to backport I suppose ! ✔️

sebastienlagarde (approved):

sebastienlagarde:
https://fogbugz.unity3d.com/f/cases/1289240/

PR #2598, opened 2020-11-12
Fixed an issue with the clear coat and SSR(RTR) and a ghosting issue with RTGI and RTAO.

anisunity:
Fixed the clear coat not being handled properly for SSR and RTR (case 1291654). 
Fixed ghosting in RTGI and RTAO when denoising is enabled and the RTHandle size is not equal to the Viewport size (case 1291654).
Fixed the wrong Profiling sampler being use for quality RTGI.
https://fogbugz.unity3d.com/f/cases/1289860/
https://fogbugz.unity3d.com/f/cases/1291654/

Testing status
Ran the DXR tests locally, all green.
Ran the HDRP tests locally (Still running).
QA needs to try to reproduce the steps described in the fogbugz for the ghosting to ensure it solved the issue.
QA needs to do comparisons between reflection probes and SSR/RT source for Clear coat lighting on lit to make sure that the issue is solved.


anisunity:
One of the test' result changed (SSR + CC) need to update the screenshots for that tests. The rest of HDRP tests are green locally.

remi-chapelain (approved):
Both issues seems to be properly fixed ✔️ 
Quick side note on clear coat, when we have a clear coat value, RTR/SSR switch from using the smoothness object to reflecting the smooth clear coat . However, this value seems to be arbitrary around 0.018xxx, is this hardcoded ? Is there any reason it isn't just > 0 ?

da585c37b33577f38b4df63b74295f4b

sebastienlagarde (approved):

PR #2600, opened 2020-11-12
Fixed multi-editing with new light intensity slider

adrien-de-tocqueville:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1290191/

I am not sure why this apply was needed, i didn't find a case in which undo redo would not work without it.
A function higher in the callstack is already doing the Apply inside an EndChangeCheck test

---
### Testing status

Tested the repro steps with the light explorer (which is not actually related).
and undo redo when editing and multi editing lights with various units
Tested undo redo when changing light unit
Tested changing light value with presets

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
@@ -630,6 +630,7 @@ namespace UnityEditor.Rendering.HighDefinition
             unitRect.width = k_UnitWidth + .5f;
 
             // Draw the unit textfield
+            EditorGUI.BeginChangeCheck();
             EditorGUI.PropertyField(valueRect, serialized.intensity, s_Styles.empty);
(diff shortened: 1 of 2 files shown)


johnpars (commented):
So this Apply was originally added to ensure proper Undo/Redo when switching between light unit types. 

Example:
- Create a new punctual light
- Switch from Lumen -> Candela -> EV100
- Undo

You will see that the undo apply in master, but not this branch.

Ideally, we should find a way to fix this regression for multi-editing, and also ensure that this undo functionality still works.

adrien-de-tocqueville:
It's not a pretty solution, but I managed to fix both behaviors by only doing the unit conversion in a local variable and write it back only when changed

johnpars (approved):
LGTM - While it's unfortunate that we simply can't turn the light unit slider into a pure view of the data (IE ditch the SerializedProperty completely and simply refer to the float value), it's necessary due to how the context menu works right now.

sebastienlagarde (approved):

PR #2602, opened 2020-11-12
Fix issue with density volume gizmo editing

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1291621/ 

There was an issue in the calculations of the fade when the per axis control was set to off. 

This PR fixes that (see gif for fixed and bug for broken)
fixedGizmo

**What did I test**
See above gif


FrancescoC-unity:
Uhm somthing looks off when targeting hd/bugfix with a branch off master. Will reopen a branch starting from that branch instead of master.

PR #2606, opened 2020-11-12
[HDRP][Compositor] Fix issues with image layers

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1289936/
Image layers in the compositor with a None texture were rendering black, which can be perceived as an error when used on top of another camera (with stacking). It makes more sense to make them don't render anything, until the user assigns a texture.
If they are the first layer/camera on the stack, they still clear to black.

---
### Testing status
- When adding a new image layer, as described in the repro steps, the output is unaffected, until the user assigns a texture.
- Compositor test is green


pmavridis:
The first of the PR also included some "camera management" improvements, but it turns out that they were causing some issues so I reverted those. Tests are now green.

TomasKiniulis (approved):
Not encountering black screen when adding image sublayer anymore. Seems to be good. Haven't ran into any PR specific compositor issues

sebastienlagarde (approved):

PR #2607, opened 2020-11-12
Added setting to control zclip and disable shadow pancaking

nigeljw-unity:
We should align with HDRP and use ShaderLab's DepthClip attribute to set the raster state for rendering shadow maps instead of explicitly clipping the depth value in the shader. API support for DepthClampEnable is only 60% for Vulkan devices, so we need to be able to fall back to the explicit clipping if necessary.

Requires corresponding changes to trunk for adding check for platform support of depth cliplling/clamping:
https://github.cds.internal.unity3d.com/unity/unity/pull/2226

# **Please read;**

### Checklist for PR maker

---
### Purpose of this PR

---
### Testing status

---
### Comments to reviewers


com.unity.render-pipelines.universal/Editor/ShaderGUI/Shaders/LitShader.cs
```diff
@@ -144,6 +144,20 @@ public override void AssignNewShaderToMaterial(Material material, Shader oldShad
                     material.SetTexture("_MetallicSpecGlossMap", texture);
             }
 
+            if (material.HasProperty("_ZClip"))
+            {
+                if (SystemInfo.supportsDepthClip)
+                {
+                    material.SetFloat("_ZClip", 1);
+                    //material.EnableKeyword("_ZClipEnable");
```
nigeljw-unity, line 152: I was looking at a few different ways to support the fallback, and I just left the code for using keywords in here for now for discussion purposes.

sebastienlagarde:
Hey

>We should align with HDRP and use ShaderLab's DepthClip attribute to set the raster state for rendering shadow maps instead >of explicitly clipping the depth value in the shader

Note: I am not sure exposing ZClip as an option is a good idea. In HDRP we control it automatically via our shadow system and enabled it for all but directional light (in addition in our context all HDRP platforms support it, but for URP you need to consider platform not supporting it - so not exposing it will make sense. Of course it depends on your use case).

nigeljw-unity:
> Hey
> 
> > We should align with HDRP and use ShaderLab's DepthClip attribute to set the raster state for rendering shadow maps instead >of explicitly clipping the depth value in the shader
> 
> Note: I am not sure exposing ZClip as an option is a good idea. In HDRP we control it automatically via our shadow system and enabled it for all but directional light (in addition in our context all HDRP platforms support it, but for URP you need to consider platform not supporting it - so not exposing it will make sense. Of course it depends on your use case).

Hey @sebastienlagarde , thanks for the feedback! I was thinking about adding a warning to the material properties GUI when depth clipping is enabled (and not supported) to address this concern. Since depth culling is currently enabled by default in the raster state, this would mean applications running on existing platforms where depth culling is not supported would now have the warning (which may confuse users).

The main reason I am exposing this setting is because depth clipping can cause similar issues. I expect users would see a similar problem on HDRP. On HDRP, it is encouraged to use light cookies for light extremely sources close to objects, but we do not yet support light cookies in URP. Here is a contrived example of the problem with depth clipping running in URP:

DepthClipEnabledWithoutPancaking

com.unity.render-pipelines.universal/Shaders/ShadowCasterPass.hlsl
```diff
@@ -27,10 +27,12 @@ float4 GetShadowPositionHClip(Attributes input)
 
     float4 positionCS = TransformWorldToHClip(ApplyShadowBias(positionWS, normalWS, _LightDirection));
 
+#if defined(_SHADOW_PANCAKING_ON)
```
phi-lira, line 30: The way this is define might change the existing behaviour for custom user shaders. 
F.ex, a user has a custom shader that includes ShadowCaster pass. Their material doesn't set the _SHADOW_PANCAKING_ON keyword. Once they update to this URP version their custom shadow shaders will stop doing pancaking and that will be a regression. 

What we should do is to introduce a keyword to disable shadow_pancaking instead, that way the default behaviour when not defining the keyword is to do pancaking, keeping existing behaviour for custom shaders unchanged.

phi-lira (changes requested):
A few thoughts about this.

1) Do we want to enable both ZClip and Pancaking separately? Aren't they correlated in any way? 

2) I'm worried about the workflow to change ZClip / Pancaking per-material. This means that if user wants to disable pancaking for isntance they will have to do so for all the material in the project + if they have custom shaders they will have to add custom editors for artists to disable that as well. This is also something that would add overhead to production (the art lead would have to make sure all new materials added to the project conform to the desired pancaking setting). 

It seems to be exposing ability to disable pancaking globally might be more efficient. Art / Project lead would set this setting once and all content would be consistent. If we have a global setting controlled by a keyword we could strip it avoiding bloating project with more shader variants.

com.unity.render-pipelines.universal/Shaders/ShadowCasterPass.hlsl
nigeljw-unity: Switching this to a disable definitely sounds like a good idea!

nigeljw-unity:
> A few thoughts about this.
> 
> 1. Do we want to enable both ZClip and Pancaking separately? Aren't they correlated in any way?
> 2. I'm worried about the workflow to change ZClip / Pancaking per-material. This means that if user wants to disable pancaking for isntance they will have to do so for all the material in the project + if they have custom shaders they will have to add custom editors for artists to disable that as well. This is also something that would add overhead to production (the art lead would have to make sure all new materials added to the project conform to the desired pancaking setting).
> 
> It seems to be exposing ability to disable pancaking globally might be more efficient. Art / Project lead would set this setting once and all content would be consistent. If we have a global setting controlled by a keyword we could strip it avoiding bloating project with more shader variants.

@phi-lira the problem with using a global setting is that we only really want to disable this on a per object basis for geometry that the user knows will get very near to light sources. In the original issue, a global setting would make sense since the player is moving around the environment with a flashlight, but in general, I do not think this is the case.

I do agree it will be frustrating to disable per material for existing projects, but I really doubt we are going to come across a project where they want to disable this globally (maybe per scene at best). The ideal solution for this is support for material variants in URP. I chatted with @Verasl about this previously, but would be good to get his perspective again. I am definitely concerned about custom shaders. A lot of custom shaders seem to be a modified copy of old default URP shaders, which are missing potential fixes like this.

Currently, we have both pancaking and depth clipping both enabled by default, and I want to leave this as the default state to avoid any regressions, which is why I added separate enables. They are somewhat redundant operations but they are clearly causing different functionality.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalTarget.cs
```diff
@@ -475,6 +475,7 @@ static class CoreRenderStates
             { RenderState.ZWrite(ZWrite.On) },
             { RenderState.Cull(Cull.Back), new FieldCondition(Fields.DoubleSided, false) },
             { RenderState.Cull(Cull.Off), new FieldCondition(Fields.DoubleSided, true) },
+            { RenderState.ZClip(ZClip.On) }, //, new FieldCondition(UniversalFields.ZClip, false) },
```
eh-unity, line 478: I wonder how platforms that doesn't support a feature a handled in the graph?

Remove commented code.

com.unity.render-pipelines.universal/Editor/ShaderGraph/UniversalFields.cs
```diff
@@ -23,6 +23,7 @@ internal static class UniversalFields
         public static FieldDescriptor NormalDropOffTS =       new FieldDescriptor(string.Empty, "NormalDropOffTS",        "_NORMAL_DROPOFF_TS 1");
         public static FieldDescriptor NormalDropOffOS =       new FieldDescriptor(string.Empty, "NormalDropOffOS",        "_NORMAL_DROPOFF_OS 1");
         public static FieldDescriptor NormalDropOffWS =       new FieldDescriptor(string.Empty, "NormalDropOffWS",        "_NORMAL_DROPOFF_WS 1");
+        public static FieldDescriptor ZClip =                 new FieldDescriptor(string.Empty, "ZClip",                  "_ZClipEnable 1");
```
eh-unity, line 26: Above it seems that you use `_ZClip 1`, we should be consistent.

com.unity.render-pipelines.universal/Editor/ShaderGUI/Shaders/LitShader.cs
eh-unity: At least I'm just fine with `_ZClip 1` and reference the property value .shader.
You could add an enum for the `1`. Maybe ZClip.On / ZClip.Off or Enable/Disable. Similar to Blend mode.

com.unity.render-pipelines.universal/Shaders/ShadowCasterPass.hlsl
eh-unity: That's a bit unfortunate. Having randomly positive/negative conditional defines will make a mess as a whole. Is there no other option to keep the condition positive? Can we upgrade materials to have it enabled by default?

eh-unity (approved):
From implementation perspective this seems to be fine. I added some minor improvement comments.
I don't have a solid opinion on whether this should be exposed publicly or not. Perhaps not, IF we could handle it automatically somehow.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -528,6 +554,15 @@ public static void SetMaterialKeywords(Material material, Action<Material> shadi
             if (material.HasProperty("_BumpMap"))
                 CoreUtils.SetKeyword(material, ShaderKeywordStrings._NORMALMAP, material.GetTexture("_BumpMap"));
 
+            if (material.HasProperty("_ShadowPancaking") && material.GetFloat("_ShadowPancaking") == 1.0f)
+            {
+                material.EnableKeyword("_SHADOW_PANCAKING_ON");
```
nigeljw-unity, line 559: @lukaschod should I do this differently now with the new shader striping?

ernestasKupciunas (approved):
Looks alright. No issues were found related to this PR.
I have tested URP Template Scene and Winter project.
I have also tested an issue described in this Slack thread: https://unity.slack.com/archives/C89KFUUCT/p1637744931341300
Enabled Shadow pancaking:
ShadowPancakingEnabled
Disabled shadow pancaking:
ShadowPancakingDisabled
If large triangles indeed can create an unwanted result(artifacts) on large triangles when depth clipping is enabled:
ezgif com-gif-maker

esmelusina:
@ernestasKupciunas Can you test the SG equivalent shader as well for parity? Thanks.

com.unity.shadergraph/Editor/Generation/Descriptors/RenderStateDescriptor.cs
```diff
@@ -39,7 +39,6 @@ internal static class RenderState
         public static RenderStateDescriptor ZTest(string value) => new RenderStateDescriptor { type = RenderStateType.ZTest, value = $"ZTest {value}" };
         public static RenderStateDescriptor ZWrite(ZWrite value) => new RenderStateDescriptor { type = RenderStateType.ZWrite, value = $"ZWrite {value}" };
         public static RenderStateDescriptor ZWrite(string value) => new RenderStateDescriptor { type = RenderStateType.ZWrite, value = $"ZWrite {value}" };
-        public static RenderStateDescriptor ZClip(string value) => new RenderStateDescriptor { type = RenderStateType.ZClip, value = $"ZClip {value}" };
```
esmelusina, line 42: Does this PR remove an option currently available to users? Don't we need an update path?

PR #2611, opened 2020-11-13
[HDRP] [PathTracing] Rigid animations now trigger PT dirtiness.

eturquin:
Adresses ticket https://fogbugz.unity3d.com/f/cases/1290721

Before: animated objects do not trigger path tracing accumulation reset, when changing animation frame in editor.
Now: animated objects reset path tracing accumulation, when changing animation frame in editor.

This has been tested using keyframed transforms (rotation, translation), and scrubbing through the timeline.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -99,6 +99,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed picking for materials with depth offset.
 - Fixed issue with exposure history being uninitialized on second frame.
 - Fixed issue when changing FoV with the physical camera fold-out closed.
+- Fixed path tracing accumulation not being reset when changing to a different frame of an animation.
 
 ### Changed
 - Combined occlusion meshes into one to reduce draw calls and state changes with XR single-pass.


remi-chapelain (commented):
This fixes only one part of the issue. 
Changing an object transform (position/rotation/scale) via script or timeline now triggers PT dirtiness correctly. ✔️ 
Also checked that it's the same with offscreen objects with extend camera culling checked. ✔️  

eb983d777aec94715be3514f473315cb

However, a skinned mesh renderer animation does not trigger anything. 
It works in a hacky way since if the transform changed, the animation also update since it triggers a reset, but if the only change is skinned animation, animation is frozen. 
I also tried deforming some bones with an insane amount to see if that triggers the RTAS to change size (since there's also a repaint bind to that change in size) thus triggering a reset but in vain. 

**Should I split the fogbugz in two or will you try to fix the skinned mesh in this PR ?** 

8118abe78be1f433472c43e7155d016e

sebastienlagarde (approved):

eturquin:
@remi-chapelain: The rigid transforms were manageable with a minimal change. For the skinned animations, we would need to know from C++ land that the mesh has changed (maybe with a hash on vertex positions), this will be more involved and should be handled separately.

remi-chapelain (approved):
Thanks for the answer. 
Part 1 of the issue is fixed ✔️ , I'll edit and create another fogbugz for the second part.

PR #2616, opened 2020-11-13
Fixed Volume duplication in hybrid workflow

JulienIgnace-Unity:
### Purpose of this PR
When using the hybrid renderer, the Volume objects are duplicated in the scene when editing a subscene. This is because we have both the entity representation and the editable game object representation at the same time. For all objects in this category, filtering is usually done based on the scene culling mask.
For volume however, this filtering was only done when rendering previews or prefab in isolation which caused them to be used twice.
The fix is to always (in editor) do the filtering based on scene culling mask so that the "entity" version of the object is not present while editing the subscene.

Case: https://fogbugz.unity3d.com/f/cases/1271064/

---
### Testing status
I tested the repro steps in the case and ran automatic testing.
However given the nature of the fix, it would be good to do a more thorough testing pass to see if Volumes won't start "leaking" in other scenes when they should not.
Preview and prefab editing (isolation or not) should be tested with volumes containing "additive" effects that will be visible if present twice. A good example is the one from the test case where a volume with a weight at 0.3 contained a sky with a 10 exposure which would then add up because of the 0.3 weight. A fixed exposure component with a similar weight would probably do it too.
Lookdev needs to be looked at too. Having volumes in the main scene not leak into it should be a good enough test.

@pmavridis I added you as a reviewer because I removed the scope forcing filtering that was only used in the Compositor. Could you validate this particular use case?


pmavridis:
> @pmavridis I added you as a reviewer because I removed the scope forcing filtering that was only used in the Compositor.
Looking at the code, the Compositor.cs that is using this feature is in Lookdev, so I guess @RSlysz should have a look.
The graphics compositor is not using this feature.

JulienIgnace-Unity:
My bad. Got a bit too quick on that one.
Will ask QA to test LookDev a bit too then.

TomasKiniulis (approved):
Haven't ran into any issues. Checked repro project with additional post-process and exposure volume overrides. Made sure editing prefabs does not duplicate volume effects, LookDev stays unaffected

RSlysz:
> 
> 
> Haven't ran into any issues. Checked repro project with additional post-process and exposure volume overrides. Made sure editing prefabs does not duplicate volume effects, LookDev stays unaffected

For lookdev, the scope was preventing LookDev sky to leak into the scene, and scene sky to leak into the LookDev. Have you tested this?

RSlysz (approved):
Seams ok. But be sure that there is no leak or volume profile between LookDev and scenes

sebastienlagarde (approved):

Chman (approved):

PR #2617, opened 2020-11-13
[HDRP][Compositor] Fix issue with gameview repaint

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1290622/
The game view was not re-painted properly when side-by-side with scene view.
(Note: the bug is only reproducible when game-view and scene-view were used side-by-side)

This is a change in behavior from the C++ side (the game view is now updated only once, unless you move your mouse on top of it). We already had some code in the compositor to guarantee that the game view is updated at least a few times per second, so I have updated the code to work on the new Unity version.

---
### Testing status
- Tested toggling the compositor's  preview on and off with the game view side-by-side with scene view. The game view now properly updates (no mouse move is necessary).
- This is only a small change in editor/ui code, so I did not run the tests.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Fixed probe volumes debug views.
+- Fixed issue displaying wrong debug mode in runtime debug menu UI.
+- Fixed useless editor repaint when using lod bias.
+- Fixed multi-editing with new light intensity slider.
+- Fixed issue with density volumes flickering when editing shape box.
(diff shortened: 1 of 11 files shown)


pmavridis:
There are two issues related to this bug:
- The first issue is that with the new Unity version, the game view window in the compositor was not refreshing as expected. The intended behavior for the compositor is that the gameview should always re-paint **at least** a few times per second, so people will see the changes when changing TAA modes, toggling layers etc. This PR fixes this issue. Without the fix we only get the first frame.
- There is another issue in HDRP that the first frame is black or white when game-view and scene view are side-by-side. This PR does not fix this issue, but it is something that we should also fix.

remi-chapelain (approved):
Just wanted to test with a moving object to be sure that everything repainted properly (since the repro included an empty static scene). All good ✔️

sebastienlagarde (approved):

PR #2627, opened 2020-11-13
Remove cyanish color shifting during sharpening in YCoCg in TAA

FrancescoC-unity:
This is a change to avoid some color shifting when doing extreme sharpening. Note that the sharpening code is still fairly simplistic and probably need some love at some point. 

This PR rotates back into RGB to perform the sharpening; an alternative would be performing the sharpening only on the luma component, but it looks worse than this way. 

Small before after (see cyan fringe), can provide privately better images

image
image

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fbetter-sharpening-taa-when-ycocg/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/4172383/job 

**What did I test**: 
Checked some scenes + ran tests + more I can discuss via slack.


TomasKiniulis (approved):
Looks good! Checked a few combinations of TAA settings. Getting way less cyan fringe compared to Master.
PR:
pr_TAA
Master:
MasterTAA

sebastienlagarde (approved):

PR #2628, opened 2020-11-13
Fixed some labels clipping in the Render Graph Viewer

JulienIgnace-Unity:
### Purpose of this PR
Fixed some labels in Render Graph Viewer
- Last pass label being clipped on the right.
- Resource names being clipped (they now have an ellipse)

https://fogbugz.unity3d.com/f/cases/1289905/

---
### Testing status
Tested in the viewer directly


remi-chapelain:
This is fixed ✔️ 
To be picky and push things a little further, I would just add a small margin to the biggest label here ! 😄 

image

remi-chapelain:
Also, while we're at it, could we disable the ability to click on the button under the passes (since it does nothing)

remi-chapelain (approved):

sebastienlagarde (approved):

PR #2633, opened 2020-11-16
[11.x.x] Fixing an issue with MSAA not working with Vulkan in Editor Play-Mode

ellioman:
# Purpose of this PR
MSAA is currently not working with Vulkan on Windows platform. 
The reason is that:
1. The editor renders to a Game View Render Texture with the sample count forced to 1. 
2. Vulkan has `SystemInfo.supportsMultisampleAutoResolve` set to true

The solution is to always require explicit MSAA resolve when in the editor.

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fmsaa-editor-issue


manuele-bonanno (approved):

phi-lira (approved):
Please add changelog.

ellioman:
Yamato was green on URP's side. Issues in HDRP and ShaderGraph are known and not related to this PR.

PR #2641, opened 2020-11-16
Added the possibility to disable distance based roughness for reflection probes.

anisunity:
Recently we changed the behavior of the reflection probes to have an approximation for the distance based BRDF integration (that we call distance based roughness). While the feature increases the image quality, it requires the user to be able to set up a proxy that matches roughly the geometry. That is not always possible (exteriors, oddly shapes rooms, etc.). So we added a toggle on the reflection probe itself to be able to disable the feature.
The checkbox is on by default.

Testing status
Ran the HDRP tests locally they were all green.
Compared with and without the checkbox and the behavior is the same as before the feature was added.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/HDProbeUI.Drawers.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/HDProbeUI.Drawers.cs
@@ -268,7 +268,8 @@ namespace UnityEditor.Rendering.HighDefinition
                     );
                 }
 
-                if (owner is HDReflectionProbeEditor && serialized.proxyVolume.objectReferenceValue != null)
+                // Don't display distanceBasedRoughness if the projection is infinite (as in this case we force distanceBasedRoughness to be 0 in code)
+                if (owner is HDReflectionProbeEditor && !(serialized.proxyVolume.objectReferenceValue == null && serialized.probeSettings.proxyUseInfluenceVolumeAsProxyVolume.boolValue == false))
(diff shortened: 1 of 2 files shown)


TomasKiniulis:
Hey @anisunity! Anything in particular that QA should pay attention to testing this?

anisunity:
I don't think so @TomasKiniulis, it is a quite simple PR (only couple lines)

iM0ve (changes requested):
The feature is working but I have a few UX suggestions.

1) Since it only works when Proxies are used. Gray the Distance Based Roughness when Proxy is set to None and Use Influence Volume As Proxy Volume is unchecked.
2) Move the feature into Projection Settings if it makes sense.
3) Adjust the tooltip so it clearly states that will only have a positive result if Proxy closely matches the environment 
@JordanL8 tooltip suggestion:
"When enabled, HDRP uses the assigned Proxy Volume to calculate distance based roughness for reflections. This produces more physically-accurate results if the Proxy Volume closely matches the environment"

image

com.unity.render-pipelines.high-definition/Runtime/Utilities/ProbeSettings.cs
```diff
@@ -278,6 +281,9 @@ public static ProbeSettings NewDefault()
         /// <summary>Defines if the planar reflection should support rough reflections.</summary>
         public bool roughReflections;
 
+        /// <summary>Defines if the reflection probe should support the distance based roughness.</summary>
```
JordanL8, line 284: Indicates whether the ReflectionProbe supports distance-based roughness.

JordanL8 (changes requested):
Can you add the small API summary change as well as the tooltip change Motiejus mentioned

anisunity:
Did the review corrections and merged master and relaunching the tests

iM0ve (approved):
Issue 1 - will be solved later since its a more complicated UX matter. Tracked in fogbugz
Issue 2 - Fixed
Issue 3 - Fixed

3Nk4sR4YLX

sebastienlagarde (approved):

sebastienlagarde:
Doc correction is done. Merging

PR #2643, opened 2020-11-16
Fixed the sampling of the normal buffer in the the forward transparent pass.

JulienIgnace-Unity:
### Purpose of this PR
In the render graph code path, the normal buffer was not bound globally anymore so it wasn't available in the transparent forward pass.
This PR fixes that.

---
### Testing status
Tested the repro project provided by @tlaedre 
Otherwise:
- Make a transparent shader graph that samples the normal buffer (can be done with custom code node and DecodeFromNormalBuffer function.
- Notice that the display is wrong compared to the non render graph path.


sebastienlagarde (approved):

TomasKiniulis (approved):
Works well now!

Checked repro project with Lit Shader Mode set to Forward/Deferred/Both and Render Graph enabled/disabled. Made sure that dx11 player build also does not contain this issue in Forward mode with Render Graph enabled

PR #2651, opened 2020-11-17
[ShaderGraph] [11.x] Fix texture associated state (samplerstate, TexelSize) for subgraphs, CFNs

cdxntchou:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1284036/

This PR converts the shadergraph resource-types (Texture2D, Texture2DArray, Texture3D, TextureCube, and SamplerState) to use a struct, instead of bunch of loosely associated variables. This simplifies a bunch of the argument passing, and fixes a number of issues in the generated code, where we assume we can find the associated variables by name-munging, which often does not work. This also clears up an issue with GLES2 where any explicit SamplerState used anywhere in the shader would cause it not to compile on the GLES2 platform (it will now compile, but will ignore the explicit SamplerStates and effectively always use the texture-associated sampler state).

We specially handle existing Custom Function Nodes that input or output one of these resource types, by introducing the concept of "Bare" resource types for Custom Function Nodes. The Bare resource types represent JUST the resource itself, not the new resource struct. Existing Custom Function Nodes will be versioned to use Bare resource types, so that the users existing HLSL code continues to work. And any CFN using bare resource types will flag a warning to the user that those nodes should be upgraded to use the resource structs instead, guiding the users away from using Bare resource types (as the Bare resource types have all of the bug issues we are solving with the structs).

Backport to 10.x:   https://github.com/Unity-Technologies/Graphics/pull/2577

---
### Testing status

ran all of the same test shaders from the backport PR (see link above)
verified they pass in the same way with the same combos of render pipeline and gfx API:
     (HDRP/URP/GLCore/GLES2/Vulkan/DX11/DX12)

tested custom function node upgrade workflow again explicitly, to make sure they upgrade properly.

tested VFX path renders shadergraph particles (same behavior as in backport)

Ran ray tracing tests locally.

Tested manually setting up ray traced reflections, with both reflector and reflectee being shadergraph shaders sampling textures:
image

ABV: 🔴 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252Ftextures/.yamato%252F_abv.yml%2523all_project_ci_trunk/4378022/job/pipeline

master ABV:  🔴 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252F_abv.yml%2523all_project_ci_trunk/4394230/job/pipeline

Failures match:   🟢 
* HDRP on Win_Vulkan_playmode_mono_Linear on version trunk -- both had a hang in the same way
* HDRP on Win_Vulkan_playmode_XR_mono_Linear on version trunk -- both had a hang in the same way
* HDRP on OSX_Metal_playmode_mono_Linear on version trunk -- both failed same set of tests + same hang
* HDRP on Linux_Vulkan_playmode_mono_Linear on version trunk -- both had a hang in the same way

Console Tests:
https://yamato.cds.internal.unity3d.com/jobs/339-ScriptableRenderPipelinePrivate/tree/sg%252Ffix%252Ftextures/.yamato%252F_abv.yml%2523all_project_ci_trunk/4445144/job/pipeline

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
```diff
@@ -25,6 +25,7 @@ Pass
     $splice(GraphKeywords)
 
     #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
+    #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Texture.hlsl"
```
sebastienlagarde, line 28: arf, we noticed an issue (haven't see it in the 10.x Pr), the Texture.hlsl MUST be define after RaytracingMacro.hlsl in our case as this file override the SAMPLE_TEXTURE2D_LOD / SAMPLE_TEXTURE2D Macro to work correctly with raytrace.

pbbastian (approved):

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -13,23 +13,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - Fixed an issue where nodes with ports on one side would appear incorrectly on creation [1262050]
 
-## [10.3.0] - 2020-11-06
-
-### Added
-
-### Changed
-
-### Fixed
-- Addressed C# warning 0649 generated by unassigned structure members
```
elizabeth-legros, line 23: Why was this removed?

com.unity.shadergraph/Editor/Data/Nodes/SlotValue.cs
```diff
@@ -63,14 +63,76 @@ enum ConcreteSlotValueTypePopupName
         Vector2 = ConcreteSlotValueType.Vector2,
         Float = ConcreteSlotValueType.Vector1, // This is currently the only renaming we need - rename Vector1 to Float
         Boolean = ConcreteSlotValueType.Boolean,
-        VirtualTexture = ConcreteSlotValueType.VirtualTexture
+        VirtualTexture = ConcreteSlotValueType.VirtualTexture,
+
+        BareSamplerState = 1000 + ConcreteSlotValueType.SamplerState,
```
elizabeth-legros, line 68: is this just a priority thing to make them display last? If so, could you just throw in a comment?

elizabeth-legros: Oh just saw later how thats used. Still want a comment explaining though

elizabeth-legros (approved):

LandonTownsendUnity (approved):
Approving since the developer completed their manual testing.

com.unity.shadergraph/CHANGELOG.md
cdxntchou: Ah, because there are two 10.3 sections.. I just merged them..

cdxntchou: I expect when I merge latest it will get rid of this change...

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
cdxntchou: Verified it does the correct thing -- because the SampleTexture* nodes directly call the SAMPLE_TEXTURE2D() macros on the hlsl Texture2D (not using the functions defined in Texture.hlsl), it will call the raytracing-defined macro.

Even if a user has a custom function node that is manually calling SAMPLE_TEXTURE2D(UnityTexture2D) it will still continue to work, as when raytracing redefines the macro, it will switch from calling UnityTexture2D.Sample() to calling UnityTexture2D.SampleLevel(), which works in raytracing.

com.unity.shadergraph/Editor/Data/Nodes/SlotValue.cs
cdxntchou: Yeah, want these at the bottom of the list and the rest of the enum values to match ConcreteSlotValueType.

Hmm...  I don't think the 1000 + is actually necessary...  this isn't a serialized type, so we don't need to preserve old values in any way, it's only used to populate the popup menu selecting the type in a custom function node.

Doesn't hurt though, and calls out that they are kind of special.

cdxntchou: I can add a comment noting that these are special.

PR #2655, opened 2020-11-17
update to the upgrader max version 6

martint-unity:
---
### Purpose of this PR
The upgrader was never updated to max version 6.
Added also a const for that.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
@@ -816,7 +816,6 @@ namespace UnityEngine.Rendering.Universal
             if (k_AssetVersion < 3)
             {
                 m_SoftShadowsSupported = (m_ShadowType == ShadowQuality.SoftShadows);
-                k_AssetPreviousVersion = k_AssetVersion;
                 k_AssetVersion = 3;
             }
 
@@ -826,7 +825,6 @@ namespace UnityEngine.Rendering.Universal
                 m_AdditionalLightsShadowmapResolution = m_LocalShadowsAtlasResolution;
                 m_AdditionalLightsPerObjectLimit = m_MaxPixelLights;
                 m_MainLightShadowmapResolution = m_ShadowAtlasResolution;
-                k_AssetPreviousVersion = k_AssetVersion;
                 k_AssetVersion = 4;
             }
 
@@ -836,7 +834,6 @@ namespace UnityEngine.Rendering.Universal
                 {
                     m_RendererDataList[0] = m_RendererData;
                 }
-                k_AssetPreviousVersion = k_AssetVersion;
                 k_AssetVersion = 5;
             }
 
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -869,7 +870,7 @@ public void OnAfterDeserialize()
 #if UNITY_EDITOR
         static void UpgradeAsset(UniversalRenderPipelineAsset asset)
         {
-            if(asset.k_AssetPreviousVersion < 5)
+            if(asset.k_AssetPreviousVersion < k_MaxAssetVersion)
```
Verasl, line 873: Looking at this now, it seems that this needed to be still 5, as this is the version it changed from single renderer to renderer list..

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -885,7 +886,7 @@ static void UpgradeAsset(UniversalRenderPipelineAsset asset)
                     asset.m_RendererData = null; // Clears the old renderer
                 }
 
-                asset.k_AssetPreviousVersion = 5;
+                asset.k_AssetPreviousVersion = k_MaxAssetVersion;
```
Verasl, line 889: Same here, this probably means we need to re-look at this code

Verasl (changes requested):
I think we changed a bit too much

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -102,8 +102,9 @@ public partial class UniversalRenderPipelineAsset : RenderPipelineAsset, ISerial
         ScriptableRenderer[] m_Renderers = new ScriptableRenderer[1];
 
         // Default values set when a new UniversalRenderPipeline asset is created
-        [SerializeField] int k_AssetVersion = 5;
-        [SerializeField] int k_AssetPreviousVersion = 5;
+        const int k_MaxAssetVersion = 6;
+        [SerializeField] int k_AssetVersion = k_MaxAssetVersion;
+        [SerializeField] int k_AssetPreviousVersion = k_MaxAssetVersion;
```
phi-lira, line 107: Shouldn't this be `k_MaxAssetVersion - 1`?

phi-lira: This logic is rather confusing. Is k_MaxAssetVersion current pipeline asset version?

martint-unity: It is very confusing indeed.

Verasl (approved):
LGTM, should solve some issues without having to remake the entire 'upgrader' for now

martint-unity:
closing due not feeling safe to land.

PR #2659, opened 2020-11-17
Update the ScrollView API usage to avoid deprecated interfaces

jessebarker:
Causes errors due to warnings being treated as errors

### Purpose of this PR
Attempting to update our pinned trunk version causes failures due to usage of a deprecated API (warnings treated as errors).  This updates the ScrollView usage to the newer interfaces.

---
### Testing status
Loading the test project (fails without this change).
Manual editmode test runs on macOS.

---
### Comments to reviewers
Should be straightforward.


jessebarker:
@theo-pnv I think you'll need this for CI to pass on any recent trunk version.

jessebarker:
The above failure is due to the line-endings test, the fix for which landed in a separate PR.  This one is ready to go.  It passes locally on macOS (and has nothing to do with line endings :-) ).

cdxntchou (approved):

PR #2670, opened 2020-11-18
Add auto trigger to formatting job

theopnv:
### Purpose of this PR

This makes the formatting job trigger automatically in open PRs. (This is the server-side verification for code convention compliance)

### DIFF
--- a/.yamato/_formatting.yml
+++ b/.yamato/_formatting.yml
@@ -19,4 +19,4 @@ formatting:
       -  perl ~/unity-meta/Tools/Format/format.pl --hgroot $(pwd) --dry-run .
     timeout: 1
     triggers:
-        expression: pull_request.target eq "master" AND NOT pull_request.draft
+        expression: pull_request.target eq "master"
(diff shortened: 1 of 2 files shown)


.yamato/_formatting.yml
```diff
@@ -18,3 +18,5 @@ formatting:
       -  hg clone -u stable http://hg-mirror-slo.hq.unity3d.com/unity-extra/unity-meta ~/unity-meta
       -  perl ~/unity-meta/Tools/Format/format.pl --hgroot $(pwd) --dry-run .
     timeout: 1
+    triggers:
+        expression: pull_request.target eq "master" AND NOT pull_request.draft
```
sophiaaar, line 22: is there a particular reason we wouldnt want this to run on a draft?

theo-pnv: It was to save bokken resources.
Do you think it would be ok to run it at each commit even draft?

sophiaaar: yeah i think so, especially if we turn on auto-abv again, it might be useful to sort formatting stuff out before triggering abv

theo-pnv: ok, let me remove the draft part then. thanks.

sophiaaar (approved):

PR #2672, opened 2020-11-18
Performance/separate builds

sophiaaar:
In the performance tests, we capture a lot of extra metadata. This metadata is captured in the build job. This means that, for the jobs where we run tests for multiple quality levels, the metadata gets overwritten when the next set of tests starts. This PR splits up the builds so that we do one build for each quality level, resulting in a set of metadata for each quality level.

The reason this is done in one job rather than making a new pipeline for each is to save resources. We dont need to import a fresh project or anything, we just need to do another build. This only adds an extra ~20 minutes to the job time, instead of doubling it (a single build takes 40 minutes)

### Purpose of this PR
* Build 2 players in one job for perf jobs that have tesfilters

---
### Testing status

---
### Comments to reviewers


theo-pnv (approved):
The python will probably need a little bit of refactoring as you mentioned on Slack but otherwise looks good to me 😄

sophiaaar:
cancelled unchanged jobs to save resources

PR #2675, opened 2020-11-18
[ShaderGraph] Fix bottom scrollbar on Graph Inspector not being usable, and fix window resizing issues

Nightmask3:
---
### Purpose of this PR

Bugfix for 1291760: https://fogbugz.unity3d.com/f/cases/1291760/

The fix for the scrollbar issue was trivial, only had to change a value in the .uss file for Resizable.uss and that took care of it.

The main focus of this PR is the host of the window resizing issues that existed prior to this commit (these issues were reported by Felix Herbst, an external partner).

The main change was to make it so that the graph inspector can no longer resize past the edges of the border, and also account for the WIndowDockingLayout behavior that docks graph windows to the sides of the shader graph window when dragged close to them, so that the resizing behavior and the docking behavior are no longer clashing with each other.

Specifically the issues that are fixed by this PR:

- The horizontal scrollbar issue as reported by Landon (and seen in the fogbugz link above)
- Resizing having unexpected behavior, i.e. when the inspector window was docked to the edge of the graph (like when opening the graph for the first time without having any user settings), resizing the right side of the window would resize the left instead, and similar behavior for the top and bottom would occur.
- Being able to resize the inspector window past the edges of the graph view was an oversight I made when implementing it, there is no use case ever where that makes sense, so now the ResizableElement code clamps that. 
- Doing this kind of clamping for the inspector also prevents the issue where the window could be resized past edges of the shader view, and then when clicked on would end up in a weird state where it was half in and half out of the view and the top/sides weren't clickable. (This was due to the interactions with the WindowDockingLayout code trying to clamp the window within the parent view but being unable to do so due to the window being larger than size of the parent view).

---
### Testing status

- Ran test project for shader graph, all checks passed
- Make sure that Sticky Note behavior was unaffected by these changes, created a Sticky Note and resized in all directions and past the edge of the screen view 
- Deleting user settings in the project and opening shader graph to check whether the resizing issue in inspector (i.e. if the right edge was scaled the left edge would move, and similarly for the bottom and top edges)
- Resizing the inspector as far as I could on each side, making sure that it never crossed the edges of the shader graph view.
- Docking the inspector on each side of the graph and repeating the above test to ensure no odd interactions between docking and resizing.

---
### Comments to reviewers

@marctem added a label for a 10.x backport, but also wanted your opinion on whether we should backport to whatever version Felix and his company were at (if thats not 10.x).

Also thinking about how we could automate testing for the resizing behavior stuff and other more specific UI interactions like it, and nothing really occurs to me, but would appreciate any thoughts you had on this Marc, and also @LandonTownsendUnity.

### DIFF
--- a/com.unity.shadergraph/Editor/Resources/Styles/InspectorView.uss
+++ b/com.unity.shadergraph/Editor/Resources/Styles/InspectorView.uss
@@ -30,6 +30,8 @@
     flex-direction: column;
     flex-grow: 1;
     align-items: stretch;
+    padding-bottom: 15px;
+
 }
 
 .InspectorView > .mainContainer > #content > #header {
@@ -39,6 +41,10 @@
     padding: 8px;
     border-top-left-radius: 6px;
     border-top-right-radius: 6px;
+}
+
+.InspectorView.scrollable > .mainContainer > #content > .unity-scroll-view > {
+    padding-bottom: 12;
 }
 
 .InspectorView.scrollable > .mainContainer > #content > .unity-scroll-view > .unity-scroller--horizontal {


Nightmask3:
Another thing this PR does is also move the ElementResizer and ResizableElement code into their own classes as opposed to all being lumped into StickyNote.cs as that code is now used by multiple classes. 

Also introduced a new interface called IResizable to allow the things in the shader graph with resizable behavior to opt in/out of having their resizing behavior being clamped to the parent view. Making a change to the original IResizable class in GraphView would have required a trunk change so this seemed like the easiest way to get there.

com.unity.shadergraph/Editor/Drawing/Interfaces/IResizable.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace UnityEditor.ShaderGraph.Drawing.Interfaces
+{
+    interface IResizable
```
marctem, line 3: I don't really like just using the same name - I know it's in a different namespace but seems unnecessarily confusing. Maybe "ISGResizable" or something? And you can inherit the original interface so it's clear it's an extension and then our ISGResizable classes can also be passed to GV Resizing functions (we may not do that but it's cleaner).

Ah, OK, gross - the GV one was added by VFX team, probably after we made ours here. Well, since we're pulling ours out to its own class, I think still best to inherit from the GV version just to avoid confusing duplication. =)

marctem (changes requested):
Nice job pulling those out to their own classes and good fixes. One naming change and I'm happy!

marctem:
Did you test what happens if you open a project that previously saved the subwindow positions outside the window bounds? What about resizing the graph window itself, do the subwindows move/resize to match?

LandonTownsendUnity (changes requested):
InternalInspectorResizeDifficult
The resize cursor requires a bit TOO much precision now... to the point where if the internal inspector is aligned to the left, right or bottom, the there is no way you can resize via the corner touching the edge of shadergraph, as it will instead resize the shadergraph itself.

LandonTownsendUnity:
While the resize zone now covers only around 2 pixels, the scrollbar takes up a good 7 or 8 pixels where you can click it... I feel like this should be rebalanced
Keep in mind that these UI elements work weirdly and the size of the "zone" you're setting in the USS is not the same size you end up with in the actual UI elements, so make sure to test them to make sure that it's not too hard to get the resize cursor.

LandonTownsendUnity:
InternalInspectorSGResizeBug
This bug still reproduces

Nightmask3:
> InternalInspectorResizeDifficult
> The resize cursor requires a bit TOO much precision now... to the point where if the internal inspector is aligned to the left, right or bottom, the there is no way you can resize via the corner touching the edge of shadergraph, as it will instead resize the shadergraph itself.

Oh that's a good catch, I reduced the size of the right and bottom resizing elements by a couple of pixels each but looks like I went too far, and didn't think of when they'd be docked. Will make them a little thicker and commit that for round 2 in a bit. Thanks Landon!

Nightmask3:
> InternalInspectorSGResizeBug
> This bug still reproduces

Ah good eye, this one wasn't something that I explicitly wanted to address but I can take a look at it now that you bring it up, Marc mentioned testing for this as well.

Nightmask3:
> Did you test what happens if you open a project that previously saved the subwindow positions outside the window bounds? What about resizing the graph window itself, do the subwindows move/resize to match?

All good things to test, and no I have not, so thanks for bringing it up! Landon also touched on the second test case in a comment he left so will make sure to address that in round 2.

com.unity.shadergraph/Editor/Drawing/Interfaces/IResizable.cs
Nightmask3: Oh that's a good idea, honestly didn't even occur to me to inherit from the GV version, will do that and rename ours to `ISGResizable` to avoid naming ambiguity.

com.unity.shadergraph/Editor/Drawing/Views/GraphSubWindow.cs
```diff
@@ -253,16 +264,27 @@ void SerializeLayout()
             EditorUserSettings.SetConfigValue(layoutKey, serializedLayout);
         }
 
-        void OnMoved(MouseUpEvent upEvent)
+        void OnMoveEnd(MouseUpEvent upEvent)
         {
             windowDockingLayout.CalculateDockingCornerAndOffset(layout, graphView.layout);
             windowDockingLayout.ClampToParentWindow();
 
+            // TODO: Dragger does some weird stuff when moving a graph sub window thats larger than screen bounds as it clamps it by offsetting the layout,
```
Nightmask3, line 272: Remove unneeded comments before merging to master

Nightmask3:
> Did you test what happens if you open a project that previously saved the subwindow positions outside the window bounds? 

Tested this, here's a gif of what happens (TL;DR it gets clamped within the window boundaries if scaled past the edges before closing in a previous session and user can then resize it as they need):

7hjkGNhsgD

> What about resizing the graph window itself, do the subwindows move/resize to match?

Accounting for that now in the graph sub windows code, here's what happens, basically matches up with what the preview and blackboard do:

8Hvru4Wtj1

Nightmask3:
> InternalInspectorResizeDifficult
> The resize cursor requires a bit TOO much precision now... to the point where if the internal inspector is aligned to the left, right or bottom, the there is no way you can resize via the corner touching the edge of shadergraph, as it will instead resize the shadergraph itself.

Accounting for this now by having padding added to the scrollers, here's what that looks like:

rsKH4Zuc8A

Feels fairly robust now, let me know what you think. 👍

marctem (approved):
Looks good. I don't love the scrollbar padding but it's fine for now until we get Overlays.

LandonTownsendUnity:
image
The scrollbar overlaps the bottom half of the last entry, making it very difficult to read.

Nightmask3:
> image
> The scrollbar overlaps the bottom half of the last entry, making it very difficult to read.

Addressed that with a quick .uss change to the Inspector. This is how it behaves now. 

4XFLOgKwHM

LandonTownsendUnity (approved):
Did some extra testing with properties, keywords, VFX graph settings. Made sure the bottom scrollbar never blocks the bottom element. I think this is good to release, no usability issues that I could see.

PR #2692, opened 2020-11-19
[HDRP] [Path Tracing] Added proper computation of geometric normal

eturquin:
Currently in ray tracing (and path tracing), the geomNormalWS stored in the bsdfData is just the smooth normal (before bump or normal mapping), rather than the normal of the triangle one would expect.

Some computations in path tracing would benefit from having the proper geometric normal (for example, to shade points that have a front-facing geom normal but back-facing shading normal). Otherwise, you get issues such as the one described here:
https://fogbugz.unity3d.com/f/cases/1293029/

This PR restores the geomNormal as was intended, in path-tracing only. It fixes the bug linked above (the accompanying test scene now renders properly, without black fringes around the silhouettes).


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
```diff
@@ -75,7 +75,10 @@ void ComputeSurfaceScattering(inout PathIntersection pathIntersection : SV_RayPa
 
 #ifndef SHADER_UNLIT
 
-    // Let's compute the world space position (the non-camera relative one if camera relative rendering is enabled)
+    // Override the geometric normal (otherwise, it is merely the non-mapped smooth normal)
+    GetCurrentIntersectionGeometricNormal(attributeData, bsdfData.geomNormalWS);
```
sebastienlagarde, line 79: hmm... there is a problem with this approach. You don't consider anymore the normal provided by the DCC tools or the normal computed by Morten Mikkelsen algorithm's.  What is happening in this case? 
Have you check what is happening with UV mirroring?

eturquin: This code snippet is only used in path tracing, in which context this geomNormal is used exclusively for geometric tests (is the ray shot above or below the triangle's plane). All shading computations are still performed with the shading normal.
This being said, the orientation needed to be checked against the potentially flipped input normal (when dbl-sided); this is now done.

remi-chapelain (commented):
Was about to say something about double sided normal but it has already been fixed. 
However, you cann still repro the issue when double sided is checked on the material 
a0323a61d7f30a5af8102733e71dc911

eturquin:
@remi-chapelain Ok I changed the approach ever so slightly to account for all double-sidedness cases, without having to reproduce the ApplyDoubleSidedFlipOrMirror(...) code completely, or other portions of code that should be in the materials.

remi-chapelain (approved):
All good now ✔️

sebastienlagarde (approved):
nice!

PR #2701, opened 2020-11-20
Upgrade URP materials

pbbastian:
### Purpose of this PR
#2510 added a new material version, which causes all the URP materials to change when opening our test projects. This PR adds those updated materials.

---
### Testing status
Opened projects, closed them. Will run Yamato.

---
### Comments to reviewers
Change log not updated as this is only changing test projects. Backport doesn't have to rush into a specific release, since there's nothing actually going into the package, but is more for our convenience.


ellioman (approved):

ellioman:
Sweet! Thanks for this!

marctem (approved):

pbbastian:
Formatting is only failing on known Shader Graph files

pbbastian:
Removing the backport label, as a direct backport would not be possible. If we think it's needed, we can upgrade the material versions on older branches by opening the projects.

PR #2704, opened 2020-11-20
Add new eye scene in material samples

remi-chapelain:
### Purpose of this PR
This PR adds a proper example of use for the SG_Eye graph [recently added](https://github.com/Unity-Technologies/Graphics/pull/2431) in the HDRP material samples.

image

---
### Testing status
Tried importing the sample from the branch to make sure that everything is properly imported. 
Size of Samples before 58.26mb
Size now :69.27mb

---
### Comments to reviewers
Still in draft, 
Still needs improvement regarding the disposition, number of prefabs and adding a text that would explain the specifics of this shader (the same way fabric scene is doing for example)

image

### DIFF
--- a/com.unity.render-pipelines.high-definition/Samples~/MaterialSamples/Textures/Iris_N.tga.meta
+++ b/com.unity.render-pipelines.high-definition/Samples~/MaterialSamples/Textures/Iris_N.tga.meta
@@ -7,7 +7,7 @@ TextureImporter:
   mipmaps:
     mipMapMode: 0
     enableMipMap: 1
-    sRGBTexture: 1
+    sRGBTexture: 0
     linearTexture: 0
     fadeOut: 0
     borderMipMap: 0
@@ -52,7 +52,7 @@ TextureImporter:
   alphaUsage: 1
   alphaIsTransparency: 0
   spriteTessellationDetail: -1
-  textureType: 0
+  textureType: 1
   textureShape: 1
   singleChannelComponent: 0
   flipbookRows: 1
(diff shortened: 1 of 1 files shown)


remi-chapelain:
Reduced samples from 10 to 4, reducing the texture footprint. 
Also modified the texture naming convention and format to match what's already in the samples

iM0ve (commented):
Review: **In Progress**
Looks good but have a few suggestions.

1) Add OnEnable reflection probe. Makes them pop better. Removes unwanted highlights
wXrnjEPhwf

SeanPuller:
Great addition to the material samples!

1. iris offset is 0.02 in every material although it shows the edge of the texture. 0 might be a better parameter
Unity_espRghWb0c

2. Diffusion profiles are included but not assigned in the materials

3.iris color textures could be 512x512 without loosing quality. Others are fine

remi-chapelain:
- Reduced iris texture size from 1k to 512px (saved another 5mb)
- Added onEnable probe
- Diffusion profile are now assigned in material and not hardcoded in shader graph
- Corrected the iris offset to 0 in every material 
- Made the eyes hover a bit to avoid clipping on the floor

image

sebastienlagarde:
@remi-chapelain . We were discussing with @SeanPuller  about merging all the sample Material in a single scene. Guess we should start :), you should add the iris material inside the main Material scene we already have. @SeanPuller  any thought?

SeanPuller:
I'm not sure about one scene. I think having them in separate scenes is cleaner and scales better. Having them all in the same "importable sample" (or whatever the term is) is the most important which we already have.

iM0ve (approved):
All issues fixed. Will be a great addition to the samples

sebastienlagarde (approved):

sebastienlagarde:
@remi-chapelain : I forget to ask you to update the documentation :) https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.2/manual/HDRP-Sample-Content.html

PR #2706, opened 2020-11-21
Fixed NaNs happening when upscaling ray tracing reflections in performance mode (case 1294076).

anisunity:
I am not sure how this is a regression, but has been reported as one in the case 1294076. This PR fixes the issue. This fix has already been done on master. So no need to forward port it.
https://fogbugz.unity3d.com/f/cases/1294076/

Testing status:
Tested to reproduce the nans before and after the fix. The nans do not happen anymore with the fix


sebastienlagarde (approved):

remi-chapelain (approved):
Fixed ✔️ 
Tested again in perf and quality mode, with scene and game view opened, could not repro.

PR #2711, opened 2020-11-23
Cloud System and Cloud Layer

adrien-de-tocqueville:
### Purpose of this PR

This PR introduces a cloud API to HDRP that can be used to render clouds between the sky and the fog.
The clouds are baked in the sky reflection probe and are included in the static lighting skies

It uses the volume system in a similar way to the Sky System and is acessible through the Visual Environnement override
VisualEnvironment

The PR also includes a type of clouds called CloudLayer which renders a cloud texture on top of the sky with options for a simple lighting simulation and for shadow casting by the sun light.

CloudLayer_UI

Default cloud map example (with lighting option on the right)
Lighting

Shadow casting on the template
shadows-ld

---
### Testing status

- Launched yamato
- Added a graphics test (5010)
- Tested with a few different cloud textures
- Tested static sky lighting

---
### Comments to reviewers

- Shadow texture is rendered as a cookie, which means rotating the sun around it's direction axis will rotate the shadow texture.
- Also should the shadow texture be put in the cookie atlas ?
- The lighting parameters don't seem very intuitive (step count / thickness)


RemyUnity:
Thanks Adrien for the great description !

Shouldn't the new "nested parameters" have their own dedicated PR ?

Also, what do you think that QA should look at for reviewing ?

AlixMi:
Hello, great improvements, the lighting is promising for a non volumetric solution :) 
My feedback :
* this would be nice to have the DefaultCloudMap already populating the Cloud Map parameter
* I think it would be better if we wouldn't have to copy the exposure value inside, when creating the clouds they appears black with a daylight scene for example before we rise the exposure. It would be nice if the clouds would already fetch the intensity of the sun so that they are bright enough and working with the overall exposure.
* When lighting is enabled, it would be nice to be able to tint shadows
* The blend between layers seems to be a max (that comes from me sorry) I think we should choose another type of blend so that it doesn't eat the shapes of cloud when layering them
* Have a contrast parameter on the shadows casted on the scene
* Have a speed parameter on the shadow casted on the scene when there is distortion enabled, because depending on the chosen tiling, it can be too fast. (Or make it work depending on the tiling so that it has a good parallel with the sky movement)
When cast shadows is disabled in the lighting tab, the sun is no longer taken into account for the lighting of clouds (last position storred)
* sometimes the scroll direction of the casted shadows seems to differ from what we can see around the sun

adrien-de-tocqueville:
> Shouldn't the new "nested parameters" have their own dedicated PR ?

#2745 
 
> Also, what do you think that QA should look at for reviewing ?

Maybe some testing on the static lighting sky or baked lighting, I am not 100% sure I tested that correclty.

iM0ve (changes requested):
Review status: **Done**
Last updated: 2020/12/14

**WHATS TESTED**
* Windows and mac
* Usability and default values
* Typical use cases
* Checked that clouds are included in the bake
* Influence on new project (HDRP defaults)
* Tooltips
* Automation - Adrien added graphic tests

**1) FIXED** As Alix mentioned, Cloud map should be already assigned as the default value, because:
* Without it, when you add the override it does nothing.
* Without it, if you search for cloud in the texture search, you will not find anything because it doesnt search in packages folder. So already need a lot of knowledge jus to try out the feature.

I would consider this an UX bug.

ux


**2) FIXED** Cloud Map has a default value but it doesnt work until the property is enabled.

a6Gxo9RJm9


**3) FIXED** Maybe worth specifying that the exposure is using Directional light. Not entirely clear what is it relative to

image


**4) FIXED** Could we make the tooltip a little more beginner-friendly, also does it simulate directional light absorption or more complex?

image (33)


**5) FIXED** Changing settings produce blinking when shadows are on.

7kU4EAEu5U


**6) FIXED** When shadow is enabled. The tiling setting becomes unhidden on Directional light, this is a bit hard to find. We currently have an info box for it. But maybe its possible to also dislplay these values on the Volume override too? Even if grayed out

image


**7) FIXED** There is a very noticeable seam when shadows are enabled. Is there something we can improve?

Screenshot 2020-12-03 at 15 20 39


**8) FIXED** With the default multiplier shadows do not match what you would expect to see from a cloudy sky. Making it unintuitive to use, we should tweak the default value or mutliplier code if we can.

shadows

**9) FIXED** We might want to remove the Cloud Layer component from the HDRP default settings since it no longer serves its purpose

image

adrien-de-tocqueville:
> **1)** As Alix mentioned, Cloud map should be already assigned as the default value

I agree but it's not possible to have it as a default value. I can propose a workaround:
Define a CloudLayer override in the HDRP default settings volume with the texture set to DefaultCloudMap
Copy the cloudmap in the project folder like we do for HDRI sky ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
```diff
@@ -295,6 +295,12 @@ SHADOW_TYPE EvaluateShadow_Directional( LightLoopContext lightLoopContext, Posit
         g_DebugShadowAttenuation = shadow;
 #endif
 
+    // Cloud shadows
```
JulienIgnace-Unity, line 298: I'm not sure I understand why we need specific code for this.
Directional lights already support cookies. Seems redundant to me. Also what happens if a directional light has a cookie?

adrien-de-tocqueville: so that's why it's a separate texture than the cookie, so that it's possible to use it together with a cookie

JulienIgnace-Unity: Was this requested by artist? I assume that the vast majority of cases for directional cookie is to actually fake clouds. Would that make sense to make them mutually exclusive and save the extra sample?

com.unity.render-pipelines.high-definition/Runtime/Sky/SkyManager.cs
```diff
@@ -744,7 +816,10 @@ internal void RequestStaticEnvironmentUpdate()
                 var renderingContext = cachedContext.renderingContext;
 
                 if (IsCachedContextValid(skyContext))
+                {
                     forceUpdate |= skyContext.skyRenderer.DoUpdate(m_BuiltinParameters);
+                    forceUpdate |= (skyContext.HasClouds() && skyContext.cloudRenderer.DoUpdate(m_BuiltinParameters));
```
JulienIgnace-Unity, line 821: Just to make sure I get this right, this will force an update only if parameters of the cloud layer used in the "GeBakingHashCode" function change right? But not all the parameters are in this function right? 
What about the other parameters? It seems to me that the GetHashCode of the cloud settings should also be incorporated in the hash that we compute for the sky and use the existing control flow. This way if any of the parameters other than the "baking" ones that could change the look of the cloud will also trigger a recompute if the sky is set on "OnChanged"

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
adrien-de-tocqueville: I guess you are right, this wasn't a request so let's do it directly with the cookie

com.unity.render-pipelines.high-definition/Runtime/Sky/SkyManager.cs
adrien-de-tocqueville: i included the GetHashCode of the clouds inside the hash of the sky (above in ComputeSkyHash), that way the code doesn't need much change and the update mode is controlled like before from the sky dropdown

JulienIgnace-Unity: Oh yeah missed it because I was looking at ComputeCloudHash that does not seem to be used at all. Might want to remove it if it's not necessary.
Also I'm still concerned that GetHashCode is not overridden at all in CloudLayer.cs so an "OnChanged" sky will probably not update cubemap correctly when changing the cloud layer parameters.

adrien-de-tocqueville: Nice catch for ComputeCloudHash.
The update is working when using OnChanged, because the volume component default GetHashCode function computes the hash of all parameters combined. It's easier when iterating but now I can implement it properly

JulienIgnace-Unity: Ah yeah forgot about that, thanks :)
It could be better to make it explicit indeed because sometimes we don't want ALL parameters to actually participate in the hash .

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
JulienIgnace-Unity: Maybe we can involve the artists with this one?
They might have another opinion :)
(even though I suppose all artists always want more :) )
Also we need to make sure that if we choose this path, we make it clear that enabling cloud layers overrides the regular directional light cookies.

JulienIgnace-Unity: @sebastienlagarde ^^ ?

JordanL8 (approved):

JordanL8 (approved):

iM0ve:
Updated the review above

iM0ve (approved):
All issues were resolved. For test details (doc) see above comment.

**In conclusion**: A very nice addition. Allows you with barely any knowledge to add nice looking clouds in the scene, but also leaves a lot of space for further customisation. 

clouds

com.unity.render-pipelines.high-definition/Runtime/Sky/SkyUtils.hlsl
```diff
@@ -17,12 +17,18 @@ float3 GetSkyViewDirWS(float2 positionCS)
     return normalize(viewDirWS.xyz);
 }
 
+float FastAtan2(float y, float x)
```
sebastienlagarde, line 20: We should move this function in common.hlsl (can be done in another PR)

sebastienlagarde (approved):

sebastienlagarde (approved):

PR #2719, opened 2020-11-23
Universal/runtime render pass

Nzollner:
# Status
This is a Draft and is currently still missing documentation.

# Purpose of this PR
This PR fixes the render passes not being able to be enqueued at runtime.
Which is requested by this bug case: https://fogbugz.unity3d.com/f/cases/1211195/
And to show the feature working is a sample implemented which also helps the user.

# Testing status
I tested this PR locally on Windows 10 using DX11 on UniversalGraphicsTest.
I have also tested the feature manually when creating the sample using the fix.

# Comments to reviewers
Please check the sample and come with feedback making the sample better for the user.

### DIFF
--- a/com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
+++ b/com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
@@ -34,7 +34,6 @@ public class RunPassAtRuntime : MonoBehaviour
                 RenderPipelineManager.beginCameraRendering -= EnqueRenderPass;
 
             refactiveGlass.SetFloat("_UseRenderPass", useRenderPass ? 1.0f : 0.0f);
-
         }
     }
 }
--- a/com.unity.render-pipelines.universal/Samples~/RenderPassSamples/Shaders/KawaseBlur.shader
+++ b/com.unity.render-pipelines.universal/Samples~/RenderPassSamples/Shaders/KawaseBlur.shader
@@ -17,7 +17,7 @@ Shader "Custom/RenderFeature/KawaseBlur"
             #pragma vertex vert
             #pragma fragment frag
 
-            
+
             struct Attributes
             {
                 float4 positionCS : POSITION;


ellioman:
I removed the `needs-backport-8.x` tag as we won't backport more PR's to that version.
Add the tag for 7.x if you think we should backport this fix for LTS.

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
```diff
@@ -0,0 +1,39 @@
...
+        renderPass.renderPassEvent = setting.renderPassEvent;
+        if(Input.GetKeyDown(KeyCode.C))
+        {
+            useRenderPass = !useRenderPass;
+            refactiveGlass.SetFloat("IsUsingRenderFeature", useRenderPass ? 1.0f : 0.0f);
+
+        }
+        if(useRenderPass)
+        {
+            ((UniversalRenderPipelineAsset)GraphicsSettings.currentRenderPipeline).scriptableRenderer.EnqueuePass(renderPass);
```
phi-lira, line 36: The camera can override the renderer. So we need to get the renderer from the camera. 

There are multiple ways to do it:
1) you can get a reference to the camera. Then call `camera.GetUniversalAdditionalCameraData().scriptableRenderer`
2) you can use beginCameraRendering event https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@10.2/manual/using-begincamerarendering.html to get the camera

To make it easier to users we could actually even add a GetRenderer() from the camera
so `camera.GetUniversalAdditionalCameraData().scriptableRenderer` could be just camera.GetScriptableRenderer()`

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
```diff
@@ -0,0 +1,39 @@
+using System.Collections;
+using System.Collections.Generic;
+using TMPro;
```
phi-lira, line 3: Is this necessary? Collections as well?

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
```diff
@@ -0,0 +1,39 @@
...
+public class RunPassAtRuntime : MonoBehaviour
+{
+    public KawaseBlurSettings setting;
+    KawaseBlur renderPass;
+
+    bool useRenderPass = false;
+
+    Material refactiveGlass;
+
+    // Start is called before the first frame update
```
phi-lira, line 17: Comment is not necessary.

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
```diff
@@ -0,0 +1,39 @@
...
+    void Start()
+    {
+        refactiveGlass = GetComponent<Renderer>().material;
+        renderPass = new KawaseBlur("KawaseBlur", setting);
+    }
+
+    // Update is called once per frame
```
phi-lira, line 24: Also no necessary

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/Shaders/KawaseBlur.shader
```diff
@@ -0,0 +1,70 @@
...
+        Tags { "RenderType" = "Opaque" }
+        LOD 100
+
+        Pass
+        {
+            CGPROGRAM
+            #pragma vertex vert
+            #pragma fragment frag
+            // make fog work
+            #pragma multi_compile_fog
```
phi-lira, line 20: Why fog is needed?

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/Shaders/KawaseBlur.shader
```diff
@@ -0,0 +1,70 @@
...
+
+            #include "UnityCG.cginc"
```
phi-lira, line 22: Needs to be written in SRP shader library / URP. This is using old built-in shader library.

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/KawaseBlur.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        // first pass
+        // cmd.GetTemporaryRT(tmpId1, opaqueDesc, FilterMode.Bilinear);
+        cmd.SetGlobalFloat("_offset", 1.5f);
+        cmd.Blit(cameraColorTexture, tmpRT1, settings.blurMaterial);
+
+        for (var i = 1; i < settings.blurPasses - 1; i++)
+        {
+            cmd.SetGlobalFloat("_offset", 0.5f + i);
+            cmd.Blit(tmpRT1, tmpRT2, settings.blurMaterial);
```
phi-lira, line 83: This won't work on XR. @thomas-zeng can you help?

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/Shaders/KawaseBlur.shader
phi-lira: Shader also needs to be stereo aware. @thomas-zeng

phi-lira (changes requested):
Also missing changelog.

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
Nzollner: Should I create the shortcut in this PR?

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
Nzollner: is removed in next commit

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/RunPassAtRuntime.cs
Nzollner: This is included in the next commit
But docs still missing for it

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/Shaders/KawaseBlur.shader
Nzollner: This is removed in the next commit

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/KawaseBlur.cs
thomas-zeng: XR prefers to use DrawMesh will fullscreen quad to do blit.
cmd.Blit in some cases will use built-in blit shader which doesn't work well with URP. 
With DrawMesh, we can explicit set the blit shader. See how URP SSAO handles blit here: 
https://github.com/Unity-Technologies/Graphics/blob/1d37e201e0353347878642680c61e7012b156e8f/com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs#L289

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/Shaders/KawaseBlur.shader
Nzollner: The updated shader should now use the universal shader library and is now using HLSL instead of cginc.

com.unity.render-pipelines.universal/Samples~/RenderPassSamples/BlurryRefraction/KawaseBlur.cs
Nzollner: The result is used in Shader Graph is it possible to use XR textures in shader graph?

Jonasmortensen (approved):
As more package samples are added to URP there needs to be a common template that makes it easy for users to investigate and navigate the samples. With this being the first sample landing in URP this is not an issue at the moment but it might need to be revisited in the near future.

Nzollner:
XR for MockHMD: Single pass instanced is not working but if instanced is turned off the Kawase blur will the correct result return on the left lens.

phi-lira:
Closing PR. Can be reopened if necessary.

PR #2722, opened 2020-11-23
Added a slider to control the fallback value of the directional shadow when the cascade have no coverage.

anisunity:
711_DirectionalShadowFallback

Previously, when shading a point on a ray tracing intersection. If the directional light was involved, the shadow value that was used if that point was outside of the cascade's coverage was 1.0, which lead to leaks in certain configurations (interior rooms with directional lights outside). This PR adds a slider on the ray tracing settings volume component to override that value locally.
image
This makes ray tracing much more usable when using a directional light.

Testing status
Added a new test that checks the new behavior.
Ran the DXR tests locally, all green.
Pair tested with @remi-chapelain and @pierre-unity

### DIFF
--- a/TestProjects/HDRP_DXR_Tests/Assets/Scenes/711_DirectionalShadowFallback.unity
+++ b/TestProjects/HDRP_DXR_Tests/Assets/Scenes/711_DirectionalShadowFallback.unity
@@ -166,6 +166,7 @@ MeshRenderer:
   m_CastShadows: 1
   m_ReceiveShadows: 1
   m_DynamicOccludee: 1
+  m_StaticShadowCaster: 0
   m_MotionVectors: 1
   m_LightProbeUsage: 1
   m_ReflectionProbeUsage: 1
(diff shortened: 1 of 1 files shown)


remi-chapelain (approved):
Made a commit to change the size of the reference image to avoid wasted space horizontally. (cf below). 
Tested that the test pass locally
Test and feature have already been pair-tested. ✔️ 

711_DirectionalShadowFallback

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAlgorithms.hlsl
```diff
@@ -274,12 +274,12 @@ float EvalShadow_CascadedDepth_Blend(HDShadowContext shadowContext, Texture2D te
 }
 
 
-float EvalShadow_CascadedDepth_Dither(HDShadowContext shadowContext, Texture2D tex, SamplerComparisonState samp, float2 positionSS, float3 positionWS, float3 normalWS, int index, float3 L)
+float EvalShadow_CascadedDepth_Dither(HDShadowContext shadowContext, Texture2D tex, SamplerComparisonState samp, float2 positionSS, float3 positionWS, float3 normalWS, int index, float3 L, out int shadowSplitIndex)
```
sebastienlagarde, line 277: you must not break the prototype of EvalShadow_CascadedDepth_Dither as users can use it in their shader, you need to introduce a new one

float EvalShadow_CascadedDepth_Dither(HDShadowContext shadowContext, Texture2D tex, SamplerComparisonState samp, float2 positionSS, float3 positionWS, float3 normalWS, int index, float3 L)
{
unused shadowSplitIndex
 return EvalShadow_GetSplitIndex(shadowContext, index, positionWS, alpha, cascadeCount, shadowSplitIndex);
}

float EvalShadow_CascadedDepth_Dither(HDShadowContext shadowContext, Texture2D tex, SamplerComparisonState samp, float2 positionSS, float3 positionWS, float3 normalWS, int index, float3 L, out int shadowSplitIndex)

or something like that

sebastienlagarde: also you must do the same thing for EvalShadow_CascadedDepth_Blend

sebastienlagarde:
Looks like thee fix don't work in XR: https://yamato-artifactviewer.prd.cds.internal.unity3d.com/139e1a00-74ab-4cfc-a445-261bd8f1b7b3%2FTest_report%2FUtrTestReport1%2FTestProjects%2FHDRP_DXR_Tests%2Ftest-results/TestReportV1.html

anisunity:
Did the required changes and the XR issue has been fixed by @fabien-unity in a different PR

sebastienlagarde (approved):

PR #2726, opened 2020-11-23
[11.x.x] Fixing 119_CameraToRTWithViewportRect scene for consoles

ellioman:
# Purpose of this PR
Test scene 119 is failing on consoles as the target texture is not properly cleared before rendering to it. 
This PR fixes that by specifically clearing it before we render to it.

# Yamato


ellioman:
The HDRP failures are known as is the formatting one. None are caused by this PR.

pbbastian (approved):

PR #2498, opened 2020-11-02
[HDRP][Docs] Update documentation for alpha output

pmavridis:
Small update for the alpha output docs (based on recent user questions in internal Slack)

PR #2736, opened 2020-11-24
[HDRP] Fix alpha output when atmospheric scattering is used

pmavridis:
### Purpose of this PR
The atmospheric scattering pass was overwriting the alpha channel, thus output alpha was not the opacity of objects (as expected by many compositor/verticals applications). With this PR the  atmospheric scattering shader is configured to not touch the alpha channel. 

Test scene with atmospheric scattering enabled (RGB channel):
image

Alpha channel of the output frame, before:
image

Alpha channel of the output frame, after:
image

I have also updated the alpha output docs based on feedback from @JordanL8 (it was from this PR: https://github.com/Unity-Technologies/Graphics/pull/2498)

---
### Testing status
- Tests are green (on DX11 run locally)
- The alpha output is not affected by atmospheric scattering, as shown in the test scene. 

I don't think any further testing is necessary, but I have added the QA team just in case (and for visibility)


RemyUnity:
Thanks you for the nice description Pavlos.

On QA side, do you think we should check (if you didn't) these (on top of my head) :
- If the compositor does properly use this alpha
- If we can output image with proper transparency with recorder
- Have the correct alpha if targeting to a rendertexture with a camera

Anything else ?

pmavridis:
Hey Remy, this is a good list of things to test. 
My tests (shown in the screenshots) were actually done with the compositor (by connecting the alpha channel in the output of the composition graph to display the alpha in the gameview), but I guess it still makes sense to do some extra checks.

Another thing to check is perhaps if MSAA breaks anything.

iM0ve (approved):
Thank you for the details. I think for the scope of this change the testing is sufficient.

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/OpaqueAtmosphericScattering.shader
```diff
@@ -120,6 +120,7 @@ Shader "Hidden/HDRP/OpaqueAtmosphericScattering"
             Cull Off    ZWrite Off
             Blend One SrcAlpha // Premultiplied alpha
             ZTest Less  // Required for XR occlusion mesh optimization
+            ColorMask RGB 0 // Only write on the RGB channels for RT 0 (required to preserve alpha)
```
sebastienlagarde, line 123: instead of not writing in alpha channel, you should say what to do with alpha in the blend mode

Blend SrcFactor DstFactor, SrcFactorA DstFactorA: Same as above, but use different factors for blending the alpha channel.

https://docs.unity3d.com/Manual/SL-Blend.html

i.e in the blend mode of alpha just do nothing

pmavridis: Done! 

The code path for PBR fog is always disabled for now (and this one is not using any blending), so I have tested it by manually enabling it it the code (and it was working in both R11G11B10 and 16-bit modes) .

sebastienlagarde (approved):

JulienIgnace-Unity (approved):

PR #2739, opened 2020-11-24
Remove non rendergraph code path

JulienIgnace-Unity:
### Purpose of this PR
This PR removes all the non-rendergraph code path from HDRP.
- Remove toggle in the HDRP default settings
- Remove toggle in the debug window
- Remove non render graph yamato tests

---
### Testing status
Given the size of the PR I suppose we have to trust automated testing.


RemyUnity:
Regarding the black screen issue when building a played in release mode when rendergraph was defined as default, I think this should be a manual test to do ?
Nothing else that might come up in your mind ?

TomasKiniulis (approved):
Nothing appears to be affected

Checked HDRP template with:
- Debug Modes
- Compositor
- HDRP Asset
- Volume Settings
- Frame Setting Overrides
- Player Build

theo-pnv (approved):
Approving the yamato part on behalf of SDETs but it looks like many HDRP tests are failing. (I don't know if they are on master, as I'm still out of date on what works and what doesn't on master).

sebastienlagarde (approved):

PR #2743, opened 2020-11-25
[URP 2D] Added dummy XML doc comments to all public APIs.

rainsing:
### Purpose of this PR
Added dummy XML doc comments to all 2D public APIs. This is done solely for the purpose of passing the package validation. We'll do real documentation at a later time, when we move out of Experimental namespace.

---
### Testing status
NA

---
### Comments to reviewers
NA


UnityPaul (approved):
To be improved right after cut off

rainsing:
As the verification won't fail until 2021.2, this is not needed anymore.

PR #2745, opened 2020-11-25
Nested volume components

adrien-de-tocqueville:
### Purpose of this PR

The PR add the possibility for volume components to have nested parameters
For example:
```cs
public class Component: VolumeComponent
{
    // This doesn't change
    public ClampedFloatParameter a = new ClampedFloatParameter(0.0f, 0.0f, 1.0f);

    // This is now possible
    [Serializable]
    public class NestedClass
    {
        public ClampedFloatParameter b = new ClampedFloatParameter(0.0f, 0.0f, 1.0f);
    }

    public NestedClass nested = new NestedClass();
}
```

It also adds a Init callback on the volume components to initialize static resources, for example to have a default value on a texture parameter
```cs
public class ExampleComponent : VolumeComponent
{
    static Texture s_DefaultTexture = null;
    public TextureParameter texture = new TextureParameter(s_DefaultTexture);
    static void Init()
    {
        s_DefaultTexture = //...
    }
}
```

This features are used by the [cloud layer](https://github.com/Unity-Technologies/Graphics/pull/2711)

---
### Testing status
Tested for the cloud layer
Tested that the volume debug menu takes correctly into account nested parameters.


sebastienlagarde (approved):

sebastienlagarde (commented):
Please add it to what's new in 11 page

JulienIgnace-Unity (approved):
LGTM. Might be nice to have @Chman opinion on this too.

Chman (approved):
LGTM. I'm not too fond of the magic `Init` method but I can't think of a better way to do it right now without adding more work on the user side.

PR #2748, opened 2020-11-25
Added an emissive as forward pass when the lit is in deferred mode v2

sebastienlagarde:
### Purpose of this PR
- based on https://github.com/Unity-Technologies/Graphics/pull/2527

In order to properly support APV and SSGI(RTGI). We need to be able to not impact the emissive when in deferred mode. To do so, we added the option (Force Forward) to have the emissive rendered in a separate forward pass (ForwardEmissiveForDeferred) only for object that have non null (default value SG or shader) or with a branch plugged in (SG).

- This have effect only possible when the lit model is in deferred
- The new  pass (ForwardEmissiveForDeferred) is not generated for shader graph if the emissive value is black or nothing is plugged in the master node on the emission port.
- The pass is disabled on non shader graph shader for material that have null emissive values.

The extra pass is stripped out in the player if the Lit mode is forward

Testing status
Added a Yamato test 2011 covering all usage types.
image

- Tested UX of shader graph / Material associated. Checked that the checkbox correctly display
- Checked the generated code of shader graph to see that the code is correctly removed
- Checked in RenderDoc that the material render in the correct pass.
- Checking the emissive override mode (SG Unlit don't work but it is not related to this PR), and emissive only lighting mode.

For QA:
This PR introduce a new pass to render Emissive Color in a new pass - ForwardEmissiveForDeferred - but only for Lit (SG Lit, LayeredLit, LitTessellation, LayeredLitTessellation, Lit) Material with a Surface Type Opaque.
The Emissive contribution of a Material will be rendering into the Lighting Buffer during the GBuffer pass or in an extra Emissive Color pass which is after all the other lighting based on the "Force Forward Emissive" checkbox on Material or SG.
This change of rendering order only apply for object render in the deferred pass; Transparent Material or other Material (hair, stacklit, fabric, eye, axf, unlit) aren't affected.

Plan test:

When enabling "Force Forward Emissive" in a Material with Lit emissive objects, we should get the same visual result. However there is a difference, now the SSAO algorithm will not affect the emissive and in addition, if two emissive object are on top of each other, they will contribute twice (against one without the change). The new pass can be check in a RenderDoc capture or the FrameDebugger. If Lit Material don't have a Emissive contribution, then nothing should be render in this extra pass.
It can be validated that it work correctly with the SSGI effect. SSGI effect by default remove emissive contribution of object. 

It need also to be check in a player.

---
### Testing status

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDTarget.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDTarget.cs
@@ -1365,6 +1365,7 @@ namespace UnityEditor.Rendering.HighDefinition.ShaderGraph
             type = KeywordType.Boolean,
             definition = KeywordDefinition.ShaderFeature,
             scope = KeywordScope.Global,
+            stages = KeywordShaderStage.Fragment,
         };
 
         public static KeywordDescriptor TransparentWritesMotionVector = new KeywordDescriptor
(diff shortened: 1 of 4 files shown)


sebastienlagarde:
replace by https://github.com/Unity-Technologies/Graphics/pull/3430

PR #2750, opened 2020-11-25
[HDRP] Updated what's new in HDRP 11

alelievr:
### Purpose of this PR
Update What's new with cubemap field change in volumes

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/whats-new-11.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/whats-new-11.md
@@ -16,7 +16,7 @@ For more information about the future, see the [Shadow](Shadows-in-HDRP.md) sect
 
 ### Cubemap fields in Volume components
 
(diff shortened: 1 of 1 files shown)


JordanL8 (approved):
Rephrased some bits, can you double-check the first sentence "Cubemap fields now... etc" to make sure I understood it correctly before merging?

sebastienlagarde (approved):

sebastienlagarde:
>Cubemap fields now accept both [RenderTextures](https://docs.unity3d.com/Manual/class-RenderTexture.html) and >[CustomRenderTextures](https://docs.unity3d.com/Manual/class-CustomRenderTexture.html) if they are formatted like Cubemaps. >This change affects the `HDRI Sky` and `Physically Based Sky` components and allows you to animate both skies.

the " if they are formatted like Cubemaps." is still weird. @alelievr what do you suggets?

alelievr:
Hum, " if they are using the cubmap mode / dimension" would be better I think (dimension is the correct renderTexture property name)

sebastienlagarde:
> Hum, " if they are using the cubmap mode / dimension" would be better I think (dimension is the correct renderTexture property name)

@JordanL8  ?

JordanL8:
Ah sorry I didn't get any notifications for this. Yeah if cubemap mode / dimension is more correct then sounds good :)

JordanL8:
I'll update this

JordanL8:
All good?

alelievr:
Yeah :)

sebastienlagarde (approved):

PR #2752, opened 2020-11-25
Normalize view direction in universal

elizabeth-legros:
Needed to have the view direction node return a normalized value in Shader Graph. Created templates in the Universal package to achieve this. 

Ran all Shader Graph tests locally and verified normalized view direction in editor


cdxntchou (approved):

phi-lira (changes requested):
Missing changelog. Otherwise LGTM.

LandonTownsendUnity (changes requested):
I would like documentation updates both in the View Direction node documentation and the upgrade guide.

com.unity.shadergraph/Documentation~/View-Direction-Node.md
```diff
@@ -4,6 +4,8 @@
 
 Provides access to the mesh vertex or fragment's **View Direction** vector. This is the vector from the vertex or fragment to the camera. The coordinate space of the output value can be selected with the **Space** dropdown parameter.
 
+NOTE: In versions prior to 11.0, the **View Direction** vector was not normalized in the **Universal Render Pipeline**. Version 11.0 changed this behavior, and this vector is now normalized in both the **High-Definition Render Pipeline** and the **Universal Render Pipeline**. To mimic old behavior, you can use the [Position Node](Position-Node.md) in **World** space and subtract the **Position** output of the [Camera Node](Camera-Node.md).
```
cdxntchou, line 7: Do you know if the camera position output in (local) World space in HDRP?  (i.e. 0,0,0) ?

com.unity.shadergraph/Editor/Data/Nodes/CodeFunctionNode.cs
```diff
@@ -277,65 +277,65 @@ public sealed override void UpdateNodeAfterDeserialization()
             {
                 AddSlot(slot);
             }
-            RemoveSlotsNameNotMatching(slots.Select(x => x.id));
+            RemoveSlotsNameNotMatching(slots.Select(x => x.id), true);
         }
 
-        private static MaterialSlot CreateBoundSlot(Binding attributeBinding, int slotId, string displayName, string shaderOutputName, ShaderStageCapability shaderStageCapability, bool hidden)
```
cdxntchou, line 283: Is this change related?  Or just setting up nice defaults?

cdxntchou (approved):

com.unity.shadergraph/Editor/Data/Nodes/CodeFunctionNode.cs
elizabeth-legros: change related because it wasnt getting passed in before (which would default to false when making the slots) so I just wanted to keep it as consistent as possible

com.unity.shadergraph/Documentation~/View-Direction-Node.md
elizabeth-legros: Yeah, that was something i wan going to have to double check in HDRP. I think you are right tho, and will address it

elizabeth-legros: This wording is getting changed tho, since weve added the new node

phi-lira (approved):

phi-lira:
@LandonTownsendUnity the feedback was addressed can you look at this PR again?

elizabeth-legros:
> @LandonTownsendUnity the feedback was addressed can you look at this PR again?

We are also waiting on updated Documentation for the node before we can merge

LandonTownsendUnity (changes requested):
I'm OK with the "view vector" object mode behavior being different because the old view direction behavior was normalized and that doesn't make sense for a "view vector" node. However it should be stated in the upgrade guide/documentation that you shouldn't swap out the node if it's in object space.

The behavior of "View Vector" is different from the old behavior of View Direction when in orthographic camera mode, in all spaces. This needs to be fixed. Even aside from upgrade problems, "view vector" being the vector from the position to the camera doesn't make sense in orthographic mode.

LandonTownsendUnity (changes requested):
View Vector needs to be fixed in HDRP - currently uses camera relative World transformations instead of Absolute World

Edit: actually hold on, let me check that

LandonTownsendUnity (commented):
Changelog also needs to be updated with the fact that the View Vector node was added.

LandonTownsendUnity (approved):
Now that the changelog is updated, I will approve.

cdxntchou (approved):

PR #2753, opened 2020-11-26
Removed the upper clamp values for recursive rendering and light cluster size (case 1294620)

anisunity:
- Removed the superior clamping value for the recursive rendering max ray length.
- Removed the superior clamping value for the ray tracing light cluster size.

Testing status
- Created a new recusive rendering volume, mades sure the field behaved as expected
- Created a new light cluster volume, made sure the field behaved as expected
- Ran the DXR tests locally, were all green.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Raytracing/RecursiveRenderingEditor.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Raytracing/RecursiveRenderingEditor.cs
@@ -1,7 +1,7 @@
+using UnityEngine;
 using UnityEditor.Rendering;
 using UnityEditor.Rendering.HighDefinition;
 using UnityEngine.Rendering.HighDefinition;
-using UnityEngine.Rendering;
 
 namespace UnityEditor.Experimental.Rendering.HighDefinition
 {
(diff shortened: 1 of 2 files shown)


remi-chapelain (approved):
Auto approoving this one, this seems simple enough to not have a second look at it ✔️

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/LightCluster.cs
```diff
@@ -13,7 +13,7 @@ public sealed class LightCluster : VolumeComponent
         /// Controls the range of the cluster around the camera.
         /// </summary>
         [Tooltip("Controls the range of the cluster around the camera.")]
-        public ClampedFloatParameter cameraClusterRange = new ClampedFloatParameter(10f, 0.001f, 50f);
+        public MinFloatParameter cameraClusterRange = new MinFloatParameter(10.0f, 0.001f);
```
sebastienlagarde, line 16: Sadly as this is a public value we aren't allowed to rename it... it will be a breaking change. so we will need to change the UX but keep the name

anisunity: ok will update that

PR #2762, opened 2020-11-26
Introduce concept of falling back textures to black in rendergraph + ReadWrite tag

FrancescoC-unity:
This among a more general thing fixes: https://fogbugz.unity3d.com/f/cases/1295272/

The issue was that because normal buffer is set globally is marked as read in the transparency pass. However, if no opaque objeccts were defined, the texture handle would be valid (we create it outside passes), but the content won't be as it was never written to. 

Instead of fixing the specific case, I introduced the concept of write count and fall back to black texture whenever the above case happens and that should happen automagically whenever a texture is declared to have such behaviour. This logic needs to be on the Rendergraph resource to avoid issues when an handle is copied around but is not updated. 

On top of that, since it is important for the mechanism that all Reads are defined after a Write, I replaced the somewhat ugly pattern of builder.ReadTexture(builder.WriteTexture(...))  with a compact builder.ReadWriteTexture(...)

**What did I test**: the repro case and all the DXR tests (most of the ReadWrite code was in DXR)


com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphResourceRegistry.cs
```diff
@@ -121,15 +135,15 @@ internal RendererListResource(in RendererListDesc desc)
...
         DynamicArray<IRenderGraphResource>[] m_Resources = new DynamicArray<IRenderGraphResource>[(int)RenderGraphResourceType.Count];
 
-        TexturePool                         m_TexturePool = new TexturePool();
-        int                                 m_TextureCreationIndex;
-        ComputeBufferPool                   m_ComputeBufferPool = new ComputeBufferPool();
-        DynamicArray<RendererListResource>  m_RendererListResources = new DynamicArray<RendererListResource>();
-        RenderGraphDebugParams              m_RenderGraphDebug;
-        RenderGraphLogger                   m_Logger;
-        int                                 m_CurrentFrameIndex;
+        TexturePool m_TexturePool = new TexturePool();
```
FrancescoC-unity, line 138: Yikes visual studio, gonna revert this

RemyUnity:
I don't really know what those changes might do or affect, do you have any hint about what additional tests of check QA should do ?

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphBuilder.cs
```diff
@@ -49,6 +52,20 @@ public TextureHandle UseDepthBuffer(in TextureHandle input, DepthAccess flags)
...
+                var texDimension = m_Resources.GetTextureResourceDesc(input.handle).dimension;
+                if (texDimension == TextureXR.dimension)
+                {
+                    return m_RenderGraph.defaultResources.blackTextureXR;
+                }
+                else if (texDimension == TextureDimension.Tex3D)
+                {
+                    return m_RenderGraph.defaultResources.blackTexture3DXR;
+                }
+            }
```
JulienIgnace-Unity, line 67: missing else for regular black texture?

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphBuilder.cs
```diff
@@ -61,11 +78,25 @@ public TextureHandle ReadTexture(in TextureHandle input)
...
         }
 
+        /// <summary>
+        /// Specify a Texture resource to read and write to during the pass.
+        /// </summary>
+        /// <param name="input">The Texture resource to read and write to during the pass.</param>
+        /// <returns>An updated resource handle to the input resource.</returns>
+        public TextureHandle ReadWriteTexture(in TextureHandle input)
+        {
+            CheckResource(input.handle);
```
JulienIgnace-Unity, line 94: Seems like this one is missing the IncrementWriteCount.
Wasn't it possible to put it directly in AddResourceWrite in the end?

FrancescoC-unity: Ah yes, moved it later and forgot to update this. 

I need m_Resources to call the Increment; will update in a sec

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
```diff
@@ -798,7 +799,10 @@ void RenderDBuffer(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle dec
                 passData.parameters = PrepareRenderDBufferParameters(hdCamera);
                 passData.meshDecalsRendererList = builder.UseRendererList(renderGraph.CreateRendererList(PrepareMeshDecalsRendererList(cullingResults, hdCamera, use4RTs)));
                 SetupDBufferTargets(renderGraph, passData, use4RTs, ref output, builder);
-                passData.decalBuffer = builder.ReadTexture(decalBuffer);
+                if (hdCamera.frameSettings.IsEnabled(FrameSettingsField.DecalLayers))
```
JulienIgnace-Unity, line 802: Leftovers from previous tests?
We talked about the dbuffer pass directly returning black if possible to remove this logic (and in the render pass itself)

JulienIgnace-Unity (changes requested):
Made a few comments

FrancescoC-unity:
@RemyUnity not much aside the repro case I think, there is nothing user facing that won't be caught by automation I think

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
FrancescoC-unity: Yes, will change that, was planning on doing a sanity check on stuff like that later, but might as well fix this one now.

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphBuilder.cs
FrancescoC-unity: Yup

anisunity:
From what I read, it looks good to me, didn't test it locally though

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Looks good to me.

What I've tested:
Forward and Deferred mode
Opaque Objects enabling/disabling for Cameras, Realtime Probes, Custom Reflection
Opaque Objects enabling/disabling in Player debug mode

PR #2768, opened 2020-11-27
Fixed SSGI texture allocation with RG disabled

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1293020/
SSGI works even if raytracing is not supported
removes if before RTHandles.Release because they are already done inside the function

---
### Testing status

tested the repro steps

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -772,19 +772,18 @@ namespace UnityEngine.Rendering.HighDefinition
             {
                 m_RaytracingGBufferManager.CreateBuffers();
                 m_RayCountManager.InitializeNonRenderGraphResources();
-
-                if (m_Asset.currentPlatformRenderPipelineSettings.supportSSGI)
-                {
-                    m_IndirectDiffuseBuffer0 = RTHandles.Alloc(Vector2.one, TextureXR.slices, colorFormat: GraphicsFormat.R16G16B16A16_SFloat, dimension: TextureXR.dimension, enableRandomWrite: true, useDynamicScale: true, useMipMap: false, autoGenerateMips: false, name: "IndirectDiffuseBuffer0");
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -850,24 +849,18 @@ void DestroyRenderTextures()
...
-                    RTHandles.Release(m_IndirectDiffuseHitPointBuffer);
-
                 m_RayTracingLightCluster.CleanupNonRenderGraphResources();
 
                 RTHandles.Release(m_FlagMaskTextureRT);
 
                 RaytracingManagerCleanupNonRenderGraphResources();
             }
+
+            RTHandles.Release(m_IndirectDiffuseBuffer0);
```
anisunity, line 859: The release function is safe even if the pointer is null?

anisunity:
Did you run the DXR tests locally?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
adrien-de-tocqueville: yep

adrien-de-tocqueville:
> Did you run the DXR tests locally?

i don't support DXR locally, I can launch yamato but I guess it will be launched on the bugfix branch altogether

sebastienlagarde (approved):

sebastienlagarde:
> > Did you run the DXR tests locally?
> 
> i don't support DXR locally, I can launch yamato but I guess it will be launched on the bugfix branch altogether

For this kind of thing, you should indeed run DXr test on yamato, I will do it

sebastienlagarde:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix-ssgi-no-rg/.yamato%252Fall-hdrp.yml%2523All_HDRP_trunk/4388893/job

sebastienlagarde:
good for the test , merging

PR #2769, opened 2020-11-27
[HDRP] AOV API improvements

pmavridis:
### Purpose of this PR
This PR introduces some improvements for the AOV API:
- Fresnel0 output buffer
- WorldSpacePosition output buffer
- Internal rendering of the debug views is now using the buffer format provided by the user (no change in the public AOV API). 

Bugfix: in the rendergraph path we have broken the "custom pass" AOV output, because we were not testing it :( 

---
### Testing status
The new features (and the custom pass AOV output) are tested with a new graphics test (ID 9902):
image

---
### Comments to reviewers
Many new files in this PR are related to the graphics test.


JulienIgnace-Unity (approved):

pmavridis:
Note for backport:
If we don't backport the whole PR to 10.x, we need to backport at least this commit to fix rendergraph:
https://github.com/Unity-Technologies/Graphics/pull/2769/commits/dbd4bb79b7d35b0d20e21d44c61383f454f22897

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVRequestData.cs
```diff
@@ -200,6 +201,24 @@ List<RTHandle> targets
             }
         }
 
+        internal void OverrideBufferFormatForAOVs(ref GraphicsFormat format, List<RTHandle> aovBuffers)
```
sebastienlagarde, line 204: Aren't we missing somewhere an update of the documentation? 
also feel the what's new in 11 pages

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVRequest.cs
```diff
@@ -172,6 +176,12 @@ public void FillDebugData(DebugDisplaySettings debug)
                 case DebugFullScreen.MotionVectors:
                     debug.SetFullScreenDebugMode(FullScreenDebugMode.MotionVectors);
                     break;
+                case DebugFullScreen.WorldSpacePosition:
+                    debug.SetFullScreenDebugMode(FullScreenDebugMode.WorldSpacePosition);
+                    break;
+                case DebugFullScreen.Fresnel0:
+                    debug.SetDebugViewGBuffer(MaterialDebugSettings.GetGBufferMaterialEnumByName("fresnel0"));
```
sebastienlagarde, line 183: it is not the correct way of handling this, think about multiple material (hair, fabric etc...)
you should use the debug matching this statement
[MaterialSharedPropertyMapping(MaterialSharedProperty.Specular)]

sebastienlagarde (commented):
Let's have a chat about this PR. :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVRequest.cs
pmavridis: Thanks for the comment, I will fix this as we have discussed.

pmavridis:
Note: I'm in the process of resolving the merge conflicts, so until then this is not ready for review yet!

pmavridis:
I forgot to say that this PR is ready for reviewing now. 
Changes from previous version:
- For Fresnel zero, I'm now using the MaterialSharedProperty.Specular output, and I'm making sure that we get sensible information for materials that use the metalness workflow (convert metalness to f0)
- We can now override the the render format for the internal HDRP rendering when outputting AOVs (so this will affect lighting calculations too). To do this, the user needs to call a new public API:
aovRequest.SetOverrideRenderFormat(true);
- The docs have been updated to include this information

sebastienlagarde (approved):

sebastienlagarde:
This PR is ready and green ABV, waiting for QA approval. @RemyUnity

sebastienlagarde:
Adding release management per request

sebastienlagarde:
Note: the failure on Mac and XR are from trunk and master.

pmavridis:
I have updated one last test image to make Nightly VFX_HDRP green (after discussing with Paul).

JMargevics (approved):
Made a pass. Things I've checked: 

- Tested new AOV buffers with HDRP shaders and all available master nodes (except Unlit for specular)  ✅
- Set the AOV buffer format to GraphicsFormat.R32G32B32A32 and verify in RenderDoc.  Anything after the **LorResTransparent** pass uses the correct format. ✅ (Passes before are still on 16bit, according to Pavlos this is expected for some of them since they rely on this format)
- Wrangle vertex positions in SG and see if it affects World Position AOV. ✅
- Checked if VFX mesh outputs appear correctly in new AOV buffers ✅
- Verified on MacOS ✅

I have no breaking issues so far, all good on my side ✅

pmavridis:
Just to give more context regarding the format override:
- We don't override the format of certain passes than need a specific format, like the g-buffer. Also passes like LowResTransparent always needs an alpha channel, so we also don't override that.

Having said that, I'm looking at a RenderDoc capture again to see if I have missed something.
(And I will fix the merge conflicts, since this branch is a bit old now)

pmavridis:
There was an additional issue with the format of the color pyramid generation that I have discovered after @JMargevics comments. I'm submitting the fix in a separate PR:
https://github.com/Unity-Technologies/Graphics/pull/3470
because it might affect users outside of the AOV API (just in case we want to backport).

PR #2771, opened 2020-11-27
[HDRP][Path Tracing] Fixed issues with path-traced volumetrics (light anim dirtiness, truncated volume shafts)

eturquin:
This PR addresses the following bugs:
https://fogbugz.unity3d.com/f/cases/1295222/
https://fogbugz.unity3d.com/f/cases/1295234/

The first was an issue with animated lights not triggering accumulation reset when an animation is played, the second produced volumes that looked truncated when the camera got closer to spot/cone/pyramid point lights. Both are now resolved (can be tested on scenes provided in the corresp. tickets).


remi-chapelain (approved):
Looks good to me ! ✔️ 
Tested with different type of light and motions on the repro projects

sebastienlagarde (approved):

PR #2775, opened 2020-11-27
Improve performance of OnDemand reflection probes [NEED C++]

JulienIgnace-Unity:
### Purpose of this PR
Requires Ono PR: https://ono.unity3d.com/unity/unity/pull-request/123697/_/terrain/getactiveterrainsnoalloc

When using OnDemand reflection probes, unwanted performance spikes were occuring. This PR aims at improving this by doing several things:
- Removed some GCAllocs in probe internal creation code
- Stopped cleaning up cameras used for probes to avoid reallocating them and their volume stack (which was very costly)
- Stopped updating the volume system for Lighting Override when it's not used
- Disable cleaning up of terrain camera data for probe cameras (this is what requires C++ change)

This should improve https://fogbugz.unity3d.com/f/cases/1293225/

---
### Testing status
Tested the small repro project provided in the fogbugz case.
Spikes should be much improved. There should still be smaller spikes because we do render the probe regularly (so 6 more camera renders) but antyhing else than rendering should be mostly gone.
It might be worth testing Detail Renderer (is that even supported in HDRP?) as it may have been impacted by the C++ change.
Also, checking the original user project would be good.

@Unity-Technologies/gfx-qa-hdrp here's a link to an editor with the C++ changes required to run the HDRP branch: https://bokken-artifacts-ui.bf.unity3d.com/katana-production/proj0-Build_WindowsEditor/39048765_11_01_2021_22_20_11_+0000/build/ZippedForTeamCity.zip


alelievr (approved):

TomasKiniulis (changes requested):
With the PR there is 40ms improvement, but discussing with Julien it seems that majority of ms time is still taken by "CullScriptable" on user project(https://forum.unity.com/threads/hdrp-enable-camera-is-really-slow.988695/). It still uses up to 175ms which causes frame stutters.

**Attaching my findings:**
EmptyProject with terrain enabled:
image


PR DeepProfile
image
trunk master DeepProfile
image

PR without DeepProfile
image
trunk master without DeepProfile
image

TomasKiniulis (approved):
Looks good now! Testing the user project TryCull() method goes down from 180ms to 20ms!
**trunk master**
image (3)
**PR**
image (5)


Adding a lot more additional details on terrain does not cause a significant spike anymore.

sebastienlagarde:
fogbugzz: https://fogbugz.unity3d.com/f/cases/1324062/

sebastienlagarde:
Currently waiting for this C++ PR to land: https://ono.unity3d.com/unity/unity/pull-request/123697/_/terrain/getactiveterrainsnoalloc

sebastienlagarde (approved):

PR #2776, opened 2020-11-27
Avoid issues causing faulty transitions in shadows (resulting in no shadows with unconventional aspect ratio) 

FrancescoC-unity:
Fixes the issue in https://fogbugz.unity3d.com/f/cases/1292295/

The issue is due to how the cascade spheres are and how the border is computed. As a result sometimes we try to transition when we shouldn't; this PR fixes the issue to a certain extent (with ludicrous transition percentages, usually >=90% of the full split). 

This also fixes another bug due to the precision of the dither transition that showed with small dots like this 

image

We already went through it with @remi-chapelain


remi-chapelain (approved):
Fix validated with @FrancescoC-unity before making the PR ! ✔️

sebastienlagarde (approved):

PR #2778, opened 2020-11-27
[Yamato] Disable editor pinning ABV dependency

kivistikliis:
---
### Purpose of this PR
Removes ABV dependency from editor pinning. This is equivalent to force merging editor revisions nightly, as green ABV is no longer a requirement. 

Editor pinning is useful with a stable green nightly, but given the numerous packages and their compatibility with trunk, it is rare to obtain a revision which is green for everything, thus resulting in trunk revision updates having to be manual. This causes us to be blocked/stuck on a given revision for longer times. Therefore this PR drops the ABV dependency: revisions are updated nightly no matter what. 

Additional work to this PR is being done to store green revisions per jobs instead on branch https://github.com/Unity-Technologies/Graphics/tree/yamato/disable-editor-pinning-green 

---
### Comments to reviewers
The PR allows to specify ABV dependency per editor track in the metafile, thus with multiple tracks it can be a per-track configuration. There will be 2 merge-all jobs:
- one with [CI] label, which picks abv dependencies up from metafiles, and runs by trigger
- other with [no CI] label, for which the trigger as well as ABV is removed for all tracks, essentially being a manual forced update (useful to keep in case we add ABV back to CI flow one day)

### DIFF
--- a/.yamato/_editor_pinning.yml
+++ b/.yamato/_editor_pinning.yml
@@ -118,10 +118,10 @@ editor-pinning-merge-all:
     dependencies:
       -  .yamato/_editor_pinning.yml#editor-pinning-merge-revisions-trunk
 editor-pinning-merge-all-ci:
-    name: Merge all [CI][trunk ABV]
+    name: Merge all [CI][trunk no ABV]
     triggers:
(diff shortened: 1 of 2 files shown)


sophiaaar (approved):

sophiaaar:
we will need to announce this before merging

PR #2781, opened 2020-11-30
[URP 2d] Fix inconsistent SetRenderTarget calls

rolandk-unity:
### Purpose of this PR

Before this PR, the exact way the camera render target is set differs `Render2DLightingPass` and `RenderLighting`.

- `Render2DLightingPass` used `CoreUtils.SetRenderTarget(cmd, colorAttachment, depthAttachment, ClearFlag.None, Color.white)`, which resulted in a `SetRenderTargetColorDepth_Internal(new RenderTargetIdentifier(color, 0, CubemapFace.Unknown, -1), ...)` call
- `RenderLighting` used `cmd.SetRenderTarget(renderTexture, depthTexture)`, which resulted in a `SetRenderTargetColorDepth_Internal(color, ...);` call

This is relevant, because the on the C++ side these render targets are considered different. On iOS, this can lead to situations where setting the already-set camera render target in `RenderLighting` incorrectly starts a new render pass. As a result, both the color and depth buffer are stored to the main memory and loaded back into tile memory, needlessly using bandwidth.

A short evaluation on my iPhone SE 2nd gen revealed that unifying the two SetRenderTarget calls resulted in a 1.5 MB (~1% of whole frame) bandwidth saving on the Lost Crypt project.

---
### Testing status
Manual testing on iOS and MacOS. Automated testing ongoing.

Note that the formatting checks fail on files in HDRP and Shader Graph, not any file that was touched by this PR. Not sure if it would make sense to fix formatting issues in another package in this PR.

Note that HDRP test failures are unrelated, as this PR does not touch the HDRP.

---
### Comments to reviewers
N/A

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/2D/Passes/Render2DLightingPass.cs
+++ b/com.unity.render-pipelines.universal/Runtime/2D/Passes/Render2DLightingPass.cs
@@ -158,7 +158,7 @@ namespace UnityEngine.Experimental.Rendering.Universal
             var blendStylesCount = m_Renderer2DData.lightBlendStyles.Length;
             using (new ProfilingScope(cmd, m_ProfilingDrawRenderers))
             {
-                CoreUtils.SetRenderTarget(cmd, colorAttachment, depthAttachment, ClearFlag.None, Color.white);
+                cmd.SetRenderTarget(colorAttachment, depthAttachment);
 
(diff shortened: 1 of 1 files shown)


UnityPaul (approved):

rainsing (approved):

phi-lira:
Formatting job failing on files not touched by this PR.

PR #2783, opened 2020-11-30
HDRP/Fix package version showing package after the last "verified" package

RSlysz:
After enquiering on FB issue on package version shown in wizard not the same as the one shown in package manager windows, it seams that this is expected behaviour of package lifecycle v2.
Only remaining :
- a typo (fixed in this PR)
- a design decision to make: align Wizard with Package Manager or remains with old behaviour for package last version discoverability

**UPDATE:**

Removed last package version check from wizard and add a button for shortchut check this in Package Manager. So now we have alignment.
image

### Tests done
Manually tested that:
- the version of current HDRP package appears. Tested with local package only. "(local)" should only appears at end for local packages.
- the retrieval of config package local installation still work. Changes introduced a regression on this due to asynchrone call conflict. It is handled with a stack so we should not bother on order but previous code only did it if nothing was in the stack. I dunno the reason. In the test, no local package was installed so it found it and display the installation button after the fix.
- the **Check update** button open the Package Manager window on the HDRP package. Also tested with the window already open. Not tested with the window opened on another package but this is package manager API so I assume it was tested on their side.


sebastienlagarde (approved):

sebastienlagarde:
>a design decision to make: align Wizard with Package Manager or remains with old behaviour for package last version >discoverability
if telling people that there is a new package cause us so much pain (we have get fogbugzz issue because of that few times already), then let's simply remove this feature and maybe replace by a shortcut to packman and an info box saying that latest version can be found in packman windows?

RSlysz:
**Convert back to draft:** 
I'll replace the version check with a link to package manager windows in this PR

RSlysz:
**Convert back to draft:**
Missing documentation update

RSlysz:
@JordanL8, @sebastienlagarde  I have seen that the documentation was not up to date. I updated it. Though a pass is needed on the DXR side. (VR have no change)

com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Wizard.md
```diff
@@ -97,5 +97,6 @@ When upgrading a project from the built-in render pipeline to HDRP, you need to
 
 - **Upgrade Project Materials to High Definition Materials**: Upgrades every Material in your Unity Project to HDRP Materials.
 - **Upgrade Selected Materials to High Definition Materials**: Upgrades every Material currently selected to HDRP Materials.
+- **Upgrade HDRP Materials to Latest Version:** Upgrades every Materials in your Unity Project to latest version available.
```
JordanL8, line 100: Upgrades every Material in your Unity Project to the latest version.

com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Wizard.md
```diff
@@ -8,16 +8,15 @@ To open the **Render Pipeline Wizard**, go to **Window > Render Pipeline** and s
 
 ## Packages
 
-At the top of the window, there is an information box that shows you the currently installed version of HDRP, as well as the latest version of HDRP that is compatible with your current Unity version.
+At the top of the window, there is an information text that shows you the currently installed version of HDRP. The button **Check update** provide you a quick shortcut to the **Package Manager** windows onto the HDRP package.
```
JordanL8, line 11: The **Check Update** button provides a shortcut to the HDRP window in the Package Manager.

JordanL8 (changes requested):
Rewrote a couple of bits

com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Wizard.md
RSlysz: Well, not sure but now I understand that HDRP have a window but it is in fact Package Manager windows, with HDRP package selected in the package list on the left, so the HDRP package description is shown on the right.

Perhaps this?
The **Check Update** button provides a shortcut to the HDRP package in the Package Manager window.

JordanL8: Yeah, sounds good

iM0ve:
Please specify whats tested :) Its a bit more than a typo, so I think still worth running a wizard in an empty 3D project would be beneficial.

RSlysz:
> Please specify whats tested :) Its a bit more than a typo, so I think still worth running a wizard in an empty 3D project would be beneficial.

Right sorry. I updated the PR with what was tested.

JordanL8 (approved):

iM0ve (approved):
Thanks for the update, looks good.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #2793, opened 2020-11-30
Updated the feature comparison table. Added a clarification about upgrading shaders from built-in to URP.

oleks-k:
Updated the feature comparison table.
Added a clarification about upgrading shaders from built-in to URP.


phi-lira (approved):

phi-lira:
Merging. I already confirmed with XR and AR team the changes.

PR #2798, opened 2020-12-01
Improvement to the ray traced reflection denoising

anisunity:
- Added a toggle to allow to include or exclude smooth surfaces from ray traced reflection denoising.
- Fixed issues in geometry buffer reading for half res RTR and RTGI.
- Removed the upscale radius from the RTR.
- How the history validity is properly tracked for RTR.
- Changed the denoising strategy for very rough RTR surfaces.

Testing status
Tested on various scenes
Ran the tests locally

This PR needs to be tested quite extensively before being merged. I already discussed what needs to be tested with @remi-chapelain .


anisunity:
@JordanL8 FYI

JordanL8 (approved):

remi-chapelain:
[Not finished testing]
Since parts of the denoisers are shared between RTR perf mode and RTGI perf mode and Upscale radius was removed for RTR in perf mode, would it make sense to also remove it for RTGI ? Either by removing the effect completely (since it can be very subtle especially with both denoisers ON) or hardcode the value somewhere if the cost can be negligible?

fb4993a419323014a0306fc7ad8278a6

remi-chapelain (approved):
Tried multiple scene from out Full Test pass project with different reflective setup with different smoothness transitions. ✔️

Tested with different viewport size with scene and game view, deferred and forward. ✔️ 
Took a look at the updated doc ✔️ 

Also tested RTGI in perf (because part of denoiser is shared) and difference is very subtle but can be seen as "better" but definitely not worse. ✔️
See comment above to also remove upscale radius in the RTGI also to match what is done with SSR in perf. ❗ 

Couldn't find a case/setup where the PR is worse than what was previously done. 
Biggest improvement can be seen using half resolution in performance mode. 
Checkbox that force to affect perfectly smooth surface definitely helps reducing aliasing and small glitches that attracts the eye ✔️ 

[Example](https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=fc39a40a-3652-11eb-83c8-ebb5d6f907df) of improvement at half resolution (Left is before / right is after)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingReflection.cs
```diff
@@ -520,6 +510,97 @@ static void RenderQualityRayTracedReflections(CommandBuffer cmd, RTRQualityRende
             CoreUtils.SetKeyword(cmd, "MULTI_BOUNCE_INDIRECT", false);
         }
 
+        #endregion
+
+        #region Legacy Pipeline
```
sebastienlagarde, line 515: What is "Legacy pipeline"?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/Reflections/RaytracingReflections.compute
```diff
@@ -26,13 +28,60 @@
...
+// Structure that holds everything that defines a GGX sample
+struct GGXSample
+{
+    float3 direction;
+    float pdf;
+};
+
+GGXSample GenerateGGXSampleDirection(uint2 currentCoord, float3 normalWS, float3 viewWS, float roughness, int frameIndex)
+{
+    // Create the local ortho basis
```
sebastienlagarde, line 50: is this function really specialise to this files? Shouldn't this be share.

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingReflection.cs
anisunity: Non-render graph

anisunity: this pr took too much time to get merged. @JulienIgnace-Unity removed this part of the code i think

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/Reflections/RaytracingReflections.compute
anisunity: This exact version is only used for perf rtr

anisunity:
After doing a rebase, i started getting nans. I need to investigate.

anisunity:
Merging and updating screenshots

PR #2802, opened 2020-12-01
[11.x.x] Updating 146_MSAA_RenderToBackbuffer Test scene

ellioman:
# Purpose of this PR
Using the current version of 146 makes it a bit difficult to determine whether MSAA is being applied or not, especially on console platforms. This PR fixes that by adding a new URP Asset with the render scale set to lowest: 0.1, and uses then when rendering the scene.

This change makes it much easier for us to spot whether MSAA is working or not. Example:
Screen Shot 2020-12-01 at 13 09 46
Screen Shot 2020-12-01 at 13 09 56

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fmsaa-test-improvement


manuele-bonanno (approved):
Looks much better! Thanks!

phi-lira (approved):

PR #2804, opened 2020-12-01
[VFX] Staging to master

julienf-unity:
### Purpose of this PR
**Added**
- Added new setting to output nodes to exclude from TAA
- New Sample Point cache & Sample Attribute map operators

**Changed**
- Changed the "Edit" button so it becomes "New" when no asset is set on a Visual Effect component, in order to save a new visual effect graph asset.

**Fixed**
- Forbid incorrect link between incompatible context Case 1269756
- Serialization issue with VFXSpawnerCallbacks
- Unexpected exception while trying to display capacity warning Case 1294180
- Exclude Operator, Context, Block and Subgraph from Preset Case 1232309
- Fix Case 1212002
- Fix LPPV not working in deferred
- Spaceship specific warnings
- Fix incorrect compilation condition for linux build

---
### Testing status
Green locally (except known failures from trunk)

---
### Comments to reviewers


PaulDemeulenaere (approved):
All 🟢 in Yamato (at least for VFX)

PR #2805, opened 2020-12-01
[Hold] Binned Lighting Technique for the HDRP (v3)

sebastienlagarde:
### Purpose of this PR
Add binned lighting (a.k.a. Z-binning) technique to the HDRP. Binned lighting replaces both fine-pruned tiled lighting (a.k.a. FPTL) and clustered lighting techniques. Binned lighting is essentially tiled+.

Short comparison:

| Technique               | Resolution in X&Y | Resolution in Z | Transparent geometry |
| --------------------- | ------------------- | ----------------- | ------------------------- |
| Tiled lighting           | High                     | None                | Yes                                 |
| FPTL                         | High                    | Perfect              | No                                  |
| Binned lighting       | High                    | Some                | Yes                                 |
| Clustered lighting   | Low                     | Low                   | Yes                                 |

Performance-wise, FPTL ≥ Binned lighting ≥ Tiled lighting.
Note that there is some overhead associated with filtering the light list during the lighting pass as well.

The work is incomplete.

What works:
* binning;
* classification;
* build dispatch indirect;
* Lit, TerrainLit, Deferred Lighting, Volumetric shaders;
* decals.

What doesn't:
* everything else. In particular, DeferredTile.shader does not work.

Next steps (in no particular order):
* fix ray tracing & contact shadows (which made liberal use of tiled lighting buffers);
* add the render graph support;
* add a single debug mode (more details on request).

Once all the compile errors are fixed and everything works and passes automated tests (including ray tracing and XR), we should perform manual regression testing and verify culling efficiency using the added debug mode.

The next step (once everything works) is to perform some testing on the consoles and optimize. Specifically, we must scalarize access to the data structure for AMD GPUs.

The final step is to remove the unused old code.

---
### Testing status

image

---
### Comments to reviewers
If anyone is interested, I accept offers to help (reach out to me or Sebastien).

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
@@ -158,6 +158,10 @@ namespace UnityEngine.Rendering.HighDefinition
 
             // We could decode the category from the key, but it is probably not worth the effort.
             m_Views[viewIndex].entitySortKeys[m_TotalEntityCount] = key;
+        }
+
+        public void IncrementEntityCount(BoundedEntityCategory category)
+        {
             m_EntityCountPerCategory[(int)category]++;
             m_TotalEntityCount++;
         }
@@ -837,6 +841,7 @@ namespace UnityEngine.Rendering.HighDefinition
         DynamicArray<ProcessedLightData> m_ProcessedLightData = new DynamicArray<ProcessedLightData>();
         DynamicArray<ProcessedProbeData> m_ProcessedReflectionProbeData = new DynamicArray<ProcessedProbeData>();
         DynamicArray<ProcessedProbeData> m_ProcessedPlanarProbeData = new DynamicArray<ProcessedProbeData>();
+        DynamicArray<bool> m_ProcessedShadowLightList = new DynamicArray<bool>();
 
         static readonly Matrix4x4 s_FlipMatrixLHSRHS = Matrix4x4.Scale(new Vector3(1, 1, -1));
 
@@ -2662,6 +2667,7 @@ namespace UnityEngine.Rendering.HighDefinition
             // Lights are processed in order, so we don't discards light based on their importance but based on their ordering in visible lights list.
 
             m_ProcessedLightData.Resize(cullResults.visibleLights.Length);
+            m_ProcessedShadowLightList.Resize(cullResults.visibleLights.Length);
 
             int maxLightCount      = m_MaxDirectionalLightsOnScreen + m_MaxPunctualLightsOnScreen + m_MaxAreaLightsOnScreen;
             int includedLightCount = 0;
@@ -2676,6 +2682,7 @@ namespace UnityEngine.Rendering.HighDefinition
 
                 // Then we compute all light data that will be reused for the rest of the light loop.
                 ref ProcessedLightData processedData = ref m_ProcessedLightData[lightIndex];
+                m_ProcessedShadowLightList[lightIndex] = false;
                 PreprocessLightData(ref processedData, light, hdCamera);
 
                 // Then we can reject lights based on processed data.
@@ -2744,6 +2751,7 @@ namespace UnityEngine.Rendering.HighDefinition
 
                         m_BoundedEntityCollection.AddEntitySortKey(viewIndex, processedData.lightCategory, key);
                     }
+                    m_BoundedEntityCollection.IncrementEntityCount(processedData.lightCategory);
                 }
 
                 includedLightCount++;
@@ -2867,8 +2875,9 @@ namespace UnityEngine.Rendering.HighDefinition
                     int shadowIndex = -1;
 
                     // Manage shadow requests
-                    if (additionalLightData.WillRenderShadowMap())
+                    if (additionalLightData.WillRenderShadowMap() && !m_ProcessedShadowLightList[lightIndex])
                     {
+                        m_ProcessedShadowLightList[lightIndex] = true;
                         int shadowRequestCount;
                         shadowIndex = additionalLightData.UpdateShadowRequest(hdCamera, m_ShadowManager, hdShadowSettings, light, cullResults, lightIndex, m_CurrentDebugDisplaySettings.data.lightingDebugSettings, shadowFilteringQuality, out shadowRequestCount);
 
@@ -3024,7 +3033,7 @@ namespace UnityEngine.Rendering.HighDefinition
 
                         m_BoundedEntityCollection.AddEntitySortKey(viewIndex, BoundedEntityCategory.ReflectionProbe, key);
                     }
-
+                    m_BoundedEntityCollection.IncrementEntityCount(BoundedEntityCategory.ReflectionProbe);
                     includedProbeCount++;
                 }
             }
@@ -3053,7 +3062,7 @@ namespace UnityEngine.Rendering.HighDefinition
                         ulong key = GenerateBoundedEntitySortingKey(planarProbeIndex, BoundedEntityCategory.ReflectionProbe, d, (int)GPULightType.PlanarReflection);
                         m_BoundedEntityCollection.AddEntitySortKey(viewIndex, BoundedEntityCategory.ReflectionProbe, key);
                     }
-
+                    m_BoundedEntityCollection.IncrementEntityCount(BoundedEntityCategory.ReflectionProbe);
                     includedProbeCount++;
                 }
             }
@@ -3082,8 +3091,6 @@ namespace UnityEngine.Rendering.HighDefinition
                     var category     = (BoundedEntityCategory)((sortKey >> layout.categoryOffset)  & ((1ul << layout.categoryBitCount)  - 1));
                     var gpuLightType = (GPULightType)((sortKey >> layout.lightTypeOffset) & ((1ul << layout.lightTypeBitCount) - 1));
                     var probeIndex   = (int)((sortKey >> layout.indexOffset)     & ((1ul << layout.indexBitCount)     - 1));
-
-                    Debug.Assert(category == BoundedEntityCategory.ReflectionProbe);
 
                     ProcessedProbeData processedProbe = (gpuLightType == GPULightType.PlanarReflection) ? m_ProcessedPlanarProbeData[probeIndex]
                         : m_ProcessedReflectionProbeData[probeIndex];
@@ -3184,6 +3191,7 @@ namespace UnityEngine.Rendering.HighDefinition
                         ulong key = GenerateBoundedEntitySortingKey(decalIndex, BoundedEntityCategory.Decal, decalIndex);
                         m_BoundedEntityCollection.AddEntitySortKey(viewIndex, BoundedEntityCategory.Decal, key);
                     }
+                    m_BoundedEntityCollection.IncrementEntityCount(BoundedEntityCategory.Decal);
                 }
 
                 int densityVolumeCount = (densityVolumes.data != null) ? Math.Min(densityVolumes.data.Count, m_MaxDensityVolumesOnScreen) : 0;
@@ -3199,6 +3207,7 @@ namespace UnityEngine.Rendering.HighDefinition
 
                         m_BoundedEntityCollection.AddEntitySortKey(viewIndex, BoundedEntityCategory.DensityVolume, key);
                     }
+                    m_BoundedEntityCollection.IncrementEntityCount(BoundedEntityCategory.DensityVolume);
                 }
 
                 /* ---------------------------- Step 2 ---------------------------- */
--- a/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/2x_Lighting/2008_Light_Volumetrics.unity
+++ b/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/2x_Lighting/2008_Light_Volumetrics.unity
@@ -24,7 +24,7 @@ RenderSettings:
   m_AmbientEquatorColor: {r: 0.114, g: 0.125, b: 0.133, a: 1}
   m_AmbientGroundColor: {r: 0.047, g: 0.043, b: 0.035, a: 1}
   m_AmbientIntensity: 1
-  m_AmbientMode: 4
+  m_AmbientMode: 0
   m_SubtractiveShadowColor: {r: 0.42, g: 0.478, b: 0.627, a: 1}
   m_SkyboxMaterial: {fileID: 0}
   m_HaloStrength: 0.5
@@ -43,7 +43,7 @@ RenderSettings:
 --- !u!157 &3
 LightmapSettings:
   m_ObjectHideFlags: 0
-  serializedVersion: 11
+  serializedVersion: 12
   m_GIWorkflowMode: 1
   m_GISettings:
     serializedVersion: 2
@@ -98,7 +98,7 @@ LightmapSettings:
     m_TrainingDataDestination: TrainingData
     m_LightProbeSampleCountMultiplier: 4
   m_LightingDataAsset: {fileID: 0}
-  m_UseShadowmask: 1
+  m_LightingSettings: {fileID: 0}
 --- !u!196 &4
 NavMeshSettings:
   serializedVersion: 2
@@ -118,6 +118,8 @@ NavMeshSettings:
     manualTileSize: 0
     tileSize: 256
     accuratePlacement: 0
+    maxJobWorkers: 0
+    preserveTilesOutsideBounds: 0
     debug:
       m_Flags: 0
   m_NavMeshData: {fileID: 0}
@@ -205,15 +207,50 @@ PrefabInstance:
   m_Modification:
     m_TransformParent: {fileID: 94104563}
     m_Modifications:
-    - target: {fileID: 6004892619064504655, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+    - target: {fileID: 198941061589059314, guid: 7340be8cd4dd8d34d808e2c6090d869f,
         type: 3}
-      propertyPath: m_Name
-      value: Shadow Dimmer (2)
+      propertyPath: m_Text
+      value: 0.25
       objectReference: {fileID: 0}
-    - target: {fileID: 6004892619064504655, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+    - target: {fileID: 198941061589059314, guid: 7340be8cd4dd8d34d808e2c6090d869f,
         type: 3}
-      propertyPath: m_IsActive
+      propertyPath: m_Color.rgba
+      value: 4293394358
+      objectReference: {fileID: 0}
+    - target: {fileID: 198941061589059314, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+        type: 3}
+      propertyPath: m_CharacterSize
       value: 1
+      objectReference: {fileID: 0}
+    - target: {fileID: 2366749451762349873, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+        type: 3}
+      propertyPath: pixelSize
+      value: 11
+      objectReference: {fileID: 0}
+    - target: {fileID: 2366749451762349873, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+        type: 3}
+      propertyPath: targetCamera
+      value: 
+      objectReference: {fileID: 1029596512}
+    - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+        type: 3}
+      propertyPath: m_RootOrder
+      value: 7
+      objectReference: {fileID: 0}
+    - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+        type: 3}
+      propertyPath: m_LocalScale.x
+      value: 0.41697517
+      objectReference: {fileID: 0}
+    - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+        type: 3}
+      propertyPath: m_LocalScale.y
+      value: 0.41697517
+      objectReference: {fileID: 0}
+    - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+        type: 3}
+      propertyPath: m_LocalScale.z
+      value: 0.41697517
       objectReference: {fileID: 0}
     - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
         type: 3}
@@ -232,6 +269,11 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
         type: 3}
+      propertyPath: m_LocalRotation.w
+      value: 1
+      objectReference: {fileID: 0}
+    - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
+        type: 3}
       propertyPath: m_LocalRotation.x
       value: -0
       objectReference: {fileID: 0}
@@ -244,16 +286,6 @@ PrefabInstance:
         type: 3}
       propertyPath: m_LocalRotation.z
       value: -0
-      objectReference: {fileID: 0}
-    - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
-        type: 3}
-      propertyPath: m_LocalRotation.w
-      value: 1
-      objectReference: {fileID: 0}
-    - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
-        type: 3}
-      propertyPath: m_RootOrder
-      value: 7
       objectReference: {fileID: 0}
     - target: {fileID: 5489785850083597078, guid: 7340be8cd4dd8d34d808e2c6090d869f,
         type: 3}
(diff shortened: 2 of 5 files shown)


EvgeniiG:
After merging, the render graph is now on by default, and spams validation errors every frame.
So you must disable the render graph in the HDRP asset until the support is in.

EvgeniiG:
Partial render graph support is in; the screen is no longer black.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Deferred.shader
```diff
@@ -128,17 +127,20 @@ Shader "Hidden/HDRP/Deferred"
                 // input.positionCS is SV_Position
                 float depth = LoadCameraDepth(input.positionCS.xy);
 
-                PositionInputs posInput = GetPositionInput(input.positionCS.xy, _ScreenSize.zw, depth, UNITY_MATRIX_I_VP, UNITY_MATRIX_V, uint2(input.positionCS.xy) / GetTileSize());
+                PositionInputs posInput = GetPositionInput(input.positionCS.xy, _ScreenSize.zw, depth, UNITY_MATRIX_I_VP, UNITY_MATRIX_V);
                 float3 V = GetWorldSpaceNormalizeViewDir(posInput.positionWS);
 
+                uint tile = ComputeTileIndex(posInput.positionSS);
```
sebastienlagarde, line 133: This should be move inside  LightLoop(), we must not change the prototype of LightLoop(), will break lot of stuff for users.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
```diff
@@ -6,7 +6,7 @@ namespace UnityEngine.Rendering.HighDefinition
 
     // Caution: Order is important and is use for optimization in light loop
     [GenerateHLSL]
-    enum GPULightType
+    enum GPULightType // Rename to "LightType"
```
sebastienlagarde, line 9: IIRC we haven't done it because LightType is already use in builtin

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/ClippingUtilities.hlsl
```diff
@@ -0,0 +1,317 @@
...
+// Remove when the shader compiler is clever enough to perform this optimization for us.
+#define OBTUSE_COMPILER
+
+// Clipping a plane by a cube may produce a hexagon (6-gon).
+// Clipping a hexagon by 4 planes (substitute a quad for the plane) may produce a decagon (10-gon).
+#define MAX_CLIP_VERTS     (10)
+#define NUM_VERTS          (8)
+#define NUM_FACES          (6)
+#define NUM_PLANES         (6)
+#define THREADS_PER_GROUP  (64)
```
sebastienlagarde, line 73: Mark for myself about the 64

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/DeferredTile.shader
```diff
@@ -43,7 +43,8 @@ Shader "Hidden/HDRP/DeferredTile"
 
             #define USE_INDIRECT    // otherwise TileVariantToFeatureFlags() will not be defined in Lit.hlsl!!!
 
-            #define USE_FPTL_LIGHTLIST 1 // deferred opaque always use FPTL
+            // Comment out the line to loop over all lights (for debugging purposes)
```
sebastienlagarde, line 46: is deferred tile working?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/HDShadowLoop.hlsl
```diff
@@ -208,9 +155,13 @@ void ShadowLoopMin(HDShadowContext shadowContext, PositionInputs posInput, float
                         weight      += 1.0f - shadowA;
                     }
                 }
-
-                lightData = FetchLight(lightStart, min(++i, last));
             }
+            else // GPULIGHTTYPE_TUBE
+            {
+                /* ??? */
```
sebastienlagarde, line 161: shadow aren't supported for tube currently

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -51,24 +52,303 @@ public static VisibleLightAxisAndPosition GetAxisAndPosition(this VisibleLight v
...
+                m_Views[i].entityBounds.Clear();
+
+                // Since we reset 'm_TotalEntityCount', we can leave 'entitySortKeys' alone.
+            }
+
+            // No need to clear ComputeBuffers, we will just overwrite the contents.
+        }
+
+        // Release the resources that are NOT automatically managed by C#.
+        // TODO: call this somewhere.
```
sebastienlagarde, line 135: Frightening todo here :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -121,18 +401,29 @@ class LightDefinitions
         public static uint s_ScreenSpaceColorShadowFlag = 0x100;
         public static uint s_InvalidScreenSpaceShadow = 0xff;
         public static uint s_ScreenSpaceShadowIndexMask = 0xff;
+
+        // Binned lighting
+        // For performance reasons, keep all sizes in powers of 2.
+        public static int s_CoarseTileEntryLimit = 64; // Per category; before pruning, so the number is lower in practice; the number could be made (almost always) exact at the cost of making the 'FillCoarseTiles' pass more complicated
```
sebastienlagarde, line 407: From the comment I do'nt understand what the 64 and 16 imply here, what is the limit?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -121,18 +401,29 @@ class LightDefinitions
...
+        public static int s_FineTileEntryLimit   = 16; // Per category; after pruning, so the number is exact
+        public static int s_CoarseTileSize       = 64;
+        public static int s_FineTileSize         = 8;
+        public static int s_zBinCount            = 8192;
+        public static int s_MaxReflectionProbesPerPixel = 4;
```
sebastienlagarde, line 412: >s_MaxReflectionProbesPerPixel = 4
we may need to revisit that. Even if this is what we should have, I expect many unity project to have more due to "users not caring"

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -302,6 +593,7 @@ public partial class HDRenderPipeline
         int m_MaxPunctualLightsOnScreen;
         int m_MaxAreaLightsOnScreen;
         int m_MaxDecalsOnScreen;
+        int m_MaxDensityVolumesOnScreen = 1024; // TODO
```
sebastienlagarde, line 596: Don't put the init value here

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -677,39 +976,67 @@ struct ScreenSpaceShadowData
 
         static MaterialPropertyBlock m_LightLoopDebugMaterialProperties = new MaterialPropertyBlock();
 
-        bool HasLightToCull()
+        static Vector4 GetZBinBufferEncodingParams(HDCamera hdCamera)
         {
-            return m_TotalLightCount > 0;
+            // encodingParams = { n, log2(f/n), 1/n, 1/log2(f/n) }
+            // n = 0.1. See ComputeFixedPointLogDepth.
+            float n = 0.1f;
```
sebastienlagarde, line 983: why not using camera.nearClipPlane ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1528,7 +1866,7 @@ bool EnoughScreenSpaceShadowSlots(GPULightType gpuLightType, int screenSpaceChan
 
                 // Keep track of the screen space shadow data
                 m_CurrentScreenSpaceShadowData[screenSpaceShadowIndex].additionalLightData = additionalLightData;
-                m_CurrentScreenSpaceShadowData[screenSpaceShadowIndex].lightDataIndex = m_lightList.lights.Count;
+                m_CurrentScreenSpaceShadowData[screenSpaceShadowIndex].lightDataIndex = -1; /* WILL NOT WORK IN XR */
```
sebastienlagarde, line 1869: can you elaborate here? what work / doesn't in XR?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2132,6 +2469,143 @@ internal static bool IsBakedShadowMaskLight(Light light)
...
+        // 'f' is the distance to the far plane.
+        // We consider the distance to the near plane n ≈ 0, since that plane may be oblique.
+        static int ComputeFixedPointLogDepth(float w, float f, int numBits = 16)
+        {
+            // z = Log[w/n] / Log[f/n]
+            // Undefined for (w < n, so we must clamp). This should not significantly affect the efficiency of Z-binning.
+            // Still need the distance to the near plane in order for the math to work.
+            // Setting it too low will quickly consume the available bits. 0.1 is a safe value.
+            // See https://zero-radiance.github.io/post/z-buffer/
+            const float n = 0.1f;
```
sebastienlagarde, line 2546: Same question than previously, why are you not using the camera near?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2634,16 +3155,73 @@ void PrepareGPUProbeData(CommandBuffer cmd, HDCamera hdCamera, CullingResults cu
                     m_CurrentScreenSpaceShadowData[i].valid = false;
                 }
 
+                /* ---------------------------- Step 1 ---------------------------- */
+                // Do not pre-process per view...
```
sebastienlagarde, line 3159: What do you mean?
note: avoid comment with /* -- */ in it, a it prevent to use /* for multiple ligne when we want to comment a large part of the code.
Prefer 
/* ---------------------------- Step 1 ---------------------------- */
+> 
// ---------------------------- Step 1 ----------------------------

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
EvgeniiG: This is optional. :-)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/DeferredTile.shader
EvgeniiG: Yes, you can test it yourself. Well, at least it worked before I switched to RG...

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
EvgeniiG: Coarse tiles are 64x64 pixels. Fine tiles are 8x8 pixels.
Each coarse tiles can contain at most 64 indices per category (64 punctual & 64 area & ..., etc). Fine tiles can contain 16 indices.
Since the coarse tiles are computed in two steps (fill and prune), it may not be possible to have 64 entries there (some will be pruned). We can use extra storage for indices before the pruning but this hasn't been implemented yet.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
EvgeniiG: Yeah, this was done very quickly...

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
EvgeniiG: 1. It could be oblique, which means that min-z depends on the pixel coordinate.
2. Small values of the near plane are costly in terms of precision and we only have 8192 bins, so 13 bits of precision.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
EvgeniiG: Light data index depends on the eye index.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
EvgeniiG: Answered above.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
EvgeniiG: Doing anything per-view is obviously expensive. It should be avoided if the data does not depend on the eye index.

EvgeniiG: You can change the style of comments if you like.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Deferred.shader
EvgeniiG: I don't like the idea of tileCoord in the position input.
First, most shaders do not need it.
Secondly, you can clearly see that even our internal developers do not understand it, and put random stuff in there.

tileCoord is only useful for the light loop and should be its parameter. z-bin is debatable indeed.

This PR will introduce a lot of breaking changes, anyway.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
EvgeniiG: Maybe call it from CleanupLightLoop? Not sure.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Deferred.shader
sebastienlagarde: yes, you are right. TileCoord should be in input position.
And is something related to the ligthloop. But wondering how to abstract it. Because utimately what we want is to have lightloop being able itself from PositionInput (and maybe other stuff) to get the correct param for Zbinning or any algorithm we could come in the future.

>This PR will introduce a lot of breaking changes, anyway.
Still if we can still have users code compile it is a win. We will see at the end but maybe we will introduce a prototype with an ignore parameter just to ignore the compiler complain

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/classification.compute
```diff
@@ -0,0 +1,136 @@
...
+
+TEXTURE2D_X_UINT2(_StencilTexture);
+
+/* ------------------------------ Outputs ----------------------------------- */
+
+RWStructuredBuffer<uint> g_TileFeatureFlags;
+
+/* ------------------------------ Utilities --------------------------------- */
+
+bool IsBinEmpty(uint tile, uint zBin, uint category, uint eye)
```
sebastienlagarde, line 34: Guess this function should be in TilingAndBinningUtilities.hlsl

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/classification.compute
```diff
@@ -0,0 +1,136 @@
...
+        }
+    #endif // LIGHT_CLASSIFICATION
+
+    #ifdef MATERIAL_CLASSIFICATION
+        featureFlags |= MATERIAL_FEATURE_FLAGS_FROM_GBUFFER(clampedPixelCoord);
+    #endif // MATERIAL_CLASSIFICATION
+    }
+
+#if (defined(PLATFORM_SUPPORTS_WAVE_INTRINSICS) && (PLATFORM_LANE_COUNT >= (TILE_SIZE * TILE_SIZE)))
+    featureFlags = WaveActiveBitOr(featureFlags);
```
sebastienlagarde, line 122: isn't this only working if lane count == TILE_SIZE * TILE_SIZE  ?  (and not >= ) ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1941,7 +1941,7 @@ internal void EvaluateShadowState(HDCamera hdCamera, in ProcessedLightData proce
                 bool validShadow = false;
                 if (processedLight.gpuLightType == GPULightType.Point
                     || processedLight.gpuLightType == GPULightType.Rectangle
-                    || (processedLight.gpuLightType == GPULightType.Spot && processedLight.lightVolumeType == LightVolumeType.Cone))
+                    || processedLight.gpuLightType == GPULightType.Spot)
```
sebastienlagarde, line 1944: why is this condition remove? guess it is required for screen space shadow? (poking @JulienIgnace-Unity  just in case)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
```diff
@@ -169,164 +169,296 @@ float4 SampleEnv(LightLoopContext lightLoopContext, int index, float3 texCoord,
...
+    if (!isTileEmpty) // Avoid wasted work
+    {
+        const uint zBinBufferIndex0 = ComputeZBinBufferIndex(zBinRange[0], category, unity_StereoEyeIndex);
+        const uint zBinBufferIndex1 = ComputeZBinBufferIndex(zBinRange[1], category, unity_StereoEyeIndex);
+        const uint zBinRangeData0   = _zBinBuffer[zBinBufferIndex0]; // {last << 16 | first}
+        const uint zBinRangeData1   = _zBinBuffer[zBinBufferIndex1]; // {last << 16 | first}
+
+        // Recall that entities are sorted by the z-coordinate.
+        // So we can take the smallest index from the first bin and the largest index from the last bin.
+        // TODO: while it appears to work, for a large range, this code can actually fail. :-(
```
sebastienlagarde, line 199: Question: Why don't you do two functions? one for the case of a single position (i.e we use in light loop) 
and thus const uint2 zBinEntityIndexRange = uint2(zBinRangeData0 & UINT16_MAX, zBinRangeData1 >> 16); = >onst uint2 zBinEntityIndexRange = uint2(zBinRangeData & UINT16_MAX, zBinRangeData >> 16);

and another one for the volumetric with the range. Are you trusting the compiler to do the right thing here?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
```diff
@@ -169,164 +169,296 @@ float4 SampleEnv(LightLoopContext lightLoopContext, int index, float3 texCoord,
...
+
+        if (IntervalsOverlap(tileEntityIndexRange, zBinEntityIndexRange)) // Avoid wasted work
+        {
+            const uint tileBufferBodyIndex = ComputeTileBufferBodyIndex(tile, category, unity_StereoEyeIndex);
 
-uint GetTileSize()
-{
-    return TILE_SIZE_FPTL;
-}
+            // The part (2) below will be actually executed during every function call.
```
sebastienlagarde, line 210: Really would like to see the output of the compiler here :).
this part is indeed what is suppose to be executed in the while (try loading) loop. But pretty sure compiler will not be smart enough (and the various if()  may not help) . Anyway we can check when code work.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
```diff
@@ -169,164 +169,296 @@ float4 SampleEnv(LightLoopContext lightLoopContext, int index, float3 texCoord,
...
+                uint tileEntityIndex = BitFieldExtract(tileEntityPair, 16 * (i & 1), 16); // First Lo, then Hi bits
+
+                // Entity indices are stored in the ascending order.
+                // We can distinguish 3 cases:
+                if (tileEntityIndex < zBinEntityIndexRange.x)
+                {
+                    i++; // Skip this entity; continue the (linear) search
+                    // Fine point: entities are currently sorted by Centroid.z.
+                    // So it could be that bin #0 starts with entity #1 (which is large)
+                    // while bin #1 contains both entity #0 (which is small) and entity #1.
```
sebastienlagarde, line 224: This is why I am wondering should we do a dedicated version for non range (lights) and another for range (volumetric ) ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/classification.compute
EvgeniiG: If it's greater then the remaining lanes should be inactive and thus not contribute to the result.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
EvgeniiG: LightVolume was the "thing" Morten added for clustered lighting and I removed it.

It is possible that I've made a mistake modifying this particular part of code.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
EvgeniiG: This optimization is trivial. I don't want to duplicate the code.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
EvgeniiG: Again, this is a trivial optimization. If it doesn't work then it's a compiler bug.

EvgeniiG: https://en.wikipedia.org/wiki/Loop-invariant_code_motion

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
EvgeniiG: I would prefer not to duplicate the code.

com.unity.render-pipelines.high-definition/Runtime/RenderPipelineResources/HDRenderPipelineResources.asset
```diff
@@ -142,13 +142,9 @@ MonoBehaviour:
...
       type: 3}
     momentShadowsCS: {fileID: 7200000, guid: 4dea53e2ff15ed0448817c2aa4246e53, type: 3}
-    decalNormalBufferPS: {fileID: 4800000, guid: fd532bf1795188c4daaa66ea798b8b0a, type: 3}
-    decalClearPropertyMaskBufferCS: {fileID: 7200000, guid: 1076a08965d4a91479b72599724f7fd6, type: 3}
+    shadowBlitPS: {fileID: 4800000, guid: ca059f1af4587a24b9a9eed3b66cff0f, type: 3}
     decalNormalBufferPS: {fileID: 4800000, guid: fd532bf1795188c4daaa66ea798b8b0a,
       type: 3}
-    shadowBlitPS: {fileID: 4800000, guid: ca059f1af4587a24b9a9eed3b66cff0f, type: 3}
-    decalNormalBufferPS: {fileID: 4800000, guid: fd532bf1795188c4daaa66ea798b8b0a, type: 3}
-    decalClearPropertyMaskBufferCS: {fileID: 7200000, guid: 1076a08965d4a91479b72599724f7fd6, type: 3}
```
sebastienlagarde, line 151: need to check why decalClearPropertyMaskBufferCS have disappeard

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -334,9 +334,9 @@ public enum FrameSettingsField
         /// <summary>When enabled, HDRP uses FPTL for forward opaque.</summary>
         [FrameSettingsField(3, autoName: FPTLForForwardOpaque, tooltip: "When enabled, HDRP uses FPTL for forward opaque.")]
         FPTLForForwardOpaque = 120,
-        /// <summary>When enabled, HDRP uses a big tile prepass for light visibility.</summary>
-        [FrameSettingsField(3, autoName: BigTilePrepass, tooltip: "When enabled, HDRP uses a big tile prepass for light visibility.")]
-        BigTilePrepass = 121,
+        /// <summary>When enabled, HDRP performs binning to determine the set of visible lights.</summary>
+        [FrameSettingsField(3, autoName: BinnedLighting)]
```
sebastienlagarde, line 338: isn't this only for the case of deffered? shouldn't we call this option deferred Binned lighting?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -3108,18 +3111,15 @@ RendererListDesc PrepareForwardTransparentRendererList(CullingResults cullResult
...
             CommandBuffer               cmd)
         {
-            // Note: SHADOWS_SHADOWMASK keyword is enabled in HDRenderPipeline.cs ConfigureForShadowMask
-            bool useFptl = opaque && frameSettings.IsEnabled(FrameSettingsField.FPTLForForwardOpaque);
-
-            // say that we want to use tile/cluster light loop
-            CoreUtils.SetKeyword(cmd, "USE_FPTL_LIGHTLIST", useFptl);
-            CoreUtils.SetKeyword(cmd, "USE_CLUSTERED_LIGHTLIST", !useFptl);
-            cmd.SetGlobalBuffer(HDShaderIDs.g_vLightListGlobal, lightListBuffer);
+            // TODO: we have so many SetGlobal calls like this that it makes more sense to just do it once...
```
sebastienlagarde, line 3120: you can remove the comment, we have no choice anyway as we need to bind for forward.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -477,32 +476,18 @@ class ForwardTransparentPassData : ForwardPassData
...
             data.frameSettings = frameSettings;
-            data.lightListBuffer = builder.ReadComputeBuffer(useFptl ? lightLists.lightList : lightLists.perVoxelLightLists);
-            if (!useFptl)
-            {
-                data.perVoxelOffset = builder.ReadComputeBuffer(lightLists.perVoxelOffset);
-                if (lightLists.perTileLogBaseTweak.IsValid())
-                    data.perTileLogBaseTweak = builder.ReadComputeBuffer(lightLists.perTileLogBaseTweak);
-            }
+
+            /* TODO: we shouldn't be reading these buffers if tiled lighting is disabled... */
```
sebastienlagarde, line 481: use // instead of /* */ when doing comment like this.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
```diff
@@ -779,17 +678,15 @@ void VolumetricLighting(uint3 dispatchThreadId : SV_DispatchThreadID,
...
-    uint2 tileCoord = groupOffset * VBUFFER_VOXEL_SIZE / TILE_SIZE_BIG_TILE;
+    uint2 tileCoord = groupOffset * VBUFFER_VOXEL_SIZE / TILE_SIZE;
+    uint  tile      = IndexFromCoordinate(tileCoord, TILE_BUFFER_DIMS.x);
 #else
     // No compile-time optimizations, no scalarization.
-    uint2 tileCoord = pixelCoord / TILE_SIZE_BIG_TILE;
+    uint  tile      = ComputeTileIndex(pixelCoord);
 #endif
-    uint  tileIndex = tileCoord.x + _NumTileBigTileX * tileCoord.y;
-    // This clamp is important as _VBufferVoxelSize can have float value which can cause en overflow (Crash on Vulkan and Metal)
```
sebastienlagarde, line 788: We need to keep this clamp in a way or the other, of do similar clamping, we have spend several QA and programmer hour on this one :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumeVoxelization.compute
```diff
@@ -308,19 +264,20 @@ void VolumeVoxelization(uint3 dispatchThreadId : SV_DispatchThreadID,
...
 #else
     // No compile-time optimizations, no scalarization.
-    // If _VBufferVoxelSize is not a power of 2 or > TILE_SIZE_BIG_TILE, a voxel may straddle
-    // a tile boundary. This means different voxel subsamples may belong to different tiles.
-    // We accept this error, and simply use the coordinates of the center of the voxel.
-    uint2 tileCoord = (uint2)((voxelCoord + 0.5) * _VBufferVoxelSize) / TILE_SIZE_BIG_TILE;
+    uint  tile      = ComputeTileIndex(pixelCoord);
 #endif
-    uint  tileIndex = tileCoord.x + _NumTileBigTileX * tileCoord.y;
-    // This clamp is important as _VBufferVoxelSize can have float value which can cause en overflow (Crash on Vulkan and Metal)
```
sebastienlagarde, line 322: we need to keep this clamp (see comment in volumetric lighting)

EvgeniiG: I clamp elsewhere.

PR #2808, opened 2020-12-01
Overlay camera now inherints target texture, display and viewport from base camera

lukaschod:
# **Please read;**

### Checklist for PR maker

---
### Purpose of this PR
Overlay camera now inherints target texture, display and viewport from base camera during rendering.

---
### Testing status
Added three graphics tests that test these scenarios.

## 150_CameraStacking_Canvas_DisplayMissmatch
https://fogbugz.unity3d.com/f/cases/1256936/
After Fix
image
Before Fix
image

## 151_CameraStacking_Canvas_RTMissmatch
https://fogbugz.unity3d.com/f/cases/1291872/
After Fix
image
Before Fix
image

## 152_CameraStacking_ViewportMissmatch
This issue I found just by expermenting with output options, so decided to include it too.
After Fix
image
Before Fix
image

- [ ] Check if there is no performance regression.
- [ ] Make sure no changes to scene asset happens, because of properties patching during rendering.

---
### Comments to reviewers
I actually want a bit discusion what do you think about the change and should we commit on that.

Currently from time to time we are getting issues related to camera stacking, because of some native functionality still depends on camera output properties. It is extremely hard to catch them all and even harder to fix them.

One of the solution with Felipe, we decide to patch overlay camera output properties during rendering from base camera. This PR represent this solution. Sadly we can not do this once as part of editor settings as overlay camera can be used by multiple base cameras.

With this PR we will be patching:
- Target Texture.
- Target Display.
- Viewport.

It should not also effect performance. As in most cases overlay camera has one base camera it means those properties will only change once. In case it is used by multiple base cameras calls are farely simple:

Target Texture
image
image
image

Target Display
image

Viewport
image

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
@@ -542,6 +542,15 @@ namespace UnityEngine.Rendering.Universal
                             if (baseCameraData.xr.enabled)
                                 m_XRSystem.UpdateFromCamera(ref overlayCameraData.xr, overlayCameraData);
 #endif
+
+                            // There is still some native codes that depends on camera properties
+                            // Current know native systems:
+                            // - ScriptableRendererContext.SetupCameraProperties.
+                            // - Screen space camera canvas.
+                            currCamera.targetTexture = baseCamera.targetTexture;
+                            currCamera.targetDisplay = baseCamera.targetDisplay;
+                            currCamera.rect = baseCamera.rect;
+
                             RenderSingleCamera(context, overlayCameraData, anyPostProcessingEnabled);
 
                             using (new ProfilingScope(null, Profiling.Pipeline.endCameraRendering))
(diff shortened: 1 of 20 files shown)


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -542,6 +542,15 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
...
                                 m_XRSystem.UpdateFromCamera(ref overlayCameraData.xr, overlayCameraData);
 #endif
+
+                            // There is still some native codes that depends on camera properties
+                            // Current know native systems:
+                            // - ScriptableRendererContext.SetupCameraProperties.
+                            // - Screen space camera canvas.
+                            currCamera.targetTexture = baseCamera.targetTexture;
+                            currCamera.targetDisplay = baseCamera.targetDisplay;
+                            currCamera.rect = baseCamera.rect;
```
lukaschod, line 552: I could actually add conditions if (currCamera.targetDisplay == baseCamera.targetDisplay). However native code inside takes care of it. So I can add it just to avoid scripting cost

lukaschod:
After deeper investigation: Changing properties does not directly trigger update, but any other change will result in changed serialized ouput of camera. This is not something we want, for this reason I am closing this PR

PR #2809, opened 2020-12-01
Fix Low Resolution Light Unit Icon

johnpars:
### Purpose of this PR
Tiny PR that fixes light unit icons which appear low resolution on retina / high DPI displays. The problem was due to a typo in the icon file name, and how the built-in icon loading behavior fell back to the low resolution icon.

Before:
Before

After:
After

---
### Testing status
- Tested on my retina display MacBook to confirm that the @2x icons are properly loaded.

---
### Comments to reviewers
- Just a file rename, so no code to review.


sebastienlagarde (approved):

sebastienlagarde:
Removing QA, no need for review for such a small thing. retraget to hd/bugfix

PR #2810, opened 2020-12-01
Add internal package for XR automated graphics tests

fabien-unity:
---
### Purpose of this PR
Centralize XR Graphics code facility into one internal package that will be used by all existing test projects under the SRP repo.

---
### Testing status
Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/tests%252Fxr%252Freusable-package/.yamato%252F_abv.yml%2523all_project_ci_trunk/4429555/job

---
### Comments to reviewers
The package is currently not used anywhere. A follow-up PR will connect it to the URP test project.

### DIFF
--- a/com.unity.testing.xr/Editor.meta
+++ b/com.unity.testing.xr/Editor.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 172eda4f97f610a4eb5c6a032ea5adda
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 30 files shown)


thomas-zeng (approved):

theo-pnv (approved):
Approving on behalf of the sdets, since Sophia is off.
However note that I'm not familiar with this code, and with how new packages are supposed to be made, so feel free to wait for her return if possible, or to double check with other reviewers :)
Some open-questions:
- Does this new package need some kind of test automation?
- Does it needs an entry in the [Codeowners ](https://github.com/Unity-Technologies/Graphics/blob/tests/xr/reusable-package/.github/CODEOWNERS) file?
Thanks!

fabien-unity:
Thanks @theo-pnv !
The package includes  few unit tests to confirm correct behavior that will be run with https://github.com/Unity-Technologies/Graphics/pull/2812
As this is an internal package only for testing, I don't think the codeowners file applied but Sophia can confirm :)

sebastienlagarde:
ok, given current state we can merge it.

PR #2815, opened 2020-12-02
Increase UniversalGraphicsTest 145_ManyAdditionalLightShadowsTypes threshold to pass on local machines

Wilfrid-Unity:
### Purpose of this PR

https://github.com/Unity-Technologies/Graphics/pull/1616 introduced automated Graphics Test https://github.com/Unity-Technologies/Graphics/blob/master/TestProjects/UniversalGraphicsTest/Assets/Scenes/145_ManyAdditionalLightShadowTypes.unity

While the newly added test passes within Unity's automatic testing pipeline Yamato, it fails when executed locally from the UnityEditor (Window/General/TestRunner/PlayMode).

@ellioman reported the issue and suggested an Average Correctness Threshold value that passes on local OXS machine.
Further increasing this value also makes the test pass on other local Windows machines, hence the currently proposed value: 0.0011

### Testing status
The test was confirmed to pass on local machines with this increased value.


ellioman (approved):

ellioman:
This PR only increases the threshold so running ABV is unnecessary.

PR #2816, opened 2020-12-02
2d/docs update

unity-cchu:

This is a dev draft updating the docs to reflect changes in URP 11


Sam-Unity (approved):
Will edit docs in next iteration.

phi-lira (approved):

PR #2833, opened 2020-12-03
Fixed renderer creation to work without urp asset assigned

lukaschod:
### Purpose of this PR
https://unity.slack.com/archives/GBJ7AU4V8/p1606927313209800
- Fixing renderers correctly recreate post process passes if post process data changes in urp asset.
- Fixing renderers creation to work when there is not urp asset assigned in project.
- Extending test CreateForwardRendererAssetWithoutErrors with renderer creation. 
- Adding test CreateRenderer2DAssetWithoutErrors.
- Moving changelog entry about post process data from 10.2 to 10.1, which was incorrect.

---
### Testing status
- [x] Checked if forward renderer and 2d renderer no longer drops error when created with urp asset.
- [x] Checked if extended test actually covers this sitauation (Fails without this fix and passes with this fix).
- [x] Ran universal gfx test locally.
- [ ] Checked if changed post process data applies to renderers.

---
### Comments to reviewers
Universal_OSX_Metal_playmode_mono_Linear_trunk fails on master too https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Funiversal-osx-metal.yml%2523Universal_OSX_Metal_playmode_mono_Linear_trunk/4435112/job

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/PostProcessPasses.cs
+++ b/com.unity.render-pipelines.universal/Runtime/PostProcessPasses.cs
@@ -89,4 +89,3 @@ namespace UnityEngine.Rendering.Universal
         }
     }
 }
-


com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -805,5 +802,36 @@ bool CanCopyDepth(ref CameraData cameraData)
             bool msaaDepthResolve = false;
             return supportsDepthCopy || msaaDepthResolve;
         }
+
+        /// <summary>
+        /// Recreate post process passes. If renderer already contains valid post process passes, they will be replaced by new ones.
+        /// </summary>
+        /// <param name="data">Resources used for creating passes. In case of the null, no passes will be created.</param>
+        private void RecreatePostProcessPasses(PostProcessData data)
+        {
```
lukaschod, line 811: Now we have same function in forward and 2d renderer. Maybe we should create base class that has this method? TBH I am not fan of inheritence, because very quikly we can end up with 5 levels of inheritence (This was huge problem in our old test framework, where inheritence is over-used).

com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
```diff
@@ -251,5 +249,36 @@ public override void FinishRendering(CommandBuffer cmd)
...
+            {
+                m_ColorGradingLutPass?.Cleanup();
+                m_PostProcessPass?.Cleanup();
+                m_FinalPostProcessPass?.Cleanup();
+
+                // We need to null post process passes to avoid using them
+                m_ColorGradingLutPass = null;
+                m_PostProcessPass = null;
+                m_FinalPostProcessPass = null;
+                m_CurrentPostProcessData = null;
```
lukaschod, line 272: I actually want to add this https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types. Do we support it in graphics repo?

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
phi-lira: We already have ScriptableRenderer base class. You can move it there.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -287,6 +287,10 @@ internal static class ShaderPropertyId
 
     public struct PostProcessingData
     {
+        /// <summary>
+        /// Post process data used.
+        /// </summary>
+        public PostProcessData resources;
```
phi-lira, line 293: Not ideal we have PostProcessData and PostProcessingData :( 
We should at least make it more clear what the data here is in docs, is it resources?

com.unity.render-pipelines.universal/Tests/Editor/EditorTests.cs
```diff
@@ -4,6 +4,7 @@
 using UnityEngine.Rendering;
 using UnityEngine.Rendering.Universal;
 using UnityEditor.Rendering.Universal.Internal;
+using UnityEngine.Experimental.Rendering.Universal;
```
phi-lira, line 7: Which API is requiring experimental namespace?

ellioman (approved):

erikabar (approved):
fix verified, lgtm

com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
```diff
@@ -29,6 +29,9 @@ internal class Renderer2D : ScriptableRenderer
 
         Renderer2DData m_Renderer2DData;
 
+        PostProcessData m_RendererPostProcessData;
+        PostProcessData m_CurrentPostProcessData;
```
phi-lira, line 33: Is this shared between renderers? Maybe could be moved to ScriptableRenderer

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
lukaschod: Maybe lets make it internal for now.

com.unity.render-pipelines.universal/Tests/Editor/EditorTests.cs
lukaschod: Renderer2D

com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
lukaschod: K

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
lukaschod: K

phi-lira (approved):
LGTM. The "PostProcessPasses" screams like PostProcessFeature :)

PR #2836, opened 2020-12-03
[HDRP] Fix rendering issues for the first frame

pmavridis:
### Purpose of this PR
- Exposure of the first frame is "wrong" when using the auto-exposure modes. With wrong I mean "not the same as the non-RenderGraph path". (the apply exposure pass for the first frame was not using the correct texture)
- The depth pyramid was not computed in the first frame, and effects that were using it (like AO) were wrong.
(this was caused because DepthBufferMipChainInfo is computed during render graph execute (during the RT alloc), but it was used during the setup of some passes)

---
### Testing status
- Open an HDRP project that is using with auto-exposure and TAA 
(TAA is optional but exaggerates the first-frame flashing, so it is easy to see)
- Hit play
- There is no flashing in the first frame

In my test I have used the test project from this bug ticket: https://fogbugz.unity3d.com/f/cases/1287088/, but any HDRP project with auto-exposure and AO should repro.
The first-frame behavior is now the same as the non-rendergraph path (checked with RenderDoc, used a branch where the non-rendergraph was still in the repo)


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
```diff
@@ -317,9 +317,7 @@ TextureHandle DynamicExposurePass(RenderGraph renderGraph, HDCamera hdCamera, Te
                     {
                         passData.source = builder.ReadTexture(source);
                         passData.parameters = PrepareApplyExposureParameters(hdCamera);
-                        RTHandle prevExp;
-                        GrabExposureHistoryTextures(hdCamera, out prevExp, out _);
-                        passData.prevExposure = builder.ReadTexture(renderGraph.ImportTexture(prevExp));
+                        passData.prevExposure = renderGraph.ImportTexture(GetPreviousExposureTexture(hdCamera));
```
pmavridis, line 320: @FrancescoC-unity An alternative fix would be to have
`passData.prevExposure =nextExposureHandle;`
This was my first fix, but in the end I have used the GetPreviousExposureTexture call because it was also used in the non-rendergraph version.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
```diff
@@ -52,6 +52,12 @@ bool NeedClearGBuffer()
 
         HDUtils.PackedMipChainInfo GetDepthBufferMipChainInfo()
         {
+            // Initialize the MipChainInfo if nececcary. This will be necessary when rendering the first frame.
+            if (m_DepthBufferMipChainInfo.textureSize == Vector2Int.zero)
+            {
+                m_DepthBufferMipChainInfo.ComputePackedMipChainInfo(new Vector2Int(RTHandles.maxWidth, RTHandles.maxHeight));
```
pmavridis, line 58: The problem here is that m_DepthBufferMipChainInfo was initialized during RT alloc with a callback, which has not been called yet during the render-graph setup of the first frame. So any pass that was using this was getting zeros (uninitialized data).
I'm not very familiar yet with RenderGraph, so I'm not sure if there is a better fix. 
@JulienIgnace-Unity If you can think something better please let me know.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
FrancescoC-unity: GetPreviousExposureTexture is the correct thing to call as it setups the history system correctly.

FrancescoC-unity: However isn't that grabbed already by GrabExposureRequiredTextures ? If I read the code right, no need to do the get again and perform the import again, you should have the handle in prevExposureHandle

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
FrancescoC-unity: Question for @JulienIgnace-Unity , shouldn't the callback be called upon first usage and hence allocation of the RT?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
```diff
@@ -52,6 +52,12 @@ bool NeedClearGBuffer()
 
         HDUtils.PackedMipChainInfo GetDepthBufferMipChainInfo()
         {
+            // Initialize the MipChainInfo if nececcary. This will be necessary when rendering the first frame.
```
FrancescoC-unity, line 55: Typo nececcary

pmavridis: Thanks! it's funny how I had it correctly the second time I used the word in the same comment.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
pmavridis: If I use the prevExposureHandle, then the ApplyExposure pass will run with the default EV100 exposure texture and this creates a very bright flashing in the first frame (which we were not getting in the non-rendergraph path). This is also the case if we don't change anything.

My understanding (and what we were doing in the non-render graph case) is that we have to use the exposure texture that we have just computed with the code in the previous lines (the dynamic-exposure pass). I get this behavior with the GetPreviousExposureTexture or if I use the nextExposureHandle directly (this is where the dynamic-exposure pass stores the result), but not with the prevExposureHandle.

FrancescoC-unity: Gotcha, my memory is fuzzy here, but if it was mimicking what was in the non-rendergraph path, all good 👍

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
pmavridis: An alternative fix is to move all calls to GetDepthBufferMipChainInfo() inside the render callbacks/functions (so RT alloc will happen first). However it would be easy to re-introduce the first-frame bug if someone calls GetDepthBufferMipChainInfo() outside of a render callback in the future. That's why I opted for the fix that you see.

JulienIgnace-Unity: @FrancescoC-unity yes it will but only when executing the graph. So any code relying on those values during graph construction will have wrong values.
Also this fix could lead to some more errors. It uses the global RTHandles system which might not always be the one for this texture. If we move this to history buffers for example this will break.
To me it seems that it would be easier to just call ComputePackedMipChainInfo with the HDCamera actualWidth/Height like it's done currently but at the beginning of ExecuteWithRenderGraph.
This way it should always work properly.

pmavridis: > To me it seems that it would be easier to just call ComputePackedMipChainInfo with the HDCamera actualWidth/Height like it's done currently but at the beginning of ExecuteWithRenderGraph.
> This way it should always work properly.

Ok thanks, I will do that, test it, and push an update.

pmavridis: And yes, this suggestion has worked. Thanks!

JulienIgnace-Unity (approved):

FrancescoC-unity (approved):

sebastienlagarde (approved):

PR #2837, opened 2020-12-03
Fix regex that excludes files from the formatting

theopnv:
### Purpose of this PR

Some files weren't caught by the formatting job, resulting in breaking commands when performing vendoring operations on Ono.
Example: 
Here's the formatting job before fixing the regex: only one file is reported
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/tools%252Ftest-formatting-fix/.yamato%252F_formatting.yml%2523formatting/4443952/job

Here's the formatting job after fixing the regex: more files are reported
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/tools%252Ftest-formatting-fix/.yamato%252F_formatting.yml%2523formatting/4459056/job


sophiaaar:
perhaps do a double check with regex101.com if you havent already

theo-pnv:
We want the regex to exclude files:
- Inside folders that finish with a tilde `~` (eg Documentation~)
- AND whose name does not include the `&` character (shell commands interpret this character and break)

image

sophiaaar (approved):

PR #2840, opened 2020-12-04
[HDRP][Path Tracing] Improved volumetric scattering sampling

eturquin:
Addressing https://fogbugz.unity3d.com/f/cases/1295187/

Volumetric scattering positional sampling (along the ray) has been streamlined: now at least half the samples go to the volume, on rays that hit a surface, regardless of transmittance or albedo. This has a dramatic effect on convergence speed.

Note that this PR incorporates changes from another PR:
https://github.com/Unity-Technologies/Graphics/pull/2771


remi-chapelain (commented):
The issue about convergence seems to be fixed, I couldn't find a case where you couldn't converge properly with a dir light. ✔️ 

However, the more you increase the fog attenuation value, the more the results starts to diverge from what we can see in rasterization. 

From what I seem to understand in rasterization, increasing the Fog Attenuation Value decreases Fog density thus improves visibility though the fog. Basically, the higher the value, the less fog you see. 

But when using the Path Tracer, increasing this value past a certain point has little to no effects when there's geometry in front of you leaving you with a an unexpected result because the density seems to decrease overall but not everywhere; (cf GIF fog behind the first pillar stays basically the same, whereas in raster, fog is basically non existent at this point)

33306db6e9c9e4d5345380e8482963d1

Example with value ranging from 50 to 5000 (default value in template)

__Fog Attenuation Value__ **50**
Fog_Att50

__Fog Attenuation Value__ **500**
Fog_Att500

__Fog Attenuation Value__ **5000**
Fog_Att5000

As you can see, in raster, the fog basically disappear at one point.

eturquin:
The mismatch issue are more likely tied to the values of extinction and scattering I am getting from HDRP, rather than the sampling mode itself (unless these is a bug there, which is always possible...). But I'd think that it was less apparent before, because it was so noisy. I'm looking into it.

sebastienlagarde (approved):

sebastienlagarde:
@eturquin  / @remi-chapelain  What is the state of this PR? Should we merge this one and have further issue fix in another PR?

remi-chapelain:
Sampling issue is fixed, but now it revealed (meaning that it was already there but we couldn't see it as well) an underlying issue that @eturquin is investigating if the parameters we got from the volume are taken into account properly to see if that could help make the path tracer and the raster version diverge less with a high Attenuation value

remi-chapelain (approved):
After discussing with @motiejus-viltrakis and @emmanuelt, I think we're good to merge this fix and I'll create a fogbugz about the attenuation thing

PR #2848, opened 2020-12-04
Fixed Render Graph error caused by recent Shared Resource PR

JulienIgnace-Unity:
### Purpose of this PR
Fix an issue causing render graph errors in some particular circumstances caused by this PR https://github.com/Unity-Technologies/Graphics/pull/2546
Details:
We now store the output TextureHandle of shadow atlases. This is due to both the fact that we may need to keep it if the resource is shared (allocation happens only once) but also because creation of the resource when not shared can happen at two different times in the code. So we keep it around and check validity. Problems occurs if after being used for a few frame we start not using the atlas anymore then the handle would still be a valid one even though it does not match any valid resources.
That's why we now invalidate it ervery frame if it's not shared.

---
### Testing status
Tested locally: Force reproduction of the issue by removing the frame validity check in the code (this would delay the issue in the normal code) and added a directional light, removed it to make the cascade atlas invalid and enabled again. This would reproduce the issue, it did not occur anymore after the fix.
Also ran graphics tests.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.RenderGraph.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.RenderGraph.cs
@@ -209,6 +209,10 @@ namespace UnityEngine.Rendering.HighDefinition
 
         public void InvalidateOutputIfNeeded()
         {
+            // Since we now store the output TextureHandle (because we only want to create the texture once depending on the control flow and because of shared textures),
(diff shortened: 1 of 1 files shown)


FrancescoC-unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
Confirmed by @iM0ve that it was ok.

iM0ve:
Yes for the record, its a weird case where you keep editor running and within an hour there would be an error and rendering breaks. To test it I just left my project running for 3 days and the issue did not occur.

PR #2063, opened 2020-09-30
[Shader Graph] Suppress exception when deleting inactive nodes that are not in the node dictionary.

esmelusina:
---
### Purpose of this PR
Fix for: https://fogbugz.unity3d.com/f/cases/1259801/
Suppress exception when deleting inactive nodes that are not contained in the node dictionary. The exception is benign and not useful to the end-user in these circumstances.

---
### Testing status
Manual Testing: Bug does not repro as described in the bug report.

PR #2853, opened 2020-12-04
[Shader Graph] Suppress exception when deleting inactive nodes that are not in the node dictionary (remake).

esmelusina:
### Purpose of this PR
Fix for: https://fogbugz.unity3d.com/f/cases/1259801/
Suppress exception when deleting inactive nodes that are not contained in the node dictionary. The exception is benign and not useful to the end-user in these circumstances.

---
### Testing status
Manual Testing: Bug does not repro as described in the bug report.

---
### Comments to reviewers
This is a duplicate of : https://github.com/Unity-Technologies/Graphics/pull/2063 for srp2core considerations.


esmelusina:
Just an experiment testing against the automatic formatter in yamato.

PR #2863, opened 2020-12-07
[2021.2 / 12.x.x] Fix for Opaque Texture + MSAA being rendered black on old Apple GPUs

manuele-bonanno:
### Purpose of this PR
fix for [ISS case 1247423](https://fogbugz.unity3d.com/f/cases/1247423/).

new API SystemInfo.supportsStoreAndResolveAction (bump to 2021.1.0a10 min required version) [PR](https://ono.unity3d.com/unity/unity/pull-request/117384/_/platforms/mobile/store_and_resolve_action_api#).

Older Apple GPUs don't support ResolveAndStore store actions. Because of this we fall back to a Resolve store action, which discards the multisampled RT contents. Any following Load load action will then load a discarded RT, resulting in a black texture.

Because of this we need to disable MSAA when performing a separate Opaque Texture pass on hardware without StoreAndResolve support (i.e. Apple GPUs <= A8).

---
### Testing status
Tested locally using the fogbugz issue repro project.

**Before**:
image

**After**:
image


manuele-bonanno:
the reason why the "fixup" code was added to InitializeAdditionalCameraData is that only at that point we have requiresOpaqueTexture available for us to check.

Ideally though after discussing with @pbbastian we decided to see if the initialization code order can be rearranged so I can move the fixup code in CreateRenderTextureDescriptor. Working on it

manuele-bonanno:
All the test failures are mostly HDRP and VFX.
The Universal graphics test failures are also happening in master and are not related to this change

pbbastian (approved):
Code looks good to me :)

oleks-k (approved):
Looks good with my change to md file :)

phi-lira (approved):
Approved as workaround to fix this issue. Howevere this doesn't fix the case a render target switch happen other than the one for the opaque color copy. We should think of a solution for the long term. Moving discussion to slack.

vaidasma:
Verified as fixed with iPhone 5s (SoC: A7, OS: 12.4.8)

vaidasma (approved):

marctem (approved):

manuele-bonanno:
the formatting failure is because of:
" Needs formatting: com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/DecalSortingInputsUIBlock.cs"

which is not modified by this CS, so it is unrelated

ellioman:
Yamato issues not caused by this PR.

PR #2868, opened 2020-12-08
[ShaderGraph] Fix for blackboard property scrolling not moving the containing scroll view

Nightmask3:
### Purpose of this PR

Bugfix for 1293632: https://fogbugz.unity3d.com/f/cases/1293632/

This was a request from a partner game studio conveyed through DRM, making it an enterprise support task.

Before this bugfix, when dragging and dropping a blackboard property in the list to reorder it, there was an issue that would become most apparent when dealing with shader graphs that have a large number of properties.

Because the ScrollView visual element doesn't automatically handle the scrolling behavior when an item contained by it is being dragged around, you have to drag the property to the top by repeatedly dragging it and manually scrolling with the mouse, which is really bad UX.

This PR fixes that so the drag/drop behavior in a large list now looks like this:

5yT0Dy3qQF

There are now two more visual elements in the hierarchy of the Blackboard, which I call the scroll boundary regions. 

When the user starts a drag/drop operation on a blackboard property, these regions are shown, and upon the user entering either of the regions with the cursor while the drag operation is still in progress, the list scrolls in the appropriate direction.

The speed of the scroll is a constant right now, could be a user setting we expose and also increase/decrease based on the middle mouse scroll input, but would consult with Alex before adding something like that. 

In order to accomplish this, this PR introduces a couple of new classes like SGBlackboard and SGBlackboardSection to replace their GraphView equivalents. The SGBlackboard is also made a subclass of the GraphSubWindow class that already exists (and which inherits from GraphView)

This is needed because: 

- Right now, with how the BlackboardProvider wraps around the Blackboard and how the two are tightly coupled in various places due to the Blackboard class being in the GraphView module in trunk and hence not something we can easily edit, makes it hard to do things "the right way". Examples of this being how we need to add visual elements to the blackboard hierarchy through code, which defeats the purpose of the .UXML files and making the layouts of visual elements data-driven. 

- Adding an SGBlackboard class solves this issue by replicating all the functionality of the GV Blackboard, and wrapping around GraphSubWindow but ultimately providing us our own class that we can quickly iterate on for Shader Graph needs. 

- The Blackboard currently stands out in the Shader Graph due to how it is only resizable from one direction (the bottom). This PR fixes that due to the GraphSubWindow framework handling the resizing functionality, as well as the UXML, styling, editor layout persistence and scrolling behavior.

I58i2oZrkM

- The idea to create an SG specific Blackboard wrapper class around GraphView is not unique, it is in fact also seen in the VFX Graph code. And its what allows them to do things like blackboard categories, which is a feature that I will be working on soon. Hence it seemed to dovetail well. 

- The other motivator here is the coming transition to GTF, and by making the GraphSubWindow class the parent class with the GraphElement inheritance, it consolidates the GraphView dependencies in a single class for our graph sub-windows, which should make swapping it out with a GTF base class easier, as long as all the functionality GraphView provided is maintained/replaced by its GTF equivalents. 

This PR also fixes a few other existing issues:

- Deleting a property node while your mouse is over it leaves the property highlighted in the blackboard: https://fogbugz.unity3d.com/f/cases/1238635/

- Default Vector1 value in blackboard can't be edited via mouse drag: https://fogbugz.unity3d.com/f/cases/1202831/

- Fixed an issue where, when blackboard properties were dragged and then the user hit the "Escape" key, the drag indicator would still be visible 

- Fixed an issue where renaming blackboard properties through the Blackboard wouldn't actually change the underlying property name

---
### Testing status

- Ran manual tests (Edit Mode) in the ShaderGraph test project and everything passed.

- Tested drag/dropping for both Properties and Keywords to ensure everything was working as intended.

- Tested drag/dropping and various edge cases like hitting Escape after drag, ending the drag operation without dropping, or dropping onto the graph instead of scrolling to ensure the boundary regions always get properly hidden after a drag/drop operation is interrupted

- Tested drag/dropping with blackboard window of various sizes to make sure the behavior and visual appearance of the scroll boundary regions was consistent at all sizes

- Tested all possible blackboard operations like moving/adding/deleting/duplicating to ensure new SGBlackboard and SGBlackboardSection classes were replicating functionality of their GraphView equivalents with no regressions

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,23 +1,23 @@
 # Code Ownership File
 
 # Universal RP General
-/com.unity.render-pipelines.universal/ @unity/universal-rp-team
+/com.unity.render-pipelines.universal/ @Unity-Technologies/universal-rp-team
 
 # Universal RP 2D Renderer
-/com.unity.render-pipelines.universal/Editor/2D @unity/2d-graphics
-/com.unity.render-pipelines.universal/Runtime/2D @unity/2d-graphics
-/com.unity.render-pipelines.universal/Shaders/2D @unity/2d-graphics
-/com.unity.render-pipelines.universal/Editor/ShaderGraph/MasterNodes @unity/2d-graphics
+/com.unity.render-pipelines.universal/Editor/2D @Unity-Technologies/2d-graphics
+/com.unity.render-pipelines.universal/Runtime/2D @Unity-Technologies/2d-graphics
+/com.unity.render-pipelines.universal/Shaders/2D @Unity-Technologies/2d-graphics
+/com.unity.render-pipelines.universal/Editor/ShaderGraph/MasterNodes @Unity-Technologies/2d-graphics
 
 # Shader Graph
-/com.unity.shadergraph/ @unity/shader-graph
+/com.unity.shadergraph/ @Unity-Technologies/shader-graph
 
 # Test systems
-/.yamato/ @unity/gfx-sdets
-/katana-ci/ @unity/gfx-sdets
-/Tools/ @unity/gfx-sdets
-/com.unity.testframework.graphics/ @unity/gfx-sdets
-/.github/ @unity/gfx-sdets
+/.github/ @Unity-Technologies/gfx-sdets
+/.yamato/ @Unity-Technologies/gfx-sdets
+/katana-ci/ @Unity-Technologies/gfx-sdets
+/Tools/ @Unity-Technologies/gfx-sdets
+/com.unity.testframework.graphics/ @Unity-Technologies/gfx-sdets
 
 # LWRP (Legacy)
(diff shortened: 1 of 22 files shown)


cdxntchou (changes requested):
Issues:
* If you start dragging a property, drag it over the graph (like you're going to place a property node) then hit escape, the drag areas don't disappear -- this is minor, can fix in a separate PR.
* when you delete a property, it starts generating errors (and seems to break the UI) -- should fix this

rest looks good

Nightmask3:
> Issues:
> 
> * If you start dragging a property, drag it over the graph (like you're going to place a property node) then hit escape, the drag areas don't disappear -- this is minor, can fix in a separate PR if you want.
> * when you delete a property, it starts generating errors (and seems to break the UI) -- should fix this
> 
> rest looks good

Latest commit fixes these two issues, let me know what you think 👍

cdxntchou (approved):

PR #2869, opened 2020-12-08
Fixed FXAA when render-target has no alpha channel to store luminance.

kaychang-unity:
---
### Purpose of this PR
Fixed FXAA not working when the render-target has no alpha channel to store luminance.
Uber pass normally store luminance in alpha channel, which is then used in the final pass by FXAA.
This does not work on Switch which uses R11G11B10 as a HDR format or on platform where the quality setting use R11G11B10 as HDR format. I think the workaround for users has been to use the "keep alpha" checkbox, as it avoid storing luminance in alpha channel. the new fix make it "automatic", as well final pass does not need to sample alpha texture of the input texture again.

---
### Testing status
Ran a test project manually.
I don't expect new failure in gfx-tests (is this tested anywhere?).

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumeVoxelization.compute
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumeVoxelization.compute
@@ -120,8 +120,8 @@ void FillVolumetricDensityBuffer(PositionInputs posInput, uint tileIndex, Jitter
     // list of object indices. Note that while objects are already sorted by type, we don't know the
     // number of each type of objects (e.g. lights), so we should remember to break out of the loop.
     volumeCount = g_vBigTileLightList[MAX_NR_BIG_TILE_LIGHTS_PLUS_ONE * tileIndex];
-	// On Metal for unknow reasons it seems we have bad value sometimes causing freeze / crash. This min here prevent it.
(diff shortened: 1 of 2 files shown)


Chman (approved):

com.unity.postprocessing/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Fix for a compilation error in the Depth of Field shader on Linux.
+- Fix FXAA when running on render-targets without alpha channel: force using green channel as luminance
```
Chman, line 16: Note: this should go in 3.0.2, not 3.0.1 (which has already been released).

Chman (changes requested):

com.unity.postprocessing/CHANGELOG.md
kaychang-unity: Indeed, I moved to to 3.0.2.

kaychang-unity:
Should it be backported to earlier versions to 10.x ?

Chman:
It's not necessary to backport it, PPv2 packages are compatible back to 2018.4 anyway so people can easily upgrade when using previous versions of Unity.

Chman (approved):

kaychang-unity:
@Chman Could you merge the PR or should we wait for @sebastienlagarde to review it?

sebastienlagarde (approved):

PR #2877, opened 2020-12-09
[8.x.x] Fixed URP bug causing HDR skyboxes to render as LDR on XR

einaros:
### Purpose of this PR

Fixes bug introduced in f19a2240af8660e1bf89fde01a356d8d5e4a1d97, which resulted in HDR skyboxes (and more?) not rendering as such on XR. Caused by camera render texture descriptors not being assigned properly.

The change pretty much reintroduces a code block which was there prior to the mentioned commit.

### Steps to reproduce the issue

Import XR, URP, add an HDR skybox. Run and watch it not be rendered as HDR.


phi-lira:
We are not supporting 8.x.x packages anymore. You can upgrade to 10.x.x or master. If this issue is not fixed in master please open a PR to targetting master.

PR #2880, opened 2020-12-10
Universal/fix aces filter

kaychang-unity:
### Purpose of this PR
This fixes ACES tonemapping not correctly working on Nintendo Switch.
The generated LUT had artefacts due to fp16 precision.
I tested Android Vulkan devices (Google Pixel 2, Glaxy S9 on ARM GPU), as well as iOS Metal (Iphone8), but the issue does not happens on these platforms.

---
### Testing status
I ran test project from https://unity3d.atlassian.net/browse/TRND-1496.

---
### Comments to reviewers
This will not break any platform and affect Switch only by increasing precision of LUT calculation for ACES tonemapping, which will not introduce new bugs.

### DIFF
--- a/com.unity.render-pipelines.core/CHANGELOG.md
+++ b/com.unity.render-pipelines.core/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Version Updated
 The version number for this package has increased due to a version update of a related graphics package.
 
+### Fixed
+- Fixed ACES tonemaping for Nintendo Switch by forcing some shader color conversion functions to full float precision.
+
 ## [10.2.0] - 2020-10-19
 
 Version Updated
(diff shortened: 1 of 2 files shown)


kaychang-unity:
Do you need backporting for this bug fix? If yes, which versions?

com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
```diff
@@ -254,7 +254,11 @@ real YCoCgCheckBoardEdgeFilter(real centerLum, real2 a0, real2 a1, real2 a2, rea
 }
 
 // Converts linear RGB to LMS
+#if defined(SHADER_API_SWITCH) // Must use full float precision on Switch to avoid artefact when using ACES tonemapping
 real3 LinearToLMS(real3 x)
```
Wilfrid-Unity, line 258: Did you mean to use float3 line 258 too?

kaychang-unity: Yes you are right, thanks. Actually the function causing problem is LMSToLinear(), but I thought it was better to replicate the change to fp32 calculation on both functions.

Wilfrid-Unity (approved):

com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
```diff
@@ -254,7 +254,11 @@ real YCoCgCheckBoardEdgeFilter(real centerLum, real2 a0, real2 a1, real2 a2, rea
 }
 
 // Converts linear RGB to LMS
+#if defined(SHADER_API_SWITCH) // Must use full float precision on Switch to avoid artefact when using ACES tonemapping
```
eh-unity, line 257: Should these just be float3s always, won't the fp16 be a problem on any platform that defines the real3 as fp16?

eh-unity (commented):
I wonder why it isn't a problem on the other mobile devices? do they define real3 as float3 even if they support true half format?

Also this goes into the .core, so it probably needs HDRP team review as well.

sebastienlagarde (approved):

PR #2886, opened 2020-12-11
Fix tangentOS implicit truncation warning

garettbass:
### Purpose of this PR
When creating a new URP Lit shader graph, changing the input color in any way results in an HLSL warning, `implicit truncation of vector type`, in the generated `BuildVertexDescriptionInputs()` function, because `output.ObjectSpaceTangent` is of type `float3` and `input.tangentOS` is of type `float4`.  To avoid the warning, an `.xyz` swizzle is used to explicitly convert to `float3`.

Please note that while this warning may not affect behavior of the shader on PC, it can become an error when generating GLSL shaders for other platforms.  Ideally, warnings in shader graph shaders should only be produced by parts of the graph that were created by users.

---
### Testing status
I created a new URP Lit shader graph, and adjusted the input color value.  I observed that the generated shader reported a warning, `implicit truncation of vector type`, in `BuildVertexDescriptionInputs()`.  I fixed the template, `BuildVertexDescriptionInputs.template.hlsl`, then reimported the shader graph, and the warning no longer appeared.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this package are documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [11.0.1] - 2020-10-21
+
+### Added
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


garettbass:
I would appreciate help passing the Labeler check.

alindmanUnity:
migrated change to internal PR to bypass issues with testing #2974

PR #2890, opened 2020-12-14
Universal/bugfix/runtime render pass fix

Nzollner:
### Purpose of this PR
This PR fixes the render passes not being able to be enqueued at runtime.
Which is requested by this bug case: https://fogbugz.unity3d.com/f/cases/1211195/
And to show the feature working is a sample implemented which also helps the user.
---
### Testing status
I tested this PR locally on Windows 10 using DX11 on UniversalGraphicsTest.
I have also tested the feature manually when creating the sample using the fix.
---
### Comments to reviewers
This PR is 1 out of 2 PRs where the first contains the fix and the second contains the sample code using this fix.


phi-lira (approved):

Nzollner:
The only error are a non-formatted HDRP code which I haven't touched but lies in master.

PR #2897, opened 2020-12-14
Sg/swizzle node UI alignment

tracychen1234:

### Checklist for PR maker

---
### Purpose of this PR
Update current Swizzle node UI for better alignment with other graph tools.
image

---
### Testing status
Tested with different vector size input values and mask input edge cases.
Succeed in both local Paly Mode/Edit Mode tests (reconnected the swizzle node in tests).

---
### Comments to reviewers


LandonTownsendUnity (changes requested):
We can't just replace nodes like this - a shadergraph using the old version of the swizzle node will not work correctly when upgraded to this branch. Either the node needs to upgrade to an equivalent version of the new node (which may be difficult to code and to test) or another solution may be making it an entirely new node, like "Swizzle inline". Since this is feature work, we may want to discuss how we should do this.

com.unity.shadergraph/Editor/Data/Nodes/Channel/SwizzleNode.cs
```diff
@@ -22,151 +24,105 @@ public SwizzleNode()
...
+        //2.mask cant be character other than "xyzw"
+        //3.If the input is not valid, wont genetate shader code and give errors
+        public bool ValidateMaskInput(int InputValueSize)
         {
-            get { return m_GreenChannel; }
-            set
+            bool MaskInputIsValid = true;
+            char[] MaskChars = _maskInput.ToCharArray();
+            char[] AllChars  = {'x', 'y' , 'z', 'w'};
+            List<char> CurrentChars = new List<char>();
```
cdxntchou, line 53: Instead of copying from the AllChars arrayr into the CurrentChars list, you can simply initialize the CurrentChars list to be what you want:
List<char> CurrentChars = new List<char>() { 'x', 'y', 'z', 'w' };

LandonTownsendUnity (changes requested):
Swizzle node also lacks a preview now, which the earlier version had:
image

com.unity.shadergraph/Editor/Data/Nodes/Channel/SwizzleNode.cs
```diff
@@ -22,151 +24,105 @@ public SwizzleNode()
...
-
-        [SerializeField]
-        TextureChannel m_AlphaChannel;
-
-        [ChannelEnumControl("Alpha Out")]
-        public TextureChannel alphaChannel
-        {
-            get { return m_AlphaChannel; }
-            set
+            if (MaskChars.Length == 0 || MaskChars.Length > 4)
```
cdxntchou, line 67: Even though it's not strictly necessary for Length, I generally like to do tests that will catch errors for any possible value.  So in this case:
if (MaskChars.Length <= 0 || MaskChars.Length > 4)

tracychen1234:
> We can't just replace nodes like this - a shadergraph using the old version of the swizzle node will not work correctly when upgraded to this branch. Either the node needs to upgrade to an equivalent version of the new node (which may be difficult to code and to test) or another solution may be making it an entirely new node, like "Swizzle inline". Since this is feature work, we may want to discuss how we should do this.

Working on backward compatibilities

tracychen1234:
> Swizzle node also lacks a preview now, which the earlier version had:
> image

Took out the preview window for better alignment with vfx graph, also we now have a Preview node so I think it's fine to take it off?

LandonTownsendUnity:
> > Swizzle node also lacks a preview now, which the earlier version had:
> > image
> 
> Took out the preview window for better alignment with vfx graph, also we now have a Preview node so I think it's fine to take it off?

I think VFX graph lacks a preview because the nature of how VFX graph works, node previews don't really work... but I think the node should have a preview in Shadergraph

com.unity.shadergraph/Editor/Data/Nodes/Channel/SwizzleNode.cs
```diff
@@ -22,151 +24,107 @@ public SwizzleNode()
...
-            });
-        }
-
-        public override void CollectPreviewMaterialProperties(List<PreviewProperty> properties)
-        {
-            base.CollectPreviewMaterialProperties(properties);
-            // Encode swizzle values into an integer
-            var value = ((int)redChannel) | ((int)greenChannel << 2) | ((int)blueChannel << 4) | ((int)alphaChannel << 6);
-            properties.Add(new PreviewProperty(PropertyType.Float)
+                owner.AddValidationError(objectId, "Invalid mask!", ShaderCompilerMessageSeverity.Error);
```
marctem, line 109: No exclamation points in error messages please. It's not exciting =). Can we give them a hint what would be valid instead? (Like, "Invalid Mask. Valid values are r, g, b.")

com.unity.shadergraph/Documentation~/Swizzle-Node.md
```diff
@@ -17,15 +18,12 @@ Channel dropdown parameters are dynamic depending on the length of the input vec
...
 
 ## Generated Code Example
 
 The following example code represents one possible outcome of this node.
 
 ```
-float4 _Swizzle_Out = In.xzyw;
-```
\ No newline at end of file
+float4 _Swizzle_Out = float4 (In.y,In.z,In.w,In.x);
```
marctem, line 29: Is the float cast here slower than the old explicit swizzle code?

com.unity.shadergraph/Editor/Data/Nodes/Channel/SwizzleNode.cs
```diff
@@ -22,151 +24,107 @@ public SwizzleNode()
...
+                sb.AppendLine(string.Format("{0} {1} = float4 (0, 0, 0, 0);", outputSlotType, outputName));
```
marctem, line 110: Does this work with the precision controls?

com.unity.shadergraph/Editor/Data/Nodes/Channel/SwizzleNode.cs
```diff
@@ -22,151 +24,107 @@ public SwizzleNode()
...
+                    if (i != 0)
+                        outputValue += ",";
+
+                    if (i < _maskInput.Length)
+                        outputValue += inputValue + "." + _maskInput[i];
+
+                    if (i >= _maskInput.Length)
+                        outputValue += "0";
+                }
+                sb.AppendLine("{0} {1} = float4 ({2});", outputSlotType, outputName, outputValue);
```
marctem, line 126: Same question about precision.

marctem (changes requested):

com.unity.shadergraph/Documentation~/Swizzle-Node.md
```diff
@@ -17,15 +18,12 @@ Channel dropdown parameters are dynamic depending on the length of the input vec
 
 | Name        | Type           | Options  | Description |
 |:------------ |:-------------|:-----|:---|
-| Red out      | Dropdown | Red, Green, Blue, Alpha (depending on input vector dimension) | Defines which input channel should be used in the output's red channel |
-| Green out      | Dropdown | Red, Green, Blue, Alpha (depending on input vector dimension) | Defines which input channel should be used in the output's green channel |
-| Blue out      | Dropdown | Red, Green, Blue, Alpha (depending on input vector dimension) | Defines which input channel should be used in the output's blue channel |
-| Alpha out      | Dropdown | Red, Green, Blue, Alpha (depending on input vector dimension) | Defines which input channel should be used in the output's alpha channel |
+| Mask     | Inputfield | X, Y, Z, W (depending on input vector dimension) | The swizzle mask is a combination of one to four characters that can be x, y, z, or w. The size of output value depends on the length of the mask input.|
```
Nightmask3, line 21: I'm also able to use r, g, b, a as mask inputs. So would add a mention of that here.

Nightmask3 (approved):
This looks good! Tested it out a decent amount and can't find any issues.

Also really like how the new swizzle node works, can totally see myself using it a lot. 
Really good stuff Tracy! 👍

Nightmask3:
> > > Swizzle node also lacks a preview now, which the earlier version had:
> > > image
> > 
> > 
> > Took out the preview window for better alignment with vfx graph, also we now have a Preview node so I think it's fine to take it off?
> 
> I think VFX graph lacks a preview because the nature of how VFX graph works, node previews don't really work... but I think the node should have a preview in Shadergraph

I do agree that an on-the-same-node preview would be really good to have for quick preview of what the swizzle operation output was, as otherwise I'd have to connect to a graph output or preview node which adds a few more clicks. 

While testing this out I found myself wanting a preview to be able to confirm that the values were in fact changing as I modified the mask.

Buuut we do also have a preview node for a reason I guess so I can see arguments for going either way. 

I guess that does raise the question as to why we would even have a preview node if all nodes had their own previews, so thats a potential reason to exclude the preview from the swizzle node. 

The previews do add some visual noise to the graph as a whole, so that's another motivation to exclude them.

tracychen1234:
> > > > Swizzle node also lacks a preview now, which the earlier version had:
> > > > image
> > > 
> > > 
> > > Took out the preview window for better alignment with vfx graph, also we now have a Preview node so I think it's fine to take it off?
> > 
> > 
> > I think VFX graph lacks a preview because the nature of how VFX graph works, node previews don't really work... but I think the node should have a preview in Shadergraph
> 
> I do agree that an on-the-same-node preview would be really good to have for quick preview of what the swizzle operation output was, as otherwise I'd have to connect to a graph output or preview node which adds a few more clicks.
> 
> While testing this out I found myself wanting a preview to be able to confirm that the values were in fact changing as I modified the mask.
> 
> Buuut we do also have a preview node for a reason I guess so I can see arguments for going either way.
> 
> I guess that does raise the question as to why we would even have a preview node if all nodes had their own previews, so thats a potential reason to exclude the preview from the swizzle node.
> 
> The previews do add some visual noise to the graph as a whole, so that's another motivation to exclude them.
> 
> Ideally we'd just be able to toggle previews on every node as the user needed so maybe we could just post-pone this discussion till we can circle back on a more graph-wide convention.

Added preview. Thanks for the feedback, Sai! I didn't realize that VFX graph lacking preview is because of their nature, thanks Landon! The preview on the node does serve as a role of providing positive/negative feedback and I feel it's quite important for a node like swizzle to give visual feedback. And I also agree with Sai that too many previews might cause some visual noise. But the user can always collapse the on-node preview window (sounds less effort comparing to plug in a preview node every time). Also asked Alex, she's good with adding a preview.

com.unity.shadergraph/Editor/Data/Nodes/Channel/SwizzleNode.cs
tracychen1234: Good point! Changed to `sb.AppendLine(string.Format("{0} {1} = 0;", outputSlotType, outputName));`

com.unity.shadergraph/Editor/Data/Nodes/Channel/SwizzleNode.cs
tracychen1234: Took off the exclamation mark. And for the error message what do you think about this?
Screenshot (4)

LandonTownsendUnity (changes requested):
Needs to validate that you are not mixing and matching rgba with xyzw. After recent changes, doing this causes compile errors. Also, I think it might be a nice improvement to usability to automatically convert the input string to lowercase so that users won't be confused if they try something like "XYZ" and it doesn't work.
image

tracychen1234:
https://github.com/Unity-Technologies/Graphics/pull/3040 is the correct one

PR #2899, opened 2020-12-14
Fixed cmd.Blit disables SPI shader keyword

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR
- Fixed cmd.Blit disables SPI shader keyword when it blits from non XR texture to another non XR texture.

---
### Testing status
- ABV

---
### Comments to reviewers
Fixed the issue by correctly configure builtin property SinglePassStereoMode. 
cmd.Blit has logic to resume to original stereo mode. If not set, the function resumes stereo mode to none and disables shader keyword.


phi-lira:
Can you share more details about this PR. It seems like using cmd.Blit might break XR in some cases and we fix URP side. How do we prevent users to do the same thing and break URP?

PR #2900, opened 2020-12-15
Core RP preview camera background color

martint-unity:
---
### Purpose of this PR
Fixing the issues people are experiencing with a darker background for the previews.
https://fogbugz.unity3d.com/f/cases/1297865/

Also just moving the gizmo color into a sub menu and make sure we can reset gizmo color

---
### Testing status
Opened up both URP and HDRP and created Prefabs and checking previews that the color is the default gray.

---
### Comments to reviewers


com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
```diff
@@ -27,26 +36,61 @@ public static Color volumeGizmoColor
             }
         }
 
+        public static Color previewBackgroundColor
```
madsnyholm, line 39: This Preview background color cannot be a User Preference.
See discussion https://unity.slack.com/archives/C0A9NNY9W/p1608024382240300?thread_ts=1607848679.236200&cid=C0A9NNY9W

madsnyholm (changes requested):
Please remove the Preview background color from Preferences and the Settings Provider. It needs to be a const color, since there are no decencies setup that will ensure all assets are reimported if the background color changes. 
See discussion https://unity.slack.com/archives/C0A9NNY9W/p1608024382240300?thread_ts=1607848679.236200&cid=C0A9NNY9W

madsnyholm (approved):
Looks good.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

RSlysz (approved):
lgtm

TomasKiniulis (commented):
Taking a look on HDRP side. If I update an older project where preview was black then prefab still holds black background even after doing a reimport, same goes to Fbx and material has to be clicked to change background. Would it be possible to have this fixed in the same PR? 

upgradedProjectConvert

martint-unity:
@TomasKiniulis You will have to do a Re-import all for this to take affect on old projects.

erikabar (approved):
Tested Material, Mesh, Prefab, Model Project preview and Inspector preview backgrounds, the issue is fixed.

TomasKiniulis (approved):
Approving this, although I see a potential issue where users might be confused why background colors vary between old assets and newly created ones on project upgrade without "Reimport All", but there doesn't seem to be other way to resolve this

phi-lira (approved):

PR #2902, opened 2020-12-15
Moving post process data back to renderer

lukaschod:
### Purpose of this PR
Bring back post process data back to renderer.

Reason:
Currently all the renderer properties are independent from urp asset and only during the actual rendering urp supplies it option. With post processing stripping all the post process data is kept in urp asset this way breaking this concept.
For this reason I and Felipe decided that maybe we should keep post process stripping option it in renderer after all. This would also make sense more if we would decide to move renderer passes into features.
It was uncovered with PR https://github.com/Unity-Technologies/Graphics/pull/2833.

---
### Testing status
- [ ] Run gfx tests locally.
- [ ] Check if stripping did not regressed.

---
### Comments to reviewers
As from functionality wise nothing changed much, I simply removed path that was taking urp asset post process data as renderer post process pass was still working.

Foward Renderer
image

Renderer2D
image

When post processing disabled data property is hidden
image


phi-lira (approved):

rainsing (approved):

Verasl (approved):
LGTM

simon-engelbrecht-soerensen (approved):
Looks good. Only issue I found was that upgrading with PP enabled in URPAsset, will not be enabled in renderer after upgrading, and might lead to some confusion with users. Not sure if feasible to do anything about however, therefore the approval.

TheoWong-pixel:
UI looks fine for now, spoke with @lukaschod to indent the data field to make it more clear the relationship between the enable options and the data option.  Generally speaking we want to always show all the available fields and grey them out when disabled instead of popping them in, but seems like this is but best option considering the backend.

TheoWong-pixel (approved):
UI looks fine for now, spoke with @lukaschod to indent the data field to make it more clear the relationship between the enable options and the data option. Generally speaking we want to always show all the available fields and grey them out when disabled instead of popping them in, but seems like this is but best option considering the backend.

PR #2912, opened 2020-12-16
Minor improvement to ssgi

anisunity:
- Changed the convergance time of ssgi to 16 frames.
- Changed the preset values for ssgi.
- Fixed an issue with half res ssgi upscale.
ContentExample070
ContentExample071

Trying to improve a bit the quality of the SSGI.

Testing status.
Checked the result on the botanist scene and it is an improvement.
This requires a bit more testing from QA on more various scenes.


sebastienlagarde (approved):

remi-chapelain (approved):
The fix on half resolution is a good catch. I managed to repro and verified that nothing unexpected happens after the fix. ✔️ 

I tested the rest of the improvement on various effects and overall, it's more noticeable by default with the new presets and less subtle. ✔️ 

I also tested ghosting on a moving object because the convergence time has been doubled but the difference between before and after is barely noticeable. ✔️ 

However, after a discussion with @anisunity, I'd like to take advantage of this PR to modify the name and the tooltip of the "object thickness" parameter (since it also have been increased) because, it's confusing from a user perpective and makes users thinks that the more thickness there is, the less rays will be "valid", whereas it's the opposite. 

I can suggest this "**depth buffer tolerance**", or even just "**tolerance**". 
"Sets the tolerance for the screen space global illumination rays to pass behind objects. A greater value will allow more rays to be valid, increasing the intensity of the effect."

Also, there's no tooltip for the filter radius !

PR #2916, opened 2020-12-16
Changed the message when the graphics device doesn't support ray tracing (case 1287355).

anisunity:
https://fogbugz.unity3d.com/f/cases/1287355/
Changed the message to be different depending if the api is dx12 or not.

Testing status 
Tested on a dx11 project.


remi-chapelain (approved):
Given the scope, I'd say testing is sufficient for this one

sebastienlagarde (approved):

PR #2920, opened 2020-12-16
Fixed a bug in the sphere-aabb light cluster (case 1294767).

anisunity:
https://fogbugz.unity3d.com/f/cases/1294767/
There was a tiny issue in building the light cluster, this fixes it.

Testing status
The issue doesn't repro anymore on the shared dataset
Local DXR tests are still green.


remi-chapelain (approved):
Given the specific edge case of the issue, looks like testing that it does not repro with the dataset is sufficient ✔️

sebastienlagarde (approved):

PR #2923, opened 2020-12-17
Yamato - Fix editor pinning branch

theopnv:
### Purpose of this PR

Today's 2021.1/staging nightly used 2021.2 because the job responsible for updating the pinned revision of the editor in our metafiles failed. This is because the editor-pinning branch as specified in the old configuration (`2021.1/staging/ci/editor-pinning`) could not be created. Because the branch `2021.1/staging` already exists, and git can't handle both a branch and a folder named similarly.

So the solution is to name the editor-pinning branch differently:
`2021.1/staging/ci/editor-pinning` becomes ` 2021.1/ci/editor-pinning`.

I also created the ` 2021.1/ci/editor-pinning` branch on Github: https://github.com/Unity-Technologies/Graphics/tree/2021.1/ci/editor-pinning

### DIFF
--- a/.yamato/_editor_pinning.yml
+++ b/.yamato/_editor_pinning.yml
@@ -8,7 +8,7 @@
 ---
 
 editor-pinning-target-to-ci:
-    name: Merge 2021.1/staging to 2021.1/staging/ci/editor-pinning
+    name: Merge 2021.1/staging to 2021.1/ci/editor-pinning
     agent:
         type: Unity::VM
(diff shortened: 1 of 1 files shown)


theo-pnv:
Merging with no review, being the only SDET available today.
This PR has low risks of worsening the situation, given that the 2021.1 nightly doesn't work in the current state anyway.
I will keep an eye on this and revert if anything goes wrong.

PR #2926, opened 2020-12-17
Fixed the planar depth texture not being properly created and rendered to (case 1299617).

anisunity:
https://fogbugz.unity3d.com/f/cases/1299617/

Testing status
The issues doesn't repro in release mode and in play mode (also hapenned sometimes)
Currently running the hdrp tests locally


sebastienlagarde (approved):

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Works as expected now.

Rough Reflections are displayed properly in release/debug builds and editor after build. Also checked that building release with rough reflections disabled and only then enabling rough reflections still displays them correctly

PR #2929, opened 2020-12-17
Fixed an issue when trying to open a look dev env library when Look Dev is not supported.

JulienIgnace-Unity:
### Purpose of this PR
Fixed an issue when trying to open a look dev env library when Look Dev is not supported.
https://fogbugz.unity3d.com/f/cases/1285104/

The button is now greyed out as it should be when Look Dev is not supported.

---
### Testing status
Tested the repro steps available in the case.


sebastienlagarde (approved):

RSlysz (approved):

remi-chapelain (approved):
On behalf of @TomasKiniulis , trusting the developer on this one !

PR #2930, opened 2020-12-17
[HDRP] Change the behavior of custom passes when the volume is disabled

alelievr:
### Purpose of this PR
Fix this case: https://fogbugz.unity3d.com/f/cases/1300022/

To fix the domain reload issue, I changed the behavior of the custom pass volume:  
Before the fix, disabling a custom pass volume, did not really impact the passes in the volume, they were just not executed.
Now with this fix, the function `Cleanup()` is called when the `Custom Pass Volume` component is disabled, which allows releasing the resources allocated in the passes.

Note that it shouldn't change the execution behavior of the passes.

---
### Testing status
Tested that the Setup and Cleanup function of the custom passes are executed correctly when:  
- domain reload happens
- enter play mode
- disable gameobject
- disable custom pass volume component


pmavridis (approved):
Looks good. I have double-checked that OnDisable is called every time we trigger a destroy, so the change makes sense.

sebastienlagarde (approved):

PR #2933, opened 2020-12-17
Fixed shader graph not supporting indirectdxr multibounce (case 1294694)

anisunity:
The keyword was missing from shader graphs which made it is malfunction.
https://fogbugz.unity3d.com/f/cases/1294694/

Testing status
Reimported the shader and the multibounce works properly now.


sebastienlagarde (approved):

alelievr (approved):

remi-chapelain (approved):
Quickly tested all HDRP material in SG and "Regular" to be sure. 
Made me think that it would be a good idea to test this behavior in at least one of our test scene. 
I'll try and modify one to integrate this when I have time.

PR #2936, opened 2020-12-17
[ShaderGraph] Fix property reference names to not use GUIDs

cdxntchou:
### Checklist for PR maker

---
### Purpose of this PR
When adding properties to the shadergraph blackboard, they are given "reference" names (names used in the HLSL code, and as identifiers for manipulating them in C# script) that include a GUID, which looks like a large string of random numbers to the user:

image

As a result, these names are not very user friendly, and we've received a lot of complaints about constantly having to change the reference names to something simpler and more readable.

With this PR, any newly created properties will instead use the new reference naming scheme, which keeps the reference name as close to the display name as possible, while maintaining it as a unique and valid HLSL identifier:

image

For any pre-existing properties, users can opt to upgrade the default naming scheme, via right click menu on the reference label:
image

This PR also adds the "Reset Reference" option (that removes overrides) to the same right click menu.
It also adds a "bold" style to the label when overridden:
image

This PR also addresses a bug where boolean keywords that set override ref names would not show up in the Material inspector.  https://fogbugz.unity3d.com/f/cases/1300454/  

---
### Testing status

- [x] Tested that pre-existing properties and keywords (of all types) maintain their pre-existing reference name, whether using an override or the default name.
- [x] Tested that pre-existing properties can be reset to default name, and upgraded to use the new default naming scheme, via the right click menus on the ref name label.
- [x] Tested that newly created properties are created using the new default naming scheme, and can be overridden and reverted to default via the right click menus.
- [x] Tested creating new properties via the blackboard "+" menu with different names, duplicate names, default names.  Always resulted in properties with non-duplicate display names, and non-duplicate reference names.
- [x] Tested modifying display names of existing properties, attempting to get duplicate reference names via different means (undo/redo, ref name overrides on some properties and not others, resetting overrides) -- no detected way to get duplicate reference names.
- [x] Tested copy/pasting properties from other graphs, with unique or colliding names, using default or override on the ref name.  No issues with duplicate ref names in the results.  ** Copies are NOT upgraded to use new naming scheme. **
- [x] Tested duplicating properties on the blackboard -- names are made unique.  ** Copies are NOT upgraded to use new naming scheme. **   ** COPIES OF DEFAULT REF NAMES MAY OVERRIDE THE REF NAME **
- [x] Tested drag/drop properties from other graphs, with unique or colliding names, using default or override on the ref name.  No issues with duplicate ref names in the results.  ** Copies are upgraded to use new naming scheme. **
- [x] Tested ConvertToProperty on float node, vector4 node -- result is correct and no duplicate naming.
- [ ] Tested Keyword reference names pass all of the same tests as properties above.
    ** KEYWORD DON'T BOLD OVERRIDES, DONT HAVE RIGHT CLICK MENU ***
- [ ] Tested that modifying display names can undo/redo correctly, including the default reference name change.
- [ ] Tested that modifying reference name (override) can undo/redo correctly
- [ ] Tested that resetting the reference name (removing override) can undo/redo correctly
- [ ] Tested that upgrading the a pre-existing property's default reference name (using new naming convention) can undo/redo correctly.
- [ ] Tested that new default property reference names are reflected in the generated shaders correctly, and work.
- [ ] Verified that new default property reference names are serialized to disk correctly.
- [ ] Tested that new default named keywords are reflected in the generated shaders correctly, and work.
- [ ] Verified that new default keyword reference names are serialized to disk correctly.
- [ ] Test Reset Reference on blackboard property
- [ ] Test blackboard property scrolling moving the containing scroll view

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
@@ -1284,6 +1284,32 @@ namespace UnityEditor.ShaderGraph
             }
         }
 
+        public string DeduplicateGraphInputReferenceName(ShaderInput input, string newName)
+        {
+            if (string.IsNullOrEmpty(newName))
+                newName = "_";
+
+            string name = newName.Trim();
+            if (string.IsNullOrEmpty(name))
+                name = "_";
+
+            if (Regex.IsMatch(name, @"^\d+"))
+                name = "_" + name;
+
+            name = Regex.Replace(name, @"(?:[^A-Za-z_0-9])|(?:\s)", "_");
+
+            switch (input)
+            {
+                case AbstractShaderProperty property:
+                    return GraphUtil.DeduplicateName(properties.Where(p => p != property).Select(p => p.referenceName), "{0}_{1}", name);
+                case ShaderKeyword keyword:
+                    name = name.ToUpper();
+                    return GraphUtil.DeduplicateName(keywords.Where(p => p != input).Select(p => p.referenceName), "{0}_{1}", name.ToUpper());
+                default:
+                    throw new ArgumentOutOfRangeException();
+            }
+        }
+
         public void SanitizeGraphInputReferenceName(ShaderInput input, string newName)
         {
             if (string.IsNullOrEmpty(newName))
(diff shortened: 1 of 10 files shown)


LandonTownsendUnity:
Do we really think it's safe to make this change while Unity doesn't have a system to protect the reference name to being set to a reserved word, and causing errors?

cdxntchou:
Made a new branch for this.  New PR here:   https://github.com/Unity-Technologies/Graphics/pull/3232

PR #2939, opened 2020-12-17
Fix light for-loops when running on WebGL1

brendan-duncan:
---
### Purpose of this PR
Additional lights were not working with WebGL1 due to the use of a variable "lightCount" used in the light iteration for-loop.
For WebGL1 only, use a constant for-loop. This checks for a new shadercompiler define, UNITY_PLATFORM_WEBGL, to isolate the change to just WebGL. If UNITY_PLATFORM_WEBGL is not defined, the changes will have no affect.

---
### Testing status
Manual tested multiple lights on WebGL1 and non-WebGL1 platforms.

---
### Comments to reviewers


com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -783,8 +789,14 @@ half3 VertexLighting(float3 positionWS, half3 normalWS)
...
 #ifdef _ADDITIONAL_LIGHTS_VERTEX
     uint lightsCount = GetAdditionalLightsCount();
+    #ifdef _USE_WEBGL1_LIGHTS
+    for (int lightIndex = 0; lightIndex < _WEBGL1_MAX_LIGHTS; ++lightIndex)
+    {
+        if (lightIndex >= lightsCount) break;
+    #else
     for (uint lightIndex = 0u; lightIndex < lightsCount; ++lightIndex)
     {
+    #endif
```
phi-lira, line 799: Suggestion, since this code is repeated in a few places and users might want to write custom shaders that work on WebGL 1, maybe we could provide a macro, LIGHT_LOOP_BEGIN(lightsCount) / LIGHT_LOOP_END or something similar?

phi-lira (changes requested):
Missing changelog.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
brendan-duncan: Changed the code to use LIGHT_LOOP_BEGIN/END macros

phi-lira (approved):

PR #2947, opened 2020-12-18
Fixed an issue with the frame count management for the volumetric fog (case 1299251).

anisunity:
https://fogbugz.unity3d.com/f/cases/1299251/
After a recent change related to the frame count management. The Frame count was not updated properly anymore in the scene view editor.  Which lead to a various set of issues related to the volumetrics. By using the camera frame count we guarantee that the frame count is incremented at every render with that camera.

Testing status
The issue doesn't repro locally anymore.
Local tests are green, except for 9402 AlphaToMask which doesn't have to do anything with the code changed...

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -30,19 +30,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed computation of geometric normal in path tracing (case 1293029).
 - Fixed issues with path-traced volumetric scattering (cases 1295222, 1295234).
 - Fixed the default background color for previews to use the original color.
-- Fixed AOV api in render graph (case 1296605)
-- Fixed a small discrepancy in the marker placement in light intensity sliders (case 1299750)
-- Fixed issue with VT resolve pass rendergraph errors when opaque and transparent are disabled in frame settings.
-- Fixed a bug in the sphere-aabb light cluster (case 1294767).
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
```diff
@@ -255,6 +255,12 @@ static internal void SafeDestroy(ref RenderTexture rt)
             }
         }
 
+        static uint VolumetricFrameIndex(HDCamera hdCamera)
+        {
+            // Here we do modulo 14 because we need the enable to detect a change every frame, but the accumulation is done on 7 frames (7x2=14)
+            return hdCamera.GetCameraFrameCount() % 14;
```
sebastienlagarde, line 261: don't we have any constant value somewhere with this 7 ? Afraid to have hardcoded value hidden in a function like this.

anisunity: right now, no when the thing is accumulation frame index is computed its explicitely written frameIndex % 7

adrien-de-tocqueville (approved):
looks good, but the frame count management in HDRenderPipeline is still broken, so we probably still have problem with other effects using accumulation like TAA

TomasKiniulis (approved):
Tested repro project and template scene with Always Refresh toggled On/Off, and various Scene/Game view layouts in Playmode and edit mode. Seems to be working correctly now

sebastienlagarde (approved):

PR #2953, opened 2020-12-18
First WIP introduction of Adaptive Probe Volumes. 

FrancescoC-unity:
This PR introduce a first iteration over APV, it is not yet in a usable state, just to have some core code in master. 
APV stands for Adaptive Probe Volume, it replaces Probe Volumes v1.  It works essentially by adaptively placing probes around the scene, subdividing as required. 

The sampling code is simpler as it is a single lookup on the indirection structure and finally samples the SH coefficients from a storage. 

The PR includes: 

 - Basic workflow (might contain bugs) with asset and profile production. 
 - Basic baking and placement (will see improvements) 
 - Lightloop sampling of the APV structure
 - Volumetrics sampling.
 - Material AO support. 
 - L0 and L1 sampling code (everything does L1, volumetrics do L0)
 - Some debug visualization. 
 - Multiple scene handling for additive loading of different assets. 
 
The loading in player needs still to be handled with user-side script, but possible with the simple asset that gets produced by the baking process. 

A lot is meant to be changed in a near future, so the usage is not recommended. 
It is of course turned off by default.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
@@ -219,22 +219,10 @@ void EvaluateAdaptiveProbeVolume(in float3 posWS, in float3 normalWS, in float3 
         bakeDiffuseLighting, backBakeDiffuseLighting);
 }
 
-// Without a normal we only evaluate L0, also we have an option to undo the constant coefficients that are there due to the cosine convolution.
-void EvaluateAdaptiveProbeVolume(in float3 posWS, bool undoCosineRescale, out float3 bakeDiffuseLighting)
+void EvaluateAdaptiveProbeVolume(in float3 posWS, out float3 bakeDiffuseLighting)
 {
     APVResources apvRes = FillAPVResources();
     bakeDiffuseLighting = EvaluateAdaptiveProbeVolumeL0(posWS, float3(0.0f, 0.0f, 0.0f), apvRes);
-    if (undoCosineRescale)
-    {
-        float invC0 = 3.54490770181f; // 1 / (0.5 * sqrt(1/Pi))
-        bakeDiffuseLighting *= invC0;
-    }
 }
-
-void EvaluateAdaptiveProbeVolume(in float3 posWS, out float3 bakeDiffuseLighting)
-{
-    EvaluateAdaptiveProbeVolume(posWS, false, bakeDiffuseLighting);
-}
-
 
 #endif // __PROBEVOLUME_HLSL__
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
@@ -470,8 +470,9 @@ float3 EvaluateVoxelDiffuseGI(PositionInputs posInput, JitteredRay ray, float t0
     posInput.positionWS = ray.originWS + t * ray.jitterDirWS;
 
     float3 apvDiffuseGI;
-    const bool undoCosineRescaling = true; // We want to undo the effect of the convolution with cosine kernel.
-    EvaluateAdaptiveProbeVolume(GetAbsolutePositionWS(posInput.positionWS), undoCosineRescaling, apvDiffuseGI);
+    // Note here we don't need undo the convolution by the cosine kernel as we only sample L0 and don't support anisotropy yet, so undoing cosine and convolve with phase function
+    // for L0 term it'd just cancel out as both are just constant factors. 
+    EvaluateAdaptiveProbeVolume(GetAbsolutePositionWS(posInput.positionWS), apvDiffuseGI);
 
     // It is possible that some invalid probes are sampled due to how APV is laying the probes
     // For now the safest way is to ignore the NaN data when sampled.


FrancescoC-unity:
All tests pass locally.

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2020.1-to-2020.2.md
```diff
@@ -203,6 +203,8 @@ In the Decal shader, the property `_MetallicRemapMin` have been added, and `_Met
 
 From 10.x, a new pass ScenePickingPass have been added to all the shader and master node to allow the editor to correctly handle the picking with tesselated objects and backfaced objects.
 
+From 10.x, if material ambient occlusion needs to be applied to probe volume GI and the material is deferred, the material needs to define `HAS_PAYLOAD_WITH_UNINIT_GI` constant and a function `float GetUninitializedGIPayload(SurfaceData surfaceData)` that returns the AO factor that is desired to be applied. No action is needed for forward only materials or if no material AO needs to be applied to probe volume GI.
```
sebastienlagarde, line 206: you can remove this statement, 
1) it is not the right location (upagre 2020.1 to 2020.2)
2) we haven't ship the feature, so we don't care.

com.unity.render-pipelines.high-definition/Runtime/Debug/InstancedProbeShader.shader
```diff
@@ -0,0 +1,97 @@
...
+    {
+        Tags { "RenderType" = "Opaque" }
+        LOD 100
+        Pass
+        {
+            CGPROGRAM
+            #pragma vertex vert
+            #pragma fragment frag
+            #pragma multi_compile_instancing
+            #include "UnityCG.cginc"
```
sebastienlagarde, line 17: We must convert this shader to new style SRP way, i.e don't use UnityCg.cginc. Can be done in next PR

com.unity.render-pipelines.high-definition/Runtime/Debug/InstancedProbeShader.shader
```diff
@@ -0,0 +1,97 @@
...
+
+                o.vertex = UnityObjectToClipPos(v.vertex * _ProbeSize);
+                o.normal = UnityObjectToWorldNormal(v.normal);
+
+                return o;
+            }
+
+            float3 evalSH(float3 normal, float4 SHAr, float4 SHAg, float4 SHAb)
+            {
+                float4 normalPadded = float4(normal, 1);
```
sebastienlagarde, line 60: not sure why you don't reuse the functoin provided in EntityLighting.hlsl?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -899,6 +896,10 @@ void InitializeLightLoop(IBLFilterBSDF[] iBLFilterBSDFArray)
                 Shader.EnableKeyword("SCREEN_SPACE_SHADOWS_OFF");
             }
 
+            if (ShaderConfig.s_EnableProbeVolumes == 1)
+            {
+                ProbeReferenceVolume.instance.InitProbeReferenceVolume(1024, m_Asset.currentPlatformRenderPipelineSettings.probeVolumeMemoryBudget, new Vector3Int(1024, 64, 1024));
```
sebastienlagarde, line 901: what are the hardcoded value 1024 and other? (can be fix in next pr)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -494,62 +494,61 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
 #if SHADEROPTIONS_ENABLE_PROBE_VOLUMES == 1
-    bool uninitialized = IsUninitializedGI(builtinData.bakeDiffuseLighting);
-    builtinData.bakeDiffuseLighting = uninitialized ? float3(0.0, 0.0, 0.0) : builtinData.bakeDiffuseLighting;
-
+    bool uninitializedGI = IsUninitializedGI(builtinData.bakeDiffuseLighting);
     // If probe volume feature is enabled, this bit is enabled for all tiles to handle ambient probe fallback.
-    // No need to branch internally on _EnableProbeVolumes uniform.
-    if (featureFlags & LIGHTFEATUREFLAGS_PROBE_VOLUME)
+    // Even so, the bound resources might be invalid in some cases, so we still need to check on _EnableProbeVolumes.
+    bool apvEnabled = featureFlags & LIGHTFEATUREFLAGS_PROBE_VOLUME && _EnableProbeVolumes;
```
sebastienlagarde, line 500: use (featureFlags & LIGHTFEATUREFLAGS_PROBE_VOLUME) && _EnableProbeVolumes for clarity

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -494,62 +494,61 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+    if (!apvEnabled)
     {
-#if !SHADEROPTIONS_PROBE_VOLUMES_ADDITIVE_BLENDING
-        if (uninitialized)
-#endif
+        builtinData.bakeDiffuseLighting = (uninitializedGI && !apvEnabled) ? float3(0.0, 0.0, 0.0) : builtinData.bakeDiffuseLighting;
```
sebastienlagarde, line 503: So we need to define something here. what disabling APV mean?
For mean disabling APV mean we fallback on the previous probe system (and not to black)
So disabling APV shouldn't require this code above. It should just be when we sample probe, we use lightprobe or put an uninitializedGI  mark depends if APV is enabled or not.

sebastienlagarde: nevermind , misread the code :P

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
```diff
@@ -1,484 +1,243 @@
...
-#endif
 
-void ProbeVolumeComputeOBBBoundsToFrame(OrientedBBox probeVolumeBounds, out float3x3 obbFrame, out float3 obbExtents, out float3 obbCenter)
+float3 EvaluateAmbientProbe(float3 normalWS)
 {
-    obbFrame = float3x3(probeVolumeBounds.right, probeVolumeBounds.up, cross(probeVolumeBounds.right, probeVolumeBounds.up));
-    obbExtents = float3(probeVolumeBounds.extentX, probeVolumeBounds.extentY, probeVolumeBounds.extentZ);
-    obbCenter = probeVolumeBounds.center;
+    real4 SHCoefficients[7];
+    SHCoefficients[0] = unity_SHAr;
```
sebastienlagarde, line 72: We should create a function with this one, the exact same code is used in builtinGIUtilities.hlsl (next PR)

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
```diff
@@ -1,484 +1,243 @@
...
+        bakeDiffuseLighting += L0;
+        backBakeDiffuseLighting += L0;
     }
     else
-#endif
     {
-        float3 sampleOutgoingRadiance = SHEvalLinearL0L1(normalWS, coefficients.data[0], coefficients.data[1], coefficients.data[2]);
-        return sampleOutgoingRadiance;
+        // no valid brick, fallback to ambient probe
+        bakeDiffuseLighting = EvaluateAmbientProbe(normalWS);
```
sebastienlagarde, line 190: See my comment above, same code is use for both front and back lighting in builtinGIUtilitiers.hlsl

com.unity.render-pipelines.high-definition/Runtime/Debug/InstancedProbeShader.shader
FrancescoC-unity: To answer this and the before, @thefranke  lifted this code from another project altogether. We indeed need to rewrite it anyway to be more SRP complaint.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
FrancescoC-unity: The first 1024 is just how much the pool allocator uses as size for the brick pool. It is a value that we need to tweak and see on profiling if it makes a difference changing it, realistically we just don't need it as a parameter of the Init function, but would need investigation whether it makes sense for different pipelines (not different projects) to  have different values. 

Note that the value is not really that hardcoded, as in it is a fixed number in a constructor, but then the value itself is correctly named in the rest of the code. 


The second set is again just whatever is the default for the index dimension of the reference volume. It will be overwritten at first load if it mismatches what's on the profile. We can give it a name and store as static somewhere indeed.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
```diff
@@ -457,6 +460,31 @@ VoxelLighting EvaluateVoxelLightingLocal(LightLoopContext context, uint groupIdx
...
+{
+#if SAMPLE_PROBE_VOLUMES
+    float tOffset, weight;
+    ImportanceSampleHomogeneousMedium(rndVal, extinction, dt, tOffset, weight);
+
+    float t = t0 + tOffset;
+    posInput.positionWS = ray.originWS + t * ray.jitterDirWS;
+
+    float3 apvDiffuseGI;
+    EvaluateAdaptiveProbeVolume(GetAbsolutePositionWS(posInput.positionWS), apvDiffuseGI);
```
sebastienlagarde, line 473: Random question: I guess EvaluateAdaptiveProbeVolume evaluate the lighting with the lambert consine convolution coefficient in it. We should remove it to apply on volumetric. Similar to what is done with Ambient probe in SetPreconvolvedAmbientLightProbe() function

FrancescoC-unity: Good point

com.unity.render-pipelines.high-definition/Runtime/Debug/InstancedProbeShader.shader
FrancescoC-unity: This wasn't considered a blocker for opening the PR though.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
FrancescoC-unity: This the function :) 
I just need to move this to the builtinGIUtilities and use there.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -207,6 +207,12 @@ float GetAmbientOcclusionForMicroShadowing(BSDFData bsdfData)
     return sourceAO;
 }
 
+#define HAS_PAYLOAD_WITH_UNINIT_GI
+float GetUninitializedGIPayload(SurfaceData surfaceData)
```
sebastienlagarde, line 211: I feel bad about this one, we will need to add it to all our lit material (hair, fabric, eye, staacklit...). But guess we have no choice. Problem is very similar to 
GetAmbientOcclusionForMicroShadowing().
Let's let it for now for the experimental development, but not sure we have alternative. At least we should change the name to GetAmbientOcclusionFromSurface() so we can reuse it for other purpose

sebastienlagarde (approved):
Some random comment

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
FrancescoC-unity: We don't  need to add it to everything that's why I added the define, we only need to add it to deferred materials.

FrancescoC-unity: The reason why I avoided using AmbientOcclusion in that name is exactly to allow potential reuse for something else.

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2020.1-to-2020.2.md
sebastienlagarde: I now understand why it is there, should be in 2021.1 to 2021.2 upgrading guide (that we haven't created)

PR #2955, opened 2020-12-18
Project skybox without perspective for ortho cameras

adrien-de-tocqueville:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1264647

project the sybox in a similar way to what builtin does for orthographic cameras
Alternative to #1936, which is less correct but more usable

the result is now independant of the camera size.

### Testing status

compared result with what's on builtin

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
@@ -140,14 +140,14 @@ namespace UnityEngine.Rendering.HighDefinition
         internal static float ProjectionMatrixAspect(in Matrix4x4 matrix)
             => - matrix.m11 / matrix.m00;
 
-        internal static Matrix4x4 ComputePixelCoordToWorldSpaceViewDirectionMatrix(float verticalFoV, Vector2 lensShift, Vector4 screenSize, Matrix4x4 worldToViewMatrix, bool renderToCubemap, float aspectRatio = -1)
+        internal static Matrix4x4 ComputePixelCoordToWorldSpaceViewDirectionMatrix(float verticalFoV, Vector2 lensShift, Vector4 screenSize, Matrix4x4 worldToViewMatrix, bool renderToCubemap, float aspectRatio = -1, bool isOrthographic = false)
         {
             Matrix4x4 viewSpaceRasterTransform;
-            aspectRatio = aspectRatio < 0 ? screenSize.x * screenSize.w : aspectRatio;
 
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
```diff
@@ -142,33 +142,46 @@ internal static float ProjectionMatrixAspect(in Matrix4x4 matrix)
...
         internal static Matrix4x4 ComputePixelCoordToWorldSpaceViewDirectionMatrix(float verticalFoV, Vector2 lensShift, Vector4 screenSize, Matrix4x4 worldToViewMatrix, bool renderToCubemap, float aspectRatio = -1)
         {
+            Matrix4x4 viewSpaceRasterTransform;
             aspectRatio = aspectRatio < 0 ? screenSize.x * screenSize.w : aspectRatio;
 
-            // Compose the view space version first.
-            // V = -(X, Y, Z), s.t. Z = 1,
-            // X = (2x / resX - 1) * tan(vFoV / 2) * ar = x * [(2 / resX) * tan(vFoV / 2) * ar] + [-tan(vFoV / 2) * ar] = x * [-m00] + [-m20]
-            // Y = (2y / resY - 1) * tan(vFoV / 2)      = y * [(2 / resY) * tan(vFoV / 2)]      + [-tan(vFoV / 2)]      = y * [-m11] + [-m21]
+            if (verticalFoV < 0)
```
sebastienlagarde, line 148: @adrien-de-tocqueville  could you add an explicit argument instead of hacking with verticalFoV ? :)

i.e ComputePixelCoordToWorldSpaceViewDirectionMatrix(.... , bool isOrthographic = false).

and be a bit more versose on the comment:
// For ortho cameras, project the skybox with no perspective
(maybe add fobugzz link, thanks)

adrien-de-tocqueville: I replicated the aspectRatio hack haha, but ok

sebastienlagarde: oh haven't noticed it :).  but let's be cleaner

TomasKiniulis (approved):
Checked multiple scene views with variety of field of view settings. Looks good! Thanks for the change, nice that we have it consistent between pipelines now.
https://user-images.githubusercontent.com/50582134/106606989-71cb7f80-656b-11eb-9f43-48d8b987ba28.mp4

PR #2956, opened 2020-12-18
Fixed various Look Dev issues after exiting Playmode

JulienIgnace-Unity:
### Purpose of this PR
This PR fixes a number of issues with the Look Dev tool after exiting playmode.
Original issue can be found here: https://fogbugz.unity3d.com/f/cases/1293262/

The root of the problem is that the Look Dev stores its state in serialized field in order to be able to have a persistent layout on domain reload etc.
The issue is that the reference to this state was stored in many different places and after exiting playmode it would become invalid but still accessible (because Unity does weird things like this). Basically resetting the Look Dev in whatever last correct serialized state it was in before entering playmode. This is why for example the layout could reset to single view after exiting playmode.

After fixing this issue, there were many consequences like HDRI not updating, comparison gizmo not working etc. I tracked and fixed all the issues like this I found.

---
### Testing status
After finding out that most issues were after exiting playmode I tested all this after doing so:
- Tested the original issue
- Tested changing HDRI in view 1 or 2 / side by side etc
- Moving the camera
- Changing current objet
- Comparison gizmo

---
### Comments to reviewers
@Unity-Technologies/gfx-qa-hdrp  as I had to change quite a bit of internal code I think it would be good to have a thorough pass on the Look Dev to make sure everything is still ok.
We especially need to test everything AFTER exiting playmode in particular as this is where most issues occured.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/LookDev/CameraController.cs
+++ b/com.unity.render-pipelines.core/Editor/LookDev/CameraController.cs
@@ -131,7 +131,6 @@ namespace UnityEditor.Rendering.LookDev
 
         public CameraController(DisplayWindow window, Action focused)
         {
-            m_CameraState = null;
             m_Window = window;
             m_Focused = focused;
         }
@@ -466,8 +465,6 @@ namespace UnityEditor.Rendering.LookDev
         public SwitchableCameraController(DisplayWindow window, Action<ViewIndex> focused)
             : base(window, null)
         {
-            m_FirstView = null;
-            m_SecondView = null;
             m_CurrentViewIndex = ViewIndex.First;
 
             m_Focused = () => focused?.Invoke(m_CurrentViewIndex);
--- a/com.unity.render-pipelines.core/Editor/LookDev/Compositor.cs
+++ b/com.unity.render-pipelines.core/Editor/LookDev/Compositor.cs
@@ -154,7 +154,6 @@ namespace UnityEditor.Rendering.LookDev
             StageCache stages)
         {
             m_Displayer = displayer;
-            m_Contexts = null;
 
             m_RenderDataCache = new RenderingData[2]
             {
--- a/com.unity.render-pipelines.core/Editor/LookDev/LookDev.cs
+++ b/com.unity.render-pipelines.core/Editor/LookDev/LookDev.cs
@@ -27,8 +27,6 @@ namespace UnityEditor.Rendering.LookDev
         /// </summary>
         internal static Context currentContext
         {
-            //get => s_CurrentContext ?? (s_CurrentContext = LoadConfigInternal() ?? defaultContext);
-
             //Lazy init: load it when needed instead in static even if you do not support lookdev
             get
             {


com.unity.render-pipelines.core/Editor/LookDev/CameraController.cs
```diff
@@ -129,13 +129,18 @@ virtual protected bool isDragging
             }
         }
 
-        public CameraController(CameraState cameraState, DisplayWindow window, Action focused)
+        public CameraController(DisplayWindow window, Action focused)
         {
-            m_CameraState = cameraState;
+            m_CameraState = null;
```
RSlysz, line 134: This is optional as the default value should be null

com.unity.render-pipelines.core/Editor/LookDev/CameraController.cs
```diff
@@ -458,16 +463,24 @@ class SwitchableCameraController : CameraController
...
 
-        public SwitchableCameraController(CameraState cameraStateFirstView, CameraState cameraStateSecondView, DisplayWindow window, Action<ViewIndex> focused)
-            : base(cameraStateFirstView, window, null)
+        public SwitchableCameraController(DisplayWindow window, Action<ViewIndex> focused)
+            : base(window, null)
         {
-            m_FirstView = cameraStateFirstView;
-            m_SecondView = cameraStateSecondView;
+            m_FirstView = null;
+            m_SecondView = null;
```
RSlysz, line 470: same

com.unity.render-pipelines.core/Editor/LookDev/Compositor.cs
```diff
@@ -150,17 +150,16 @@ public bool pixelPerfect
 
         public Compositer(
             IViewDisplayer displayer,
-            Context contexts,
             IDataProvider dataProvider,
             StageCache stages)
         {
             m_Displayer = displayer;
-            m_Contexts = contexts;
+            m_Contexts = null;
```
RSlysz, line 157: same

com.unity.render-pipelines.core/Editor/LookDev/LookDev.cs
```diff
@@ -27,8 +27,21 @@ internal static IDataProvider dataProvider
         /// </summary>
         internal static Context currentContext
         {
+            //get => s_CurrentContext ?? (s_CurrentContext = LoadConfigInternal() ?? defaultContext);
```
RSlysz, line 30: Should be removed

RSlysz (approved):
Looks good to me

sebastienlagarde (approved):

JMargevics (changes requested):
Hey, so I've tried this branch and I can't assign any Env. Libraries on it. Same works fine on Master. 

❌ This branch:
https://user-images.githubusercontent.com/50581102/102788155-de6b3000-43aa-11eb-9e24-91a4675489cb.mp4

✔ Master:
https://user-images.githubusercontent.com/50581102/102788160-e0cd8a00-43aa-11eb-8a21-76524993e0df.mp4

PR #2962, opened 2020-12-18
[2021.2] [ShaderGraph] Disabling failing test for IsNaN node

cdxntchou:
### Purpose of this PR
Fixing the ShaderGraph tests on OSX metal by disabling the test of the IsNAN node:

image

IsNaN and general NaN behavior is surprisingly platform / GPU / compiler specific, and needs much further investigation to solve properly.

In the meantime we can just kill the test.

---
### Testing status
Ran ShaderGraph CI tests:

OSX metal:  🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252Fdisable-isnan-test/.yamato%252Fshadergraph-osx-metal.yml%2523ShaderGraph_OSX_Metal_playmode_mono_Linear_trunk/4662488/job

Full shaderGraph PR job: 🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252Fdisable-isnan-test/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/4666866/job/pipeline

---
### Comments to reviewers

### DIFF
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/IPhonePlayer/Metal/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/IPhonePlayer/Metal/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6c5b6b0ef820edc743de84e2588af3620f1ad3081be0a96c4fb3a2805e3d5651
-size 7322
+oid sha256:205146432f93b8afaf4f9e3385da2146ed243e91db3a3dbfbe04ed2cbbb84c6e
+size 6913


esmelusina (approved):

jessebarker (approved):
Looks good (especially on Yamato)

LandonTownsendUnity (changes requested):
I don't like the action of just making the test return green - it makes it very unclear to anyone looking at the scene that the test is not actually being run. I think the proper way to do this is to disable the gameobject the test material is on so that it can be clear to someone looking at the scene that we don't have coverage on that node. This will require updating the reference images but I think it's the right way to do it.

LandonTownsendUnity (approved):
PR needs to make it in to get CI green again and we don't have time to wait for a proper fix, so approving despite concerns. Chris said he will make 2 tickets, one for fixing the IsNaN node and one for properly disabling the test if we can't get the node fixed.

PR #2970, opened 2020-12-21
Fix Clearcoat on Stacklit breaks when URP is imported into the project

sebastienlagarde:
### Purpose of this PR
Fix case https://fogbugz.unity3d.com/f/cases/1297806/

This PR fix an issue when a Material with a cloat is created in HDRP or Universal and the other pacakages is imported as well.
It seems that creating a new BlockField which have same name in both pipeline cause conflict. Sharing this blockfield solved the issue. Note that the issue is about both stacklit and lit and for both CoatMask and CoatSmoothness.

It seems that a general rules, we must share BlockField that have similar name if we want to avoid conflict between the two pipelines, not sure why.

Adding Jennifer for visibiltiy about cross pipeline issue.

---
### Testing status
Out side of the repro step of the bug I have check that both stackLit (left) and Lit (right) are working correctly. And that both CoatMask and CoatSmoothness work correctly.

Before the fix the generated code between working version and non working version (i.e after import of URP) only differ by the CoatMask field. 

image

image

When importing URP we get this

After the fix when importing URP we get this
image

And the generated code match.

---
### Comments to reviewers


jenniferd-unity (approved):
I dont know enough about CodeFields to judge that portion of the code, but looks fine from a general perspective

TomasKiniulis (approved):
Validating that all Coat options are working with StackLit and Lit with and without universal package added. No pr related issues seems to occur.

Verasl (approved):
LGTM

elizabeth-legros (approved):

PR #2971, opened 2020-12-21
[2021.1] Workaround Android GLES3 failing tests

tomzig16:
### Purpose of this PR

This is backport for #2904 

Android GLES3 tests are failing due to known issue: https://fogbugz.unity3d.com/f/cases/1297080/ . This regression was introduced with the certain gpu recorder fix from the profiler team. Enabling "Frame Timing Stats" fixes this issue as with it enabled Unity does not use GPU recorder (which was regressed in previously mentioned bug).
This PR only enables "Frame Timing Stats" settings in Android player settings, so no other platform should be affected.

---
### Testing status
Manual + Yamato tests on GLES3
ABV

---
### Comments to reviewers
This is temporary change, once it is fixed this PR will be reversed.
I have changed only one setting but for some reason more changes to the `ProjectSettings.asset` file were made (targeted change is on line 140). ABVs are green so I suspect its due to different Unity version that was opened with?


theo-pnv (approved):

theo-pnv:
Approved but it seems there is a conflict on `TestProjects/UniversalGraphicsTest/Assets/Test/TestFilters/TestCaseFilters.asset`.
I reran the pipeline to get rid of the instabilities.

PR #2982, opened 2020-12-22
Fixed an issue with material using distortion from ShaderGraph init after Material creation (case 1294026)

sebastienlagarde:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1294026/

---
### Testing status
Tested the case describe in the fogbugzz, works now

---
### Comments to reviewers


TomasKiniulis (approved):
Looks good. Checked repro and new created distortion unlit/lit shadergraph , no issues occur anymore either when creating material from shadergraph selected or setting it manually from material

PR #2986, opened 2020-12-23
[SRP] Uniformise all ShaderGraph/SubShaderGraph and FBX extension to be lower case

sebastienlagarde:
### Purpose of this PR
 This PR uniformize all extensions of assets in SRP repository to be lower case so it works correctly with Linux and other case sensitive platform as we tend to mix and match shadergraph and ShaderGraph extension.
In the C# code it change any mixed or uppercase hardcoded extension as lower case.

---
### Testing status
Automatic test should be enough to validate that change is safe.

---
### Comments to reviewers


theo-pnv (approved):

SteveTheodore (approved):
Seems to hit all the files in the package

ellioman (approved):

elizabeth-legros (approved):

PR #2989, opened 2020-12-23
Update 5001_Fog_FogFallBack scene to correct bad behavior

anisunity:
Change the way the test 5001_Fallback was done. It was using a single camera and rendering multiple times during a single frame different images which is not compatible with the way HDRP handles its history system.

This restores the previous screenshot behavor, but given that the cameras have moved by epsilon we need to regenerate screenshots for this tests on all platforms.

Done with @remi-chapelain


sebastienlagarde (approved):

remi-chapelain (approved):
Already pair-tested with @anisunity

PR #2996, opened 2021-01-06
Per-stage shader keywords in HDRP

FrancescoC-unity:
With this PR HDRP will use per-stage keywords what that means is that we can do stuff like `multi_compile_fragment` to have only the fragment shader variant computed. It is very important to define one separate line per stage that needs to include the keyword. This work for both multi compiles and shader_features 

This is has to be done manually for code shaders (I did a pass on our shaders, being conservative to be safe) and I added a field to have that easily generated in SG. 

This reduces the number of compiled shader programs with all the benefits that come from that. 

I ran test manually and they pass.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit.shader
@@ -209,9 +209,9 @@ Shader "HDRP/TerrainLit"
 
             // In deferred, depth only pass don't output anything.
             // In forward it output the normal buffer
-            #pragma multi_compile_fragment _ WRITE_NORMAL_BUFFER
+            #pragma multi_compile _ WRITE_NORMAL_BUFFER
             #pragma multi_compile_fragment _ WRITE_DECAL_BUFFER
-            #pragma multi_compile_fragment _ WRITE_MSAA_DEPTH
+            #pragma multi_compile _ WRITE_MSAA_DEPTH
 
             #define SHADERPASS SHADERPASS_DEPTH_ONLY
             #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLitTemplate.hlsl"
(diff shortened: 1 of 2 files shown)


com.unity.shadergraph/Editor/Data/Util/KeywordUtil.cs
```diff
@@ -65,11 +66,57 @@ public static string ToDeclarationString(this KeywordDefinition keywordDefinitio
...
+            if ((keyword.stages & KeywordShaderStage.Geometry) == KeywordShaderStage.Geometry)
+                dstStages.Add(KeywordShaderStage.Geometry);
+            if ((keyword.stages & KeywordShaderStage.RayTracing) == KeywordShaderStage.RayTracing)
+                dstStages.Add(KeywordShaderStage.RayTracing);
+
+            return dstStages.Count > 1;
+
+        }
+
+        // TODO: BROKEN FIX!!!!
```
alelievr, line 92: what is broken? :)

alelievr (approved):
LGTM!

(with the TODO fixed)

com.unity.shadergraph/Editor/Data/Util/KeywordUtil.cs
FrancescoC-unity: Oops this escaped my cleanup :D Will remove immediately!

cdxntchou (approved):

sebastienlagarde (approved):

sebastienlagarde (approved):

sebastienlagarde:
Adding QA: QA please try to load a project like SpaceShip or Palace, and build a player a check that everything is fine. Thanks

iM0ve (approved):
I have upgraded Amalienborg Palace project and also built a Player. Did not notice any visual issues or any new console errors

image

sebastienlagarde (approved):

LandonTownsendUnity (approved):
Just made sure that URP still works properly when HDRP is in project.

PR #2997, opened 2021-01-06
fix broken renderer data inspector

martint-unity:
---
### Purpose of this PR
Fixed the issue that caused the Renderer Data inspector to break after adding Renderer Object and another Renderer Feature after that.

Fogbugz case
https://fogbugz.unity3d.com/f/cases/1301085/


com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -75,6 +75,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue where MSAA did not work in Editor Game View on Windows with Vulkan.
 - Fixed issue where selecting and deselecting Forward Renderer asset would leak memory [case 1290628](https://issuetracker.unity3d.com/issues/urp-scriptablerendererfeatureeditor-memory-leak-while-interacting-with-forward-renderer-in-the-project-window)
 - Fixed the default background color for previews to use the original color.
+- Fixed an issue where inspector of Renderer Data would break after adding Renderer Object renderer feature and then adding another renderer feature.
```
phi-lira, line 78:
```suggestion
- Fixed an issue where the inspector of Renderer Data would break after adding RenderObjects renderer feature and then adding another renderer feature.
```

phi-lira (approved):
LGTM. Minor changelog suggestions

martint-unity:
Previous test run had known failures that are failing in master and not because of this PR
Test started again because of changelog update and I will turn them off to not waste bandwidth

PR #3000, opened 2021-01-06
Move APV to Core and remove shader config dependency + auto load in player without external script

FrancescoC-unity:
This PR does a couple of things: 

 - Remove the dependency on shader config. Substituted it with a multi_compile in shader. 
 - Added a multi compile for L1 and L2, at the moment L2 doesn't do anything different, it is just a setup for when we'll implement. 
 - Move practically all files to core. 
 - Added warning in editors if anything but HDRP is used. 
 - Changed visibility for several things. 
 - Added some documentation for stuff that is public. 
 - Made it so the asset gets loaded directly from the component that is used in editor. Previously the system needed an external per-project script, now is not needed. 
 - Deleted some code that was dead.


sebastienlagarde (approved):

thefranke (approved):
Nice work, thanks!

PR #3003, opened 2021-01-06
Fix for android input node tests

cdxntchou:
There were issues with android test failing with compression changes to textures, as well as an instability with one test. Updated instability and made textures uncompressed. Updated all reference images.

Backported https://github.com/Unity-Technologies/Graphics/pull/3053


jessebarker (approved):
The new images look good.  Hopefully the yamato results are consistent run-to-run with these.

PR #3019, opened 2021-01-07
Fix issue with shadow mask and area lights

FrancescoC-unity:
The issue stems from the fact that we kill the shadow mask when NdotL <= 0, which is completely valid for punctual lights, but less so for area lights in real time where we don't have a single valid light direction for the whole thing (being an area there are a lot of valid directions). 

Fogbugz: https://fogbugz.unity3d.com/f/cases/1303523/ and https://fogbugz.unity3d.com/f/cases/1300426/


sebastienlagarde (approved):

remi-chapelain (approved):
Issue is fixed. 
Tested on the data set with all type of light to be sure, that nothing broke ✔️

PR #3020, opened 2021-01-07
Changed the clamping approach for RTR and RTGI (in both perf and quality) to improve visual quality.

anisunity:
Do to the very small budget of samples and execution time that we have, we need to clamp the values before denoising. The previous strategy for denoising de-saturated the values a bit too much. This PR changes the strategy for reflections and GI improving singificantly the visual quality for low clamp values.

image

Testing status
Tested with a bunch of lighting setups, made sure that the changes made on the tests made sense and updated the screenshots


remi-chapelain (approved):
Tested on custom projects with RTGI and RTR
Great improvement especially visible when using a low clamp value where some color and hue information were lost/changed before this PR ✔️

sebastienlagarde (approved):

PR #3023, opened 2021-01-07
Disabled specular occlusion for what we consider medium and larger scale rtao > 1.25 with a 25cm falloff interval.

anisunity:
Specular occlusion is currently applied on indirect specular no matter what. This PR disables specular occlsuion on RTAO for wider ambient occlusion RayLength > 1.5meters.
https://fogbugz.unity3d.com/f/cases/1304315/

Testing status
Ran the tests locally, made sure the updated screenshots made sense and tested ssao vs rtao


iM0ve (changes requested):
The issue is fixed. However, I found one concern.

Maybe it is expected, but It seems that the Debug mode now does not match the Ambient occlusion fix. Do we have to make the change for the debug mode too? If this is intended, let me know and I will approve.

**Concern**
6lByVQ34DR

Before PR:
Before

After PR:
After

**Whats tested:**
* Tested if the issue in fb was addressed
* Debug mode
* Other areas of the scene to make sure AO is still working as intendeded

sebastienlagarde (approved):

sebastienlagarde:
>Maybe it is expected, but It seems that the Debug mode now does not match the Ambient occlusion fix. Do we have to make the >change for the debug mode too? If this is intended, let me know and I will approve.
Hey, yes it is expected and in particular, the SSAO mode don't show what the specular occlusion is.

iM0ve (approved):
Approving since the concern I found is expected. No other issues

sebastienlagarde:
@anisunity This PR fail on standalone 001 template, I don't understand why. For now reverted from hd/bugfix

sebastienlagarde:
ok, fixed default value and re-integrated the PR

PR #3025, opened 2021-01-08
fix blackboard nullref when duplicating node with blackboard closed

elizabeth-legros:
There was a nullref when checking selection in MaterialGraphView.GetIndiciesToInsert. Added a nullcheck to avoid checking size on a null IEnumerable.

Verified the issue does not replicate with after fix

https://fogbugz.unity3d.com/f/cases/1294430


Nightmask3 (approved):
Looks good, thanks for the fix! Can't imagine this has any side effects. 👍

PR #3027, opened 2021-01-08
Fixed the condition on temporal accumulation in the reflection denoiser (case 1303504).

anisunity:
After a recent changes that depending on the smoothness of a surface changes the strategy of rtr denoising, the conditions to reject history was wrong. this PR fixes it.
https://fogbugz.unity3d.com/f/cases/1303504/

Testing status
Tried to repro the issue on the dataset after the fix, can't repro.
DXR tests are green locally.


remi-chapelain (approved):
Tested on a bigger project (the one that helped found this issue) and this fixes it ✔️

sebastienlagarde (approved):

PR #3028, opened 2021-01-08
Changed the name from the Depth Buffer Thickness to Depth Tolerance for SSGI (case 1301352).

anisunity:
https://fogbugz.unity3d.com/f/cases/1301352/


remi-chapelain (approved):
Clearer that way ! ✔️ 
Would have liked a longer tooltip explaining but it's too advanced to explain without over-simplifying and/or being misleading.

sebastienlagarde (approved):

PR #3032, opened 2021-01-08
Change the source value for the ray tracing frame index iterator from m_FrameCount to the camera frame count (case 1301356).

anisunity:
Due to recent changes, the m_FrameCount is updated once for every rendered camera image. This changes the source iterator from m_Frame count to camera frame count for correct scene view results and solves the issue in the fogbugz
https://fogbugz.unity3d.com/f/cases/1301356/

Testing status
We do not see the artifacts anymore on the test
DXR tests are green locally


remi-chapelain (approved):
testing seems sufficient on this one ✔️

sebastienlagarde (approved):

PR #3039, opened 2021-01-08
Fix a flickering issue related to moving shadow receivers (case 1302392) and a refactoring to avoid computing the same value for every effect.

anisunity:
This PR includes twp seperate changes:
- Rejecting the history when the receiver is moving (as the signal will be changing on the target obejct) (case 1302392)
- Doing the history validation pass once for the whole frame and using it for every effect instead of doing it for every effect.

Testing status 
I ran the DXR tests locally they are all green.
Opened couple scenes and made sure the reprojection looks fine.(Shadows, AO, RTGI)
Unfortunately, the tests do not capture temporal denoising this would require a bit more manual testing from @remi-chapelain to make sure it behaves normally after the PR.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
@@ -649,7 +649,7 @@ namespace UnityEngine.Rendering.HighDefinition
 
         internal HDTemporalFilter GetTemporalFilter()
         {
-            if (m_TemporalFilter == null)
+            if (m_TemporalFilter == null && m_RayTracingSupported)
             {
                 m_TemporalFilter = new HDTemporalFilter();
                 m_TemporalFilter.Init(m_Asset.renderPipelineRayTracingResources);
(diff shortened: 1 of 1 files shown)


remi-chapelain (commented):
Tested multiple settings for the 3 raytraced effects concerned (RTGI, RT Shadows, RTAO) in the repro project and in another bigger real world project.

The ghosting seems to have completely disappeared when the "receiver" is moving in RTGI / RTAO and RT Shadows except RT Area Light shadow. It's the only case that does not seem affected by this fix. (see gif)
88beecb0e878e4fcabe75675c9561f52

Despite ghosting being fixed, I have the feeling it's maybe a bit to "agressive" on discarding the history, is this possible to fine tune the threshold for history validity ?
I've noticed that the history gets discarded instantly if the receiver is moving, even if we cannot see it moving with out eyes. I think it would be a bit better to increase a bit the tolerance to be able to NOT discard history if it's moving very slowly, some ghosting can be tolerated especially if it's moving really really slowly. 
I tried fiddling around with the constant in TemporalFilter compute shader but not luck 

You can see the difference between before the PR and after

**Before** 
before_low

**After**
after_low

anisunity:
No, not possible to "tune" it, we don't have access to that piece of information

remi-chapelain (approved):
Concerns about unaffected area lights won't be fixed in this PR, it will be the object of future work. 
As for the very low moving threshold, it can introduce some new issues, (unwanted/unjustified noise if the receiver has a very low velocity basically) but I believe this is still an improvement compared to previous implementation. 

Approving ✔️

sebastienlagarde (approved):

anisunity:
Unfortuantely can't be a trade off. Either it ghosts or it doesn't

PR #3041, opened 2021-01-08
[ShaderGraph] [bugfix 1304501] Fix for ShaderGraph hang, and slightly better handling of exceptions

cdxntchou:
### Purpose of this PR
Fix for bug:  https://fogbugz.unity3d.com/f/cases/1304501/

In certain cases, the TabbedView code will add/remove a null Target UI element to the UI hierarchy, resulting in an exception thrown all the way up the call stack, skipping a lot of shadergraph code and corrupting shadergraph.

This PR checks for null before Adding or Removing the Target.

It also changes one exception handling location to handle all exceptions, so that necessary ShaderGraph code doesn't get skipped when exceptions like this are thrown.

---
### Testing status

Tested that the exception handling change, by itself, results in exceptions being called out as errors in the log, but does not corrupt the ShaderGraph behavior.

Tested that the full change fixes the issue (windows editor) -- no longer any errors/exceptions being thrown, or ShaderGraph hang behavior.

Tested that the "node settings" tab view activation on selection still appears to be working, even when the inspector view is not visible initially.

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -1,4 +1,4 @@
-﻿# Changelog
+# Changelog
 All notable changes to this package are documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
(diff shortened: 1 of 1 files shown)


jessebarker (approved):
LGTM.  Sai should confirm.

Nightmask3 (approved):
Tested to make sure the bug behavior is fixed and doesn't repro. Can confirm it does not. Looks good!

LandonTownsendUnity (approved):
Looks like it definitely fixes the bug. Let's get this merged.

PR #3043, opened 2021-01-08
Added a new control slider on RTR and RTGI to force the LOD Bias on both effects.

anisunity:
For flickering and performance reasons we've decided to expose the ability to override the LOD Bias for specific ray tracing effects.
Here is an example with reflections.
image

Testing status
Testing on mutliple scenes, works fine.
DXR tests are green locally

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/GlobalIllumination.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/GlobalIllumination.cs
@@ -86,7 +86,7 @@ namespace UnityEngine.Rendering.HighDefinition
         /// </summary>
         [Tooltip("The LOD Bias HDRP applies to textures in the global illumination. A higher value increases performance and makes denoising easier, but it might reduce visual fidelity.")]
         public ClampedIntParameter textureLodBias = new ClampedIntParameter(7, 0, 7);
-        
+
         /// <summary>
         /// Controls the length of GI rays.
         /// </summary>
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflection.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflection.cs
@@ -122,7 +122,7 @@ namespace UnityEngine.Rendering.HighDefinition
         /// Defines the LOD Bias for sampling all the textures.
         /// </summary>
         public ClampedIntParameter textureLodBias = new ClampedIntParameter(1, 0, 7);
-        
+
         /// <summary>
         /// Controls the length of reflection rays.
         /// </summary>


anisunity:
@JordanL8  this requires some doc update

com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/HDScreenSpaceReflectionEditor.cs
```diff
@@ -80,6 +82,7 @@ public override void OnEnable()
         static public readonly GUIContent k_RayTracingText = EditorGUIUtility.TrTextContent("Ray Tracing (Preview)", "Enable ray traced reflections.");
         static public readonly GUIContent k_ReflectSkyText = EditorGUIUtility.TrTextContent("Reflect Sky", "When enabled, SSR handles sky reflection.");
         static public readonly GUIContent k_LayerMaskText = EditorGUIUtility.TrTextContent("Layer Mask", "Layer mask used to include the objects for screen space reflection.");
+        static public readonly GUIContent k_LodBiasText = EditorGUIUtility.TrTextContent("Lod Bias", "Defines the LOD Bias for sampling all the textures. A higher value will increase peformance and make denoising easier but may reduce visual fidelity.");
```
JordanL8, line 85: The LOD Bias for sampling all the textures. A higher value increases performance and makes denoising easier, but it might reduce visual fidelity.

remi-chapelain:
Next time please mention in the PR which ray tracing effect are impacted please :)
I assume the two concerned effects are Ray Traced Reflection and Ray Traced Global Illumination based on the changes

TomasKiniulis (commented):
Sharing some minor findings:
- "Lod Bias" can be easily confused with geometry LODs, which I thought is the case. Maybe it would be better to have it named "Mipmap level"? Or "Texture LOD Bias"? To state clearly that textures are affected.
- For RTGI, I cannot notice the impact of "LOD Bias", what could be the setup to have the effect noticeable? I suggest moving the slider to the bottom of RTGI list, as the effect does not seem to give a significant difference in this case

com.unity.render-pipelines.high-definition/Runtime/Lighting/GlobalIllumination.cs
```diff
@@ -98,6 +98,12 @@ public int filterRadius
         [Tooltip("Defines the layers that GI should include.")]
         public LayerMaskParameter layerMask = new LayerMaskParameter(-1);
 
+        /// <summary>
+        /// Defines the LOD Bias for sampling all the textures.
+        /// </summary>
+        [Tooltip("Defines the LOD Bias for sampling all the textures. A higher value will increase peformance and make denoising easier but may reduce visual fidelity.")]
```
JordanL8, line 104: Sets the LOD Bias HDRP applies to textures in the reflection. A higher value increases performance and makes denoising easier, but it might reduce visual fidelity.

JordanL8:
I added documentation to the ssr and ssgi docs. I'm not sure where it is in the Inspector so can you have a look a the docs and move the Lod Bias entry up/down in the table if its in the wrong place? Thanks

TomasKiniulis (commented):
@anis can you check the comments? Anything else besides these findings seems to be good

anisunity:
Updated the PR docs and merged master @sebastienlagarde

TomasKiniulis (approved):
Thanks for the Texture Lod bias name change. After discussing with Anis, I see a significant difference with emissive textures containing small contrasting details. Placement for the bias slide seems to make sense

sebastienlagarde (approved):

sebastienlagarde (changes requested):
We need to reset after each effect who use it the _RayTracingLodBias to 0 otherwise other effect like recursive rendering will be affected

anisunity:
Updated and fixed the PR

sebastienlagarde (approved):

PR #3048, opened 2021-01-11
bump version to 7.5.3

sebastienlagarde:
all in titlle

### DIFF
--- a/com.unity.render-pipelines.core/CHANGELOG.md
+++ b/com.unity.render-pipelines.core/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [7.5.3] - 2021-11-1
+
+Version Updated
+The version number for this package has increased due to a version update of a related graphics package.
+
 ## [7.5.2] - 2020-11-16
 
 Version Updated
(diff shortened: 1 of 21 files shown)


com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [7.5.3] - 2021-11-1
```
theo-pnv, line 7: iirc we had some validation tests issues in the past with dates with only one digit (like this one).
Please add a `0` to the date: `2021-11-01`

theo-pnv: Note: For all packages

theopnv (changes requested):
Do lines 4 and 9 of the visual effect graph testing package need to be bumped as well? https://github.com/Unity-Technologies/Graphics/blob/srp-7-5-3/com.unity.testing.visualeffectgraph/package.json#L4

com.unity.render-pipelines.core/CHANGELOG.md
phi-lira: looking at the history, shouldn't the date/month be reversed as well? 
`2021-01-11`

phi-lira (changes requested):

com.unity.render-pipelines.core/CHANGELOG.md
sebastienlagarde: good point!

sebastienlagarde:
adressed pr feedback

theopnv (approved):

sebastienlagarde:
thanks for the update of testing vfx package

PR #3049, opened 2021-01-11
Update changelog to have 2020.1 and 2020.2 version

sebastienlagarde:
all in title

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Version.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Version.hlsl
@@ -1,9 +1,9 @@
+// The old version number system below is deprecated whith Graphic Packages that have move as core package of Unity.
+// User should rely on the Macro UNITY_VERSION now to detect which version of Unity is coupled to the current set of pipeline shader
+// Example of usage #if UNITY_VERSION >= 202120 to check if the version is above or equal 2021.2
 #define SHADER_LIBRARY_VERSION_MAJOR 11
 #define SHADER_LIBRARY_VERSION_MINOR 0
 
 #define VERSION_GREATER_EQUAL(major, minor) ((SHADER_LIBRARY_VERSION_MAJOR > major) || ((SHADER_LIBRARY_VERSION_MAJOR == major) && (SHADER_LIBRARY_VERSION_MINOR >= minor)))
 #define VERSION_LOWER(major, minor) ((SHADER_LIBRARY_VERSION_MAJOR < major) || ((SHADER_LIBRARY_VERSION_MAJOR == major) && (SHADER_LIBRARY_VERSION_MINOR < minor)))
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -4,7 +4,12 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [11.0.0] - 2020-10-21
+## [2022.2.0] - 2021-11-1
```
theo-pnv, line 7: Please add a `0` when it's a single-digit date, we had issues with validation tests in the past.
Example: `2021-11-01`

theo-pnv: Like in this test https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/srp-bumpversion/.yamato%252F_projectcontext.yml%2523test_all_project_windows_trunk/4831948/job

theo-pnv: Note: For all packages

com.unity.render-pipelines.high-definition-config/CHANGELOG.md
```diff
@@ -4,7 +4,12 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [11.0.0] - 2020-10-21
+## [2022.2.0] - 2021-11-1
```
theo-pnv, line 7: Shouldn't it be 2021.2.0 instead? 
Note: For all packages

theo-pnv (changes requested):

stramit (commented):
I believe that because we are no longer doing package releases we do NOT need the date. It would be good to remove that.

sebastienlagarde:
> I believe that because we are no longer doing package releases we do NOT need the date. It would be good to remove that.

But I guess package validation suite if we don't have a date...

com.unity.render-pipelines.high-definition-config/CHANGELOG.md
sebastienlagarde: good catch!

com.unity.render-pipelines.core/CHANGELOG.md
sebastienlagarde: good point

theo-pnv:
I don't know if we would still be respecting the KeepAChangelog format if we removed the dates. https://keepachangelog.com/en/1.0.0/
image

theo-pnv (approved):

phi-lira (approved):

PR #3050, opened 2021-01-11
Fixed some render texture leaks.

JulienIgnace-Unity:
### Purpose of this PR
Fixed various render texture leaks.
- Volumetric history buffers were never released
- Texture from the 3D atlas were not disposed of properly
- Render Graph deallocation scheme of stale texture was too aggressive, causing texture to deallocate/reallocate every frame when off-screen cameras were present (this also caused some unwanted duplicate textures in memory).

Fixes this issue: https://fogbugz.unity3d.com/f/cases/1295828/

---
### Testing status
Tried the repro described in the case and also made multiple sequences of entering and leaving playmode while looking at the memory profiler to check that no render textures were leaking anymore.
No further QA should be required.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -66,7 +66,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed LayerMask editor for volume parameters.
 - Fixed the condition on temporal accumulation in the reflection denoiser (case 1303504).
 - Fixed box light attenuation.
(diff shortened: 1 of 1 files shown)


alelievr (approved):

Lex4art:
Hope this will fix those - https://fogbugz.unity3d.com/default.asp?1302641_plje0fv5aekhop4t and https://fogbugz.unity3d.com/default.asp?1302629_83dl8u4t94pnudpv**

JulienIgnace-Unity:
@Lex4art it does seem that the leak you observed is the one I fixed with volumetric fog indeed.

sebastienlagarde (approved):

JMargevics (approved):

PR #3055, opened 2021-01-11
[HDRP] Fix issue with compositor related custom passes

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1305330/

Some custom post process passes were still active after disabling the compositor (and their name was missing in RenderDoc captures.)

---
### Testing status
- Compositor test is green
- After disabling the compositor, the passes are not executed anymore, as seen in RenderDoc:
image

---
### Comments to reviewers
Note that these passes decide if they will be executed or not based on some info from the camera. Ideally I would like to add a proper disabled state on them, but right now we cannot do that because the IsActive() API call does not have access to the camera. This is easy to fix, but I don't think that updating the API is worth it for only this use case. The downside is that old projects with any left-over passes might still have this overhead, until the user removes them manually from the HDRP asset.


alelievr (approved):

sebastienlagarde (approved):

JMargevics (approved):
Talked to @alelievr and he verified this on his side. No additional QA pass needed 👍

PR #3056, opened 2021-01-11
Fix box light attenuation

adrien-de-tocqueville:
### Purpose of this PR

Note: Still need to investigate why tests with shadows are failing

fix both issues raised in https://fogbugz.unity3d.com/f/cases/1299693/

The attenuation was divided by the the distance to the box light, which would make it drop very quickly and go above the light intensity when close to the light

Also removes two unused variables.

Note: this changes quite significantly the behaviour of box lights with range attenuation

Before            |  After
:-------------------------:|:-------------------------:
before | after

---
### Testing status

Tested the repro project.
Tried various light ranges, with light close to objects and further than range, with range attenuation enabled and disabled and verified all works as expected.


TomasKiniulis (approved):
Looks good to me. Tested multiple box lights overlapping each other while toggling range attenuation. No issues found.

https://user-images.githubusercontent.com/50582134/104289320-a777ce00-54c1-11eb-9f66-28c83c386e96.mp4

sebastienlagarde (approved):

RinaldoTjan:
This looks good. It solved the over brightening issue.
While the fall off fade out a bit fast to my liking, but It does render quite a nice smooth fall off effect.
If no fall off is desired people will disable the attenuation anyway.
I'd say problem solved.

PR #3057, opened 2021-01-11
Hd/fix 1299116 tesselation cull

kecho:
### Purpose of this PR
Bug fix for dissapearing camera triangles on lit tesselation materials. The culling algorithm used for tesselated geometry was incorrect, and if a camera would be inside a triangle then there is a likelahood for this geometry to dissapear.
This PR fixes the problem by implementing a correct triangle / frustum plane rejection algorithm.

For more info check: https://fogbugz.unity3d.com/f/cases/1299116/

More technical explanation:
The rejection algorithm used before was testing individual triangle edges being out of the frustum. If a camera is inside a triangle, then it would be considered culled, thus triggering the issue.
The new rejection algorithm just tests the 3 vertices of the triangle instead.

---
### Testing status
Tested using the scene attached (see the fogbugz scene attached). Also, created a new test that is able to replicate the problem.
The new test is called 1224_Lit_Tesselation_Culling.
Tested on (screenshots submitted):
* Win32 Vulkan, Dx11, Dx12
* Linux Vulkan
* OSX Metal

Performance: Added 1 register on depth prepass and opaque passes for tesselation hull constant function evaluation

---
### Comment for Reviewers
A few notes to be aware:
* Culling occurs during the tesselation constant hull evaluation, by returning 0,0,0 for the tesselation factors
* The bug is tricky to repro, it can only happen with a mesh with a big single triangle. Easiest way to repro in unity is by adding a cube (not a plane), since the cube has 2 tris per face.
* This CL ensured to preserve the 'optimized' logic, which prevents from running redundant culling tests for shadow passes.

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/2505_Area_Light_ShadowMask_Baking.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/2505_Area_Light_ShadowMask_Baking.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:e0e690ee5292c6439ddaae68790866b71c540162dbdbdbda543470533d68fe8c
-size 289
+oid sha256:c659967f1e80fb2be10daf61fc01895038d670334235b94a39639892b717f5b0
+size 136318


kecho:
I should have made this against hd/bugfix. Newb mistake

sebastienlagarde (approved):

sebastienlagarde:
PR is ready, waiting for QA

JMargevics (approved):
Bug does not reproduce anymore with this PR 👍
Additionally made a quick pass on tessellated shaders + displacement. No issues on my side ✅

PR #3059, opened 2021-01-11
Fix undo redo on layered lit editor

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1301111/

The uv checkbox is now serialized in the material meta file, alongside the layer materials
This way, they are remembered and can be undone.

Also add a fallback to initialize values when meta file doesn't contain any serialized data, so that undo now works when you edit uv or layer material for the first time after creation

Also removes some duplicate code

---
### Testing status
Tested undo redo on layered lit material layer list


TomasKiniulis (approved):
Thanks! Checked LayeredLit and LayeredLitTessellation, in both shaders assigned Layer Materials get undone and applied correctly after undo/redo now.

sebastienlagarde (approved):

PR #3061, opened 2021-01-11
Fix Light Intensity UI Prefab Override Application (1299563)

johnpars:
### Purpose of this PR
Case: https://fogbugz.unity3d.com/f/cases/1299563/

This PR is an overall fix/improvement to the behavior of applying prefab overrides for the "Intensity" field in the light UI (note, this field allows users to apply the override on intensity _and_ light unit setting, depending on what changed). 

Before:
Before

After:
After

---
### Testing status
I manually created a light prefab and confirmed the new intended behavior. I changed light intensity/light unit and confirmed these changes are now registered to the prefab, and it's possible to apply the override anywhere in the two lined inspector rect.

---
### Comments to reviewers
A part of this fix required a very strange/nebulous swap of two `EditorGUI.BeginProperty` calls. It's still unclear to me why this was necessary, my guess is that calling one before the other would simply overwrite the state of the previous. It may be indicative of some kind of deeper bug that lies within this call, maybe something to look into with editor devs if this kind of problem ever occurs again.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fixed GC allocations from XR occlusion mesh when using multipass.
 - Fixed XR depth copy when using MSAA.
+- Fixed after post process custom pass scale issue when dynamic resolution is enabled (case 1299194).
 - Fixed an issue with light intensity prefab override application not visible in the inspector (case 1299563).
+
+### Changed
+- Change the source value for the ray tracing frame index iterator from m_FrameCount to the camera frame count (case 1301356).
 
 ## [11.0.0] - 2020-10-21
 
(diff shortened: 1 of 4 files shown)


sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -599,11 +599,17 @@ static void DrawLightIntensityGUILayout(SerializedHDLight serialized, Editor own
             Rect labelRect = lineRect;
             labelRect.width = EditorGUIUtility.labelWidth;
 
+            // Expand to reach both lines of the intensity field.
+            labelRect.height *= 2f;
```
RSlysz, line 603: To be 100% accurate, the interline is missing here. You should have 2* singleLineHeight +2 (or 4 don't remember exactly and the property is not accessible in user land.)

RSlysz (approved):
Looks good to me.

This is highly nebulous. Normally there is no ordering on the EditorGUI.BeginProperty AFAIK as long as they are all targeting same Behaviour. I'm pretty sure that changing the height should be enough. But in the same times, it cannot be harmful AFAIK.

iM0ve (approved):
Very minor change. I opened to check the same as in the gif provided, but did this on mac. No issues found, good to go.

PR #3062, opened 2021-01-12
Filter UniversalGraphicsTests

phi-lira:
Disable Test 035 as it's currently failing on all Standalone platforms. 
 Fogbugz: https://fogbugz.unity3d.com/f/cases/1304675


martint-unity (changes requested):
Could you please add the ID to the fogbugz case

martint-unity (approved):

nigeljw-unity (approved):

PR #3070, opened 2021-01-12
[HDRP] Added a RenderGraph pass that resets the camera size after the dynamic res upscale

alelievr:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1299194/ and https://forum.unity.com/threads/viewport-is-incorrect-in-a-custom-pass-running-in-afterpostprocess-when-using-dynamic-resolution.1008029/

Context: When the dynamic resolution is enabled, HDRP renders everything in a smaller viewport up to the final post-process pass which is just after the "After PP" post-process injection point.  
After post-process, opaque materials are rendered before post processes in an offscreen target and composited back during this final pass.  
So the custom pass after post-process is the only pass that renders something after the final pass that does the upscaling of the image for dynamic res.

Because of that, the after post-process custom passes are rendered with a smaller viewport than expected and there is no upscale pass after this one so it leads scale issue during the custom pass rendering.

To fix this, I added a pass that resets the camera size after the dynamic res upscale pass, so everything rendered after that is using the correct scale.

---
### Testing status
Tested that there is no more scale issue with a simple after post-process custom pass and dynamic res enabled (with a fixed resolution scale multiplier).  
Also verified that there is no CG alloc in the RG pass.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fafter-post-process-dynamic-res/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/4886416/job

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/1x_Materials/1224_Lit_Tesselation_Culling.meta
+++ b/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/1x_Materials/1224_Lit_Tesselation_Culling.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 3e15a4def182a184d9dafc2122db4c90
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/1x_Materials/1224_Lit_Tesselation_Culling.unity
+++ b/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/1x_Materials/1224_Lit_Tesselation_Culling.unity
@@ -1,0 +1,345 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+OcclusionCullingSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 2
+  m_OcclusionBakeSettings:
+    smallestOccluder: 5
+    smallestHole: 0.25
+    backfaceThreshold: 100
+  m_SceneGUID: 00000000000000000000000000000000
+  m_OcclusionCullingData: {fileID: 0}
+--- !u!104 &2
+RenderSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 9
+  m_Fog: 0
+  m_FogColor: {r: 0.5, g: 0.5, b: 0.5, a: 1}
+  m_FogMode: 3
+  m_FogDensity: 0.01
+  m_LinearFogStart: 0
+  m_LinearFogEnd: 300
+  m_AmbientSkyColor: {r: 0.212, g: 0.227, b: 0.259, a: 1}
+  m_AmbientEquatorColor: {r: 0.114, g: 0.125, b: 0.133, a: 1}
+  m_AmbientGroundColor: {r: 0.047, g: 0.043, b: 0.035, a: 1}
+  m_AmbientIntensity: 1
+  m_AmbientMode: 0
+  m_SubtractiveShadowColor: {r: 0.42, g: 0.478, b: 0.627, a: 1}
+  m_SkyboxMaterial: {fileID: 0}
+  m_HaloStrength: 0.5
+  m_FlareStrength: 1
+  m_FlareFadeSpeed: 3
+  m_HaloTexture: {fileID: 0}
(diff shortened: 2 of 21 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -1535,6 +1537,33 @@ void RenderGizmos(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle colo
             return executed;
         }
 
+        class BindResetCameraSizeForAfterPostProcess
```
JulienIgnace-Unity, line 1540: Most if not all other pass data are called XXXPassData so I think we should rename this for consistency.

JulienIgnace-Unity (changes requested):
Requested a rename but otherwise lgtm. Hopefully we won't have unforseen side effects. Tagging @FrancescoC-unity in case you can think of anything like that.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
alelievr: fixed

JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -1535,6 +1537,33 @@ void RenderGizmos(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle colo
...
 
+        class ResetCameraSizeForAfterPostProcessPassData
+        {
+            public HDCamera hdCamera;
+            public ShaderVariablesGlobal shaderVariablesGlobal;
+        }
+
+        void ResetCameraSizeForAfterPostProcess(RenderGraph renderGraph, HDCamera hdCamera, CommandBuffer commandBuffer)
+        {
+            if (hdCamera.allowDynamicResolution && HDUtils.hdrpSettings.dynamicResolutionSettings.enabled)
```
FrancescoC-unity, line 1548: Better to go through the Dynamic resolution handler method: DynamicResolutionEnabled ( `DynamicResolutionHandler.instance.DynamicResolutionEnabled ` )

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -1535,6 +1537,33 @@ void RenderGizmos(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle colo
...
+                using (var builder = renderGraph.AddRenderPass("Reset Camera Size After Post Process", out ResetCameraSizeForAfterPostProcessPassData passData))
+                {
+                    passData.hdCamera = hdCamera;
+                    passData.shaderVariablesGlobal = m_ShaderVariablesGlobalCB;
+                    builder.AllowPassCulling(false);
+
+                    builder.SetRenderFunc(
+                        (ResetCameraSizeForAfterPostProcessPassData data, RenderGraphContext ctx) =>
+                        {
+                            data.shaderVariablesGlobal._ScreenSize = new Vector4(data.hdCamera.finalViewport.width, data.hdCamera.finalViewport.height, 1.0f / data.hdCamera.finalViewport.width, 1.0f / data.hdCamera.finalViewport.height);
```
FrancescoC-unity, line 1559: Do we need updating the RTHandle scale stuff before pushing the cbuffer?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
alelievr: Thanks, fixed :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
alelievr: Good catch! indeed the RTHandleScale needs to be updated if we want to sample low-res buffers from before the upscale

sebastienlagarde (approved):

FrancescoC-unity (approved):

PR #3072, opened 2021-01-12
[HDRP]Improvements and fixes on Volumes Components Inspectors

alex-vazquez:
### Purpose of this PR
This PR is fixing these bugs
- https://fogbugz.unity3d.com/f/cases/1297795/
- https://fogbugz.unity3d.com/f/cases/1303477/
- https://fogbugz.unity3d.com/f/cases/1297789/

### Changes
- Fixed spacing between property fields due to a harcoded Rect
- ALL/NONE changed to Toggles and maintain the state
- When ALL/NONE are selected we check if we should affect to the additional properties
- Adding localization to undo stack
- Adding unit tests
- Bunch of test checking changes done on additional properties of volumes components
- Testing the Volume component editors to check the set and get of the overrides states.
- Update CHANGELOG.md

### Testing status
- Spacing of properties on Volume Components editors
- ALL/NONE are update only the correct overrides (when advanced mode is off they do not affect to advanced properties)
- ALL/NONE mantain the state if all the overrides are checked or if any of them are checked.

### Comments to reviewers
Documentation changes will be sync with @JulienIgnace-Unity


com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -176,9 +196,18 @@ internal void Init(VolumeComponent target, Editor inspector)
...
+
+            InitParameters();
+
             OnEnable();
         }
 
+        void InitParameters()
+        {
+            m_VolumeNotAdditionalParameters = new List<VolumeParameter>();
+            VolumeComponent.FindParameters(target, m_VolumeNotAdditionalParameters, field => field.GetCustomAttribute<AdditionalAttribute>() == null);
```
RSlysz, line 208: @JulienIgnace-Unity Shouldn't you use this attribute too for your PR on the displays of them? Can be perhaps usefull if we have a equivalent to default inspector for volumes.

com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
```diff
@@ -195,10 +200,14 @@ public virtual void Override(VolumeComponent state, float interpFactor)
...
-            SetAllOverridesTo(parameters, state);
+            SetOverridesTo(parameters, state);
         }
 
-        void SetAllOverridesTo(IEnumerable<VolumeParameter> enumerable, bool state)
+        /// <summary>
+        /// Sets the override state of the given parameters on this component to a given value.
+        /// </summary>
+        /// <param name="state">The value to set the state of the overrides to.</param>
+        public void SetOverridesTo(IEnumerable<VolumeParameter> enumerable, bool state)
```
RSlysz, line 210: Isn't having this public is redudant with SetAllOverridesTo?

If usefull keep this change, if not really remove it. We only expose what is needed as exposed API are hard to change.

com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
```diff
@@ -81,7 +81,9 @@ public class VolumeComponent : ScriptableObject
         /// <summary>
         /// Extracts all the <see cref="VolumeParameter"/>s defined in this class and nested classes.
         /// </summary>
-        static void GetParameters(object o, List<VolumeParameter> parameters)
+        /// <param name="parameters">The list filled with the parameters.</param>
+        /// <param name="filter">If you want to filter the parameters</param>
+        public static void FindParameters(object o, List<VolumeParameter> parameters, Func<FieldInfo, bool> filter = null)
```
RSlysz, line 86: missing documentation on `o`

Public API must be fully documented in code (even returns if not void). We have test on the API documentation actually. 

If the visibility change is only to have access in VolumeComponent, normaly Editor package see internals on Runtime version of the package usually in Unity. So it would have worked with internal normally.

(optional) Also it do not search, but deliver them up to an optional filter. Get have my preference in this case.

RSlysz (changes requested):
My main concern is that the current test freeze the code a little to much. I would have prefered to only test the feature than actually test all component and froze the additional fields eddition without editing your test.

Then check the API visibility and if public is good, fix the missing argument in the documentation.

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -75,6 +75,25 @@ public VolumeComponentEditorAttribute(Type componentType)
...
+
+        Vector2? m_OverrideToggleSize;
+        internal Vector2 overrideToggleSize
+        {
+            get
+            {
+                m_OverrideToggleSize ??= CoreEditorStyles.smallTickbox.CalcSize(Texts.overrideSettingText);
+                return m_OverrideToggleSize.Value;
+            }
+        }
```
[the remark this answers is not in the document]
alex-vazquez: The thing here you should call this method on `OnInspectorGUI`, that's why I did it like this

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
alex-vazquez: We were discussing this, an we will look at the merge.

com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
alex-vazquez: I need to be public. See `VolumesComponentEditor.SetOverridesTo`

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -75,6 +75,25 @@ public VolumeComponentEditorAttribute(Type componentType)
     /// <seealso cref="VolumeComponentEditorAttribute"/>
     public class VolumeComponentEditor
     {
+        class Texts
```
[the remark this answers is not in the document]
alex-vazquez: I saw `Texts` using it in other classes, for me using the word `Styles` for the class name makes me think that they are colors, sizes, and here I am holding `GUIContent` not how they look like. I will changed as in our package is called `Styles`.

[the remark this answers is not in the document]
jenniferd-unity: i think other classes in HDRP at least use _Styles_ maybe it is best to go with that for this portion of code?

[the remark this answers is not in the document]
alex-vazquez: Yeah, we will mantain `Styles` to avoid confusion in the project. So be will be always using `Styles` for our package :)

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -75,6 +75,25 @@ public VolumeComponentEditorAttribute(Type componentType)
...
+        {
+            public static GUIContent overrideSettingText { get; } = EditorGUIUtility.TrTextContent("", "Override this setting for this volume.");
+            public static GUIContent allText { get; } = EditorGUIUtility.TrTextContent("ALL", "Toggle all overrides on. To maximize performances you should only toggle overrides that you actually need.");
+            public static GUIContent noneText { get; } = EditorGUIUtility.TrTextContent("NONE", "Toggle all overrides off.");
+
+            public static string toggleAllText { get; } = L10n.Tr("Toggle All");
+        }
+
+        Vector2? m_OverrideToggleSize;
```
jenniferd-unity, line 87: I am not familiar with that notation. what does the ? mean after the type

alex-vazquez: This is making a nullable value type (since c# 8 you can do it on reference types).

Basically it allow us to set null to the variable(default value)

I am using it here to allow me to check if it has been initialized.

For instance imagine a `bool` that is binded to UI. bool will only have true(yes) or false(no) as possibilities, but you want to make sure that the user selects if the option is wanted or not. So, you could not default to `false`, So instead of having another bool to check if the the user has selected yes or no for this, you make a `bool?`
now our `bool?` has three states null, true, false. This will allow us to check easily if the user has selected a value for it.

So, to obtain the value of a nullable reference or value type you could do it by `.HasValue` (but it think performance wise will be the same as doing `==null`), and to obtain the value stored in our case a `Vector2` simply use .Value.

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
```diff
@@ -0,0 +1,109 @@
...
+                .SetName("VolumeComponentNoAdditionalAttributes"),
+            new TestCaseData(typeof(VolumeComponentAllAdditionalAttributes))
+                .Returns(new string[2] {"parameter1", "parameter2"})
+                .SetName("VolumeComponentAllAdditionalAttributes"),
+            new TestCaseData(typeof(VolumeComponentMixedAdditionalAttributes))
+                .Returns(new string[2] {"parameter2", "parameter4"})
+                .SetName("VolumeComponentMixedAdditionalAttributes"),
+        };
+
+        [Test, TestCaseSource(nameof(s_AdditionalAttributesTestCaseDatas))]
```
alex-vazquez, line 88: @RSlysz , thank you for this, I forgot totally about it

com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
alex-vazquez: Made internal and fix in the documentation, thanks.

com.unity.render-pipelines.core/Runtime/Common/CoreAttributes.cs
```diff
@@ -13,4 +13,12 @@ public class DisplayInfoAttribute : Attribute
         /// <summary>Display order used in UI.</summary>
         public int order;
     }
+
+    /// <summary>
+    /// Attribute used to customize UI display to allow properties only be visible when "Show Additional Properties" is selected
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Field)]
```
JulienIgnace-Unity, line 20: I'm wondering if we should call this attribute "AdditionalProperty". "Additional" alone seems to vague.

JulienIgnace-Unity (changes requested):
One comment about the name of the Additional Attribute otherwise lgtm.

com.unity.render-pipelines.core/Runtime/Common/CoreAttributes.cs
alex-vazquez: Perfect, I will rename it to `AdditionalProperty`. Thanks

JulienIgnace-Unity (approved):

RSlysz (approved):
Better on the test side. 📗  lgtm :)

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
jenniferd-unity: thanks for the tip!

jenniferd-unity (approved):
Nice first PR :) definitely good improvements there!

alex-vazquez:
Reviewed with @MaddalenaUnity , green light from she

iM0ve (approved):
**Before PR**

a

**After PR**

b

**Concerns:**
* The 2px change increases the vertical gaps between checkboxes, making already long volume overrides already longer. Discussed with UX that this is an expected side effect
* The horizontal gaps between checkboxes also increased due to alignment with NONE field. Maybe we shouldnt align with NONE? See image below cc @MaddalenaUnity 

Screenshot 2021-01-25 at 17 01 29

**Whats tested:**
Since its just a visual change I only did basic testing of described features and Undo check.

**Conclusion**
These are minor concerns and fine to merge it as is, but would be good to consider them if we can.

alex-vazquez:
@iM0ve @MaddalenaUnity 

Updating with your comments:

volumesallalignright

PR #3074, opened 2021-01-12
Fix additional data editors with non editable objects

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1303466/

When inspecting a readonly gameobject without additional data component, the editor will crash because it can't add the additional data component.
Added an import post processor that adds the relevant additional data components during import (they are not read only yet at this moment).
With this version the displayed UI is the one from HDRP, and not thie builtin editor like i proposed before

---
### Testing status
Tested the repro file which contains a light and a camera
Tested with a reflection probe component


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightEditor.cs
```diff
@@ -66,6 +73,12 @@ void OnUndoRedo()
 
         public override void OnInspectorGUI()
         {
+            if (m_AdditionalLightDatas == null)
+            {
+                base.OnInspectorGUI();
```
RSlysz, line 78: I am not sure we want to do that.

If I get it well here we show the base inspector so the built-in one even if we are in HDRP. It can get confusing for users. 
I think we should add a warning message in the same time, like in the form of an helpbox before the Inspector draws. To be confirmed with UX. ( @MaddalenaUnity )
(same for other types)

RSlysz (commented):
It is technicaly good, but need a UX pass on this change to be approved.

sebastienlagarde:
@adrien-de-tocqueville what is the status of this PR?

adrien-de-tocqueville:
I talked with maddalena about it and she wasn't fond of the solution. I tried to duplicate the asset as temporary hidden gameobject (which is no longer readonly) to display the correct editor. It seems to work but I wait for the next bugdays to properly finsh it

sebastienlagarde (approved):

sebastienlagarde:
adding Jennifer for awareness as it add a model post processor when you have HDRP package present.

TomasKiniulis (approved):
Testing is sufficient enough. Approving

com.unity.render-pipelines.high-definition/Editor/AssetProcessors/ModelPostProcessor.cs
```diff
@@ -0,0 +1,32 @@
...
+                    var hd = c.gameObject.AddComponent<AdditionalT>();
+                    if (initDefault != null)
+                        initDefault(hd);
+                }
+            }
+        }
+
+        void OnPostprocessModel(GameObject go)
+        {
+            AddAdditionalData<Camera, HDAdditionalCameraData>(go, HDAdditionalCameraData.InitDefaultHDAdditionalCameraData);
```
jenniferd-unity, line 27: So if I read this correctly, as long as we have the HDRP package present then importing a model will add HD additional data components. 
If we remove the package then what happens, is it okay to leave them?

RSlysz: Yes. Did you test with HDRP package installed but the RP in use is Universal? Could it work in this case?

RSlysz: Also, Universal have the same additional component for their Camera. What happens if both Universal and HD is installed and the prefab as a Camera, a UniversalAdditionalCameraData and a HDAdditionalCameraData?

TomasKiniulis:
Approved too fast. We still need to check the test case with both HDRP and URP packages installed.

com.unity.render-pipelines.high-definition/Editor/AssetProcessors/ModelPostProcessor.cs
adrien-de-tocqueville: You are right, I didn't think about it
Also since the bug also happens on URP, i need to fix it for both

adrien-de-tocqueville: So now if a pipeline is present in the project it will always add the additional data, otherwise if you switch active pipeline after model import, the object is not reimported and you still have the errors

iM0ve (changes requested):
After loading up a clean project (deleted library) that has URP and HDRP preinstalled, I am getting a ModelPostProcessor error

Screenshot 2021-02-03 at 16 38 19

adrien-de-tocqueville:
@iM0ve oops, that was a bad copy paste, fixed

RSlysz (approved):

iM0ve (approved):

sebastienlagarde:
note to me: redirect this PR to master on hd/bugfix is merge

phi-lira (approved):

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -797,6 +797,27 @@ public static T[] GetAdditionalData<T>(UnityEngine.Object[] targets, Action<T> i
             return data;
         }
 
+        /// <summary>Add the appropriate AdditionalData to the given GameOject and its children containing the original component</summary>
```
svens-unity, line 800: Typo in `GameOject `

PR #3042, opened 2021-01-08
Convert HDRP tests to use MockHMD for XR

fabien-unity:
---
### Purpose of this PR
- refactor XR tests to be able to use MockHMD and the full XR SDK path
- disable some tests until further investigation

---
### Testing status
[Yamato run](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/tests%252Fxr%252Fconvert-to-mockhmd-2/.yamato%252Fall-hdrp.yml%2523Nightly_HDRP_trunk/4815973/job) is similar to master (mac is failing).

---
### Comments to reviewers
There's a lot a files but it's mostly boilerplate and Yamato configuration

PR #3075, opened 2021-01-12
Fix XR depth copy and MSAA

fabien-unity:
---
### Purpose of this PR
- fix XR depth copy with MSAA and RenderGraph
- fix errors when trying to use PathTracing with XR single-pass
- fix errors when no occlusion meshes are present with RenderGraph

---
### Testing status
- locally tested with the template and Oculus plugin
- this code is extracted from https://github.com/Unity-Technologies/Graphics/pull/3042 and already passed Yamato.

---
### Comments to reviewers
Code fixes extracted from https://github.com/Unity-Technologies/Graphics/pull/3042 so that we can also backport them.


sebastienlagarde (approved):

sebastienlagarde:
one XR test is failing

PR #3090, opened 2021-01-13
Hd/fix backplate globalcubemap

JulienIgnace-Unity:
### Purpose of this PR
This PR removes rendering of backplates inside the cubemap generated for lightng in the HDRI sky.
The reason is that it did not make sense conceptually since the backplate needs a position and the cubemap is rendered at infinity. This caused artefacts in reflections coming from the sky cubemap.
This will fix https://fogbugz.unity3d.com/f/cases/1276692/

---
### Testing status
Tested the repro scene provided in the case.
@Unity-Technologies/gfx-qa-hdrp  it could be good to test a bit more thoroughly the backplate feature as I'm not 100% familiar with it so I might have missed some cases. Also please test local reflection probes. They should still contain the backplate.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Sky/SkyManager.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Sky/SkyManager.cs
@@ -960,7 +960,7 @@ namespace UnityEngine.Rendering.HighDefinition
                 }
 
                 if (preRenderSky)
-                    skyContext.skyRenderer.PreRenderSky(m_BuiltinParameters, false, hdCamera.camera.cameraType != CameraType.Reflection || skyContext.skySettings.includeSunInBaking.value);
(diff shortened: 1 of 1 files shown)


skhiat (approved):

iM0ve (approved):
No issues found, seems good to go.

Before PR:
After the reflection probe update/bake parts of the reflection would go black. Also, backplate from HDRI would be captured by global reflection probe.
wIdYrSXdoS


After PR:
No more black parts. The backplate reflection capture was removed from global reflection probe, now reflection of the backplate should be captured by placing a local probe in the scene.
1TqaZx6rSi

Whats tested:
* Repro project to reproduce black artifacts
* Capturing reflections of the backplate using local probes

JulienIgnace-Unity:
closed and reopened here: https://github.com/Unity-Technologies/Graphics/pull/3111

PR #3098, opened 2021-01-13
mobile optimisation: avoid rendering skybox in a different encoder

manuele-bonanno:
### Purpose of this PR
URP is rendering skybox in a separate pass, this is inefficient as it require to flush the target to main memory. Skybox should be rendered at the same time as opaque.

This optimisation was discussed with Apple.

---
### Testing status
local testing + Yamato

---
### Comments to reviewers
Separate Skybox RenderPass:

image


manuele-bonanno:
The Linux Vulkan STandalone test 132 failure is a known issue on master so not related to this PR

manuele-bonanno:
closing PR for now as discussed with Felipe and Elvar, as this can cause regressions for users that need to copy the depth texture

phi-lira:
We discussed the proper fix and will add the item to the roadmap. In short we need to make sure we don't regress existing projects but change behaviour for new ones or project that use new asset.

PR #3099, opened 2021-01-13
mobile optimisation: reduce the size of the fragment input struct of …

manuele-bonanno:
### Purpose of this PR
Right now URP lighting shader (PBR, Blinn phong shading, Terrain lighting, Baked lighting) fragment input struct on average has at least 7-8 variables. In the PBR lit shader it has the following types:
float2, float2, float3, float3, float4, float3, half4, float4, float3, float4.

The aim is to reduce the size of the fragment input struct (by changing the type of variable to smaller one, for example `half`, or reducing the number of the variables in the struct itself) to free up some write bandwidth performance on vertex shader pass.

This optimisation was discussed with Apple.

---
### Testing status
local + Yamato

---
### Comments to reviewers

Vertex Shader write bandwidth before:

image

Vertex Shader write bandwidth after:

image

### DIFF
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -49,6 +49,7 @@ The version number for this package has increased due to a version update of a r
 - Re-enabled implicit MSAA resolve to backbuffer on Metal MacOS.
 - Changed Post Process Data to bool. When it is no enabled all post processing is stripped from build, when it is enabled you can still override resources there.
 - Converted XR automated tests to use MockHMD.
+- Reduced the size of the fragment input struct of the Terrain and Forward lighting shaders.
 
 ### Fixed
 - Fixed an issue where additional lights would not render with WebGL 1


manuele-bonanno:
The Linux Vulkan STandalone test 132 failure is a known issue on master so not related to this PR

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitPasses.hlsl
```diff
@@ -47,16 +47,20 @@ struct Varyings
...
-    float4 tangent                  : TEXCOORD4;    // xyz: tangent, w: viewDir.y
-    float4 bitangent                : TEXCOORD5;    // xyz: bitangent, w: viewDir.z
+    half4 normal                    : TEXCOORD3;    // xyz: normal, w: viewDir.x
+    half4 tangent                   : TEXCOORD4;    // xyz: tangent, w: viewDir.y
+    half4 bitangent                 : TEXCOORD5;    // xyz: bitangent, w: viewDir.z
 #else
-    float3 normal                   : TEXCOORD3;
-    float3 viewDir                  : TEXCOORD4;
+    half3 normal                    : TEXCOORD3;
+    half3 viewDir                   : TEXCOORD4;
```
phi-lira, line 55: Isn't it worth it to move viewDir computation to fragment like the other shaders?

phi-lira (approved):
LGTM. Please add changelog entry.

phi-lira:
ABV was green in previous changeset, just changelog was changed.

PR #3100, opened 2021-01-13
[HDRP][Compositor] Fix issues with compositor's undo

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1305633/
https://fogbugz.unity3d.com/f/cases/1307170

When undoing the creation of a layer, we revert the state of the compositor to a previous state (with Undo.RecordObject), but we never remove any internal cameras that were created for this layer. These "orphan" cameras are still pointing to the compositor's RT, and this could create errors when removing these RTs (in my case I did not get a crash but an error when following the repro steps).

I fixed this issue by having a list of all compositor-allocated camera's and by doing a clean-up step in the compositor's undo callback.

---
### Testing status
I have tested this on a new HDRP outdoors scene (from the templates). I could not repro a crash, but I got an error message after the undo and the Undo actions had wrong descriptions. After this PR:
- The description of the undo action now shows correctly as "Undo Add compositor sublayer":
image
(instead of "undo enable compositor" as seen in the video )
- no crash or error code in the console after the repro steps.


com.unity.render-pipelines.high-definition/Editor/Compositor/CompositorWindow.cs
```diff
@@ -114,13 +114,13 @@ void OnGUI()
                 Undo.RegisterCreatedObjectUndo(compositor.outputCamera.gameObject, "Create Compositor");
                 Undo.RegisterCreatedObjectUndo(go, "Create Compositor");
             }
-            else if (compositor)
+            else if (compositor && (compositor.enabled != enableCompositor))
             {
                 string message = enableCompositor ? "Enable Compositor" : "Disable Compositor";
                 Undo.RecordObject(compositor, message);
                 compositor.enabled = enableCompositor;
```
pmavridis, line 121: Note: This assignment was always creating an undo action, so we now do it only if necessary.

adrien-de-tocqueville (approved):

sebastienlagarde (approved):

remi-chapelain (commented):
Repro case is fixed indeed but I still managed to create an additionnal error after creating one of each layer and "Undoing" my way back to the compositor enabling. 

https://user-images.githubusercontent.com/57442369/104937264-d86a6e00-59ad-11eb-96bc-0e1d0e94cf05.mp4

pmavridis:
Thanks for the finding, I have updated the PR to fix this issue.
I have also made the following changes:
- The selected layer does not change on Undo
- Also fixed a similar issue with Undo on AOVs: https://fogbugz.unity3d.com/f/cases/1307170 )

remi-chapelain:
Last commits did the trick, approving. ✔️

remi-chapelain (approved):

sebastienlagarde (approved):

PaulDemeulenaere:
This PR causes an issue with formatting job, fixed at https://github.com/Unity-Technologies/Graphics/pull/3187

PR #3103, opened 2021-01-13
Fixed fogbugz case 1303572

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR
- Fixed fogbugz case 1303572 (https://fogbugz.unity3d.com/f/cases/1303572/)
---
### Testing status
- ABV
- QA
- Tested manually with repro project and verified the issue is fixed.

---
### Comments to reviewers
- The issue is that updateVolume is updated too late (after InitializeCameraData) and because of that, cameraData does not have requireDepthTexture enabled. This caused CoC pass to fail. 
- The issue is introduced by https://github.com/Unity-Technologies/Graphics/commit/647e4729c23dfb911cbf6120b76273aaa87dd8e2 where InitializeCameraData call is moved ahead of updateVolume.


fabien-unity (approved):
LGTM

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -475,6 +475,8 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
 
             bool isStackedRendering = lastActiveOverlayCameraIndex != -1;
 
+            // Update volumeframework before initializing additional camera data
+            UpdateVolumeFramework(baseCamera, baseCameraAdditionalData);
```
phi-lira, line 479: There are other calls to UpdateVolumeFramework happening after InitializeCameraData, shouldn't they be changed as well?
 
 What's in the InitializeCameraData that depends on the UpdateVolumeFramework?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -505,7 +507,6 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
             //It should be called before culling to prepare material. When there isn't any VisualEffect component, this method has no effect.
             VFX.VFXManager.PrepareCamera(baseCamera);
 #endif
-            UpdateVolumeFramework(baseCamera, baseCameraAdditionalData);
 #if ADAPTIVE_PERFORMANCE_2_0_0_OR_NEWER
```
phi-lira, line 510: @PaulDemeulenaere for awareness. Is there any issue with calling UpdateVolumeFramework before VFXManager.PrepareCamera

PaulDemeulenaere: PrepareCamera will not access to any global settings, it will mainly update material, if camera isn't modified by UpdateVolumeFramework I don't expect any side effect.

PaulDemeulenaere (approved):
🟢 I checked the implementation of https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.core/Runtime/Volume/VolumeManager.cs#L15 & https://github.com/Unity-Technologies/Graphics/blob/76a39325d61876686f11f13c2320a4e0732f7d7b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs#L581
It won't have any effect on VFX, thanks for the notification.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
thomas-zeng: Hey felipe. It's the [CheckPostProcessForDepth](https://github.com/Unity-Technologies/Graphics/blob/c6038de9fbc56cbc73ca43d112e79dd0bac725c3/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs#L620) called in [InitializeAdditionalCameraData](https://github.com/Unity-Technologies/Graphics/blob/c6038de9fbc56cbc73ca43d112e79dd0bac725c3/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs#L831
) that depends on the UpdateVolumeFramework. 
UpdateVolumeFramework will update volume manager with the current camera. And that need to happen before the CheckPostProcessForDepth call.

phi-lira: Makes sense now. it also seems to make sense that we should guarantee that all calls to UpdateVolumeFramework should happen before InitializeAdditionalCameraData, Otherwise we might get other issues in the future?

phi-lira (changes requested):
We should fix the other calls of UpdateVolumeFramework to make sure it happens before initializing camera data.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
thomas-zeng: I agree. I did a search in Graphics repo looking for UpdateVolumeFramework use cases. There are 3 places that it is called. I verified with the PR change, all 3 cases have UpdateVolumeFramework  called before InitializeAdditionalCameraData. I think we are good.

nigeljw-unity:
Just pinging this PR for update.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
thomas-zeng: Hey @phi-lira, could you review my last comment about this PR and give some feedback :)? Thanks!

phi-lira (approved):

thomas-zeng:
I believe the failed tests are known issue mentioned here: https://unity.slack.com/archives/CFZERNTS7/p1612772986193600?thread_ts=1612748357.191200&cid=CFZERNTS7

thomas-zeng:
Hey @nigeljw-unity 
I have followed up with reviewers and this PR now has all reviewer's greenlight.
I also launched ABV, there are couple failures that I believe are known issues mentioned here https://unity.slack.com/archives/CFZERNTS7/p1612772986193600?thread_ts=1612748357.191200&cid=CFZERNTS7

Could you please give this PR a go? Thanks

nigeljw-unity:
Hey Thomas, sorry, this fell off my radar. I did go through it a few weeks ago, but needed to follow up digging through the yamato failures to verify they were expected (which I did and they looked good). I fixed the changelog conflict. It kicked off yamato again. I just want to be diligent and go through the yamato failures on lastest master. I'll do this later today when yamato finishes. Ping me if I don't follow up right away!

nigeljw-unity:
@thomas-zeng there is an HDRP regression with the formatter, can you pull master into this? The fix was pushed yesterday.

thomas-zeng:
Hi @nigeljw-unity,
I have merged master into the PR branch :)

PR #3104, opened 2021-01-13
[ShaderGraph] Removing "public" API that was never meant to be public

cdxntchou:
### Purpose of this PR
ShaderGraph makes public a lot of it's internal structures, that were never meant to be public API; but were made public for the purposes of allowing HDRP and URP packages to use them.

We marked a lot of it by putting it in the "Internal" namespace, but now with the documentation requirements we can't keep them public.

This PR converts them all to "Internal" and instead makes use of the "InternalsVisibleTo" construct to allow HDRP and URP to continue using them.

I used this list to find all of the public API points:   https://docs.google.com/spreadsheets/d/1U9ai0fAF9KTI5yTIky5fn_nLOUndxIUzhrJZdYnlzao/edit#gid=1243794439

---
### Testing status
Tested that ShaderGraph still opens (PC, windows) on a selection of different graphs, and standard lit and unlit shadergraphs still operate in HDRP and URP.

Running FULL Yamato pass:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Fclose-the-gates/.yamato%252F_abv.yml%2523all_project_ci_trunk/4878369/job/pipeline

master equivalent:

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Attributes/InspectableAttribute.cs
+++ b/com.unity.shadergraph/Editor/Data/Attributes/InspectableAttribute.cs
@@ -3,7 +3,7 @@ using System;
 namespace UnityEditor.ShaderGraph.Drawing
 {
     [AttributeUsage(AttributeTargets.Property)]
-    public class InspectableAttribute : Attribute
+    internal class InspectableAttribute : Attribute
     {
         // String value to use in the Property name TextLabel
         public string labelName { get; private set; }
--- a/com.unity.shadergraph/Editor/Data/Attributes/SGPropertyDrawerAttribute.cs
+++ b/com.unity.shadergraph/Editor/Data/Attributes/SGPropertyDrawerAttribute.cs
@@ -3,7 +3,7 @@ using System;
 namespace UnityEditor.ShaderGraph.Drawing
 {
     [AttributeUsage(AttributeTargets.Class)]
-    public class SGPropertyDrawerAttribute : Attribute
+    internal class SGPropertyDrawerAttribute : Attribute
     {
         public Type propertyType { get; private set; }
 
--- a/com.unity.shadergraph/Editor/Data/Enumerations/Precision.cs
+++ b/com.unity.shadergraph/Editor/Data/Enumerations/Precision.cs
@@ -7,7 +7,7 @@ namespace UnityEditor.ShaderGraph.Internal
         Half,
     }
 
-    public enum ConcretePrecision
+    internal enum ConcretePrecision
     {
         Single,
         Half,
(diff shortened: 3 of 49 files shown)


jessebarker (approved):
I don't feel comfortable auditing the actual list of APIs to make sure we haven't missed anything, but the changes look solid.

phbarralis:
Hi,
Asset store packages like this one ([](https://assetstore.unity.com/packages/tools/particles-effects/shadergraph-essentials-141671)) will have to resort to ugly hacks to work after this PR. They are an essential part of Unity ecosystem (just read the reviews on this asset!).
What are you thoughts on this?

Thanks, trying to be constructive here :)

marctem:
Hi, @phbarralis - It's absolutely a fair question. These APIs were never intended to be public in their current form. Most of them (I suspect the ones you use) are in the "ShaderGraph.Internal" namespace which was our attempt to make clear that they were unsupported and would change. We are preparing to change them in the near future, which is why we need to make them internal. However, we are going to evaluate the timing and see if there is a way to leave them as-is until we have a functional replacement. Thank you for your constructive tone; we really want everyone to succeed with Unity and Shader Graph!

phbarralis:
> Hi, @phbarralis - It's absolutely a fair question. These APIs were never intended to be public in their current form. Most of them (I suspect the ones you use) are in the "ShaderGraph.Internal" namespace which was our attempt to make clear that they were unsupported and would change. We are preparing to change them in the near future, which is why we need to make them internal. However, we are going to evaluate the timing and see if there is a way to leave them as-is until we have a functional replacement. Thank you for your constructive tone; we really want everyone to succeed with Unity and Shader Graph!

I do understand that these API where not meant to be used and are in the ShaderGraph.Internal namespace :)
I also know that many things are already internal in ShaderGraph, so this is only following the existing trend.

You do have to understand where this is coming from. Before Q2 2019 the entire ShaderGraph API was public and open. Several asset store package where developed and used by game devs on top of it. It was then put internal, **with no replacement** (and it's been almost 2y now). 
I know some publishers decided to discontinue their assets, and others (like myself) preferred to continue support it, because users where already relying on my assets to create their game. Even if it means undocumented breaking changes every version and work on my end. 

With the github repo going private (even if it's reverted for now) and this PR, I believe it sends the wrong signal. Why makes our lives much harder or impossible, when all we want is the same thing as you: everyone to succeed with Unity, SRPs and ShaderGraph. 

Thanks for your time and your understanding, hopefully you can related to my point of view :)

marctem:
After discussion with internal and external users, we have decided to leave these classes public for now, with the explicit understanding that they are unsupported, until we have a viable replacement API that we can support long-term for target customization.

PR #3108, opened 2021-01-14
[HDRP] Fixed lookdev reload bug when viewing a scene object

alelievr:
### Purpose of this PR
Fixed a bug that was causing objects from the scene to disappear from lookdev when relayouting the editor.
https://fogbugz.unity3d.com/f/cases/1300916/

---
### Testing status
Played a bit with lookdev with both assets on the disk and from the scene, dual-mode view, entering play mode, quit editor while a scene object is loaded or close/re-open window. Didn't notice any unexpected behaviors.

I'm not very confident with my ability to test the lookdev so it might be good to have a QA validate this PR as well.


sebastienlagarde (approved):

TomasKiniulis (changes requested):
@alelievr I can't get the fix to work. Can you take a second look at this? Made sure your changes are present, tried reloading project with clean library, but the issue is still present

https://user-images.githubusercontent.com/50582134/104702586-ec4c7080-571e-11eb-9ea1-bceed10864b2.mp4

alelievr:
Hum, yeah it's missing half of the bugfix. Probably because of the merge mess yesterday

alelievr:
@TomasKiniulis I re-added the missing bit, it should work now

TomasKiniulis (approved):
Thanks! This seems good. checked maximising screen when lookdev is docked while switching through HDRis and multiple libraries, checked various layouts. No issues are introduced.

RSlysz (approved):

PR #3115, opened 2021-01-14
Universal/soft particles ortho

eh-unity:
### Purpose of this PR
Fix orthographic soft particles.

---
### Testing status
Ran tests locally.

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fixed an issue where render scale was breaking SSAO in scene view. [case 1296710](https://issuetracker.unity3d.com/issues/ssao-effect-floating-in-the-air-in-scene-view-when-2-objects-with-shadergraph-materials-are-on-top-of-each-other)
 - Fixed GC allocations from XR occlusion mesh when using multipass.
+- Fixed an issue where soft particles did not work with orthographic projection. [case 1294607](https://issuetracker.unity3d.com/product/unity/issues/guid/1294607/)
 
 ## [11.0.0] - 2020-10-21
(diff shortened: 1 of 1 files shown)


richardkettlewell (approved):
I'm not a reviewer on this but just wanted to comment to say nice change!

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesFunctions.hlsl
```diff
@@ -224,6 +224,16 @@ half3 MixFog(real3 fragColor, real fogFactor)
     return MixFogColor(fragColor, unity_FogColor.rgb, fogFactor);
 }
 
+// Linear depth buffer value between [0, 1] or [1, 0] to eye depth value between [near, far]
+real LinearDepthToEyeDepth(real rawDepth)
```
phi-lira, line 228: My opinion is that we should add an override on Common.hlsl of LinearEyeDepth that handles orthographic. It's rather confusing that we have a similar function name in URP library and none of those functions specific if ortho is supported or not. 

Also it seems this one takes rawDepth unreversed depth while the others in Common.hlsl take depth.

phi-lira (changes requested):

phi-lira (approved):

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesFunctions.hlsl
eh-unity: We can't add an override as the value "space" is not encoded into the float in anyway, except by convention. This rescales linear depth (or reverse linear depth) into near<far view/eye space linear depth.
Functions in common.hlsl convert and rescale non-linear depth into view/eye space linear depth.
This also uses unity global constants directly, while common.hlsl functions are more generic and take everything as parameters.
REVERSE_Z-if is required to support OpenGL(ES).

PR #345, opened 2020-05-01
Fix for issue with Default material not available when Models are imported at project startup.

Thomasch-unity3d:
### **Please read**
**PR workflow guidelines**
* SRP ABV will start automatically on Yamato when you open your PR
* Changes to docs and md files will **not** trigger ABV jobs 
* Consider making use of **draft PRs** if you are not 100% sure that your PR is ready for review
* ABV will restart if you add a new commit to a branch with an open PR (hence why you should consider using draft PRs)
* Adding [skip ci] (case insensitive) to the title of PRs will stop any jobs being trigger automatically - you will need to open Yamato and find your branch to run ABV
* You can also add [skip ci] to commit messages to prevent CI from running on that push
* Add [cancel old ci] to your commit message if you've made changes you want to test and no longer need the previous jobs

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [ ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR.
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR

Fixes case 1141488 where Materials generated by the model importer fallback to the legacy default material rather than HDRP's default material.
The issue here is that HDRenderPipelineEditorResources fails to load using LoadAssetAtPath.
Using LoadSerializedFileAndForget succeeds but imples that changes to the HDRenderPipelineEditorResources asset are not picked up until Unity is restarted.

---
### Testing status

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in developer mode, you have a button at end of resources that check the paths)
- Other: 

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics

Any test projects to go with this to help reviewers?

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #3123, opened 2021-01-15
Hd/fix wizard runtime resources

RSlysz:
### Purpose of this PR
Fix bug https://fogbugz.unity3d.com/f/cases/edit/1287627 caused by a regression from PR https://github.com/Unity-Technologies/Graphics/pull/345 from a change after reviews. Bug was: in Wizard the runtime resource cannot be fixed once broken.

---
### Testing status
Manually tested that wizard fix work as intended.

---
### Comments to reviewers
@sebastienlagarde we must backport this to all version having the above mentioned PR.


sebastienlagarde (approved):

iM0ve (approved):
Minor issue with a workaround just top reopen the project. Developer tested that the wizard now runs properly. I think its safe to merge, no additional testing is required.

PR #3127, opened 2021-01-15
[HDRP][Path Tracing] Added alpha channel to path traced results

eturquin:
Addresses https://fogbugz.unity3d.com/f/cases/1304187/

The alpha channel of path-traced images was always set to 1.
Now, if there is no sky, it will use the alpha value of the clear color on camera ray misses (set to 0 by default), to match rasterization behaviour.

This change has been tested with and without presence of sky, and with a clear color with alpha going from 0 to 1:
- with a sky: output alpha is 1
- without a sky: output alpha is equal to alpha of the clear color
- if a ray hit a geometry, alpha is 1 (no matter if the material has refraction or not)

Note that for now, volumetrics are ignored and won't affect the alpha.


pmavridis (approved):

pmavridis:
Perhaps it is worth backporting this to 10.x? The PR seems simple-enough/back-portable.

sebastienlagarde (commented):
Agree for backporting to 10.x. 
The PR miss a changelog
and we should update the doc about alpha to say that it also work with path tracer 

https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.2/manual/Alpha-Output.html

eturquin:
> The PR miss a changelog
> and we should update the doc about alpha to say that it also work with path tracer

Ok, adding both.

sebastienlagarde (approved):

remi-chapelain (approved):
Tested with a camera outputting the path traced result to a render texture and then using the texture in a transparent and/or a opaque alpha clipped material. ✔️ 
Tested that it only worked with Sky setup to None ✔️ 
Worth noting a few things even though they're expected :
- HDRP Color buffer needs to be setup with an alpha channel (R16B16G16A16) for this PR to have any effect
- Doesn't work with Unity Recorder (Recorder does not care about alpha)
- Doesn't work with [CaptureScreenshot](https://docs.unity3d.com/ScriptReference/ScreenCapture.CaptureScreenshot.html) (Game view discards alpha)

Thanks to @pmavridis who helped me test this. 

https://user-images.githubusercontent.com/57442369/104930929-0a77d200-59a6-11eb-8884-cf4c72f8dfd2.mp4

PR #3128, opened 2021-01-15
[URP 2D] MSAA store actions

rolandk-unity:
---
### Purpose of this PR
This PR ensures that the correct `RenderBufferStoreAction` for all render targets using MSAA in the 2DRP.

Previously, `RenderBufferStoreAction.Store`, which is translated into "store and resolve" for MSAA targets by the low-level code, was used whenever setting a render target that needs to be stored, regardless of whether or not the multisampled content of the RT is still needed later in the frame. After this PR,  we use `RenderBufferStoreAction.Resolve` instead if the RT in question is being set for the last time in the frame. This prevents the multisampled contents being stored when they are not needed anymore and, ultimately, reduces the bandwidth usage when using MSAA on mobile platforms.

---
### Testing status
The test failures in Android_Vulkan_Standalone_il2cpp_Linear and 056_2D_Lights 056_2D_Lights_Shader_Graph are a known issue, which is probably fixed by https://github.com/Unity-Technologies/Graphics/pull/3146.

---
### Comments to reviewers
Setting the correct `RenderBufferStoreAction` is straightforward for normal maps, and in the unlit setting. For a lit main render target, it is a bit more challenging. Essentially, we need to identify the last `SetRenderTarget` call within the frame, use `RenderBufferStoreAction.Resolve` there, and `RenderBufferStoreAction.StoreAndResolve` in other locations.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
+++ b/com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
@@ -332,7 +332,7 @@ namespace UnityEngine.Experimental.Rendering.Universal
                     if (light.lightType == Light2D.LightType.Global)
                         continue;
 
-                    if (!light.renderVolumetricShadows)
+                    if (light.volumeIntensity <= 0.0f || !light.volumeIntensityEnabled)
                         continue;
 
                     var topMostLayerValue = light.GetTopMostLitLayer();


unity-cchu (approved):

com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
```diff
@@ -267,10 +267,11 @@ private static void RenderLightSet(IRenderPass2D pass, RenderingData renderingDa
             }
         }
 
-        public static void RenderLightVolumes(this IRenderPass2D pass, RenderingData renderingData, CommandBuffer cmd, int layerToRender, int endLayerValue, RenderTargetIdentifier renderTexture, RenderTargetIdentifier depthTexture, List<Light2D> lights)
+        public static void RenderLightVolumes(this IRenderPass2D pass, RenderingData renderingData, CommandBuffer cmd, int layerToRender, int endLayerValue,
```
phi-lira, line 270: This will be a public API breakage? Maybe that's ok if the 2D renderer is still experimental, but otherwise something to note.

phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
rolandk-unity: Actually, the class is internal, even if the method is public.

simon-engelbrecht-soerensen (approved):

PR #3131, opened 2021-01-15
[HDRP] Fix error in Depth Of Field near radius blur calculation

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1306228/
There was an error depending on the value of near max radius.

The near max radius controls the number of dilation passes. The problem was that the computation of passes during RG execution was taking resolution scale into account, but not during the RG pass setup. 

---
### Testing status
- Moving the near max radius slider is not causing any problems anymore (tested on the project attached in the bug ticket)


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -1498,6 +1498,12 @@ DepthOfFieldParameters PrepareDoFParameters(HDCamera camera)
             return parameters;
         }
 
+        static void GetDoResolutionScale(in DepthOfFieldParameters dofParameters, out float scale, out float resolutionScale)
```
FrancescoC-unity: I assume Do**F**  here.

Also, does it need to be static?

pmavridis: Yeah, I already fixed it :)

FrancescoC-unity:
LGTM

FrancescoC-unity (approved):

sebastienlagarde (approved):

PR #3137, opened 2021-01-15
fix 1299233 ies resize (CG.Alloc)

skhiat:
Fix:
https://fogbugz.unity3d.com/f/cases/1299233/ (cf. Video)
image
Repro bug:
To see Atlas Display Cookie Atlas (make sure Clear Cookie Atlas is *NOT* checked)
1. Change IES Size to 64
2. Change IES Size to 512
Notice when clear cookie atlas we don't have the issue.

This PR fix this issue.
Make sure that don't break the others features of the Atlas (for example Animated RenderTexture).
NZueNdZniI


sebastienlagarde:
@skhiat there is still issue with your PR, there is file not suppose to be there

sebastienlagarde:
use https://github.com/Unity-Technologies/Graphics/pull/3243

PR #3146, opened 2021-01-18
[URP 2D] Fix: Normal map rendering may corrupt depth on mobile

rolandk-unity:
### Purpose of this PR

This PR fixes an issue with the normal map rendering in the 2DRP when using a depth buffer. 

When a depth buffer is present, it is shared between the normal map rendering pass and the main render pass. Previously, the normal map rendering would use a `BufferStoreAction.DontCare` for this depth buffer, giving permission to the graphics driver to potentially corrupt its contents. This could lead subsequent draw calls to the main render target (or subsequent normal map draws) having a corrupt depth buffer.

This PR addresses the issue by changing the store action to `Store`.

---
### Testing status
Automated testing ongoing

---
### Comments to reviewers
Ongoing discussion: https://unity.slack.com/archives/C28K9NSTB/p1610715031014900


rainsing (approved):
Thanks for taking care of this!

PR #3150, opened 2021-01-18
Cleanup of Render Graph code (WIP)

JulienIgnace-Unity:
### Purpose of this PR
This PR is a first pass of cleanup of HDRP code now that we have only the render graph code path.
It focus only on the prepass passes (HDRenderPipeline.Prepass.cs)
- Merge XXXParameters to XXXPassData for all passes
- Removed PrepareXXXParameters and moved it to the pass build code.
- Removed most static sub functions
- Moved relevant code from HDRenderpipeline.cs to HDRenderpieline.Prepass.cs
- Replaced all render texture setup code by render graph helpers to avoid boiler plate code
- Removed all useless pass data parameters
- Moved prepass specific materials to the prepass file.
- Split depth prepass in two (prepass for deferred materials and prepass for forward/forwardOnly materials)
- Split ray tracing prepass in two explicit passes: depth prepass and flag mask pass.

---
### Testing status
This PR only move code around. As such automated tests should be enough.

### Note to reviewers
Main focus of the review should be to assess if those changes I made are for the better or if some things should be done differently to improve the code even further. Overall the code should be more compact and readable (almost 300 less lines of code) than before. Also navigation should be improved since we removed a lot of intermediate functions.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
@@ -323,48 +323,6 @@ namespace UnityEngine.Rendering.HighDefinition
             }
         }
 
-        class RayTracingFlagMaskPassData
-        {
-            public FrameSettings frameSettings;
-            public TextureHandle depthBuffer;
-            public TextureHandle flagMask;
-            public RendererListHandle opaqueRenderList;
-            public RendererListHandle transparentRenderList;
-            public bool clear;
-        }
-
-        TextureHandle RenderRayTracingFlagMask(RenderGraph renderGraph, CullingResults cull, HDCamera hdCamera, TextureHandle depthBuffer)
-        {
-            if (!hdCamera.frameSettings.IsEnabled(FrameSettingsField.RayTracing))
-                return renderGraph.defaultResources.blackTextureXR;
-
-            RecursiveRendering recursiveSettings = hdCamera.volumeStack.GetComponent<RecursiveRendering>();
-            if (!recursiveSettings.enable.value)
-                return renderGraph.defaultResources.blackTextureXR;
-
-            // This pass will fill the flag mask texture. This will only tag pixels for recursive rendering for now.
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
```diff
@@ -280,19 +288,28 @@ void BindMotionVectorPassColorBuffers(in RenderGraphBuilder builder, in PrepassO
...
+        class UpdateColorMaskForUnlitPassData
+        {
+            public ColorWriteMask colorWriteMask;
+        }
+
+        void UpdateColorMaskForUnlit(RenderGraph renderGraph, HDCamera hdCamera)
+        {
+            using (var builder = renderGraph.AddRenderPass<UpdateColorMaskForUnlitPassData>("UpdateColorMaskForUnlit", out var passData))
+            {
+                // Disable write to normal buffer for unlit shader (the normal buffer binding change when using MSAA)
```
sebastienlagarde, line 300: normal write buffer for unlit is enabled when unlit is a shadow matte (only supported with shader graph), so this comment is wrong.

FrancescoC-unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDProfileId.cs
```diff
@@ -18,7 +17,8 @@ internal enum HDProfileId
         Distortion,
         AccumulateDistortion,
         ApplyDistortion,
-        DepthPrepass,
+        ForwardDepthPrepass,
```
sebastienlagarde, line 20: we should rename to ForwardMaterialDepthPrepass and DeferredMaterialDepthPrepass  guess it will be more understandable

sebastienlagarde: ah nevermind, I was confuse, keep it like that

sebastienlagarde (approved):

anisunity (approved):
Looked quickly, looks good to me

PR #3151, opened 2021-01-18
Adds support for the PlayStation 5 platform

robinb-u3d:
### Purpose of this PR
This PR adds support for the PlayStation 5 platform to the SRP packages and the PPv2 package.

---
### Testing status
* HDRP - Local testing with the HDRP template project: Passed (PS5 and PS4).
* PS5 graphics tests: Running...
* 

---
### Comments to reviewers
PPV2 changes
* the optimizer is cleaver enough to produce the same result without us having to explicitly specify producing the same output in most cases. Screen space reflections is once case where it is slightly different but does still collapse down to the correct assembly just with a slightly different codegen pattern.
* PS5 build target is wrapped to allow this package to work even in older versions of Unity as PS5 will not show up as a valid enum member there.

### DIFF
--- a/com.unity.template-universal/Packages/com.unity.template.universal/CHANGELOG.md
+++ b/com.unity.template-universal/Packages/com.unity.template.universal/CHANGELOG.md
@@ -6,9 +6,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [12.0.0] - 2021-01-11
 
-### Added
-- Support for the PlayStation 5 platform has been added.
-
 ### Changed
 - Update SRP package to 12.0.0
 


rthompson-unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
Note: For the backport we need to wait for the various C++ PR to land, could you links all those PR here? thanks

Chman (approved):

robinb-u3d:
PS5 support is available in the following public builds on Unity so backports for the SRP versions that support those can go ahead making sure to update the minimum supported version in the packages otherwise developers will get compile errors where we reference PS5 in enums.
2020.2.3f1
2021.1.0a8
2021.2

Landing in public 2019.4 (so SRP 7.X) is in progress and I'll ping you on the PR when it it's ready to land so you can track it's progress so don't backport that version just yet.

phi-lira (approved):

sebastienlagarde (approved):

sebastienlagarde:
Don't know why Universal RP have been added whereas Felipe approved. Also the SDETS is only because a change in exclusion file. I am forcing the merge as Yamato is ok and PR approve.

sebastienlagarde:
@robinb-u3d  could you prepare a PR for 7.x.x/release? This one can't be cherry pick, there is a lot of conflcit. Thanks

sebastienlagarde:
+ I am not sure if the code is available on 19.4?

robinb-u3d:
@sebastienlagarde I'll bring up a separate PR for 7.x.x/release. Code has not landed yet though is in progress.

PR #3159, opened 2021-01-18
Fix ShaderGraph Custom Editor GUI in URP and HDRP

alelievr:
### Purpose of this PR
Issue: https://unity.slack.com/archives/CAR8G94EL/p1610921687007300

This PR fixes the Custom Editor GUI field in the ShaderGraph inspector (the field value was ignored).
image

---
### Testing status
- Tested that custom editor works in ShaderGraph for both URP and HDRP.
- Tested that default inspectors sill works as expected.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Version Updated
 The version number for this package has increased due to a version update of a related graphics package.
 
+### Fixed
+- Fixed the Custom Editor GUI field in the Graph settings that was ignored.
+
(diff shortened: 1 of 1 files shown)


phi-lira (approved):

sebastienlagarde (approved):

LandonTownsendUnity (approved):
Validated that nothing goes wrong when the materials with custom GUIs are viewed in built in.
Validated expected results with custom GUI when HDRP, URP, or both targets are in the shader, in HDRP and URP.
In a shader with both targets, validated that when two different custom GUIs are set, the right one shows up in the right SRP.
Validated that the correct GUI appears when no custom GUIs are set.
Validated that when a shader has both targets and one of them has a custom GUI, the other one loads the default GUI properly.

PR #3167, opened 2021-01-19
Added some misc utils to deal with SH2 with more intuitive functions

FrancescoC-unity:
Added some verbose functions to access and set data on SphericalHarmonicsL2  which might be not super-intuitive at first read otherwise. 

EDIT: Tested and all good seems like.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/SphericalHarmonics.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/SphericalHarmonics.cs
@@ -210,7 +210,7 @@ namespace UnityEngine.Rendering
         /// <param name ="sh">The SphericalHarmonicsL2 data structure to get information from.</param>
         /// <param name ="index">The index of the coefficient that is requested (must be less than 9).</param>
         /// <returns>The value of the requested coefficient.</returns>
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

com.unity.render-pipelines.core/Runtime/Lighting/SphericalHarmonics.cs
```diff
@@ -68,4 +68,152 @@ public override int GetHashCode()
...
+            sh[2, index] = coefficient.z;
+        }
+
+        /// <summary>
+        /// Get a spherical harmonics coefficient.
+        /// </summary>
+        /// <param name ="sh">The SphericalHarmonicsL2 data structure to get information from.</param>
+        /// <param name ="index">The index of the coefficient that is requested (must be less than 9).</param>
+        /// <returns>The value of the requested coefficient.</returns>
+        public static Vector3 GetCoefficient(ref SphericalHarmonicsL2 sh, int index)
```
FrancescoC-unity, line 213: This should not be ref, going to fix in a minute.

thefranke (approved):

PR #3169, opened 2021-01-19
[HDRP][URP] Added warning message in Projector component inspector UI that the component is not supported

arttu-peltonen:
### Purpose of this PR
This PR fixes the following bug: https://fogbugz.unity3d.com/f/cases/1300327/

---
### Changes
- Added custom inspectors for both URP and HDRP, that make it clear that the component is not supported and disables the inspector UI.
- Added changelog under both URP and HDRP.

HDRP:
projector-warning-hdrp
URP:
projector-warning-urp

---
### Testing status
Tested manually that the intended message appears in Projector component inspector depending on whether HDRP or URP is the active render pipeline. If no render pipeline asset is assigned (=builtin pipeline used), Projector component inspector is unchanged.

---
### Comments to reviewers
- Checked with UX that the warning message content and appearance are OK.
- In my opinion, it is difficult for a user migrating a project from builtin render pipeline into SRP to be aware that Projector component does not work anymore. With this change, it will be clear why the component is not working if they happen to select a Projector GameObject, but we are not addressing the "discoverability problem". Let me know if you have ideas to improve on this, even if that might be outside the scope of this PR.
- My first PR, don't hesitate to let me know if I'm doing anything wrong :)

### DIFF
--- a/com.unity.render-pipelines.core/Editor/ProjectorEditor.cs
+++ b/com.unity.render-pipelines.core/Editor/ProjectorEditor.cs
@@ -7,11 +7,11 @@ namespace UnityEditor.Rendering
     [CanEditMultipleObjects]
     class ProjectorEditor : Editor
     {
-        const string k_Message = "The active render pipeline does not support the Projector component. If using HDRP, use the Decal Projector component instead.";
+        static readonly GUIContent k_Message = EditorGUIUtility.TrTextContent("The active render pipeline does not support the Projector component. If using HDRP, use the Decal Projector component instead.");
 
         public override void OnInspectorGUI()
         {
-            EditorGUILayout.HelpBox(k_Message, MessageType.Warning);
+            EditorGUILayout.HelpBox(k_Message.text, MessageType.Warning);
 
             using (new EditorGUI.DisabledScope(true))
             {


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HighDefinitionRenderPipelineProjectorEditor.cs
```diff
@@ -0,0 +1,22 @@
+using UnityEngine;
+using UnityEngine.Rendering.HighDefinition;
+
```
sebastienlagarde, line 3: Regarding file name, use the shortcut:

HighDefinitionRenderPipelineProjectorEditor -> HDProjectorEditor

Note: I suppose you put the name of the pipeline in class name to avoid collision when the two pipe are present?
There is namespace here UnityEditor.Rendering.HighDefinition. So question, do we really need to have different name?

Guess typeof(HDRenderPipelineAsset) is enough to not collide, so we could name both editor in both package 
ProjectorEditor.

Last question, do we need two files at all?
Maybe we can put the editor declaration in Core package without specifiy the type of RP asset ?
 [CustomEditorForRenderPipeline(typeof(Projector))]
 And it will apply to all pipeline?
 
 (cc @jenniferd-unity )

sebastienlagarde: Note: I am not a UI guy, so it is just random guess

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineProjectorEditor.cs
```diff
@@ -0,0 +1,22 @@
+using UnityEngine;
+using UnityEngine.Rendering.Universal;
+
+namespace UnityEditor.Rendering.Universal
+{
+    [CustomEditorForRenderPipeline(typeof(Projector), typeof(UniversalRenderPipelineAsset))]
+    [CanEditMultipleObjects]
+    public class UniversalRenderPipelineProjectorEditor : Editor
```
phi-lira, line 8: This class doesn't need to be public. Otherwise we have to document / generate API docs for it.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HighDefinitionRenderPipelineProjectorEditor.cs
arttu-peltonen: Thanks for the comment. I tried to follow existing naming conventions, but I agree the names are a bit convoluted already. I would be fine with just using "ProjectorEditor" for both since they are in different namespaces.

I initially thought about putting it in RP Core, but then you can't customize the error message based on which RP is active (since UniversalRenderPipelineAsset and HDRenderPipelineAsset are not necessarily available). Unless there's some other way to do that neatly.

We could also decide that the RP-specific message is not required, in which case we could have it in RP Core. Originally suggested text included the conditional "... If HDRP is used, use the Decal Projector component instead" which is less clean but conveys the same information.

phi-lira: **Maybe we can put the editor declaration in Core package without specifiy the type of RP asset ?**

This is valid concern, it would also make the editor consistent for all SRP, including custom.

RSlysz: I agree with the renaming. Plus, up to now, when we named HDxxx or Universalxxx, it was because xxx was something in HDRP or UniversalRP. But here it is not the case. We just override for clarification.

And if we want to deactivate it only for HDRP and UniversalRP, this is ok. But if we want to deactivate it also for any CustomRP, I think Seb as a point here. What is the intent regarding custom SRP?

phi-lira (changes requested):

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HighDefinitionRenderPipelineProjectorEditor.cs
RSlysz: For customization of message, if all in CoreRP, you can grab the render pipeline asset currently used and check its type I guess.

JulienIgnace-Unity (approved):
LGTM but maybe we can actually have only one file in Core like Seb said and query something like "currentPipeline.supportsProjectors" ? This would require a new API though...
If we don't want to go that far then it's probably fine to have this copy pasted as this is not something we are going to change often anyway.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HighDefinitionRenderPipelineProjectorEditor.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        const string k_Message = "The active render pipeline (HDRP) does not support the Projector component. Use the Decal Projector component instead.";
+
+        public override void OnInspectorGUI()
+        {
+            EditorGUILayout.HelpBox(k_Message, MessageType.Warning);
+
+            using (new EditorGUI.DisabledScope(true))
+            {
+                DrawDefaultInspector();
```
RSlysz, line 18: In this special case this is OK.

Though, here is a few things to take into consideration.
- This only works if there is no specific editor for Projector. If there is a dedicated editor for it in built-in, display can differ from default one. (can be different placement, tooltips, edit mods etc).
- Simplest is to find if there is a [CustomEditor(typeof(Projector))] somewhere in Unity. If there is one, you will have differences. In this case, I recommend instead to create an Inspector for Projector using Editor.CreateEditor and display it here to be sure the display remain the same.

Here I think just a small comment to explain why we do not have to do this could help in case of maintenance needed on it.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HighDefinitionRenderPipelineProjectorEditor.cs
```diff
@@ -0,0 +1,22 @@
+using UnityEngine;
+using UnityEngine.Rendering.HighDefinition;
+
+namespace UnityEditor.Rendering.HighDefinition
+{
+    [CustomEditorForRenderPipeline(typeof(Projector), typeof(HDRenderPipelineAsset))]
+    [CanEditMultipleObjects]
+    public class HighDefinitionRenderPipelineProjectorEditor : Editor
```
RSlysz, line 8: Editor class are usually private/internal.

We want to only expose in the user API what have meaning for them.

RSlysz (changes requested):
Check what is the intend for Custom SRP

Check visibility of the Editor class.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HighDefinitionRenderPipelineProjectorEditor.cs
arttu-peltonen: Thanks, I'll add a comment that there's no custom editor for Projector.

> For customization of message, if all in CoreRP, you can grab the render pipeline asset currently used and check its type I guess.
As far as I understand, existence of CoreRP package doesn't imply either URP or HDRP package are available, which is where those types are defined. Can I do this without hard-coding the typenames ("UniversalRenderPipelineAsset" and "HighDefinitionRenderPipelineAsset")?

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HighDefinitionRenderPipelineProjectorEditor.cs
arttu-peltonen: Maybe someone has better info about this, but I've understood "Projector" is bit of a legacy feature. So if neither URP nor HDRP support it, it would make sense to me that by default custom SRP don't support it either. I guess they could always choose to write a custom inspector that enables the UI if really needed.

Sounds to me like benefits of having only one custom editor for all SRPs are bigger than the drawback of not being able to tailor the message specifically for HDRP.

sebastienlagarde: yup, we are all align :)

arttu-peltonen:
Updated based on feedback, new generic UI looks like this:
image (1)

alex-vazquez (approved):

com.unity.render-pipelines.core/Editor/ProjectorEditor.cs
```diff
@@ -0,0 +1,23 @@
+using UnityEngine;
+using UnityEngine.Rendering;
+
+namespace UnityEditor.Rendering
+{
+    [CustomEditorForRenderPipeline(typeof(Projector), typeof(RenderPipelineAsset))]
+    [CanEditMultipleObjects]
+    class ProjectorEditor : Editor
+    {
+        const string k_Message = "The active render pipeline does not support the Projector component. If using HDRP, use the Decal Projector component instead.";
```
jenniferd-unity, line 10: should it be a GUI_Content to support localization? i am not sure how this works

jenniferd-unity (approved):
Added comment about Localization but it may already be correct. Otherwise all good!

com.unity.render-pipelines.core/Editor/ProjectorEditor.cs
arttu-peltonen: @alex-vazquez flagged this for me as well. I looked into it and initially thought localization is not supported by the API, but looking back into it I found some use cases from the main editor codebase that do exactly that. I'll fix this.

RSlysz (approved):
lgtm

com.unity.render-pipelines.core/Editor/ProjectorEditor.cs
```diff
@@ -0,0 +1,23 @@
+using UnityEngine;
+using UnityEngine.Rendering;
+
+namespace UnityEditor.Rendering
+{
+    [CustomEditorForRenderPipeline(typeof(Projector), typeof(RenderPipelineAsset))]
+    [CanEditMultipleObjects]
+    class ProjectorEditor : Editor
+    {
+        static readonly GUIContent k_Message = EditorGUIUtility.TrTextContent("The active render pipeline does not support the Projector component. If using HDRP, use the Decal Projector component instead.");
```
phi-lira, line 10: Nitpicking, could be improved for follow up PR, but you could check which RP is running at runtime if it's HDRP, URP, or something else and then display the approapriate message.

phi-lira (approved):

sebastienlagarde (approved):

sebastienlagarde:
Hey, so Yamato is green, the various error in Yamato come from Master not from the PR. Good to merge.

PR #3172, opened 2021-01-19
Lighting/pv2 l2 support

thefranke:
### Purpose of this PR
This PR adds *basic* L2 support. The current code is not compressing the range of L2 coefficients in any way, which forces a different layout of the coefficients in the textures than the current naming suggests. This has two downsides:

1. L0 cannot be queried with a single texture tap and instead requires three.
2. L1 and L2 cannot use textures with fewer bits.

A second PR will add those fixes. This PR further depends on [this pull request](https://github.com/Unity-Technologies/Graphics/pull/3167) which adds a nicer way of retrieving and setting coefficients.

---
### Testing status
Manual testing done with the temple scene. 

---
### Comments to reviewers
To activate L2, go to the HDRP asset and change the coefficient range in the ProbeVolume section using the drop-down.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
+++ b/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
@@ -169,9 +169,9 @@ namespace UnityEngine.Rendering
                     int j = bakingCells[c].probeIndices[i];
                     SphericalHarmonicsL2 shv = sh[j];
 
-                    // It can be shown that |L1_i| <= |2*L0|
-                    // Precomputed Global Illumination in Frostbite by Yuriy O'Donnell.
-                    // https://media.contentapi.ea.com/content/dam/eacom/frostbite/files/gdc2018-precomputedgiobalilluminationinfrostbite.pdf
+                    // Compress the range of all coefficients but the DC component to [0..1]
+                    // Upper bounds taken from http://ppsloan.org/publications/Sig20_Advances.pptx
+                    // Divide each coefficient by DC*f to get to [-1,1] where f is from slide 33
                     for (int rgb = 0; rgb < 3; ++rgb)
                     {
                         var l0 = sh[j][rgb, 0];
@@ -179,29 +179,22 @@ namespace UnityEngine.Rendering
                         if (l0 == 0.0f)
                             continue;
 
-                        // TODO: Enlarging range by 2 because we're using irradiance probes. Need to derive correct factors and remove this.
-                        // Compare ProbeVolume.hlsl
-                        l0 *= 2.0f;
+                        // TODO: We're working on irradiance instead of radiance coefficients
+                        //       Add safety margin 2 to avoid out-of-bounds values
+                        float l1scale = 1.7320508f; // 3/(2*sqrt(3)) * 2
+                        float l2scale = 3.5777088f; // 4/sqrt(5) * 2
 
                         // L_1^m
-                        shv[rgb, 1] = sh[j][rgb, 1] / (l0 * 2.0f) + 0.5f;
-                        shv[rgb, 2] = sh[j][rgb, 2] / (l0 * 2.0f) + 0.5f;
-                        shv[rgb, 3] = sh[j][rgb, 3] / (l0 * 2.0f) + 0.5f;
-                        
+                        shv[rgb, 1] = sh[j][rgb, 1] / (l0 * l1scale * 2.0f) + 0.5f;
+                        shv[rgb, 2] = sh[j][rgb, 2] / (l0 * l1scale * 2.0f) + 0.5f;
+                        shv[rgb, 3] = sh[j][rgb, 3] / (l0 * l1scale * 2.0f) + 0.5f;
+
                         // L_2^-2
-                        shv[rgb, 4] = sh[j][rgb, 4] / (l0 * 3.75f) + 0.5f;
-
-                        // L_2^-1
-                        shv[rgb, 5] = sh[j][rgb, 5] / (l0 * 3.75f) + 0.5f;
-
-                        // L_2^0
-                        shv[rgb, 6] = sh[j][rgb, 6] / (l0 * 0.3125f * 4) + 0.5f;
-
-                        // L_2^1
-                        shv[rgb, 7] = sh[j][rgb, 7] / (l0 * 3.75f) + 0.5f;
-
-                        // L_2^2
-                        shv[rgb, 8] = sh[j][rgb, 8] / (l0 * 0.9375f) + 0.5f;
+                        shv[rgb, 4] = sh[j][rgb, 4] / (l0 * l2scale * 2.0f) + 0.5f;
+                        shv[rgb, 5] = sh[j][rgb, 5] / (l0 * l2scale * 2.0f) + 0.5f;
+                        shv[rgb, 6] = sh[j][rgb, 6] / (l0 * l2scale * 2.0f) + 0.5f;
+                        shv[rgb, 7] = sh[j][rgb, 7] / (l0 * l2scale * 2.0f) + 0.5f;
+                        shv[rgb, 8] = sh[j][rgb, 8] / (l0 * l2scale * 2.0f) + 0.5f;
 
                         // Assert coefficient range
                         for (int coeff = 1; coeff < 9; ++coeff)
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
```diff
@@ -79,12 +79,31 @@ APVConstants LoadAPVConstants( StructuredBuffer<int> index )
...
 float3 DecodeSH(float l0, float3 l1)
 {
-    return (l1 - 0.5) * 4.0 * l0;
+    // TODO: Enlarging range by 2 because we're using irradiance probes. Need to derive correct factors and remove this.
+    // Compare ProbeGIBaking.cs
+    return (l1 - 0.5f) * 2.0f * l0 * 2.0f;
 }
 
-float4 DecodeSH(float l0, float4 l1)
+void DecodeSH_L2(float l0, inout float4 l2_R, inout float4 l2_G, inout float4 l2_B, inout float4 l2_C)
```
FrancescoC-unity, line 87: I would just adding a comment here about the coefficients

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -166,40 +166,52 @@ private static void OnAdditionalProbesBakeCompleted()
...
-                        }
+                        if (l0 == 0.0f)
+                            continue;
+
+                        // TODO: Enlarging range by 2 because we're using irradiance probes. Need to derive correct factors and remove this.
+                        // Compare ProbeVolume.hlsl
+                        l0 *= 2.0f;
+
+                        // L_1^m
+                        shv[rgb, 1] = sh[j][rgb, 1] / (l0 * 2.0f) + 0.5f;
```
FrancescoC-unity, line 187: Again, I'd add comments to specify what is going on (the link above only shows L1)

sebastienlagarde: we should define fully the constant: 0.9375f

for example in AmbientProbeConvolutoin.compute we do
#define fC0 (rsqrt(PI * 4.0) * rsqrt(PI * 4.0))  // Equivalent (0.282095 * (1.0 / (2.0 * sqrtPI)))
#define fC1 (rsqrt(PI * 4.0 / 3.0) * rsqrt(PI * 3.0)) // Equivalent to (0.488603 * (sqrt ( 3.0) / ( 3.0 * sqrtPI)))
#define fC2 (rsqrt(PI * 4.0 / 15.0) * rsqrt(PI * 64.0 / 15.0)) // Equivalent to (1.092548 * (sqrt (15.0) / ( 8.0 * sqrtPI)))
#define fC3 (rsqrt(PI * 16.0 / 5.0) * rsqrt(PI * 256.0 / 5.0)) // Equivalent to (0.315392 * (sqrt ( 5.0) / (16.0 * sqrtPI)))
#define fC4 (rsqrt(PI * 16.0 / 15.0) * rsqrt(PI * 256.0 / 15.0)) // Equivalent to  (0.546274 * 0.5 * (sqrt (15.0) / ( 8.0 * sqrtPI)))

help a lot to understand where the constant come from.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
```diff
@@ -212,25 +258,51 @@ public static void FillDataLocation(ref DataLocation loc, SphericalHarmonicsL1[]
...
+                            // Second texture will have L1_G coefficients in RGB and L1R.y in the alpha channel
+                            Color L1G_L1Ry = new Color(L1G.x, L1G.y, L1G.z, L1R.y);
+                            // Third texture will have L1_B coefficients in RGB and L1R.z in the alpha channel
+                            Color L1B_L1Rz = new Color(L1B.x, L1B.y, L1B.z, L1R.z);
+
+                            loc.TexL0_L1rx.SetPixel(ix, iy, iz, L0_L1Rx);
+                            loc.TexL1_G_ry.SetPixel(ix, iy, iz, L1G_L1Ry);
+                            loc.TexL1_B_rz.SetPixel(ix, iy, iz, L1B_L1Rz);
+
+                            if (bands == ProbeVolumeSHBands.SphericalHarmonicsL2)
```
sebastienlagarde, line 276: Not sure to understand why we go through all SH2 index
for (int brickIdx = 0; brickIdx < shl2.Length; brickIdx += kBrickProbeCountTotal)

and test 
if (bands == ProbeVolumeSHBands.SphericalHarmonicsL2)

(haven't read the full code, just seems weird)

sebastienlagarde: naively I will have except something like:

for (int brickIdx = 0; brickIdx < (bands == ProbeVolumeSHBands.SphericalHarmonicsL2) ? shl2.Length : shl1.Length; brickIdx += kBrickProbeCountTotal)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
sebastienlagarde: Same remark than in C#.   Currently we could define the constant in C# and generate it for the hlsl code, so we are sure they match

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
```diff
@@ -61,25 +77,54 @@ APVConstants LoadAPVConstants( StructuredBuffer<int> index )
...
+    l2_C.x = (l2_C.x - 0.5) * 0.9375f * l0;
+    l2_C.y = (l2_C.y - 0.5) * 0.9375f * l0;
+    l2_C.z = (l2_C.z - 0.5) * 0.9375f * l0;
+}
 
 #define APV_USE_BASE_OFFSET
 
 // We split the evaluation in several steps to make variants with different bands easier.
-float3 EvaluateAPVL0(APVResources apvRes, float3 uvw)
+float3 EvaluateAPVL0(APVResources apvRes, float3 uvw, out float L1Rx)
```
sebastienlagarde, line 112: guess we should also add a 

float3 EvaluateAPVL0(APVResources apvRes, float3 uvw)
{
float unused;
    return EvaluateAPVL0(apvRes, uvw, out unused);
}

in case we only need EvaluateAPVL0 (for volumetric)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
FrancescoC-unity: The SH2 structure is always filled (to be able to not having to rebake in case a setting on asset is changed), but we load only the relevant bands to the runtime structures.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
FrancescoC-unity: That is not the public API, the volumetric already have their own simplified "public" API that still works ( https://github.com/Unity-Technologies/Graphics/blob/cc188aeec07e783aeaabab72f3d443cdd7de8a56/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl#L301 or https://github.com/Unity-Technologies/Graphics/blob/cc188aeec07e783aeaabab72f3d443cdd7de8a56/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl#L252 if the resources are passed separately)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
thefranke: Done and done

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
thefranke: Done and done

sebastienlagarde (approved):

FrancescoC-unity (approved):

PR #3141, opened 2021-01-15
[2021.2] Particle Shaders: Adding Depth & Depth Normals passes

ellioman:
# Purpose of this PR
Particles currently do not contribute to Depth or DepthNormal textures. This causes an issue with SSAO when the `After Opaque` option is enabled as it makes them look transparent.

This PR adds Depth and DepthNormals passes to particle shaders. I had to add a `#define _BumpScale 1.0` to the input hlsl files for simplelit and unlit in order to be able to call `SampleNormalTS(...)` using the same fragment shader for all particles.

A test scene will be added to the URP Graphics Test project in a followup PR.

# QA
### What has been tested
- Tested in Editor (OS X 10.14.6) with a custom test scene to be added in a followup PR in URP's Graphics Test Project
- Ran the Test Runner and made sure all tests were a success

### What needs testing
- Make sure particles render to depth and/or normals texture on all our supported platforms and API's
- Make sure this change hasn't regressed particle rendering

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Ffeature%252Fparticles-depth-and-normals

PR #3174, opened 2021-01-19
[2021.2] SSAO: NormalMaps, support for Unlit & Particles plus various improvements

ellioman:
# Purpose of this PR
This PR introduces various improvements and features to SSAO. 

### Open issues
- Android GLES3: Instability with the test results often showing capturing half of the screen. I've disabled this platform in the test filters and will create a issue for it. I will link to the case in the test filters file once the issue has been created.
- Linux Vulkan: Issue 1305639 is still open and affects the new test scenes. They are filtered out until the case is fixed.

### Quality of downscaled SSAO improved
Screen Shot 2021-01-19 at 16 50 21

### Performance
- SSAO final Texture is now `R8` instead of `ARGB32`
- Performance improved in the SSAO Passes, shaving almost a 1ms on an iPhone 8.

### Bugfixes
- Fixes an issue where Depth Prepass was not added when SSAO was set to Depth Only. 

### Main changes
- DepthNormal Passes are now split into separate .hlsl files for each shader type
- DepthNormal passes now take the NormalMaps/Detail/Parallax into account if present
- Unlit shaders now receive SSAO
- Particle Shaders now receive SSAO (Depth & DepthNormal passes added in #3141)
- Intensity is no longer a slider but a float field

### Other changes
- Added a OutputTextureFeature renderer feature and shaders to the test project for the depth and normals scenes
- Frame Timings settings were disabled for Android because of issue [1316285](https://issuetracker.unity3d.com/issues/opengles3-android-urp-scenes-are-rendered-after-a-period-of-time-of-black-screen-or-not-rendered-at-all-on-the-build)
- Blit Type setting is now set to Always because of an issue with two cameras and a full screen quad (under investigation)

# Improved test scenes and setup for SSAO
Previous scenes have been removed and new ones have been added. 

#### Removed test scenes
  - 140_SSAO_DepthOnly_Projection
  - 141_SSAO_DepthOnly_Orthographic
  - 143_SSAO_DepthNormals_Orthographic
  - 144_SSAO_RenderToBackBuffer
  - 145_SSAO_Deferred_Projection
  - 146_SSAO_Deferred_Orthographic
  - 147_SSAO_Deferred_RenderToBackBuffer

#### New test scene setup (Tests Projection & Orthographic cameras)
  - 200_DepthDeferred
  - 200_DepthForward
  - 201_DepthNormalsDeferred
  - 201_DepthNormalsForward
  - 202_SSAO_Depth
  - 203_SSAO_DepthNorma
  - 204_SSAO_Deferred
  - 205_SSAO_BackBuffer

#### The new scenes now test
- Lit Shader (Without Normal maps, with normal maps, alpha clipped, normalmap + detail + parallax)
- ComplexLit (Without Normal maps, with normal maps, alpha clipped, normalmap + detail + parallax)
- SimpleLit (Without Normal maps, with normal maps, alpha clipped)
- BakedLit (Without Normal maps, with normal maps, alpha clipped)
- Unlit
- Particles Lit (Without Normal maps, with normal maps, alpha clipped)
- Particles Simple Lit (Without Normal maps, with normal maps, alpha clipped)
- Particles Unlit (Without Normal maps, with normal maps, alpha clipped)
- SpeedTree 7 (Without Normal maps, with normal maps)
- SpeedTree 8 (Without Normal maps, with normal maps)
- Terrain
- Terrain Waving Grass
- Three different ShaderGraph shaders

# QA
### What has been tested
- Tested in Editor (OS X 10.14.6) with the new test scenes, made sure all pass
- Tested a build on an iPhone 8 (iOS 14.3) with one of the new test scenes
- Tested a build on Galaxy S10+ (Mali-G76 MP12) with GLES3 and Vulkan
- Tested a build on Galaxy Note10+ 5G (Adreno 640) with GLES3 and Vulkan

### What needs testing
- Test if there are any regressions with these changes, especially on mobile platforms & API's
- Test performance difference

### Performance Docs
- [Android](https://docs.google.com/spreadsheets/d/1BwQxOJXhzk4X6GeSvPqUlbuzKSFYtH_N_oIpxKsvtbg/edit#gid=1452499847&range=A1:D1)
- [iOS](https://docs.google.com/spreadsheets/d/149ac28KGkTC3G-AtvwrPVs8v57B_eo0he8FcYT05HmA/edit#gid=0)

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fssao-normalmaps-and-improvements


TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/200_Assets/Scripts/OutputTextureFeature.cs
```diff
@@ -0,0 +1,97 @@
+using UnityEngine;
```
ellioman, line 1: This Renderer Feature is used to display the Depth, Normals or SSAO texture in the new test scenes. 
It is only in the projects, not part of public URP codebase.

TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/200_Assets/Scripts/SwitchRenderingMode.cs
```diff
@@ -0,0 +1,23 @@
+using System.Collections;
```
ellioman, line 1: This script changes lighting mode in the new test scenes to prevent me having to create a separate renderer for each test scene

TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/200_Assets/Shaders/OutputDepthNormalsTexture.shader
```diff
@@ -0,0 +1,58 @@
+Shader "Hidden/Test/OutputDepthNormalsTexture"
```
ellioman, line 1: This shader is used with the OutputTextureFeature to display the Normals texture.

TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/200_Assets/Shaders/OutputDepthTexture.shader
```diff
@@ -0,0 +1,62 @@
+Shader "Hidden/Test/OutputDepthTexture"
```
ellioman, line 1: This shader is used with the OutputTextureFeature to display the depth texture.

TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/200_Assets/Shaders/OutputSSAOTexture.shader
```diff
@@ -0,0 +1,58 @@
+Shader "Hidden/Test/OutputDepthNormalsTexture"
```
ellioman, line 1: This shader is used with the OutputTextureFeature to display the SSAO texture.

com.unity.render-pipelines.universal/Shaders/Utils/ScreenSpaceAmbientOcclusion.shader
```diff
@@ -127,15 +127,14 @@ Shader "Hidden/Universal Render Pipeline/ScreenSpaceAmbientOcclusion"
...
-                #include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl"
                 #include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl"
 
                 half4 FragAfterOpaque(Varyings input) : SV_Target
                 {
                     UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
                     AmbientOcclusionFactor aoFactor = GetScreenSpaceAmbientOcclusion(input.uv);
-                    half occlusion = aoFactor.directAmbientOcclusion;
+                    half occlusion = aoFactor.indirectAmbientOcclusion;
```
ellioman, line 137: Using the "After Opaque" feature should apply the full AO effect.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/DepthNormalsOnlyPass.hlsl
```diff
@@ -24,14 +24,29 @@ half4 frag(PackedVaryings packedInput) : SV_TARGET
     #endif
 
     #if defined(_GBUFFER_NORMALS_OCT)
-    float3 normalWS = normalize(unpacked.normalWS);
-    float2 octNormalWS = PackNormalOctQuadEncode(normalWS);           // values between [-1, +1], must use fp32 on some platforms
-    float2 remappedOctNormalWS = saturate(octNormalWS * 0.5 + 0.5);   // values between [ 0,  1]
-    half3 packedNormalWS = PackFloat2To888(remappedOctNormalWS);      // values between [ 0,  1]
-    return half4(packedNormalWS, 0.0);
+        float3 normalWS = normalize(unpacked.normalWS);
+        float2 octNormalWS = PackNormalOctQuadEncode(normalWS);           // values between [-1, +1], must use fp32 on Nintendo Switch.
```
phi-lira, line 28: Is this a bad merge? it seems like you are reverting changes to the comment

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
```diff
@@ -164,11 +169,19 @@ internal bool Setup(ScreenSpaceAmbientOcclusionSettings featureSettings, Scripta
...
                 m_CurrentSettings = featureSettings;
-                renderPassEvent = featureSettings.AfterOpaque ? RenderPassEvent.AfterRenderingOpaques : RenderPassEvent.AfterRenderingGbuffer;
 
-                ScreenSpaceAmbientOcclusionSettings.DepthSource source = isRendererDeferred
-                    ? ScreenSpaceAmbientOcclusionSettings.DepthSource.DepthNormals
-                    : m_CurrentSettings.Source;
+                ScreenSpaceAmbientOcclusionSettings.DepthSource source;
+                if (isRendererDeferred)
+                {
+                    renderPassEvent = featureSettings.AfterOpaque ? RenderPassEvent.AfterRenderingOpaques : RenderPassEvent.AfterRenderingGbuffer;
```
phi-lira, line 176: Not related to this PR, but question:
Does SSAO need to run before skybox? We plan to move option to copy depth after skybox so we don't cause render target switch between opaque and skybox.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
```diff
@@ -279,19 +292,27 @@ public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderin
...
+                m_AOPassDescriptor = descriptor;
+                m_AOPassDescriptor.width /= downsampleDivider;
+                m_AOPassDescriptor.height /= downsampleDivider;
+                m_AOPassDescriptor.colorFormat = RenderTextureFormat.ARGB32;
+
+                m_BlurPassesDescriptor = descriptor;
+                m_BlurPassesDescriptor.colorFormat = RenderTextureFormat.ARGB32;
+
+                m_FinalDescriptor = descriptor;
+                m_FinalDescriptor.colorFormat = RenderTextureFormat.R8;
```
phi-lira, line 309: Do all platforms that support SSAO support R8 format?

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -45,81 +91,77 @@ float4 _CameraViewZExtent[2];
...
     #endif
 }
 
-float CompareNormal(float3 d1, float3 d2)
+half CompareNormal(half3 d1, half3 d2)
 {
-    return smoothstep(kGeometryCoeff, 1.0, dot(d1, d2));
-}
-
-float2 GetScreenSpacePosition(float2 uv)
```
phi-lira, line 95: Is this function being removed?

com.unity.render-pipelines.universal/Shaders/ComplexLit.shader
```diff
@@ -257,6 +259,8 @@ Shader "Universal Render Pipeline/Complex Lit"
             // -------------------------------------
             // Material Keywords
             #pragma shader_feature_local _NORMALMAP
+            #pragma shader_feature_local _PARALLAXMAP
+            #pragma shader_feature_local _ _DETAIL_MULX2 _DETAIL_SCALED
```
phi-lira, line 263: Can any of these features be `_fragment` cc @lukaschod

com.unity.render-pipelines.universal/Shaders/SimpleLit.shader
```diff
@@ -245,6 +245,7 @@ Shader "Universal Render Pipeline/Simple Lit"
 
             // -------------------------------------
             // Material Keywords
+            #pragma multi_compile SIMPLE_LIT
```
phi-lira, line 248: Is this supposed to be be #define SIMPLE_LIT, otherwise I don't get why we are adding a multi_compile for simple lit.

phi-lira (approved):

phi-lira:
Big PR need more reviewer coverage. I'd say to add someone from Terrain and Particles at least.

phi-lira:
Maybe also @kaychang-unity as he touched several of these files.

com.unity.render-pipelines.universal/Shaders/SimpleLit.shader
ellioman: This was leftover when splitting the DepthNormals passes. Should not be there.
Removing in next commit.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
ellioman: Yes, I was doing that to clean up a little bit as it was only used in one place.
But to avoid any issues with users potentially using that function I'm adding it back in the next commit.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
ellioman: Good point! 
I'm not sure so I'll add a `SystemInfo.SupportsRenderTextureFormat(R8)` check for that to be on the safe side.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/DepthNormalsOnlyPass.hlsl
ellioman: Yes and no. Only the comment seems to be incorrect here, the code is the same (apart from me indenting things) from the original.
Will fix the comment in the next commit.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -249,62 +305,62 @@ void SampleDepthNormalView(float2 uv, out float depth, out float3 normal, out fl
...
+    for (int s = 0; s < SAMPLE_COUNT; s++)
     {
-        #if defined(SHADER_API_D3D11)
-            // This 'floor(1.0001 * s)' operation is needed to avoid a DX11 NVidia shader issue.
-            s = floor(1.0001 * s);
-        #endif
-
         // Sample point
-        float3 v_s1 = PickSamplePoint(uv, randAddon, s); // (kchang) should we rotate this "random" vector to world space?
+        half3 v_s1 = PickSamplePoint(uv, s); // (kchang) should we rotate this "random" vector to world space?
```
eh-unity, line 328: Is there an answer to this question?
Since there are multiple spaces, we should document the spaces for position variables. Either use our space 
postfix convention like WS, CS etc. or just add comments.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -319,63 +375,58 @@ float4 SSAO(Varyings input) : SV_Target
 // Geometry-aware separable bilateral filter
 half4 Blur(float2 uv, float2 delta) : SV_Target
 {
-    float4 p0 = SAMPLE_BASEMAP(uv                 );
-    float4 p1a = SAMPLE_BASEMAP(uv - delta * 1.3846153846);
-    float4 p1b = SAMPLE_BASEMAP(uv + delta * 1.3846153846);
-    float4 p2a = SAMPLE_BASEMAP(uv - delta * 3.2307692308);
-    float4 p2b = SAMPLE_BASEMAP(uv + delta * 3.2307692308);
+    half4 p0 =  (half4) SAMPLE_BASEMAP(uv                 );
+    half4 p1a = (half4) SAMPLE_BASEMAP(uv - delta * 1.3846153846);
```
eh-unity, line 379: How are these constants obtained? Are these UV adjustments to use bilinear sampling to do part of the blur? We should have a named constants for them.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -319,63 +375,58 @@ float4 SSAO(Varyings input) : SV_Target
...
     #endif
 
-    float w0  =                                           0.2270270270;
-    float w1a = CompareNormal(n0, GetPackedNormal(p1a)) * 0.3162162162;
-    float w1b = CompareNormal(n0, GetPackedNormal(p1b)) * 0.3162162162;
-    float w2a = CompareNormal(n0, GetPackedNormal(p2a)) * 0.0702702703;
-    float w2b = CompareNormal(n0, GetPackedNormal(p2b)) * 0.0702702703;
+    half w0  =                                           half(0.2270270270);
+    half w1a = CompareNormal(n0, GetPackedNormal(p1a)) * half(0.3162162162);
+    half w1b = CompareNormal(n0, GetPackedNormal(p1b)) * half(0.3162162162);
```
eh-unity, line 396: Named constants.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -319,63 +375,58 @@ float4 SSAO(Varyings input) : SV_Target
 // Geometry-aware separable bilateral filter
 half4 Blur(float2 uv, float2 delta) : SV_Target
```
eh-unity, line 376: We could easily need this for some other effect too. Can we generalize it in a meaningful way? Or break into "building block" functions?

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainDetailLit.shader
```diff
@@ -247,7 +249,7 @@ Shader "Hidden/TerrainEngine/Details/UniversalPipeline/Vertexlit"
                 surfaceData.occlusion = 1.0;
 
                 InputData inputData = (InputData)0;
-                inputData.normalWS = half3(0, 1, 0); // need some default to avoid division by 0.
+                inputData.normalWS = input.NormalWS;
```
eh-unity, line 252: can input.NormalWS be (0,0,0)? Is the previous comment relevant here?

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
ellioman: I think that's a good idea.
My next SSAO PR will tackle different Blur Options, so I'll split the Blurring code into a separate file and see if I can be more general.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -45,81 +91,82 @@ float4 _CameraViewZExtent[2];
...
 // filter. The higher value, the more sensitive it is.
-static const float kGeometryCoeff = 0.8;
+static const half kGeometryCoeff = half(0.8);
 
 // The constants below are used in the AO estimator. Beta is mainly used for suppressing
 // self-shadowing noise, and Epsilon is used to prevent calculation underflow. See the
-// paper (Morgan 2011 http://goo.gl/2iz3P) for further details of these constants.
-static const float kBeta = 0.002;
-#define EPSILON         1.0e-4
+// paper (Morgan 2011 https://bit.ly/3uAPRgz) for further details of these constants.
```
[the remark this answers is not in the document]
ellioman: Yeah, the link was broken so I replaced it. 
Think you're right about having a direct link so I'll fix that in the next commit.

manuele-bonanno (approved):
the SSAO.hlsl changes address the ALU performance issues we found on iOS. Good stuff!

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -15,15 +15,61 @@ SAMPLER(sampler_BaseMap);
 SAMPLER(sampler_ScreenSpaceOcclusionTexture);
 
 // Params
-float4 _SSAOParams;
+half4 _SSAOParams;
```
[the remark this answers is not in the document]
ellioman: They could be, but that will require work on the C# side for the constant buffer vs uniform approaches as it's not supported everywhere. But could have a performance improvement so I'll note this down as a future task.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
ellioman: Agree! 
I've removed the comment as we shouldn't have discussions in the codebase and noted it down for the future work.
Allow me to tweak the variable naming with proper postfixes in the followup PR.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -45,81 +91,82 @@ float4 _CameraViewZExtent[2];
...
-float2 GetScreenSpacePosition(float2 uv)
+half CompareNormal(half3 d1, half3 d2)
 {
-    return uv * SCREEN_PARAMS.xy * DOWNSAMPLE;
+    return smoothstep(kGeometryCoeff, half(1.0), dot(d1, d2));
 }
 
 // Trigonometric function utility
-float2 CosSin(float theta)
+half2 CosSin(half theta)
```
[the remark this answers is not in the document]
ellioman: Good point. As I'd like not to touch many more files in this PR, allow me to do that in the follow up PR.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
ellioman: Noted! I'll adress this in the Blur PR for SSAO.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
ellioman: Noted. Will adress in the follow-up PR.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/DepthNormalsOnlyPass.hlsl
```diff
@@ -24,14 +24,29 @@ half4 frag(PackedVaryings packedInput) : SV_TARGET
...
+        return half4(packedNormalWS, 0.0);
     #else
-    float3 normalWS = NormalizeNormalPerPixel(unpacked.normalWS);
-    return half4(normalWS, 0.0);
+        // Retrieve the normal from the bump map or mesh normal
+        #if defined(_NORMALMAP)
+            #if _NORMAL_DROPOFF_TS
+                // IMPORTANT! If we ever support Flip on double sided materials ensure bitangent and tangent are NOT flipped.
+                float crossSign = (unpacked.tangentWS.w > 0.0 ? 1.0 : -1.0) * GetOddNegativeScale();
+                float3 bitangent = crossSign * cross(unpacked.normalWS.xyz, unpacked.tangentWS.xyz);
```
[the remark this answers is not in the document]
ellioman: This is basically copy/paste of what ShaderGraph uses for normals. From what I've seen it doesn't, if it does then we'll need to address that in a separate PR.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
```diff
@@ -279,19 +292,27 @@ public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderin
...
 
+                // Set up the descriptors
+                RenderTextureDescriptor descriptor = cameraTargetDescriptor;
+                descriptor.msaaSamples = 1;
+                descriptor.depthBufferBits = 0;
+
+                m_AOPassDescriptor = descriptor;
+                m_AOPassDescriptor.width /= downsampleDivider;
+                m_AOPassDescriptor.height /= downsampleDivider;
+                m_AOPassDescriptor.colorFormat = RenderTextureFormat.ARGB32;
```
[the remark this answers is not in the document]
ellioman: Because they output AO + Normals.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
```diff
@@ -279,19 +292,27 @@ public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderin
...
+
+                m_BlurPassesDescriptor = descriptor;
+                m_BlurPassesDescriptor.colorFormat = RenderTextureFormat.ARGB32;
+
+                m_FinalDescriptor = descriptor;
+                m_FinalDescriptor.colorFormat = SystemInfo.SupportsRenderTextureFormat(RenderTextureFormat.R8) ? RenderTextureFormat.R8 : RenderTextureFormat.ARGB32;
```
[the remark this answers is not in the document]
ellioman: I will need to do some investigation on this, which will fit well with the follow-up PR. 
If R5 is enough precision then we could go R8 -> R5G6B5 -> RGBA32 depending on whether the platform supports it or not.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -15,15 +15,61 @@ SAMPLER(sampler_BaseMap);
...
+
 float4 _SourceSize;
 float4 _ProjectionParams2;
-float4x4 _CameraViewProjections[2]; // This is different from UNITY_MATRIX_VP (platform-agnostic projection matrix is used). Handle both non-XR and XR modes.
-float4 _CameraViewTopLeftCorner[2]; // TODO: check if we can use half type
 float4 _CameraViewXExtent[2];
 float4 _CameraViewYExtent[2];
 float4 _CameraViewZExtent[2];
 
+static half RandomUV[40] =
```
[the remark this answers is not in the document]
ellioman: I've added a comment explaining how they were calculated. 
As this is only used by SSAO and tightly bound to the SSAO random values, I'd like to keep it in this file. For now at least.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -45,81 +91,82 @@ float4 _CameraViewZExtent[2];
 // Constants
 // kContrast determines the contrast of occlusion. This allows users to control over/under
 // occlusion. At the moment, this is not exposed to the editor because it's rarely useful.
-static const float kContrast = 0.6;
+static const half kContrast = half(0.6);
```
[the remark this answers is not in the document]
ellioman: Good point. I've added a comment about the range being from 0..1.

com.unity.render-pipelines.universal/Shaders/ComplexLit.shader
lukaschod: Sadly not, it requires viewDirectionTS. Unless we drop it and calculate in fragment, I do support that code path too

lukaschod: #if (defined(_NORMALMAP) || (defined(_PARALLAXMAP) && !defined(REQUIRES_TANGENT_SPACE_VIEW_DIR_INTERPOLATOR))) || defined(_DETAIL)
#define REQUIRES_WORLD_SPACE_TANGENT_INTERPOLATOR
#endif

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainDetailLit.shader
ellioman: I'll find someone from the Terrain team to look at this for me and confirm that for me.

com.unity.render-pipelines.universal/ShaderLibrary/DeclareNormalsTexture.hlsl
```diff
@@ -23,8 +23,8 @@ float3 LoadSceneNormals(uint2 uv)
     float3 normal = LOAD_TEXTURE2D_X(_CameraNormalsTexture, uv).xyz;
 
     #if defined(_GBUFFER_NORMALS_OCT)
-    half2 remappedOctNormalWS = Unpack888ToFloat2(normal); // values between [ 0,  1]
-    half2 octNormalWS = remappedOctNormalWS.xy * 2.0h - 1.0h;    // values between [-1, +1]
+    float2 remappedOctNormalWS = Unpack888ToFloat2(normal); // values between [ 0,  1]
```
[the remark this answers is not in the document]
ellioman: Since all of the code there is done in float I don't think I need to add a comment about half.
These lines were causing a precision loss warning in our shader so I removed the half variables.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/DepthNormalsOnlyPass.hlsl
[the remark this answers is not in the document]
eh-unity: I would do a quick double check on to see the quality. Maybe output the normals with both settings and do diff in (in code or image editor) to see if there's any difference.

com.unity.render-pipelines.universal/ShaderLibrary/DeclareNormalsTexture.hlsl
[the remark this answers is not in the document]
eh-unity: Ah yes, I missed the return type. It's all good then.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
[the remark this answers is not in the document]
eh-unity: I suppose for the reconstructing normal SSAO needs to output them for the bilateral blur.
Do we really need to output the normals always? 
Why do the blur passes output normal?
Could we reduce it's format size? reuse normals from AO pass? 
If we do that then reading the original AO is waste for all but the first blur (we're reading the blurred AO) maybe separate them into different targets on some platforms.

Shouldn't we just read the normals if they're present (we sample it in AO shader)
Also if some other effect needs normals in the future, shouldn't we just pass them in and read?

Do they need to be good quality, could we get away with less? Reduced format size? Maybe R8G8?
We are going to blur the AO anyway and possibly do half or quarter res AO. Normal quality might not matter in the end.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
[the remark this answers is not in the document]
eh-unity: Maybe call them SSAORandomUV then.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -45,81 +91,82 @@ float4 _CameraViewZExtent[2];
...
-    float sn, cs;
+    half sn, cs;
     sincos(theta, sn, cs);
-    return float2(cs, sn);
+    return half2(cs, sn);
 }
 
 // Pseudo random number generator with 2D coordinates
-float UVRandom(float u, float v)
+half UVRandom(float u, int sampleIndex)
```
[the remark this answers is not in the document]
eh-unity: This seems SSAO specific, maybe call it SSAO_UVRandom, so it's clear.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -171,12 +222,12 @@ float3 ReconstructViewPos(float2 uv, float depth)
 // High:   5 taps on each direction: | z | x | * | y | w |
 // https://atyuwen.github.io/posts/normal-reconstruction/
 // https://wickedengine.net/2019/09/22/improved-normal-reconstruction-from-depth/
-float3 ReconstructNormal(float2 uv, float depth, float3 vpos)
```
eh-unity, line 174: This seems generic, maybe put it in Universal utility functions.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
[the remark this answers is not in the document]
ellioman: Makes sense. Will do.

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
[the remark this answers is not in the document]
ellioman: Renamed it to GetRandomUVForSSAO :)

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
ellioman: That's the plan. It will be done by @lukaschod as he's doing similar things with the Decal work.
He's also going to look if some improvements can be made.

kaychang-unity (approved):

com.unity.render-pipelines.universal/ShaderLibrary/SSAO.hlsl
```diff
@@ -33,93 +84,94 @@ float4 _CameraViewZExtent[2];
...
+    const float2 positionSS = GetScreenSpacePosition(uv);
+    const half gn = half(InterleavedGradientNoise(positionSS, sampleIndex));
+
+    const half u = frac(UVRandom(half(0.0), sampleIndex) + gn) * half(2.0) - half(1.0);
+    const half theta = (UVRandom(half(1.0), sampleIndex) + gn) * half(TWO_PI);
+
+    return half3(CosSin(theta) * sqrt(half(1.0) - u * u), u);
 }
 
 float RawToLinearDepth(float rawDepth)
```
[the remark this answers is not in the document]
ellioman: Nice! 
That function doesn't handle orthographic cameras unfortunately.
But to improve the code I've removed the RawToLinearDepth() function and call yours for Projection Cameras.
This also exposed a fault in the code where the _ORTHOGRAPHIC keyword was flipped.

eh-unity (approved):

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainDetailLit.shader
lumonix: It might make more sense if the consuming code (e.g. in the normalization code) handles the zero length normal since we don't really know here what is coming in or how it gets used in URP.

Alternatively here we can do a length-squared check to see if input.NormalWS is zero length, but that cost only makes sense if we think the consuming code won't already handle that condition.

erikabar (approved):
Android performance is similar or better with this PR. https://docs.google.com/spreadsheets/d/1BwQxOJXhzk4X6GeSvPqUlbuzKSFYtH_N_oIpxKsvtbg/edit#gid=1452499847&range=A1:D1

iOS will be tested by mobile team. 

no regressions found in win/macOS

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitDepthNormalsPass.hlsl
```diff
@@ -0,0 +1,117 @@
...
+            nrm += splatControl.a * UnpackNormalScale(SAMPLE_TEXTURE2D(_Normal3, sampler_Normal0, IN.uvSplat23.zw), _NormalScale3);
+
+            // avoid risk of NaN when normalizing.
+            #if HAS_HALF
+                nrm.z += 0.01h;
+            #else
+                nrm.z += 1e-5f;
+            #endif
+
+            normalTS = normalize(nrm.xyz);
```
broepstorff, line 95: Can we extract the normal map mixing here to a function, and make SplatmapMix() call the same (shared) function? Not sure if anywhere else also does the same mixing logic....

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitDepthNormalsPass.hlsl
```diff
@@ -0,0 +1,117 @@
...
+    #if defined(_NORMALMAP) && !defined(ENABLE_TERRAIN_PERPIXEL_NORMAL)
+        half3 normalWS = TransformTangentToWorld(normalTS, half3x3(-IN.tangent.xyz, IN.bitangent.xyz, IN.normal.xyz));
+    #elif defined(ENABLE_TERRAIN_PERPIXEL_NORMAL)
+        half3 viewDirWS = IN.viewDir;
+        float2 sampleCoords = (IN.uvMainAndLM.xy / _TerrainHeightmapRecipSize.zw + 0.5f) * _TerrainHeightmapRecipSize.xy;
+        half3 normalWS = TransformObjectToWorldNormal(normalize(SAMPLE_TEXTURE2D(_TerrainNormalmapTexture, sampler_TerrainNormalmapTexture, sampleCoords).rgb * 2 - 1));
+        half3 tangentWS = cross(GetObjectToWorldMatrix()._13_23_33, normalWS);
+        half3 normalWS = TransformTangentToWorld(normalTS, half3x3(-tangentWS, cross(normalWS, tangentWS), normalWS));
+    #else
+        //half3 viewDirWS = IN.viewDir;
```
broepstorff, line 108: please remove this unused line

broepstorff (changes requested):
Looks good from the terrain side of things. Just a few code cleanup requests

-Brad (I'm a new Graphics engineer hire on the Terrain team)

aleksandrasdzikia (approved):
Looks good, no issues found. Ran graphics test projects from **UniversalGraphicsTest_Foundation**, **UniversalGraphicsTest_Terrain**, **UniversalGraphicsTest_Lighting** (Scenes from 200 to 205).  Also, ran **URP Examples** to check that there are no new issues. No visual or functionality issues. 

Also there's a slight improvement on iOS performance, tests done by @ernestasKupciunas : https://docs.google.com/spreadsheets/d/149ac28KGkTC3G-AtvwrPVs8v57B_eo0he8FcYT05HmA/edit#gid=0

Devices used: OnePlus Nord, Galaxy Note10 USA, Huawei P20, Huawei P40 lite, iPhone SE.

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainDetailLit.shader
```diff
@@ -183,6 +183,7 @@ Shader "Hidden/TerrainEngine/Details/UniversalPipeline/Vertexlit"
                 half4   Color           : TEXCOORD2; // Vertex Color
                 half4   LightingFog     : TEXCOORD3; // Vetex Lighting, Fog Factor
                 float4  ShadowCoords    : TEXCOORD4; // Shadow UVs
+                half3   NormalWS        : TEXCOORD5; // Shadow UVs
```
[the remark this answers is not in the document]
ellioman: Good catch. Fixed in next commit.

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitDepthNormalsPass.hlsl
ellioman: Done.

broepstorff (approved):

PR #2751, opened 2020-11-25
[HDRP] density volumes improvements

alelievr:
### Purpose of this PR

⚠️ Waiting for Volumetric performance tests to land (ETA: ~2 weeks)

- Allows the Density Volume component to take a RenderTexture / Custom Render Texture as volume mask
- Add colored volume masks
- Higher-resolution volume masks (up to 256 cube configured in the HDRP settings)
- Added a falloff mode for density volume blend distance (linear or exponential)

Also, improve the 3D Texture atlas to support different 3D texture resolutions and RGBA 3D textures and real-time GPU copy for 3D render textures.

New volumetric settings in the HDRP asset:
image

Falloff mode:
image
Note that the falloff mode doesn't affect the Fade Distance interpolation (which is always linear).

Also added a debug mode to visualize the density volume 3D atlas:
image

With selection support:
image

---
### Testing status

Modified the fog volume test with 3D textures of different resolutions and colors:
image

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fdensity-volumes-improvements/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/4610582/job/full-pipeline

### Note:
If you want to author texture 3D in unity I recommend this tool: https://github.com/alelievr/Mixture.
You can ask me for installation :)

PR #3191, opened 2021-01-20
[HDRP] fix nullref when chaging RP from HDRP to URP

alelievr:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1306291/

That was introduced by the density volume improvement PR #2751 

---
### Testing status
I tested switching from HDRP to URP in an empty project and the nullref doesn't appear anymore.


sebastienlagarde (approved):

LandonTownsendUnity (approved):
Looks like it fixes the issue. No risk because it's just adding a null reference check, so I'm approving

LandonTownsendUnity:
Also, talked with dev of PR and backports aren't needed because the regression is exclusive to a feature added to 2021.2. @sebastienlagarde if you have reason to believe that's not the case please add the tags back.

sebastienlagarde:
Thanks @LandonTownsendUnity . Independently of repro you found, the code is safer, so we will still backport it just in case.

PR #3195, opened 2021-01-20
Change Light Unit Slider Value Ranges

johnpars:
### Purpose of this PR
This is a small PR that nudges some light unit slider value ranges at the request of lighting artist. 

It also fixes an issue where the Outdoor HDRP Scene template had a light configured for **130K Lux**, and the slider's limit was **120K Lux**, thus indicating caution to the user. The value range was extended so that the caution is no longer displayed:

Before:
image

After:
image

Also, the Exposure Sunlit Scene value range was changed from **12-16 EV** to **12-15 EV**, and the preset changed to **13 EV**.

---
### Testing status
I locally created a new scene with the outdoor HDRP template and confirmed that the new 130K value range no longer registers caution to the user. I also locally confirmed the changes to the Exposure Sunlit value range.

---
### Comments to reviewers
~


sebastienlagarde (approved):

iM0ve (approved):

iM0ve (approved):
No further testing should be required

PR #3200, opened 2021-01-20
Fix XR depth copy

fabien-unity:
---
### Purpose of this PR
Make sure the depth buffer is properly copied for each views so that devices like Oculus or Varjo can make use of them for better reprojection or Mixed Reality usage.

---
### Testing status
Locally tested with HDRP template on Win10 (d3d11) with the Varjo Analytics debug tool as described in the bug (https://fogbugz.unity3d.com/f/cases/1286908/).

---
### Comments to reviewers
@JulienIgnace-Unity : can you double-check the depth copy code/shader for non XR usage ? Maybe add a QA member to this PR to run some more tests ?


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -446,7 +449,8 @@ void CopyXRDepth(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle depth
                             RTHandle depthRT = data.depthBuffer;
 
                             mpb.SetTexture(HDShaderIDs._InputDepth, data.depthBuffer);
-                            mpb.SetVector(HDShaderIDs._BlitScaleBias, depthRT.rtHandleProperties.rtHandleScale / DynamicResolutionHandler.instance.GetCurrentScale());
+                            mpb.SetVector(HDShaderIDs._BlitScaleBias, new Vector4(data.dynamicResolutionScale, data.dynamicResolutionScale, 0.0f, 0.0f));
```
JulienIgnace-Unity, line 452: Does this work if you start playing with the window size? I'm not sure the dynamicResolutionScale includes the RTHandle scale so I would expect this break when resizing the window.

fabien-unity: I tested different scenarios including resizing editor and standalone windows, and all seems good!

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
@fabien-unity is there a way to produce a test for this so we don't break it again?

fabien-unity:
> @fabien-unity is there a way to produce a test for this so we don't break it again?

Yes but it's not trivial. I've added it to our roadmap!

ryanhy-unity:
Did a quick check with a new project (mock and Oculus) and didn't see any issues with this PR.  Seeing as Fabien has run through the bug scenario with Varyo and verified the fix as well I think this looks good from my point of view.

ryanhy-unity (approved):

PR #3202, opened 2021-01-20
Move Volume Gizmo color from it's own panel to colors panel

alex-vazquez:
### Purpose of this PR
Moving `Volume Gizmo Color` from Preferences->Colors->SRP to Preferences->Colors (under group scene)

volumecolorgizmo

https://fogbugz.unity3d.com/f/cases/1307495

A new method has been added to allow easly adding colors to this panel. See `RegisterSceneColor`

---
### Testing status
Testing that volumes gizmo color are updated properly on both pipelines.

---
### Comments to reviewers
Need to add documentation(I was unable to see the if there were references to this color setting on the docs)

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
+++ b/com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
@@ -44,10 +44,10 @@ namespace UnityEngine.Rendering
         }
 
         /// <summary>
-        /// Adds a <see cref="PrefColor"/> into the <b>Preferences > Colors</b> panel./>
+        /// Adds a <see cref="PrefColor"/> into the **Preferences > Colors** panel./>
         /// </summary>
-        /// <param name="name">The name the color has in the <b>Colors</b> panel. This is in the format of 'group/name'.</param>
-        /// <param name="defaultColor">The initial color to use for the new entry in the <b>Colors</b> panel. This is also the value Unity uses when it resets the colors to their defaults.</param>
+        /// <param name="name">The name the color has in the **Colors** panel. This is in the format of 'group/name'.</param>
+        /// <param name="defaultColor">The initial color to use for the new entry in the **Colors** panel. This is also the value Unity uses when it resets the colors to their defaults.</param>
         public static Func<Color> RegisterPreferenceColor(string name, Color defaultColor)
         {
             if (string.IsNullOrEmpty(name))


jenniferd-unity (commented):
I dont know enough about C# to be able to assess RegisterSceneColor. However, the PR was not explicit if this was tested under URP too. 
Also is there a test project for those kind of functionalities?

alex-vazquez:
@jenniferd-unity this works well on both URP & HDRP. I ignore if there is a test project for this. Maybe @RSlysz  or @martint-unity might know.

RSlysz:
Testing editor is hard. We have not full flexibility on this. I am not sur you can test that this while appears in the color panel. But you can test that the method/object you grab using reflection have not changed in signature with a unit test, which is a start.

martint-unity (approved):
LGTM

com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
```diff
@@ -10,80 +15,54 @@ namespace UnityEngine.Rendering
...
+            {
+                Debug.LogError("Unable to find property `Color` on object `UnityEditor.PrefColor`");
+                return () => defaultColor;
+            }
+
+            s_ColorPref.Push(colorPref);
+
+            MemberExpression colorProperty = Expression.Property(Expression.Constant(colorPref, prefColorType), colorInfo);
+
+            return Expression.Lambda<Func<Color>>(colorProperty).Compile();
```
RSlysz, line 65: It appears I forgot a case in the initial implementation:
If we load a package that use this while the Preference > Color is open, I am not sur it will redraw.

So make sure you test this. And if it not redraws, that means you must explicitly check if it is open and redraw it after each ColorPref insertion. Ask me if you don't see how to do this.

com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
```diff
@@ -10,80 +15,54 @@ namespace UnityEngine.Rendering
...
 
         static CoreRenderPipelinePreferences()
         {
-            Load();
+            GetColorPrefVolumeGizmoColor = RegisterSceneColor("Volume Gizmo", s_VolumeGizmoColorDefault);
         }
 
-        static void Load()
+        static readonly ConcurrentStack<object> s_ColorPref = new ConcurrentStack<object>();
+        internal static Func<Color> RegisterSceneColor(string name, Color defaultColor)
```
RSlysz, line 43: If this code is meant to be used in other SRP, we need a public access on this (example, how to use it for HDRP decals?) Otherelse, we will need to ammend this file each time or grabb this method with reflection in other SRP packages.

RSlysz (changes requested):
The ColorPref should not be reimplemented in each SRP. Here in case there is something only in one SRP, we need either to reimplement it inside ore amend the CoreRP file with information that is usefull only for one SRP. We should clarify this.

com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
```diff
@@ -10,80 +15,60 @@ namespace UnityEngine.Rendering
...
+            PropertyInfo colorInfo = prefColorType.GetProperty("Color");
+
+            var colorPref = Activator.CreateInstance(prefColorType, name, defaultColor.r, defaultColor.g, defaultColor.b, defaultColor.a);
+            s_ColorPref.Push(colorPref);
+
+            MemberExpression colorProperty = Expression.Property(Expression.Constant(colorPref, prefColorType), colorInfo);
+
+            // Make sure that the new preference color is being loaded into the Preference/Colors panel
+            MethodInfo loadMethod = prefColorType.GetMethod("Load");
+            loadMethod.Invoke(colorPref, null);
```
RSlysz, line 68: Nice fix

RSlysz (approved):

com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
```diff
@@ -10,80 +15,60 @@ namespace UnityEngine.Rendering
...
 
         static CoreRenderPipelinePreferences()
         {
-            Load();
+            GetColorPrefVolumeGizmoColor = RegisterPreferenceColor("Scene/Volume Gizmo", s_VolumeGizmoColorDefault);
         }
 
-        static void Load()
+        /// <summary>
+        /// This function provides a way of adding a <see cref="PrefColor"/> into the panel 'Preferences\Colors' />
```
JordanL8, line 47: Adds a <see cref="PrefColor"/> into the **Preferences > Colors** panel.

com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
```diff
@@ -10,80 +15,60 @@ namespace UnityEngine.Rendering
...
+        /// </summary>
+        /// <param name="name">The name that the color will have on the panel, in form of 'group/name'</param>
```
JordanL8, line 49: The name the color has in the **Colors** panel. This is in the format of 'group/name'.

com.unity.render-pipelines.core/Runtime/Utilities/CoreRenderPipelinePreferences.cs
```diff
@@ -10,80 +15,60 @@ namespace UnityEngine.Rendering
...
+        /// <param name="defaultColor">Used for first time, and to restore the default value</param>
```
JordanL8, line 50: The initial color to use for the new entry in the **Colors** panel. This is also the value Unity uses when it resets the colors to their defaults.

JordanL8 (changes requested):

JordanL8:
Package API docs use markdown formatting within the main XML tags so you don't need to use \<b> etc... Can you replace the HTML with the markdown equivalent? I think this is only bold for these changes so \<b> > \*\* but for more information on formatting for package API docs, see [here](https://confluence.unity3d.com/pages/viewpage.action?spaceKey=DOCS&title=API+docs+content+and+formatting#heading-Formatting)

JordanL8:
Can you ping me when you've done this? ^^

JordanL8 (approved):

RemyUnity (approved):
LGTM

PR #3206, opened 2021-01-20
Fixed a regression introduced when changing the clamping algorithm for RTR and RTGI (case 1307479).

anisunity:
https://fogbugz.unity3d.com/f/cases/1307479/

A previous PR  introduced a regression (extra unexposing) in RTGI quality. This reverts that issue.

Testing status:
Made sure it works with exposure != 0
Updated the concerned screenshots
Added a new test to cover the problem for the future.


remi-chapelain (approved):
Fixed ✔️ 
Took a quick look at the updated tests also ✔️

sebastienlagarde (approved):

PR #3208, opened 2021-01-21
Update Hair Shader Preset Documentation

johnpars:
### Purpose of this PR
Tiny documentation PR that corrects some of the language for the Hair Shader Preset properties.

The documentation makes reference to a non-existent secondary specular color property. In reality this secondary specular color is computed under the hood within in the graph, and not an exposed option. Since we cannot easily modify the graph at this point without the threat of breaking it for users, this PR simply tries to improve the explanation of some properties in the documentation for the shader preset. 

---
### Testing status
~

---
### Comments to reviewers
~

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/hair-shader.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/hair-shader.md
@@ -49,7 +49,7 @@ New Materials in HDRP use the [Lit shader](Lit-Shader.md) by default. To create 
 | **Smoothness UV Scale Transform** | Sets the tiling rate (xy) and offsets (zw) for the Smoothness Mask Map. |
 | **Smoothness Min**                | Set the minimum smoothness for this Material.                |
 | **Smoothness Max**                | Set the maximum smoothness for this Material.                |
-| **Specular Color**                | Set the representative color of the highlight which will be used to drive both the primary specular highlight color which is mainly monochrome and the secondary specular highlight color which is chromatic.|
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Documentation~/hair-shader.md
```diff
@@ -49,10 +49,10 @@ New Materials in HDRP use the [Lit shader](Lit-Shader.md) by default. To create
 | **Smoothness UV Scale Transform** | Sets the tiling rate (xy) and offsets (zw) for the Smoothness Mask Map. |
 | **Smoothness Min**                | Set the minimum smoothness for this Material.                |
 | **Smoothness Max**                | Set the maximum smoothness for this Material.                |
-| **Specular Color**                | Set the color of the primary specular highlight.             |
-| **Specular Multiplier**           | Modifies the **Specular Color** by this multiplier.          |
+| **Specular Color**                | Set the representative color of the highlight which will be used to drive both the primary specular highlight color which is mainly monochrome and the secondary specular highlight color which is chromatic.|
```
Vic-Cooper, line 52: Set the representative color of the highlight that Unity uses to drive both the primary specular highlight color, which is mainly monochrome, and the secondary specular highlight color, which is chromatic.

Vic-Cooper (changes requested):
Made one suggested change.

sebastienlagarde (approved):

sebastienlagarde:
Change done, merging.

PR #3213, opened 2021-01-21
[HDRP] Fix locale diffusion profile shader property value in ShaderGraph

alelievr:
### Purpose of this PR
Fix https://unity.slack.com/archives/CC68SH5SB/p1611213720003000

---
### Testing status
I tested that the diffusion profile still works correctly (expose a diffusion profile in ShaderGraph and link it to the diffusion profile slot of a translucent shader).


RemyUnity (approved):
Tested on my machine with a french windows (using , as decimal separator), and I can confirm that the fix is working.

sebastienlagarde (approved):

PR #3226, opened 2021-01-21
Cleanup of Render Graph code (WIP2)

JulienIgnace-Unity:
### Purpose of this PR
This PR is a continuation of https://github.com/Unity-Technologies/Graphics/pull/3150
It focuses on HDRP debug passes and follows the same pattern as previous PR.
Notable information:
- Split the monolithic LightLoop Overlay pass into several passes
- Moved all debug code from several files to HDRenderpipeline.Debug.cs

---
### Testing status
Tested most of the debug overlays/rendering modes locally.
From my testing only DepthPyramid full screen mode did not work but it's the same on master. SingleShadowMap UseSelection mode seems to have some problems but it's similar on master.

---
### Comments to reviewers
@Unity-Technologies/gfx-qa-hdrp even though I tested manually it would be good to have another pass at testing debug modes. This PRs changes code related to all our debug overlays and some full screen debug mode. So testing Material, Lighting and Rendering debug panels would be good.


sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
```diff
@@ -6,15 +6,197 @@ namespace UnityEngine.Rendering.HighDefinition
...
+            set
+            {
+                if (value)
+                    m_DebugDisplaySettings.SetDebugLightingMode(DebugLightingMode.VisualizeCascade);
+                else
+                    m_DebugDisplaySettings.SetDebugLightingMode(DebugLightingMode.None);
+            }
+        }
+
+        static bool NeedColorPickerDebug(DebugDisplaySettings debugSettings)
```
FrancescoC-unity, line 86: Why is this static and the other not? Is it called from outside and not visible in this PR?

FrancescoC-unity (commented):
LGTM (though github diff in this kind of PR makes things a bit hard)

FrancescoC-unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
JulienIgnace-Unity: No reason in particular I guess. It was just copy pasted as is.
A case could be made that they should all be static since they are kind of "pure" functions anyway

TomasKiniulis (approved):
I tend to get consistent rendergraph errors when I set Tile/Cluster Debug to Tile and Material section Atrributes to Texccord2. Happens on ~~both PR and~~(fixed) master. Had Rendergraph issues occur with DepthPyramid, what was already mentioned in description as similar between master and pr too. So I think it's safe to have this merged.

https://user-images.githubusercontent.com/50582134/107016628-d58dbc80-67a6-11eb-83c1-b916c179a707.mp4

PR #3227, opened 2021-01-21
[HDRP] Update eye sample scene

remi-chapelain:
### Purpose of this PR
This PR fixes two issues in the eye material sample scene : 
- When going into play mode, they were two warnings about a misssing script (The referenced script on this Behaviour (Game Object 'SceneIDMap') is missing!). Scene has been edited manually to remove the phantom gameobject
- If you wanted to use a sample at a proper scale (25mm instead of 1 meter in the sample) you had some issue about lighting. Scene have been re-arranged and an explicative text have been added to ensure it's clear for users how to use them and get correct results. 

image

Updated text
image

---
### Testing status
- Tested in a fresh project to import material sample from the branch. 
- Clicked on play to see if warning were not still happening.

### Comments to reviewers
I'd like to be sure that the text makes sense and does not contain obvious mistakes

### DIFF
--- a/.yamato/_green_job_revisions_trunk.metafile
+++ b/.yamato/_green_job_revisions_trunk.metafile
@@ -81,140 +81,140 @@ all_vfx_urp_ci_-_trunk:
       updated_at: '2020-12-09 23:30:57.365726'
       version: ''
 nightly_shadergraph_-_trunk:
-  updated_at: '2021-01-21 07:31:03.270537'
+  updated_at: '2021-01-22 07:33:04.418113'
   last_green_revisions:
     android:
-      revision: 453023b6d02d56c12563da02011e59faad0a1cca
-      updated_at: '2021-01-21 00:25:51.925686'
+      revision: 45138d0401c56bc9aa6dce7b52f0b51dc53ebace
+      updated_at: '2021-01-22 01:53:46.036233'
       version: ''
     ios:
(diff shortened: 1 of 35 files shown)


iM0ve (approved):
Performed tests should be enough for the scope of the change

com.unity.render-pipelines.high-definition/Samples~/MaterialSamples/Eye.unity
```diff
@@ -123,6 +123,184 @@ NavMeshSettings:
...
+  m_text: 'The diffusion profiles in this sample are calibrated for one meter eye
+    balls.
+
+    To use them in a real scale scenario, roughly 25mm eye balls, you
+    need to adjust the world scale in the diffusion profiles accordingly.
+
+
+    In
+    addition, Directional lights will have difficulty to light such small objects,
+    instead prefer punctual lights to have correct results.'
```
Vic-Cooper, line 173: Suggested change: 
Unity calculates the diffusion profiles in this sample for eyeballs that have a diameter of one meter.

In reality, eyeballs have a diameter of roughly 25mm. To use the diffusion profiles at a realistic scale you need to adjust the world scale in the diffusion profiles accordingly. To create lighting that works with such small objects, use punctual lights instead of directional lights.

Vic-Cooper (approved):
Great! :)

sebastienlagarde (approved):

PR #3231, opened 2021-01-21
Only base camera in a camera stack can use deferred.

kaychang-unity:
### Purpose of this PR
As discussed in the past, only base camera type should allow deferred rendering mode.
Overlay camera types are now always rendered in forward rendering mode.
UI has been updated to warn the user when an overlay camera is using deferred rendering mode.

---
### Testing status
Ran UniversalGraphicsTest project.

---
### Comments to reviewers
The update UI looks like this when there is an invalid camera in the stack:

image


kaychang-unity:
ABV is green (same as master branch).

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -258,6 +258,15 @@ static GUIContent TempContent(string text, string tooltip, Texture i)
 
         GUIContent m_NameContent = new GUIContent();
 
+        bool IsNonBaseCameraDeferred(UniversalAdditionalCameraData additionalCameraData)
+        {
+            if (additionalCameraData.renderType == CameraRenderType.Base)
+                return false;
+
+            ScriptableRenderer renderer = additionalCameraData.scriptableRenderer;
+            return renderer is ForwardRenderer && ((ForwardRenderer)renderer).renderingMode == RenderingMode.Deferred;
```
phi-lira, line 267: Suggestion to make code more readable use `as` operator:
`ForwardRenderer renderer = additionalCameraData.scriptableRenderer as ForwardRenderer;`
`return (renderer != null) ? renderer.renderingMode == RenderingMode.Deferred : false;`

phi-lira (approved):

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Added support for the PlayStation 5 platform.
 
+### Changed
```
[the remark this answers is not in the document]
kaychang-unity: Ok.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
kaychang-unity: Done.

martint-unity:
Default warning.
image

Changed warning.
image

We can not change this warning message. The warning message is for cameras that can not be used in a stack. Like you can not have a Base camera on the stack. 
This change make no sense when not using Deferred. SO this is a no go. 
The image above makes no sense for a user that is using forward.

martint-unity (changes requested):
The warning makes no sense when in forward.

kaychang-unity:
@martint-unity I updated the UI as requested.

martint-unity (approved):

PR #3237, opened 2021-01-22
Change the tooltip for color shadows and semi-transparent shadows (case 1307704).

anisunity:
https://fogbugz.unity3d.com/f/cases/1307704/

Testing status
Checked both texts in the editor.


remi-chapelain (approved):
Looks good ✔️

PR #3244, opened 2021-01-22
Hd/no rt tests v3

RemyUnity:
---
### Purpose of this PR
As the graphic tests were done by capturing the rendering through a render texture, is was impossible to capture accumulation effects.
This PR enables backbuffer capture for the HDRP graphic tests framework, particularly usefull for DXR denoising and pathtracing.
It also adds a "wait for frame count multiple" setting in the test settings to enforce a sync point for capturing.
The references images were updated following those changes.

---
### Testing status
Purely automated tests.

---
### Comments to reviewers
/

### DIFF
--- a/com.unity.testing.hdrp/TestRunner/HDRP_GraphicTestRunner.cs
+++ b/com.unity.testing.hdrp/TestRunner/HDRP_GraphicTestRunner.cs
@@ -39,7 +39,9 @@ public class HDRP_GraphicTestRunner
         // Grab the HDCamera
         HDCamera hdCamera = HDCamera.GetOrCreate(camera);
 
-        if (settings.ImageComparisonSettings.UseBackBuffer)
+        bool useBackBuffer = settings.ImageComparisonSettings.UseBackBuffer;
+
+        if (useBackBuffer)
             GameViewUtils.SetGameViewSize(settings.ImageComparisonSettings.TargetWidth, settings.ImageComparisonSettings.TargetHeight);
 
         Time.captureFramerate = settings.captureFramerate;
@@ -67,7 +69,7 @@ public class HDRP_GraphicTestRunner
             settings.doBeforeTest.Invoke();
 
             // Wait again one frame, to be sure.
-            yield return new WaitForEndOfFrame();
+            yield return WaitFunction( useBackBuffer );
         }
 
         hdCamera.Reset();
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
```diff
@@ -518,12 +518,7 @@ static internal bool ValidRayTracingHistory(HDCamera hdCamera)
 
         internal int RayTracingFrameIndex(HDCamera hdCamera)
         {
-        #if UNITY_HDRP_DXR_TESTS_DEFINE
-            if (Application.isPlaying)
-                return 0;
-            else
-        #endif
-            return hdCamera.IsTAAEnabled() ? hdCamera.taaFrameIndex : (int)m_FrameCount % 8;
+            return (int)hdCamera.cameraFrameCount % 8;
```
anisunity, line 521: you'll need to merge master into your branch and take that version

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingReflection.cs
```diff
@@ -144,6 +144,14 @@ static void RTReflectionDirectionGeneration(CommandBuffer cmd, RTReflectionDirGe
 
         #region Deferred Lighting
 
+        RTHandle RequestReflectionHistoryBuffer(CommandBuffer cmd, HDCamera hdCamera)
```
anisunity, line 147: Not sure why this is in this PR

anisunity (changes requested):

anisunity:
A change required and a question. Otherwise i think it looks good

com.unity.testing.hdrp/TestRunner/HDRP_GraphicTestRunner.cs
```diff
@@ -62,8 +73,27 @@ public IEnumerator Run(GraphicsTestCase testCase)
...
+            else
+            {
+                // Given that we will render two frames, we need to compensate for them in the waiting
+                // After this line, the next frame will be frame 0.
+                while (((hdCamera.cameraFrameCount + 2) % (uint)settings.frameCountMultiple) != 0) yield return new WaitForEndOfFrame();
+            }
+        }
+
+        // Force clear all the history buffers
+        hdCamera.RequestClearHistoryBuffers();
```
JulienIgnace-Unity, line 96: I'm not sure this is the right way to do that.
With this code, we clear all history buffers to black regardless of what the correct default value is. It makes things more consistent but probably wrong in many cases.
A lot of the render loop code relying on history buffers actually checks for first frame or existence of the history buffer to make special cases for when history is not correct. With this way of doing things, this won't work.
I suggest that instead of clearing to black we just dealloc all history buffers. This should mimic better what happens when we actually creates a new camera. Pinging @anisunity as I think it was you who did this?

anisunity: Indeed for all ray tracing buffers black is the default value so it works, but if there is other buffers that require specific values it won't work. deallocating would work yes.

sebastienlagarde:
replace by https://github.com/Unity-Technologies/Graphics/pull/5067

PR #3254, opened 2021-01-25
[Yamato] Exclude docs in expression triggers

sophiaaar:
### Purpose of this PR
* Excluded docs changes from per-project expression triggers
* Updated the trigger on the package jobs (I don't think this worked before)

---
### Testing status
* Changed some md files and pushed to confirm that no jobs started
* Updated the package job trigger and pushed, confirmed package jobs started

---
### Comments to reviewers
Un-drafting early so I can test the trigger


theo-pnv (approved):
Can you add the Documentation exclusion for the core package too? https://github.com/Unity-Technologies/Graphics/blob/yamato/docs-trigger/.yamato/_abv.yml#L26

PR #3273, opened 2021-01-27
Universal bugfix multiediting fix lights

martint-unity:
Purpose of this PR
Fixes the multi editing issue in this fb case
https://fogbugz.unity3d.com/f/cases/1289620/

Testing status
Created several cameras and tested this property locally.


martint-unity:
Still have the issues @simon-engelbrecht-soerensen showed here
https://github.cds.internal.unity3d.com/unity/Graphics/pull/2992
the ui doesnt update properly... will work on that.

Verasl:
Are these changes supposed to work between light types? like having a spot and a point light selected at the same time? In this case, currently the bias settings are not even visible.

martint-unity:
Still have the issues @simon-engelbrecht-soerensen showed here
https://github.cds.internal.unity3d.com/unity/Graphics/pull/2992
the ui doesnt update properly... will work on that.
I will not do that in this PR. We should make another bug-report for that. Since it looks like it is more involved than I initialy thought.
@Verasl I Don't think they should work between light types.

Verasl (approved):
LGTM

simon-engelbrecht-soerensen (approved):
Works as expected!

simon-engelbrecht-soerensen (approved):
lgtm

phi-lira:

PR #3275, opened 2021-01-27
Disabling GI denoiser in tests

lukaschod:
### Purpose of this PR
Disabling GI denoised in tests to reduce halo effect in universal graphics tests.

---
### Testing status
- [x] Added test that ensures no new lighting setting will be created with denoiser.


ellioman (approved):
Great. Nice one adding the test! 👍

phi-lira (approved):

PR #3276, opened 2021-01-27
[HDRP] performance tests improvements

alelievr:
### Purpose of this PR
- Fix HDRP static analysis tests
- Add new performance tests related to timings
- Add multiple resolutions for memory tests

Here are the performance tests that have been implemented:
For CPU / GPU and memory:
- Materials
- Lighting
- Effect (post process, custom pass, compositor)
- Other

Note that all the tests above are executed with both Forward and Deferred render settings.


theo-pnv (approved):
Approving the yml/yamato part

sebastienlagarde (approved):

PR #3295, opened 2021-01-29
Fix shaders incompatible with our DXC MSL and Vulkan backends

Dredhog:

### Checklist for PR maker

---
### Purpose of this PR
To make SRP shaders compatible with our DirectX shader compiler (DXC) backend (and the MSL and Vk backends in particular as they impose additional limitations on shaders beside the ones satisfied for DX12 with earlier PRs of this nature)

DXC limitations (for all API targets):
*  Inability to use `writes` to constant buffers  (this is a limitation of DXC in general when targeting modern HLSL which is required to use certain new features), our particle shaders heavily rely on this

DXC Metal backend limitations:
* Lack of Tesselation support (not implemented on our end)
* Lack of `VFACE` support - only `SV_IsFrontFace` as it translates to MSL's `front_facing`

DXC Vk/Metal limitations
* Inability to use min() inside array length declarations (dxc bug https://github.com/microsoft/DirectXShaderCompiler/issues/3355)

This PR does not yet fully fix the `_ABV for SRP` for a custom editor revision where DXC is the default compiler as:
* many tests rely on DX11 to build projects (DXC shaders don't work there)
* There still appear to be some legitimate issues DXC from running all test suites (caused by issues in our DXC impl. and not in shaders)
* Also, some test images vary very little (the diff image is black from what I cant tell) possibly due to some codegen/precision issues.
* SubsurfaceScattering actually works better with DXC for MSL (looks like DX11/Vk). My guess this is due to identical memory layouts between APIs when using DXC (all use DX layout)

---
### Testing status
* Ran [_ABV for SRP repository (all PR jobs) - CUSTOM-REVISION (without DXC)](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/dxc-msl-vk-compatibility/.yamato%252F_abv.yml%2523all_project_ci_CUSTOM-REVISION/5084739/job/pipeline) 
with almost the same failures as were on master [at the time of branching off ](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252F_abv.yml%2523all_project_ci_trunk/4696415/job/pipeline)
* The URP tests 035_Shader_TerrainShaders.png is failing on more platforms with this PR, but I'm not yet sure if that's because it might be unstable as it used to fail before)
---

### DIFF
--- a/com.unity.render-pipelines.universal/Shaders/XR/XRMirrorView.hlsl
+++ b/com.unity.render-pipelines.universal/Shaders/XR/XRMirrorView.hlsl
@@ -30,7 +30,7 @@ Varyings VertQuad(Attributes input)
     output.positionCS.xy = output.positionCS.xy * float2(2.0f, -2.0f) + float2(-1.0f, 1.0f); //convert to -1..1
 
     //Using temporary as writing to global _ScaleBias.w is prohibited when compiling with DXC
-    float scaleBiasW = _ScaleBias.w; 
+    float scaleBiasW = _ScaleBias.w;
 #if UNITY_UV_STARTS_AT_TOP
     // Unity viewport convention is bottom left as origin. Adjust Scalebias to read the correct region.
     scaleBiasW = 1 - _ScaleBias.w - _ScaleBias.y;


com.unity.render-pipelines.high-definition/Runtime/Debug/DebugViewTiles.shader
```diff
@@ -50,7 +50,7 @@ Shader "Hidden/HDRP/DebugViewTiles"
             int _ClusterDebugMode;
 
             StructuredBuffer<uint> g_TileList;
-            Buffer<uint> g_DispatchIndirectBuffer;
+            StructuredBuffer<uint> g_DispatchIndirectBuffer; // Has to be a StructuredBuffer (not Buffer) to work with DXC https://github.com/KhronosGroup/SPIRV-Cross/issues/1362
```
sebastienlagarde, line 53: My memory is fuzzy, but I guess there was a reasons to use Buffer instead of StructuredBuffer all over the places. Guess StructuredBuffer was not working on some platforms or in some context. Anyone remember? @mmikk @jessebarker @FrancescoC-unity @EvgeniiG

sebastienlagarde: (This comment is for the whole PR, not this particular line)

aras-p: @sebastienlagarde I see that a bunch of these `Indirect arguments have to be in a _buffer_, not a structured buffer` were added by Rune 4 years ago. But why they have to be in buffer? Maybe they had to be in buffer back then, but since then some code got tweaked/fixed and they no longer have to?

aras-p (approved):

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugViewTiles.shader
jessebarker: I don't recall any specific issues with Buffer vs StructuredBuffer and their translations to MSL or GLSL/SPIR-V.  With Metal, we've typically had more issues based on data type or pixel format rather than how they're organized (apart from issues around texture atomics).  The failures I'm seeing on Yamato don't seem related to shaders and there are a lot of passing suites for Vulkan and macOS, so I'm guessing this is ok.

EvgeniiG: Structured buffers that are not 128-bit aligned have terrible performance on Nvidia. Not sure about AMD.

Dredhog: This source change is only needed to generate correct MSL.

Not sure if the change is harmful for Metal with the old compiler though. When compiling this to MSL with the old compiler backend (FXC+HLSLcc) the output changes from:
`const device uint *TestRWBuffer [[ buffer(0) ]]`
to
```
struct TestRWBuffer_Type {    uint value[1]; };
const device TestRWBuffer_Type *TestRWBuffer [[ buffer(0) ]] // << in the entry point argument list
```
so my guess would be that this change doesn't have perf implication on Metal as this is still a pointer to an array of uints in the end ( correct me if I'm wrong as I'm just guessing ).

Do you think we could maybe use macros for [RW]Buffer which would only be defined to [RW]StructuredBuffer on DXC+Metal (or maybe Metal with all compilers?) to avoid use of unaligned StructuredBuffer on other APIs?

@EvgeniiG What is our current minimal buffer alignment, do we often use buffers not aligned to 128B?

EvgeniiG: In the current shader file, the size of the structure is sizeof(unit), so it is not 128-bit aligned.

A bigger concern (which is possibly what Rune referred to) is that I don't believe there is a guarantee about the exact memory layout of structured buffers, so interpreting it as a regular array of float3s or float4s (which is what an indirect dispatch does) may potentially be unsafe.

_"Structured identifies the Resource as a single contiguous block of memory with no mipmaps, nor array slices, but it does have a structure size (stride), so that it represents an array of structures. Implementations can take advantage of knowing there is a fixed structure size in the way they lay out the memory physically (hidden from the application)."_

https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#StructuredBuffer

Dredhog: Sorry for the delay, had been attending to other PRs the past week.

From what I hear the main concern is with the change to StructuredBuffers are:
1) Using them for Indirect buffers for D3D APIs could _potentially_ not work on some hardware where SBs don't have an array of structures memory layout as that's something the d3d11 spec does not enforce
2) Structured buffers can be less performant on NVIDIA hardware than regular buffers

I could either:
1) Leave the code as is
2) Modify the changes I did to indirect buffers in particular by adding some macro like INDIRECT_BUFFER and RW_INDIRECT_BUFFER, which would only be  defined to StructuredBuffer and RWStructured for the combo of DXC+Metal (as the above 2 concerns are for D3D or NVIDIA, neither of which concern Metal)
4) Some other suggestion

So @sebastienlagarde, @EvgeniiG what's your verdict on what should be done?

Dredhog: Ok, so I randomly had an epiphany, that this replacement can be done with the shader compiler preprocessor with `-DRWBuffer=RWStructuredBuffer` `-DBuffer=StructuredBuffer` without needing to add ad-hoc macros or change shader source.

I undid the manual shader changes in f9697a37a717e65cb80cb27c0aaa78c3cea583d4.

jessebarker (approved):
LGTM once the tests pass.

Pauliusd01 (approved):
LGTM, I've used this SRP branch to test and compare DXC functionality with no obvious issues.

aras-p:
Ok yeah now without the Buffer changes the diff is tiny, go go go! 🟢

sebastienlagarde (approved):

sebastienlagarde:
update to lastet master + fixed formatting, ready to merge after Yamato success

PR #3304, opened 2021-01-29
Don't apply sRGB blit when rendering to a Camera.targetTexture

hkr:
Case 1307710 uses URP to render to RenderTexture (via Camera.targetTexture). On Android versions that don't support a sRGB backbuffer this can cause problems.

### Checklist for PR maker

---
### Purpose of this PR
Fix case 1307710

---
### Testing status
URP tests and manual test with the repro project

---
### Comments to reviewers


erikabar (approved):
dev already tested with repro project, approving

phi-lira (approved):

PR #3306, opened 2021-01-30
fix issue with shaderPass defines occurring in post graph instead of …

esmelusina:
### Purpose of this PR
Move ShaderPass definitions into pre graph, so that the SHADERPASS define works correctly for graph nodes and custom function nodes.

---
### Testing status
Manual testing; verifying SHADERPASS define works correctly in custom function node. We don't currently have graphics testing that covers this case specifically, but testing framework improvements will enable that.


RemyUnity:
As I initially reported the issue on slack I checked the fix. LGTM in urp, left is a sphere with default shader, and right is one with vertices pushed only in the shadowcaster pass.
image

It does weird stuff in HDRP, but that's maybe expected : 
image

martint-unity (approved):

LandonTownsendUnity (approved):

PR #3316, opened 2021-02-01
URP ShaderGraph mobile optimizations to add half support and lower bandwidth usage

manuele-bonanno:
### Purpose of this PR
These optimizations were discussed with Apple.

mobile optimizations to lower bandwidth:
completely removed viewDirectionWS variable declaration from StructField.Varyings struct and also the usage of the variable from the shader template as well.

For WorldSpaceViewDirection that is used in the ShaderGraph node, replaced the calculation from:

normalize(input.viewDirectionWS) 

to:

 normalize(GetWorldSpaceViewDir(input.positionWS)

Also changed the field dependency of the WorldSpaceViewDirection node to now requiring StructFields.Varyings.positionWS variable.

Improvements (running on iPhone SE - 1334x750 res, BoatAttack - benchmark static island scene, Low Quality):

Before:
GPU Time: 20.634 ms~
L2 cache read per frame: 114.31 mb~ / per frame
L2 cache write per frame: 37.33 mb~ / per frame
After:
GPU Time: 20.151 ms
L2 cache read per frame: 112.43 mb~ / per frame
L2 cache write per frame: 35.36 mb~ / per frame

---
### Testing status
local + iOS device + Yamato
all Shadergraph tests are green, the only URP tests failing are also failing on master so are not broken by this PR. Tests ran locally on an earlier trunk build were successful

---


manuele-bonanno:
the failed URP tests are also failing on master and don't seem related to this PR. Shadergraph tests are all green

manuele-bonanno:
closing for the time being while discussing some potential platform issues

PR #3317, opened 2021-02-01
Mention in TAA doc that a certain use case will lead to problems.

FrancescoC-unity:
Unfortunately the history buffers are allocated per camera. If we have multiple game views with different aspect ratios,  but using the same camera game object, the history will be stomped on. 

The solution would be factoring in the aspect ratio when requesting a new camera, however that will inevitably lead to undesirable results when rescaling the viewport in a normal use case.

Since the use case of having two game views using the _same_ camera is fairly fringe, we'd rather not tax the main path to accommodate this. To make this clear I added a line in the docs.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Anti-Aliasing.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Anti-Aliasing.md
@@ -45,6 +45,10 @@ To select TAA for a Camera:
 
 When using the same Camera GameObject for multiple Game Views TAA may not work as expected due to limitations of the history buffer system. Multiple game views using different Cameras will however work as expected.
 
+### Limitations
+In the Editor, if multiple Game views use the same Camera, TAA may not work as expected due to limitations of the history buffer system. However, if you use multiple Game views, where each Game view uses a unique Camera, TAA works as expected.
+
+
 <a name="SMAA"></a>
 
 ## Subpixel morphological anti-aliasing (SMAA)


JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Documentation~/Anti-Aliasing.md
```diff
@@ -43,6 +43,8 @@ To select TAA for a Camera:
 1. Select the Camera in the Scene view or Hierarchy and view it in the Inspector.
 2. In the General section, select Temporal Anti-aliasing (TAA) from the Anti-aliasing drop-down.
 
+When using the same Camera GameObject for multiple Game Views TAA may not work as expected due to limitations of the history buffer system. Multiple game views using different Cameras will however work as expected.
```
sebastienlagarde, line 46: maybe add this into a "limitatoin section" : cc @JordanL8 ?

JordanL8: Yeah, a limitations section per anti-aliasing effect (when necessary) sounds good. For this one:

\### Limitations
In the Editor, if multiple Game views use the same Camera, TAA may not work as expected due to limitations of the history buffer system. However, if you use multiple Game views, where each Game view uses a unique Camera, TAA works as expected.

JordanL8 (changes requested):

JordanL8 (approved):

PR #3164, opened 2021-01-18
updated lighting settings

martint-unity:
---
### Purpose of this PR
Fixes for lighting after changes landed in trunk. 
Now the lighting for these scenes are off and needs to be corrected

PR #3318, opened 2021-02-01
[10.x.x Backport] Updated lighting settings #3164

ellioman:
Backport of #3164


lukaschod (approved):
Checked if there is no accidently difference between original branch

ellioman:
The two Yamato issues are known and not related to this PR (056).

PR #3323, opened 2021-02-01
Fix issues with light layers issues when editing multiple lights

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1308284 

The issue seemed to be related to the fact that we were syncing rendering layers when multiple different light layers were selected for different lights, leading to essentially a random one between the selected to update for all the lights (wrongly).  This fixes the issue. 

EDIT: I tested the fogbugz repro, + I tried adding another light and try the same thing (so 3 lights).


sebastienlagarde (approved):

TomasKiniulis:
Hey @FrancescoC-unity! Could you give a short description of what was tested?

FrancescoC-unity:
> Hey @FrancescoC-unity! Could you give a short description of what was tested? Anything additional that should be checked besides validating the repro case?

Oops sorry yes, added.

TomasKiniulis (approved):
Thanks! Just in case, did a quick check in repro case switching RP assets one with light layers enabled other with disabled, toggling light layers on/off in asset and reselecting multiple lights. Also toggling more different light layers in multiple lights. No issues found!

PR #3324, opened 2021-02-01
Cleanup CameraCaptureBridge API Remove legacy connection to recorder

jeanblouin:
### Purpose of this PR
This PR remove the legacy connection to the Unity Recorder which is not required anymore, since the recorder now uses the available CameraCaptureBridge API.

The PR removes the old reflection code that is now obsolete.

---
### Testing status
- Run HDRP_Tests locally 
- Validated that the unity recorder works correctly with this change local
- Yamato tests

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Utilities/CameraCaptureBridge.cs
+++ b/com.unity.render-pipelines.core/Runtime/Utilities/CameraCaptureBridge.cs
@@ -1,4 +1,3 @@
-
 using System;
 using System.Collections.Generic;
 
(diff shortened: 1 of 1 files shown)


sebastienlagarde:
@jeanblouin does it need to be backported ?  on 2021.1 ? on 2020.2? thanks

jeanblouin:
> @jeanblouin does it need to be backported ? on 2021.1 ? on 2020.2? thanks

Yes, that would be a good idea.

sebastienlagarde (approved):

PR #3329, opened 2021-02-01
Hdrp/fix probe blend distance UI

kecho:
### Purpose of this PR
Blend distance of reflection probes was being bound by max. Instead, it should be bound by the min of the size.
This causes some annoyance when editing the blend distance of a reflection probe.

---
### Testing status
Tested manually by adding / removing a probe

Fogbugz: https://fogbugz.unity3d.com/f/cases/1306525/


FrancescoC-unity:
This looks fine to me, but I am maintainer of the system, so gonna add a person that usually deals with this kind of UI stuff alongside my LGTM 🟢

FrancescoC-unity (approved):

sebastienlagarde (approved):
You miss a changelog

RSlysz (approved):

iM0ve (approved):
Verified that the probe now updates the same way using the slider and inputting a number

https://user-images.githubusercontent.com/36502659/106742687-07c3e080-6626-11eb-99d8-ca98d360672c.mp4

PR #3330, opened 2021-02-01
Improving the screen space global illumination.

anisunity:
In this PR I tried to improve the screen space global illumination. The improvements need validation from @remi-chapelain. I also removed the depth thickness value from the UI.

image

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/HDRenderPipeline.ScreenSpaceGlobalIllumination.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/HDRenderPipeline.ScreenSpaceGlobalIllumination.cs
@@ -370,8 +370,8 @@ namespace UnityEngine.Rendering.HighDefinition
             public TextureHandle outputBuffer;
         }
 
-        TextureHandle ConvertSSGI(RenderGraph renderGraph, HDCamera hdCamera, bool halfResolution, 
-            TextureHandle depthPyramid, TextureHandle stencilBuffer, TextureHandle normalBuffer, 
+        TextureHandle ConvertSSGI(RenderGraph renderGraph, HDCamera hdCamera, bool halfResolution,
+            TextureHandle depthPyramid, TextureHandle stencilBuffer, TextureHandle normalBuffer,
             TextureHandle inputBuffer0, TextureHandle inputBuffer1)
         {
             using (var builder = renderGraph.AddRenderPass<ConvertSSGIPassData>("Convert SSGI", out var passData, ProfilingSampler.Get(HDProfileId.SSGIConvert)))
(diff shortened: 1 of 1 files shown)


remi-chapelain (commented):
Spent some time testing it in various project (HDRP Template, Raytraceing Scene, Living in light, HDRP Automated test) and while this is better in terms of intensity and less subtle than what we have, it introduces unwanted artefacts and lighting in dark/corner areas. 

This has been pair-reviewed with @anisunity and a new version with the possibility for the user to bypass the validity mask to blend the SSGI result will be added to accomodate if the user is using lightmaps or not. 

Here are some example (Brighter image are after the PR, dimmer are before)

5f3a7ff518f03a26f3859592416d54bd

78c28181633dff92befa828006f112b0

anisunity:
image

anisunity:
image

remi-chapelain (commented):
Testing still in progress. 

For now, I've taken a deep look at projects when there's no lightmaps. So far, we have a result that is clearly better in terms of intensity and more inline with what users will expect for the feature.

Still needs some testing in other projects with the blending of lightmaps, when some objects are moving and also want to take a quick look at performance because even though the filtering cost has been reduced, the quality settings have changed a lot. 

One concern for now is the upgrade of current projects; This might be an underlying issue of our volume system. Since the max clamped value for filter radius has increased, if you upgrade a project, you need to remove the override and re-add it for the max slider to be correct. 

https://user-images.githubusercontent.com/57442369/106936287-b1859900-671c-11eb-96c7-2f3d0610655c.mp4

Here's some static examples for now !

**Without lightmaps**
4e99a10f57e9803ca690dc55317c57e1 (1)

e89289d5cc82738917fc5c5f814b9909

**With lightmaps**
41685975ab633082bc6984503c9f1bf2

remi-chapelain (approved):
I updated my previous [comment](https://github.com/Unity-Technologies/Graphics/pull/3330#pullrequestreview-583680774), this is good on my side, already a big improvement over what we already have. 

I just have the concern about the upgrade of some projects since the clamped range is serialized in the volume asset, you need to remove/re-add or reset the override for the min/max values to be updated. The quality settings still works but the UI is stuck with the old values. 

Still approved and I'll report something about it since it's not related to this PR.

remi-chapelain (commented):
Made one last pass on it after the support of the light probe volume v2. Even though, it's super experimental at the moment, the two effects supports each other very well and give proper results. 

The only thing that I'd like to be modified here is documentation, since SSGI doesn't use a blending algorithm anymore and totally replaces lightmaps now, the same way RTGI does ,it can be interesting to understand that lightmaps won't help anymore.
 
Also could be great to add a bit about its intended usage/workflow to be sure that no ones expect only using SSGI will make magic and invent information it doesn't have. 

Here's some quick test with the probe volume v2 :

https://user-images.githubusercontent.com/57442369/107611678-d01bf080-6c44-11eb-8c42-cd335ff6d36c.mp4

https://user-images.githubusercontent.com/57442369/107611627-b5e21280-6c44-11eb-81c7-863efc7c6f52.mp4

sebastienlagarde (approved):

sebastienlagarde:
@anisunity PR have conflict with Julien PR that have been merged, can you merge master? thanks
Also the Formatting job don't pass the PR need a formatting pass

JordanL8 (approved):

PR #3331, opened 2021-02-01
[HDRP]Fixed parameters order on inspectors for Volume Components without custom editor

alex-vazquez:
### Purpose of this PR
This PR is fixing this
- https://fogbugz.unity3d.com/f/cases/1294683/
#### Before PR
image
#### After PR:
image
---
### Testing status
- Check that the `VolumeParameters` are being displayed in the same way as they are declared
---
### Comments to reviewers
This PR is simply backporting the `VolumeComponentInspector.cs` file from https://github.com/Unity-Technologies/Graphics/pull/2745

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
+++ b/com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
@@ -150,7 +150,7 @@ namespace UnityEditor.Rendering
                 .Where(
                     t => t.IsDefined(typeof(VolumeParameterDrawerAttribute), false)
                     && !t.IsAbstract
-                    );
+                );
 
             // Store them
             foreach (var type in types)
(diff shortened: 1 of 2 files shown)


sebastienlagarde (approved):

jenniferd-unity (approved):

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -202,37 +212,27 @@ public override int Compare((GUIContent displayName, int displayOrder, Serialize
...
+                        name = attr[0].name;
+                        order = attr[0].order;
+                    }
 
-                var parameter = new SerializedDataParameter(property);
-                m_Parameters.Add((new GUIContent(name), order, parameter));
-            }
-            m_Parameters.Sort(new ParameterSorter());
+                    var parameter = new SerializedDataParameter(t.Item2);
+                    return (new GUIContent(name), order, parameter);
```
RSlysz, line 232: If I get it well, here will be the field name right? Shouldn't it be localizable?

RSlysz (approved):
Pre approved but check localization.

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
alex-vazquez: yeah, that's true, I will create a another task to improve this. Thanks Rémi

TomasKiniulis (approved):
Checked fogbugz case custom volume, also created a new one and made sure that existing Volume overrides are not affected. Looks good! Order is correct now in all volumes

martint-unity (approved):

PR #3332, opened 2021-02-01
[HDRP] Fix error when increasing the maximum planar reflection limit

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1306530/

The global constant buffer is using a number of fixed sized arrays related to planar reflections, like this one:
`float4x4 _Env2DCaptureVP[32];`
The size of the array is controlled from the C# side with the _`ShaderVariablesGlobal.s_MaxEnv2DLight`_ constant.
(changing the constant in the code requires to re-generate the shader includes, and it is not accessible in user scripts)

In practice this imposes a hard limit on the planar reflections that HDRP can have on screen at once.

In the HDRP asset UI we now enforce this limitation regarding the array size and I have updated the documentation/tooltips regarding this limitation.

--- 
### Testing status
- The error now does not repro, but we also cannot set the max value for planar reflections higher than ShaderVariablesGlobal.s_MaxEnv2DLight.


sebastienlagarde (approved):

JulienIgnace-Unity (approved):

iM0ve (approved):
Testing should be sufficient for the scope of the PR

PR #3333, opened 2021-02-01
Fix uninitialized variable for shadow cascades

adrien-de-tocqueville:
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1273362/

---
### Testing status

tested with the repro project linked on the fogbugz page
Now using the frustum planes don't reduce the amount of culled objects

This PR: https://github.com/Unity-Technologies/Graphics/pull/1616
made the unintialized variable have a default value of 1 by not using the planes (unintended side effect which would effectively give a more efficient culling, unrelated to the fact the planes are present)

---
### Comments to reviewers

I am not sure of who to put in URP team to review that
Also, what should be the backports ?


sebastienlagarde:
Note: This PR must target master. However it include change from hd/bugfix, so it will need to wait to merge hd/bugfix before merging this PR

Wilfrid-Unity (approved):
Thanks Adrien, I checked that the shadows were culled as expected in the repro project.

For changes related to URP shadows, I suggest pinging @lukaschod, @simon-engelbrecht-soerensen

com.unity.render-pipelines.universal/Runtime/ShadowUtils.cs
```diff
@@ -57,6 +57,9 @@ public static bool ExtractDirectionalLightMatrix(ref CullingResults cullResults,
             shadowSliceData.resolution = shadowResolution;
             shadowSliceData.shadowTransform = GetShadowTransform(shadowSliceData.projectionMatrix, shadowSliceData.viewMatrix);
 
+            // This used to be fixed to .6f, but is now configureable.
```
eh-unity, line 60: Perhaps add to the comment that this is "culling sphere radius multiplier", so it's clear to the next reader. I had do a search to understand that as "**cascade blend culling** factor" seems like something "completely" different.

eh-unity (approved):
I assume hd/bugfix is now merged? Should target master?
I also suggested a minor comment improvement.
Otherwise looks good to me.

com.unity.render-pipelines.universal/Runtime/ShadowUtils.cs
adrien-de-tocqueville: sure. Also in the end i changed the value to 0.6 because I think i makes more sense, since it was the original value

sebastienlagarde (approved):

PR #3335, opened 2021-02-01
Hide shadow resolution value

adrien-de-tocqueville:
### Purpose of this PR

Goal is to have the same dropdown for shadow resolution bewteen cloud layer and vloumetric clouds following a request from qa

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Sky/CloudSystem/CloudLayer/CloudLayer.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Sky/CloudSystem/CloudLayer/CloudLayer.cs
@@ -51,16 +51,14 @@ namespace UnityEngine.Rendering.HighDefinition
     /// </summary>
     public enum CloudShadowsResolution
     {
-        /// <summary>Size 32</summary>
-        CloudShadowsResolution32 = 32,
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

remi-chapelain:
Still commenting even if this is already merged but there was a few more things we discussed to change in this overrides to try and be consistent with the volumetric clouds one : 
- [ ] - Use Low / Med / High for shadow resolution (done in this PR)
- [ ] Add information on the tooltip when enabling shadows that the shadow will override the dir light cookie (you can't have both shadows from two clouds systems) here's the current tooltip from volumetric clouds : "Enable/Disable the volumetric clouds shadow. This will override the cookie of your directional light and the cloud layer shadow (if active)."
- [ ] Change Scroll Direction to Scroll Orientation to be consistent with the now Wind Orientation in the override
- [ ] Finally, since both override can work together (top clouds sheet with cloud layer and medium and low level with volumetric clouds), should we just remove the dropdown from visual environment that implies they're both mutually exclusive ?

adrien-de-tocqueville:
>should we just remove the dropdown from visual environment that implies they're both mutually exclusive ?

It's still possible for users to make custom clouds with it. We should rathe make it explicit that volumetric clouds are not selected from there

PR #3340, opened 2021-02-02
Reset ambient probe upon switching to very different skies

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1306396/ (note it had nothing to do with reflection probes) 

The issue is that Sky system will use previous ambient probe when a new context is created; this is often correct when marginal changes happen, however if a radical change is made the previous probe might be largely incorrect. 

The issue was due to the fact that Volumetric used ambient probe, when the previous probe was used a super over-blown image when switching between afternoon and night will be produced until async readback is done. This will recover, however the many effects that rely on previous frame will now be "corrupted" with the overblown image and the final image won't be able to recover. 

This change introduces a heuristic to determine if previous sky settings was too different from current, if it was then we reset the ambient probe, which will be black but is recoverable.

What did I test: the repro case + changing intensity of skies by small values and notice that the ambient probe is not reset to black in those cases.


com.unity.render-pipelines.high-definition/Runtime/Sky/HDRISky/HDRISky.cs
```diff
@@ -169,6 +169,24 @@ public override int GetHashCode()
...
+        /// Determines if the SkySettings is very different from another. This is going to be used to determine whether
+        /// to reset completely the ambient probe instead of using previous one when waiting for current data upon changes.
+        /// In addition to the checks done with the base function, this HDRISky override checks whether the cubemap parameter
+        /// has changed if both settings are HDRISky.
+        /// </summary>
+        /// <param name="otherSettings">The settings to compare with.</param>
+        /// <returns>Whether the settings are deemed very different.</returns>
+        public virtual bool IsVeryDifferentFrom(SkySettings otherSettings)
+        {
+            if (otherSettings.GetSkyRendererType() != GetSkyRendererType())
```
JulienIgnace-Unity, line 182: Can't we put this part in the base SkySetting implementation?

FrancescoC-unity: And we should! I added it last minute and I thought I did put it there; moving now.

FrancescoC-unity: Also made a mistake there, I missed the override keyword.

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Checking with @iM0ve testing is sufficient. Approving.

com.unity.render-pipelines.high-definition/Runtime/Sky/HDRISky/HDRISky.cs
```diff
@@ -169,6 +169,21 @@ public override int GetHashCode()
...
 
+        /// <summary>
+        /// Determines if the SkySettings is very different from another. This is going to be used to determine whether
+        /// to reset completely the ambient probe instead of using previous one when waiting for current data upon changes.
+        /// In addition to the checks done with the base function, this HDRISky override checks whether the cubemap parameter
+        /// has changed if both settings are HDRISky.
+        /// </summary>
+        /// <param name="otherSettings">The settings to compare with.</param>
+        /// <returns>Whether the settings are deemed very different.</returns>
+        public override bool IsVeryDifferentFrom(SkySettings otherSettings)
```
sebastienlagarde, line 180: I am just worried that we introduce a public API name "IsVeryDifferentFrom". Summoning @JordanL8  in case he have an idea of a better name :) 

Should we have instead the notion of threshold. Like :   NeedRecomputeAmbientProbe(SkySettings otherSettings) or a name like that?  haveSettingsChange

JordanL8: Yeah I can't say that I'm a fan of IsVeryDifferentFrom. The summary says its just used to determine whether to reset the ambient probe, but can it be used for other things? If so, I'd say stick to the is different route but call it something like **DivergesFrom**.

FrancescoC-unity: The problem with NeedRecomputeAmbientProbe is that it makes no sense in the context of comparing skySettings without having the otherSkySetting being exactly the previous sky setting for the same sky. However that cannot be enforced at this level. 

What we are just saying with that function is that those are different, the summary mentions ambient probe because that's how we use it, but it is mentioned as a note for our use. 

I can put DivergesFrom, but that is not correct either as even with a small difference we have a divergence. This is true only if there is a big divergence. 

I am open to better names :D

JordanL8: Are you just open or are you very open? :p Although **Very** tells the users its more than a small difference, they'd still have to look at the docs to find out what Very actually means (like what threshold for what properties etc...). Maybe just ignore Very and mention in the summary what the sky settings need to be in order to be different. That way the information is in the most visible place because the summary shows up in the user's IDE when they type the method.

FrancescoC-unity: But not everyone reads documentation :D If I have .DivergesFrom(other) returning false for something that diverges less than a certain amount I'd be very confused and angry D:

JordanL8: True, I guess with very there they would have to look up the docs to figure out what it means, but without it, they could assume its just like one property and get angry when it isn't as you say. How about Significantly? SignificantlyDivergesFrom or IsSignificantlyDifferentFrom?

FrancescoC-unity: This I like! Will change

sebastienlagarde (approved):

PR #3341, opened 2021-02-02
Fix light frustum planes

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1304395/

issue was that the planes were extracted form the projection matrix, but after it has been transformed into GPU version, which gave wrong results for some planes

---
### Testing status
Tested the repro steps of the project
Yamato has some red tests but they are red on master or work locally (eg. 2005_Light_Color_Temp) so it seems fine


sebastienlagarde (approved):

TomasKiniulis (approved):
I think tested repro steps are sufficient. If nothing additional compared to master is failing on yamato this seems good to me.

PR #3343, opened 2021-02-02
The RTAO's history is now discarded if the occlusion caster was moving (case 1303418).

anisunity:
This PR solves partially the issue (the rest of the issue is solved by this PR) https://github.com/Unity-Technologies/Graphics/pull/3039 This discards the history of the intersected occluder is moving.
https://fogbugz.unity3d.com/f/cases/resolve/1303418/

Testing status
Tried the repro shared by the QA, doesn't repro anymore.
DXR tests are green locally.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Removed the material pass probe volumes evaluation mode.
 - Unifying the history validation pass so that it is only done once for the whole frame and not per effect.
 - Updated the tooltip for the Decal Angle Fade property (requires to enable Decal Layers in both HDRP asset and Frame settings) (case 1308048).
+- The RTAO's history is now discarded if the occlusion caster was moving (case 1303418).
 
 ## [11.0.0] - 2020-10-21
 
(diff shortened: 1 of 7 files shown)


sebastienlagarde:
@anisunity does it need to be backported?

sebastienlagarde (approved):

remi-chapelain (commented):
I have the same type of remarks than the RT [shadows threshold thing](https://github.com/Unity-Technologies/Graphics/pull/3039#pullrequestreview-566513981). 

6fa36c4609a8c90824fd7875f39ab25f

At **very** low velocity, nothing has changed.  ✔️ 
At high velocity, history is discarded and we don't get severe ghosting with RTAO. ✔️ 
At low velocity however, this is where you hit the velocity threshold and if you are right around it, it will flicker back and forth between discarding history and not and each time you can see a "jump". 
I'm not sure there's a real solution here apart from having another parameter like in the TAA that would be a speed rejection threshold of some sort. 
Just wanted to point that it can be a trade off in some specific situation.

**Comparison at high velocities**
47a6953071df1443d81127445a3a5730 (1)

**BEFORE THE PR**
ezgif-3-f8fbe14254b3

**AFTER THE PR** 
ezgif-3-f8310b70c1cc

**BEFORE THE PR**
7e45cf0335d16e3387bf47cf86bee310

**AFTER THE PR** 
ezgif-3-a212405fcf2b

remi-chapelain (approved):
Approving, since I believe this is still an improvement.

PR #3348, opened 2021-02-02
Fixed the display name of a Volume Parameter when is defined the attribute InspectorName

alex-vazquez:
---
### Purpose of this PR
The inspectors were shown like this:

image

When the user has defined them like this:

```
[Header("High noise")][Header("High noise")]    [InspectorName("Size and occurance"), Tooltip("Increase to make patches SMALLER, and frequent")] public ClampedFloatParameter _HighNoiseSpaceFreq = new ClampedFloatParameter(0.1f, 0.1f, 1f);
 [InspectorName("Wobble frequency"), Tooltip("Increase to make patches wobble more often")] public ClampedFloatParameter _HighNoiseTimeFreq = new ClampedFloatParameter(0.05f, 0f, 2f);
 [InspectorName("Pit depth"), Tooltip("Increase to make patches \"deeper\"")] public ClampedFloatParameter _HighNoiseAmplitude = new ClampedFloatParameter(0.4f,0f, 4f);
 [InspectorName("Wind strength"), Tooltip("Increase to make patches move faster")] public ClampedFloatParameter _HighNoiseMoveSpeedScale = new ClampedFloatParameter(0.2f, 0f, 10f);
```
So the attribute `InspectorName` was not properly readed to show the correct label for those properties.

After the changes in this PR, they appear correctly.

image

---
### Testing status

- Checking that the display name for Volume Parameters of Volume Components without a custom inspector are shown correctly.
- Added unit test.

---
### Comments to reviewers
It was missing an entry to check the `InspectorNameAttribute`, I've changed it to be a switch, and I extracted it to a method to be able to call it from a Unity test.

---
### Comments for QA
Grab the test project from [here](https://fogbugz.unity3d.com/f/cases/1294683/|https://fogbugz.unity3d.com/f/cases/1294683/)

and check the `TestVolumeComponent` on the inspector.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
+++ b/com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
@@ -2,10 +2,10 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using UnityEditor.AnimatedValues;
 using UnityEngine;
 using UnityEngine.Assertions;
 using UnityEngine.Rendering;
-using UnityEditor.AnimatedValues;
 
 namespace UnityEditor.Rendering
 {
@@ -350,7 +350,7 @@ namespace UnityEditor.Rendering
             // Display every field as-is
             foreach (var parameter in m_Parameters)
             {
-                if (parameter.displayName.text != "")
+                if (!string.IsNullOrEmpty(parameter.displayName.text))
                     PropertyField(parameter.param, parameter.displayName);
                 else
                     PropertyField(parameter.param);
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -457,36 +457,53 @@ protected void PropertyField(SerializedDataParameter property)
...
-        protected void PropertyField(SerializedDataParameter property, GUIContent title)
+        /// <param name="property">The property to obtain the attributes and handle the decorators</param>
+        /// <param name="title">A custom label and/or tooltip that might be updated by <see cref="TooltipAttribute"/> and/or by <see cref="InspectorNameAttribute"/></param>
+        void HandleDecorators(SerializedDataParameter property, GUIContent title)
         {
-            // Handle unity built-in decorators (Space, Header, Tooltip etc)
+            if (!property.attributes.Any())
+                return;
+
             foreach (var attr in property.attributes)
```
RSlysz, line 469: Not sure you gain anything here. If there is any attribute, you will exit quickly of the foreach. So there is no gain to check if this is empty right above.

Also, if the attribute property do some operation, now it is done two times instead of just once. As this is reflection API I would be extra careful with this.

RSlysz (approved):
Nice test :)

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
alex-vazquez: Removed, thanks :)

jenniferd-unity (approved):

martint-unity (approved):

iM0ve (approved):
**Whats tested:**
* Tested that InspectorName attribute is now used to display name, instead variable name previously.
* If InspectorName does not exist, the variable name is still used as fallback.
* Parameters are now correctly grouped using the Header attribute.
* Upgraded the project from Core to HDRP to see if there is any unexpected behaviour after update.
* For above testing the fogbuz sample was used, to verify that the users issue is gone

Screenshot 2021-02-04 at 12 13 39

PR #3350, opened 2021-02-02
Fix keywords with fbx importer

adrien-de-tocqueville:
### Purpose of this PR

fix https://fogbugz.unity3d.com/f/cases/1289869/

removed some logic from the fbx importer aimed at setting the keywords and passes, and added a call to ResetMaterialKeywords, which does the same.

Also added a call to ResetMaterialKeywords in OnPostProcessMaterial which then function called when the model uses import mode "Standard (legacy)", which apparently is the origin of the bug according to the fogbugz page

Finally added the support of converting "ReflectionFactor" to "_Metallic" property, which was not handled (from my test by exporting an fbx material using blender)

---
### Testing status

Tested with the car model in the repro project and custom models made in blender
I am not aware if there are automatic testing for the fbx importer but if there are we should launch them


sebastienlagarde (commented):
I know that the fogbugz is only about FBX, but currently I guess it is the same issue with Arnold, and 3DS Max physical Material, could you check? thanks

adrien-de-tocqueville:
The problem was related to the model import settings, so it's handled by the new OnPostprocessMaterial for any file format (not specific to fbx)
I guess we could still add the resetmaterialkeywords for other material types, but they are still added automatically when openning the inspector, and I don't have materials of those types to test

sebastienlagarde:
> they are still added automatically when openning the inspector

ok I see. So fine for me with current PR

sebastienlagarde (approved):

TomasKiniulis (approved):
I think checking custom models and repro case is enough. Don't see what else could be tested here. Approving. Although If you'd like additional test pass please let me know.

PR #3352, opened 2021-02-02
Fixed lightmaps not working properly with shader graphs in ray traced reflections (case 1305335).

anisunity:
There was an issue with attribute binding for lightmaps.
https://fogbugz.unity3d.com/f/cases/1305335/

Testing status
The issues doesn't happen anymore after the fix on the shared dataset.


anisunity:
Needs backport imo

alelievr (approved):

sebastienlagarde (approved):

remi-chapelain (approved):
Looking at the test case in the fogbugz, it seems sufficient enough to approve without double checking ! ✔️

PR #3354, opened 2021-02-02
Fix white flash on camera cuts with volumetric 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1270283/ 

The issue stems from the fact that the reprojection uses _previous_ frame exposure, so not only we need to wait for reprojection data to be valid but also the previous exposure needs to be valid. 

What did I test: The repro case both scrubbing and in play mode + I verified that reprojection still happens after the two frames we skip it (by checking in debugger :P )


TomasKiniulis (changes requested):
Hey @FrancescoC-unity! Please give a short description of what was tested and what additional testing is needed

FrancescoC-unity:
> Hey @FrancescoC-unity! Please give a short description of what was tested and what additional testing is needed

Done!

TomasKiniulis:
> > Hey @FrancescoC-unity! Please give a short description of what was tested and what additional testing is needed
> 
> Done!

Thanks!

TomasKiniulis (approved):
Checked with @iM0ve since he already looked into the issue. Testing status is sufficient. Approving!

PR #3355, opened 2021-02-02
Hdrp/fix tesselation wireframe

kecho:
### Purpose of this PR
Fixing bug were wireframe view does not match the material hdrp wireframe view.
World positions in SCENEPICKING are absolute. Because the _CameraViewMatrix is special, is pushed through HDRP only once, and can't be resubmitted on Unity native rendering, it must be aware of the SCENEPICKING pass.
When the scene picking pass is on, and if we indeed pushed a relative matrix, then we must apply a correction, since the world position coming from scenepicking is always absolute.

---
### Testing status
* Tested in demo level and repro steps of bug
d.

Fogbugz: https://fogbugz.unity3d.com/f/cases/1307745/


com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitTessellation.shader
```diff
@@ -410,9 +410,9 @@ Shader "HDRP/LitTessellation"
             #define SCENEPICKINGPASS
             #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl"
             #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl"
+            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/PickingSpaceTransforms.hlsl"
```
kecho, line 413: I had to move this include here, since the tesselation algorithms are located inside LitData.hlsl.

kecho:
Applied changes suggested by @sebastienlagarde 
Let me know if this works

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl
```diff
@@ -233,4 +233,22 @@ uint2 ComputeFadeMaskSeed(float3 V, uint2 positionSS)
     return fadeMaskSeed;
 }
 
+float4x4 _GetCameraViewProjMatrixRaw()
```
sebastienlagarde, line 236: Note, you must rely on testing SHADEROPTIONS_CAMERA_RELATIVE_RENDERING for absolute or not.
+ don't use "_" for function, we use it only (for legacy rasons) for constant buffer value.

So here is my suggestion.
remove the change in 
ShaderVariablesFunctions.hlsl

replace the code in PickingSpaceTransforms.hlsl by


// The scene picking pass is not under the control of SRP. The camera matrix is not camera relative
// when using _CameraViewProjMatrix from scene picking pass we must remove the translation to be camera relative
float4x4 ScenePickingGetCameraViewProjMatrix()
{
    float4x4 translationMatrix = {
            { 1.0 ,0.0 , 0.0, -_WorldSpaceCameraPos.x },
            { 0.0 ,1.0 , 0.0, -_WorldSpaceCameraPos.y },
            { 0.0 ,0.0 , 1.0, -_WorldSpaceCameraPos.z },
            { 0.0 ,0.0 , 0.0, 1.0} };

    return mul(_CameraViewProjMatrix, translationMatrix);
}

#if (SHADEROPTIONS_CAMERA_RELATIVE_RENDERING != 0)
#define _CameraViewProjMatrix ScenePickingGetCameraViewProjMatrix()
#endif

And thus we don't need to touch _CameraViewProjMatrix .  (note that _CameraViewProjMatrix  is used in an upcoming VFX PR, so with above suggestion it will still work. We have only issue with scene picking, so let's fix it only for it.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitTessellation.shader
sebastienlagarde: yes it make sense. 
So just to be coherent, we should move it in all shaders (axf.sahder, litTessellation.sahder, Lit.shader, LayeredLit.shader, LayeredLitTessellation.shader) and we also need to change it for shader graph.

this is located in HDSahderPasses.cs

here

            IncludeCollection GenerateIncludes()
            {
                var includes = new IncludeCollection();

                includes.Add(CoreIncludes.CorePregraph);
                includes.Add(CoreIncludes.kPassPlaceholder, IncludeLocation.Pregraph);
                includes.Add(CoreIncludes.CoreUtility);
                includes.Add(CoreIncludes.kShaderGraphFunctions, IncludeLocation.Pregraph);
                includes.Add(CoreIncludes.kPickingSpaceTransforms, IncludeLocation.Pregraph);
                includes.Add(CoreIncludes.kPassDepthOnly, IncludeLocation.Postgraph);

                return includes;
            }
            
            move  includes.Add(CoreIncludes.kPickingSpaceTransforms, IncludeLocation.Pregraph); below includes.Add(CoreIncludes.CorePregraph);

kecho: Good call.

kecho: Done.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl
kecho: Done

sebastienlagarde (approved):

sebastienlagarde (commented):
please add a changelog

FrancescoC-unity (approved):

TomasKiniulis (approved):
Perfect! No issues occur anymore with LitTessellation and LayeredLitTessellation oribiting around an object in wireframe display. Made sure that shadergraph lit and axf shaders are unaffected as well.

PR #3361, opened 2021-02-03
[Yamato] remove force-push from update script

kivistikliis:
---
### Purpose of this PR
Removes deprecated force-push argument from update script


.yamato/script/gfx_sdet_tools_revision.txt
```diff
@@ -1 +1 @@
-33fc12e281d192275bd5b3021cb941e9d8ee11a8
+98903c0062b36e3d4780f162bf0a6e97d98b8f18
```
theo-pnv, line 1: Shouldn't it be `55d2a90` ? (HEAD of gfx-sdet-tools)

theo-pnv (approved):

.yamato/script/gfx_sdet_tools_revision.txt
kivistikliis: I've used the revision of the commit before I merged it to master, so it includes the correct changeset

PR #3362, opened 2021-02-03
Added the default quality settings to the HDRP asset for RTAO, RTR and RTGI (case 1304370).

anisunity:
https://fogbugz.unity3d.com/f/cases/1304370/

Testing status
Checked that the new submenus display correctly.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.cs
@@ -293,7 +293,7 @@ namespace UnityEditor.Rendering.HighDefinition
 
                 --EditorGUI.indentLevel;
             }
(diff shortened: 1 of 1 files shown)


remi-chapelain (approved):
Tried to change the settings to see if it applies correctly when changing quality settings ✔️ 
Modified a bit the labels and order to be consistent between settings and actual naming on the overrides. 

I'll report this later but I thought of 2 things

First, 
RTAO has its SSAO quality setting fallback, 
RTR has its SSR quality setting fallback, 
RTGI has **NOT** its quality setting fallback whereas there's some quality settings in SSGI. 
image

Secondly, we should indent everything in the quality setting in between low/med/high, for it to be clearer and easier to read
image

remi-chapelain:
Fixed the #2 mentioned in the previous comment in the last commit.

sebastienlagarde (approved):

PR #2595, opened 2020-11-12
[HDRP] Fix runtime debug UI showing wrong entry

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1290638/ 

The issue was that the change of the value and the redraw of the UI upon change was off-sync. This fixes that. 

**What did I test**
 Scrolled through debug menu options and verify the name displayed matches what's happening on screen.

PR #3370, opened 2021-02-03
Fixed debug panel reseting when going through enum items

JulienIgnace-Unity:
### Purpose of this PR
This PR is a fix for https://fogbugz.unity3d.com/f/cases/1307100/

It was caused by two olders PRs:
https://github.com/Unity-Technologies/Graphics/pull/2162
https://github.com/Unity-Technologies/Graphics/pull/2595

The original issue fixed by those PR is no longer reproducible (from what we can tell, it was just caused by the fact that runtime debug shaders were not included in the build) so I just reverted those faulty PRs.

---
### Testing status
Tested locally as well as with @remi-chapelain


remi-chapelain (approved):
Already verified using the branch before the PR ✔️

sebastienlagarde (approved):

PR #3371, opened 2021-02-03
Add explicit control for double sided GI 

FrancescoC-unity:
Added option to control if a material is considered double sided for GI or not to be on par with what Builtin has. 

image

This allows for independent control. 

What did I test. 

I used this scene setup, where if double sided for GI is **not used**  will look like

image

If it is used it looks like 

image

Given that, I tried both with a Lit material and a Lit from shadergraph to try all the new options with the material double sided option ON and OFF (all combinations).

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/2109_GI_DoubleSided.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/2109_GI_DoubleSided.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b2cefeb8eba706cd1c7f3e76a4b8cac8f5bb9de584861d4be39246f0392e7833
-size 6224
+oid sha256:af4127c6b976b709be5429aee973a82675c586bf81951bcc94fae49f4646b028
+size 64738
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/2109_GI_DoubleSided.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/2109_GI_DoubleSided.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b2cefeb8eba706cd1c7f3e76a4b8cac8f5bb9de584861d4be39246f0392e7833
-size 6224
+oid sha256:8fb0eae82f8e8e15027a7cc3feb78aaca6540d3763679e71eb2a2cceab9b765e
+size 65005
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/2109_GI_DoubleSided.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/2109_GI_DoubleSided.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b2cefeb8eba706cd1c7f3e76a4b8cac8f5bb9de584861d4be39246f0392e7833
-size 6224
+oid sha256:a27bbe970de3f5a9a5928148e214179ad42be3723084d75f75d1e4435a966953
+size 64737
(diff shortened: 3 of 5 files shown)


TomasKiniulis:
@FrancescoC-unity I think surface options doc page should be updated https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.3/manual/Surface-Type.html and the PR could benefit from a dedicated automated test scene, also maybe it would be possible to have spaces in input field options? For example "Forces On" instead of "ForcesOn".

FrancescoC-unity:
I'll update the doc! I will also try to add a test, will have to remember in a bit :D 

Regarding the space, unfortunately I don't think we can.

sebastienlagarde (commented):
Sanity check. QA you should test migration. I expect we don't need any migratoin code because the default value (0) of the enum is matchmaterial so the same than the old behavior. But better to QA it.

Also for QA be sure to test with both ShaderGraph and non shader graph shader.

radishface:
Thanks for implementing this! I would suggest renaming the enums in the UI to:

- `MatchMaterial` to `Auto`
- `ForcesOn` to `On`
- `ForcesOff` to `Off`

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDSubShaderUtilities.cs
```diff
@@ -218,6 +218,16 @@ public static void AddDoubleSidedProperty(PropertyCollector collector, DoubleSid
                 hlslDeclarationOverride = HLSLDeclaration.UnityPerMaterial,
                 value = new Vector4(1, 1, -1, 0)
             });
+            collector.AddShaderProperty(new Vector1ShaderProperty
+            {
+                enumNames = { "MatchMaterial", "ForcesOn", "ForcesOff" }, // values will be 0, 1 and 2
```
radishface, line 223: Please rename enum values in the UI to:
`MatchMaterial` to `Auto`
`ForcesOn` to `On`
`ForcesOff` to `Off`

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/SurfaceOptionUIBlock.cs
```diff
@@ -103,6 +103,7 @@ internal static class Styles
             // Lit properties
             public static GUIContent doubleSidedNormalModeText = new GUIContent("Normal Mode", "Specifies the method HDRP uses to modify the normal base.\nMirror: Mirrors the normals with the vertex normal plane.\nFlip: Flips the normal.");
             public static GUIContent depthOffsetEnableText = new GUIContent("Depth Offset", "When enabled, HDRP uses the Height Map to calculate the depth offset for this Material.");
+            public static GUIContent doubleSidedGIText = new GUIContent("Double-Sided GI", "Determines how double-sided GI is handled.");
```
radishface, line 106: Please make sure that the tooltip matches our [documentation page](https://docs.unity3d.com/ScriptReference/Material-doubleSidedGI.html):

> When enabled, the lightmapper accounts for both sides of the geometry when calculating Global Illumination. Backfaces are not rendered or added to lightmaps, but get treated as valid when seen from other objects. When using the Progressive Lightmapper backfaces bounce light using the same emission and albedo as frontfaces. (Currently this setting is only available when baking with the Progressive Lightmapper backend.)

radishface (changes requested):
Changes look good! Only had a couple of UI-related suggestions to add.

radishface (approved):
Thank you for implementing changes so quickly. Approved!

sebastienlagarde:
Note for me: must be retarget to master once hd/bugfix is merge

iM0ve (changes requested):
**Whats tested:**
* Upgrading + shadergraph
* Lit, LayeredLitTesselation, Unlit, 
* Raytraced SSGI
* Path tracing

**Concerns**

1) **[Fixed]** The option does not exist for Unlit? Expected?

Unlit


2) **[Expected]** Has no effect on Raytraced SSGI or Path tracing. I assume this is expected because the tooltip says its only lightmapping. But want to clarify if we want parity for raytracing

Path

**Conclusion:**
I found no major issues and upgrading works. Just want to clarify the concerns and will approve.
Update: issues addressed/fixed

FrancescoC-unity:
> **Whats tested:**
> 
> * Upgrading + shadergraph
> * Lit, LayeredLitTesselation, Unlit,
> * Raytraced SSGI
> * Path tracing
> 
> **Concerns**
> 
> 1. The option does not exist for Unlit? Expected?
> 
> Unlit
> 
> 1. Has no effect on Raytraced SSGI or Path tracing. I assume this is expected because the tooltip says its only lightmapping. But want to clarify if we want parity for raytracing
> 
> Path
> 
> **Conclusion:**
> I found no major issues and upgrading works. Just want to clarify the concerns and will approve.

Indeed this is just for baked GI, I'd personally keep Ray tracing "proper". The reason why you want this separate is mostly tied to how the baking process works. 

I guess Unlit should be there, I'll fix it up

iM0ve (approved):
Updated the review

sebastienlagarde:
screenshots have been updated

PR #3372, opened 2021-02-03
Upgraded URP asset

phi-lira:
### Purpose of this PR
 Fixed OnDemand scheduler failure while trying to load URP.asset file

### Testing status
 Runned https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Fvfx_urp-win-dx11.yml%2523Build_VFX_URP_Win_DX11_Standalone_PerformanceTests_mono_Linear_trunk/5164983/job 3 times in a row

### Comments to reviewers
 I printed out the name of the asset that was causing the OnDemand schedule assert. That pointed out to URP.asset file. 
 It seems updating the asset causes the test to be fixed but I'm not convinced this is the real issue. Following up with AssetDatabase team but at least this means we get CI running again.


PaulDemeulenaere (approved):
We discussed about this change with @fabien-unity, it avoids a random issue with AssetDB.
LGTM 🟢

PR #3373, opened 2021-02-03
Hd/fix input registering domain reload

RSlysz:
### Purpose of this PR
Improve performance for DebugMenu input registering system at each domain reload. FB: https://fogbugz.unity3d.com/f/cases/1311108/

I found that there is way more call to serialization (which is the bottle neck here) than needed. So I reworked the checking algorithm to use the less possible serialization. Checking was here to be sure we do not register duplicate entry. Now it is done once on all the cached entry instead of one time per new entry.
So basically we previously had a complexity of roughly O(n.m) where n was the pre existing entry amount and m the inserted entry amount. And with this change we are in O(n+k) where k is the entry that we can safely insert without creating duplicate (k < m, often 0). (Complexity is computed against serialization call to get element in the array)

Also I delayed the actual register to next editor frame. This in order to let a chance to pack in case of several class request to register an input at initialization. The costly operation of the new algorithm is the caching of pre existing entries. If we can afford to prevent being done several time, it worth it.

And finally, the InputSystem package is supported by the DebugMenu but in this case, it still try to add inputs in the asset used by the legacy system. I skipped this operation as it is not needed.

In addition, I added warnings when using InputRegistering while the InputSystem package is in use.

---
### Testing status

In test project given with the FB above, there is 529 entries in ProjectSettings > InputManager

Before fix
image
image

After fix (without delaying next frame to have meaningful results) without InputSystem package
image
image

Tested it is totally skipped when InputSystem is in use. Tested with the project from FB https://fogbugz.unity3d.com/f/cases/1306751/ which use a lot more entry in the input asset. Times goes to ~1min (2754 entries) to almost nothing. 

And DebugMenu still working after change in both project (with and without the InputSystem package).

**Note:** tests have been done before changing from using GetArrayElementAtIndex to Next wich use even less computation time. So expect even better results.

---
### Comments to reviewers

Note: the perfect fix should be to have a C# API for this. Now that there is a dedicated package for this, I dunno if any development on the subject is something to think of or not.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Inputs/InputRegistering.cs
+++ b/com.unity.render-pipelines.core/Runtime/Inputs/InputRegistering.cs
@@ -59,12 +59,12 @@ namespace UnityEngine
 
         static void AddEntriesWithoutCheck(SerializedProperty spAxes, List<InputManagerEntry> newEntries)
         {
-            int startRange = spAxes.arraySize;
-            int endRange = startRange + newEntries.Count;
-            spAxes.arraySize = endRange;
+            int endOfCurrentInputList = spAxes.arraySize;
+            spAxes.arraySize = endOfCurrentInputList + newEntries.Count;
 
-            for (int i = startRange; i < endRange; ++i)
-                CopyEntry(spAxes.GetArrayElementAtIndex(i), newEntries[i]);
+            SerializedProperty spAxis = spAxes.GetArrayElementAtIndex(endOfCurrentInputList);
+            for (int i = 0; i < newEntries.Count; ++i, spAxis.Next(false))
+                CopyEntry(spAxis, newEntries[i]);
         }
 
         // Get a representation of the already registered inputs
@@ -72,11 +72,10 @@ namespace UnityEngine
         {
             int size = spAxes.arraySize;
             List<(string name, InputManagerEntry.Kind kind)> result = new List<(string name, InputManagerEntry.Kind kind)>(size);
-            for (int i = 0; i < size; ++i)
-            {
-                var spAxis = spAxes.GetArrayElementAtIndex(i);
+
+            SerializedProperty spAxis = spAxes.GetArrayElementAtIndex(0);
+            for (int i = 0; i < size; ++i, spAxis.Next(false))
                 result.Add((spAxis.FindPropertyRelative("m_Name").stringValue, (InputManagerEntry.Kind)spAxis.FindPropertyRelative("type").intValue));
-            }
             return result;
         }
 


com.unity.render-pipelines.core/Runtime/Inputs/InputRegistering.cs
```diff
@@ -76,18 +149,39 @@ public static void RegisterInputs(List<InputManagerEntry> entries)
...
+        public static void RegisterInputs(List<InputManagerEntry> entries)
+        {
+#if ENABLE_INPUT_SYSTEM && ENABLE_INPUT_SYSTEM_PACKAGE
+            Debug.LogWarning("Trying to add entry in the legacy InputManager but using InputSystem package. Skiping.");
+            return;
+#else
+            s_PendingInputsToRegister.AddRange(entries);
+
+            //delay the call in order to do only one pass event if several different class register inputs
+            EditorApplication.delayCall += DelayedRegisterInput;
```
RSlysz, line 182: This line is replaced by `DelayedRegisterInput();` in the test to compare algorithm in same conditions.

com.unity.render-pipelines.core/Runtime/Inputs/InputRegistering.cs
```diff
@@ -66,8 +57,76 @@ static void WriteEntry(SerializedProperty spAxes, InputManagerEntry entry)
             spAxis.FindPropertyRelative("joyNum").intValue = (int)entry.joystick;
         }
 
-        public static void RegisterInputs(List<InputManagerEntry> entries)
+        static void AddEntriesWithoutCheck(SerializedProperty spAxes, List<InputManagerEntry> newEntries)
         {
+            int startRange = spAxes.arraySize;
```
jenniferd-unity, line 62: note for next time: i got confused because expected a range here. so could you confirm startRange is basically firstEntryIndex and endRange is lastEntryIndex?

com.unity.render-pipelines.core/Runtime/Inputs/InputRegistering.cs
```diff
@@ -76,18 +135,39 @@ public static void RegisterInputs(List<InputManagerEntry> entries)
...
+        public static void RegisterInputs(List<InputManagerEntry> entries)
+        {
+#if ENABLE_INPUT_SYSTEM && ENABLE_INPUT_SYSTEM_PACKAGE
+            Debug.LogWarning("Trying to add entry in the legacy InputManager but using InputSystem package. Skiping.");
+            return;
+#else
+            s_PendingInputsToRegister.AddRange(entries);
+
+            //delay the call in order to do only one pass event if several different class register inputs
+            EditorApplication.delayCall += DelayedRegisterInput;
```
jenniferd-unity, line 168: does this mean we could end up calling this function multiple times in the same frame?
also when do we unregister the call?

com.unity.render-pipelines.core/Runtime/Inputs/InputRegistering.cs
RSlysz: Yes I was refering it at range like in the list.AddRange(...) method.
So startRange = firstEntryPositionInInputListIndex and endRange = lastEntryPositionInInputListIndex

Just first/lastEntryIndex is unclear if it is in the newEntries list or in the spAxes array. So I'll keep range and add a comment.

com.unity.render-pipelines.core/Runtime/Inputs/InputRegistering.cs
RSlysz: No it means you can call it multiple time in a frame, it will all be batched in delayed call.
Example:
Frame1, two different classes call register input
```
RegisterInputs(new List<InputManagerEntry>
      {
            new InputManagerEntry { name = "cmd1",  kind = InputManagerEntry.Kind.KeyOrButton, btnPositive = "left ctrl",   altBtnPositive = "joystick button 8" },
            new InputManagerEntry { name = "cmd2",  kind = InputManagerEntry.Kind.KeyOrButton, btnPositive = "backspace",   altBtnPositive = "joystick button 9" }
      });
      
      
RegisterInputs(new List<InputManagerEntry>
      {
            new InputManagerEntry { name = "cmd3",  kind = InputManagerEntry.Kind.KeyOrButton, btnPositive = "left alt",   altBtnPositive = "joystick button 1" }
      });
```
Delayed  calls:
- register cmd1, cmd2 and cmd3 at the same time
- nothing to register, so exit early

So we inspected the already registered inputs only once instead of two time when done at same frame.



And for unregistering, the mechanism where not here prior and was never requested. I assume it is not needed in this case. But it can be implemented in a similar way. I assume this was done to add inputs in the InputManager.asset only and not remove them.

RSlysz:
I have seen Martin comment on the thread of the bug that suggest to use Next instead of GetArrayElementAtIndex for better performances. I'll amend this PR with this.

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Seems good! Did a validation check with user project adjusting InputManager entries. HDAdditionalCameraData stays at about 60ms now instead of about 2 sec. Haven't encountered any PR related issues.

My test results:
Before PR:
image
With PR:
image

sebastienlagarde:
@RSlysz : There is a problem with this PR, I have reverted it from hd/bugfix.
Failure is here: https://yamato-artifactviewer.prd.cds.internal.unity3d.com/92253399-6df5-498f-806d-ad672245efce%2Flogs%2FTestProjects%2FHDRP_Tests%2Ftest-results/TestReportV1.html

PR #3374, opened 2021-02-03
[HDRP][PathTracing] Added a clamping mode parameter to path tracing options

eturquin:
This PR addresses https://fogbugz.unity3d.com/f/cases/1310514/

It introduces a new "Intensity Clamping" enum parameter to the path tracer, which can bet set to:
- Indirect Only: the default mode, same behaviour as before, ie the clamping is only applied to indirect path segments
- Direct and Indirect: clamping is applied to all path segments
- None: no clamping at all

This last mode guarantees a better match with baked solutions in cases where a large part of the lighting comes from indirect, in which case clamping will reduce noise, but also as an undesirable side effect darken the result significantly.

It has been tested on the scene used in bug case 1310514, and the path-traced image without clamping gives us a much closer match indeed (see screenshots on the corresp. ticket).

Waiting for feedback about this solution; an alternative would be to make the "Maximum Intensity" parameter a value with min=0 but no max (so no slider anymore), in which case you could set a huge value to "disable" clamping. However, this would not provide the possibility to apply clamping either on indirect only, or direct+indirect.

Doc and changelog will be taken care of after the most suitable solution has been chosen.

Also note that intensity was kept rather than the specialised exposure slider, as it expresses a post-exposition pixel value (but before post-processes).

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/PathTracing/PathTracingEditor.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/PathTracing/PathTracingEditor.cs
@@ -16,6 +16,7 @@ namespace UnityEditor.Experimental.Rendering.HighDefinition
         SerializedDataParameter m_MaxSamples;
         SerializedDataParameter m_MinDepth;
         SerializedDataParameter m_MaxDepth;
+        SerializedDataParameter m_IntensityClamping;
         SerializedDataParameter m_MaxIntensity;
 
         public override void OnEnable()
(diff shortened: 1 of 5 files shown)


eturquin:
There seems to be a preference (including from myself) for simply extending the range of the original attribute (possibly to [0; inf[ and removing the slider), rather than introducing a new attribute. Let's wait a bit for more opinions.

sebastienlagarde (approved):

eturquin:
Even though it did solve the initial bug, this change shouldn't be merged as is, as a different UI option would be preferable.

iM0ve (changes requested):
The current implementation fixes the issue but also introduces other corner cases. I think we should iterate more on this. If possible use the Exposure slider and link the Maximum intensity non arbitrary numbers, so its easier for users to know what value to choose for different lighting conditions. 

A simpler less user friendly solution would be to have the slider changed to an Int value. So that you could input any number that fits your scene. With the current implementation you can only choose Unlimited or Max clamp intensity 100

eturquin:
A simpler change will be implemented instead: change the current slider+value field to a basic value field, with no max value. A more detailed tooltip will explain the tradeoff between lower max for less fireflies, and higher values if darkening it too important.

PR #3375, opened 2021-02-03
[ShaderGraph] [2021.2] Added sanitization to virtual texture's layer reference names.

joshua-davis:

### Checklist for PR maker

---
### Purpose of this PR
Fix for bug https://fogbugz.unity3d.com/f/cases/1304146/

The layer reference name on virtual textures could have invalid characters if you paste them in (normally sanitized during typing). This change calls the hlsl name sanitizing function on the layers on change.

---
### Testing status
Manually tested by following the repro steps (paste invalid characters into the layer reference name).

---
### Comments to reviewers


LandonTownsendUnity (approved):
Only bug I found is that the collision checker will check what the name was previously (for instance if you have __1 and paste in %^ you'll get __2 instead of __1 again). This is probably too minor to hold back the fix so I'll approve. If you think it would be an easy fix you can fix it.

joshua-davis:
> Only bug I found is that the collision checker will check what the name was previously (for instance if you have __1 and paste in %^ you'll get __2 instead of __1 again). This is probably too minor to hold back the fix so I'll approve. If you think it would be an easy fix you can fix it.

Fixed. In addition I noticed on my first attempt that doing this would leave the old text there until I clicked off and on the property again. I believe I've fixed this as well but it's worth testing.

LandonTownsendUnity (approved):

jessebarker (approved):
LGTM

PR #3380, opened 2021-02-04
[Universal] RenderPass conversion

jonuuukas:
### Purpose of this PR
Initial PR for the URP RenderPass conversion. This adds the yet limited feature in a hidden manner as it still needs work. This converts current Forward+Deferred in a 1:1 manner - ScriptableRenderPass corresponds to RenderPass. This is not feasible performance-wise on TBDRs, but is a good starting point, as we will continue implementing RenderPass lazy merging. MSAA implementation is not here as well, though we are including some of the related methods (such as ConfigureResolveTarget and etc.)

All of the API and variables are internal, one can enable the checkbox to enable native RenderPass by adding "ENABLE_RENDER_PASS_UI" to the Scripting define symbols. This will create a new checkbox in the renderer assets which can be used to enable RenderPass during runtime or before. 

This is work in progress, so documentation and QA passes will be done in the future when we will officially be exposing this API from internal to public
---
### Testing status
This is still very experimental and didn't have a QA round as a lot of work still has to be done. However, on the editor, ~124 tests are green, with few known failures that require changes in the engine-side. 
---
### Comments to reviewers


kaychang-unity:
Is ScriptableRenderPass.srpDescriptor a mechanism to get information on render-targets that are not available through RenderTargetIdentifier, such as graphicsFormats, width, height, ... ? It is a shame RenderTargetidentifier is so limited.

Anyway, I don't think I am the most qualified to comment on the design of of RenderPass and how it integrates into URP code.

kaychang-unity (approved):

jonuuukas:
@kaychang-unity - srpDescriptor currently serves that purpose, yeah. However, when the transition to RTHandles happens, we will be removing this class, as the RTHandles carry all the required information for the RenderPass execution.

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -55,6 +56,7 @@ public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderin
         {
             if (this.allocateDepth)
                 cmd.GetTemporaryRT(depthAttachmentHandle.id, descriptor, FilterMode.Point);
+            srpDescriptor.ConfigureDescriptor(GraphicsFormat.DepthAuto, descriptor.width, descriptor.height, 1, true);
             ConfigureTarget(new RenderTargetIdentifier(depthAttachmentHandle.Identifier(), 0, CubemapFace.Unknown, -1));
```
phi-lira, line 60: API wise I have some concerns with this. ScriptableRenderPassDescriptor has a mix of render pass data and attachment data (format). 
 
 We are configuring attachment data in two APIs (ConfigureDescriptor (format) and ConfigureTarget (surface handle))

phi-lira: So now you have overloads of ConfigureDescriptor and ConfigureTarget taking arrays of attachment data. 
My suggestion is to let ConfigureTarget configure all the per-attachment information, maybe when we support RTHandle it can just take an array of that as that provides the GraphicsFormat you need, but meanwhile an internal overload to take GraphicsFormat and RenderTargetIdentifiers could work?

The other fields you have in `ScriptableRenderPassDescriptor` (width, height, samples) why do we need a descriptor for that instead of storing in the ScriptableRenderPass itself? 

So if user wants to query width of a render pass API would be simpler `myPass.width` instead of `myPass.srpDescriptor.width`?

com.unity.render-pipelines.universal/Runtime/Passes/GBufferPass.cs
```diff
@@ -68,10 +70,11 @@ public override void Configure(CommandBuffer cmd, RenderTextureDescriptor camera
                 RenderTextureDescriptor gbufferSlice = cameraTextureDescriptor;
                 gbufferSlice.depthBufferBits = 0; // make sure no depth surface is actually created
                 gbufferSlice.stencilFormat = GraphicsFormat.None;
-                gbufferSlice.graphicsFormat = m_DeferredLights.GetGBufferFormat(i);
+                gbufferSlice.graphicsFormat = gbufferFormats[i];
                 cmd.GetTemporaryRT(m_DeferredLights.GbufferAttachments[i].id, gbufferSlice);
```
phi-lira, line 74: I guess this is one example to justify my worry with the API design you have today.

This entire piece of code will be replaced by RTHandle in the future. So then gbufferSlice.graphicsFormat will not exist. If you have two sets of API taking per-attachement data (ConfigureDescriptor(formats), ConfigureTarget(rtIDs)) then you have to loop the RTHandle array just to get those arrays. 

So, in essence making your API today pass the GraphicsFormat array in ConfigureTarget will make it easy to switch to ConfigureTarget taking an RTHandle array once we support it.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPassDescriptor.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace UnityEngine.Rendering.Universal
+{
+    internal sealed class ScriptableRenderPassDescriptor
+    {
+        internal int width;
+        internal int height;
+        internal int sampleCount;
+        internal GraphicsFormat[] formats;
+        internal bool isLastPass;
+        internal bool isDepthOnly;
```
phi-lira, line 13: Considering GraphicsFormat comes from RTHandle. Why not having the other settings in the ScriptableRenderPass itself?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -189,6 +193,8 @@ public ScriptableRenderPass()
             overrideCameraTarget = false;
             isBlitRenderPass = false;
             profilingSampler = new ProfilingSampler(nameof(ScriptableRenderPass));
+            useNativeRenderPass = true;
```
phi-lira, line 196: This API is internal, what does it mean for custom passes that can't change this userNativeRenderPass?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -724,7 +770,62 @@ void ExecuteRenderPass(ScriptableRenderContext context, ScriptableRenderPass ren
...
+                bool useDepth = m_ActiveDepthAttachment == RenderTargetHandle.CameraTarget.Identifier() && (!(isLastPassToBB || (isLastPass && cameraData.camera.targetTexture != null)));
+
+                var attachments =
+                    new NativeArray<AttachmentDescriptor>(useDepth && !renderPass.srpDescriptor.isDepthOnly ? validColorBuffersCount + 1 : 1, Allocator.Temp);
+
+                for (int i = 0; i < validColorBuffersCount; ++i)
+                    attachments[i] = m_ActiveColorAttachmentDescriptors[i];
+
+                if (useDepth && !renderPass.srpDescriptor.isDepthOnly)
+                    attachments[validColorBuffersCount] = m_ActiveDepthAttachmentDescriptor;
```
phi-lira, line 788: This could be encapsulated in the ScriptableRenderPass when configuring attachments we could already cache attachment descriptors?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -724,7 +770,62 @@ void ExecuteRenderPass(ScriptableRenderContext context, ScriptableRenderPass ren
...
+                int height = renderPass.srpDescriptor.height != -1 ? renderPass.srpDescriptor.height : desc.height;
+                sampleCount = renderPass.srpDescriptor.sampleCount != -1
+                    ? renderPass.srpDescriptor.sampleCount
+                    :
+#if UNITY_EDITOR
+                    !isLastPassToBB ? sampleCount : 1;
+#else
+                    //In case MSAA is disabled on camera, we need to still check the QualitySettings, as backbuffer will be created with that setting in mind
+                    isLastPassToBB? QualitySettings.antiAliasing : sampleCount;
+#endif
```
phi-lira, line 802: This sampleCount logic and last pass is very convoluted.

phi-lira (changes requested):
Made some comments, but have not finished review yet. Let's catch up tomorrow to go through comments? It might be easier to discuss and hear your thoughts over a call.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
manuele-bonanno: most of this stuff wasn't needed anymore and was related to an old FinalBlitPass conversion. Simplified the code

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
jonuuukas: TODO: add a comment

com.unity.render-pipelines.universal/Editor/UniversalRendererDataEditor.cs
```diff
@@ -74,6 +81,13 @@ public override void OnInspectorGUI()
             }
             EditorGUI.indentLevel--;
             EditorGUILayout.Space();
+#if ENABLE_RENDER_PASS_UI
```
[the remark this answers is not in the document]
jonuuukas: At this time this feature will regress the performance on these platforms, so we shouldn't really show it. This could be done after lazy merging is done, but for now - lets keep it hidden

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
jonuuukas: SrpDescriptor changes were reverted and now the API is using ConfigureTarget overloads to get this data

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPassDescriptor.cs
jonuuukas: SrpDescriptor changes were reverted and now the API is using ConfigureTarget overloads to get this data

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
jonuuukas: At the end, we expect to have a preprocessing mechanism that would determine whether pass is eligible or not to be inside a RenderPass, so this is temporary. Also, Native RenderPass has to be enabled in the renderer asset and having this value as true won't really affect anything unless it's explicitly enabled

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
jonuuukas: This is temporary code as it will be changing in the future with lazy merging being introduced. We technically could cache something per-pass, but that will change as soon as we start grouping the render passes

jonuuukas:
@Unity-Technologies/mobile-qa - please do a sanity check on this PR. Checking BoatAttack and UniversalRenderingExamples should be enough. The feature itself shouldn't be tested now, we are only interested whether default behaviour has regressed or got broken with this PR. So basically opening and running these projects on few platforms should be enough

ernestasKupciunas (approved):
Looks good on the mobile side. There were no issues related to these changes, and the default behaviour seems to work as expected.

Testing doc: https://docs.google.com/document/d/1fM3mptWJuI4hvk78xSXi4nOwoswl1ZUWCriNsGOlMMk/edit?usp=sharing

Devices under testing:
VLNQA00108, LGE LG Leon 4G LTE, 5.0.1, Snapdragon 410 MSM8916, Adreno 306

manuele-bonanno (approved):

phi-lira (approved):

ellioman:
Failures on Yamato are known and not caused by this PR.

PR #3381, opened 2021-02-04
Use a custom HelperURL attributes to allow an automatic generation of package ver

FrancescoC-unity:
This adds two attributes (one for Core + One for HDRP - URP uses HelpURL attribute with a constant already) that auto generates t he right pattern of URL without having to specify a constant explictely as the package version is obtained through pacman. 

Note: the feature needed to have this was **not** backported to 2020.2 so I couldn't test it on yamato. That said

**What did I test**: I checked that the ? links on components are just there and they form the right URL (note that the link it will create leads to non-existent page as the documentation has not been published yet for this version, but it looks like: e.g. https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@12.0/manual/Override-Exposure.html ) 
I did not test _all_ the modified things but a bunch of them (from both core and HDRP)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/LocalVolumetricFog.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/LocalVolumetricFog.cs
@@ -161,7 +161,7 @@ namespace UnityEngine.Rendering.HighDefinition
     } // class LocalVolumetricFogParameters
 
     /// <summary>Local Volumetric Fog class.</summary>
-    [HelpURL(Documentation.baseURL + Documentation.version + Documentation.subURL + "Local-Volumetric-Fog" + Documentation.endURL)]
+    [HDRPHelpURLAttribute("Local-Volumetric-Fog")]
     [ExecuteAlways]
     [AddComponentMenu("Rendering/Local Volumetric Fog")]
     public partial class LocalVolumetricFog : MonoBehaviour


com.unity.render-pipelines.core/Runtime/Documentation.cs
```diff
@@ -1,21 +1,61 @@
...
         /// <summary>
         /// Current version of the documentation.
         /// </summary>
-        public const string version = "11.0";
+        public static string version
+        {
+            get
+            {
+                PackageInfo packageInfo = PackageInfo.FindForAssembly(typeof(DocumentationInfo).Assembly);
+                return packageInfo == null ? fallbackVersion : packageInfo.version.Substring(0, 4);
```
sebastienlagarde, line 56: nice! We will not have to update this number anymore, cool :)

sebastienlagarde:
Adding @JordanL8 and QA for visibility but there is nothing to check here.

sebastienlagarde:
note to me: retarget the PR on master once hd/bugfix is merge

TomasKiniulis (approved):
Looking through description and the file changes this seems good! Approving

JordanL8 (approved):
Awesome!

johnpars (approved):
LGTM :)

phi-lira (approved):

sebastienlagarde (approved):

sebastienlagarde:
Waiting for https://ono.unity3d.com/unity/unity/pull-request/122373/_/assetpipeline/asset-matching-crash-fix/1319521

PR #3387, opened 2021-02-04
Add new API to CachedShadowManager  and fix WouldFitInAtlas

FrancescoC-unity:
This PR introduces a couple of things: 

**New public API**

 - A new WouldFitInAtlas that takes directly a light. 
 - A LightHasBeenPlacedInAtlas to verify if a light has found its placement in the cached shadow atlas. 

**Fixes**
- Would fit in atlas would previously would wrongly return true even if only one face of a point light shadow fit in the atlas. 

EDIT: 

Added even more API:

- LightHasBeenPlaceAndRenderedAtLeastOnce needed as light OnDemand don't render automatically upon placement so they might contain stale data and the user might want to know about it.


FrancescoC-unity:
Note: there might be a slight issue with the new API, please don't merge I'll have to do a small pass on Monday.

sebastienlagarde (approved):

sebastienlagarde:
@FrancescoC-unity guess we should update the documentation about sahdow caching as well to talk about this new feature

FrancescoC-unity:
Yes, will add the docs when I do a small pass on it (likely Wednesday as it coincide with a meeting I will have)

sebastienlagarde:
I have remove the QA on this PR as it is not something we can easily test. It was tested by Bornfire for validation instead.

PR #3389, opened 2021-02-04
[HDRP][PathTracing] High resolution sky on camera ray misses

eturquin:
This addresses: https://fogbugz.unity3d.com/f/cases/1304114

Before:
On camera ray misses, path tracing was using the reflection routines, which rely on a low res cube map to compute their value. This resulted in a sky "background" that was looking lower res in path tracing than in rasterization.

Now:
We now render once per accumulation (on the first frame) a sky texture matching the output buffer resolution, and use this instead on the path-traced camera ray misses. As a result, the sky will appear the same as with rasterization, at a minimal cost.

Extras:
Code has been cleaned up here and there for better compliance with rendergraph (in particular, the texture used to store the path traced result at each frame is now properly managed).

Testing:
The new approach has been tested with the scene from the aforementioned Fogbugz case.
Further tests should make sure that it works properly with all kinds of skies, and in particular that we get the expected resolution improvement on HDRIs.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
```diff
@@ -306,26 +322,52 @@ class RenderPathTracingData
...
+                passData.hdCamera = hdCamera;
+                passData.colorBuffer = builder.WriteTexture(skyBuffer);
+                passData.depthTexture = builder.WriteTexture(CreateDepthBuffer(renderGraph, true, false));
+                passData.debugDisplaySettings = m_CurrentDebugDisplaySettings;
+                passData.skyManager = m_SkyManager;
+
+                builder.SetRenderFunc(
+                    (RenderSkyPassData data, RenderGraphContext context) =>
+                    {
+                        data.skyManager.RenderSky(data.hdCamera, data.sunLight, data.colorBuffer, passData.depthTexture, data.debugDisplaySettings, context.cmd);
```
pmavridis, line 365: passdata.depthTexture should be data.depthTexture, otherwise we will get GCAlloc during rendering as far as I know.
(we had similar issues in many places in the past)

pmavridis (approved):
Looks good to me. 
Rasterizing the sky only for the first frame of the accumulation is a good optimization, but might not work so well with accumulation motion blur on a rotating camera. If it turns out to be a problem, we can easily fix it, so I would not worry about this now.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
JulienIgnace-Unity: Yes, this will generate GCAlloc. Simple way to check that you don't capture any variables in the lambda is to hover the mouse on the  => in visual studio. It will tell you if anything is captured.

eturquin:
> Rasterizing the sky only for the first frame of the accumulation is a good optimization, but might not work so well with accumulation motion blur on a rotating camera. If it turns out to be a problem, we can easily fix it, so I would not worry about this now.

Good point, I've made a minor change so that we render the sky on every frame when recording (we can refine this further if needs be, but it's not likely to be a bottleneck... :D).

JulienIgnace-Unity (approved):
Just one minor issue with a variable captured otherwise lgtm.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
eturquin: Oh yeah that was a typo, thanks! Fixed now.

remi-chapelain (approved):
Great improvement. 

Tested with multiple skies (PBR, Gradient, HDRI) ✔️ 
Also tested multiple HDRIs at different resolution up to 16k ✔️ 
7296901d93432dfe835b2b9803b0de34

Interesting side effect is that we get the sun disk back with the path tracer with the PBR Sky ! 
3b10dc2c3ea0b6e36a9de5f1bddd8704

sebastienlagarde (approved):

TestProjects/HDRP_RuntimeTests/Assets/Scenes/002-HDMaterials.unity
```diff
@@ -180,6 +180,7 @@ MeshRenderer:
   m_CastShadows: 1
   m_ReceiveShadows: 1
   m_DynamicOccludee: 1
+  m_StaticShadowCaster: 0
```
sebastienlagarde, line 183: Why 002 test is change here? it should be affected. Can you revert to master version thanks

sebastienlagarde:
Note for me: need to retarget this PR on master once hd/bugfix is merge, it shouldn't be in hd/bugfix

sebastienlagarde (approved):

sebastienlagarde:
Change to Draft. Waiting for your confirmation @eturquin to merge this one

PR #3390, opened 2021-02-04
[XPipeline]Reorder GameObject > Volumes, Light & VFX

alex-vazquez:
### Purpose of this PR

Reordering SRP related menus from the UX done by @MaddalenaUnity  and @TheoWong-pixel on this [document](https://docs.google.com/spreadsheets/d/14OzGtvhjp_nK-8LAeVlOMLQeyAf2OoZW5Ph519PWv40/edit#gid=195881870)

##### HDRP
image
image

- Changed GameObject/Rendering/Density Volume to GameObject/Volume/Density Volume
- Changed GameObject/Rendering/Decal Projector to GameObject/Decal Projector
- Changed GameObject/Volume/Sky and Fog Volume to GameObject/Volume/Sky and Fog Global Volume

---
### Testing status
Check that the menus appear in the same order as UX designed them


TheoWong-pixel:
Cna we push VFX up and remove the sub menu? if they are named the same, its a bit redundant.

com.unity.render-pipelines.universal/Editor/2D/Renderer2DMenus.cs
```diff
@@ -111,62 +101,62 @@ static bool CreateLightValidation()
...
 
-        [MenuItem("GameObject/Light/Spot Light 2D", false, k_MenuPriority)]
+        [MenuItem("GameObject/Light/Spot Light 2D", priority = CoreUtils.Sections.section3)]
         static void CreatePointLight2D(MenuCommand menuCommand)
         {
             CreateLight(menuCommand, Light2D.LightType.Point);
         }
 
-        [MenuItem("GameObject/Light/Global Light 2D", false, k_MenuPriority)]
+        [MenuItem("GameObject/Light/Global Light 2D",  priority = CoreUtils.Sections.section3)]
```
RSlysz, line 146: Shouldn't we enforce an order here like for other sections ?

com.unity.visualeffectgraph/Editor/VFXAssetEditorUtility.cs
```diff
@@ -62,7 +62,7 @@ static VisualEffectAssetEditorUtility()
         public const string editorResourcesFolder = "Editor/UIResources";
         public static string editorResourcesPath => VisualEffectGraphPackageInfo.assetPackagePath + "/" + editorResourcesFolder;
 
-        [MenuItem("GameObject/Visual Effects/Visual Effect", false, 10)]
+        [MenuItem("GameObject/VFX/VFX", priority = 12)]
```
RSlysz, line 65: Should this priority be a const somewhere in the VFX package? Seams to be the only menu entry. If this is the case, don't change it. It will be simpler this way.

RSlysz (approved):

com.unity.visualeffectgraph/Editor/VFXAssetEditorUtility.cs
alex-vazquez: I think is better to leave it here, as is the only menu of VFX on GameObjects

com.unity.render-pipelines.universal/Editor/2D/Renderer2DMenus.cs
alex-vazquez: I've enforce it for clarification, but the alphabetical order was correct too.

TheoWong-pixel (approved):

martint-unity (approved):

com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/ReflectionMenuItem.cs
```diff
@@ -27,7 +27,7 @@ static void CreateMirrorGameObject(MenuCommand menuCommand)
             }
         }
 
-        [MenuItem("GameObject/Light/Planar Reflection Probe", priority = CoreUtils.gameObjectMenuPriority)]
+        [MenuItem("GameObject/Light/Planar Reflection Probe", priority = 22)]
```
jenniferd-unity, line 30: why do we change to a fixed number and not _CoreUtils.Sections.section2 + CoreUtils.Priorities.gameObjectMenuPriority_ for example?

com.unity.render-pipelines.universal/Editor/2D/Renderer2DMenus.cs
```diff
@@ -111,62 +101,62 @@ static bool CreateLightValidation()
             return Light2DEditorUtility.IsUsing2DRenderer();
         }
 
-        [MenuItem("GameObject/Light/Freeform Light 2D/Square", false, k_MenuPriority)]
+        [MenuItem("GameObject/Light/Freeform Light 2D/Square", priority = CoreUtils.Sections.section3 + CoreUtils.Priorities.gameObjectMenuPriority + 4)]
```
jenniferd-unity, line 104: why did we remove the 'false' (is Validate function) in this file? is not needed anymore because of the fact everything is handled nicely?

alex-vazquez: We had  `false` because we need it to be able to call the ctor with the priority, but using `priority = ` on the argument of the constructor allow you to skip it, and it will get it's default value, that for `bool` in this case is false.

com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/ReflectionMenuItem.cs
alex-vazquez: This is the only item that goes under `Light`, I thought it would be more clear to have just the hardcoded number here instead of adding a new field on the `CoreUtils.Priorities`.

MaddalenaUnity (approved):

jenniferd-unity (approved):

iM0ve (approved):
Loaded the PR up for a quick check.
* Checked there are any related warnings/errors
* Created a density volume to see if the creation still works
* Checked that the new ordering is implemented

peeweek:
The VFX item should be at least named Visual Effect as we never refer as VFX but Visual Effect

Also, there are plans to add more items to the visual effects category in the future.

sebastienlagarde (commented):
Note: Don't know if we want to consider it or not but a "Density volume" is very different from the other type of volume. Density volume isn't part of the volume system, it is an object made to affect volumetric fog, but isn't a "volume". So not sure we should move it inside the same category as volume.

VladNeykov:
The VFX Graph has both a component and an asset. Does this PR aim at also renaming the asset-creation in the project folder?
Feels odd if these two have different names.
image

alex-vazquez:
@TheoWong-pixel , @sebastienlagarde , I've created another issue to tackle the Density Volume.

https://jira.unity3d.com/browse/XPIPELINE-90

I will create another PR with those changes

PR #3391, opened 2021-02-04
Removal of most old TextureFormat/RenderTextureFormat

JulienIgnace-Unity:
### Purpose of this PR
The goal of this PR is to remove references to RenderTexture and RenderTextureFormat enumerations from the Core and HDRP packages in preparation for their deprecation.
Most of them have been removed successfully but two problem remain with the current GraphicsFormat API.
- There is no equivalent to Alpha8 TextureFormat
- It's not possible to create properly a depth buffer (shadow map or not) without using the RenderTextureFormat.Depth/ShadowMap formats.
So all of those instances have been left in the code until the proper API is available.

---
### Testing status
Ran automated tests locally + yamato.


com.unity.render-pipelines.high-definition/Editor/Core/TextureCombiner/TextureCombiner.cs
```diff
@@ -241,7 +204,7 @@ public Texture2D Combine(string savePath)
             combinerMaterial.SetVector("_BRemap", m_remapings[2]);
             combinerMaterial.SetVector("_ARemap", m_remapings[3]);
 
-            RenderTexture combinedRT = new RenderTexture(xMin, yMin, 0, RenderTextureFormat.ARGBFloat, RenderTextureReadWrite.sRGB);
+            RenderTexture combinedRT = new RenderTexture(xMin, yMin, 0, GraphicsFormat.R32G32B32A32_SFloat);
```
FrancescoC-unity, line 207: So this is losing the sRGB qualifier here, can we do something or the format just doesn't have an equivalent?

com.unity.render-pipelines.high-definition/Editor/Sky/HDRISky/HDRISkyEditor.cs
```diff
@@ -81,7 +81,7 @@ public override void OnEnable()
             var hdrp = HDRenderPipeline.defaultAsset;
             if (hdrp != null)
                 m_IntegrateHDRISkyMaterial = CoreUtils.CreateEngineMaterial(hdrp.renderPipelineResources.shaders.integrateHdriSkyPS);
-            m_ReadBackTexture = new Texture2D(1, 1, TextureFormat.RGBAFloat, false, false);
+            m_ReadBackTexture = new Texture2D(1, 1, GraphicsFormat.R32G32B32A32_SFloat, TextureCreationFlags.None);
```
FrancescoC-unity, line 84: Before this was not linear, now seems it is, again as before do we have an option?

com.unity.render-pipelines.high-definition/Editor/Core/TextureCombiner/TextureCombiner.cs
JulienIgnace-Unity: In this kind of cases I assumed the sRGB flag was a mistake as float formats are always linear anyway.

FrancescoC-unity: I wondered if before we did some weird conversion internally, but yes I guess we are good like this

FrancescoC-unity (approved):

PR #3394, opened 2021-02-04
Bugfix 1304162: When importing a shader graph make sure to log any er…

joshua-davis:
…rors as some were getting dropped.

### Checklist for PR maker

---
### Purpose of this PR
Some errors in shader graph weren't getting reported at all. These errors aren't in a node so they can't be detected earlier. The errors can be fetched via ShaderUtils though, so on import grab and report the errors. 

---
### Testing status
Create a virtual texture property and give it an invalid layer reference name (via pasting not typing). On saving the shader graph this should print an error to the console.

---
### Comments to reviewers
Unfortunately, there doesn't seem to be a way to limit the console message so it only shows up once. That means this will print a message every time you hit save, but this seems better than not printing anything. In light of this, the preview was also left alone as it was deemed to be too noise if it logged every time.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Newly created properties and keywords will no longer use obfuscated GUID-based reference names in the shader code [1300484]
 - Fixed ParallaxMapping node compile issue on GLES2
 - Fixed a selection bug with block nodes after changing tabs [1312222]
+- Fixed some shader graph compiler errors not being logged [1304162].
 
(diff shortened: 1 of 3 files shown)


jessebarker (approved):
Once the formatting issue is resolved, this looks good to go.

ghost (approved):
The main concern that required this change has now been addressed with the line number added
image
image
It is now much easier for users to discover exactly what is going wrong and fix it.
There was talk about removing the "shadergraph has at least one warning" but it was discovered that deprecated nodes doesn't throw that, and a fix is in process to remove the warnings from the shadergraph that was causing the bug so ABV will pass. The failures were caused by the view direction node but I think that the warnings should be there because users need to be notified about the new behavior because under certain situations it will change the result of their graphs, so we decided to keep that in.

Everything else looks good. Approving.

PR #3396, opened 2021-02-04
Fix Sub-shadow rendering with cached shadows + Fix PCSS filtering issues on first frame or cached shadows

FrancescoC-unity:
This PR fixes two bugs: 

 - https://fogbugz.unity3d.com/f/cases/1308464/
 - Fix an issue with PCSS filtering being wrong on cached shadows. (Was already fixed in the past https://github.com/Unity-Technologies/Graphics/pull/1498/files but got undone with other changes) 

**What did I test**: the repro cases for both cases (old one linked in the other PR) + the PCSS issue verified with client.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

iM0ve (approved):
Verified that the customer case is solved


https://user-images.githubusercontent.com/36502659/108343457-a5054400-71e4-11eb-8f89-6ccaca4bc92b.mp4

PR #3405, opened 2021-02-05
[ShaderGraph][2021.2] Update rectangle node to better handle detail at a distance

jessebarker:
---
### Purpose of this PR
This PR addresses an issue where the rectangle node could generate visual artifacts at the edges of an object.
    - Existing method doesn't handle detail at a distance (large screen space derivatives) well
    - Add a new control to the rectangle node to switch between methods
    - Existing method is "fastest" and the new method is "nicest"
    - Test project uses default of "fastest", but users can pick "nicest"
      for best results.

---
### Testing status
Manual testing according to the original bug report using both methods now available on the node.
Yamato tests to come.

---
### Comments to reviewers
I'd like to determine whether the names I've chosen around the new control on the node are acceptable before updating documentation and the changelog.


cdxntchou (approved):

jessebarker:
Ok.  I've added documentation on the new control as well as a changelog entry for this.  Unless folks want to bikeshed on the name of the control or the description, this should be good to go.

alindmanUnity (approved):
UX approval

LandonTownsendUnity (approved):
Wasn't able to make the results of "nicest" diverge from "fastest" in any scenario, including very large or very small values, aside from fixing the white pixels caused by the bug. Tested unchanging UVs for NANs. Upgrading worked as expected.

cdxntchou:
> 
> 
> Wasn't able to make the results of "nicest" diverge from "fastest" in any scenario, including very large or very small values, aside from fixing the white pixels caused by the bug. Tested unchanging UVs for NANs. Upgrading worked as expected.

It should be significantly at distance (when the rectangle is very small on screen).

PR #3408, opened 2021-02-05
HDRP Graphic Test Fixes for Metal

brandon-unity3d:
### Purpose of this PR
The Desktop Graphic team is evaluating the graphic tests that are currently disabled in an effort to support the SRP teams through fixing issues that maybe occurring in the graphic device layer.  The list of disabled tests and their status can be found in the following google sheet.
https://docs.google.com/spreadsheets/d/11iY_cbOn9Mt1jba24295RU5zAnEAay1h/edit#gid=442307131

We found that many many of the metal/osx tests have been disabled therefore we shifted our focus to those graphic tests.

There were a couple of issues we encountered.  First, for some of the tests like 1900_AlphaTest_SG_a, the console would emit the following error message "[Error] Metal: Error creating pipeline state (Hidden/HDRP/Blit)".  This is because the render target asset that the shader's fragment shader writes to was of a UINT format however the fragment shader outputs a float4.  Therefore the asset needed to be in SFLOAT color format.

Secondly, we found several invalid graphic test images for several of the tests that were mostly black.  However, on other platforms they looked more correct.

---
### Testing status
We tested on Yamato as well as locally.  We ran a full ABV for all platforms covered by the ABV.

---
### Comments to reviewers

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/1550_3dsMax.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/1550_3dsMax.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:20deceb4023a54291028dac962d2569f84bf3b974d0ed0ab010d5ac81c4cbcd8
-size 168638
+oid sha256:f782e5a9bb710dcbf2d078e983b9e2baff4817537e155ad3de355ae7f3401dda
+size 150550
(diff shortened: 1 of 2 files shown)


sebastienlagarde (approved):

sebastienlagarde:
Thanks for looking at those issues! And thanks for the fixes of render targets:

>We found that many many of the metal/osx tests have been disabled therefore we shifted our focus to those graphic tests.
Note that a lot of test (in particular all the MotionVectorXXX) are instable. If you re-enable them all, the behavior we were observing is that "sometimes", some object disappear. If you run multiple time Yamato, you will get different error on different screenshots with missing objects. Pretty sure there is a synchronization / barrier issue but we only reproduce it on Yamato

NetteVI:
Thanks for the info Sebastien! We're keeping an eye on the previously disabled tests and seeing if they ever fail on our branch on Yamato, and we'll keep track to see if they still exhibit instabilities.

There's one issue with this PR currently in that the new 2501_LightLayers image for Metal is still not quite correct. It's related to an area lights on Catalina issue that is affecting about 10 other HDRP metal tests right now, so I'm working to resolve that issue in another commit first before moving this PR forward.

sebastienlagarde:
note: you need to merge master to be able to use the latest trunk which fix various issue, right now with current code you will still get a lot of error.

NetteVI (approved):

PR #3425, opened 2021-02-08
Fix undo preview with inline value

elizabeth-legros:
There was a bug that would cause PreviewManager to incorrectly use an old version of a node for populating preview property values. Fix by changing how we store our node to preview lookup, and rearranging our order of operations in handling graph changes.
https://fogbugz.unity3d.com/f/cases/1311862#BugEvent.1072930741
Tested undoing inline values, edge connections, and the specific bug case


cdxntchou (approved):

ghost (approved):
I did a bit of exploratory testing and it seems fine, along with fixing the bug. Approving.

PR #3428, opened 2021-02-09
Fixing resolution problems with CAS. CAS now renders exactly the reso…

kecho:
Fixing Contrast Adaptive Sharpening issues with Software & Hardware DRS 
* When CAS was applied in final, it was skipping samples due to the fact that positionSS was being clamped as an integer. This caused heavy pixelation and aliasing artifacts.
* The wrong size was being passed into CAS, and the wrong RTScaleHandle was being applied. This was causing the source resolution to be much higher, if there is a max/min resolution.
* Removing very weird code around hardware drs. During finalpass, if hw drs is on, the previous scale was used. This scale was sometimes garbage. The assumption from the comment is that during hw drs we want to just blit the entire buffer, to avoid viewport artifacts around edges. I disabled bias scaling during UI blit for this reason.

Comparison CAS

Another comparison of hardware vs software before these changes:
image

And a comparison of hardware and software after the fixes, they now match as expected:
image

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
+++ b/com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
@@ -109,6 +109,24 @@ namespace UnityEngine.Rendering
             }
         }
 
+        public Vector2 GetResolvedScale()
+        {
+            if (!m_Enabled || !m_CurrentCameraRequest)
+            {
+                return new Vector2(1.0f, 1.0f);
+            }
+
+            float scaleFractionX = m_CurrentFraction;
+            float scaleFractionY = m_CurrentFraction;
+            if (!m_ForceSoftwareFallback && type == DynamicResolutionType.Hardware)
+            {
+                scaleFractionX = ScalableBufferManager.widthScaleFactor;
+                scaleFractionY = ScalableBufferManager.heightScaleFactor;
+            }
+
+            return new Vector2(scaleFractionX, scaleFractionY);
+        }
+
         /// <summary>
         /// Set the scaler method used to drive dynamic resolution.
         /// </summary>
@@ -234,6 +252,7 @@ namespace UnityEngine.Rendering
 
         /// <summary>
         /// Applies to the passed size the scale imposed by the dynamic resolution system.
+        /// Note: this function has the side effect of caching the last scale size.
         /// </summary>
         /// <param name="size">The starting size of the render target that will be scaled by dynamic resolution.</param>
         /// <returns>The parameter size scaled by the dynamic resolution system.</returns>
@@ -246,21 +265,26 @@ namespace UnityEngine.Rendering
                 return size;
             }
 
-            float scaleFractionX = m_CurrentFraction;
-            float scaleFractionY = m_CurrentFraction;
-            if (!m_ForceSoftwareFallback && type == DynamicResolutionType.Hardware)
-            {
-                scaleFractionX = ScalableBufferManager.widthScaleFactor;
-                scaleFractionY = ScalableBufferManager.heightScaleFactor;
-            }
+            Vector2Int scaledSize = ApplyScalesOnSize(size);
+            m_LastScaledSize = scaledSize;
+            return scaledSize;
+        }
 
-            Vector2Int scaledSize = new Vector2Int(Mathf.CeilToInt(size.x * scaleFractionX), Mathf.CeilToInt(size.y * scaleFractionY));
+        /// <summary>
+        /// Applies to the passed size the scale imposed by the dynamic resolution system.
+        /// Note: this function is pure (has no side effects), this function does not cache the pre-scale size
+        /// </summary>
+        /// <param name="size">The size to apply the scaling</param>
+        /// <returns>The parameter size scaled by the dynamic resolution system.</returns>
+        public Vector2Int ApplyScalesOnSize(Vector2Int size)
+        {
+            Vector2 resolvedScales = GetResolvedScale();
+            Vector2Int scaledSize = new Vector2Int(Mathf.CeilToInt(size.x * resolvedScales.x), Mathf.CeilToInt(size.y * resolvedScales.y));
             if (m_ForceSoftwareFallback || type != DynamicResolutionType.Hardware)
             {
                 scaledSize.x += (1 & scaledSize.x);
                 scaledSize.y += (1 & scaledSize.y);
             }
-            m_LastScaledSize = scaledSize;
 
             return scaledSize;
         }
--- a/com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
+++ b/com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
@@ -219,8 +219,12 @@ namespace UnityEngine.Rendering
 
             if (DynamicResolutionHandler.instance.HardwareDynamicResIsEnabled() && m_HardwareDynamicResRequested)
             {
-                float xScale = (float)DynamicResolutionHandler.instance.finalViewport.x / GetMaxWidth();
-                float yScale = (float)DynamicResolutionHandler.instance.finalViewport.y / GetMaxHeight();
+                Vector2Int maxSize = new Vector2Int(GetMaxWidth(), GetMaxHeight());
+                // Making the final scale in 'drs' space, since the final scale must account for rounding pixel values.
+                var scaledFinalViewport = DynamicResolutionHandler.instance.ApplyScalesOnSize(DynamicResolutionHandler.instance.finalViewport);
+                var scaledMaxSize = DynamicResolutionHandler.instance.ApplyScalesOnSize(maxSize);
+                float xScale = (float)scaledFinalViewport.x / (float)scaledMaxSize.x;
+                float yScale = (float)scaledFinalViewport.y / (float)scaledMaxSize.y;
                 m_RTHandleProperties.rtHandleScale = new Vector4(xScale, yScale, m_RTHandleProperties.rtHandleScale.x, m_RTHandleProperties.rtHandleScale.y);
             }
             else
--- a/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
@@ -164,15 +164,20 @@ namespace UnityEngine.Rendering.HighDefinition
             public CustomPostProcessVolumeComponent customPostProcess;
         }
 
-        TextureHandle GetPostprocessOutputHandle(RenderGraph renderGraph, string name)
+        TextureHandle GetPostprocessOutputHandle(RenderGraph renderGraph,  string name, bool dynamicResolution = true)
         {
-            return renderGraph.CreateTexture(new TextureDesc(Vector2.one, true, true)
+            return renderGraph.CreateTexture(new TextureDesc(Vector2.one, dynamicResolution, true)
             {
                 name = name,
                 colorFormat = m_ColorFormat,
                 useMipMap = false,
                 enableRandomWrite = true
             });
+        }
+
+        TextureHandle GetPostprocessUpsampledOutputHandle(RenderGraph renderGraph, string name)
+        {
+            return GetPostprocessOutputHandle(renderGraph, name, false);
         }
 
         void FillBloomMipsTextureHandles(BloomData bloomData, RenderGraph renderGraph, RenderGraphBuilder builder)
@@ -294,9 +299,9 @@ namespace UnityEngine.Rendering.HighDefinition
                     }
                     else
                     {
-                        passData.tmpTarget1024 = builder.CreateTransientTexture(new TextureDesc(1024, 1024, true, false)
+                        passData.tmpTarget1024 = builder.CreateTransientTexture(new TextureDesc(1024, 1024, false, false)
                             { colorFormat = GraphicsFormat.R16G16_SFloat, enableRandomWrite = true, name = "Average Luminance Temp 1024" });
-                        passData.tmpTarget32 = builder.CreateTransientTexture(new TextureDesc(32, 32, true, false)
+                        passData.tmpTarget32 = builder.CreateTransientTexture(new TextureDesc(32, 32, false, false)
                             { colorFormat = GraphicsFormat.R16G16_SFloat, enableRandomWrite = true, name = "Average Luminance Temp 32" });
 
                         builder.SetRenderFunc(
@@ -320,7 +325,7 @@ namespace UnityEngine.Rendering.HighDefinition
                         passData.prevExposure = builder.ReadTexture(renderGraph.ImportTexture(GetPreviousExposureTexture(hdCamera)));
 
                         TextureHandle dest = GetPostprocessOutputHandle(renderGraph, "Apply Exposure Destination");
-                        passData.destination = builder.WriteTexture(dest);;
+                        passData.destination = builder.WriteTexture(dest);
 
                         builder.SetRenderFunc(
                             (ApplyExposureData data, RenderGraphContext ctx) =>
@@ -604,7 +609,7 @@ namespace UnityEngine.Rendering.HighDefinition
                     passData.nextMVLen = TextureHandle.nullHandle;
 
                     TextureHandle dest = GetPostprocessOutputHandle(renderGraph, "Post-DoF TAA Destination");
-                    passData.destination = builder.WriteTexture(dest);;
+                    passData.destination = builder.WriteTexture(dest);
 
                     builder.SetRenderFunc(
                         (TemporalAntiAliasingData data, RenderGraphContext ctx) =>
@@ -678,7 +683,7 @@ namespace UnityEngine.Rendering.HighDefinition
                     }
 
                     TextureHandle dest = GetPostprocessOutputHandle(renderGraph, "Motion Blur Destination");
-                    passData.destination = builder.WriteTexture(dest);;
+                    passData.destination = builder.WriteTexture(dest);
 
                     builder.SetRenderFunc(
                         (MotionBlurData data, RenderGraphContext ctx) =>
@@ -863,8 +868,8 @@ namespace UnityEngine.Rendering.HighDefinition
                 {
                     passData.source = builder.ReadTexture(source);
                     passData.parameters = PrepareContrastAdaptiveSharpeningParameters(hdCamera);
-                    TextureHandle dest = GetPostprocessOutputHandle(renderGraph, "Contrast Adaptive Sharpen Destination");
-                    passData.destination = builder.WriteTexture(dest);;
+                    TextureHandle dest = GetPostprocessUpsampledOutputHandle(renderGraph, "Contrast Adaptive Sharpen Destination");
+                    passData.destination = builder.WriteTexture(dest);
 
                     passData.casParametersBuffer = builder.CreateTransientComputeBuffer(new ComputeBufferDesc(2, sizeof(uint) * 4) { name = "Cas Parameters" });
 
--- a/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
@@ -3248,6 +3248,10 @@ namespace UnityEngine.Rendering.HighDefinition
             public int initKernel;
             public int mainKernel;
             public int viewCount;
+            public int inputWidth;
+            public int inputHeight;
+            public int outputWidth;
+            public int outputHeight;
         }
 
         CASParameters PrepareContrastAdaptiveSharpeningParameters(HDCamera camera)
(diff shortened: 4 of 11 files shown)


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
```diff
@@ -83,7 +84,7 @@ Shader "Hidden/HDRP/FinalPass"
             UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
             float2 positionNDC = input.texcoord;
-            uint2 positionSS = input.texcoord * _ScreenSize.xy;
+            float2 positionSS = input.texcoord * _ScreenSize.xy;
```
kecho, line 87: This is the gnarliest problem: uint2 was basically destroying any interpolation between samples, since _ScreenSize is much smaller than the viewport of this pixel shader.

sebastienlagarde (commented):
Please add a changelog. Let Francesco comment as he is more familiar with this code than me.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -3454,9 +3468,8 @@ FinalPassParameters PrepareFinalPass(HDCamera hdCamera, BlueNoise blueNoise, boo
             {
                 if (dynResHandler.HardwareDynamicResIsEnabled())
                 {
-                    var scaledSize = dynResHandler.GetLastScaledSize();
-                    backBufferRect.width = scaledSize.x;
-                    backBufferRect.height = scaledSize.y;
+                    backBufferRect.width = destinationHandle.rt.width;
```
FrancescoC-unity, line 3471: This is not always correct. This is used only when PostProcessIsFinalPass is false, that means that the texture is created using our own system.
The HW dynamic res doesn't necessarily need to render to the final resolution of the destination target. Imagine this scenario: 

- Camera 0 (e.g. main camera): has a final resolution of 4k, Dynamic Res at 50%. Then this code is fine as it sets to the desired final res. 
- Camera 1 (e.g. realtime reflection probe): has a final resolution of 512x512. Destination now is 4k as that is the biggest resolution found (given how RTHandle works), then this code will set the viewport to the full 4k, having then wrong UVs calculated. 

The code that was there before looks wrong to me (maybe I am forgetting something @JulienIgnace-Unity  ?), but a solution I think could be just getting the output from final resolution (queriable from the dyn res handler) instead of the buffer size.

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
```diff
@@ -219,8 +219,10 @@ public void SetReferenceSize(int width, int height, MSAASamples msaaSamples, boo
 
             if (DynamicResolutionHandler.instance.HardwareDynamicResIsEnabled() && m_HardwareDynamicResRequested)
             {
-                float xScale = (float)DynamicResolutionHandler.instance.finalViewport.x / GetMaxWidth();
-                float yScale = (float)DynamicResolutionHandler.instance.finalViewport.y / GetMaxHeight();
+                var scaledFinalViewport = DynamicResolutionHandler.instance.GetScaledSize(DynamicResolutionHandler.instance.finalViewport);
+                var scaledMaxViewport = DynamicResolutionHandler.instance.GetScaledSize(new Vector2Int(GetMaxWidth(), GetMaxHeight()));
```
FrancescoC-unity, line 223: This is a bit dangerous as it will cache the original size as GetMaxWidth which is not necessarily what we want. 

Is this to have the ceiling applied before the scale calculation? Because if not outside for the side effect of caching the original scale this is equivalent to 

finalViewport * fraction / maxViewport * fraction == finalViewport / maxViewport which as there before

FrancescoC-unity: If it is indeed what you wanted to do (minus the caching), we might want to add a new API to get the same data without side effect.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -3348,12 +3357,14 @@ FinalPassParameters PrepareFinalPass(HDCamera hdCamera, BlueNoise blueNoise, boo
...
             RTHandle                afterPostProcessTexture,
-            RenderTargetIdentifier  destination,
+            RTHandle                destinationHandle,
             RTHandle                alphaTexture,
             CommandBuffer           cmd)
         {
+
             // Final pass has to be done in a pixel shader as it will be the one writing straight
             // to the backbuffer eventually
+            RenderTargetIdentifier destination = destinationHandle;
```
FrancescoC-unity, line 3367: No need to do this intermediate step, you can get the rt directly from RTHandle

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
FrancescoC-unity: Indeed shouldn't have been used by the CAS part, but everything else that uses positionSS  should still be uint2 as we use texture loads with it and never samples.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
```diff
@@ -860,7 +860,13 @@ TextureHandle ContrastAdaptiveSharpeningPass(RenderGraph renderGraph, HDCamera h
                 {
                     passData.source = builder.ReadTexture(source);
                     passData.parameters = PrepareContrastAdaptiveSharpeningParameters(hdCamera);
-                    TextureHandle dest = GetPostprocessOutputHandle(renderGraph, "Contrast Adaptive Sharpen Destination");
+                    TextureHandle dest = renderGraph.CreateTexture(new TextureDesc(Vector2.one, false, true)
```
FrancescoC-unity, line 863: I am wondering if at this point we should allocate this texture with explicit texture size rather than a scale of 1 ( matching the _OutputTextureDimensions passed to CAS )

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
kecho: Not needed anymore, indeed.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: I tried this with realtime reflection probes. The reason why it works there is because DRS is disabled. This code path only gets executed if DRS is on, DRS is hardware mode, and we have a pass after FinalPass (blit into some UI). The reason for this from what I understand is to reduce complexity by using the entire resolution of the target. I could remove this altogether though, and achieve the same results. Any reason why we'd want to keep this code? Maybe @JulienIgnace-Unity  can answer?

FrancescoC-unity: > probes. The reason why it works there is because DRS is disabled

Yes, maybe my example was bad, but same thing can happen with any secondary camera that uses dyn res too (as long we have dyn res on two things with different target final resolutions we'll have problems)

kecho: Actually, @FrancescoC-unity  this is completely unnecesary once I fixed the side effect issue below :)

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
kecho: It is indeed what we want. The reason why that equality you typed is not the same is because of rounding:

round(finalViewport * fraction) / round(maxViewport * fraction) != finalViewport/maxViewport
The rations need to consider the partial pixel of the downsampled resolution, otherwise you get some weird lines in the edges.
Hope that made sense, I am now writting proper pure functions for those APIs. The side effect did cause some issues, thank you for pointing that out!

FrancescoC-unity: Yup wrote in the comment that if it was rounding then yeah then we want (The equation was an _if not_ ) :D 

But yeah, we probably want something without side effect, shame we'd need to make that public. So I'd suggest to have an helper to just get the fraction and then something internal to HDRP only to do the same scaling as the handler.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: Ok, using finalViewport works like a charm. But wouldnt this be equivalent of deleting this entire block of code? (the backbuffer rect is set to the finalViewport from the camera, which is not the DRS viewport, is the proper final viewport).
It gets set on line 572 of HDCamera.cs
            DynamicResolutionHandler.instance.finalViewport = new Vector2Int((int)finalViewport.width, (int)finalViewport.height);

FrancescoC-unity: I am indeed a bit confused on why that code is there to start with :/ Waiting on the Julien summon he might know why.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
kecho: hmm maybe. Gotta test with Software resolution, and ensuring that in dx11 we dont reallocate a texture.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
kecho: This has been removed in new version

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: ok, on the meantime slapped a TODO there. Hopefully we can nuke this whole block and simplify things a bit :)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: Removed

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
kecho: Added some new APIS. LMK what you think.

kecho:
> Please add a changelog. Let Francesco comment as he is more familiar with this code than me.

Updated the changelog !

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -274,6 +282,29 @@ public float GetCurrentScale()
             return (m_Enabled && m_CurrentCameraRequest) ? m_CurrentFraction : 1.0f;
         }
 
+        /// <summary>
+        /// Gets the resolved scale percentages.
+        /// Note: this function is pure, and has no side effects.
+        /// </summary>
+        /// <returns>The resolved drs scales (from 0 to 1, for x and y). These might change depending on what the hardware supports.</returns>
```
FrancescoC-unity, line 289: In documentation we should avoid acronyms (Dynamic Resolution here instead of DRS)

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -274,6 +282,29 @@ public float GetCurrentScale()
...
+        public Vector2 GetResolvedScale()
```
FrancescoC-unity, line 290: Do we need this API as public? The less we have public the better is for us as it will allow us to easily modify with time. I think with the no-side effect scale thingie we don't need this as public

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -246,21 +247,28 @@ public Vector2Int GetScaledSize(Vector2Int size)
...
+            m_LastScaledSize = scaledSize;
+            return scaledSize;
+        }
 
-            Vector2Int scaledSize = new Vector2Int(Mathf.CeilToInt(size.x * scaleFractionX), Mathf.CeilToInt(size.y * scaleFractionY));
+        /// <summary>
+        /// Applies to the passed size the scale imposed by the dynamic resolution system.
+        /// Note: this function is pure (has no side effects).
+        /// </summary>
+        /// <param name="resolvedScales">Scale amount to use. Optionally, could utilize GetScaledSize function to get the scales desired, but this function is safer
```
FrancescoC-unity, line 259: Not sure about the safer, as the idea is that if a custom SRP uses the system, then the GetScaleSize side effect is needed to keep the system functioning as long as it is called at the right time. 

Also no parameter here is resolvedScales in the function signature.  

I'd expand the API doc here to say that this doesn't cache the last scale size or the pre-scale size (i.e. the cachedOriginalSize )

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
FrancescoC-unity: On second thought it is indeed probably not a good idea, let's keep like this

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: private now, done!

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: fixed documentation. Done!

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: Fixed this! done

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
```diff
@@ -294,9 +294,9 @@ TextureHandle DynamicExposurePass(RenderGraph renderGraph, HDCamera hdCamera, Te
                     }
                     else
                     {
-                        passData.tmpTarget1024 = builder.CreateTransientTexture(new TextureDesc(1024, 1024, true, false)
+                        passData.tmpTarget1024 = builder.CreateTransientTexture(new TextureDesc(1024, 1024, false, false)
```
FrancescoC-unity, line 297: I am _sure_ this was not set to do hw dyn res to start with, so it makes me think lots of stuff got messed up during the render graph transition; let me know if you want an extra pair of eyes but I think we need to do a pass on all the new TextureDesc;

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
JulienIgnace-Unity: As far as I can tell, the viewport size should indeed always be finalViewport here (as we want anything rendered after post processes at full rez. Debug modes mostly). I don't think it should be destinationHandle.rt.width though as it would not work with 2 cameras using dynamic rez and different output resolution).
The other important thing here is setting x,y to 0,0 since all subsequent debug assumes viewports starts at (0,0)

FrancescoC-unity: Let's nuke those lines then!

kecho: 💪

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
kecho: So there are two ways of fixing it for hardware:
1) We disable DRS on this target, and carry on. This would make exposure match %100 software.
2) We keep this as a DRS target, but we have to change the code internally, and also the dispatch size must match the size of this target (its hard coded to 1024, which causes just a piece of the frame to be put in the exposure input).

Let me kow your thoughts. I like option # 1 better, because exposure remains consistenish between software and hardware DRS

FrancescoC-unity: It should be 1 and it was 1 before rendergraph I am sure :-) 

Hence why I am saying it might be that we might have made similar port issues from before rendergraph to rendergraph (and we should make a pass to check if it is indeed the case)

kecho: 💪

JulienIgnace-Unity: Damn, the carnage continues!
It's not my fault, it's the reviewers :p

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
```diff
@@ -164,16 +164,29 @@ class CustomPostProcessData
             public CustomPostProcessVolumeComponent customPostProcess;
         }
 
-        TextureHandle GetPostprocessOutputHandle(RenderGraph renderGraph, string name)
+        TextureHandle GetPostprocessOutputHandle(TextureHandle previousTexture, RenderGraph renderGraph, string name)
```
FrancescoC-unity, line 167: I am not so sure on this. Both this and the GetPostprocessUpsampledOutputHandle are identical if not for the hardware dynamic res is enabled. 

I think I'd rather have that being an explicit parameter rather than relying on the source. 

So something like 


GetPostprocessOutputHandle(RenderGraph renderGraph, string name, bool enableHWDynamicRes) 

and then caller can be explicit like 

TextureHandle dest = GetPostprocessOutputHandle(renderGraph, "Motion Blur Destination", enableHWDynamicRes: true);

and 

                    TextureHandle dest = GetPostprocessOutputHandle(renderGraph, "Contrast Adaptive Sharpen Destination", enableHWDynamicRes: true);

FrancescoC-unity: We can still have GetPostprocessUpsampledOutputHandle as a verbose wrapper around  GetPostprocessOutputHandle(renderGraph, name, enableHWDynamicRes: false)

kecho: done!

kecho:
Running universal and hdrp tests

FrancescoC-unity (approved):

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
@kecho . Please add in the What's new 12.0 .md file (in Documentation~ folder) (in the improvement section the change. i.t it is for the users to be aware that there is an improvement of the quality of dynamic res. Thanks

kecho:
> @kecho . Please add in the What's new 12.0 .md file (in Documentation~ folder) (in the improvement section the change. i.t it is for the users to be aware that there is an improvement of the quality of dynamic res. Thanks

Done. Let me know if the grammar / wording is ok

sebastienlagarde:
Added @JordanL8  for doc review.

JordanL8 (approved):
Reworded a few bits on the what's new page. Let me know if I misunderstood anything

kecho:
> Reworded a few bits on the what's new page. Let me know if I misunderstood anything

Wow very nice job. Clear as water!

TomasKiniulis (approved):
Looks solid checking from my side, Contrast Adaptive Sharpen lives up to its name with this PR! Compared between master in Amelienborg palace project PR results do look clearer and less washed out.

What I checked:
- Multiple games views with different resolutions sets
- Made sure other upscale filters are still working correctly
- Checked that there's no significant differences switching Dynamic Resolution Type between Hardware and Software
- Checked with Force Screen Percentage Enabled disabled in Playmode and Edit Mode
- Made sure that switching between multiple HDRP assets with different Dynamic Res settings does not introduce issues
- Contrast Adaptive Sharpen filter works correctly in Player build

Master:
master
PR:
PR

PR #3442, opened 2021-02-09
Fix to reduce the amount of allocations required when using MSVO and Dynamic Resolution

robinb-u3d:
### Purpose of this PR
This is a port of from the original Post Processing repo https://github.com/Unity-Technologies/PostProcessing/pull/901 now that Unity has been fixed so that we can make these changes. See the CHANGELOG.md file changes for a description of the version requirements to make use of this bugfix. Fixes case 1285605.

The existing dynamic resolution strategy manually scales the temporary intermediate textures to the correct size required for the current dynamic resolution scale. This allows the effect to run correctly and for a given fixed scaling factor will keep the memory usage constant however if dynamic resolution is being scaled over time then every time it changes scaling factor the temporary intermediate textures will be reallocated causing an increase in memory pressure while this happens and possibly generating more VRAM fragmentation (depending on the platform).

This PR make the temporary intermediate textures be full size dynamic resolution textures themselves. The behavior of this PR and the existing behavior should match when dynamic resolution is not being used. However when dynamic resolution is being used this PR changes the following:

We now allocate the temporary intermediate textures at full size all the time so for a fixed dynamic resolution scaling of 0.5 this PR will increase VRAM usage as we will now allocate all textures at 1.0 size and then make sure of dynamic resolution to access a 0.5 scaled version using the same memory.
The temporary intermediate textures are now dynamic, this means they may have less chance of matching other temporary textures used by other systems and so increase the memory usage of this project.
Changing the dynamic scaling factor will no longer cause these textures to be reallocated thus VRAM usage should be consistent no matter what level of dynamic scaling factor is used and reduce memory fragmentation and allocation overheads.
Fixing this behavior showed up a bug in how Unity handles viewport setting when using dynamic resolution on some render targets on dynamic resolution platforms. This PR requires a fixed Unity version to work correctly otherwise the temporary intermediate textures are not rendered to correctly when the scale is not 1.0. These bugfixes have now landed into Unity see the CHANGELOG.md file.

---
### Testing status
Testing with the reproduction project from case 1285605 on the original platform and monitoring the VRAM memory usage shows that the fixed version has a constant VRAM usage with the same visual effect compared against Post Processing V2 3.0.1.


Chman (approved):
We'll add the post-processing gfx tests back next week to make sure everything is alright but it's looking good to me.

JoJo2nd (approved):

robinb-u3d:
@Chman do you want me to wait for you to add the tests and have them pass before merging this in?

Chman:
I would prefer we wait for the tests, yes. We'll try to get them up and running sometimes this week.

Chman:
Yes, tests are in a branch that we're currently testing but we ran into some issues with Yamato etc so it's taking longer than it should :(

PR #3444, opened 2021-02-10
[XPipeline]Unify Material editors sections on both pipelines

alex-vazquez:
---
### Purpose of this PR

Using the same Material Headers on URP as HDRP is using. This unifies the right spacing to fix this [issue](https://jira.unity3d.com/browse/HDRP-492)

image

---
### Testing status
- Test that state (open/collapsed) are properly stored.
- Test that advanced is collapsed by default

---
### Comments to reviewers
Move some code to Core from HDRP and using it on URP


martint-unity (approved):

com.unity.render-pipelines.core/Editor/Material/MaterialHeaderScopeList.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Draws all the <see cref="MaterialHeaderScopeItem"/> with its information stored
+        /// </summary>
+        /// <param name="materialEditor"><see cref="MaterialEditor"/></param>
+        /// <param name="material"><see cref="Material"/></param>
+        public void DrawHeaders(MaterialEditor materialEditor, Material material)
+        {
+            if (material == null)
+                throw new ArgumentNullException(nameof(material));
```
jenniferd-unity, line 43: should it be an exception or a regular error handling (console print)?

alex-vazquez: In this case I think it makes more sense to throw exceptions, because is a developer error.

sebastienlagarde (approved):

alex-vazquez:
@Unity-Technologies/gfx-qa-hdrp . Quick look at HDRP materials to make sure they look the same, no crashes and no regressions
@Unity-Technologies/gfx-qa-foundation . Check URP materials and that they behave in the same way as before, but with the new headers.

Note: Previous collapsabled states  saved with another version will not be recovered as the key has changed.

simon-engelbrecht-soerensen (approved):
Seems to be working as expected in URP.

remi-chapelain (approved):
Took a quick look at the material samples in HDRP Classic & SG
Everything looks the same in editor and inspector, nothing unexpected
Approving ✔️

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayerListUIBlock.cs
```diff
@@ -39,28 +39,7 @@ internal class Styles
...
-            if (layerNumber == 4)
-            {
-                materialEditor.SetExpandedAreas((uint)ExpandableBit.ShowLayer3, true);
-            }
-            if (layerNumber >= 3)
-            {
-                materialEditor.SetExpandedAreas((uint)ExpandableBit.ShowLayer2, true);
-            }
-            materialEditor.SetExpandedAreas((uint)ExpandableBit.ShowLayer1, true);
-        }
```
RSlysz, line 63: Be careful with this. You require extend test with HDRP/LayeredLit and HDRP/LayeredLitTesselation
This ShowLayerX are not for expended but for reallly showing the LayerX

RSlysz (changes requested):
Do thorough check of layered material in HDRP. There is high probability that this PR break them.

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayerListUIBlock.cs
alex-vazquez: This code was only called in the setter, and the setter was never being called, besides that the code for a layer number 4 will expand to `ShowLayer2` as the condition is wrong.

RSlysz: Truth. Code should have changed

RSlysz (approved):

PR #3453, opened 2021-02-10
[2021.2] SSAO: Fixing an issue where sometime the pass would not contain the material

ellioman:
# Purpose of this PR
Sometimes after importing a renderer with the SSAO renderer feature, the SSAO would not run.
The reason is that the Pass never received the material even though the feature had it.

The fix is to always send the material to the pass so we make sure it's always up do date.

# Testing status
Tested locally with a repro project.


Jonasmortensen (approved):
This fixes issues with SSAO not working for imported package samples.

svens-unity (approved):

ellioman:
Last commit was a fix for a changelog conflict. It was green before that so I cancelled the jobs.

PR #3457, opened 2021-02-10
Fixed an issue with first frame of ao causing some kind of ghosting e…

JulienIgnace-Unity:
### Purpose of this PR
This PR fixes an issue where first frame of AO could exhibit some kind of ghosting artefacts.
This was caused by the depth buffer mip chain info being computed the wrong way in some cases.

Repro steps:
- Load a scene with AO (HDRP_Test AO test scene is a good example)
- Set it to half resolution with temporal enabled
- Show the AO in the scene view (with a different angle than game view)
- Enter playmode and exit.
- Notice the ghosting artefacts on AO

Ao_firstFrame

---
### Testing status
Tested the repro step above and ran automated tests.


FrancescoC-unity (approved):

iM0ve (approved):
Looks safe. I also quickly loaded up a project to change some AO values to see it still works as expected

sebastienlagarde (approved):

sebastienlagarde:
@JulienIgnace-Unity  missing changelog

PR #3462, opened 2021-02-10
Clean up shader ui

adrien-de-tocqueville:
### Purpose of this PR

This PR cleans up and removes a lot of duplicate code in the HDRP shader inspectors
The aim is mainly to prepare the material variants work.

It also fixes some issues, mainly with the emission intensity in the Lit shader and decal shader
- when multi editing with mixed units, inputs are disabled
- when enabling/disabling "use emission intensity", color is convert from hdr <-> ldr space (previously switching on then off would make you loose the data you input)

I also changed the behavior of the miminim and maximum steps of pixel displacement lit shader.
Previously putting a maximum lower than mimum changed the minimum
And minimum could not go above maximum
Now putting a minimum above maximum also raises the maximum
The min and max step field were previously float values, I changed them to integer input because it makes more sense

---
### Testing status
Tested the modified properties in the inspector

---
### Comments to reviewers
There are maybe some strange corner cases with multiediting of the emission properties

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/EmissionUIBlock.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/EmissionUIBlock.cs
@@ -146,13 +146,15 @@ namespace UnityEditor.Rendering.HighDefinition
             bool intensityIsMixed = unitIsMixed || emissiveIntensity.hasMixedValue;
 
             using (new EditorGUILayout.HorizontalScope())
-            using (new EditorGUI.DisabledScope(unitIsMixed))
             {
                 EditorGUI.showMixedValue = intensityIsMixed;
                 EmissiveIntensityUnit unit = (EmissiveIntensityUnit)emissiveIntensityUnit.floatValue;
 
                 if (unitIsMixed)
-                    materialEditor.ShaderProperty(emissiveIntensity, Styles.emissiveIntensityText);
+                {
+                    using (new EditorGUI.DisabledScope(unitIsMixed))
+                        materialEditor.ShaderProperty(emissiveIntensity, Styles.emissiveIntensityText);
+                }
                 else
                 {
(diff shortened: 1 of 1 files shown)


sebastienlagarde (commented):
This is great change, will need to review this more carefully. Also those change will not be catch by automatic test, so we need to be sure that QA carefully test it. Can you list the property affected by the cahnge that the QA should take care of? thanks

sebastienlagarde (approved):
Very good change thanks

For QA pay special attention to Emissive edition (try to go from Emissive intensity to HDR color, back and forth and undo. THanks

sebastienlagarde:
Adding Antoine for double check.

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/EmissionUIBlock.cs
```diff
@@ -122,7 +120,7 @@ public override void OnGUI()
             }
         }
 
-        void UpdateEmissiveColorFromIntensityAndEmissiveColorLDR()
+        public static void UpdateEmissiveColorFromIntensityAndEmissiveColorLDR(MaterialEditor materialEditor, Material[] materials)
```
alelievr, line 123: missing API doc

alelievr (approved):

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/EmissionUIBlock.cs
adrien-de-tocqueville: nice catch, I changed them to internal because they shouldn't be public anyway

JMargevics (approved):
I've imported our material samples and played with all shader types. No issues or broken properties ✅
Also checked the fixes:

Displacement Min/Max fix. Tried entering floats as well :) 
ucO8xvGBQC

Can't multi-edit emission values
image

Units are converted correctly.
RB2Qc1nasa

PR #3465, opened 2021-02-11
Vulkan Standalone Player MSAA to Backbuffer Fix

manuele-bonanno:
### Purpose of this PR
MSAA to backbuffer (no intermediate passes, so no CopyColor/HDR/PostProc) is broken on Vulkan Windows Standalone? The code handling the explicit resolve (no intermediate ResolveAA pass) has been there for quite a long time.
Editor won't show this issue because of the extra FinalBlit to SceneView/PlayView render targets, so no BB involved there

As a fix, we force the ExplicitResolve on non-mobile platforms which have multisampleAutoresolve = true (all Vulkan platforms except Android)

---
### Testing status
local testing: Vulkan Editor, Vulkan WinPlayer, Android Vulkan, Mac Editor, MacPlayer, iOS Player
Yamato: ABV

---
### Comments to reviewers
To reproduce make sure MSAA is enabled and no intermediate passes are added and Post Processing is OFF.

image

Windows Standalone result before the fix:
image

Windows Standalone result after the fix:
image


ellioman (approved):

erikabar (approved):
tested Windows Vulkan and DX11 with Template and UniversalRenderingExamples projects, issue is fixed and no regressions found.

PR #3281, opened 2021-01-28
Custom Interpolator Feature

esmelusina:
### Purpose of this PR
Allows users to make, use, and modify custom vertex-to-fragment interpolators through ShaderGraph.

---
### Testing status
Thoroughly tested by QA. A separate PR for graphics tests will be made, with haste to take advantage of ongoing improvements to our testing framework (and avoid blocking image refs).

---
### Comments to reviewers
If you are an HDRP or URP reviewer-- this is a big PR for ShaderGraph, feel free to review only the changes to your packages.

PR #3467, opened 2021-02-11
Refactor VFX / SG Integration (use any target)

johnpars:
### Purpose of this PR
Allow users to target VFX Graph with HDRP Master Node/Stack in ShaderGraph. Specifically it unlocks the following to users:
- Support any HDRP material type (StackLit, Hair, etc.) for any VFX Output Context.
- Support Vertex Stack in VFX (Replacing [this PR](https://github.cds.internal.unity3d.com/unity/vfx-graphics/pull/112)).
- Improve VFX Output Context UI by re-using Material UI blocks for surface option configuration.

https://user-images.githubusercontent.com/28882975/110500813-944f3c00-80c7-11eb-86c2-30bb24f98d8f.mp4

The refactor is done with the following technical goals in mind:
1) Consolidate the various VFX **.template** files into a single template, reducing code duplication and making VFX more easily maintainable if there are changes to HDRP shading architecture.
2) Phase out the need for HDRP's specific VFX Context implementations (i.e. the "Output Particle **Lit** Mesh" etc.). Output contexts themselves should only describe the primitive/task type and not also suggest the material type- this is the responsibility of the shader graph.
3) Use Shader Graph as the VFX shader generator and share as much information as possible with the render pipeline's (sub)target. This lets VFX automatically scale with HDRP whenever a new sub-target/material type is introduced. (Prior to this PR, the Visual Effect target would have to manually re-implement the work for each material type- and so far only supported lit/unlit).
4) Serialize the generated VFX material within the VFX asset. Prior to this PR the material was generated on the fly and all of the render state was statically known at compile time. The problem with this is VFX must re-implement all of the render state configuration itself. By serializing the material we can easily re-use the HDRP material inspector UI blocks which are largely responsible for configuring render state and surface options.
5) Keep the original VFX Shader Graph Integration. For now, this PR simply offers a new generation path, with plans for a future PR to support URP as well (for 21.2).  

### Usage
1) Ensure **Improved ShaderGraph Generation** is enabled in the Visual Effects Preferences panel. It should be enabled by default.

image

2) Select the new **Support VFX Graph** toggle in within the HDRP Target Graph Inspector and save the asset.

image

3) In VFX Graph, assign the VFX compatible shader graph into the **Shader Graph** field. Note that exposed fragment and vertex properties in the Shader Graph will be exposed in the Graph View and configurable per-particle (this is exactly how the current SG support works).

image

4) Configure the effect's render state & surface properties in the output context's inspector. Note that we now serialize materials within the VFX asset, and can re-use the material inspector generated by the shader graph. This allows us to holistically configure render state in HDRP between VFX and normal material renderers, and convey that information to the user in a familiar way.

image

---
### Testing 
From a user point of view, this PR introduces a slight adjustment in workflow.
From a quality/testing point of view this PR introduces various things that will require validation in various systems (HDRP, SG, VFX). 

- Manually verified that all HDRP sub targets compile for all VFX output types (mesh),
- Manually verified motion vectors for all output types.
- Manually verified support for vertex stack (and motion vectors for vertex stack + VFX animation at the same time).
- Manually verified render state configuration via the new Context UI's material inspector.
- Manually verified multiple output contexts in the same system.
- Manually verified the new workflow of toggling VFX support in graph, and linking to it from VFX.
- Manually verified support for the original Visual Target / integration.
- Made a standalone build
- [Test Document](https://docs.google.com/document/d/1DkYmO_sQsRaAaB0s4DI9QxXlDqmoK5jt1xs8WyE78Ig/edit?usp=sharing) 
- [Yamato](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fsg-vfx-integration-improved)
- Added new test **34_ShaderGraphGeneration** to VFX_HDRP. @VladNeykov plans to add more robust testing in the [FTP](https://github.cds.internal.unity3d.com/unity/Graphics-QA-FTP-VFX).

---
### Comments to reviewers
- Note that this PR does not support the HDRP Decal target. (though should, with some more work).
- I have left various specific remarks in the PR for discussion.
- ~This PR is dependent on an incoming C++ PR from the VFX team for VFX Material Serialization. In the meantime I have made a temporary workaround C++ branch that injects the material render state: https://ono.unity3d.com/unity/from-git/changelog?branch=graphics%2Fvfx%2Finject-material-state~
- The PR is dependent on this C++ PR to land the material serialization: https://ono.unity3d.com/unity/unity/pull-request/122141/_/graphics/vfx/feature/material-serialization
- This is a pretty big PR so please only review relevant areas to your domain if you'd like.
- This branch was made at the same time as Shader Graph Custom Interpolators (#3281 ). It's unclear how easy it would be to leverage that feature in this PR. Perhaps in a future PR we could use it.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
@@ -19,7 +19,6 @@ Pass
 
     // Pragmas
     $splice(PassPragmas)
-    #pragma enable_d3d11_debug_symbols
 
     // Keywords
     $splice(PassKeywords)
--- a/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VFX/ConfigMesh.template.hlsl
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VFX/ConfigMesh.template.hlsl
@@ -2,7 +2,7 @@
 
 bool GetMeshAndElementIndex(inout AttributesMesh input, inout AttributesElement element)
 {
-    uint index = input.instanceID;
+    uint index = VFX_GET_INSTANCE_ID(input);;
 
     if (ShouldCullElement(index))
         return false;


com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDSubTarget.cs
```diff
@@ -119,6 +121,9 @@ public override void Setup(ref TargetSetupContext context)
 
         protected SubShaderDescriptor PostProcessSubShader(SubShaderDescriptor subShaderDescriptor)
         {
+            if (VFXSubTarget.IsConfigured())
+                subShaderDescriptor = VFXSubTarget.PostProcessSubShader(subShaderDescriptor);
```
johnpars, line 125: Here is the main entry point in HDRP to invoke VFX to post process the current sub target.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDSubTarget.cs
```diff
@@ -193,7 +198,10 @@ public override void GetFields(ref TargetFieldContext context)
         {
             // Common properties between all HD master nodes
             // Dots
-            context.AddField(HDFields.DotsInstancing,      systemData.dotsInstancing);
+            context.AddField(HDFields.DotsInstancing, systemData.dotsInstancing);
+
+            // VFX Setup
+            VFXSubTarget.GetFields(ref context);
```
johnpars, line 204: VFX also needs to configure fields to enable the primary graph VFX include, as well as setup the necessary VFX output type.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDSubTarget.cs
```diff
@@ -91,8 +92,9 @@ public override void Setup(ref TargetSetupContext context)
         {
             context.AddAssetDependency(kSourceCodeGuid, AssetCollection.Flags.SourceDependency);
             context.AddAssetDependency(subTargetAssetGuid, AssetCollection.Flags.SourceDependency);
+            var inspector = VFXSubTarget.IsConfigured() ? VFXHDRPSubTarget.Inspector : customInspector;
             if (!context.HasCustomEditorForRenderPipeline(typeof(HDRenderPipelineAsset)))
-                context.AddCustomEditorForRenderPipeline(customInspector, typeof(HDRenderPipelineAsset));
+                context.AddCustomEditorForRenderPipeline(inspector, typeof(HDRenderPipelineAsset));
```
johnpars, line 97: In case of VFX generation we override the inspector with the one suited for VFX Context UI.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
```diff
@@ -31,7 +32,22 @@ Pass
...
+    // Omit the HDRP FragInput struct definition since VFX provides an arbitrary number of interpolators
+    // (this is because properties are evaluated by the graph on the fly in the vertex stage)
+    // These interpolates must be transfered into the FragInputs so they are available for the Surface evaluation.
+    // Thus we procedurally generate a FragInput struct.
+    // Warning: Omitting this means we no longer define the GetVaryingsDataDebug..
+    #ifndef HAVE_VFX_MODIFICATION
     #include "Packages/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/FragInputs.hlsl"
+    #else
+    // For now, move the full struct splice to this point (contains FragInputs).
+    $splice(PassStructs)
```
johnpars, line 48: Not very happy with how I currently move the entire struct splice up to this point, eager to hear thoughts about it.  This also might have unforeseen issues that will later be caught in the CI. 

Reminder: we must do this because FragInputs contains all of the incoming interpolator that got configured by VFX. Thus we must override FragInputs and define all of the structs here so that anything dependent on FragInputs can still get it.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
```diff
@@ -168,6 +187,7 @@ Pass
     // --------------------------------------------------
     // Build Graph Inputs
 
+	$features.graphVFX:     $include("VFX/Config.template.hlsl")
```
johnpars, line 190: Main entry to the VFX graph implementation

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/SharedCode.template.hlsl
```diff
@@ -17,6 +17,11 @@ FragInputs BuildFragInputs(VaryingsMeshToPS input)
     $FragInputs.texCoord3:          output.texCoord3 = input.texCoord3;
     $FragInputs.color:              output.color = input.color;
 
+    #ifdef HAVE_VFX_MODIFICATION
+    // FragInputs from VFX come from two places: Interpolator or CBuffer.
+    $splice(VFXSetFragInputs)
```
johnpars, line 22: Is it possible to re-use the new custom interpolators feature for this? VFX interpolants are basically just custom interpolators...

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VertexAnimation.template.hlsl
```diff
@@ -39,12 +39,16 @@ VertexDescriptionInputs AttributesMeshToVertexDescriptionInputs(AttributesMesh i
...
 
-// This is used for injecting the define below.
-$splice(CustomInterpolatorPreVertex)
+    // This is used for injecting the define below.
+    $splice(CustomInterpolatorPreVertex)
 
 
 
-    AttributesMesh ApplyMeshModification(AttributesMesh input, float3 timeParameters
+    AttributesMesh ApplyMeshModification(AttributesMesh input,
```
johnpars, line 47: Invocation of this call is getting pretty messy with both VFX and custom interpolants trying to get their info in

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VertexAnimation.template.hlsl
```diff
@@ -57,7 +61,17 @@ $splice(CustomInterpolatorPreVertex)
     $VertexDescriptionInputs.TimeParameters: vertexDescriptionInputs.TimeParameters = timeParameters;
 
     // evaluate vertex graph
+#if defined(HAVE_VFX_MODIFICATION)
+    GraphProperties properties;
```
johnpars, line 65: GraphProperties is defined once and zero initialized. The VFX Element just configures the necessary vertex properties which the vertex description function will use. This is very similar to the element pixel properties configuration.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDShaderUtils.cs
```diff
@@ -83,13 +83,19 @@ internal enum ShaderID
         /// False: unknown shader used in material
         /// </returns>
         public static bool ResetMaterialKeywords(Material material)
+        {
+            var shaderID = GetShaderEnumFromShader(material.shader);
+            return ResetMaterialKeywords(material, shaderID);
+        }
+
+        internal static bool ResetMaterialKeywords(Material material, ShaderID shaderId)
```
johnpars, line 91: A new internal version of this function, so that VFX can manually provide the ShaderID from the current Shader Graph asset it is using. The original implementation of this function would fail to configure the material for VFX, since none of that meta data is stored in the VFX-generated shader.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
```diff
@@ -107,6 +107,21 @@ PackedVaryingsType MotionVectorVS(inout VaryingsType varyingsType, AttributesMes
...
+        // Re-construct the VFX mesh (i.e. in case of procedural output like planar primitives).
+        // And fetch element index to sample the previous element matrix.
+        AttributesElement element;
+        ZERO_INITIALIZE(AttributesElement, element);
+
+        GetMeshAndElementIndex(inputMesh, element);
+
+        // NOTE: We have to re-evaluate the attributes here, not good.
+        // TODO: organize things so that VFX mesh and attributes get computed once.
+        GetElementData(element);
```
johnpars, line 120: MotionVectorVS requires us to re-evaluate all of the VFX attributes + graph information (needless cost on performance) to support motion vectors for VFX animation + vertex stack. This happens because currently none of the work done in VertMesh to configure the mesh gets transferred here. This PR should not merge until we confirm how we do that.

com.unity.shadergraph/Editor/Data/Graphs/AbstractShaderProperty.cs
```diff
@@ -22,11 +22,19 @@ public bool gpuInstanced
             set {}
         }
 
-        internal virtual string GetHLSLVariableName(bool isSubgraphProperty)
+        internal virtual string GetHLSLVariableName(bool isSubgraphProperty, GenerationMode mode)
         {
+            if (mode == GenerationMode.VFX)
+                return $"PROP.{referenceName}";
```
johnpars, line 28: To support the GraphProperties struct that gets passed into the pixel/vertex functions, we need to modify the HLSL variable name to look up into that input struct.

com.unity.shadergraph/Editor/Data/Interfaces/GenerationMode.cs
```diff
@@ -3,7 +3,8 @@ namespace UnityEditor.ShaderGraph
     enum GenerationMode
     {
         Preview,
-        ForReals
+        ForReals,
+        VFX
```
johnpars, line 7: Introduce a third Generation Mode specific for VFX. This is currently required because VFX must configure vertex/surface description functions per-instance.

com.unity.shadergraph/Editor/Generation/Processors/Generator.cs
```diff
@@ -887,6 +902,17 @@ void GenerateShaderPass(int targetIndex, PassDescriptor pass, ActiveFields activ
                 spliceCommands.Add("Debug", debugBuilder.ToCodeBlock());
             }
 
+            // --------------------------------------------------
+            // Additional Commands
+
+            if (pass.additionalCommands != null)
```
johnpars, line 908: Introduce the AdditionalCommand descriptors to unlock VFX to splice in all of the complex work it does for property mapping between functions evaluation and assignments.

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
```diff
@@ -338,10 +338,14 @@ internal static string GetShaderText(string path, out List<PropertyCollector.Tex
         }
 
 #if VFX_GRAPH_10_0_0_OR_NEWER
-        // TODO: Fix this
+        // TODO: Fix this - VFX Graph can now use ShaderGraph as a code generation path. However, currently, the new
```
johnpars, line 341: Currently we must still support the ShaderGraphVfxAsset to maintain compatibility with the Visual Effect target. Thus this PR is also still slightly dependent on some of the information kept within this container (fragment/vertex properties). We also rely on it to filter out only shader graphs that support VFX in the object selector.

com.unity.visualeffectgraph/Editor/Compiler/VFXCodeGenerator.cs
```diff
@@ -362,10 +367,190 @@ static private void SubstituteMacros(StringBuilder builder)
             }
         }
 
+        internal static void BuildContextBlocks(VFXContext context, VFXContextCompiledData contextData,
```
johnpars, line 370: Lines 370-532 were made to break up and re-use some of the components of VFXCodeGenerator; but pretty much all of this should be consolidated into something simpler as there is a lot happening that is redundant.

com.unity.visualeffectgraph/Editor/Compiler/VFXCodeGenerator.cs
```diff
@@ -362,10 +367,190 @@ static private void SubstituteMacros(StringBuilder builder)
...
+
+            buildFragInputsGeneration = fragInputsGeneration.ToString();
+        }
+
         static private StringBuilder Build(VFXContext context, string templatePath, VFXCompilationMode compilationMode, VFXContextCompiledData contextData, HashSet<string> dependencies)
         {
             if (!context.SetupCompilation())
                 return null;
+
+            if (context is VFXShaderGraphParticleOutput shaderGraphContext &&
```
johnpars, line 539: Main entry point from VFX-side of the new Shader Graph generation path.

com.unity.visualeffectgraph/Editor/Compiler/VFXCodeGenerator.cs
```diff
@@ -556,6 +741,40 @@ static private StringBuilder Build(VFXContext context, string templatePath, VFXC
...
+            var graph = new GraphData
+            {
+                messageManager = new MessageManager(), assetGuid = AssetDatabase.AssetPathToGUID(path)
+            };
+            MultiJson.Deserialize(graph, textGraph);
+            graph.OnEnable();
+            graph.ValidateGraph();
+
+            // Configure the state of VFX target generation utils.
+            using (new VFXSubTarget.CompilationScope(context, contextData))
```
johnpars, line 765: Configure the static state of the VFXSubTarget with the context and context data, before invoking the shader graph generator.

com.unity.visualeffectgraph/Editor/Compiler/VFXGraphCompiledData.cs
```diff
@@ -845,6 +845,25 @@ private void GenerateShaders(List<GeneratedCodeData> outGeneratedCodeData, VFXEx
                     descs[i].source = generated.content.ToString();
                     descs[i].name = fileName;
                     descs[i].compute = generated.computeShader;
+
+                    if (generated.computeShader)
```
johnpars, line 849: This (ln 849-866) will be removed after the C++ PR lands. It is a mechanism to support faux material serialization to unblock development and allow testing everything else.

com.unity.visualeffectgraph/Editor/Data/VFXDataParticle.cs
```diff
@@ -191,8 +192,188 @@ public BucketInfo[] GetBucketLayoutInfo()
         private List<int> m_BucketOffsets = new List<int>();
     }
 
+    // [NOTE-VFX-MATERIALS]
```
johnpars, line 195: Note that this will be removed after the C++ PR lands. It is a mechanism to support faux material serialization to unblock development and allow testing everything else.

com.unity.visualeffectgraph/Editor/Data/VFXDataParticle.cs
```diff
@@ -191,8 +192,188 @@ public BucketInfo[] GetBucketLayoutInfo()
...
+        [SerializeField]
+        private VFXMaterialCollection m_MaterialCollection = new VFXMaterialCollection();
+
+        public Material GetOrCreateMaterial(VFXContext context) => m_MaterialCollection.GetOrCreate(context);
+
+        public void SyncContextMaterial(VFXContext context) => m_MaterialCollection.TrySyncRenderStateProperties(context);
+
+        static VFXMapping InjectMaterialState() => new VFXMapping("injectMaterialState", 1);
+
+        static void AppendMaterialParameters(Material material, List<VFXMapping> parameters, VFXContextCompiledData contextData, VFXExpressionGraph expressionGraph)
```
johnpars, line 319: This will also be removed after the C++ PR lands. It is a temporary hack to unblock development/testing that uses the VFXMappings to inject the state of the material into the underlying C++ material created by VFX.

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXSubTarget.cs
```diff
@@ -0,0 +1,67 @@
...
+                s_Context    = null;
+            }
+        }
+
+        internal static bool IsConfigured() => s_Configured;
+
+        internal static SubShaderDescriptor PostProcessSubShader(SubShaderDescriptor descriptor)
+        {
+            // TODO: Move generic VFX sub shader processing in here and break up the callback into SRP-specific portions (like FragInputs struct).
+            return OnPostProcessSubShader?.Invoke(descriptor, s_Context, s_Data) ?? descriptor;
```
johnpars, line 43: Forward the context + context data to the current render pipeline's implementation of VFX Sub Target.

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubTarget.cs
```diff
@@ -0,0 +1,374 @@
...
+        // Configures an HDRP Subshader with a VFX context.
+        private static SubShaderDescriptor PostProcessSubShader(SubShaderDescriptor subShaderDescriptor, VFXContext context, VFXContextCompiledData data)
+        {
+            var attributesStruct = GenerateVFXAttributesStruct(context, VFXAttributeType.Current);
+            var sourceAttributesStruct = GenerateVFXAttributesStruct(context, VFXAttributeType.Source);
+
+            // Defer to VFX to generate various misc. code-gen that ShaderGraph currently can't handle.
+            // We use the AdditionalCommand descriptors for ShaderGraph generation to splice these in.
+            // ( i.e. VFX Graph Block Function declaration + calling, Property Mapping, etc. )
+            GenerateVFXAdditionalCommands(
```
johnpars, line 31: This call is responsible for assembling all of the various code-gen that VFX is responsible for (graph function evaluation, assignment, etc) and organizing them into additional command descriptors.

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubTarget.cs
```diff
@@ -0,0 +1,374 @@
...
+namespace UnityEditor.Rendering.HighDefinition.ShaderGraph
+{
+    [InitializeOnLoad]
+    static class VFXHDRPSubTarget
+    {
+        internal const string Inspector = "Rendering.HighDefinition.VFXShaderGraphGUI";
+
+        static VFXHDRPSubTarget()
+        {
+            VFXSubTarget.OnPostProcessSubShader += PostProcessSubShader;
```
PaulDemeulenaere, line 19: I realized this could be a future design issue : OnPostProcessSubShader & VFXHDRPSubTarget are static, the OnPostProcessSubShader is appended with first C# compilation.
HDRP & URP can already cohabit within the same project (but there isn't any VFXURPSubTarget)
So either we should register/unregister this callback or simply handle a verification in "PostProcessSubShader" (the second option is probably the simplest).
It can be done later and this is maybe something you already anticipated since you added this comment :
`https://github.com/Unity-Technologies/Graphics/pull/3467/files#diff-42390dfcfd4234620ef9565c7a776d24cd6f05e33c1ddc99da21165533a4cc38R42`
Not really an issue yet.

com.unity.visualeffectgraph/Editor/Data/VFXDataParticle.cs
```diff
@@ -191,8 +192,188 @@ public BucketInfo[] GetBucketLayoutInfo()
...
+        private Dictionary<VFXContext, Material> m_MaterialMap = new Dictionary<VFXContext, Material>();
+        private Dictionary<VFXContext, RenderStateProperties> m_RenderStatePropertyMap = new Dictionary<VFXContext, RenderStateProperties>();
+
+        public Material GetOrCreate(VFXContext context)
+        {
+            Material mat;
+
+            if (!m_MaterialMap.TryGetValue(context, out mat))
+            {
+                mat = new Material(VFXResources.defaultResources.shader) { name = k_RenderSettingsMaterialName };
```
PaulDemeulenaere, line 220: I'm getting an exception here while loading VisualEffect_HDRP : VFXResources.defaultResources.shader is null

PaulDemeulenaere: It will be addressed by the new solution from @julienf-unity anyway

com.unity.visualeffectgraph/Editor/Data/VFXDataParticle.cs
```diff
@@ -191,8 +192,188 @@ public BucketInfo[] GetBucketLayoutInfo()
...
+        {
+            [SerializeField]
+            private List<string> m_PropertyNames = new List<string>();
+
+            [SerializeField]
+            private List<float> m_PropertyValues = new List<float>();
+
+            public RenderStateProperties(Material material)
+            {
+                var shader = material.shader;
```
PaulDemeulenaere, line 255: This access sometime generates this exception : 
MissingReferenceException: The object of type 'Material' has been destroyed but you are still trying to access it.
Your script should either check if it is null or you should not destroy the object.
UnityEditor.VFX.VFXMaterialCollection+RenderStateProperties..ctor (UnityEngine.Material material) (at G:/Unity/Dev_VFX_Extra/com.unity.visualeffectgraph/Editor/Data/VFXDataParticle.cs:255)
UnityEditor.VFX.VFXMaterialCollection.CacheRenderStates () (at G:/Unity/Dev_VFX_Extra/com.unity.visualeffectgraph/Editor/Data/VFXDataParticle.cs:295)
UnityEditor.VFX.VFXMaterialCollection.OnBeforeSerialize () (at G:/Unity/Dev_VFX_Extra/com.unity.visualeffectgraph/Editor/Data/VFXDataParticle.cs:285)
UnityEditor.EditorApplication:Internal_CallUpdateFunctions()
Same, it should simpler once the material is handled in VFXAsset.

com.unity.visualeffectgraph/Editor/Data/VFXDataParticle.cs
johnpars: Indeed, for the temporary material mechanism there have been issues such as this and the other one you noted, where I did not invest the time to fix since Julien has the new solution incoming

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
johnpars: CC @sebastienlagarde for discussion - basically this code forces us to re-evaluate all of the VFX code block a second time to reconstruct AttributesElement. Ideally we just share what gets created by VertMesh, but doing this would have some design implications for the rest of the times we invoke VertMesh etc.

VladNeykov:
Sharing the WIP [Test Doc](https://docs.google.com/document/d/1DkYmO_sQsRaAaB0s4DI9QxXlDqmoK5jt1xs8WyE78Ig/edit#).
Still lots left to test, but it has already a few actionable points.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
julienf-unity: I'm not sure I understand why this would be is needed? Why do we need the attributes?

johnpars: Attributes are required because we need to evaluate the vertex stack for the previous frame (to compute vertex animation motion vectors)

johnpars: But again if we just re-use the attributes from VertMesh this is not necesarry to reconstruct a second time

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
```diff
@@ -117,9 +132,13 @@ PackedVaryingsType MotionVectorVS(inout VaryingsType varyingsType, AttributesMes
...
         previousMesh.positionOS = effectivePositionOS;
 
-        previousMesh = ApplyMeshModification(previousMesh, _LastTimeParameters.xyz
+        previousMesh = ApplyMeshModification(previousMesh,
+            _LastTimeParameters.xyz
     #if defined(USE_CUSTOMINTERP_APPLYMESHMOD)
             , varyingsType.vmesh
+    #endif
+    #if defined(HAVE_VFX_MODIFICATION)
+            , element
```
johnpars, line 141: @julienf-unity The attributes must be passed to the previous frame's vertex stack

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
julienf-unity: Reevaluation of attributes will probably be stripped away by the compiler anyway as exactly identical expressions. But this should be only relevant for HAVE_MESH_MODIFICATION right?

There's is more a space problem here: inputMesh.positionOS should be in particle space and not in global VFX space. So TransformMeshToPreviousElement should not touch it. It should be only a matter of abstracting what is objectToWorld and previousObjectToWorld matrices.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
```diff
@@ -19,6 +19,7 @@ Pass
 
     // Pragmas
     $splice(PassPragmas)
+    #pragma enable_d3d11_debug_symbols
```
PaulDemeulenaere, line 22: N.B. : Think to remove the debug symbol

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
sebastienlagarde: Note: I am wondering if the custom interpolator feature of SG can be compatible with VFX?
https://github.com/Unity-Technologies/Graphics/pull/3281/files#diff-a1d073cb0296b152768b86b960e65e0a456db463b2ec858f7bbcc5a5c0fe675eR35

what do you think, can we simply add CustomInterpolators customInterpolators in the fraginputs of VFX?

sebastienlagarde: >Not very happy with how I currently move the entire struct splice up to this point, eager to hear thoughts about it.
I have for now no better idea, maybe we can revisit this in a second time, but gess that for now it is fine like this

johnpars:
Note that this PR will merge the code from #3744

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/SharedCode.template.hlsl
sebastienlagarde: you mean merge it in CustomInterpolatorVaryingsToFragInputs ? If we can, we should

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VertexAnimation.template.hlsl
sebastienlagarde: unless we can generalise VFX as just a list of interpolant and merge it with regualr custom interpolator feature I guess we have no choice?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
sebastienlagarde: just to clarify, you are worrying about the re-call to GetElementData(element); ?

johnpars: Yes- I worry about evaluating the entire graph blocks a second time to re-construct the element definition (which contains attributes that control the previous frame vertex stack, as well as the index to the previous element matrix). Unfortunately it's not very easy to share what gets done in VertMesh with MotionVectorVS

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VertexAnimation.template.hlsl
johnpars: Indeed we do not have a choice unless we can merge it with custom interpolator. Somewhat related, I have discussed with @PaulDemeulenaere about generalizing a lot of the content of this PR (for example 90% of what is in HDRP package for VFX generation can be moved into the VFX package and shared by all render pipelines). But it will be a sizeable amount of work and we are leaning on doing a later PR to generalize the code.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/VertMesh.hlsl
```diff
@@ -136,13 +136,33 @@ VaryingsMeshType VertMesh(AttributesMesh input, float3 worldSpaceOffset)
     UNITY_SETUP_INSTANCE_ID(input);
     UNITY_TRANSFER_INSTANCE_ID(input, output);
 
+#if defined(HAVE_VFX_MODIFICATION)
+    AttributesElement element;
+    ZERO_INITIALIZE(AttributesElement, element);
+
+    // Vertex might be culled early, thus we ensure the varyings are initialized to zero.
+    ZERO_INITIALIZE(VaryingsMeshType, output);
```
sebastienlagarde, line 144: This call is redudant with the previous one above. 

 VaryingsMeshType output;
    ZERO_INITIALIZE(VaryingsMeshType, output);

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
sebastienlagarde: We have done a call with John about this one

com.unity.shadergraph/Editor/Generation/Enumerations/InterpolationModifier.cs
```diff
@@ -0,0 +1,12 @@
+namespace UnityEditor.ShaderGraph
+{
+    [GenerationAPI]
+    internal enum InterpolationModifier
+    {
+        Linear,
```
sebastienlagarde, line 6: I am wondering if change related to modifier may conflict with this PR here: https://github.com/Unity-Technologies/Graphics/pull/3468/files#diff-13f5ff3fe703f5b644e6dd0281cd393c6987650f8cefa0d1f4e596332e95ebb6R105

johnpars: Hm, indeed it probably would. The question will be if that PR should make use of this PR's introduction of a formal Interpolation Modifier to the API, or just wait for that PR to merge and then I will merge with it and untangle the conflicts

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VFX/Config.template.hlsl
```diff
@@ -0,0 +1,296 @@
...
+#endif
+
+    return size3;
+}
+
+#if HAS_STRIPS
+#define PARTICLE_IN_EDGE (id & 1)
+float3 GetParticlePosition(uint index)
+{
+    struct Attributes attributes = (Attributes)0;
```
PaulDemeulenaere, line 100: Why this `struct` here ? Does it have any meaning in HLSL or is it a typo ?

JordanL8 (approved):
All relevant documentation has been added.

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubTarget.cs
```diff
@@ -0,0 +1,387 @@
...
+        {
+            Current,
+            Source
+        }
+
+        static string[] kVFXAttributeStructNames =
+        {
+            "Attributes",
+            "SourceAttributes"
+        };
```
PaulDemeulenaere, line 228: The "Attributes" naming can be confusing with the "builtin" struct field which has a completly different purpose : 
 https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.shadergraph/Editor/Generation/TargetResources/StructFields.cs#L8
 Maybe we should prefix them.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VFX/Config.template.hlsl
johnpars: Good catch, this function was a simple copy/paste from the original template which contains the same thing:
https://github.com/Unity-Technologies/Graphics/blob/70d8490a694d9443549fc2619b5677238522d552/com.unity.visualeffectgraph/Shaders/VFXParticleCommon.template#L259

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubTarget.cs
johnpars: Yes we should prefix these or perhaps splice the contents directly into AttributesElement

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
```diff
@@ -85,7 +85,11 @@ void MotionVectorPositionZBias(VaryingsToPS input)
 #endif
 }
 
+#ifdef HAVE_VFX_MODIFICATION
+PackedVaryingsType MotionVectorVS(inout VaryingsType varyingsType, AttributesMesh inputMesh, AttributesPass inputPass, AttributesElement inputElement)
```
sebastienlagarde, line 89: just to be sure we avoid error in case of future change, it is better to do it like this:

PackedVaryingsType MotionVectorVS(inout VaryingsType varyingsType, AttributesMesh inputMesh, AttributesPass inputPass, 
#ifdef HAVE_VFX_MODIFICATION
, AttributesElement inputElement
#endif
)

I know it could looks like a matter of taste, but often, as the other part of the code isn't compile, we forget to update both side.

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
```diff
@@ -3,26 +3,144 @@
...
+
+                    // This will correctly handle the configuration of keyword and pass setup.
+                    m_MaterialEditor.OnInspectorGUI();
+
+                    materialChanged = EditorGUI.EndChangeCheck();
+                }
+            }
+
+            // Indicate caution to the user if transparent motion vectors are disabled and motion vectors are enabled.
+            if ((m_MaterialEditor.target != null) &&
```
PaulDemeulenaere, line 98: Thi code should be in the previous `if (m_MaterialEditor != null)` scope

kroatoa:
Hi @johnpars,
You write

> Have not yet updated or created new tests.
what is your plan for that?

> Note that this PR does not support the HDRP Decal target. (though should, with some more work).

Is that work planned in this PR? If not when will it happen? The intent behind the question is to find out if it will be done before 21.2 code freeze in May.

UnityCatalina:
Hi @johnpars! Is there a [Jira Roadmap Epic](https://unity3d.atlassian.net/issues/?filter=17139) for this? 
Also, in terms of UX, it looks like there are several checkboxes that need turning on, in 2 different windows, before the user is allowed to use this feature. 
Is there a reason for this? Are there any performance or UX drawbacks to just having it on all the time?

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
johnpars: Indeed, and @vlad-unity caught an issue caused by it here: https://docs.google.com/document/d/1DkYmO_sQsRaAaB0s4DI9QxXlDqmoK5jt1xs8WyE78Ig/edit?disco=AAAALpPkIcs

It should now be resolved.

johnpars:
Hi @UnityCatalina, we track the feature on this Jira task: https://jira.unity3d.com/browse/HDRP-236

I will follow up with information regarding yours and @kroatoa remarks soon

johnpars:
@kroatoa 

> Is that work planned in this PR? If not when will it happen?

We will not ship Decal support in this PR or for 21.2. VFX Team is tracking Decal support in this task: https://jira.unity3d.com/browse/GVFXG-58

> what is your plan for that?

This PR will be coming with new test coverage for VFX shaders using this new ShaderGraph generation path in HDRP. We will test all HDRP subtargets (barring Decal) with all VFX output types, considering all render state configurations. It will also be a good idea to test for parity between VFX Shaders and normal Material shaders, so I will look into that as well.

Note that the VFX team is tracking URP support with this new generation path in this task: https://jira.unity3d.com/projects/GVFXG/issues/GVFXG-123

kroatoa (approved):
Thanks for clarifying the plans on which renderer pipelines this targets.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
sebastienlagarde: ping to not forget :)

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VFX/Config.template.hlsl
```diff
@@ -0,0 +1,302 @@
...
+{
+    uint index;
+    Attributes attributes;
+#if HAS_STRIPS
+    uint relativeIndexInStrip;
+    StripData stripData;
+#endif
+};
+
+bool ShouldCull(uint index)
```
sebastienlagarde, line 64: side note: the function name: GetSize / ShouldCull are extremely generic and could conflict with users function or thing like that. Maybe worse being carefult with name

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/VFX/Config.template.hlsl
sebastienlagarde: We have Macro for zero initilaization that you should used: ZERO_INITIALIZE()

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDShaderUtils.cs
sebastienlagarde: poking @slunity  for awareness of this coming change of ResetMaterialKeywords

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPBinder.cs
```diff
@@ -14,14 +20,156 @@ class VFXHDRPBinder : VFXSRPBinder
         public override string SRPAssetTypeStr  { get { return typeof(HDRenderPipelineAsset).Name; } }
         public override Type SRPOutputDataType  { get { return typeof(VFXHDRPSubOutput); } }
 
-        public override void SetupMaterial(Material mat)
+        static Dictionary<HDShaderUtils.ShaderID, string> s_ShaderNames = new Dictionary<HDShaderUtils.ShaderID, string>()
+        {
+            { HDShaderUtils.ShaderID.SG_Lit,      "Lit"      },
```
sebastienlagarde, line 25: Poking @slunity for this change as well

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPBinder.cs
```diff
@@ -14,14 +20,156 @@ class VFXHDRPBinder : VFXSRPBinder
...
+        public override void SetupMaterial(Material mat, bool hasMotionVector = false, bool hasShadowCasting = false, ShaderGraphVfxAsset shaderGraph = null)
         {
             try
             {
-                HDShaderUtils.ResetMaterialKeywords(mat);
+                if (shaderGraph != null)
+                {
+                    // Recover the HDRP Shader Enum from the VFX Shader Graph.
+                    var shaderID = GetShaderEnumFromShaderGraph(shaderGraph);
+                    HDShaderUtils.ResetMaterialKeywords(mat, shaderID);
```
sebastienlagarde, line 68: Last poke @slunity

sebastienlagarde (approved):
validation for the hdrp part

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPBinder.cs
slunity: Thanks for the heads up. Well the new HDSubTargets in my changes can directly provide the resetter function so even better.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/SharedCode.template.hlsl
```diff
@@ -68,6 +61,26 @@ SurfaceDescriptionInputs FragInputsToSurfaceDescriptionInputs(FragInputs input,
...
+#else
+    $SurfaceDescriptionInputs.ObjectSpacePosition:       output.ObjectSpacePosition =         TransformWorldToObject(input.positionRWS);
+    $SurfaceDescriptionInputs.ObjectSpaceTangent:        output.ObjectSpaceTangent =          TransformWorldToObjectDir(output.WorldSpaceTangent);
+    $SurfaceDescriptionInputs.ObjectSpaceBiTangent:      output.ObjectSpaceBiTangent =        TransformWorldToObjectDir(output.WorldSpaceBiTangent);
+    $SurfaceDescriptionInputs.ObjectSpaceViewDirection:  output.ObjectSpaceViewDirection =    TransformWorldToObjectDir(output.WorldSpaceViewDirection);
+
+    #if defined(SHADER_STAGE_RAY_TRACING)
+    $SurfaceDescriptionInputs.ObjectSpaceNormal:         output.ObjectSpaceNormal =           mul(output.WorldSpaceNormal, (float3x3) ObjectToWorld3x4());
+    #else
+    $SurfaceDescriptionInputs.ObjectSpaceNormal:         output.ObjectSpaceNormal =           normalize(mul(output.WorldSpaceNormal, (float3x3) UNITY_MATRIX_M));           // transposed multiplication by inverse matrix to handle normal scale
```
johnpars, line 80: Note that I have added this normalize for classic object space normal. Before it wasn't normalized which causes issues in some kinds of meshes. We can remove it if desired, but note that there is a discrepancy between VFX object normal and classic object normal without it.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/SharedCode.template.hlsl
```diff
@@ -68,6 +61,26 @@ SurfaceDescriptionInputs FragInputsToSurfaceDescriptionInputs(FragInputs input,
     $SurfaceDescriptionInputs.FaceSign:                  output.FaceSign =                    input.isFrontFace;
     $SurfaceDescriptionInputs.TimeParameters:            output.TimeParameters =              _TimeParameters.xyz; // This is mainly for LW as HD overwrite this value
 
+#if defined(HAVE_VFX_MODIFICATION)
```
johnpars, line 64: Here we organize treatment of object space data (VFX or classic shader).

johnpars:
Re-requesting your review @sebastienlagarde for a final look at two recent commits (**08cc749** and **0091153**) which implement the new flow of vertex transformation and handling of "particle space as object space"

com.unity.visualeffectgraph/Editor/Compiler/VFXCodeGenerator.cs
cdxntchou: Hax!   :)

Instead of storing the replacement sub target info in a global variable...
Could we just modify the graph itself here, removing the old subtarget and adding our replacement?

Or, I think ideally, we make a version of GetShaderText that takes a replacement target (or target array?) -- when called with a non-null replacement target it uses that instead of the targets embedded in the graph.  That would let us replace the target more directly here.

com.unity.shadergraph/Editor/Data/Interfaces/GenerationMode.cs
cdxntchou: Yeah, Ideally we would want to find a way to make this more generic, so you can configure it all from the Target/Subtarget API level. but I think that's a more architectural task for ShaderGraph.  This should be fine for now.

cdxntchou (changes requested):
Awesome!

Main suggestion is fix the indirect way that VFXSubTarget / VFXHDRPSubTarget inject their information via global variables.  I think we could just replace the Targets/SubTargets in the graph (assuming it's a temp graph, not serialized or cached or anything).  Or even better directly pass down the desired Targets/SubTargets to the code gen, and have code gen use those instead of graph-embedded targets.

Secondary is if we could reduce the VFX specific code in ShaderGraph, i.e. maybe replace WorksWithVFX() Target function with an interface?

I think the VFX generation mode is fine for now, fixing that would be a much larger task.

I didn't look at the specifics of the shader template code -- hopefully HDRP and VFX have checked it!  :)

LandonTownsendUnity (changes requested):
Graphs with custom interpolators throw errors and fail when used in VFX graph
image

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
johnpars: Late reply on my thoughts to this- but overall I think it is definitely possible to utilize custom interpolators to transfer the VFX properties (it would also solve this procedural FragInputs since the way custom interpolators handle it, the just have a substruct wrapped in a preprocessor). 

But, I think it is out of the scope of this PR to do it. As you say, a second time/PR seems correct for this

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/HDShaderGUI.cs
```diff
@@ -106,7 +106,7 @@ protected static void SynchronizeShaderGraphProperties(Material material)
         {
             var defaultProperties = new Material(material.shader);
             foreach (var floatToSync in floatPropertiesToSynchronize)
-                if (material.HasProperty(floatToSync))
+                if (material.HasProperty(floatToSync) && defaultProperties.HasProperty(floatToSync))
```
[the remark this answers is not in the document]
johnpars: Indeed I do remember specifically hitting an issue with this line originally failing to catch a non-existent default property (I did not investigate too much into the why) for the Unlit shader (`RequiresSplitLighting` I believe). So, I added this case to the predicate to protect against the case of a non-existent default property, and haven't encountered the issue since (though it seemed to happen randomly and was hard to reproduce).

com.unity.visualeffectgraph/Editor/GraphView/VFXViewPreference.cs
```diff
@@ -98,6 +109,7 @@ public override void OnGUI(string searchContext)
                 using (new SettingsWindow.GUIScope())
                 {
                     LoadIfNeeded();
+                    m_GenerateOutputContextWithShaderGraph = EditorGUILayout.Toggle(new GUIContent("Improved Shader Graph Generation", "When enabled, any shadergraph shaders can be assigned to Visual Effect outputs as long as ‘Support VFX Graph’ is enabled in the ShaderGraph’s Graph Inspector."), m_GenerateOutputContextWithShaderGraph);
```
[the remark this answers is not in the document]
johnpars: I suppose not. Technically a user could still just use the legacy VFX target as that still works. (However if they make a graph with VFX Support enabled + a legacy VFX Target, the former takes precedence).

johnpars:
Hi @cdxntchou, thanks for the review - I have addressed some of the minor notes and wanted to expand with more detail on my approach to a couple larger ones:

1) Replacement of WorksWithVFX from Target API and Introduction of IMaySupportVFX: https://github.com/Unity-Technologies/Graphics/pull/3467/commits/ae768333ca8db4718c2f27e4502eb6914e396629
2) Fixing the indirect hax:
a. Generator now accepts an optional target array: https://github.com/Unity-Technologies/Graphics/pull/3467/commits/937b26f576b471371c96c6a241a5d0eb7e1162d9
b. Introduce a second interface to transfer VFX context data to targets (this is less SG related): https://github.com/Unity-Technologies/Graphics/pull/3467/commits/8a7d91ba85198759bb5afb03109974ae6dbb2c63
c. Using A and B to replace the static globals with explicit target configuration: https://github.com/Unity-Technologies/Graphics/pull/3467/commits/0128cef4e89228dd6b8b44d436fc7171061b9d25

TestProjects/VisualEffectGraph_HDRP/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -230,8 +230,6 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/AllTests/VFXTests/GraphicsTests/28_FlipBookArray.unity
     guid: 69ce8d3ad279c734cbea097f07ebb1fc
-    path: Assets/AllTests/VFXTests/GraphicsTests/28_CameraProject.unity
```
PaulDemeulenaere, line 233: This one is a bad merge, I renabled it in an incoming PR

TestProjects/VisualEffectGraph_HDRP/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -241,10 +239,10 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/AllTests/VFXTests/GraphicsTests/31_CameraBuffer.unity
     guid: 022b924fd58e94e41bcc5a24218370fe
-  - enabled: 0
-    path: Assets/AllTests/VFXTests/GraphicsTests/32_ExcludeFromTAA.unity
-    guid: 8f9c50f62a67d6d429a71cb9ad4a3f57
```
PaulDemeulenaere, line 246: This test has been temporarily removed, you should resolve the editor settings change manually : There is a proess which is removing unchecked test in editorSettings

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubOutput.cs
```diff
@@ -44,6 +44,14 @@ protected override IEnumerable<string> filteredOutSettings
                     yield return "transparentRenderQueue";
                 else
                     yield return "opaqueRenderQueue";
+
+                if (owner is VFXShaderGraphParticleOutput shaderGraphOutput &&
```
PaulDemeulenaere, line 48: Minor :  it could be more appropriate to merge this condition with `!supportsQueueSelection` or even integrate this test in `supportsQueueSelection`, I'm not sure it make sense but it can be more consitent.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDSubTarget.cs
```diff
@@ -176,8 +186,15 @@ protected SubShaderDescriptor PostProcessSubShader(SubShaderDescriptor subShader
                 // Set default values for HDRP "surface" passes:
                 if (passDescriptor.structs == null)
                     passDescriptor.structs = CoreStructCollections.Default;
+
                 if (passDescriptor.fieldDependencies == null)
-                    passDescriptor.fieldDependencies = CoreFieldDependencies.Default;
+                {
+                    // For now we handle VFX here since the only target that specifies field dependencies is decal, which is currently not supported.
```
julienf-unity, line 192: This sounds a bit like a hack

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPBinder.cs
```diff
@@ -14,14 +20,156 @@ class VFXHDRPBinder : VFXSRPBinder
...
+
+        // List of shader properties that currently are not supported for exposure in VFX shaders.
+        private static readonly Dictionary<Type, string> s_UnsupportedShaderPropertyTypes = new Dictionary<Type, string>()
+        {
+            { typeof(DiffusionProfileShaderProperty), "Diffusion Profile" },
+            { typeof(VirtualTextureShaderProperty),   "Virtual Texture"   },
+            { typeof(GradientShaderProperty),         "Gradient"          }
+        };
+
+        public override bool IsGraphDataValid(GraphData graph)
```
julienf-unity, line 146: This should probably be in the base class because it does not seem HDRP specific

julienf-unity: But I guess this can be done later when implementing URP

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDShaderPasses.cs
```diff
@@ -280,22 +280,32 @@ RenderStateCollection GenerateRenderState()
...
+                        HDStructFields.AttributesMesh.uv1,
+                        HDStructFields.AttributesMesh.color,
+                        HDStructFields.AttributesMesh.uv2,
+                        HDStructFields.AttributesMesh.uv3,
+                        HDStructFields.FragInputs.tangentToWorld,
+                        HDStructFields.FragInputs.positionRWS,
+                        HDStructFields.FragInputs.texCoord1,
+                        HDStructFields.FragInputs.texCoord2,
+                        HDStructFields.FragInputs.texCoord3,
+                        HDStructFields.FragInputs.color,
```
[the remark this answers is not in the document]
johnpars: I am not sure either, this is how it was done for HDRP prior to the PR and the modifications here were simply necessary to handle the Unlit case where the compiler optimization caused z-fighting artifacts.

If the PR is green for all tests I suggest we investigate this in a future PR since it is somewhat orthogonal to VFX's refactor, imo. cc @sebastienlagarde

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubOutput.cs
johnpars: Good point, missed that case :)

cdxntchou (approved):

cdxntchou:
> 
> 
> Hi @cdxntchou, thanks for the review - I have addressed some of the minor notes and wanted to expand with more detail on my approach to a couple larger ones:
> 
>     1. Removal of `WorksWithVFX` from Target API and Introduction of `IMaySupportVFX`: [ae76833](https://github.com/Unity-Technologies/Graphics/commit/ae768333ca8db4718c2f27e4502eb6914e396629)
> 
>     2. Fixing the indirect hax:
>        a. Generator now accepts an optional target array: [937b26f](https://github.com/Unity-Technologies/Graphics/commit/937b26f576b471371c96c6a241a5d0eb7e1162d9)
>        b. Introduce a second interface `IRequireVFXContext` to transfer VFX context data to targets (this is less SG related): [8a7d91b](https://github.com/Unity-Technologies/Graphics/commit/8a7d91ba85198759bb5afb03109974ae6dbb2c63)
>        c. Using A and B to replace the static globals with explicit target configuration: [0128cef](https://github.com/Unity-Technologies/Graphics/commit/0128cef4e89228dd6b8b44d436fc7171061b9d25)
> 
> 
> Note that there are now two interfaces. One is intended to allow a Target to delegate the shader graph importing behavior (since it must generate the `ShaderGraphVfxAsset`). The second is intended to allow the exchange of VFX data to a target. I needed to split them into two, otherwise there would have been some tricky dependency issues between SG & VFX.

hmm, yeah I see.  We can only merge these two interfaces once we get rid of the need for the extra VFX data coming out of the shadergraph importer...

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/SharedCode.template.hlsl
```diff
@@ -63,6 +61,26 @@ SurfaceDescriptionInputs FragInputsToSurfaceDescriptionInputs(FragInputs input,
...
+    // VFX maintains an internal coordinate space and here we handle the case of going from world to element space (object space to the user).
+    $SurfaceDescriptionInputs.ObjectSpacePosition:       output.ObjectSpacePosition =         TransformWorldToElement(input.worldToElement, input.positionRWS);
+    $SurfaceDescriptionInputs.ObjectSpaceTangent:        output.ObjectSpaceTangent =          TransformWorldToElementDir(input.worldToElement, output.WorldSpaceTangent);
+    $SurfaceDescriptionInputs.ObjectSpaceBiTangent:      output.ObjectSpaceBiTangent =        TransformWorldToElementDir(input.worldToElement, output.WorldSpaceBiTangent);
+    $SurfaceDescriptionInputs.ObjectSpaceViewDirection:  output.ObjectSpaceViewDirection =    TransformWorldToElementDir(input.worldToElement, output.WorldSpaceViewDirection);
+    $SurfaceDescriptionInputs.ObjectSpaceNormal:         output.ObjectSpaceNormal =           TransformWorldToElementDir(input.worldToElement, output.WorldSpaceNormal);
```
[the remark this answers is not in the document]
johnpars: This is now fixed via the abstraction of UNITY_MATRIX_M and UNITY_MATRIX_I_M as well as the new SG mechanism to require transform for certain nodes or data.

julienf-unity (commented):
Looks like the PR is in a good state now. Goal is to merge asap (this week or beginning of next week).
There are still 2 specific issues to fix on DX12 and Vulkan before landing (which seems specific to new SG integration according to QA)

sebastienlagarde (approved):

VladNeykov (approved):
Everything green on the [test doc](https://docs.google.com/document/d/1DkYmO_sQsRaAaB0s4DI9QxXlDqmoK5jt1xs8WyE78Ig/edit#), thanks for the many iterations and your hard work on this @johnpars !

PaulDemeulenaere (changes requested):
The head still have an unexpected "#pragma enable_d3d11_debug_symbols"
No real impact on execution beside build size.

johnpars:
Yamato is green. Marking ready for review

julienf-unity (approved):

PaulDemeulenaere (approved):
I double check the discussion and everything has been resolve.
LGTM, we have to be sure that Nightly_VFX_HDRP is 🟢 but also HDRP, we got this failure on last [run](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fsg-vfx-integration-improved/.yamato%252Fhdrp-osx-metal.yml%2523HDRP_OSX_Metal_playmode_cache_mono_Linear_trunk/6540064/job) with OSX :
```[06:26:38.461 Information] [260/262] TestFinish
	name: UnityEngine.TestTools.Graphics.Tests.ImageAssertTests.PerPixel_BadB_Throws
	result: PASSED
	duration: 0.010073 seconds
[06:26:45.241 Information] LogEntry
	severity: Warning
	message: 61 TextureIDs were not freed
	stacktrace: 
	line: 1734
	file: ./Runtime/GfxDevice/GfxDevice.cpp```
I don't think it's related.
Great job @johnpars ! This refactor will unleash a lot of technical constraint we had with VisualEffectGraph integration.
FYI, I started to port your approach to URP (see this [branch](https://github.com/Unity-Technologies/Graphics/tree/HDRP/sg-vfx-integration-improved-to-urp)) but it's still in progress. I will come back to you once I have something working in both pipeline.

sebastienlagarde:
Re-launched nightly with custom revision that should be green: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fsg-vfx-integration-improved/.yamato%252Fall-hdrp.yml%2523Nightly_HDRP_CUSTOM-REVISION/6546711/job/pipeline

LandonTownsendUnity (approved):

PR #3468, opened 2021-02-11
Support of interpolators for SV_POSITION in shader graph and a conservative depth offset mode.

anisunity:
On an internal user request, we added the possibility to specify that the depth offset is always away from the base one. That allows for potential performance gains due to it not breaking the early depth test mechanic. The option is available on shader graph under the depth offset section (or in the material at the same position).
image

This required to modify the shader graph code to allow for interpolators for the SV_Position Attribute.

By default the option is disabled.

This requires some docs update @JordanL8 FYI

Testing status:
It not possible to add an automated tests for this unfortunately as it doesn't do anything visible.
The only testing I did was take a pix capture with the option available.
image

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
@@ -344,7 +344,7 @@ Shader "HDRP/Lit"
     #ifndef _SURFACE_TYPE_TRANSPARENT
     #define _DEFERRED_CAPABLE_MATERIAL
     #endif
-    
+
     // In this shader, the heightmap implies depth offsets away from the camera.
     #ifdef _HEIGHTMAP
     #define _CONSERVATIVE_DEPTH_OFFSET


sebastienlagarde (commented):
we need to add the option on regular lit shader as well (as we support parallax mapping there). I will suggest that currently we force the option for them in this case

JordanL8 (approved):

JordanL8 (approved):

alelievr (approved):

com.unity.render-pipelines.high-definition/Documentation~/master-stack-unlit.md
```diff
@@ -111,6 +111,7 @@ Depending on the [Graph Settings](#graph-settings) you use, Shader Graph can add
 [!include[](snippets/shader-properties/surface-options/alpha-to-mask.md)]
 [!include[](snippets/shader-properties/surface-options/double-sided.md)]
 [!include[](snippets/shader-properties/surface-options/ss-depth-offset.md)]
+[!include[](snippets/shader-properties/surface-options/conservative-depth-offset.md)]
```
sebastienlagarde, line 114: we don't have conservative depth for unlit iirc

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -54,6 +54,14 @@
 #define RAY_TRACING_OPTIONAL_ALPHA_TEST_PASS
 #endif
 
+#if _CONSERVATIVE_DEPTH_OFFSET
```
sebastienlagarde, line 57: guess it should be #ifdef _CONSERVATIVE_DEPTH_OFFSET

sebastienlagarde (commented):
We need to add the conservative depth to Lit.shader, LayeredLit.shader, LitTessellation.shader, LayeredLitTessellation.shader

anisunity:
@sebastienlagarde did the review changes you asked for

com.unity.render-pipelines.high-definition/Documentation~/master-stack-unlit.md
anisunity: If we have depth offsets on unlit shader, i don't see why we couldn't have the conservative constraint for them aswell

sebastienlagarde: ah, indeed - sorry made a confusion - unlit.shader don't have a need for it as we don't allow to control depth offset there BUT unlit master node have it. Is the current PR handling this case?

RemyUnity:
> It not possible to add an automated tests for this unfortunately as it doesn't do anything visible.

Wouldn't it be possible to test it by having an object A with a negative pixel depth offset behind of a normal one, B ?
If conservative is disabled, the object A should be partially visible, and if it's enabled not ?

sebastienlagarde (approved):

sebastienlagarde:
Waiting for QA and reply fo Anis to @RemyUnity  question

sebastienlagarde:
We also need to add this change to the what's new page and the upgrade guide

com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
```diff
@@ -102,7 +102,7 @@ internal static void GenerateShaderStruct(StructDescriptor shaderStruct, ActiveF
 
                         //check for a semantic, build string if valid
                         string semantic = subscript.HasSemantic() ? $" : {subscript.semantic}" : string.Empty;
-                        structBuilder.AppendLine($"{subscript.type} {subscript.name}{semantic};");
+                        structBuilder.AppendLine($"{subscript.interpolation} {subscript.type} {subscript.name}{semantic};");
```
sebastienlagarde, line 105: Note for me: tracking change compare to this PR: https://github.com/Unity-Technologies/Graphics/pull/3467/files#diff-13f5ff3fe703f5b644e6dd0281cd393c6987650f8cefa0d1f4e596332e95ebb6R105

anisunity:
If the user provides a negative depth offset and declared it conservative it is an undefined behavior, meaning the pixels will contain trash data and that cannot be tested.

cdxntchou (approved):
Looks good

TomasKiniulis (approved):
Double checking Conservative option with all SG Master nodes to be sure no visual issues or errors occur. Looks good.

anisunity:
An instability in the test running was hit. Merged master into this, re-formatted and reran the tests

PR #3472, opened 2021-02-11
Bugfix 1287335: Enum keywords would fail when an entry was given a du…

joshua-davis:
…plicate name that differed only by case. In addition there were various symbols that could be used that would break enum keywords due to how the material enum property editor sanitized display names.

https://fogbugz.unity3d.com/f/cases/1287335/

### Checklist for PR maker

---
### Purpose of this PR
Originally the bug was that giving two enum keywords the same display name with different cases would make the keyword not function properly.

After digging deeper, the bigger problem is that the material property editor uses the display name to enable the keyword, since the names didn't match (e.g. `A(1)` vs. `A_1_`) the keyword would never get set. This was the root cause with the duplicate names as one would have a display name of `a` but a reference name of `A_1`.

To fix this, enum keyword display names are now sanitized to match the same rules that the material editor expects, this way the same reference name is looked up. That is only allowing alpha numeric, spaces, and underscores.

---
### Testing status
- [x] Tested enums with duplicate display names: [`A`, `a`]
- [x] Partially invalid display names: `A(1)`
- [x] Empty display names: `""`
- [x] Whitespace display names: `" "`

---
### Comments to reviewers
While changing the rules for names could have the potential to break things, the rules are only changed for things that were already invalid hlsl identifiers. Anything that was valid before will have the same reference name and anything else didn't work.

Also, during investigation, I determined that display names that were empty or completely white space would break things. In this case, one entry would be missing (e.g. if you make: ["A", "", "C"] then what's displayed is ["A", "C"]).

### DIFF
--- a/com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
+++ b/com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
@@ -23,7 +23,7 @@ namespace UnityEditor.ShaderGraph.Drawing.Inspector.PropertyDrawers
     class ShaderInputPropertyDrawer : IPropertyDrawer
     {
         internal delegate void ChangeExposedFieldCallback(bool newValue);
-        internal delegate  void ChangeDisplayNameCallback(string newValue);
+        internal delegate void ChangeDisplayNameCallback(string newValue);
         internal delegate void ChangeReferenceNameCallback(string newValue);
         internal delegate void ChangeValueCallback(object newValue);
         internal delegate void PreChangeValueCallback(string actionName);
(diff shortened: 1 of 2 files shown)


com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
```diff
@@ -1208,18 +1210,25 @@ void KeywordAddCallbacks()
...
                 if (EditorGUI.EndChangeCheck())
                 {
-                    keyword.entries[index] = new KeywordEntry(index + 1, displayName, referenceName);
+                    displayName = displayName.Trim();
+                    referenceName = GetSanitizedReferenceName(displayName.ToUpper());
+                    var duplicateIndex = FindDuplicateReferenceNameIndex(entry.id, referenceName);
+                    if (duplicateIndex != -1)
+                    {
+                        var duplicateEntry = ((KeywordEntry)m_KeywordReorderableList.list[duplicateIndex]);
+                        Debug.LogWarning($"Display name '{displayName}' will create the reference name '{referenceName}' which conflicts with the existing display name '{duplicateEntry.displayName}' at index {duplicateIndex}.");
```
LandonTownsendUnity, line 1225: This wording is a little big confusing, can you change it to something like
"Display name '{displayName}' will create the same reference name '{referenceName}' as the existing display name '{duplicateEntry.displayName}' at index {duplicateIndex}."

ghost (changes requested):

jessebarker (approved):
LGTM.

ghost (approved):

cdxntchou (approved):

cdxntchou:
Little bit non-ideal that it only shows the error in the console.   But it's such a low-occurrence thing, most users are unlikely to ever run into it, so I think it's ok.

PR #3474, opened 2021-02-11
[ShaderGraph][12] Add the line number to the message shown by a node badge

jessebarker:
---
### Purpose of this PR
This is an addendum to PR 3410.  It was pointed out after merging that we were not retaining the same details around node errors as the user would get from a console message.  So, even though a conscious choice was made not to use the console for these messages, we want to give developers a better hint as to what is wrong with a node.

---
### Testing status
Manual testing as per https://fogbugz.unity3d.com/f/cases/1296291/
Yamato tests to come.

---
### Comments to reviewers
This is the simplest version of this change (without changing the interface for AttachMessage).  When the message is not specific to the code, we will still output "at line 0" (as that is what is set on ShaderMessage).

### DIFF
--- a/com.unity.shadergraph/Editor/Drawing/Views/GraphEditorView.cs
+++ b/com.unity.shadergraph/Editor/Drawing/Views/GraphEditorView.cs
@@ -860,7 +860,8 @@ namespace UnityEditor.ShaderGraph.Drawing
                 else
                 {
                     var foundMessage = messageData.Value.First();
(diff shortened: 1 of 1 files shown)


LandonTownsendUnity (approved):
Thank you for adding this! Pretty simple change, shouldn't need much testing.

joshua-davis (approved):

marctem (approved):
Looks fine, foundMessage.line is always initialized.

PR #3450, opened 2021-02-10
[2021.2] Updating reference image for 005_LitBakedEmission on DX11 playmode

ellioman:
# Purpose of this PR
* Fixing a broken test because of a incorrect reference image.
* Deleting an unused reference image 

# Yamato

PR #3478, opened 2021-02-12
[2021.2] Redoing fixes for URP Split Test Projects

ellioman:
# Purpose of this PR
This PR is doing the following:
1. Redoing the Testfilters changes done in #3464 
2. Redoing the reference image update done in #3450 
3. Updating reference image for 144 SSAO on Windows Vulkan & iPhone
4. Cleaning up the Test Filters file, which had a lot null scene entries because of the split project

# Yamato


svens-unity (approved):

hdb-unity (approved):
Looks like this will fix:
- 005 scene in lighting
- 144 scene in postpro
- update test filters

ellioman:
The failures on Yamato are known or cancelled tests that are not affected by this PR.

PR #3482, opened 2021-02-12
[2021.2] Cleaning TestFilters files in the Split Projects.

ellioman:
# Purpose of this PR
After splitting the test project, our test filters files have many null references as the scenes have been moved to other projects.

#3478 cleaned up the one in foundation. This PR cleans up the 2D, Lighting, Post Processing and Terrain projects.


svens-unity (approved):
changes to images should go away once the other PR has landed?

hdb-unity (approved):
I understand there is minimal risk of halo effect from:
- 4x reference image
- 1x meta file

Otherwise, makes the test filters more readable.

ellioman:
Failures on Yamato are not related to this PR.

PR #3484, opened 2021-02-12
DLSS in HDRP - Deep Learning Super Sampling integration

kecho:
**Integrating NVIDIA's Deep Learning Super Sampling into HDRP**

### Purpose of this PR
* Adding new DLSS upscaler into HDRP.
* Adding the concept of upscaling before post process. 
* Post process pipeline massaging and refactor of destination size. Added the concept of post process size scales, for sampling and rendering, since now the dynamic resolution upscaler could run before or after post processes.
* Adding new UI controls of DLSS
* Adding new tests images for DLSS
* Added new debug panel in the render pipeline analysis pane, to check the internal state of NVIDIA unity plugin.

### Testing status:
To test, just get the latest trunk version, which contains DLSS backend features.

The following steps have been tested locally:

1.  Open the package manager, select in the filter "Built-in" , ensure to enable the "NVIDIA" core native module in your project.
2.  Ensure you dont have render doc on otherwise, the feature wont be detected. In the HDRP quality settings, right under Dynamic Resolution section, enable "tick on the Enable NVIDIA Deep Learning Super Sampling Checkbox"
3.  Ensure your desired camera has Dynamic Resolution Enabled (check in the inspector the checkbox is on)
4.  Read the info box that will appear. This info box will let you know if DLSS is detected or not on this session. If not, then the fallback regular antialiasing is used, and the standard DRS path is executed.
5.  Inspect / play around with software/hardware dx12/dx11/vulkan  APIs of the effect

This effect must not be present in any other platform other than Windows, and in the case any GPU other than NVIDIA's is present, the effect must not be enabled.

If you are measuring performance, please ensure that you are looking at GPU time, using the GPU profiler, and not the CPU profiler. Note too that DLSS wont make your frame run faster unless your scene is highly bound by gpu draw calls.
One example is spaceship demo.

### Known Testing Issues:
We have 1 known crash, vulkan on intel CPU is now crashing %100 of the time, this is being addressed in a separate PR:
https://ono.unity3d.com/unity/unity/pull-request/125566/_/graphics/fix-nvapi-vulkan-crash


com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
```diff
@@ -221,19 +226,54 @@ public void SetReferenceSize(int width, int height, MSAASamples msaaSamples, boo
...
+            return CalculateRatioAgainstMaxSize(viewport.x, viewport.y);
+        }
+
+        internal Vector2 CalculateRatioAgainstMaxSize(int width, int height)
+        {
+            Vector2 maxSize = new Vector2Int(GetMaxWidth(), GetMaxHeight());
+            return new Vector2((float)width / maxSize.x, (float)height / maxSize.y);
+        }
+
+        internal Vector2 CalculateScaledRatioAgainstMaxSize(int width, int height)
```
FrancescoC-unity, line 272: I don't think we need one overload Vector2Int and one with the two separate, I'd keep only the Vector2Int , but either are fine I think

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -316,6 +321,17 @@ float2 ClampAndScaleUVForBilinear(float2 UV)
     return ClampAndScaleUV(UV, _ScreenSize.zw, 0.5f);
 }
 
+// This is assuming full screen buffer and .
```
FrancescoC-unity, line 324: Just noticed I never finished typing this comment apparently (from above, I guess this is copy paste) 

`// This is assuming full screen buffer and half a texel offset for the clamping. `

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
```diff
@@ -221,19 +226,54 @@ public void SetReferenceSize(int width, int height, MSAASamples msaaSamples, boo
...
+            Vector2 scales;
+            if (DynamicResolutionHandler.instance.HardwareDynamicResIsEnabled() && m_HardwareDynamicResRequested)
+            {
+                scales = CalculateScaledRatioAgainstMaxSize(DynamicResolutionHandler.instance.finalViewport);
+            }
+            else
+            {
+                scales = CalculateRatioAgainstMaxSize(width, height);
             }
+            m_RTHandleProperties.rtHandlePostProcessScale = new Vector4(scales.x, scales.y, 0.0f, 0.0f);
```
FrancescoC-unity, line 253: Missing the previous frame ones (`m_RTHandleProperties.rtHandlePostProcessScale.x` and `m_RTHandleProperties.rtHandlePostProcessScale.y ` )

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
```diff
@@ -221,19 +226,54 @@ public void SetReferenceSize(int width, int height, MSAASamples msaaSamples, boo
...
+        internal void SetPostProcessScale(int width, int height)
+        {
+            Vector2 scales;
+            if (DynamicResolutionHandler.instance.HardwareDynamicResIsEnabled() && m_HardwareDynamicResRequested)
+            {
+                scales = CalculateScaledRatioAgainstMaxSize(DynamicResolutionHandler.instance.finalViewport);
+            }
+            else
+            {
+                scales = CalculateRatioAgainstMaxSize(width, height);
```
FrancescoC-unity, line 251: This is because the rounding is different in HW mode? Why not calling the CalculateScaledRatioAgainstMaxSize in this case too?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
```diff
@@ -882,6 +900,69 @@ TextureHandle ContrastAdaptiveSharpeningPass(RenderGraph renderGraph, HDCamera h
...
+                    name = "Resolved Color Buffer",
+                    colorFormat = renderGraph.GetTextureDesc(inputColor).colorFormat,
+                    useMipMap = false,
+                    enableRandomWrite = true
+                });
+                passData.outputTextures.color = builder.WriteTexture(passData.outputTextures.color);
+
+                passData.outputTextures.depthBuffer = renderGraph.CreateTexture(new TextureDesc(Vector2.one, false, true)
+                {
+                    name = "Resolved Depth Buffer",
```
FrancescoC-unity, line 923: I know this is all tmp, but worth making a note here to know that most of the PP down the line sample the pyramid and not the depth buffer directly,

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -70,6 +70,29 @@ private enum SMAAStage
...
+        Vector2Int m_FullRes = new Vector2Int(1,1);
+        Vector2Int m_DownsampledRes = new Vector2Int(1,1);
+
+        internal enum ResolutionGroup
+        {
+            Downsampled,
+            Full
+        }
+
+        internal ResolutionGroup ResGroup { set; get; }
```
FrancescoC-unity, line 84: This should be camelCase as per conventions (same for below Viewport and UpsamplerSchedule)

FrancescoC-unity: I wonder though, can these be private instead?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -70,6 +70,29 @@ private enum SMAAStage
...
         readonly HableCurve m_HableCurve;
 
+        //Viewport information
+
+        Vector2Int m_FullRes = new Vector2Int(1,1);
+        Vector2Int m_DownsampledRes = new Vector2Int(1,1);
+
+        internal enum ResolutionGroup
+        {
+            Downsampled,
```
FrancescoC-unity, line 80: I'd maybe chose a different name as Downsampled here is downsampled only in the context of dynamic resolution, when no dynamic res is on it might be confusing. 

I don't have yet a better name in mind that is not maga verbose, but will try to think about it a bit :D

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -70,6 +70,29 @@ private enum SMAAStage
...
+
+        internal Vector2Int Viewport { get { return  ResGroup == ResolutionGroup.Full ? m_FullRes : m_DownsampledRes; } }
+
+        internal enum UpsamplerScheduleType
+        {
+            BeforePost,
+            AfterPost,
+        }
+
+        internal UpsamplerScheduleType UpsamplerSchedule { set; get; }
```
FrancescoC-unity, line 94: I think this information needs to stay with the dynamic res handler

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -293,6 +322,10 @@ public void BeginFrame(CommandBuffer cmd, HDCamera camera, HDRenderPipeline hdIn
             // Override full screen anti-aliasing when doing path tracing (which is naturally anti-aliased already)
             m_AntialiasingFS        &= !m_PathTracing.enable.value;
 
+            // Antialiasing of any sort is disabled on DRS Pre post processes.
+            // The assumption here is that upsamplers that work before post processes should perform anti aliasing as well in one go.
```
FrancescoC-unity, line 326: Uhm this might be problematic without first changing the UI a bit, this will disregard what the user selects in the options and is not great UX.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -3299,13 +3336,67 @@ static void DoContrastAdaptiveSharpening(in CASParameters parameters, CommandBuf
...
+            return parameters;
+        }
+
+        static void DoUpsampleScene(
+            UpsampleSceneParameters parameters, CommandBuffer cmd,
+            RTHandle sourceColor, RTHandle sourceDepth, RTHandle sourceMotionVectors,
+            RTHandle outputColor, RTHandle outputDepth, RTHandle outputMotionVectors)
+        {
+            var mainKernel = parameters.mainUpsampleKernel;
+            if (mainKernel < 0)
```
FrancescoC-unity, line 3368: I think an hard fail is better here since there is no scenario when this is fine :D

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -693,6 +705,15 @@ internal static void ResetAllHistoryRTHandleSystems(int width, int height)
                 }
             }
         }
+    
+        unsafe internal void UpdateScalesAndScreenSizesCB(ref ShaderVariablesGlobal cb)
+        {
+            cb._ScreenSize = screenSize;
+            cb._PostProcessScreenSize = postProcessScreenSize;
+            cb._RTHandleScale = RTHandles.rtHandleProperties.rtHandleScale;
+            cb._RTHandlePostProcessScale = RTHandles.rtHandleProperties.rtHandlePostProcessScale;
```
FrancescoC-unity, line 714: We might need the _RTHandleScaleHistory  equivalent too as we create history for some PP passes too.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -1526,6 +1526,34 @@ void RenderGizmos(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle colo
...
+
+        internal void UpdatePostProcessScreenSize(RenderGraph renderGraph, HDCamera hdCamera, int postProcessWidth, int postProcessHeight)
+        {
+            using (var builder = renderGraph.AddRenderPass("Update RT Handle Scales CB", out UpdatePostProcessScreenSizePassData passData))
+            {
+                passData.hdCamera = hdCamera;
+                passData.shaderVariablesGlobal = m_ShaderVariablesGlobalCB;
+                passData.postProcessWidth  = postProcessWidth;
+                passData.postProcessHeight = postProcessHeight;
+                builder.AllowPassCulling(false);
```
FrancescoC-unity, line 1545: No need for this line

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -298,10 +298,15 @@ float GetIndirectSpecularMultiplier(uint renderingLayers)
 
 // Functions to clamp UVs to use when RTHandle system is used.
 
-float2 ClampAndScaleUV(float2 UV, float2 texelSize, float numberOfTexels)
+float2 ClampAndScaleUV(float2 UV, float2 texelSize, float numberOfTexels, float2 multiplier)
```
FrancescoC-unity, line 301: A super nitpicky point here, but to be consistent with how we call these things, I'd prefer `scale `here

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -316,6 +321,17 @@ float2 ClampAndScaleUVForBilinear(float2 UV)
     return ClampAndScaleUV(UV, _ScreenSize.zw, 0.5f);
 }
 
+// This is assuming full screen buffer and .
+float2 ClampAndScaleUVForBilinearPostProcessTexture(float2 UV)
+{
+    return ClampAndScaleUV(UV, _ScreenSize.zw, 0.5f, _RTHandlePostProcessScale.xy);
+}
+
+float2 PostProcessTexelSize()
```
FrancescoC-unity, line 330: I think this doesn't contain what the functions says and also shouldn't (should be previous frame scale to be consistent with rest)

FrancescoC-unity: Probably _PostProcessScreenSize.zw but I don't think we necessarily need this function as we already assume the texel size is in  _ScreenSize.zw

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
kecho: So this is a bit tricky: we want to apply the scale when the target size is the final viewport, and we dont when it is not.
It doesnt make sense to apply the scale when the size is the 'smaller' drsd, since its already scaled.

kecho: I have to figure out whats the best way to approach this

kecho: I think the fundamental problem is that somewhere, we are dealing with scales that are not rounded to the pixel, I will do a check on the entire pipe, we might be able to remove this stupid complexity

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: Maybe the group names are BeforeDRS, AfterDRS?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: So the schedule type goes back to the DRS handler? Itd have to become public then

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: Yes indeed, there is a bit of things we have to do in terms of UI.
For example, we still want the user to specify a fallback config, because existance of <name of secret upscaler> might not be guaranteed. I will do another pass on UI and add the dummy upscaler as a selection, we can do a UI pass.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
kecho: cool, ill do another pass

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
kecho: Ok I have a more elegant fix for this. Upcoming in the next push

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
FrancescoC-unity: Yeah I think that makes sense

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
kecho: Got it

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
FrancescoC-unity: A bit wary on the abbreviation, I think it makes more sense if we flip it on its head and say BeforeDynamicResUpscale  / AfterDynamicResUpscale   

Dunno naming is hard :D

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: done, privatizing :D

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: moved to drs handler

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
kecho: Done

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
kecho: Done. May I ask then, when are render passes culled? Since this pass is not using any render targets (are render passes withouth any targets automatically not culled?)

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
kecho: Done

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
FrancescoC-unity: This indeed feels wrong on second read (was ealry morning first time around :D ) Though I see most of the passes we do that just push constants don't have that set. 

Paging in rendergraph owner @JulienIgnace-Unity

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: Done

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: @FrancescoC-unity I have done some UI settings. First of all, I am calling this upscaler 'Prepost Upscaler' but later we will replace it with the proper name, since legally I cant mention it here ;) so assume that.
The UI Without enabling DRS for the camera now looks like this 
image

1) User selects DRS checkbox on camera
2) User sees the new technique as a checkbox. User presses it
3) Upon pressing it, the Post Anti-Aliasing selected gets indented, and changes its text to "Fallback Post Anti-Aliasing"

The reason for this fallback is because this upscaler could not be supported by the user's hardware. So the option must exist along a fallback.

I added the possibility of "Detected" in the text "Using PrePostProcess Upscaler" text, which would indicate the hardware running this would support it.

Let me know your thoughts

FrancescoC-unity: Sounds good to me as a concept indeed, but given how the settings on the upscaling and dynamic resolution parameters are on the Asset (we don't allow per-camera settings for now), if we can detect what filtering is selected (and we should be able to?), we can skip the checkbox altogether and maybe just display info box + the fallback nomenclature.

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
kecho: Done

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
kecho: FIxed

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
kecho: FIxed

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: Updated this. Will continue conversaton wtih Maddalena

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
```diff
@@ -217,23 +222,37 @@ public void SetReferenceSize(int width, int height, MSAASamples msaaSamples, boo
                 lastFrameMaxSize = new Vector2(GetMaxWidth(), GetMaxHeight());
             }
 
+            var scales = CalculateRatioAgainstMaxSize(m_RTHandleProperties.currentViewportSize);
```
kecho, line 225: Ping @FrancescoC-unity 
What do you think of this refactor? I think is more elegant to just divide times the prescaled max.
This way you preserve the correct rations when you establish a smaller reference size. Let me know.

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -65,6 +65,31 @@ public class DynamicResolutionHandler
         private PerformDynamicRes m_DynamicResMethod = null;
         private static DynamicResolutionHandler s_Instance = new DynamicResolutionHandler();
 
+        /// <summary>
+        /// The scheduling mechanism to apply upscaling.
+        /// </summary>
+        public enum UpsamplerScheduleType
+        {
+            /// <summary>
```
JulienIgnace-Unity, line 73: Just a fyi, if you want to make enum doc more compact you can use this syntax as well:
/// <summary>Description.</summary>
This way it's only 2 lines per value instead of 4:)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
JulienIgnace-Unity: I would tend to agree with Francesco here. I think Before/After works better in all cases.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -550,7 +571,7 @@ static void DoCopyAlpha(in DoCopyAlphaParameters parameters, RTHandle source, RT
         {
             cmd.SetComputeTextureParam(parameters.copyAlphaCS, parameters.copyAlphaKernel, HDShaderIDs._InputTexture, source);
             cmd.SetComputeTextureParam(parameters.copyAlphaCS, parameters.copyAlphaKernel, HDShaderIDs._OutputTexture, outputAlphaTexture);
-            cmd.DispatchCompute(parameters.copyAlphaCS, parameters.copyAlphaKernel, (parameters.hdCamera.actualWidth + 7) / 8, (parameters.hdCamera.actualHeight + 7) / 8, parameters.hdCamera.viewCount);
+            cmd.DispatchCompute(parameters.copyAlphaCS, parameters.copyAlphaKernel, (parameters.hdCamera.actualWidth + 7) / 8, (parameters.hdCamera.actualWidth + 7) / 8, parameters.hdCamera.viewCount);
```
JulienIgnace-Unity, line 574: Is this change expected? Looks like a copy/paste mistake

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -70,6 +70,21 @@ private enum SMAAStage
...
+
+        private enum ResolutionGroup
+        {
+            Downsampled,
+            Full
+        }
+
+        private ResolutionGroup resGroup { set; get; }
+
+        private Vector2Int viewport { get { return resGroup == ResolutionGroup.Full ? m_FullRes : m_DownsampledRes; } }
```
JulienIgnace-Unity, line 86: "viewport" might be a bit confusing.
In Unity terms, viewport.xy is the offset. Here we use it as a size so I first thought there was a mistake.
Maybe "viewportSize" ?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -1300,6 +1321,7 @@ struct DepthOfFieldParameters
             public BlueNoise.DitheredTextureSet ditheredTextureSet;
 
             public HDCamera camera;
+            public Vector2Int viewport;
```
JulienIgnace-Unity, line 1324: Same here (comment on viewport name above)

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
```diff
@@ -217,23 +222,40 @@ public void SetReferenceSize(int width, int height, MSAASamples msaaSamples, boo
...
-                m_RTHandleProperties.rtHandleScale = new Vector4(scaleCurrent.x, scaleCurrent.y, scalePrevious.x, scalePrevious.y);
+                m_RTHandleProperties.rtHandleScale = new Vector4(scales.x, scales.y, scalePrevious.x, scalePrevious.y);
             }
+
+            SetPostProcessScale(m_RTHandleProperties.currentViewportSize.x, m_RTHandleProperties.currentViewportSize.y);
+        }
+
+        internal void SetPostProcessScale(int width, int height)
+        {
+            Vector2 scales = CalculateRatioAgainstMaxSize(new Vector2Int(width, height));
```
JulienIgnace-Unity, line 242: I'm not sure I get this part.
I don't see where we initialize rtHandlePostProcessScale but here we fill it with values from itself?
Am I missing something? :p

JulienIgnace-Unity: Also, from what I see, it's only used to update the global constant buffer.
Is there a point to store this in the RTHandleSystem then?
I'd prefer that we keep RTHandleSystem as agnostic as possible. I don't feel like post processing is something that should be present in there as the RTHandleSystem is not really supposed to know this kind of stuff.

kecho: moved this to the hdcamera

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: Fixed.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: Renamed it to viewportSize

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -1644,6 +1644,34 @@ void RenderGizmos(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle colo
...
+            public int postProcessHeight;
+            public HDCamera hdCamera;
+            public ShaderVariablesGlobal shaderVariablesGlobal;
+        }
+
+        internal void UpdatePostProcessScreenSize(RenderGraph renderGraph, HDCamera hdCamera, int postProcessWidth, int postProcessHeight)
+        {
+            using (var builder = renderGraph.AddRenderPass("Update RT Handle Scales CB", out UpdatePostProcessScreenSizePassData passData))
+            {
+                passData.hdCamera = hdCamera;
```
JulienIgnace-Unity, line 1659: In theory passes with no resource output are supposed to be culled by the graph. It has been disabled for now but I want to revisit this at some point. Since this pass has no output, you should add builder.AllowCulling(false) here to make it future prool :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDDynamicResolutionPlatformCapabilities.cs
```diff
@@ -4,29 +4,19 @@ namespace UnityEngine.Rendering.HighDefinition
...
+        /// True if the render pipeline detected DLSS capable platform. False otherwise.
+        /// </summary>
+        public static bool DLSSDetected { get { return m_DLSSDetected; } }
 
-        public static bool GetFlag(Flag flag)
-        {
-            return (s_FeatureFlags & (1 << (int)flag)) != 0;
-        }
+        /// <summary>
+        /// User only variable, to control wether DLSS is on or off programmatically / from user code..
```
FrancescoC-unity, line 13: I think user code here is implicit given that docs is for user, I'd just say 

"Variable used to control whether DLSS is on or off"

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDDynamicResolutionPlatformCapabilities.cs
```diff
@@ -4,29 +4,19 @@ namespace UnityEngine.Rendering.HighDefinition
...
+        /// </summary>
+        public static bool DLSSEnabled { set { m_DLSSEnabled = value; } get { return m_DLSSEnabled && m_DLSSDetected; } }
```
FrancescoC-unity, line 15: I wonder if we want to split the set and the get here, DLSSEnabled set to true can retrieve immediately false if dlss is not detected, splitting them would give us the option to give explicit naming

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -25,6 +25,18 @@ public enum DynamicResScalePolicyType
...
+    /// The source slots for dynamic resolution scaler. Defines registers were the scalers assigned are stored. By default the User one is always used
+    /// </summary>
+    public enum DynamicResScalerSlot
+    {
+        /// <summary> Scaler slot set by the function SetDynamicResScaler</summary>
+        User,
+        /// <summary> Scaler slot set by the function SetSystemDynamicResScaler</summary>
+        System,
+        /// <summary> total number of scaler slots </summary>
+        Count
```
FrancescoC-unity, line 38: Don't need this in C#, you can get the count with stuff like

Enum.GetValues(typeof(DynamicResScalerSlot) ).Length;

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -257,14 +279,32 @@ public Vector2 GetResolvedScale()
...
+        /// </summary>
+        /// <param name="scaler">The delegate used to determine the resolution percentage used by the dynamic resolution system.</param>
+        /// <param name="scalerType">The type of scaler that is used, this is used to indicate the return type of the scaler to the dynamic resolution system.</param>
+        static public void SetSystemDynamicResScaler(PerformDynamicRes scaler, DynamicResScalePolicyType scalerType = DynamicResScalePolicyType.ReturnsMinMaxLerpFactor)
+        {
+            s_ScalerContainers[(int)DynamicResScalerSlot.System] = new ScalerContainer() { type = scalerType, method = scaler };
+        }
+
+        /// <summary>
+        /// Set the active dynamic scaler.
```
FrancescoC-unity, line 302: I would add a bit more here

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
FrancescoC-unity: Still valid ^ :)

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -78,6 +102,17 @@ private void Reset()
         /// </summary>
         public Vector2Int finalViewport { get; set; }
 
+        /// <summary>
+        /// By default, dynamic resolution scaling is turned off automatically when the source matches the final viewport (100% scale).
+        /// That is, DynamicResolutionEnabled and SoftwareDynamicResIsEnabled will return false if the scale is 100%.
+        /// For certain upscalers, we dont want this behavior since they could possibly include anti aliasing and other quality improving post processes.
+        /// Setting this to true will eliminate this behavior.
+        /// </summary>
+        public bool allowDynamicResolutionOnAllPercentages
```
FrancescoC-unity, line 111: I feel the name might be confusing, I'd change it to something like "Run Upscaler filter on full resolution" (not exactly this, but in this vein)

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -402,14 +482,19 @@ public Vector2Int GetScaledSize(Vector2Int size)
 
         /// <summary>
         /// Applies to the passed size the scale imposed by the dynamic resolution system.
+        /// This function uses the internal resolved scale from the DRS system.
```
FrancescoC-unity, line 485: I would not use acronyms in the doc here (DRS vs Dynamic Resolution Handler)

com.unity.render-pipelines.core/Runtime/Debugging/NvidiaDebugView.cs
```diff
@@ -0,0 +1,266 @@
+using System;
```
FrancescoC-unity, line 1: Question: any reason this is in core vs HDRP given that DLSS is supported only in HDRP?

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
FrancescoC-unity: I think we discussed this, it's been 2 months now and I forgot :D

FrancescoC-unity:
Have we tested this with custom post processes?

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: As discussed left it there. If we dont use it then:
1) We could just not use the Count when we instantiate the array, but then we lose the compiler protection, if say we add another member, the array containing the slots will mismatch. Opening up for a crash
2) If we decide to use reflection, the array containing the slots must then be initialized in a static block, I rather avoid it because every time I use a static block I get some issues with IL2CPP (sometimes reflection data isnt ready) and theres some really strange attributes to add in order to ensure correct initialization
3) An alternative is having a Last member, pointing Last to System, and instantiating the array as Last +1. This sounds a bit too much typing too. In which case for me the best option is #1, drop all compiler safety in favor of cleaner API, since this enum is likely to not be touched further.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDDynamicResolutionPlatformCapabilities.cs
kecho: This has been fixed

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDDynamicResolutionPlatformCapabilities.cs
kecho: Removed user control of this. In a separate PR we must figure out how to override at runtime DRS Settings in general (right now we can't)
For now added camera overrides, but still is not ideal (user might want to turn it all off)

com.unity.render-pipelines.core/Runtime/Textures/RTHandleSystem.cs
kecho: Yes, this was done to preserve correct ratios between destination. What you want, always is a scaled max, a scaled current resolution and compute the ratio of that, because of rounding (a texel of rounding in the smaller res means a lot more texels at full res).
This means that if we dont do this, the aspect ratio at smaller resolutions wabbles. This is only noticible in resolutions < %30 scales. I do think is important to have becasue DLSS does pretty well that low so its likely users will be pushing resolution down a lot.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
kecho: @julienb-unity I still dont know, but I do think I should leave the AllowPassCulling here right?

FrancescoC-unity: Wrong @JulienIgnace-Unity  :D

julienb-unity: @kecho I think you referenced the wrong Julien... 😃

kecho: LOL!! sorry man!

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
kecho: this has been fixed!

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.RenderGraph.cs
kecho: This has been reworked.

JordanL8 (approved):

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: Added the docs. CHanged this to         
/// <summary>
/// Sets the active dynamic scaler slot to be used by the runtime when calculating frame resolutions.
/// See DynamicResScalerSlot for more information.
 /// </summary>

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: Changed this to runUpscalerFilterOnFullResolution

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: switched to full name instead of acronym

FrancescoC-unity (approved):
All good to go [Provided the custom PP change lands shortly after]

TestProjects/HDRP_Tests/Packages/packages-lock.json
```diff
@@ -13,7 +13,7 @@
       "url": "https://packages.unity.com"
     },
     "com.unity.burst": {
-      "version": "1.4.1",
+      "version": "1.5.0",
```
sebastienlagarde, line 16: Don't push this file

sebastienlagarde: (I should remove it from github)

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -159,6 +194,31 @@ private static DynamicResolutionHandler GetOrCreateDrsInstanceHandler(Camera cam
             return instance;
         }
 
+        /// <summary>
+        /// The scheduling mechanism to apply upscaling.
+        /// </summary>
+        public enum UpsamplerScheduleType
+        {
+            /// <summary>
+            /// Indicates that upscaling must happen before post processing.
```
sebastienlagarde, line 203: You should include in the description what is the difference :), i.e with current description it is not possible to  know why chose one option rather than the others

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.cs
```diff
@@ -897,6 +897,12 @@ internal bool AddDiffusionProfile(DiffusionProfileSettings profile)
 
 #endif
 
+        [SerializeField]
+        internal string DLSSProjectId = "231313132";
```
sebastienlagarde, line 901: Maybe use 000000 to highlight that it is not setup? (if it is allowed by the API)

sebastienlagarde (approved):
very minor remark, very great works!

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
```diff
@@ -47,5 +54,5 @@ Use this section to select which custom [Diffusion Profiles](Diffusion-Profile.m
...
 | Shader Variant Log Level              | Use the drop-down to select what information HDRP logs about Shader variants when you build your Unity Project. • Disabled: HDRP doesn’t log any Shader variant information.• Only HDRP Shaders: Only log Shader variant information for HDRP Shaders.• All Shaders: Log Shader variant information for every Shader type. |
 | Lens Attenuation Mode                 | Set the attenuation mode of the lens that is used to compute exposure. With imperfect lens some energy is lost when converting from EV100 to the exposure multiplier, while a perfect lens has no attenuation and no energy is lost. |
-| Light Layer Names                     | The name displayed on Lights and Meshes when using [Light Layers](Light-Layers.md). |
-| Decal Layer Names                     | The name displayed on decals and Meshes when using [Decal Layers](Decal.md). |
+| Use DLSS Custom Project Id            | When set to true, a custom project ID can be specified for the NVIDIA Deep Learning Super Sampling module. This setting is only available when the NVIDIA native package is enabled on the project. |
```
JordanL8, line 57: ```Controls whether to use a custom project ID for the NVIDIA Deep Learning Super Sampling module. If you enable this property, you can use **DLSS Custom Project Id** to specify a custom project ID.<br/>This property only appears if you enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project.```

What about when its disabled? Does it just use some default value or generate a unique one for your project?

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
```diff
@@ -47,5 +54,5 @@ Use this section to select which custom [Diffusion Profiles](Diffusion-Profile.m
...
+| DLSS Custom Project Id                | The custom project ID to be used for the NVIDIA Deep Learning Super Sampling module. This setting is only available when the NVIDIA native package is enabled on the project. |
```
JordanL8, line 58: ```The custom project ID to use for the NVIDIA Deep Learning Super Sampling module. This setting is only available when the NVIDIA native package is enabled on the project.<br/>This property only appears if you enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project.```

JordanL8 (changes requested):
Added a couple of suggestions and have one question

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
kecho: No, so if its disabled, it will utilize the default project ID which is internal in unity.

JordanL8 (approved):

kroatoa (approved):

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
```diff
@@ -47,5 +54,6 @@ Use this section to select which custom [Diffusion Profiles](Diffusion-Profile.m
...
 | Lens Attenuation Mode                 | Set the attenuation mode of the lens that is used to compute exposure. With imperfect lens some energy is lost when converting from EV100 to the exposure multiplier, while a perfect lens has no attenuation and no energy is lost. |
-| Light Layer Names                     | The name displayed on Lights and Meshes when using [Light Layers](Light-Layers.md). |
-| Decal Layer Names                     | The name displayed on decals and Meshes when using [Decal Layers](Decal.md). |
+| Use DLSS Custom Project Id            | Controls whether to use a custom project ID for the NVIDIA Deep Learning Super Sampling module. If you enable this property, you can use **DLSS Custom Project Id** to specify a custom project ID.<br/>This property only appears if you enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project. |
```
JordanL8, line 57: From reading the stuff from Paul in slack, I think the first bit is fine. We should just add information about what happens when disabling the property. So:
`Controls whether to use a custom project ID for the NVIDIA Deep Learning Super Sampling (DLSS) module. If you enable this property, you can use **DLSS Custom Project Id** to specify a custom project ID. If you disable this property, Unity generates a unique project ID. <br/>This property only appears if you enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project.`

I also think its important to add info about the ID being unique but I'm not sure if it _should_ or _must_ be unique yet. I've asked in slack

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Asset.md
```diff
@@ -79,6 +79,10 @@ These settings control the draw distance and resolution of the decals atlas that
 | **Property**                    | **Description**                                              |
 | ------------------------------- | ------------------------------------------------------------ |
 | **Enable**                      | Enable the checkbox to make HDRP support dynamic resolution in your Unity Project. |
+| **- Enable DLSS**               | Enable the checkbox to make HDRP support NVIDIA Deep Learning Super Sampling (DLSS).<br/>This property only appears if you enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project. |
+| **-- Performance Quality**      | Use the drop-down to select which performance mode DLSS operates on. The options are:<br/>&#8226; **Balanced**: - Balances performance with quality.<br/>&#8226; **MaxPerf**: - Fast performance, lower quality.<br/>&#8226; **MaxQuality**: - High quality, lower performance.<br/>&#8226; **UltraPerformance**: - Fastest performance, lowest quality.<br/>&#8226; **UltraQuality**: - Highest quality, lowest performance. |
```
remi-chapelain, line 83: > | **-- Performance Quality**      |

I Believe this has change to "mode" now, to better match what we're doing in raytracing perf/quality mode

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Asset.md
```diff
@@ -79,6 +79,10 @@ These settings control the draw distance and resolution of the decals atlas that
...
+| **-- Use Optimal Settings**     | Enable the checkbox to make DLSS control the Sharpness and Screen Percentage automatically. |
+| **-- Sharpness**                | Use the slider to set the pixel sharpness of the DLSS upscaler. |
```
remi-chapelain, line 85: I suggested another tooltip on editor side to prevent us using the same word (shaprness) we're trying to define and give a bit more info on what's at stake here. 

"Controls how the DLSS upsampler will render edges on the image. More sharpness usually means more contrast and clearer image but can increase flickering and fireflies. This setting is ignored if ‘Use Optimal Settings’ is checked."

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Camera.md
```diff
@@ -37,6 +37,12 @@ The HDRP Camera shares many properties with the Unity's [standard Camera](https:
...
+| **Allow DLSS**                  | Controls whether this camera can use NVIDIA Deep Learning Super Sampling (DLSS). To use DLSS, enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project and enable DLSS in your [HDRP Asset](HDRP-Asset.md). |
+| **- Use Custom Quality**        | Indicates whether this Camera overrides the DLSS quality mode specified in the [HDRP Asset](HDRP-Asset.md). If you enable this property, HDRP uses the below **Performance Quality** property. |
+| **-- Performance Quality**      | Specifies the performance quality mode override for performance quality mode for DLSS.The options are:<br/>&#8226; **Balanced**: - Balances performance with quality.<br/>&#8226; **MaxPerf**: - Fast performance, lower quality.<br/>&#8226; **MaxQuality**: - High quality, lower performance.<br/>&#8226; **UltraPerformance**: - Fastest performance, lowest quality.<br/>&#8226; **UltraQuality**: - Highest quality, lowest performance.<br/>This property only appears if you enable **Use Custom Quality**. |
```
remi-chapelain, line 42: Same comment here about naming of the property and only 4 mode instead of 5

com.unity.render-pipelines.high-definition/Documentation~/deep-learning-super-sampling-in-hdrp.md
```diff
@@ -0,0 +1,63 @@
...
+To change the DLSS quality mode for your whole project:
+
+1. Select the HDRP Asset that has DLSS enabled and view it in the Inspector.
+2. Go to **Rendering** > **Dynamic Resolution** > **DLSS** and set the **Performance Quality** property to the quality mode you want.
+
+To override the DLSS quality mode for a particular Camera:
+
+1. In the Hierarchy or Scene view, select a Camera and view it in the Inspector.
+2. Select **Use Custom Quality**.
+3. Set the **Performance Quality** property to the quality mode you want.
```
remi-chapelain, line 63: Same comment about property naming "mode"

remi-chapelain (approved):
PR is in good shape. Go for merge once little things from documentation will be updated. 💚 

Did a few performances tests to ensure that there's a net gain (up to 35/40%) even though we trust Nvidia for this. 

A few things were improved including : 
- Issue about quality settings not taken into account
- More info on tooltips
- Quality settings naming for consistency with RTX naming
- Ease of use for installing Nvidia package (without having to go to package manager) and switching architecture to 64 bits

image

image

Will enter issues for those 2 concerns that doesn't block nor prevent anyone from using DLSS. 
- The order of the quality settings is misleading. It would make more sense to go from Best perf to best quality (or inverse) instead of jumping from perf to balance, to quality, to perf again. This require a cpp PR, can be done later, not blocking anything. 
- Also found [an issue about black pixels](https://unity.gyazo.com/d7bb07ade9872dd21eb26fc2f89b1b6b) on one frame when switching DLSS ON that would be related to how we handle RT in Vulkan. 

image


Here's the [test document](https://docs.google.com/document/d/1W828t9dDrkSPVleLCfi-rD08aw-qCu7xUtr6oqu-9nA/edit#) for more information about what's tested and discussions.

PR #3489, opened 2021-02-12
[VFX/URP] 2DRenderer integration

PaulDemeulenaere:
---
### Purpose of this PR
Basic integration of 2D Renderer :
- Add missing pass "Universal2D"
- Add sorting control in renderer editor
_sorting_layer

*Other* :
- Fix line rendering issue here 469280a2d517862dfc82fdeebc7c82de46d6c5c3
_URP_issue_line

It was an equivalent issue of SubPixelAA (see [here](https://docs.google.com/presentation/d/1scY5qsRrkB1GbJ2FwXVKTT8JU_NEa30bMNazxrf4ED8/edit#slide=id.g9dbad1aa66_0_11))

---
### Testing status
Added a graphic test :
_Graphic_Test

It covers :
- The standard primitive output : Mesh, Cube, Planar, Line, Point, with Opaque & Transparent.
- Basic integration of shaderGraph (the wagon train, is a particle strip + shaderGraph)
- Ordering using either layerId & layerOffset

Public Yamato ⏳ 
[Internal ](https://github.cds.internal.unity3d.com/unity/ScriptableRenderPipelinePrivate/pull/83)Yamato ⌛ 

---
### Comments to reviewers
@JordanL8 I updated the documentation at https://github.com/Unity-Technologies/Graphics/pull/3489/commits/8be763fe63b09d021eebdc4dcd5abefe7ec15acf but it's minimal, I'm expecting some update after the QA pass.

@Rus-Unity @VladNeykov Feel free to use this [channel](https://unity.slack.com/archives/G01DWMX1Z43/p1613377871002700) if needed : I'm not really confortable with 2D in general, I'm maybe missing obvious issues.

@julienf-unity I added `PassForwardCommon.template` which is the same content than `PassForward.template` to share code between "classic" forward & `Universal2D`

**Side note:**  I would like to filter out sortingPriority/sortingInLayer only when there is a Renderer2DData somewhere but we don't have a dedicated API (yet ?), see this [conversation](https://unity.slack.com/archives/C89KFUUCT/p1612799052163400).


JordanL8 (approved):

VladNeykov:
Looking good, only 1 small VFX-specific issue noted on the [Test doc](https://docs.google.com/document/d/1R-ABOwJ8IGhrc0Mug8RMMb2gA0RwzfS5PxCzaBm_27Q/edit#)
Not sure what the DoD is, so we've noted also a few potentially missing features which, if outside of the scope of this PR, can maybe be added to the backlog.

com.unity.visualeffectgraph/Shaders/RenderPipeline/Universal/Templates/VFXParticleBasicCube.template
```diff
@@ -8,5 +8,6 @@
 		${VFXInclude("Shaders/ParticleHexahedron/PassDepth.template"),IS_OPAQUE_PARTICLE}
 		${VFXInclude("Shaders/ParticleHexahedron/PassBasicForward.template")}
 		${VFXInclude("Shaders/ParticleHexahedron/PassShadowCaster.template"),USE_CAST_SHADOWS_PASS}
+		${VFXIncludeRP("Templates/ParticleHexahedron/PassForward2D.template")}
```
julienf-unity, line 11: So we generate pass 2D always with URP ? 

Isnt it mutually exclusive with regular forwardPass (in that case it's just a matter a renaming the pass) in a project or can a project have both 2D and 3D renderer?

Also can we filter the 2D pass if 2D renderer is not used or is this unknown ?

julienf-unity: Is it done like this in unlit shader graph? (Always generate both passes)

PaulDemeulenaere: > So we generate pass 2D always with URP ?

Yes, we are always generating both, at some point, I would like to add a specific option to generate one, another or both but it would require to move SRP output in URP package (to allow serialization of specific data for URP).

> Also can we filter the 2D pass if 2D renderer is not used or is this unknown ?

For now, it's unknown, I've asked [here](https://unity.slack.com/archives/C89KFUUCT/p1612799052163400) but since we can switch the URP asset in runtime, checking the actual list of target isn't enough.

> Isnt it mutually exclusive with regular forwardPass (in that case it's just a matter a renaming the pass) in a project or can a project have both 2D and 3D renderer?

You can definitely have 2D & 3D renderer cohabiting, a camera can render using 2D while another, in the same scene, can render using 3D. Not sure if it's a use case but you can imagine render your 3D Scene and a minimap using 2D of the same scene.
By the way, in the project VisualEffect_URP, we are now mixing the classic 3D forward renderer and the 2D Renderer.

> Is it done like this in unlit shader graph? (Always generate both passes)

A shadergraph unlit is visible (by default, there is maybe a node settings somewhere which provide exclusion).
If you are taking the example of [Lit.shader](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Shaders/Lit.shader) both pass are declared :
- Universal2D https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Shaders/Lit.shader#L336
- UniversalForward https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Shaders/Lit.shader#L372 (!= UniversalForwardOnly since it supports deferred)

... but for `Unlit.shader`, the "main" pass doesn't declare any `LightMode` :
https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Shaders/Unlit.shader#L28

... Looking at the documentation : https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@10.2/manual/urp-shaders/urp-shaderlab-pass-tags.html
=> ```If you do not set the LightMode tag in a Pass, URP uses the SRPDefaultUnlit tag value for that Pass.```
I missed this detail, maybe we can use the "SRPDefaultUnlit" fallback for both passes, it could drastically reduce the amount of change for this PR but, on the other hand, it will prevent us to exclude a pass later.

See also integration of `SRPDefaultUnlit` in
- URP renderer : https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs#L47
- 2D renderer : https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Runtime/2D/Passes/Render2DLightingPass.cs#L29

VladNeykov (approved):
No outstanding issues found. As mentioned in the [test doc](https://docs.google.com/document/d/1R-ABOwJ8IGhrc0Mug8RMMb2gA0RwzfS5PxCzaBm_27Q/edit?ts=60364ead), there are a number of features integral to the 2D workflow which we currently do not support and should probably take a swing at in a future PR.
Thanks for adding this support, it was easily in the top most requested tasks by users.

PaulDemeulenaere:
@Rus-Unity & @julienf-unity : Do you have a moment to take a new look at this MR ? Feel free to poke me on slack.
The build standalone is failing from the last week merge, I'm waiting for trunk merge to re-update this change.

com.unity.visualeffectgraph/Shaders/RenderPipeline/Universal/Templates/VFXParticleBasicCube.template
PaulDemeulenaere: We still have the possibility to use `SRPDefaultUnlit` instead of `UniversalForwardOnly`, it will  reduce the file change but it will prevent us to do any specific change to 2D or 3D passes in the future.

UnityPaul (approved):

PaulDemeulenaere:
Test has been re-run at https://github.com/Unity-Technologies/Graphics/pull/3489/commits/452f66ae6ad8ca598d9d3f9119bfaa2088a5a617 🟢 
Ready to be merge.

PR #3490, opened 2021-02-12
[Shader Graph][12.x.x][2021.2] Make sure to emit float values for properties as decimal

jessebarker:
### Purpose of this PR
This PR addresses https://fogbugz.unity3d.com/f/cases/1313932/

Backports:
- #3490
- #5969
- #5968

The root issue is that the UI is capturing small decimal float values in scientific notation (e.g., 0.00001 -> 1E-05).
Property declarations in ShaderLab do not accept this syntax (other intiializations and constant declaration in the body of the shader are fine with scientific notation).  The fix is to update Vector1 and Color properties to use NodeUtils.FloatToShaderValueShaderLabSafe rather than NodeUtils.FloatToShaderValue (other properties were already safe).
N.b. Not all Shader Graph properties translate to ShaderLab properties (e.g., Gradient)

---
### Testing status

- [x] Testing according to https://fogbugz.unity3d.com/f/cases/1313932/ (float/Vector1 only called out)
- [x] Tested color and vector property types (caught the color case this way as well as inspection of the code).
Yamato tests coming

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Graphs/ColorShaderProperty.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/ColorShaderProperty.cs
@@ -36,7 +36,7 @@ namespace UnityEditor.ShaderGraph.Internal
 
         internal override string GetPropertyBlockString()
         {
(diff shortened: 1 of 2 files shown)


joshua-davis (approved):

LandonTownsendUnity (approved):
No regressions I don't think, and it fixes the bugs

cdxntchou (approved):

PR #3501, opened 2021-02-15
assign renderpipelineasset and quality of life improvements

martint-unity:
---
### Purpose of this PR
Adding quality of life improvements.
On A Renderer you can now click this button
image

to get a UI to assign and/or un-assign from Pipeline Assets
image

On a Pipeline Asset you now have these 2 buttons
image
Which will let you easily assign the pipeline asset you have selected to either the Project Settings Graphics or to the current Quality level.

On a Camera you can now directly jump to the assigned renderer to quickly see / change settings
image

---
### Testing status
Katana

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/ScriptableRendererDataEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/ScriptableRendererDataEditor.cs
@@ -135,19 +135,6 @@ namespace UnityEditor.Rendering.Universal
                 }
             }
         }
-
-        // void AssignRendererToRPAsset(string rpName)
-        // {
-        //     foreach (var asset in rpAssets)
-        //     {
-        //         if (asset.name == rpName)
-        //         {
-        //             Debug.Log("Adding");
-        //             asset.AddRendererToRendererDataList(target as ScriptableRendererData);
-        //             return;
-        //         }
-        //     }
-        // }
     }
 
     [CustomEditor(typeof(ScriptableRendererData), true)]
@@ -180,7 +167,6 @@ namespace UnityEditor.Rendering.Universal
         [SerializeField] private bool falseBool = false;
         List<Editor> m_Editors = new List<Editor>();
 
-        string[] options = {};
         private void OnEnable()
         {
             m_RendererFeatures = serializedObject.FindProperty(nameof(ScriptableRendererData.m_RendererFeatures));


com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -295,6 +293,80 @@ void DrawQualitySettings()
...
+                    message = $"This will assign {asset.name} to the current Graphics Settings.\nRight now nothing is assigned. Do you want to assign this asset?";
+                    buttonText = "Assign";
+                }
+                else
+                {
+                    message = $"This will assign {asset.name} to the current Graphics Settings.\nRight now {pipelineAsset.name} is assigned. Do you want to replace it?";
+                    buttonText = "Replace";
+                }
+
+                if (GraphicsSettings.defaultRenderPipeline != (RenderPipelineAsset)target)
```
alex-vazquez, line 321: You already had the cast in `asset`, you can also extract the cast outside of both buttons.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -295,6 +293,80 @@ void DrawQualitySettings()
...
+                if (pipelineAsset == null)
+                {
+                    message = $"This will assign {asset.name} to the current Graphics Settings.\nRight now nothing is assigned. Do you want to assign this asset?";
+                    buttonText = "Assign";
+                }
+                else
+                {
+                    message = $"This will assign {asset.name} to the current Graphics Settings.\nRight now {pipelineAsset.name} is assigned. Do you want to replace it?";
+                    buttonText = "Replace";
+                }
```
alex-vazquez, line 319: You can extract this to it's own method:
```
bool PromptAssingTo(string assetName, string title, RenderPipelineAsset asset)
        {
            if (asset == null)
            {
                return EditorUtility.DisplayDialog(
                    $"Assign to {title}",
                    $"This will assign {assetName} to the current {title}.\nRight now nothing is assigned. Do you want to assign this asset?",
                    "Assign", "Cancel");
            }
            
            return EditorUtility.DisplayDialog($"Assign to {title}", $"This will assign {assetName} to the current {title}.\nRight now {asset.name} is assigned. Do you want to replace it?", "Replace", "Cancel");
        }
```

Then you can simply:

`if (GraphicsSettings.defaultRenderPipeline != asset && PromptAssingTo(asset.name, "Graphics Setting", pipelineAsset))`

and 

`if (QualitySettings.renderPipeline != asset && PromptAssingTo(asset.name, "Quality Settings", qualityAsset))`

avoiding also the string operations if the conditions are not matched.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -295,6 +293,80 @@ void DrawQualitySettings()
...
+
+                if (QualitySettings.renderPipeline != asset)
+                {
+                    if (EditorUtility.DisplayDialog("Assign to Quality Setting", message, buttonText, "Cancel"))
+                    {
+                        // Reflection to be able to get the QualitySettings Object.
+                        // We need the object to do record an Undo action here when switching RenderPipelineAsset
+                        Object obj = QualitySettings.GetQualitySettings();
+                        Undo.RecordObject(obj, $"Assigned {asset.name}");
```
alex-vazquez, line 358: Undo Stack should be translated: `L10n.Tr("Assigned  ") + asset.name`

com.unity.render-pipelines.universal/Editor/ScriptableRendererDataEditor.cs
```diff
@@ -1,14 +1,142 @@
...
+    static class RemoveRendererDataDialogText
+    {
+        public static readonly string title = "Remove Renderer Data from URP Asset";
+        public static readonly string message = "This change might change the renderer data list entries";
+        public static readonly string proceed = "Proceed";
+        public static readonly string ok = "Ok";
+        public static readonly string cancel = "Cancel";
+    }
+
+    public class AssignToRendererDataWindow : EditorWindow
```
alex-vazquez, line 24: - I think it would be better if this class is moved to it's own file, also I do not see dependencies on Universal, if we use the base class `RenderPipelineAsset`, so we can move it to core directly.

- I opened the window, and it has a lot of height, that it wasn't usefull. What about using a [PopupWindow](https://docs.unity3d.com/ScriptReference/PopupWindow.html)? Where you can calculate all the Toggle + the Buttons and return the proper height?

- Missing documentation for public class

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -494,98 +566,108 @@ void DrawAdaptivePerformance()
...
+            if (list.index != m_DefaultRendererProp.intValue)
             {
-                bool shouldUpdateIndex = false;
-                // Checking so that the user is not deleting  the default renderer
-                if (li.index != m_DefaultRendererProp.intValue)
-                {
-                    // Need to add the undo to the removal of our assets here, for it to work properly.
-                    Undo.RecordObject(target, $"Deleting renderer at index {li.index}");
+                // Need to add the undo to the removal of our assets here, for it to work properly.
+                Undo.RecordObject(target, $"Deleting renderer at index {list.index}");
```
alex-vazquez, line 646: Localization on undo stack is missing

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -443,6 +444,27 @@ public ScriptableRenderer GetRenderer(int index)
...
         }
 
+        /// <summary>
+        /// Returns a renderer data from the current pipeline asset
+        /// </summary>
+        /// <param name="index">Index to the renderer data. If invalid index is passed, the default renderer is returned instead.</param>
+        /// <returns></returns>
+        public ScriptableRendererData GetRendererData(int index)
+        {
+            if (index == -1)
```
alex-vazquez, line 454: Why the case that I do `GetRenderData(-2)` returns `null` and if I do `GetRenderData(-1)` gives me the default?

In which cases this would happen?

Also if -1 is a special case to return the default one, we can add it as default value of the argument, this will allow us to get the default by simply calling the method.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -478,7 +510,56 @@ string RendererDataDisplayName(ScriptableRendererData data)
...
+            if (CanRemoveFromRendererDataList(rendererData))
+            {
+                int index = RendererDataIndex(rendererData);
+                if (index != -1)
+                {
+                    UpdateDefaultRendererValue(index);
+                }
+
+                // Lets remove the renderer data
+                var rendererDataList = new List<ScriptableRendererData>(m_RendererDataList);
```
alex-vazquez, line 531: If the index is -1 you do not need to copy and delete, besides that you can simply call:

`rendererDataList.RemoveAt(index)`, and avoid another lookup for the renderer.

alex-vazquez (changes requested):

ValentinHume (changes requested):
The added buttons work but there are a few issues :
Undo not working after assigning a Renderer Data to a Render Pipeline Asset
Cannot build the player, seems to be linked to the editor scripts
No warning when trying to unassign a Renderer Data set as Default 
Here's is the test document with more infos : https://docs.google.com/document/d/1CSQO_nw_CWv4IzzAFmzepCQADPfFkv41i2xYzhGKdq4/edit?usp=sharing

martint-unity:
Closing due to changes in UX needs to be sorted

PR #3505, opened 2021-02-15
Hd/deferred debug visualization

JarkkoUnity:
### Purpose of this PR
Adds support to visualize materials that are rendered in deferred pass. Materials rendered in deferred pass show in green while the rest shows in red in this mode. You can enable this mode in Unity editor: Window => Analysis => Render Pipeline Debugger => Material => Properties => DeferredMaterials

---
### Testing status
Created a simple test scene to test the following shader combos for proper result:
- HDRP/Lit, HDRP/LitTessellation, HDRP/LayeredLit, HDRP/LayeredLitTessellation, for both opaque and transparent modes (green when opaque, red when transparent)
- HDRP/TerrainLit (always green)
- HDRP/Unlit (always red)
- Shader Graph with HDRP/Lit opaque and transparent modes

Also HDRP SampleScene seems to properly visualize deferred vs forward materials


JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.hlsl
```diff
@@ -83,6 +83,14 @@ void GetPropertiesDataDebug(uint paramId, inout float3 result, inout bool needLi
             result = float3(1.0, 0.0, 0.0);
 #else
             result = float3(0.0, 0.0, 0.0);
+#endif
+            break;
+
+        case DEBUGVIEWPROPERTIES_DEFERRED_MATERIALS:
+#ifdef _DEFERRED_CAPABLE_MATERIAL
+            result = float3(0.0, 1.0, 0.0);
```
sebastienlagarde, line 91: So you miss to take into account the FrameSettings.
In FrameSettings we can disable or not the render in deferred. If FrameSettings are setup to forward then we should show material as forward.
result = debugDisplay.IsLitShaderModeDeferred ? float3(0.0, 1.0, 0.0) : float3(1.0, 0.0, 0.0);

And you need to add it to the command buffer.

+ TerrainLit as you mention :)

JarkkoUnity: I intentionally didn't consider the FrameSettings for the visualization. I.e. if the purpose of this mode is to inform user which materials _can_ be rendered in deferred, but not necessarily which ones are rendered in deferred for the current camera. So that user can see how optimized the scene is for deferred rendering regardless of the active mode. I can change this and update the comment though, if it's preferred to obey the active mode instead.

sebastienlagarde (approved):

sebastienlagarde:
Note: Yamato is good, the issue are coming from master branch, infrastructure or C++, not related to this PR, ready to go after QA approval.

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.hlsl
JarkkoUnity: FYI, after discussion with artists changed the visualization to take FrameSettings into account

TomasKiniulis (changes requested):
From functionality perspective this looks solid! Tested Deferred and Forward materials in HDRP_Template and Amalienborg palace.
Made sure, that feature works with multiple Scene/Game views switching HDRP asset between Forward/Deferred/Both and with "Both" mode debug view properly displays changed Default Frame Settings and Camera frame settings.
https://user-images.githubusercontent.com/50582134/109181462-9abee900-7794-11eb-8de3-d6e12635d92e.mp4

Some minor nitpicks.
If possible, would be great to have debug view transparent overlay on top of regular shading instead of having it completely solid. I think in the current state:
- Users will tend to switch between this debug view and no debug view just to get better idea what are they looking at.
- Current colors are hard on the eyes. Perhaps red and green could be less saturated? Or if we had debug view transparent that would greatly reduce eye strain.

Also, I think this needs documentation in Debug Modes Material section: https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.3/manual/Render-Pipeline-Debug-Window.html

sebastienlagarde:
> * If possible, would be great to have debug view transparent overlay on top of regular shading

yes, we have discuss it and decided to not go with it for this task as it is way more complex to implement this way and this task was onboarding for Jarkko.

So @JarkkoUnity  all is missing is the doc part:
>Also, I think this needs documentation in Debug Modes Material section:
thanks

TomasKiniulis:
> > * If possible, would be great to have debug view transparent overlay on top of regular shading
> 
> yes, we have discuss it and decided to not go with it for this task as it is way more complex to implement this way and this task was onboarding for Jarkko.

Okay, what about reducing saturation of the colors? Something like in the picture attached? Comparing between the top and bottom I'd choose bottom :D (and again sorry for nitpicking)
Untitled-1

sebastienlagarde:
We could replace the color, but then I would prefer to ask the artists what they like (we use this same red and green color at various place so if we change it for one effect we must change it for all effect. We can do that in a separate PR about improving debug menu.

TomasKiniulis:
Spoke with Jarkko. 
- Looks like we don't have doc for any Properties options, so these options should be documented in a separate PR.
- After testing PR in Spaceship demo it exposed that Transparent Additive Materials are displayed yellow instead of pure red. Which does not appear to be expected.
https://user-images.githubusercontent.com/50582134/109281060-80345080-7824-11eb-83a3-58311934ac44.mp4

sebastienlagarde (approved):

sebastienlagarde:
@TomasKiniulis is it good with lastest change of Jarrko to move forward? We will keep this PR and mode small for now. thanks

TomasKiniulis (approved):
Looks good. Checking with Jarkko, seems like there's no good way to address the yellow color with transparent additive at the moment. Agree that it should be kept as is.

PR #3510, opened 2021-02-15
Remove unnecessary project files from split projects.

hdb-unity:
Removes auto-generated files that are not required to allow split projects to function correctly.


hdb-unity:
5x failed CI
- 3x CUSTOM-REVISION not supplied (mistaken run)
- 2x failures that have no data to compare from base

phi-lira (approved):

PR #3515, opened 2021-02-16
Allow Normal Bias on APV to change without needing a rebake

FrancescoC-unity:
Pretty much as the title says, before this change, to see the effects of changing normal bias previously a rebake was needed. This change allows to see the effect at the first scene/game view refresh after the change.


FrancescoC-unity:
NOTE: There is now an issue with serialization, do not merge please.

sebastienlagarde (approved):

PR #3518, opened 2021-02-16
[Yamato] Use revision variables in update jobs

sophiaaar:
### Purpose of this PR
* The update projects live in a different repository, so it's convenient to be able to edit the revision to checkout via the Yamato UI when testing new changes.
* It also means the Yamato jobs won't break if someone pushes a change to the update project repository

---
### Testing status
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/yamato%252Fadd-vars-to-update-job
* Tested with a defined variable
* Tested editing variable in the UI

---
### Comments to reviewers

### DIFF
--- a/.yamato/_abv.yml
+++ b/.yamato/_abv.yml
@@ -41,8 +41,6 @@ all_project_ci_nightly_trunk:
          rerun: on-new-revision
       -  path: .yamato/all-urp_performance_lostcrypt.yml#PR_URP_Performance_LostCrypt_trunk
          rerun: on-new-revision
-      -  path: .yamato/all-urpupdate.yml#PR_URPUpdate_trunk
-         rerun: on-new-revision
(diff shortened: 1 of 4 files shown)


.yamato/config/universal_perf_boatattack.metafile
```diff
@@ -1,5 +1,6 @@
 project:
   name: URP_Performance_BoatAttack
+  repo: BoatAttack
```
kivistikliis, line 3: Do we use this somewhere? Just wondering, but it's not a blocker

kivistikliis (approved):

.yamato/config/universal_perf_boatattack.metafile
sophiaaar: oh I didn't realise I even changed that, thank you

PR #3539, opened 2021-02-18
Move resolver resize logic to HDRP

robin-demoor:

### Checklist for PR maker

---
### Purpose of this PR
We currently perform the resolver resizing logic of the RTHandles in the engine itself. To make VT SRP agnostic we want to perform as little logic in the engine and move that to HDRP itself so users can make their own VT system without having to mimic/use the RTHandles system. This logic will be removed from core once this PR has landed.
---
### Testing status

---
### Comments to reviewers


sebastienlagarde (approved):

sebastienlagarde:
Hey, it miss a changelog, also if I understand correctly it is also related to a C++ change? do you have a link for the PR? thanks

robin-demoor:
I will add a changelog and this is the ono PR https://ono.unity3d.com/unity/unity/pull-request/121456/_/graphics/vt-resolver-update-size

This PR will have to land before we land the ono one.

RemyUnity:
So, this PR doesn't introduce any change from user perspective ?
Should QA check test here, or pay a particular attention ?

robin-demoor:
This PR itself doesn't change anything. The trunk PR that follows will change the behavior for people who have a custom pipeline with Virtual Texturing.

I tested locally with scripts that resize everything to a few random both valid and invalid sizes and that seemed to work/log what is expected. I don't think there is much extra to test for.

robin-demoor:
Closing PR for now as it is blocked by https://fogbugz.unity3d.com/f/cases/1306556/

RemyUnity:
> I tested locally with scripts that resize everything to a few random both valid and invalid sizes and that seemed to work/log what is expected. I don't think there is much extra to test for.

Thanks ! Next time remember to add this information directly in the PR description :)

RemyUnity (approved):

JulienIgnace-Unity (commented):
I'm not sure I understand the purpose of this PR.
What was on the C++ side that is going to be removed?
If I understand the PR description it seems that there was some logic based on the RTHandle system behavior? Is that correct?
The width and height that we passed to the resolver UpdateSize function were just the current viewport size so I'm not sure where the RTHandle specific part is here. I'm probably just missing something though :)
Is this the part where we just ensure that the resolver size can only grow?  (cf CalculateResolverDimensions). Isn't that a behavior that you'd want regardless of using RTHandle or not? (I assume that by having it only grow, it avoids reallocation of internal buffers when switching between cameras with different resolution).

robin-demoor:
@JulienIgnace-Unity The issue is that we were doing this "not downsizing" in Unity Engine and had complaints from people that the resolver was not the size they requested because they didn't know this logic was happening. (Someone was implementing Virtual Texturing in URP) That is why we chose to make the engine side as low-level as possible and we don't want to perform this logic without people knowing.  So we moved the logic to HDRP and the engine side now just accepts whatever width/height you give it.  TLDR; Nothing changes for people using HDRP this is just a change for people implementing VT elsewhere.

(trunk PR: https://ono.unity3d.com/unity/unity/pull-request/121456/_/graphics/vt-resolver-update-size)

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
Why is there change of universal  shader graph in this PR ?

JulienIgnace-Unity:
> Why is there change of universal shader graph in this PR ?

There was a merge mistake that broke the PR and I redid the merge. Don't know why it got there. It's just a case change for .shadergraph though so probably not an issue.

robin-demoor:
I made a mistake when merging in master which Julien Ignace helped fix but the end result still has these 2 files with a change in case. The .ShaderGraph at the end is not .shadergraph

sebastienlagarde:
> I made a mistake when merging in master which Julien Ignace helped fix but the end result still has these 2 files with a change in case. The .ShaderGraph at the end is now .shadergraph

ok got it, it is currently the githook system who have automatically fixed it

hdb-unity (approved):
Not tested locally (based on earlier discussion in this PR).

Checking changes to UniversalGraphicsTest

Functionally, don't think it will cause an issue...

If possible, can you restore the filename as it is in master so it won't make any changes when merged?

sebastienlagarde:
Tried to fix the shader graph files issue but don't work. It shouldn't be a problem in practice, the files are unchanged.

PR #3540, opened 2021-02-18
Graphics SG unit test for IsBack/FrontFace

JarkkoUnity:
### Purpose of this PR
Test that IsFrontFace shader graph node is working properly for Lit and Unlit shaders

---
### Testing status
Ran the unit test in editor (Win/DX11) and Yamato for different platforms


remi-chapelain (commented):
Do you think it will be a good addition for the test to include a double sided material ? 
For back/front faced culled material, the node will always return the same result for the object so maybe having a double sided alpha clipped sphere to test a most complex case can be interesting ? 

image

remi-chapelain (approved):
Looks good to me ! Thanks for taking the time to add the more complex usecase !

sebastienlagarde (approved):

PR #3542, opened 2021-02-18
[Xpipeline]Improvements on Volume Additional Properties

alex-vazquez:
### Purpose of this PR
- Volume components without custom editor and parameters with `[AdditionalProperty]` were not being handled by their default editor
- Removing the `BeginAdditionalProperty` and `EndAdditionalProperty` of custom inspectors as this is handled by the default/base `VolumeComponentEditor` class, also this behaviour has been changed to a `IDisposable`
- Now, the base method `PropertyField` returns if a property has been drawn, this allow inherited classes to simply add check based on other properties(for example `DOFEditor`)
- Fix correct alignment of Additional Properties.

---
### Testing status
- Check that the properties that are defined as additional are shown/hidden when they 'Show Additional Properties' is changed.
- Check that the additional properties are correctly right-aligned to non additional properties


com.unity.render-pipelines.high-definition/Editor/PostProcessing/DepthOfFieldEditor.cs
```diff
@@ -120,15 +120,10 @@ void DrawQualitySettings()
...
-                if (BeginAdditionalPropertiesScope())
-                {
-                    PropertyField(m_Resolution);
-                    PropertyField(m_HighQualityFiltering);
-                    PropertyField(m_PhysicallyBased);
-                    if (m_PhysicallyBased.value.boolValue == true)
-                        EditorGUILayout.HelpBox(Styles.InfoBox, MessageType.Info);
-                }
-                EndAdditionalPropertiesScope();
+                PropertyField(m_Resolution);
```
JulienIgnace-Unity, line 123: With this change, the help box will still show up with or without additional properties enabled so we still need a scope here.

com.unity.render-pipelines.high-definition/Editor/Sky/AtmosphericScattering/FogEditor.cs
```diff
@@ -157,18 +150,16 @@ public override void OnInspectorGUI()
...
-                        PropertyField(m_Anisotropy, s_AnisotropyLabel);
-                        if (m_Anisotropy.value.floatValue != 0.0f)
-                        {
-                            EditorGUILayout.Space();
-                            EditorGUILayout.HelpBox(
-                                "When the value is not 0, the anisotropy effect significantly increases the performance impact of volumetric fog.",
-                                MessageType.Info, wide: true);
-                        }
+                    PropertyField(m_DirectionalLightsOnly);
+                    if (PropertyField(m_Anisotropy, s_AnisotropyLabel) && m_Anisotropy.value.floatValue != 0.0f)
```
JulienIgnace-Unity, line 156: Same here

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -145,36 +147,19 @@ public bool showAdditionalProperties
             }
         }
 
-        /// <summary>
-        /// Start a scope for additional properties.
-        /// This will handle the highlight of the background when toggled on and off.
-        /// </summary>
-        /// <returns>True if the additional content should be drawn.</returns>
-        protected bool BeginAdditionalPropertiesScope()
```
JulienIgnace-Unity, line 153: Do we really want to remove that from public namespace?
Anytime a users uses anything else than PropertyFields, the additional properties implementation won't work.

com.unity.render-pipelines.high-definition/Editor/PostProcessing/DepthOfFieldEditor.cs
alex-vazquez: Will be shown with additional properties, what it won't have is the highlight

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
alex-vazquez: I can do the scope protected if you want.

com.unity.render-pipelines.high-definition/Editor/Sky/AtmosphericScattering/FogEditor.cs
alex-vazquez: Yeah, if the Anisotropy is not rendered it will return false and therefore the content won't be rendered

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
alex-vazquez: I think is not released, that's why I removed. But correct me if I am wrong

jenniferd-unity (approved):

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
JulienIgnace-Unity: I don't think it's released yet indeed. But my point is more that users can (and probably will) have UI code without PropertyFields (I know we had quite a few at least for a while) so if we don't expose this anymore, they won't be able to properly implement additional properties.
Or maybe I'm missing something here?

com.unity.render-pipelines.high-definition/Editor/PostProcessing/DepthOfFieldEditor.cs
JulienIgnace-Unity: Yeah but don't we want the background highlight to be consistent?

JulienIgnace-Unity: (I haven't seen how it looks like but I would assume that we'd rather have the whole background be highlighted)

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
alex-vazquez: I've recovered the functions.

JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Editor/PostProcessing/MotionBlurEditor.cs
```diff
@@ -55,38 +53,31 @@ public override void OnInspectorGUI()
 
             PropertyField(m_MaxVelocityInPixels);
             PropertyField(m_MinVelInPixels);
+            PropertyField(m_DepthCmpScale);
 
-            if (BeginAdditionalPropertiesScope())
-            {
-                PropertyField(m_DepthCmpScale);
+            EditorGUILayout.Space();
+            EditorGUILayout.LabelField("Camera Velocity", EditorStyles.miniLabel);
```
TomasKiniulis, line 59: This gets shown when Additional Properties are toggled off and shown twice When Additional Properties are toggled on.
https://user-images.githubusercontent.com/50582134/109020061-4ba88380-76c2-11eb-9c9f-039d55c395c6.mp4

TomasKiniulis (changes requested):
Properties appear correctly aligned comparing to master. 

Master result:
master (2)
PR:
PR_Change

Only one minor issue with LabelField "Camera Velocity" is left

com.unity.render-pipelines.high-definition/Editor/PostProcessing/MotionBlurEditor.cs
alex-vazquez: Fixed, could you check please? Thank you :)

TomasKiniulis (approved):
Looks great now! Thanks @alex-vazquez!

com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -9,10 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fixed missing warning UI about Projector component being unsupported (case 1300327).
 - Fixed the display name of a Volume Parameter when is defined the attribute InspectorName
-- Calculating correct rtHandleScale by considering the possible pixel rounding when DRS is on
```
martint-unity, line 12: Are you supposed to delete an entry from the changelog?

martint-unity (approved):
LGTM

PR #3548, opened 2021-02-18
[Test Projects] Remove Terrain Shader tests from UniversalGraphics project

he-ax:
### Purpose of this PR
Remove terrain shader tests from UniversalGrahicsTest project. They should only be in UniversalTGraphicsTest_Terrain project.


HedvigAxelsson:
Entire project will be removed, no need for this PR.

PR #3556, opened 2021-02-18
[bugfix1251348] Fix for Node Search Window not showing some compatible ports when dragging to open

tracychen1234:
### Checklist for PR maker

---
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1251348/

---
### Testing status
The reported issue is fixed--The Checkerboard node shows up when dragging out from a Vector4 Node input slot.
Succeed in local tests (PlayMode& EditMode).

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Graphs/MaterialSlot.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/MaterialSlot.cs
@@ -259,35 +259,11 @@ namespace UnityEditor.ShaderGraph
 
         public bool IsCompatibleStageWith(MaterialSlot otherSlot)
         {
-            var candidateStage = otherSlot.stageCapability;
-            var downStreamNodeStageCapability = ShaderStageCapability.All;
-            var connectedStage = new List<ShaderStageCapability>();
-            var downstreamNodeList = new List<AbstractMaterialNode>();
-            bool isBlockNodeConnected = false;
-            NodeUtils.GetDownsteamNodesForNode(downstreamNodeList, otherSlot.owner);
-            foreach (var n in downstreamNodeList)
-            {
-                if (n.GetType() == typeof(BlockNode))
-                {
-                    isBlockNodeConnected = true;
-                    var inputSlots = n.GetInputSlots<MaterialSlot>();
-                    connectedStage.Add(inputSlots.First().stageCapability);
-                }
-            }
-
-            if (isBlockNodeConnected)
-            {
-                if (!connectedStage.Contains(ShaderStageCapability.Fragment))
-                {
-                    downStreamNodeStageCapability = ShaderStageCapability.Vertex;
-                }
-                else if (!connectedStage.Contains(ShaderStageCapability.Vertex))
-                {
-                    downStreamNodeStageCapability = ShaderStageCapability.Fragment;
-                }
-            }
-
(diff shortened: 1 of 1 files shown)


LandonTownsendUnity (changes requested):
Works for fragment stage exclusive nodes like Checkerboard, but Vertex ID (a vertex stage exclusive node) currently does not show up unless the node you're dragging the edge out from connects to the vertex stage

com.unity.shadergraph/Editor/Data/Graphs/MaterialSlot.cs
```diff
@@ -259,8 +259,29 @@ public bool IsCompatibleWith(MaterialSlot otherSlot)
...
+                {
+                    var inputSlots = n.GetInputSlots<MaterialSlot>();
+                    connectedStage.Add(inputSlots.First().stageCapability);
+                }
+                else
+                {
+                    downStreamShaderSatge = otherSlot.stageCapability;
+                }
+            }
+            if (!connectedStage.Contains(ShaderStageCapability.Fragment))
```
Nightmask3, line 279: I'm not sure if this is true so it's a question, is it possible to get around having to store the shader stages for all downstream nodes and directly check within the loop itself as you go? 

What I'm thinking is that instead you could directly check in the loop if the BlockNode is of type Vertex or of type Fragment if its encountered, and if so then you could directly set the downStreamShaderStage and then break out of the loop.

I see that connectedStage is only ever added to when you encounter a BlockNode, and it seems to me that you could only ever have BlockNodes of either Fragment or Vertex stageCapability in the list of downstream nodes correct?

Because we already do prevent you from connecting to a Vertex stage block node from a Fragment stage node (and vice versa), so it seems like you could check if either of those are encountered and set the downStreamShaderStage directly to that value when first encountered.

Let me know if that makes sense, might be completely wrong

Nightmask3 (changes requested):

Nightmask3:
> Works for fragment stage exclusive nodes like Checkerboard, but Vertex ID (a vertex stage exclusive node) currently does not show up unless the node you're dragging the edge out from connects to the vertex stage

Good catch, I imagine this could be fixed by also checking if: `candidateStage == ShaderStageCapability.All`

com.unity.shadergraph/Editor/Data/Graphs/MaterialSlot.cs
tracychen1234: I find NodeUtils.GetEffectiveShaderStageCapability() seems are checking the effective shader stage downstream and upstream(if its a output slot). I tried and it seemed working! So I used it to check rather than what I wrote previously.

LandonTownsendUnity (approved):
Tested for correct behavior for Vertex exclusive nodes, Fragment exclusive nodes, Vertex exclusive subgraphs, Fragment exclusive subgraphs, and properties in the following cases:
1. Coming from an add node not plugged into anything
2. Coming from an add node plugged into a vertex phase block
3. Coming from an add node plugged into a fragment phase block
4. Coming from a vertex phase block
5. Coming from a fragment phase block

Nightmask3 (approved):
Looks good! Nice find of that util function for checking up and downstream, didn't know that was a thing! 😄

PR #3557, opened 2021-02-18
Unify XR code into SRP Core [22.1]

fabien-unity:
---
### Purpose of this PR
- converge XR implementations from URP and HDRP into one code base in SRP Core
- expose new set of API (into the experimental namespace) to be used by Scriptable Render Pipelines

---
### Testing status
Tested locally with the template projects on various configurations :
- Win10+Oculus+d3d11 in Editor and Standalone
- Android+Quest

---
### Comments to reviewers
This is an important refactor, please take time to review carefully and pay extra attention to the exposed public API and naming convention.


sebastienlagarde (approved):
approve for HDRP part

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,374 @@
...
+        // Keep track of only one XR layout
+        static XRLayout s_Layout = new XRLayout();
+
+    #if ENABLE_VR && ENABLE_XR_MODULE
+        static List<XRDisplaySubsystem> s_DisplayList = new List<XRDisplaySubsystem>();
+        static XRDisplaySubsystem s_Display;
+    #endif
+
+        // MSAA level (number of samples per pixel) shared by all XR displays
+        static int s_MsaaLevel = 1;
```
sandy-carter-unity, line 27: Can we use the `MSAASamples` enum for this?

fabien-unity: Good point, done!

com.unity.render-pipelines.core/Runtime/XR/XRPass.cs
```diff
@@ -0,0 +1,262 @@
...
+        {
+            return m_Views[viewIndex].textureArraySlice;
+        }
+
+        /// <summary>
+        /// Queue up render commands to enable single-pass techniques.
+        /// Note: depending on the platform and settings, either single-pass instancing or the multiview extension will be used.
+        /// </summary>
+        /// <param name="cmd"></param>
+        public void StartSinglePass(CommandBuffer cmd)
```
jenniferd-unity, line 140: Any reason why we prefer to check for enabled keyword instead of #if ENABLE_VR && ENABLE_XR_MODULE?

com.unity.render-pipelines.core/Runtime/XR/XROcclusionMesh.cs
```diff
@@ -0,0 +1,173 @@
...
+        static readonly ProfilingSampler k_OcclusionMeshProfilingSampler = new ProfilingSampler("XR Occlusion Mesh");
+
+        internal XROcclusionMesh(XRPass xrPass)
+        {
+            m_Pass = xrPass;
+        }
+
+        internal void SetMaterial(Material mat)
+        {
+            m_Material = mat;
```
jenniferd-unity, line 21: I wonder if this created a copy of the material instead of holding a reference. shouldnt we use ref keyword?

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,372 @@
...
+
+namespace UnityEngine.Rendering
+{
+    /// <summary>
+    /// Used by render pipelines to communicate with XR SDK.
+    /// </summary>
+    public static class XRSystem
+    {
+        // Valid empty pass when a camera is not using XR
+        public static readonly XRPass emptyPass = new XRPass();
```
jenniferd-unity, line 16: if this is public static, shouldnt we use k_emptyPass? I am not sure about code convention here.
also we should probably document it with <summary>

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,372 @@
...
+                s_MirrorViewMaterial = CoreUtils.CreateEngineMaterial(mirrorViewPS);
+
+            if (XRGraphicsAutomatedTests.enabled)
+                SetLayoutOverride(XRGraphicsAutomatedTests.OverrideLayout);
+        }
+
+        /// <summary>
+        /// Used by the render pipeline to communicate to the XR device how many samples are used by MSAA.
+        /// </summary>
+        /// <param name="msaaSampleCount"></param>
```
jenniferd-unity, line 80: wrong parameter name :)

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,372 @@
...
+        /// Returns the number of samples (MSAA) currently configured on the XR device.
+        /// </summary>
+        /// <returns></returns>
+        public static MSAASamples GetDisplayMSAASamples()
+        {
+            return s_MSAASamples;
+        }
+
+        /// <summary>
+        /// Used by the render pipeline to scale the render target on the XR device.
```
jenniferd-unity, line 106: Maybe doc could mention how to use it. I assume 1.0f corresponds to 1:1 / 100% scale but clarification could help our users

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,372 @@
...
+        /// <summary>
+        /// Used by the render pipeline to initiate a new rendering frame through a XR layout.
+        /// </summary>
+        public static XRLayout NewLayout()
+        {
+            RefreshDeviceInfo();
+
+            if (s_Layout.GetActivePasses().Count > 0)
+            {
+                Debug.LogWarning("Render Pipeline error : XRSystem.EndLayout() was not called!");
```
jenniferd-unity, line 128: The warning is confusing. Could we be explicit that you are calling the function right after?

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,372 @@
...
+            {
+                s_Display.zNear = zNear;
+                s_Display.zFar = zFar;
+            }
+        #endif
+        }
+
+        internal static void SetDisplaySync()
+        {
+            // Disable vsync on the main display when rendering to a XR device.
```
jenniferd-unity, line 181: Is this documented anywhere?

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,372 @@
...
+            // XRTODO : expose public property to control this behavior
+            QualitySettings.vSyncCount = 0;
+
+            // On Android and iOS, vSyncCount is ignored and all frame rate control is done using Application.targetFrameRate.
```
jenniferd-unity, line 185: the comment doesnt match the code below. Should you also test against iOS platform?

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        // XRTODO : expose as public API
+        static void SetLayoutOverride(Action<XRLayout, Camera> action)
+        {
+            s_LayoutOverride = action;
+        }
+
+        // Disable legacy VR system before rendering first frame
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSplashScreen)]
+        static void XRSystemInit()
```
jenniferd-unity, line 204: I am curious - does this work?
I thought that first frame was always builtin since the SRP are not yet initialized. but since we are checking the serialized value it may work?

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,372 @@
...
+        static XRView BuildView(XRDisplaySubsystem.XRRenderPass renderPass, XRDisplaySubsystem.XRRenderParameter renderParameter)
+        {
+            // Convert viewport from normalized to screen space
+            Rect viewport = renderParameter.viewport;
+            viewport.x      *= renderPass.renderTargetDesc.width;
+            viewport.width  *= renderPass.renderTargetDesc.width;
+            viewport.y      *= renderPass.renderTargetDesc.height;
+            viewport.height *= renderPass.renderTargetDesc.height;
+
+            // XRTODO : remove this line and use XRSettings.useOcclusionMesh instead when it's fixed
```
jenniferd-unity, line 339: do you have a fogbugz for this?

com.unity.render-pipelines.core/Runtime/XR/XRView.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    internal readonly struct XRView
+    {
+        internal readonly Matrix4x4 projMatrix;
+        internal readonly Matrix4x4 viewMatrix;
+        internal readonly Rect viewport;
+        internal readonly Mesh occlusionMesh;
+        internal readonly int textureArraySlice;
+
+        internal XRView(Matrix4x4 projMatrix, Matrix4x4 viewMatrix, Rect viewport, Mesh occlusionMesh, int textureArraySlice)
```
jenniferd-unity, line 11: should we use ref for occlusionMesh?

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
```diff
@@ -228,8 +228,6 @@ public class DebugData
             public float fullscreenDebugMip = 0.0f;
             /// <summary>Index of the light used for contact shadows display.</summary>
             public int fullScreenContactShadowLightIndex = 0;
-            /// <summary>XR single pass test mode.</summary>
-            public bool xrSinglePassTestMode = false;
```
jenniferd-unity, line 232: shouldnt we first deprecate it and redirerect to the new path?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1091,6 +1090,9 @@ protected override void Render(ScriptableRenderContext renderContext, Camera[] c
 #endif
             Terrain.GetActiveTerrains(m_ActiveTerrains);
 
+            XRSystem.singlePassAllowed = m_Asset.currentPlatformRenderPipelineSettings.xrSettings.singlePass;
```
jenniferd-unity, line 1093: cant we have the singlePassAllowed directly check for that value?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/XRMirrorView.hlsl
```diff
@@ -1,11 +1,11 @@
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
 
-TEXTURE2D_X(_BlitTexture);
+TEXTURE2D_X(_SourceTex);
```
jenniferd-unity, line 4: i wonder if it has an impact for our users that we are changing the name of those variables. what do you think?

jenniferd-unity (approved):
Pre-approving - but had a few comments about some copies + documentation

com.unity.render-pipelines.core/Runtime/XR/XRPass.cs
fabien-unity: It's 2 different mechanisms:
- the runtime check + settings is done with xr.enabled
- the conditional compilation is just a way to strip out the XR code on platform who do not support XR, but even if the platform supports XR, it does not imply XR is used or active for this particular frame

com.unity.render-pipelines.core/Runtime/XR/XROcclusionMesh.cs
fabien-unity: Classes in C# (like Material) are reference type by default so no need for ref keyword here.

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: Will fix 👍

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: Will change 👍

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: will change 👍

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: I think for public static it's not required, but I'll add the documentation.

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: Not yet, I'll add a todo !

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: will fix

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: Yes this has been working like this for 2 years now. I don't think we have a better way but I'm open to ideas.

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: no, I'll see if we can add one

com.unity.render-pipelines.core/Runtime/XR/XRView.cs
fabien-unity: Mesh is class so the ref it not required here, and we don't modify it

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
fabien-unity: I'll re-add it with the Obsolete keyword. There's no alternative for now, but this was only really used internally and not required anymore.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
fabien-unity: I think we could once we migrate the setting to a common Core setting ? right now, `GlobalXRSettings` is from HDRP only but `XRSystem.singlePassAllowed` is in Core.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/XRMirrorView.hlsl
fabien-unity: This was really used only internally so there should be 0 impacts for users.

thomas-zeng (approved):

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,377 @@
...
+#if ENABLE_VR && ENABLE_XR_MODULE
+using UnityEngine.XR;
+#endif
+
+namespace UnityEngine.Rendering
+{
+    /// <summary>
+    /// Used by render pipelines to communicate with XR SDK.
+    /// </summary>
+    public static class XRSystem
```
mikesfbay, line 13: Rather than removing XRSystem from URP and forcing both URP and HDRP to use the same XRSystem class, we should allow XR to have different features that are on URP vs HDRP. HDRP and URP have different needs, with HDRP more console and PC-centric while URP is more mobile-centric.

Rather than a single XRSystem class, we can create XR base class in Core that has basic common features. URP can subclass from XR base class to allow for customizations and experimental features geared toward Quest/mobile.

Specifically for XRSystem:

1. Create a base class XRBaseSystem in core and let XRSystem in each pipeline subclass from XRBaseSystem.

2. This allows experimental features such as late latching to first populate and be developed in URP (adding to URP's XRSystem). For late latching, HDRP won't have any immediate need for late latching given PC targets at 90Hz with a much faster GPU. Oculus also has no plans to support late latching for Rift.

3. There are other differences between HDRP (console/PC) and URP (mobile) such as the need for memoryless textures and Vulkan subpasses on mobile, that allowing subclassing of XRBaseSystem will allow more flexible render pipeline features.

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,377 @@
...
+        /// <param name="renderScale">A value of 1.0f represents 100% of the original resolution.</param>
+        public static void SetRenderScale(float renderScale)
+        {
+        #if ENABLE_VR && ENABLE_XR_MODULE
+            SubsystemManager.GetInstances(s_DisplayList);
+
+            foreach (var display in s_DisplayList)
+                display.scaleOfAllRenderTargets = renderScale;
+        #endif
+        }
```
mikesfbay, line 119: Scaling the render target for dynamic resolution should be avoided since XR must reallocate the eye texture whenever the scale changes. This has caused many developer performance issues before.
URP/Quest uses viewport scale and fixed foveated resolution for dynamic resolution. 
This refactor is a good time to drop SetRenderScale from URP and make it specific to HDRP if it still needs it.

RenderScale would be one more reason to create subclass from a base XRBaseSystem class to be used by each pipeline.

com.unity.render-pipelines.core/Runtime/XR/XRBuiltinShaderConstants.cs
```diff
@@ -0,0 +1,72 @@
...
+        static readonly int k_StereoCameraProjection    = Shader.PropertyToID("unity_StereoCameraProjection");
+        static readonly int k_StereoCameraInvProjection = Shader.PropertyToID("unity_StereoCameraInvProjection");
+        static readonly int k_StereoMatrixV             = Shader.PropertyToID("unity_StereoMatrixV");
+        static readonly int k_StereoMatrixInvV          = Shader.PropertyToID("unity_StereoMatrixInvV");
+        static readonly int k_StereoMatrixP             = Shader.PropertyToID("unity_StereoMatrixP");
+        static readonly int k_StereoMatrixInvP          = Shader.PropertyToID("unity_StereoMatrixInvP");
+        static readonly int k_StereoMatrixVP            = Shader.PropertyToID("unity_StereoMatrixVP");
+        static readonly int k_StereoMatrixInvVP         = Shader.PropertyToID("unity_StereoMatrixInvVP");
+        static readonly int k_StereoWorldSpaceCameraPos = Shader.PropertyToID("unity_StereoWorldSpaceCameraPos");
+
```
mikesfbay, line 18: Since built-in shaders will be gradually removed from URP (ie. Skybox), this built-in shader constants file should remain in URP instead of promoting to Core. Moving to Core would only confuse developers to rely more on built-in constants instead of URP ones.  HDRP also don't use built-in so there's no need to share this file with HDRP.

mikesfbay (changes requested):

com.unity.render-pipelines.core/Runtime/XR/XRBuiltinShaderConstants.cs
fabien-unity: The built-in constants are required on both pipelines for backwards compatibility with older shaders. I don't think we can get rid of them until we get rid of the builti-in pipeline.

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: This API method is currently called from each pipeline but we can remove it from URP if required. It's an optional call. I believe it's good to expose it publicly for the custom pipelines also. If URP no longer requires it, then we just don't call the method.

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
fabien-unity: I think it's worth considering this design as a second step, if we encounter a good use case.
But for now, with this class being static, it's easy to handle the differences directly in the respective render pipeline.
Adding inheritance right now will complexify the API and will make cross-pipeline code harder to deal with.

sandy-carter-unity (approved):

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
mikesfbay: Here's a good use case. The new late latching for URP adds several functions to XRSystem and adds late latching states to XRPass. These shouldn't go into core pipeline since late latching isn't supported on HDRP nor needed in HDRP as I mentioned above. 
If each pipeline can subclass from a common XRBaseSystem, new XR features like late latching can exist in one pipeline without polluting other pipelines.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -157,14 +148,14 @@ public UniversalRenderPipeline(UniversalRenderPipelineAsset asset)
...
             }
 
-#if ENABLE_VR && ENABLE_XR_MODULE
-            XRSystem.UpdateRenderScale(asset.renderScale);
-#endif
+
+            // Configure initial XR settings
+            MSAASamples msaaSamples = (MSAASamples)Mathf.Clamp(Mathf.NextPowerOfTwo(QualitySettings.antiAliasing), (int)MSAASamples.None, (int)MSAASamples.MSAA8x);
+            XRSystem.SetDisplayMSAASamples(msaaSamples);
+            XRSystem.SetRenderScale(asset.renderScale);
```
phi-lira, line 157: XRSystem doesn't depend on the modules anymore I guess? That's nice.

com.unity.render-pipelines.universal/Runtime/XR.meta
```diff
@@ -1,8 +0,0 @@
-fileFormatVersion: 2
```
phi-lira, line 1: Was this a dangling meta, I can't find the script file?

com.unity.render-pipelines.universal/Runtime/XR/XRLayout.cs
```diff
@@ -1,26 +0,0 @@
-// Helper API to create custom XR layout
```
phi-lira, line 1: Are you missing deleting the .meta for this script?

phi-lira (approved):
LGTM. Thanks for also removing the obsolete properties / functions.

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
mikesfbay: Any thoughts on my above comments? By merging in late latching and moving XRSystem to Core, late latching is exposed to HDRP and any user defined SRP's, which we don't support or tested with late latching. 

There's also late latching code changed during the merge, which needs to be dealt with (I'll do a comparison).

mikesfbay (changes requested):

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        /// <summary>
+        /// Used by the render pipeline to begin late lacthing mechanism.
+        /// </summary>
+        public static void BeginLateLatching(Camera camera, XRPass xrPass)
+        {
+#if ENABLE_VR && ENABLE_XR_MODULE
+            // Only support late latching for multiview on Android
+            if (s_Display != null && Application.platform == RuntimePlatform.Android && xrPass.viewCount == 2)
```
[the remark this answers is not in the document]
fabien-unity: Ok I'll update the line.

com.unity.render-pipelines.universal/Runtime/XR/XRLayout.cs
fabien-unity: The whole folder is gone, there should be no left-over meta files now.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
fabien-unity: XR code stills needs the modules to be present but I moved the conditional compilation inside the core package to have a cleaner URP/HDRP integration.
If the modules are not present, the methods will be stripped out by the IL2CPP dead code removal.

com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs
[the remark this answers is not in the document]
mikesfbay: Can you also update the comment line above without the "Android" requirement?
"// Only support late latching for multiview on Android"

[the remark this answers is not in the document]
fabien-unity: Oh right! done.

theopnv (approved):
The gfx-sdets team is only owning the codeowners file in this PR. Please refer to the other reviewers for the other files. Thanks!

DiannaUnity (approved):
Version: 2022.1.0a8.1153
Revision: trunk a6d7f27874c0
Built: Sat, 04 Sep 2021 19:12:24 GMT

Tested with Android Pixel 5 and iOS iPhone 12 Pro

I used the ARFoundation samples project and there was no crashing, I checked with a new project and it appears not to affect Mobile negatively, LGTM 👍

JasonCostanza:
@fabien-unity Will this be backported at all? Reason I ask, this is going to master and thusly paired with 21.2+ of the Unity editor. If it will backport to 21.1 then I will need to test both WMR Provider and the OpenXR provider. If it's just staying in 21.2 then I will only be testing OpenXR provider as the WMR provider is no longer available. Can you clarify the plan there?

fabien-unity:
@JasonCostanza this PR is targeting trunk (2022.x) + SRP 13 but *might* be backported to 2021.2 / SRP 12. I don't think we'll backport to 2021.1.

ryanhy-unity (approved):
All gfx xr automation tests passed for URP:
   gfx tests:
      hdrp: rift (instanced)
      urp:  Rift (instanced), Quest (multiview), OpenXR (on WMR headset, multipass and instanced)
   functional: Rift, Quest, OpenXR, MockHMD, WMR, all stereo rendering modes
   performance: this was blocked in the xr automation currently.  Please advise if we need to test this and I can probably get a workaround (issue is in the setup scripts and not related to changes here)

manual testing:
  urp: tested Quest with all rendering modes using end to end xr automated test, no issues
   hdrp: tested Rift with all rendering modes using hdrp template scene, no issues

JasonCostanza (changes requested):
Version: 2021.2.0b12.3482
Revision: 2021.2/staging 8137c0cefaf0
Built: Wed, 08 Sep 2021 00:45:04 GMT

commit 5a609f4c2da103dbbf24909a1149372d747af465 (HEAD -> xr/unify-into-core-1, origin/xr/unify-into-core-1)

I tested on WMR and seeing unique failures to render things properly compared to HL2. The devices are not appearing to display the same results using the identical UWP app solution

WMR Failures occur in the following scenes of the Functional Test Suite
- Line Renderer
- Post Render
- Projectors
- Render Texture
- Replacement Shader
- Target Buffers
- Stencil Buffer

On HL2 these scenes display as I would expect to see them. For these reasons, I am not confident I can pass the PR as WMR HMD's would not display equal results to the HL2 HMD

The above listed tests can be found here:
https://github.cds.internal.unity3d.com/unity/xr.manualtests.functional

fabien-unity:
@JasonCostanza Can you confirm the same tests are passing with Graphics master version ? I don't have this HW to test on.

fabien-unity:
I did a last (hopefully) update so that we can extend XRPass for URP-specific use cases (late latching, compositor layers, etc).
@ryanhy-unity would you be able to re-run the test suite with this latest changes ?

simon-engelbrecht-soerensen (approved):
Tested with mac/win in all API's in standalone builds/editor. Working as expected.

com.unity.render-pipelines.universal/Runtime/XR/XRSystemUniversal.cs
```diff
@@ -0,0 +1,69 @@
...
+                cmd.UnmarkLateLatchMatrix(CameraLateLatchMatrixType.InverseViewProjection);
+                xrPass.hasMarkedLateLatch = false;
+            }
+#endif
+        }
+
+        internal static void MarkShaderProperties(CommandBuffer cmd, XRPassUniversal xrPass, bool renderIntoTexture)
+        {
+#if ENABLE_VR && ENABLE_XR_MODULE
+            if (xrPass.isLateLatchEnabled && xrPass.canMarkLateLatch)
```
mikesfbay, line 53: Previously, MarkLateLatchShaderProperties is called in UpdateGPUViewAndProjectionMatrices() under the condition that cameraData.xr.singlePassEnabled is true. Currently, that's no longer the case. We need to restore it back as before:

    internal void UpdateGPUViewAndProjectionMatrices()
        {
     ...
            if (cameraData.xr.singlePassEnabled)
            {
               ...
                if (cameraData.xr.canMarkLateLatch)
                    MarkLateLatchShaderProperties(cmd, ref cameraData);
            }
        }

Also, there's the plan to change xrPass.isLateLatchEnabled based on return value of BeginRecordingIfLateLatched(). In that case, xrPass.isLateLatchEnabled would strictly be based on XRDisplay's capabilities rather than being multiview or not (currently). Thus, it's better to also add the additional cameraData.xr.singlePassEnabled check here.

Fyi, thanks for adding this XRSystemUniversal container!

mikesfbay (changes requested):

com.unity.render-pipelines.universal/Runtime/XR/XRSystemUniversal.cs
fabien-unity: Actually, it is redundant.
isLateLatchEnabled can only be set to true if singlePassEnabled is true (see BeginLateLatching).
So whenever isLateLatchEnabled is true, we can be sure that singlePassEnabled is also true.

mikesfbay: You didn't finish reading my last comment: 
"Also, there's the plan to change xrPass.isLateLatchEnabled based on return value of BeginRecordingIfLateLatched(). In that case, xrPass.isLateLatchEnabled would strictly be based on XRDisplay's capabilities rather than being multiview or not (currently). Thus, it's better to also add the additional cameraData.xr.singlePassEnabled check here."

Thus, we should keep my previous logic the same during this refactor so that engine's BeginRecordingIfLateLatch changes would just work without further changes to URP.

mikesfbay: Upon more discussion, we can leave this line as is, and later we can add vendor-specific policy decisions to xrUniversal.canMarkLateLatch (i.e. whether a vendor (Oculus, MS) allows late latching in singlePass or multipass or both). With new BeginRecordingIfLateLatched changes, isLateLatchEnabled will still be based on the vendor capability flag.

mikesfbay:
@amaestas13 Since late latching code paths have changed in this refactor, please re-test with the late latch Stage 3 URP test project and LL SRP Batcher (Stage 4) test project (Stage 4 PR has landed in latest trunk). Let me know if you have any questions.

sebastienlagarde:
Hey, any update on this PR?

daves-unity (approved):

amaestas13 (approved):
Ran testing on late latching and tested oculus locally.
Tested with both 1.11 and 2.0 and everything looks to be good. 

With the previous testing done by @ryanhy-unity in automation i feel I can sign off on this. 

Sorry for the delay

JasonCostanza (approved):
2022.1.0a15.2209
Revision: trunk 49120632d05b
Built: Tue, 09 Nov 2021 06:50:39 GMT

commit 549af26548c1138f2bf23cd07d61f87424a72136 (HEAD -> xr/unify-into-core-1, origin/xr/unify-into-core-1)

Compared HL2 and WMR using the same project and no longer see any differences between the devices like I did before.

mikesfbay (approved):

phi-lira (approved):

phi-lira:
@fabien-unity please run URP PR before merging this one. You might have already did that but it's hard to see as there are a few merges of master to it.

2kCarlos:
@fabien-unity  (CC: @alxdncn )
I was wondering if this new XRLayout API is used in rendering yet, or just the C# side so far?
Is there a way to get a RenderTexture (or multiple) of all the view(s) rendered?

For example, if I setup a XRLayout with 1 XRPass and 8 XRViews, can I actually...
1. Get 1 RenderTexture with all 8 views' render results in it, and/or,
2. Get 8 RenderTextures with the render results in them? (Perhaps in a texture array?)

If so, a code example would be amazing! :)

fabien-unity:
@2kCarlos ,
So far, it's only partially used internally (see https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.core/Runtime/XR/XRSystem.cs#L94).

The XRLayout API is not ready and is only internal for the moment and we don't have a timeline to make it usable for end users.
You could try to hack something my changing most of the internal keywords to public but we don't support this use case for now.
The idea would be to clear the default layout, and then add the new XRPass and XRView as you wish.
But after a deeper look, you'll face more issues and I'm not sure it's good avenue for your project.
I think instead you might want to look at implementing a XR SDK Plugin to achieve what you need.

cecarlsen:
Hi @fabien-unity I found your comment from on Feb 23, 2022 In my research into setting custom view projection matrices for instanced XR. I'd like to have a look at the XR SDK. But the link to the sign up page in [the manual](https://docs.unity3d.com/6000.2/Documentation/Manual/xr-sdk.html) is broken. How can I access the XR SDK? Ideally I'd like to see how MockHMD is implemented.

fabien-unity:
cc @thep3000 could you follow up about the broken link ?

cecarlsen:
@fabien-unity and @thep3000  So. Now the link in the Unity 6 manual is gone. I guess that also sort of counts as a fix. But how do I access the XR SDK and example code? We desperately need this as it seems to be the _only_ way to set custom view matrices.

fabien-unity:
@thomas-zeng can you take in this feedback ? Any anything changed since this PR landed ?

thomas-zeng:
Hi @cecarlsen ,

Have you looked into a couple of helper functions in XR that can update custom view projection data for shaders?
`UpdateBuiltinShaderConstants`
`SetBuiltinShaderConstants`

And here is how URP uses these functions to setup view/proj matrix
https://github.com/Unity-Technologies/Graphics/blob/d6894e53d4ee279e61ff3b856a719a799ac0586b/Packages/com.unity.render-pipelines.universal/Runtime/FrameData/UniversalCameraData.cs#L52

cecarlsen:
Thanks @thomas-zeng I hadn't seen the XRBuiltinShaderConstants. In my case, I'm trying to make this work in HDRP. Calling UpdateBuiltinShaderConstants() in a BeforeRendering custom pass does nothing - perhaps its overwritten by HDCamera when it uploads constants in ExecuteRenderRequest() (not sure which of the two calls is first). It this even a viable approach. Or would I have to modify HDRP, or write a XR Provider plugin?

Regarding the XR SDK, the [documentation ](https://docs.unity3d.com/6000.0/Documentation/Manual/xrsdk-provider-setup.html)says: _"The SDK package includes an example project that builds sample plug-ins."_. How do I access this SDK and samples code? The com.unity.xr.management package has a "Example XR Management implementation" sample, but it only shows the basics on the C# side.

cecarlsen:
Thanks [@thomas-zeng](https://github.com/thomas-zeng) I hadn't seen the XRBuiltinShaderConstants. In my case, I'm trying to modify the single pass stereo view projection matrices in HDRP, and I can see XRBuiltinShaderConstants is also used there. But mimicking HDCamera.PushBuiltinShaderConstantsXR in a BeforeRendering custom pass does nothing. Is the HDCamera updating the constants after BeforeRendering? Is this even a viable approach? Or would I have to modify HDRP, or write a XR Provider plugin?

Regarding the XR SDK, the [documentation ](https://docs.unity3d.com/6000.0/Documentation/Manual/xrsdk-provider-setup.html)says: "The SDK package includes an example project that builds sample plug-ins.". How do I access this example project? The com.unity.xr.management package has a "Example XR Management implementation" sample, but that can hardly be it as it only shows the basics on the C# side.

cecarlsen:
Hi @thomas-zeng I found a [hack](https://discussions.unity.com/t/custom-single-pass-stereo-matrices-in-hdrp-how/1624042/6) – but it involves modding both HDRP and SRP Core which is far from optimal. I've received some hints on how to implement an XR Provider, but it seems overly complicated for just setting a couple of matrices. I hope functional version of SetStereoProjectionMatrix() and SetStereoViewMatrix() will appear in the future. Preferably with the ability to change the XR layout and define more than two views.

PR #3448, opened 2021-02-10
Universal/split project 20.2 autosplit

hdb-unity:
### Purpose of this PR

This PR copies scenes from UniveralGraphicsTest **into** UniversalGraphicsTest_2D, UniversalGraphicsTest_Foundation, UniversalGraphicsTest_Lighting, UniversalGraphicsTest_Terrain, UniversalGraphicsTest_PostPro.

The aims are to enable better collaboration by **separating test projects by team**, or at least by concern. Also, to **reduce iteration time** and **CI resource load** by allowing developers to more easily run tests relating to the areas of their code changes; instead of running the entire suite each run locally or in CI. The entire suite can and should then be run after iteration to give the full coverage.

This PR lands split projects into master branch.

---
### Testing status

**Manual Tests**: What did you do?
- [x] Run split projects in CI
- [x] Run original project in CI
- [x] Run split projects locally on MacOS
- [x] Tested backport process from these split projects to universal/split-project-20.1 and universal/split-project-19.3
- [x] Reverted changes commit to UniversalGraphicsTest project (it should NOT be changed by this PR) 

---
### Comments to reviewers
- CI jobs should continue to use UniversalGraphicsTest after this PR
- This PR should not change any code outside of TestProjects
- This PR should not change any projects in TestProjects, except UniversalGraphicsTest_2D, UniversalGraphicsTest_Foundation, UniversalGraphicsTest_Lighting, UniversalGraphicsTest_Terrain, UniversalGraphicsTest_PostPro
- There are tasks to complete after this PR (as discussed with reviewers)

PR #3558, opened 2021-02-19
Copy test scene 146_AdditionalLightShadowCustomResolution to test project UniversalGraphicsTest_Lighting

Wilfrid-Unity:
### Purpose of this PR

https://github.com/Unity-Technologies/Graphics/pull/2126 added automated test scene https://github.com/Unity-Technologies/Graphics/blob/master/TestProjects/UniversalGraphicsTest/Assets/Scenes/146-AdditionalLightShadowCustomResolution.unity

https://github.com/Unity-Technologies/Graphics/pull/3448 copied project UniversalGraphicsTest scenes to 5 smaller test projects. However test scene 146-AdditionalLightShadowCustomResolution.unity was not included during this copy.
I think this might be caused by the hyphen in the scene name, that might have confused the tooling.

This pull request copies test scene 146-AdditionalLightShadowCustomResolution.unity from test project UniversalGraphicsTest to test project UniversalGraphicsTest_Lighting (and renames it to use an underscore instead of the hyphen).

### DIFF
--- a/TestProjects/UniversalGraphicsTest_Lighting/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/146_AdditionalLightShadowCustomResolution.png
+++ b/TestProjects/UniversalGraphicsTest_Lighting/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/146_AdditionalLightShadowCustomResolution.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:ede8664bf70f457df19f659a83b05272dd643caadb9870d1bdddb1b476161fbd
+size 84889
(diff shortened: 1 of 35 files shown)


phi-lira:
Is that test scene testing real time lighting only? If so it should be in _Foundation instead.
@hdb-unity maybe we should rename _Lighting to GI / BakedLighting?

Wilfrid-Unity:
> Is that test scene testing real time lighting only? If so it should be in _Foundation instead.
> @hdb-unity maybe we should rename _Lighting to GI / BakedLighting?

Yes it only tests real-time lighting (so does 145_ManyAdditionalLightShadowTypes that is also in split-project "Lighting" currently).

hdb-unity (approved):
These changes look good, new scene passes locally in playmode.

I'll make note to rename this scene in the backport of split project, too.

PR #3564, opened 2021-02-19
Removing duplicated property in lit shader

lukaschod:
### Purpose of this PR
Removing duplicated property _GlossMapScale in lit shader, it was there accidently.
https://fogbugz.unity3d.com/f/cases/1315032/

---
### Testing status
Extended test ValidateShaderResources to check also if there is duplicated properties in our urp shaders.
In future warning will be added in shader system.


Verasl (changes requested):
This property is not actually used, my guess is a bad merge at some point maybe as to why it still exists in the properties(the non-hide in inspector) not to mention this also seems to be in the ComplexLit shader too, also un-utilised.
I would suggest we look at removing the non-hidden ones, and keep the one listed in the 'ObsoleteProperties'.

lukaschod:
> This property is not actually used, my guess is a bad merge at some point maybe as to why it still exists in the properties(the non-hide in inspector) not to mention this also seems to be in the ComplexLit shader too, also un-utilised.
> I would suggest we look at removing the non-hidden ones, and keep the one listed in the 'ObsoleteProperties'.

Thats a good catch, fixed

Verasl (approved):
LGTM

PR #3565, opened 2021-02-19
Enable 119 test on Android Vulkan

tomzig16:
### Purpose of this PR
This PR enables 119 test on Android vulkan as it was previously disabled due to instability.
I have reenabled it and ran couple of times the tests - no failures. I suspect maybe it was happening due to big load, which was lowered when projects were split. In any case, keeping test filtered out and a bug with huge user pain reported makes little sense to me. With PR I will be closing 1252270 ticket and opening a Jira ticket in Mobile Graphics team to log this incident. 

---
### Testing status
[Yamato branch](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/urp%252Fandroid%252Fenable-119)
Couple of tests on Yamato to try to reproduce, no failures so far.


phi-lira (approved):

tomzig16:
Removing needs backport because this one does not reproduce on newer samsung devices. Once we add them to 10.x this test will be enabled back with them, as I go through all tests (filtered and unfiltered)

PR #3562, opened 2021-02-19
[HDRP] Change "All" to "Scene" in Force Forward Emissive menu label + udpate documentation

sebastienlagarde:
### Purpose of this PR
Change "All" to "Scene" in Force Forward Emissive menu label + udpate documentation

---
### Testing status
Describe what manual/automated tests were performed for this PR

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #3566, opened 2021-02-19
Added bold formatting for menu items and added the all > scene option renames

JordanL8:
### Purpose of this PR
Adds work done in #3562 to the what's new page and added bold formatting.


sebastienlagarde (commented):
@JordanL8 there is no need to mention the change in Force Emissive Forward as it wasn't existing before. it is a new feature for 21.2

sebastienlagarde (approved):

JordanL8:
Ahh right, thanks!

PR #3570, opened 2021-02-19
[2022.1] Universal: Fixing 146 MSAA test scene.

ellioman:
# Purpose of this PR
Previously when I modified this test scene in #2802, I broke the backbuffer part of the test.
This PR fixes that to make sure we are testing BackBuffer MSAA properly.


manuele-bonanno (approved):
Looks good! Thanks for fixing it!

PR #3572, opened 2021-02-19
[XPipeline]Camera editor unification on HDRP

alex-vazquez:
### Purpose of this PR
https://jira.unity3d.com/browse/XPIPELINE-84

[UX Document](https://docs.google.com/document/d/1XlfH365Ak4xGECRSS4QYShWHfHrd6N5qfb7mrar-7Tc/edit)

Unify headers and controls across SRP's, this PR is for HDRP Camera Editor

image

---
### Testing status
Check that values are updated and restored.

---
### Comments to reviewers
The editor of the clear camera color was always visible because the condition was commented.


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Drawers.cs
```diff
@@ -533,8 +536,8 @@ static void Drawer_FieldDepth(SerializedHDCamera p, Editor owner)
         static void Drawer_FieldClear(SerializedHDCamera p, Editor owner)
         {
             EditorGUILayout.PropertyField(p.clearColorMode, clearModeContent);
-            // if(p.clearColorMode.GetEnumValue<HDAdditionalCameraData.ClearColorMode>() == HDAdditionalCameraData.ClearColorMode.BackgroundColor) or no sky in scene
-            EditorGUILayout.PropertyField(p.backgroundColorHDR, backgroundColorContent);
+            if(p.clearColorMode.GetEnumValue<HDAdditionalCameraData.ClearColorMode>() == HDAdditionalCameraData.ClearColorMode.Color)
```
jenniferd-unity, line 539: why did we change from background color to Color here?
also should it appear/disappear or be grayed out?

alex-vazquez: - In built-in and URP it disapears. Should we change the behaviour?
- `Color` is the new name, background does not exist anymore, I think we changed somehow and we commented it until they did the replacement. But we forgot to remove the comment.

RSlysz: It was replaced long time ago because we don't only clear the background. BackgroundColor make sens in a painter rendering algorithm but we don't really do that so it can be misleading.

Could you check with UX for that. Before it was more : we should hide less possible and prefer disabled fields. Then we start to hide in more options as there was too many of them. I am unsur if it is good or not to hide this. We don't hide fields based on other fields most of the time.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Skin.cs
```diff
@@ -6,13 +6,15 @@ namespace UnityEditor.Rendering.HighDefinition
     static partial class HDCameraUI
     {
         const string generalSettingsHeaderContent = "General";
-        const string physicalSettingsHeaderContent = "Physical";
+        const string renderingSettingsHeaderContent = "Rendering";
+        const string physicalSettingsHeaderContent = "Physical Camera";
+        const string environmentSettingsHeaderContent = "Environment";
         const string outputSettingsHeaderContent = "Output";
 
         const string clippingPlaneMultiFieldTitle = "Clipping Planes";
```
RSlysz, line 14: Sorry for that. Seams that we miss L10n here. Can you convert this or add a FB for those fields please?

RSlysz (approved):

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Drawers.cs
TheoWong-pixel: It's okay to hide the field when the input is not relevant to the current selection. 

We should only be showing + greying out fields when selection is not available but the information on what is being active / used is still necessary.

TheoWong-pixel: But can we have Background Color indented under Background Type?

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Skin.cs
alex-vazquez: added localization :)

jenniferd-unity (approved):

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Drawers.cs
alex-vazquez: @TheoWong-pixel , I've updated the picture with the new editor :)

RSlysz (approved):
lgtm

alex-vazquez:
@Unity-Technologies/gfx-qa-hdrp , Could you do a quick pass on the properties and see that they appear correctly? Many thanks

iM0ve (approved):
Looks good

Whats tested:
* I compared the Before and After camera properties
* I used some of the more advanced properties to see if they still work
* Upgraded a simple scene

**Before**

Before

**After**

After

**Testing if Phys Depth of Field, Exposure and TAA still work**


https://user-images.githubusercontent.com/36502659/109512362-b257e380-7aac-11eb-8d21-58dec974f378.mp4

PR #3578, opened 2021-02-19
Improved a bit the area cookie behavior for higher smoothness values to reduce artifacts.

anisunity:
image

When area light cookies are not relatively diffuse texture (white dot in the middle for instance), artifacts can start to appear (check screenshot). This PR reduces those artifacts a bit for high smoothness values, but they remain for rougher surfaces.

Testing
Tested with textures shared by users there is a significant improvement.
Test needs to be rerun and there is probably screenshots


remi-chapelain (changes requested):
The PR improves slightly the quality. It's not much, but it's honest work :). 
Couldn't find any case where it was worse. 
The lower the intensity of the area light and the higher the resolution of the cookie, the more you'll be able to see improvements. 

I also tested with IES cookie ✔️ 

However, I'm blocking because there's one issue with SG Stacklit that seems to ignore the smoothness of the material now. (stacklit is the middle one on the top row)

26bba5870013c6329dafac2eb709cf3e

**Low Intensity**
9da288e1cdbaaf7aa93e479e2b0052a1

**High Intensity**
9da288e1cdbaaf7aa93e479e2b0052a1

**Low Resolution**
05bd82e02c5c1e110221bab0634c6f5a

remi-chapelain (approved):
Looks good to me now that the stack lit have been fixed ! ✔️

sebastienlagarde (approved):

sebastienlagarde:
Don't merge for now as it cause issue in test

anisunity:
There was an issue with the default behavior that i fixed

anisunity:
Updated Dx11, Dx12 and WinVulkan screenshots. Linux didn't launch. Relaunching the tests

PR #3579, opened 2021-02-19
Dynamic Resolution Tests

kecho:
### Purpose of this PR
* Adding new dynamic resolution tests, for software and hardware on all platforms.
* DRS can now run on render to texture cameras.
* Adding new callback for CustomPostProcess pass. This callback is required since we change the filter mid frame. Its the only possible way I found to set the filter on a per camera basis.

image

### Comments to reviewers
This PR includes a new callback in the CustomPass. If we dont wan't to expose this, please provide feedback on other possible ways we could change the filter mid frame.
Another possibility as well is to create 8 separate tests, and keep the filter on a per frame basis (not per camera basis). This however would add 8 new tests.

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/4091_DRS-Software.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/4091_DRS-Software.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:7aea5daf529c8577a2b8a590348e6d4c2c06c230715382a69af71cc748f9d169
+size 933103
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/4091_DRS-Software.png.meta
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/4091_DRS-Software.png.meta
@@ -1,0 +1,108 @@
+fileFormatVersion: 2
+guid: 94e3b861d03b9a94b9da5408d42879f9
+TextureImporter:
+  internalIDToNameTable: []
+  externalObjects: {}
+  serializedVersion: 11
+  mipmaps:
+    mipMapMode: 0
+    enableMipMap: 0
+    sRGBTexture: 1
+    linearTexture: 0
+    fadeOut: 0
+    borderMipMap: 0
+    mipMapsPreserveCoverage: 0
+    alphaTestReferenceValue: 0.5
+    mipMapFadeDistanceStart: 1
+    mipMapFadeDistanceEnd: 3
+  bumpmap:
+    convertToNormalMap: 0
+    externalNormalMap: 0
+    heightScale: 0.25
+    normalMapFilter: 0
+  isReadable: 1
+  streamingMipmaps: 0
+  streamingMipmapsPriority: 0
+  vTOnly: 0
+  grayScaleToAlpha: 0
+  generateCubemap: 6
+  cubemapConvolution: 0
+  seamlessCubemap: 0
+  textureFormat: 1
+  maxTextureSize: 2048
+  textureSettings:
+    serializedVersion: 2
+    filterMode: 1
+    aniso: 1
+    mipBias: 0
+    wrapU: 0
+    wrapV: 0
+    wrapW: 0
+  nPOTScale: 0
+  lightmap: 0
+  compressionQuality: 50
+  spriteMode: 0
+  spriteExtrude: 1
+  spriteMeshType: 1
+  alignment: 0
+  spritePivot: {x: 0.5, y: 0.5}
+  spritePixelsToUnits: 100
+  spriteBorder: {x: 0, y: 0, z: 0, w: 0}
+  spriteGenerateFallbackPhysicsShape: 1
+  alphaUsage: 1
+  alphaIsTransparency: 0
+  spriteTessellationDetail: -1
+  textureType: 0
+  textureShape: 1
+  singleChannelComponent: 0
+  flipbookRows: 1
+  flipbookColumns: 1
+  maxTextureSizeSet: 0
+  compressionQualitySet: 0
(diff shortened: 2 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/CustomPass/CustomPassBookmarkEventContext.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace UnityEngine.Rendering.HighDefinition
```
kecho, line 3: Will add proper docs here, if you guys agree on this callback arguements / naming

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/CustomPass/CustomPass.cs
```diff
@@ -213,7 +213,7 @@ internal void ExecuteInternal(RenderGraph renderGraph, HDCamera hdCamera, Cullin
             this.owner = owner;
             this.currentRenderTarget = targets;
             this.currentHDCamera = hdCamera;
-
+            OnRenderEvent(new CustomPassRenderEventContext() { hdCamera = hdCamera, eventType = CustomPassRenderEventContext.EventType.OnExecute });
```
sebastienlagarde, line 216: Adding @alelievr  for review on adding callback in custom pass

sebastienlagarde (approved):
I am ok with the change on my side, but prefer to let @alelievr  judge about the added callback or if we want to do it another way

sebastienlagarde:
Note: you should trigger Yamato

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/CustomPass/CustomPass.cs
alelievr: @kecho I'm not sure to see why we need to add another user callback for such case, there is already the `Execute()` function of the custom pass that should be enough.

Additionally, your callback is outside of the render graph, so I'm really not sure when it would be executed.

Also, I don't understand the purpose of the enum `CustomPassRenderEventContext.EventType`. We already have the injection point available in the custom pass API: https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.3/api/UnityEngine.Rendering.HighDefinition.CustomPass.html#UnityEngine_Rendering_HighDefinition_CustomPass_injectionPoint

kecho: Hey @alelievr , the problem with Execute is that it happens too late. It happens in the execution of the render graph itself. I need a callback that happens during render graph construction, I thought having some sort of 'event' hook would make it easy to add other injection points. 
Alternatively, I need a callback that ticks once per camera, before performing any rendering.  This tick must happen during render graph construction, prior to running the render graph.
I need to specify a new upsampling filter per camera, and this must happen before render graph execution, otherwise its too late.

alelievr: Okay, so indeed, in that case, we don't have yet an API that provides access during the construction of the render graph. We still have the plan to expose such API when we'll move the custom passes to render graph (with full API on the user side).

So right now, I'd say that it's better to not expose the AAPI you added because we won't be able to remove it before the replacement lands (if it's public, it needs to be documented and kept alive for migration purposes during multiple major versions :/ )

Just to make sure, right now this callback is only used during our tests? If that's the case, we can either make it internal and then replace it with a proper API when it lands. Or you can use the This callback:
https://docs.unity3d.com/ScriptReference/Rendering.RenderPipelineManager-beginCameraRendering.html

It's called at the very beginning of the frame before the culling, would this work for you?

sebastienlagarde: >I need a callback that ticks once per camera, before performing any rendering / render graphing.
We have BeginCameraRendering (https://docs.unity3d.com/2021.1/Documentation/ScriptReference/Rendering.RenderPipeline.BeginCameraRendering.html) but not sure if it fit all your need.

kecho: @sebastienlagarde  this delegate will work! However I will have to make  some refactors on how the state of the DynamicResolutionManager gets propagated. Will have to run it by @FrancescoC-unity  but sounds like the way to go then.

kecho: Did the changes, worked like a charm. @alelievr  won't need the custom pass callback. Thanks for checking in. Will wait on a review from Francesco now :)

FrancescoC-unity: Did you push said changes? I see only formatting as recent push

kecho: woops! sorry just pushed again

FrancescoC-unity: Oh the change looks trivial to me, I am good with it 🟢

FrancescoC-unity (approved):

alelievr (approved):

sebastienlagarde (approved):

sebastienlagarde:
Need to update screenshots now :)

kecho:
Final screenshots have been updated. Triggering yamato one more time.

sebastienlagarde:
Test is green, I will move forward and merge it as it can prevent to break code. If QA have feedback, please raise them, thanks.

JMargevics (commented):
Checked already merged. Looks good 👍

PR #3583, opened 2021-02-22
[XPipeline]Remove dot color parameter from MaterialHeaderScope and Layered Materials Improvements.

alex-vazquez:
---
### Purpose of this PR
- Remove the ColorDot parameter of `MaterialHeaderScope` this allow us avoiding generating `GUIContent` every time for adding the spacing.
- Instead we are generating the `GUIContents` once with an icon from a Texture2D with the color, and every time that we are drawing the scope we are setting the proper icon size.
- Fix vertical spacing on LayeredLit

##### Before PR
image

##### After PR
image

---
### Testing status
Check that a material using `HDRP/LayeredLit`shader is properly shown

---
### Comments to reviewers
The `Texture2D.red` has alpha, that's why I created another red in our `CoreEditorStyles`.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
+++ b/com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
@@ -20,11 +20,11 @@ namespace UnityEditor.Rendering
         /// <summary>Style for a minilabel button</summary>
         public static readonly GUIStyle miniLabelButton;
 
-        /// <summary><see cref="Texture2D"/> 1x1 pixels with red color</summary>
+        /// <summary><see cref="Texture2D"/> 1x1 pixel with red color</summary>
         public static readonly Texture2D redTexture;
-        /// <summary><see cref="Texture2D"/> 1x1 pixels with green color</summary>
+        /// <summary><see cref="Texture2D"/> 1x1 pixel with green color</summary>
         public static readonly Texture2D greenTexture;
-        /// <summary><see cref="Texture2D"/> 1x1 pixels with blue color</summary>
+        /// <summary><see cref="Texture2D"/> 1x1 pixel with blue color</summary>
         public static readonly Texture2D blueTexture;
 
         /// <summary> PaneOption icon </summary>


com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -71,7 +71,37 @@ static CoreEditorUtils()
...
+        {
+            Texture2D tex2 = new Texture2D(1, 1);
+            tex2.SetPixels32(new Color32[]{color});
+            tex2.Apply();
+            return tex2;
+        }
+
+        /// <summary>Changes the icon size of the <see cref="EditorGUIUtility"/>, and restore it back</summary>
+        public class IconSizeScope : IDisposable
+        {
```
jenniferd-unity, line 89: you should also add documentation or make them internal

jenniferd-unity (changes requested):
Need to add documentation or change to internal the new additions to CoreEditorUtils.cs

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
alex-vazquez: I saw that `EditorGUIUtility` already has a IconSizeScope :)

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/DetailInputsUIBlock.cs
```diff
@@ -76,13 +75,12 @@ internal class Styles
         /// <param name="layerIndex">Current layer index to display. 0 if it's not a layered shader</param>
         /// <param name="features">Features of the block.</param>
         /// <param name="dotColor">Subheader dot color. See Layered Lit UI subheader for more info.</param>
-        public DetailInputsUIBlock(ExpandableBit expandableBit, int layerCount = 1, int layerIndex = 0, Features features = Features.All, Color dotColor = default(Color))
+        public DetailInputsUIBlock(ExpandableBit expandableBit, int layerCount = 1, int layerIndex = 0, Features features = Features.All)
```
jenniferd-unity, line 78: @RSlysz I am not sure here if it is okay to remove the parameter from the constructor, what do you think?

jenniferd-unity (commented):
Ready to approve if it is okay to remove a parameter from the constructor

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/DetailInputsUIBlock.cs
RSlysz: I am not sure. The code have changed a lot after @alelievr change it to be blocks.
I would say that it was a default value that was used formerly to not draw the dot. But if alex identified the fields where it is used, I'm pretty sure it is ok to remove it now.

com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
```diff
@@ -88,6 +95,10 @@ static CoreEditorStyles()
             contextMenuIcon = new GUIContent(paneOptionsIcon, contextTooltip);
             contextMenuStyle = new GUIStyle("IconButton");
 
+            redTexture = CoreEditorUtils.CreateColoredTexture2D(Color.red);
+            greenTexture = CoreEditorUtils.CreateColoredTexture2D(Color.green);
+            blueTexture = CoreEditorUtils.CreateColoredTexture2D(Color.blue);
```
RSlysz, line 100: Not sure it will be used outside of HDRP though

RSlysz (approved):
It is a little less performant but at scale that have no impact and in Editor context. It integrate better this way. Nice work.

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayersUIBlock.cs
```diff
@@ -17,15 +17,17 @@ public class LayersUIBlock : MaterialUIBlock
...
-                "Layer 2",
-                "Layer 3",
+                EditorGUIUtility.TrTextContent(" Main layer", icon: Texture2D.whiteTexture),
+                EditorGUIUtility.TrTextContent(" Layer 1", icon: CoreEditorStyles.redTexture),
+                EditorGUIUtility.TrTextContent(" Layer 2", icon: CoreEditorStyles.greenTexture),
+                EditorGUIUtility.TrTextContent(" Layer 3", icon: CoreEditorStyles.blueTexture),
             };
 
-            public static readonly ExpandableBit[] layerExpandableBits =
+            public static ExpandableBit[] layerExpandableBits { get; } =
```
alelievr, line 30: I'm interested to know why did you change this read-only field to a property?

Properties are more expensive to access than fields because of the underlying function they generate, furthermore, the data inside the layerExpandableBits can be modified at runtime with a getter like this (which doesn't seem to be useful in this case).

Is there something I'm missing here?

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/DetailInputsUIBlock.cs
alelievr: Yeah, I guess it's okay since they aren't used anymore with this PR.
But be careful as it's a public API breaking change it can't be backported

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayersUIBlock.cs
```diff
@@ -17,15 +17,17 @@ public class LayersUIBlock : MaterialUIBlock
...
+            public static Vector2 layerIconSize = new Vector2(5, 5);
+
+            public static GUIContent[] layers { get; } =
             {
-                "Main layer",
-                "Layer 1",
-                "Layer 2",
-                "Layer 3",
+                EditorGUIUtility.TrTextContent(" Main layer", icon: Texture2D.whiteTexture),
+                EditorGUIUtility.TrTextContent(" Layer 1", icon: CoreEditorStyles.redTexture),
```
alelievr, line 25: There is a Texture2D.redTexture in Unity already so you don't need to create a new one:
https://docs.unity3d.com/ScriptReference/Texture2D-redTexture.html

com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
alelievr, line 100: IIRC, Texture2D assets are not garbage collected, they need to be explicitly destroyed otherwise they'll leak in the editor.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -72,9 +72,21 @@ static CoreEditorUtils()
         }
 
         // UI Helpers
+
+        /// <summary>Creates a 1x1 <see cref="Texture2D"/> with a plain <see cref="Color"/></summary>
+        /// <param name="color">The color to fill the texture</param>
+        /// <returns>a <see cref="Texture2D"/></returns>
+        public static Texture2D CreateColoredTexture2D(Color color)
+        {
+            Texture2D tex2 = new Texture2D(1, 1);
```
alelievr, line 81: I think it would be better to mark this texture as HideAndDontSave so its content survives a domain reload which is useful if you keep a reference of it in a window that can also survive a domain reload:
https://docs.unity3d.com/ScriptReference/HideFlags.HideAndDontSave.html

alelievr (changes requested):
Left a few comments (the leak should be addressed)

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayersUIBlock.cs
alex-vazquez: `Texture2D.redTexture` has alpha 0.

```
inline void Red(Texture2D *tex, unsigned char *data, int x, int y, int maxX, int maxY)
{
    data[0] = 255;
    data[1] = 0;
    data[2] = 0;
    data[3] = 0;
}
```

I need to create my own red without alpha.

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayersUIBlock.cs
alex-vazquez: I only changed because I didn't like that the field is named layerExpandableBits instead of k_layerExpandable bits. therefore I changed to a property to have a correct naming convention.

com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
alex-vazquez: I will leave them here for the future.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
alex-vazquez: Done thanks,

com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
alex-vazquez: Added `DestroyInmediate` of this textures.

com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
```diff
@@ -20,11 +20,17 @@ public static class CoreEditorStyles
         /// <summary>Style for a minilabel button</summary>
         public static readonly GUIStyle miniLabelButton;
 
-        static readonly Texture2D paneOptionsIconDark;
-        static readonly Texture2D paneOptionsIconLight;
+        /// <summary><see cref="Texture2D"/> 1x1 pixels with red color</summary>
```
jenniferd-unity, line 23: pixel is singular here :)

alex-vazquez: ups, yeah, thanks :)

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/DetailInputsUIBlock.cs
jenniferd-unity, line 78: since it is public we should probably leave the previous constructor as is (with Color dotcolor), mark it as obsolete and call the new constructor without the Color dotColor, what do you think?

alex-vazquez: Perfect, I will do that.

alex-vazquez: @jenniferd-unity , I've tried to do it obsolete but as this is an argument with default parameter I am getting.

```
![image](https://user-images.githubusercontent.com/76204843/109274320-a7365680-7813-11eb-8cab-0964169d7a4b.png)
```
What do you think is the best here to do?, leave the argument and do not use it, or just remove it?

jenniferd-unity (approved):

TomasKiniulis (approved):
Looks good! Haven't ran into any issues tweaking LayeredLit Surface Inputs, Layer parameters. Material is properly updated.

PR #3585, opened 2021-02-22
Trigger scene repaints twice on automatic exposure settings changed

FrancescoC-unity:
Usually apparently a scene repaint is called upon changes of the volume settings. However, automatic exposure modes will need two frames to correctly display the result (the automatic exposure is based on previous frame); with this change we force the repaint twice if a setting that requires it happens. 

Fix for: https://fogbugz.unity3d.com/f/cases/1315508/

**What did I test**: the repro case and trivial variants of it.


sebastienlagarde (approved):

FrancescoC-unity:
I messed up with merges. Closing and creating a fresh PR.

PR #3589, opened 2021-02-22
[XPipeline]UniversalRenderPipelineSerializedCamera class

alex-vazquez:
### Purpose of this PR
Extract all serialized parameters of a camera
This will help us to unify the code between URP and HDRP (WIP for the camera editor)

---
### Testing status
Check that the camera values are properly updated and restored

---
### Comments to reviewers
Just wrapping all the serialized properties in a class
Renaming the camera editor


jenniferd-unity (approved):
Code looks good - but i would add Jira link and additional description before submitting to explain the PR's motivation

com.unity.render-pipelines.universal/Editor/Camera/URPSerializedCamera.cs
```diff
@@ -0,0 +1,62 @@
+using UnityEngine.Rendering.Universal;
+
+namespace UnityEditor.Rendering.Universal
+{
+    internal class URPSerializedCamera
+    {
+        internal SerializedObject serializedObject;
```
pbbastian, line 7: All the `internal` members (properties, constructors, fields included) should be `public` when the type is `internal`. Also, this one should probably also be a get-only property. If not that, then a readonly field, as changing the value would leave it inconsistent with all the other values.

com.unity.render-pipelines.universal/Editor/Camera/URPSerializedCamera.cs
```diff
@@ -0,0 +1,62 @@
...
+        internal void Update()
+        {
+            serializedObject.Update();
+        }
+
+        internal void Apply()
+        {
+            serializedObject.ApplyModifiedProperties();
+
+        }
```
pbbastian, line 60: I think these should not be here, as they are basically just duplicating the `SerializedObject` API. Instead, the API user should call them directly on the `serializedObject`.

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added View Vector node to mimic old behavior of View Direction node in URP.
 - Added support for the PlayStation 5 platform.
 - Enabled deferred renderer in UI.
+- Added a wrapper class for all serialized properties of a Camera (URPSerializedCamera)
```
pbbastian, line 12: No need for this line, as it is an internal change

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -21,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Move Assets/Create/Rendering/Universal Render Pipeline/Forward Renderer to Assets/Create/Rendering/URP Forward Renderer
 - Removing unused temporary depth buffers for Depth of Field and Panini Projection.
 - Optimized the Bokeh Depth of Field shader on mobile by using half precision floats.
+- Changed UniversalRenderPipelineCameraEditor to URPCameraEditor
```
pbbastian, line 25: No need for this line, as it is an internal change

pbbastian (changes requested):

com.unity.render-pipelines.universal/Editor/Camera/URPSerializedCamera.cs
```diff
@@ -0,0 +1,50 @@
+using UnityEngine.Rendering.Universal;
+
+namespace UnityEditor.Rendering.Universal
+{
+    internal class URPSerializedCamera
```
RSlysz, line 5: In HDRP we already have it separated. It is called SerializedHDCamera. We should align to have same naming everywhere I think if the intend is to align separation.

Also, default visibility for a class is already internal. For nested class it would have been private but here the visibility does no change at all.

If the aim is to use the separated class for embedding all the serialization burden, it would make sens to have an Update() and Apply() method that redirect on the serializedObject. This way, the serializedObject can be hidden and all serialization specific code managed here. Especially, we deal with two set of data: the ones in the camera and the ones in the additionaldatas. You can sync both almost all the time so it give credit to have it here in its own Update/Apply.

The naming right now is a little confusing:
- m_SerializedCamera (name of the variable in code) is in fact the serialization on the additional data component
- settings handle datas of the legacy camera


Note that we do this in HDRP because:
- we have a lot of fields
- we  have really specific serialization process sometimes that need dedicated code.
- it allows to make a clear delimitation between data / drawer / logic. We made this by constructing a static library drawer HDCameraUI and thus the editor logic remain in the Editor class. But it complexity things so as long as class are not that long, I advice against doing this.

I am not 100% sure URP will profit from this change as it is right now.

com.unity.render-pipelines.universal/Editor/Camera/URPSerializedCamera.cs
```diff
@@ -0,0 +1,50 @@
...
+            volumeTrigger = serializedObject.FindProperty("m_VolumeTrigger");
+            renderPostProcessing = serializedObject.FindProperty("m_RenderPostProcessing");
+            antialiasing = serializedObject.FindProperty("m_Antialiasing");
+            antialiasingQuality = serializedObject.FindProperty("m_AntialiasingQuality");
+            stopNaN = serializedObject.FindProperty("m_StopNaN");
+            dithering = serializedObject.FindProperty("m_Dithering");
+            clearDepth = serializedObject.FindProperty("m_ClearDepth");
+            cameraType = serializedObject.FindProperty("m_CameraType");
+            var o = new PropertyFetcher<UniversalAdditionalCameraData>(serializedObject);
+            cameras = o.Find("m_Cameras");
```
RSlysz, line 44: I bielieve you dont need a property fetcher for only one field...
You could either use it for all field or rewrite this one this way:
cameras = serializedObject.FindProperty("m_Cameras");

RSlysz (commented):
Technically it is ok beside confusing namings. Not sure of the interest here though.

What do you mean by unifying camera editors? Let's hopes it will not add to much burden on the refactor I am doing on cameras :/

com.unity.render-pipelines.universal/Editor/Camera/URPSerializedCamera.cs
alex-vazquez: Changed thanks @RSlysz

com.unity.render-pipelines.universal/Editor/Camera/URPSerializedCamera.cs
alex-vazquez: The idea here is preparing URP for the future as well, this will help us identify the common code, and tooltips, etc.

Thank you for your feedback.

com.unity.render-pipelines.universal/Editor/Camera/URPCameraEditor.cs
```diff
@@ -13,7 +13,7 @@ namespace UnityEditor.Rendering.Universal
 {
     [CustomEditorForRenderPipeline(typeof(Camera), typeof(UniversalRenderPipelineAsset))]
     [CanEditMultipleObjects]
-    class UniversalRenderPipelineCameraEditor : CameraEditor
+    class URPCameraEditor : CameraEditor
```
martint-unity, line 16: DO we want to change this name? Everything else we have in our code is Universal....

martint-unity (approved):
Approving, But I would like an answer on my comment :)

com.unity.render-pipelines.universal/Editor/Camera/URPSerializedCamera.cs
```diff
@@ -0,0 +1,55 @@
+using UnityEngine.Rendering.Universal;
+
+namespace UnityEditor.Rendering.Universal
+{
+    internal class URPSerializedCamera
```
martint-unity, line 5: UniversalRenderPipelineSerializedCamera?

alex-vazquez: Yeah, I will change that too, many thanks for the heads up :)

simon-engelbrecht-soerensen (approved):
Looks to be working as expected!

PR #3591, opened 2021-02-22
Fix NaNs issues with Volumetric Clouds on some platforms

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1316066/ 

The way the reprojection happens samples in an [-1...1] range, however that could lead to sample the LDS out of range. On some platforms this will result in unexpected results leading to either NaNs or very large random values causing corruption. 

 This PR adds a clamp to the address so that sampling outside LDS is not possible. 

**Note**: This is not against hd/bugfix as the volumetric clouds are not in that branch yet. 
**Note2:** I see that the feature is disabled on Xbox, I am not certain if this issue was the only reason why so I did not re-enable it. 

**What did I test**: launched on Xbox several times and noticed it does not repro anymore. (Before it either gives colourful corruption or black screen). 

@ QA: Can you please check again on Metal? (you'll need to revert a couple of lines of code, see the latest commit on this PR and remove the Metal part too)


TomasKiniulis (approved):
For Xbox side I think validation by dev is sufficient. Approving, but let me know in case it does need an extra Xbox player build.

FrancescoC-unity:
> For Xbox side I think validation by dev is sufficient. Approving, but let me know in case it does need an extra Xbox player build.

Could you guys check on metal? I wonder if this was the same issue.

remi-chapelain:
> > For Xbox side I think validation by dev is sufficient. Approving, but let me know in case it does need an extra Xbox player build.
> 
> Could you guys check on metal? I wonder if this was the same issue.

@iM0ve will take a look !

iM0ve (changes requested):
So after modifying the code to enable metal the Volumetric clouds works (we still need a commit to push these changes, because I did it locally).

A few probably PR unrelated observations:
* 2 metal warnings regarding clouds
* The transition between clouds is very "ghosty"


https://user-images.githubusercontent.com/36502659/108730157-ec1c6d80-7533-11eb-814d-2f7041e2055c.mov

FrancescoC-unity:
> So after modifying the code to enable metal the Volumetric clouds works (we still need a commit to push these changes, because I did it locally).
> 
> A few probably PR unrelated observations:
> 
> * 2 metal warnings regarding clouds
> * The transition between clouds is very "ghosty"
> 
>  Screen.Recording.2021-02-22.at.17.30.03.mov

This is unrelated, please file a separate bug.

FrancescoC-unity:
I am going to re-enable Metal too soon

iM0ve (approved):
Updated my review

anisunity (approved):
Good catch, lgtm

sebastienlagarde (approved):

PR #3593, opened 2021-02-22
Fix alpha output in AOVs and debug views when using shadow matte

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1311830/

Shadow matte was not outputting the correct alpha in debug view and in AOVs.

---
### Testing status
I have used this test scene, where the ground is using a shadow matte shader:
image

And after this PR, we get the following alpha output when selecting "Alpha" in the "Common material properties" debug view  :
image

---
### Comments to reviewers
There is a performance hit now when outputting alpha with a debug view, because we also render the shadow maps (even when a shadow matte is not in the view). Perhaps this could be optimized in the future with the RendererList API.


skhiat (approved):
Look good to me

sebastienlagarde (approved):

remi-chapelain (approved):
Testing seems sufficient on this one ! ✔️

PR #3594, opened 2021-02-22
[HDRP][PathTracing] Changed maximum intensity parameter from clamped to positive value

eturquin:
This addresses: https://fogbugz.unity3d.com/f/cases/1310514/

This PR changes the type of Path Tracer's maximum intensity parameter from a clamped value (0 to 100), to a positive value.
This allows to set it to arbitrarily large values, effectively deactivating the clamp, and solving the kind of issues exposed in the ticket above.
In addition, a more detailed tooltip has been added.

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/2x_Lighting/2011_ForwardEmissiveForDeferred/UnlitSGEmissive.shadergraph
+++ b/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/2x_Lighting/2011_ForwardEmissiveForDeferred/UnlitSGEmissive.shadergraph
@@ -1,5 +1,5 @@
 {
-    "m_SGVersion": 2,
+    "m_SGVersion": 3,
     "m_Type": "UnityEditor.ShaderGraph.GraphData",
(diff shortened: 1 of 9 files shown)


iM0ve (approved):
**Conclusion**
I think this change allows more flexibility for pathtracing, allowing it to work better in physically-based scenarios. With a small downside that the tool will require bit more knowledge to use, this is alleviated with the new tooltip.

Whats tested:
* The Intensity parameter is now a possitive float instead of a slider
* Inputting values <0 is not allowed
* Tooltip
* The issue described in fogbuz is fixed. Users can now decide to have a noisier image but prevent it from looking underexposed.

**Here are my comparisons:**
Path tracing off

cVxlD1uUf2

Maximum Intensity 100 000

ol9qdS57aU

Maximum Intensity 1000

qOZZLV0b4v

Maximum Intensity 100 (Max before PR)

9it4SQCLtO

Maximum Intensity 10 (default value)

YuOkUTlmFJ

sebastienlagarde (approved):

sebastienlagarde:
Note: retarget this PR to amster when hdbugfix is merge as it must not be backported to 10.x (due to breaking change api)

PR #3596, opened 2021-02-22
[VFX] Fix VFX graph shader code generation error with empty textures in shader graph

gabrieldelacruz:
---
### Purpose of this PR
There was a regression recently, when generating vfx graph shader code from shader graph.
If there is a texture node that is empty, we are not able to properly declare the texture in the shader code. To solve this we had to pass the TextureDimension along with the texture (even though it is redundant when we have a one) to create the correct VFXExpression.

---
### Testing status
The error it is very easy to repro, we just had to create an empty sample texture node with no texture. Tested the fix with all different types of TextureDimension available (Texture2D, Texture2DArray, Texture3D, Cubemap).

---
### Comments to reviewers
Texture2D has different kinds of default textures, but in VFX graph it will always default to gray. This was like this before, so it is not a regression. There is already a fogbugz for that (https://fogbugz.unity3d.com/f/cases/1114588/)


julienf-unity (approved):

julienf-unity:
For reviewers, this is quite an urgent fix as it is a HUP regression on 10.x. We'll make a 10.3.2 package with this fix (and one other) asap

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
```diff
@@ -341,24 +341,25 @@ public override VFXExpressionMapper GetExpressionMapper(VFXDeviceTarget target)
...
-                        foreach (var tex in shaderGraph.textureInfos.Where(t => t.texture != null).OrderBy(t => t.name))
+                        foreach (var tex in shaderGraph.textureInfos)
                         {
-                            switch (tex.texture.dimension)
+                            switch (tex.dimension)
                             {
+                                default:
                                 case TextureDimension.Tex2D:
-                                    mapper.AddExpression(new VFXTexture2DValue(tex.texture.GetInstanceID(), VFXValue.Mode.Variable), tex.name, -1);
+                                    mapper.AddExpression(new VFXTexture2DValue(tex.instanceID, VFXValue.Mode.Variable), tex.name, -1);
```
julienf-unity, line 350: Not needed in this PR but I wonder why Mode is set to variable. This is only textures withing shader graph that are not exposed right? So having them as constant will allow potential optimization opportunities on our side (like not resetting them always for instance)

gabrieldelacruz: Just checked the history and it's been like that since the creation of this file. Maybe there is a reason for it?

VladNeykov:
Getting `Shader error in 'Hidden/VFX/SG/System/Output Particle Quad': cannot implicitly convert from 'Texture2D<float4>' to 'struct UnityTexture2D' at line 1391 (on d3d11)` if the SG texture is un-exposed (does not happen in 8.x). Is this something which can be addressed in this PR?
8J3HGtH0Z1

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
```diff
@@ -348,12 +348,12 @@ static ShaderGraphVfxAsset GenerateVfxShaderGraphAsset(GraphData graph)
             // we need to override graph.isSubgraph, so save old state to restore it
             // (this is not great, but whole VFX pipeline is rather hacky at the moment)
             // use try/finally to ensure it always gets restored
+            // UPDATE(2/24/2021) @gabriel.delacruz: Not required anymore, as we are not
+            // forcing graph.isSubgraph to be true. Kept in case we have to go back to that,
+            // but most likely this code will be soon replaced by direct generation from SG
             bool oldIsSubGraph = graph.isSubGraph;
```
julienf-unity, line 354: Caching this does not look needed anymore?

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
julienf-unity: Don't think there's a reason, but anyway we I'll keep that in mind and we can fix later. At the moment the only optimization it allows is to share slot in property sheet for identical textures. But for SG texture, n texture slots will still be used given how it is done anyway. So the optimization is far from massive.

julienf-unity: Oh actually maybe the reason was that! because we cannot share SG texture because even if constant they require direct namings and cannot be aliased.

julienf-unity (approved):

julienf-unity:
Changelog is missing

julienf-unity:
Also did you make sure all tests were green locally?

com.unity.shadergraph/Editor/Data/Graphs/AbstractShaderProperty.cs
```diff
@@ -100,6 +100,11 @@ internal virtual void AppendPropertyBlockStrings(ShaderStringBuilder builder)
 
         internal abstract void ForeachHLSLProperty(Action<HLSLProperty> action);
 
+        internal virtual string GetPropertyAsArgumentString(string precisionString, bool isVfxGraph)
```
cdxntchou, line 103: If this is only ever called from VFX path, rename it to GetPropertyAsArgumentStringForVFX() and drop the bool.

com.unity.shadergraph/Editor/Data/Graphs/VirtualTextureShaderProperty.cs
```diff
@@ -184,11 +184,13 @@ internal void AddTextureInfo(List<PropertyCollector.TextureInfo> infos)
             {
                 string layerRefName = value.layers[layer].layerRefName;
                 var layerTexture = value.layers[layer].layerTexture;
+                var texture = layerTexture != null ? layerTexture.texture : null;
```
cdxntchou, line 187: FYI if you didn't know about the fancy C# syntax that does the same thing:

var texture = layerTexture?.texture;

com.unity.shadergraph/Editor/Generation/Processors/Generator.cs
```diff
@@ -114,7 +114,7 @@ void BuildShader()
             {
                 m_GraphData.AddValidationError(m_OutputNode.objectId, ShaderKeyword.kVariantLimitWarning, Rendering.ShaderCompilerMessageSeverity.Error);
 
-                m_ConfiguredTextures = shaderProperties.GetConfiguredTexutres();
+                m_ConfiguredTextures = shaderProperties.GetConfiguredTextures();
```
cdxntchou, line 117: Nice.

cdxntchou (changes requested):
Looks good! I would just want to rename the function to make it clear it is a special case.

VladNeykov (changes requested):
Hi @gabrieldelacruz ,
On **HDRP** it looks good: verified both bug fixes, tried both cases with different texture types (Texture2D, Texture2DArray, Texture3D, Cubemap), ran the FTP project and a player build, did manual exploratory testing:
aCSrKmFQCK

On **URP** however both FB cases still repro and the FTP test scene fails (checked on PC and mac, @VitaSkruibyte checked on her machine as well):
image
image

com.unity.shadergraph/Editor/Data/Graphs/VirtualTextureShaderProperty.cs
gabrieldelacruz: Thanks! I was aware of that syntax, but I try to avoid it with Object derived types. Not the case here, I'll change it.

gabrieldelacruz:
> On **URP** however both FB cases still repro and the FTP test scene fails (checked on PC and mac, @VitaSkruibyte checked on her machine as well):
It seems that there was a missing include file for URP, it should be fine now.

com.unity.shadergraph/Editor/Data/Graphs/AbstractShaderProperty.cs
gabrieldelacruz: Renamed, thanks!

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
gabrieldelacruz: I removed the caching and the comment, it is more readable now

VladNeykov (approved):
No outstanding issues, URP (editor + standalone) works as well now.
Thanks for fixing these HUP bugs! \o/
fHocJuwpIk

cdxntchou (approved):

PR #3001, opened 2021-01-06
Hdrp/update decal handles pivot

RSlysz:
### Purpose of this PR
This PR add the missing customizable pivot to Decals. It is required for placement snapping and finely tuning rotation point.
This contains the following changes:
- Add Pivot in the inspector. Modifying it will decal the projection box from the transform point.
HpT1k81sHB
- Add pivot and UV edit mode. In this mode, you can modify the pivot without moving the box position. It will automatically adapt transform position for the box to remain still.
3mBL4zh7aC
uzLOuAwdmt
- Add customizable shortcut key for this new edit mode. 
Unity_IMnMN5rNlj
- Add customizable color for the Decal's Gizmo / Handle. The main point here is to be able to select different intensity which is really useful when dealing with dark environment/ really bright environment.
fv3p853aA9

This also fix:
- https://fogbugz.unity3d.com/f/cases/1299596/
- https://fogbugz.unity3d.com/f/cases/1282722/
---
### Testing status
Tested the handles of previous edit mode and new edit mode with 
- pivot not at (0,0,0)
- transform not at position (0,0,0)
- transform with a random rotation
- size different than (1,1) and projection depth different than 1
In all above cases, it seams to behave correctly.

Tested that changing the Transform's rotation make the projection volume rotate around the pivot as intended.

Tested color and face intensity change along the color set in the preference.

Tested Undo/Redo on edition by handle of the pivot point (with new edit mode)

---
### Comments to reviewers
ColorPref are internal and cannot be directly used. I kinda hacked it by using them with reflection and Activator. It is not a nice integration but it seams to work. Particularly because the color list seams to be updated each time the window is enabled. So AFAIK what I have done should work fine.

Also, the default color provided to the ColorPref is the same than now. So this should not change on users at update without them actively modifying it.

To do this I had to tweek a little the HierarchicalBox used also for DensityVolume and ReflectionProbe's gizmo/handle. The change is minor and should not affect the behaviour of those.

All modification are only on the Editor manipulation side. So there is no actual change of the projection volume computation code. Which means that there is no need to migrate any data, they should still work fine. Though a test on that could be nice to be 100% sure.

@SeanPuller for awareness

@QA After discussion with @MaddalenaUnity , the color is not under HDRP but under Scene

PR #3600, opened 2021-02-22
Hd/fix pivot size ratio percistency through 0

RSlysz:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1308338/
Also set default pivot to (0, 0, 0) (must be a miss in the refactor) as otherwise the vertex snapping cannot work.
Also discovered and fixed a recently introduced multi-edition issue when changing size in Inspector.

razpJYvW4p

---
### Testing status
Manually tested:
- ratio is kept when changing the size to 0 and go back to another value through Inspector
- saved ratio is reset when pivot is changed in Inspector
- saved ratio is reset when pivot is changed in the edit mode
- tested multi-edition: when multi editing several DecalProjector's Inspector's Size, they all rescale around their own pivot point.

Note: saved ratio is not kept when Editor is destructed (the fix is worth only for the Inspector lifetime but it cover all problematic workflow with the issue above)

---


RSlysz:
Realized there is something odd with multi edition and tweaking inspector size (both before and after PR). Enquiring more information from @SeanPuller on slack on the matter to know the best way to solve this (or not) and integrate it here.

sebastienlagarde (approved):

RSlysz:
Tomas found a flaw in my initialization logic in case of multi-edition. This need to be fixed.

fredericv-unity3d (approved):

TomasKiniulis (approved):
After the fixes PR seems good!
What I've checked:
- Not encountering any issues with projection update
- Confirmed that adjusting Projection Depth with different multiple selected object values does update input box with the same value when changed
- No GUI errors occur when adjusting pivot
- Multi-editing decals with mismatching Projection Depth do adjust values from their individual pivots now

PR unrelated issue found: https://fogbugz.unity3d.com/f/cases/1317105/

sebastienlagarde:
Note: fix for PR: https://github.com/Unity-Technologies/Graphics/pull/3001 (2021.1)

PR #3602, opened 2021-02-22
Fix Emissive Color Override Debug Display (Case 1313123)

johnpars:
### Purpose of this PR
Fix: [1313123](https://fogbugz.unity3d.com/f/cases/1313123/)

[This PR](https://github.com/Unity-Technologies/Graphics/pull/3430) introduced support for a special forward pass to write emissive in the case of deferred render path. It introduced a bug related to the debug views, specifically emission override in the lighting panel:

image

As you can see a override emissive is provided and some outputs are ignoring it when they should be responsive. There is actually two different reasons why this happened and the PR fixes both:
1) The Unlit ShaderGraph used by **Test 2011** actually was not linking to the **Emission** port but the **Base Color** port. Since this results in no emissive in the graph, the emissive defaults to 0 and thus the debug mode ignores it. Fixing this in the graph solves it:

image

2) The debug display keyword was skipped for the new force emissive pass. Making sure the keyword is properly added to the pass solves it:

image

---
### Testing status
- I toggled the new Force Forward Emissive option and confirmed that the debug display emissive override functions as expected for lit/unlit materials.
- Did some general testing with the emissive color override debug view to confirm the the issue is gone.

---
### Comments to reviewers
~


com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/SurfaceSubTarget.cs
```diff
@@ -159,7 +159,7 @@ protected override void CollectPassKeywords(ref PassDescriptor pass)
             pass.keywords.Add(CoreKeywordDescriptors.TransparentWritesMotionVector);
             pass.keywords.Add(CoreKeywordDescriptors.FogOnTransparent);
 
-            if (pass.IsLightingOrMaterial())
+            if (pass.IsLightingOrMaterial() || pass.IsForwardEmissiveForDeferred())
```
johnpars, line 162: Does it make sense to decouple the debug display check from `LightingOrMaterial`? Not sure if this new emissive pass qualifies as "lighting or material" (and also, adding it into the LightingOrMaterial check causes other issues).

sebastienlagarde: Here I will suggest to make thing clearer. 

Let's create a
pass.NeedDebugDisplay()    (or any better name :) 

with 
public static bool NeedDebugDisplay(this PassDescriptor pass) =>
 return pass.IsLightingOrMaterial() || pass.lightMode == HDShaderPassNames.s_ForwardEmissiveForDeferredStr

At least the purpose of the function is clear.

> (and also, adding it into the LightingOrMaterial check causes other issues).
yes, it shouldn't be added to IsLightingOrMaterial, it is different.

sebastienlagarde (commented):
See suggetsion. Thanks for this fix!

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/SurfaceSubTarget.cs
johnpars: Done!

sebastienlagarde (approved):

sebastienlagarde:
I will redirect this PR to master once hd/bugfix is merge as this issue only exist on master

sebastienlagarde (approved):

PR #3603, opened 2021-02-22
[HDRP] Fixed SSR with SSR_Transparent

skhiat:
Case:
https://fogbugz.unity3d.com/f/cases/1311088/
@remi-chapelain 

Fix version of the shader used for SSR + transparent:
Before:
https://user-images.githubusercontent.com/55133890/108783522-ed6a8c00-756d-11eb-81e5-f47d94b70017.mp4
After the fix:
3xSwL3a9pO

As a side effect, this PR fixes some "white flash" when switch some state related to SSR


remi-chapelain (approved):
Quickly tested multiple combinations of HDRP Asset / Frame Settings / Transparents / SSR  and camera setup to be sure that this does not repro anywhere. Looks good ✔️ 

However, while testing, I most probably found a fairly recent regression that is **NOT** introduced by this PR https://fogbugz.unity3d.com/f/cases/1316840/

sebastienlagarde (approved):

PR #3604, opened 2021-02-23
2d/custom lit

UnityPaul:
### Purpose of this PR
Exposes an internal RT for use with custom 2D Lighting model

---
### Testing status
056_2D_Lights_Shader_Graph expanded to include sprites rendered using custom lighting


unity-cchu (approved):

phi-lira (changes requested):
You shouldn't change UniversalGraphicsTest project anymore. Otherwise LGTM.

phi-lira (approved):

com.unity.render-pipelines.universal/Editor/2D/ShaderGraph/Nodes/LightTextureNode.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    enum BlendStyle
+    {
+        LightTex0,
+        LightTex1,
+        LightTex2,
+        LightTex3,
+    }
+
+    [Title("Input", "2D", "Light Texture")]
```
cdxntchou, line 18: Will this node will only work when you have a Sprite Custom Lit subtarget on the graph?

UnityPaul: it will work with any 2D sub targets... but the CustomLit adds a normal pass that's in the Lit target but not in the Unlit target...

cdxntchou: My worry here is that users will add this node to non-2D graphs, and be confused when it doesn't work.  We have the ability to flag nodes as only working on a specific Target, so it gives a nice warning tag message when used inappropriately.  It might be good to extend that to SubTargets as well, so we can flag this node as only working with the 2D subtargets...

cdxntchou: This check could be added by extending UniversalTarget.cs IsNodeAllowedByTarget() to also ask the subtarget if it allows the node.   The subtarget can check for a filter attribute similar to SRPFilterAttribute -- would have to add a new filter attribute like "SubTargetFilterAttribute",  or maybe more specifically for this use case:  "2DSubTargetOnlyAttribute"

cdxntchou (changes requested):
Make the custom lighting an option on the existing subtarget (reduces copy/paste and I think it's simpler for the user)

Add a check to disallow the new node except when using 2D subtargets.

com.unity.render-pipelines.universal/Editor/2D/ShaderGraph/Nodes/LightTextureNode.cs
UnityPaul: Done... please take a look

UnityPaul:
> Make the custom lighting an option on the existing subtarget (reduces copy/paste and I think it's simpler for the user)
> 
> Add a check to disallow the new node except when using 2D subtargets.
> 
> The rest looks great!

Done

cdxntchou (approved):
Very nice!

erikabar (approved):
lgtm

BoonKiangTan:
Feature verified on branch:

Version: 2021.2.0a8.1261
Revision: trunk 05e6d22b19f7
Built: Wed, 03 Mar 2021 03:25:48 GMT
Branch: 2d/custom-lit (6f684c4522)

- Verified 2D Light Texture node works as intended (changing Light Tex 0-4 changes which lights affect the material based on its blend style)
- No issues with shadergraph output (tried adding, subtracting, multiplying light tex with main tex, no issues)
- Disabling Custom Lit in graph settings throws error in console (intended)

BoonKiangTan (approved):

phi-lira:
Hybrid test failures not introduced in this PR.

PR #3607, opened 2021-02-23
Fix cached directional update upon rotation and add API to don't render on placement for OnDemand

FrancescoC-unity:
This PR addresses a couple of things: 

 - Add an API to allow to allow an OnDemand shadow to not render upon enabling/registering (i.e. needing to explicitly request the update even for the first time). 
 - Fix directional light update upon rotation, previously it would end up rendering every frame even with no changes going on after the first change. 
 - Expose the "Update on Light Movement" option to UI for directional lights too. 

Note: this includes the changes already merged in master from https://github.com/Unity-Technologies/Graphics/pull/3387 

**What did I test**: 

- Tried the new API by verifying that with it set to true (default) and OnDemand light will update on enabling/registering. With the new API set to false, the shadow won't render when the light is first registered with the Atlas as expected. 
- Tried the update on light movement and noticed that the directional light updates only if the angle difference is above the set threshold.


sebastienlagarde (approved):

sebastienlagarde:
will  retarget to master after merge of hd/bugfix

PR #3608, opened 2021-02-23
Decal in material debug display

adrien-de-tocqueville:
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1316568/

This PR fix the material debug display for decals (for normal, albedo, ...) on transparent/opaque objects and in forward/deferred by binding the decal buffers to the shaders
This requires computing the gpu light list during debug display for decals on transparent objects

It also fixes a wrong label on decal projector mask opacity channel under smoothness (which was labelled normal opacity channel)

And it also fixes the mask opacity option on normal when decal only affects normal (the mask opacity slider wasn't taken into account for the global fade factor computation)

---
### Testing status
tested the repro steps of the fogbug


sebastienlagarde (approved):

sebastienlagarde:
Added @JulienIgnace-Unity for RG check

JulienIgnace-Unity (approved):

remi-chapelain (approved):
I checked the 2 others issues that were fixed in this PR in another project. 
The label and the mask opacity for normal alpha source ! 
nice find.

PR #3609, opened 2021-02-23
Missing lighting quality settings for SSGI (case 1312067).

anisunity:
https://fogbugz.unity3d.com/f/cases/1312067/
Pretty straight forward addition.

Testing status
Opened an asset and made sure the values displayed where what they are expected to be


remi-chapelain (approved):
LGTM ✔️

sebastienlagarde (approved):

sebastienlagarde (commented):
Question: does this code support migration?   i.e does it break existing SSGI settings?

remi-chapelain:
Indeed, since low/med/high have changed, it will erase your settings if you were using a quality settings other than "Custom"

anisunity:
The default quality settings were not serialized before.  so it was always picking the values currently described in the code. This PR only exposes them (and makes them serializeable)

sebastienlagarde:
> I'm guessing, this will impact very few users since SSGI was hardly usable in its previous state and since performance is better now, even if the quality settings gets overriden, it will be for a better results at an equal or lower cost.. so win/win ?

So it could be acceptable but in this case we must say it in the upgrade guide for 21.2.  (that it need to be re-author in this context)

sebastienlagarde:
Note for me: need to re-merge master once-hd/bugfix is merged.

PR #3614, opened 2021-02-23
Fix a leak cause for area light cookies 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1316692/ 

The issue here is that the first blit in the convolution process is into the shared common texture used as a temp space for the rest of the convolution. To avoid some issues with whatever the previous content was previously there we add a border of 1 texel around the texture. 

**What did I test**: the repro case and noticed that now it looks the same as if that was the only cookie in the atlas. Area light tests still pass locally.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/CookieSampling.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/CookieSampling.hlsl
@@ -18,7 +18,7 @@ float2 RemapUVWithPadding(float2 coord, float2 size, float rcpPaddingWidth, floa
     float2 offset = 0.5 * (1.0 - scale);
 
     // Avoid edge bleeding for texture when sampling with lod by clamping uvs:
-    float2 mipClamp = pow(2, lod) / (size * COOKIE_ATLAS_SIZE * 2);
+    float2 mipClamp = pow(2, ceil(lod)) / (size * COOKIE_ATLAS_SIZE * 2);
     return clamp(coord * scale + offset, mipClamp, 1 - mipClamp);
 }
 


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightCookieManager.cs
```diff
@@ -181,8 +181,13 @@ Texture FilterAreaLightTexture(CommandBuffer cmd, Texture source, int finalWidth
                     m_MPBFilterAreaLights.SetInt(s_sourceMipLevel, 0);
                     m_MPBFilterAreaLights.SetTexture(s_texSource, source);
 
+
+                    // Since we blit the cookie texture into a common texture, to avoid leaking we blit with an extra border
+                    int border = 1;
                     cmd.SetRenderTarget(m_TempRenderTexture0, 0);
-                    cmd.SetViewport(new Rect(0, 0, viewportWidth, viewportHeight));
+                    cmd.SetViewport(new Rect(0, 0, viewportWidth + border, viewportHeight + border));
```
sebastienlagarde, line 188: wondering if it cause any atlas fitting issue? i.e you have a 256x256 atlas with a 256x256 texture and then it don't fit anymore because of the 1 border. Maybe we should "rescale" ? i.e add a border of 1 and render in 254x254 ? Thought

FrancescoC-unity: So as it is now it will fit, it will just not put the border around it which is fine as there is no data outside to be messed with. 

This only changes the viewport, not the size of the texture, so in your example what would happen is the border will end up outside the rendertarget, but that should be fine as in that case we only have the one thing.

remi-chapelain (changes requested):
I wanted to test a more packed atlas in terms of color constrast (less black space) and I still manage to repro easily the issue if I push the atlas to its limit.
However, it fixes the "simple" repro case in the bug and it helps a bit with the new one.  

https://user-images.githubusercontent.com/57442369/108983987-bb514b00-768f-11eb-8c01-681490351e29.mp4

here's the repro case if needed
[case_1316692_v2.zip](https://github.com/Unity-Technologies/Graphics/files/6035098/case_1316692_v2.zip)

remi-chapelain (approved):
Tested again different arrangement of cookie atlas with different cookies size. Can't repro the leak on the size of the black and white grid cookie ! 
Also quickly tested if this didn't introduced regression with punctuals light classic & IES cookie ! 

**Before**
image

**After**
image

sebastienlagarde (approved):

sebastienlagarde:
please, add a changelog

PR #3611, opened 2021-02-23
Added information about rt scene validation

JordanL8:
### Purpose of this PR
Added information about rt scene validation.

Fixes https://fogbugz.unity3d.com/f/cases/1315559

PR #3616, opened 2021-02-23
Fixed the "double sided" mode not stacking properly on multi-materials (case 1314963).

anisunity:
https://fogbugz.unity3d.com/f/cases/1314963/
The double sided is unique to every instance (meaning it cannot be specified for every submesh). This is not a bug, it a wanted behavior. This PR adds double sided verification in the check scene for ray tracing steps. It was added to the documentation in an other PR.

https://github.com/Unity-Technologies/Graphics/pull/3611

Testing status
Checked it on the scene in the repro
image


remi-chapelain (changes requested):
Can you update the docs to reflect the addition of the ray tracing menu ? or maybe @JordanL8 ?
https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@11.0/manual/Menu-Items.html#other-items

JordanL8:
Docs in here: https://github.com/Unity-Technologies/Graphics/pull/3611

remi-chapelain (approved):

sebastienlagarde (approved):

PR #3617, opened 2021-02-23
[Universal] fix case 1306432 material upgrade issue

martint-unity:
---
### Purpose of this PR
Fixes an issue in where baseshadergui has an enum value that is reversed from our materials. So there is a mismatch causing some of our upgrades to have the wrong value.
FB case 1306432

---
### Testing status
Tested with the testcase inside the FB case. 

---
### Comments to reviewers
This needs thorough testing with all differnet kind of materials that uses the smoothness source.


ValentinHume (approved):
Looks good, tested on standard and all legacy shaders using smoothness source

Verasl (approved):
LGTM

phi-lira (approved):

erikabar (approved):
approving, developer already tested with the repro project.

PR #3624, opened 2021-02-23
Volumetric Fog and Volumetric Cloud bug fixes

anisunity:
- Fixed a limit case when the camera is exactly at the lower cloud level (case 1316988).
- Fixed the various history buffers being discarded when the fog was enabled/disabled (case 1316072).

Testing status
The two issues do not repro anymore.


remi-chapelain (approved):
After discussing with @anisunity what has been tested in depth, I don't think additionnal testing is required !

sebastienlagarde (approved):

sebastienlagarde:
note for myself: retarget to master once hd/bugfix is merge, shouldn't be backported to 10.x

sebastienlagarde (approved):

PR #3627, opened 2021-02-24
Fixing planar reflection probes issues [FogBugz 1312153]

kecho:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1312153/

Fixing multiple issues regarding exposure with Planar Relfection Probes (PRP).

* Fixed issues with ExposureControl FrameSetting  in Planar reflection volume
* when autoexposure is on and exposure control is on for the PRP, the PRP uses an empty exposure texture (1x) and a stale exposure scaler for sampling.
This would cause the PRP to look sometimes bright, sometimes dark, sometimes correct.
* Unlit surfaces would also render direclty into the raw PRP hdr surface. This causes reflections of unlit to be different when ExposureControl is turned off the PRP.

image
image

* Now the PRP uses the correct exposure of its parent camera to pre-expose when auto exposure is turned on.
* Adding support to de-expose the unlit surfaces when ExposureControl is turned off on the PRP, this makes its reflection to be always the same, regarldess of the state of ExposureControl frame setting for the PRP.
image

* Moved the computation of the probe exposure to the camera. If we have say the editor and the game view open at the same time, it would cause flickering and multiple pre-exposure instabilities.
* When in auto exposure mode, this change ensures that the parent view's exposure textures are actually used for the PRP.  This change will also make it trivial to pre-expose any other proxy reflection volume if desired.
* Fixed stalling bug on exposure: when in editor mode, sometimes the exposure async queue would stall for many frames (I was seeing it go up to 800 sometimes). When you play, this issue goes away. To fix this, I added a on editor hack only that force updates each async request.
* Fixed Issues on UI rendering
When we selected fixed exposure, now the PRV is pre exposed properly, however there is no sampling multiplier used when showing the PRV in the UI:
image
The UI Also works correctly in all exposure modes (fixed & auto):
image

---
### Testing status
Expecting 2208_PlanarReflectionCullingStencil to be updated, since unlit geo reflections are now correct:
image

* Ran all Planar reflection tests. Must update 2208, running all yamato atm.

---


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -319,14 +322,7 @@ public void BeginFrame(CommandBuffer cmd, HDCamera camera, HDRenderPipeline hdIn
                     }
                 }
 
-                // Note: GetExposureTexture(camera) must be call AFTER the call of DoFixedExposure to be correctly taken into account
-                // When we use Dynamic Exposure and we reset history we can't use pre-exposure (as there is no information)
-                // For this reasons we put neutral value at the beginning of the frame in Exposure textures and
-                // apply processed exposure from color buffer at the end of the Frame, only for a single frame.
-                // After that we re-use the pre-exposure system
-                RTHandle currentExposureTexture = (camera.resetPostProcessingHistory && !isFixedExposure) ? m_EmptyExposureTexture : GetExposureTexture(camera);
```
kecho, line 327: This was one of the problems. THe logic to get the exposure texture here does not match the logic to get the exposure to sample the PRP

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -742,18 +738,24 @@ ExposureParameters PrepareExposureParameters(HDCamera hdCamera)
 
         public RTHandle GetExposureTexture(HDCamera camera)
         {
+            // Note: GetExposureTexture(camera) must be call AFTER the call of DoFixedExposure to be correctly taken into account
+            // When we use Dynamic Exposure and we reset history we can't use pre-exposure (as there is no information)
+            // For this reasons we put neutral value at the beginning of the frame in Exposure textures and
+            // apply processed exposure from color buffer at the end of the Frame, only for a single frame.
+            // After that we re-use the pre-exposure system
+            if (m_Exposure != null && camera.resetPostProcessingHistory && !IsExposureFixed(camera))
```
kecho, line 746: I moved the logic here. Now its consistent

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -742,18 +738,24 @@ ExposureParameters PrepareExposureParameters(HDCamera hdCamera)
...
+                return m_EmptyExposureTexture;
+
             // 1x1 pixel, holds the current exposure multiplied in the red channel and EV100 value
             // in the green channel
-            // One frame delay + history RTs being flipped at the beginning of the frame means we
-            // have to grab the exposure marked as "previous"
-            var rt = camera.GetPreviousFrameRT((int)HDCameraFrameHistoryType.Exposure);
+            var rt = camera.currentExposureTextures.current;
```
kecho, line 751: Abstracted the complexity here, and moved the logic to the hdCamera instead. This allows us to also inherit exposure texture when we are say, rendering a PRP camera.

com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/PlanarReflectionProbeEditor.cs
```diff
@@ -193,6 +193,7 @@ void OnOverlayGUI(Object target, SceneView sceneView)
                 // Setup the material to draw the quad with the exposure texture
                 var material = GUITextureBlit2SRGBMaterial;
                 material.SetTexture("_Exposure", exposureTex);
+                material.SetFloat("_ExposureBias", (float)Math.Log(1.0f / p.ProbeExposureValue(), 2.0));
```
kecho, line 196: This fixes the UI, so it doesnt blow up when the probe is pre-exposed.

com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/PlanarReflectionProbeEditor.cs
```diff
@@ -299,7 +300,7 @@ static void DrawCapturePositionGizmo(PlanarReflectionProbe probe)
 
             k_PreviewMaterial.SetTexture("_MainTex", probe.texture);
             k_PreviewMaterial.SetMatrix("_CaptureVPMatrix", vp);
-            k_PreviewMaterial.SetFloat("_Exposure", 1.0f);
+            k_PreviewMaterial.SetFloat("_Exposure", (float)Math.Log(1.0 / probe.ProbeExposureValue(), 2.0));
```
kecho, line 303: This fixes the UI, so it doesnt blow up when the probe is pre-exposed.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -128,6 +126,18 @@ float aspect
         // Array of names that will be used in the Render Loop to name the probes in debug
         internal string[] probeName = new string[6];
 
+        float m_ProbeExposureValue = 1.0f;
```
kecho, line 129: Moved the exposure async request logic to the camera. Instead the probe just caches the exposure value.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1804,7 +1804,7 @@ internal bool GetEnvLightData(CommandBuffer cmd, HDCamera hdCamera, in Processed
                     //capturedForwardWS.z *= -1; // Transform to RHS standard
                     m_TextureCaches.env2DCaptureForward[fetchIndex] = new Vector4(capturedForwardWS.x, capturedForwardWS.y, capturedForwardWS.z, 0.0f);
 
-                    if (probe.frameSettings.IsEnabled(FrameSettingsField.ExposureControl))
+                    if (probe.ExposureControlEnabled)
```
kecho, line 1807: We must use the setting resolved from the probe, not from the frameSettings. Using the frmaeSettings from the probe is wrong because it can be disabled (not ticking on using custom frame settings in the probe reflection component), and this code was not checking that.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -237,7 +238,34 @@ internal struct VolumetricCloudsAnimationData
         internal uint                   cameraFrameCount = 0;
         internal bool                   animateMaterials;
         internal float                  lastTime;
-        internal Camera                 parentCamera = null; // Used for recursive rendering, e.g. a reflection in a scene view.
+
+        private  Camera                 m_parentCamera = null; // Used for recursive rendering, e.g. a reflection in a scene view.
+        internal  Camera                 parentCamera { get { return m_parentCamera; } }
+
+        internal void SetParentCamera(HDCamera parentHdCam)
```
kecho, line 245: Set parent camera now is a bit more complex: it actually caches the exposure textures of the parent camera.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1559,15 +1559,32 @@ ref _cullingResults
...
+                                // For the clear color, we need to do it manually here.
+                                additionalCameraData.backgroundColorHDR = additionalCameraData.backgroundColorHDR * visibleProbe.ProbeExposureValue();
+                            }
+                            else
+                            {
+                                //the de-exposure multiplier must be used for anything rendering flatly, for example UI or Unlit.
+                                //this will cause them to blow up, but will match the standard nomralized exposure.
+                                hdParentCamera.RequestGpuDeExposureValue(GetExposureTextureHandle(hdParentCamera.currentExposureTextures.previous));
+                                visibleProbe.SetProbeExposureValue(1.0f);
+                                additionalCameraData.deExposureMultiplier = 1.0f / hdParentCamera.GpuDeExposureValue();
```
kecho, line 1586: The final goal is for unlit / UI to have an exact luminance when rendered. To achieve this, we 'blow them up' to the opposite exposure, so that when they go through exposure, its the desired nits specified in the unlit/UI material.
With pre exposure this is not necesary, since we are rendering in a consistent relative luminance.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -357,6 +385,134 @@ internal GameObject exposureTarget
...
+
+            var currentTexture = GetCurrentFrameRT((int)HDCameraFrameHistoryType.Exposure);
+            if (currentTexture == null)
+            {
+                RTHandle Allocator(string id, int frameIndex, RTHandleSystem rtHandleSystem)
+                {
+                    // r: multiplier, g: EV100
+                    var rt = rtHandleSystem.Alloc(1, 1, colorFormat: HDRenderPipeline.k_ExposureFormat,
+                        enableRandomWrite: true, name: $"{id} Exposure Texture {frameIndex}"
+                    );
```
sebastienlagarde, line 502: I guess this code need to be updated with this fix: https://github.com/Unity-Technologies/Graphics/pull/3867 can you check?

sebastienlagarde: Also Poking @FrancescoC-unity  just to have a quick check of this PR

sebastienlagarde (approved):
Please add various comment in the code (you commented the PR itself but would like to have that in the code :)

Also add a changelog

sebastienlagarde:
I don't see update of Yamato screenshots in the PR, expected?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
FrancescoC-unity: Re: #3867 the textures are different ones (my PR is on the artist driven exposure curve).

Will make a pass on the rest of PR soon

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -286,33 +284,23 @@ void BeginPostProcessFrame(CommandBuffer cmd, HDCamera camera, HDRenderPipeline
...
             {
                 // Fix exposure is store in Exposure Textures at the beginning of the frame as there is no need for color buffer
                 // Dynamic exposure (Auto, curve) is store in Exposure Textures at the end of the frame (as it rely on color buffer)
                 // Texture current and previous are swapped at the beginning of the frame.
                 bool isFixedExposure = IsExposureFixed(camera);
-                if (isFixedExposure)
+
+                //if exposure comes from the parent camera, it means we dont have to calculate / force it.
+                //Its already been done in the parent camera.
+                if (isFixedExposure && camera.exposureControlFS && camera.currentExposureTextures.useCurrentCamera && camera.currentExposureTextures.current != null)
```
FrancescoC-unity, line 295: I'd like these conditions a bit more condensed (maybe in a separete fun call like the IsExposureFixed one)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
FrancescoC-unity: Uhm I think for those use cases the objects should be simply rendered after post process, unless I am misunderstanding something here (will come to it later to re-read)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1559,15 +1559,32 @@ ref _cullingResults
                             visibleProbe.ForceRenderingNextUpdate();
                         }
 
-                        hdCamera.parentCamera = parentCamera; // Used to inherit the properties of the view
+                        hdCamera.SetParentCamera(hdParentCamera); // Used to inherit the properties of the view
+                        hdCamera.resetPostProcessingHistory = false;
```
FrancescoC-unity, line 1563: Why do we need this reset to off?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
kecho: The exposure curve PR should be fine. Will ask QA to test all exposure modes.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
kecho: hmmm not sure. I think it might be a typo. Testing without it

kecho: Actually this is needed, but is done the wrong way. resetPostProcessingHistory is always true on probe cameras, so without this it would return always an invalid exposure texture.
The correct fix is to ignore this flag if the exposure texture is inherited from the parent. Will fix this right away.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
kecho: Unlit renders directly into the pre post process color buffer. This fixes the issues when having unlit rendering into a planar reflection probe camera. A planar reflection probe camera stores radiance, hence there are no post effects in a PRP camera.

The unlit formula is:
color + emissive*preexposure, which gets rendered into the color buffer of the planar reflction probe.
If the color buffer is pre-exposed, it means that the value that you send in "color" will be exactly that value when sampling the PRP.

If the color buffer of the PRP is not pre-exposed, it means that color will be going through exposure, which will make it darker.
For unlit, i believe the intention is to have that color value be exactly that value, after we apply tone mapping in the main frame.
Thus, if we pre multiply the color times a de exposure value, once the reflection plane gets exposed by the main camera, we will have a sort of accurate unlit color. The alternative is to apply an inverse tonemapping, or even worse, do render planes on the reflection plane itself. 
I think this solution works pretty well for the use case. The only artifact is a slight delay on unlit surfaces in the reflection catching up wtih the current frame's exposure.
Hope that makes sense

iM0ve (approved):
Looks good on my part.

In addition to existing testing I tried a few things on a real-life project.

* Changing Planar Probe resoltuion
* Split Scene/Game view
* Used Unlit and Lit emissive material
* Different exposures and time of day
* Verified bug in the fobugz scene

Unity_f0nO02evl9
Unity_g2F7leNAyN
yOH3OXqlFU

FrancescoC-unity (approved):

sebastienlagarde:
note for myself: retarget this PR to target

kecho:
@sebastienlagarde this branch is ready to go! 👍

PR #3628, opened 2021-02-24
Fixing ordering of UI Frame settings [Fogbugz 1315452]

kecho:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1315452/

Fixing order of settings in the UI. Orders were a bit messed on on the attribute. This attribute is safe to modify since its entirely used for UI purposes.

---
### Testing status
Checked ordering is as expected, and enum Refraction isn't hidden anymore.

image


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -146,65 +146,66 @@ public enum FrameSettingsField
...
         TransparentsWriteMotionVector = 16,
 
-        /// <summary>When enabled, HDRP processes a refraction render pass for Cameras using these Frame Settings. This add a resolve of ColorBuffer after the drawing of opaque materials to be use for Refraction effect during transparent pass.</summary>
-        [FrameSettingsField(0, autoName: Refraction, customOrderInGroup: 15, tooltip: "When enabled, HDRP processes a refraction render pass for Cameras using these Frame Settings. This add a resolve of ColorBuffer after the drawing of opaque materials to be use for Refraction effect during transparent pass.")]
-        Refraction = 13,
+        //NOTE: Obsoletes must precede the proper enums, otherwise the compiler occludes them.
         /// <summary>When enabled, HDRP processes a refraction render pass for Cameras using these Frame Settings. This add a resolve of ColorBuffer after the drawing of opaque materials to be use for Refraction effect during transparent pass.</summary>
         [Obsolete]
         RoughRefraction = 13,
```
kecho, line 152: Moved the obsolete above. If the obsolete is below, the compiler will completely override the value of "Refraction". So even if we get the enums by name, it won't work.

victorsclui:
It looks good, and is more complete than the original fix, since we also make sure FrameSettingsFieldAttribute have the correct displayedName field.

The only thing I can think of is that since we are adding a static constructor, this sharing from the DOTS team may be of interest (https://unity.slack.com/archives/CE7DZN2H1/p1592525582345000).

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/OverridableFrameSettingsArea.cs
```diff
@@ -27,9 +27,12 @@ static OverridableFrameSettingsArea()
...
             attributesGroup = new Dictionary<int, IOrderedEnumerable<KeyValuePair<FrameSettingsField, FrameSettingsFieldAttribute>>>();
             Type type = typeof(FrameSettingsField);
-            foreach (FrameSettingsField value in Enum.GetValues(type))
+            foreach (string enumName in Enum.GetNames(type))
             {
-                attributes[value] = type.GetField(Enum.GetName(type, value)).GetCustomAttribute<FrameSettingsFieldAttribute>();
+                if (type.GetField(enumName).GetCustomAttribute<ObsoleteAttribute>() != null)
+                    continue;
+
+                attributes[(FrameSettingsField)Enum.Parse(type, enumName)] = type.GetField(enumName).GetCustomAttribute<FrameSettingsFieldAttribute>();
```
victorsclui, line 35: Since we already have static field FrameSettingsFieldAttribute.s_FrameSettingsAutoNames can we reuse this Dictionary? So we just iterate over s_FrameSettingsAutoNames here which gives us both enumName and Enum.Parse(type, enumName) from its iterator.

victorsclui (commented):
It looks good and is very complete since we also make sure FrameSettingsFieldAttribute have the correct displayedName field.

Besides the comment about static dictionary reuse, since we are adding a static constructor, this sharing from the DOTS team may be of interest (https://unity.slack.com/archives/CE7DZN2H1/p1592525582345000).

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/OverridableFrameSettingsArea.cs
kecho: good point. I would have to make this a static cacheable thing since the ordering of static constructors is not guaranteed !

kecho: Done. Latest commit should have exactly this. I looked into the post you mentioned, it might be too overkill. Well let me know your thoughts.

victorsclui: yea thinking about it editor build always runs on mono and that attribute works for il2cpp so it will not do much sorry

victorsclui (approved):

sebastienlagarde (approved):

sebastienlagarde:
Please add a changelog

kecho:
Changelog added

remi-chapelain (approved):
Just tested that the refraction frame settings worked and had the proper behavior (i.e fallback to opaque for every material that has a refraction flag)

PR #3629, opened 2021-02-24
Do not serialize UI properties on VolumeParameters

alex-vazquez:
---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1312464/

---
### Testing status
- Check the problem of the bug is solved
- Check other parameters
- Change the max to a high number, change the value to that number, decrease the max.

---
### Comments to reviewers
The main problem is that the properties used to modifiy the UI values are serialized, this is correct, but the range values are only set on the constructor,( something that is specified on the Volume Component), therefore they do not need to be serialized.

Just adding the attribute makes them to recover on whatever is on the constructor.


remi-chapelain (approved):
Testing seems sufficient on this one ! ✔️

RSlysz (approved):
as long as the value is initialized by the constructor, it is good

alex-vazquez:
@Unity-Technologies/gfx-qa-hdrp , could you please do a quick pass on this PR?

remi-chapelain (commented):
What I did test : 
- Tested to reduce max value of a ClampedParameter ✔️ 
- Tested to increase min value of a ClampedParameter ✔️ 
- Tested to modify quality settings of one override to see if upgrade behaved properly ✔️ 
- Tested to add/remove item of an enum ✔️ 

The only thing that "sort of fail" in "common" parameters is upgrading a ColorParameter from hdr:false to hdr:true, you need to reset the component (or remove/add it) to make HDR info to appear in the color picker
However, Changing parameters to showAlpha gets shown instantly in the override in this PR (In master you still need to reset the override)

https://user-images.githubusercontent.com/57442369/109840022-f7644d00-7c47-11eb-9d5c-8577f44fa54d.mp4

public ColorParameter color = new ColorParameter(Color.black, hdr: false, showAlpha: false, showEyeDropper: false);

Also, not related to this PR, but the showEyeDropper is always present wether you set it to true or false.   ¯\\_(ツ)_/¯

RSlysz (approved):

remi-chapelain (approved):
Thanks, this fixes the HDR issue. 

As for what I said about the eyeDropped, I was looking at the wrong thing, it's working as expected !

PR #3634, opened 2021-02-24
Display a warning help box when decal atlas is out of size

adrien-de-tocqueville:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1317163/

It's more a UX improvement as the issue was that the decla atlas was out of space. Now a warning message is still logged in the console but a text is also displayed in the decal projector inspector, if the affect transparent materials is ticked.
The issue only appears on transparents because opaque objects don't use the decal atlas.

image

---
### Testing status

tested on the repro project, where the decal atlas size is smaller than the needed size of the textures


remi-chapelain (approved):
I little redundancy can't hurt ✔️  :)

sebastienlagarde (approved):

PR #3644, opened 2021-02-24
[HDRP] Fix volume component nullref

alelievr:
### Purpose of this PR
Fix case https://fogbugz.unity3d.com/f/cases/1317156/

Apparently, the `AssetDatabase.SaveAssets()` function caused the serializedObject of the volume to be invalid and thus all its serialized properties were broken when we tried to access them to create the volume editor. Moving the `SaveAssets` just after the creation of the volume editor seems to fix the issue.

---
### Testing status
Tested adding and removing components on a Volume profile from a GameObject in the scene, from an asset on the disk and a preset on the disk. Everything looks correct.

---
### Comments to reviewers
@Chman Do you know why the AssetDatabase Save was made between the code that gets the new serialized property for the volume and the editor creation?

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -56,6 +56,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed decals in material debug display.
 - Fixed WouldFitInAtlas that would previously return wrong results if any one face of a point light would fit (it used to return true even though the light in entirety wouldn't fit).
 - Fixed loss of persistency of ratio between pivot position and size when sliding by 0 in DecalProjector inspector (case 1308338)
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

remi-chapelain (approved):
This looks safe and don't have much more testing to do. ✔️ 
However, discussing it with @iM0ve, we were curious to understand why this regressed ? Do you have more info about it ? :)

alelievr:
No idea :) could be an AssetDatabase behavior that changed recently too (though I'm not really sure why re-serializing the asset would cause its serialized references to break)

Chman:
Definitely looks like a regression on the ADB side. To be honest I can't remember the rational for this specific part of the code but I know it was _highly_ sensitive to the order in which things are done (at the time anyway, and it was made with ADBv1). Something to do with the order of serialization, Unity writing the file to disk, the editor refresh sequence and whatnot. As long as it works now it looks fine though.

PR #3651, opened 2021-02-25
[Fix] case 1313954: Add TestRunnerBake tag to 006 scene in _Lighting project

he-ax:
Add TestRunnerBake tag to scene 006.
Fogbugz case 1313954.
Reference images for 006 seems to already be updated.


phi-lira:

PR #3655, opened 2021-02-25
[10.x.x][VFX] Fix VFX graph shader code generation error with empty textures in shader graph

gabrieldelacruz:
---
### Purpose of this PR
Integrating PR #3596 

There were a few regression recentlys, when generating vfx graph shader code from shader graph.
- If there is a texture node that is empty, we are not able to properly declare the texture in the shader code. To solve this we had to pass the TextureDimension along with the texture (even though it is redundant when we have a one) to create the correct VFXExpression.
- Shader graph would also produce shader compile errors if we had an exposed texture.
- Finally, for URP, we were missing an include in shaders: "Packages/com.unity.render-pipelines.core/ShaderLibrary/Texture.hlsl"

---
### Testing status
QA pass on HDRP and URP.

---
### Comments to reviewers
Texture2D has different kinds of default textures, but in VFX graph it will always default to gray. This was like this before, so it is not a regression. There is already a fogbugz for that (https://fogbugz.unity3d.com/f/cases/1114588/)


julienf-unity (approved):

VladNeykov (approved):
Verified fixes, looking good

PR #3665, opened 2021-02-26
[2021.2][12.x] Shader Graph Blackboard refactor to MVVM+Redux paradigm

Nightmask3:
---
### Purpose of this PR

This PR is part of the work effort described here: 

https://jira.unity3d.com/secure/RapidBoard.jspa?rapidView=2564&projectKey=GSG&view=planning&selectedIssue=GSG-265&epics=visible&issueLimit=-1&selectedEpic=GSG-258

The PR has three purposes:

- To laterally refactor all of the Blackboard functionality to switch over to the paradigm used by Graph Tools Foundation, in preparation for the GTFO later on.

- To enable the easier implementation of blackboard categories (or) blackboard property groups, as part of the Blackboard workflow improvements and URP as default initiative. 

- Remove the cruft that has accumulated in the UI layer of Shader Graph over time by introducing new more formal constructs and patterns for how the UI state and interactions are managed.

This PR also fixes this bug:

https://fogbugz.unity3d.com/f/cases/1264894/

Three minor things this PR changes/improves:

- Inspector now has max number of elements it will display (currently set to 20 in the code)

In the inspector, in favor of having a faster UX when selecting large numbers of elements, we now limit the max number of elements you can see in the inspector to 20, and once selections cross that number, the inspector will no longer display additional elements at the bottom of the window.

This is what would happen in 2020.2 once you cross more than 30 selections (the cursor doesn't show so its a little hard to make out, but note how long it takes for each element to be highlighted before I progress to the next one, that's me as a user waiting for the application to provide the feedback of the highlight-ing/selection action before moving to my next operation)

https://user-images.githubusercontent.com/13033542/110173712-e0db0480-7db3-11eb-9fb0-3795bc7d1059.mp4

The same selection behavior now (at around 80-90 selections):

https://user-images.githubusercontent.com/13033542/110173910-2c8dae00-7db4-11eb-8c14-99a27e155a0a.mp4

- Blackboard hides the padding around the scrollbars when doing the property drag and drop + scrolling, and unhides it when drag/drop is done, so as to present a slightly cleaner UI when doing the property scrolling operations:

New behavior:

https://user-images.githubusercontent.com/13033542/110174959-d02b8e00-7db5-11eb-97fc-1ce107c48b4c.mp4

- Blackboard now properly handles selection persistence of items, with a less hackier way of doing things than before:

Old behavior:

https://user-images.githubusercontent.com/13033542/110175954-690ed900-7db7-11eb-9cb8-576543fce87f.mp4

New behavior:

https://user-images.githubusercontent.com/13033542/110176022-85127a80-7db7-11eb-8482-c031adef8103.mp4

---
### Testing status

- [x] Ran edit mode tests and they passed
- [x] Adding inputs by the '+' menu
- [x] Adding inputs by copy/pasting single and multiple inputs 
- [x] Adding inputs by duplicating inputs (both with right click menu and Ctrl+D hotkey)
- [x] Adding inputs by dragging in inputs from other graphs and subgraphs
- [x] Adding inputs of all types
- [x] Removing inputs by deleting single and multiple inputs (using Delete key)
- [x] Removing inputs by right-click menu, single and multiple inputs 
- [x] Moving inputs around using drag and drop, single and multiple inputs 
- [x] Blackboard item selection persistence after undo/redo 
- [x] Renaming blackboard properties, both through double-clicking the Pill and right click rename menu
- [x] Reset reference menu on right clicking input still works
- [x] Ensured that inputs get saved and loaded in with no data loss
- [x] Ensured that input display name being changed by inspector is reflected in blackboard
- [x] Tested blackboard property scrolling regions to ensure they still work and behave as before

### Performance Stress tests:
- [x] Tested duplicating ~100 items
- [x] Tested undoing above duplication of ~100 items
- [x] Tested deleting ~100 items
- [x] Tested undoing above deletion of ~100 items
- [x] Tested converting 100 inline nodes to properties   (This and the undo of it are somewhat slow, but tested the same thing in 2020.2 and it's identical performance wise so its not a regression)
- [x] Tested undoing above conversion of 100 inline nodes to properties
- [x] Tested converting 100 property nodes to inline nodes (This and the undo of it are somewhat slow, but tested the same thing in 2020.2 and it's identical performance wise so its not a regression)
- [x] Tested undoing above conversion of 100 property nodes to inline nodes
    
### Current bugs:

- [x] Undo/redo support for all the above actions
    - [x] Redo is broken for all actions
    - [x] Undo doesn't work for renaming blackboard items
- [x] Changing property default values throws a NullReferenceException
- [x] Performance regression on deleting items and adding items and then undoing that action (fixed, needs verification)
- [x] Regression on blackboard items being highlighted when property node associated with that item is hovered on graph (fixed, needs verification)
- [x] Convert to property and Convert to Inline actions would break the blackboard (fixed, needs verification)
---
### Notes for reviewers:

Things to do still:

- [x] Add automated tests
     - [x] Remove reflection code from `ShaderGraphUITestHelpers` 
- [x] Remove TODO comments 
- [x] Add feedback to the Inspector for having reached max inspectable elements count


LandonTownsendUnity (changes requested):
Redo is not working for any operation. Additionally, undo is not working for renaming. It says undo/redo was tested; did a later commit break it after testing?

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
```diff
@@ -1,21 +1,478 @@
...
+
+        public string NewGraphPath { get; set; }
+    }
+
+    class CopyShaderInputAction : IGraphDataAction
+    {
+        void CopyShaderInput(GraphData graphData)
+        {
+            Assert.IsNotNull(graphData, "GraphData is null while carrying out CopyShaderInputAction");
+            Assert.IsNotNull(shaderInputToCopy, "ShaderInputToCopy is null while carrying out CopyShaderInputAction");
```
marctem, line 75: I don't think we want to land Asserts in our real code - that'll throw an uncaught exception that will put the graph in a bad state. Unless they're compiled out? I'm not actually sure.

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
```diff
@@ -1,21 +1,478 @@
...
+        public BlackboardItem shaderInputToCopy { get; set; }
+
+        public BlackboardItem copiedShaderInput { get; set; }
+
+        public int insertIndex { get; set; } = -1;
+    }
+
+    class BlackboardController : SGViewController<GraphData, BlackboardViewModel>
+    {
+        // Type changes (adds/removes of Types) only happen after a full assmebly reload so its safe to make this stuff static (I think, confirm this assumption)
```
marctem, line 135: This should be true until we add custom structs - then we'll need to be more dynamic about it.

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
```diff
@@ -1,21 +1,478 @@
...
+            foreach (var shaderInputType in m_shaderInputTypes)
+            {
+                if (shaderInputType.IsAbstract)
+                    continue;
+
+                var info = Attribute.GetCustomAttribute(shaderInputType, typeof(BlackboardInputInfo)) as BlackboardInputInfo;
+                string name = info?.name ?? ObjectNames.NicifyVariableName(shaderInputType.Name.Replace("ShaderProperty", ""));
+
+                // QUICK FIX TO DEAL WITH DEPRECATED COLOR PROPERTY
+                if (name.Equals("Color", StringComparison.InvariantCultureIgnoreCase) && ShaderGraphPreferences.allowDeprecatedBehaviors)
```
marctem, line 191: We want deprecated _and_ new versions in this case, right? Looks like this code will only add the deprecated version.

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
```diff
@@ -1,21 +1,478 @@
...
+                    }
+                }
+            }
+
+            return indexPerSection;
+        }
+
+        // A map from shaderInput reference names to the viewDataKey of the BlackboardPropertyView that is used to represent them
+        // This data is used to re-select the shaderInputs in the blackboard after an undo/redo is performed
+        Dictionary<string, string> oldSelectionPersistenceData { get; set; } = new Dictionary<string, string>();
```
marctem, line 457: Don't have to make this change now, but it seems like instead of storing "oldselectiondata" in a side channel like this, we could make it part of what we save/restore directly with the undo stack - would require decoupling undo data from save data, so not sure it's a good idea.

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardSectionController.cs
```diff
@@ -0,0 +1,197 @@
...
+using BlackboardItemController = UnityEditor.ShaderGraph.Drawing.ShaderInputViewController;
+
+namespace UnityEditor.ShaderGraph.Drawing
+{
+    class MoveShaderInputAction : IGraphDataAction
+    {
+        void MoveShaderInput(GraphData graphData)
+        {
+            Assert.IsNotNull(graphData, "GraphData is null while carrying out MoveShaderInputAction");
+            Assert.IsNotNull(ShaderInputReference, "ShaderInputReference is null while carrying out MoveShaderInputAction");
```
marctem, line 21: Not sure about the Asserts.

com.unity.shadergraph/Editor/Drawing/Controllers/SGController.cs
```diff
@@ -1,30 +1,44 @@
...
     struct SGControllerChangedEvent
     {
         public ISGControlledElement target;
         public SGController controller;
-        public int change;
+        public IGraphDataAction change;
 
-        bool m_PropagationStopped;
-        public void StopPropagation()
+        private bool m_PropagationStopped;
```
marctem, line 30: Don't need explicit private.

com.unity.shadergraph/Editor/Drawing/Controllers/ShaderInputViewController.cs
```diff
@@ -0,0 +1,304 @@
...
+        {
+            ViewModel.Model = Model;
+            ViewModel.IsSubGraph = DataStore.State.isSubGraph;
+            ViewModel.IsInputExposed = (DataStore.State.isSubGraph || (Model.isExposable && Model.generatePropertyBlock));
+            ViewModel.InputName = Model.displayName;
+            switch (Model)
+            {
+                case AbstractShaderProperty shaderProperty:
+                    ViewModel.InputTypeName = shaderProperty.GetPropertyTypeString();
+                    // HACK: Handles upgrade fix for deprecated old Color property
```
marctem, line 213: I don't like prefixing with "HACK" - either it's a TODO we'll fix soon, or just remove the word hack and let the code speak for itself.

marctem (changes requested):
This is only halfway done but about as far as I'll get today. Gives you a chance to address asap; I'll finish reviewing the rest tomorrow.

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/PropertyTests.cs
```diff
@@ -114,6 +117,12 @@ public IEnumerator ResetPropertyReference()
             }
         }
 
+        [Test]
+        public void AddPropertyTests()
```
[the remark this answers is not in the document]
Nightmask3: Yes working on the test automation code for this right now, along with addressing your review comments 👍

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ShaderGraphUITestHelpers.cs
```diff
@@ -102,26 +102,24 @@ public static void TestAllElements(VisualElement container, Action<VisualElement
             }
         }
 
-        public static void SendMouseEventToVisualElement(
+        public static void MouseDownEvent(
```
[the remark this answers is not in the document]
Nightmask3: Good catch, was bad naming, I have changed the function to be named `SendMouseEvent` now, for more general use of sending mouse events by the test code.

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ShaderGraphUITestHelpers.cs
```diff
@@ -102,26 +102,24 @@ public static void TestAllElements(VisualElement container, Action<VisualElement
...
-            screenButtonPosition += eventPositionOffset;
-            using var evt = ShaderGraphUITestHelpers.MakeMouseEvent(eventType, screenButtonPosition, mouseButton);
-            evt.target = elementToNotify;
-            elementToNotify.SendEvent(evt);
+            var screenButtonPosition = GetScreenPosition(parentWindow, elementToNotify);
+            parentWindow.SendEvent(new Event() {type = eventType, mousePosition = screenButtonPosition, clickCount =  clickCount, button = (int)mouseButton, modifiers = eventModifiers});
         }
 
-        public static Vector2 GetScreenPosition(VisualElement visualElement)
+        public static Vector2 GetScreenPosition(EditorWindow parentWindow, VisualElement visualElement)
```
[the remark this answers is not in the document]
Nightmask3: Patrick from the GTFO team should be credited for bringing it to my attention, but I am glad to not be relying on an internal function any more as well!

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/SubWindowTests.cs
```diff
@@ -56,10 +55,11 @@ public void CloseGraphWindow()
             m_Window.Close();
         }
 
-        private void ToggleSubWindows(bool showBlackboard, bool showPreview)
+        private void ToggleSubWindows(bool showBlackboard, bool showPreview, bool showInspector)
```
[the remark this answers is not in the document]
Nightmask3: This is a great idea, I was finding this unwieldy as I wrote it but wasn't certain how to improve it. I can definitely get that refactor in at a later date/more convenient time.

com.unity.shadergraph/Editor/Data/Graphs/CategoryData.cs
```diff
@@ -0,0 +1,55 @@
...
+        List<Guid> m_ChildItemIDList;
+
+        public List<Guid> childItemIDList
+        {
+            get => m_ChildItemIDList;
+            set => m_ChildItemIDList = value;
+        }
+
+        [SerializeField]
+        bool m_IsExpanded;
```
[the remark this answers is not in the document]
Nightmask3: Good catch, have removed this now

com.unity.shadergraph/Editor/Data/Graphs/DataStore.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public void Dispatch(ActionType action)
+        {
+            try
+            {
+                m_Reducer(State, action);
+            }
+            catch (Exception exception)
+            {
+                Debug.Log("Ran into exception of type:" + exception + " while dispatching action of type: " + action + "to DataStore.");
```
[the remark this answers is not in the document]
Nightmask3: A fair point regarding having actionable information for the user, this isn't something that you could really fix as a user without diving into the code (if it happened).

The idea was to handle exceptions somewhat gracefully from any action that can modify the application state/core data model, and if an exception is thrown provide a descriptive error message that at least provides more context/info for a bug report than just the exception itself. 🤔 

I'm searching for a different function that writes to the editor log as you describe it but can't find any explicit references, could you mention the specific log function that you were thinking I use instead?

com.unity.shadergraph/Editor/Drawing/Blackboard/SGBlackboard.cs
```diff
@@ -1,18 +1,95 @@
...
+
+        BlackboardViewModel ViewModel
+        {
+            get => m_ViewModel;
+            set => m_ViewModel = value;
+        }
+
+        // sections are created by BlackboardSectionControllers, then assigned to this Blackboard by the BlackboardController
+        internal SGBlackboardSection PropertySection { get; set; }
+        internal SGBlackboardSection KeywordSection { get; set; }
```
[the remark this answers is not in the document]
Nightmask3: Good catch, I realized these were redundant references and could be removed. And yes in general, it should be up to the controllers to glue the various views together as opposed to the views maintaining direct hard references to each other, I think.

com.unity.shadergraph/Editor/Drawing/Blackboard/SGBlackboard.cs
```diff
@@ -23,23 +100,52 @@ class SGBlackboard : GraphSubWindow, ISelection
...
-        public Action<SGBlackboard, VisualElement, string> editTextRequested { get; set; }
+        public Action addItemRequested { get; set; }
+        public Action<int, VisualElement> moveItemRequested { get; set; }
+
+        GenericMenu m_AddPropertyMenu;
 
-        public SGBlackboard(GraphView associatedGraphView) : base(associatedGraphView)
+        public SGBlackboard(BlackboardViewModel viewModel) : base(viewModel)
         {
+            ViewModel = viewModel;
```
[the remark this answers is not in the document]
Nightmask3: I'm not certain if its a pattern that's been formally described anywhere, but you've got the core idea there, where the view model classes themselves also have a hierarchy, and the base view model interface `ISGViewModel` is what the parent class of `SGBlackboard`, i.e. `GraphSubWindow` is maintaining a reference to

The blackboard subclass (and other views in general), will maintain references to the same view model instance, but at a lower derived class level that is specific to their view model needs, i.e. for the ``SGBlackboard`, we have the `BlackboardViewModel` reference and so on.

One could argue that the base class reference to the view model is pointless, as view models typically only have utility in their specific-ness towards a given view's purpose, but I find it useful because then we can do things like this:

```
protected GraphSubWindow(ISGViewModel viewModel)
{
    ViewModel = viewModel;
    m_ParentView = ViewModel.parentView;
    ParentView.Add(this);
    .....
}
```
Where setting up data and functionality for base window management stuff like binding references to a parent view and adding the view in question to the hierarchy etc, can all be handled by the `GraphSubWindow` base class, (which is the whole purpose for that class tbh).

com.unity.shadergraph/Editor/Drawing/Blackboard/SGBlackboardSection.cs
```diff
@@ -1,25 +1,65 @@
...
+
+        BlackboardSectionController m_Controller;
+
+        BlackboardSectionViewModel m_ViewModel;
+
+        BlackboardSectionViewModel ViewModel
+        {
+            get => m_ViewModel;
+            set => m_ViewModel = value;
+        }
```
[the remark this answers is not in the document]
Nightmask3: Good catch, removed this as it was redundant

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
Nightmask3: Very good catch, negligent of me to leave these in here, have been reading up on standards for assertions in production code, i.e. **DON'T DO IT**

I have wrapped them in an `#if SG_ASSERTIONS` check that we can use to compile those instructions out of production, but we can turn them on during our local dev-work in order to make sure that anything that we do doesn't regress the graph data manipulation.

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
Nightmask3: Yes good catch, for whatever reason I thought it was an either/or behavior desired here as opposed to, you can create the deprecated Colors AND create the new proper colors, when that flag is enabled. Will change it so both get added.

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
Nightmask3: I did a deep dive into how `GraphView` handles undo/redo for selection states and it's unfortunately not great 😢  

Relies on persistence of the viewDataKey which is basically a GUID attached to the property views
but given that we frequently recreate the graphData when-ever we do an undo/redo (which basically just replaces the graphData outright), we also end up recreating the propertyViews as a result, otherwise they point to stale `shaderInput` objects. 

This code here basically is a hack on top of that but as long as we're reliant on `GraphView` I don't really see a way around that, unless we're willing to rewrite our own undo/redo handling with a stack for selections and all that entails. 

It's not impossible, but I'd hope that GTFO has a better solution in mind so we don't have to do it 😅 . 

Would be worth bringing up at the next monthly sync infact. Will make a note.

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardSectionController.cs
Nightmask3: see above comment

com.unity.shadergraph/Editor/Drawing/ViewModels/BlackboardViewModel.cs
```diff
@@ -2,22 +2,44 @@
...
+            builtInKeywordNameToAddActionMap = new Dictionary<string, IGraphDataAction>();
+            categoryInfoList = new List<CategoryData>();
+            disabledKeywordNameList = new List<string>();
         }
 
-        public override void WriteToModel()
+        public void Reset()
         {
-            throw new System.NotImplementedException();
+            propertyNameToAddActionMap.Clear();
```
marctem, line 38: Do you need to clear the other members like parentView and the titles here as well?

com.unity.shadergraph/Editor/Drawing/ViewModels/InspectorViewModel.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace UnityEditor.ShaderGraph.Drawing
+{
+    class InspectorViewModel : ISGViewModel
+    {
+        public GraphData Model { get; set; }
+
+        public VisualElement parentView { get; set; }
+
+        public void Reset()
+        {
```
marctem, line 14: Nothing to clear here?

com.unity.shadergraph/Editor/Drawing/Views/GraphSubWindow.cs
```diff
@@ -42,24 +55,34 @@ class GraphSubWindow : GraphElement, ISGResizable
...
 
-        public List<ISelectable> selection => graphView?.selection;
+        public List<ISelectable> selection
+        {
+            get
+            {
+                if (ParentView is ISelectionProvider selectionProvider)
+                    return selectionProvider.GetSelection;
+
+                Debug.Log("ERROR: GraphSubWindow: Was unable to find a selection provider. Please check if parent view of: " + name + " implements ISelectionProvider::GetSelection");
```
marctem, line 82: Not user actionable, should probably be guarded assert.

marctem (changes requested):

com.unity.shadergraph/Editor/Drawing/Blackboard/SGBlackboard.cs
```diff
@@ -89,6 +89,7 @@ BlackboardViewModel ViewModel
         bool m_IsFieldBeingDragged = false;
 
         const int k_DraggedPropertyScrollSpeed = 6;
+        const float k_OriginalScrollerPadding = 12;
```
marctem, line 92: So now with the padding removed in the other PR, we don't need this any more, right?

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
```diff
@@ -1,21 +1,513 @@
...
+#if SG_ASSERTIONS
+            Assert.IsNotNull(graphData, "GraphData is null while carrying out AddShaderInputAction");
+#endif
+            // If type property is valid, create instance of that type
+            if (blackboardItemType != null && blackboardItemType.IsSubclassOf(typeof(BlackboardItem)))
+                shaderInputReference = (BlackboardItem)Activator.CreateInstance(blackboardItemType, true);
+            // If type is null a direct override object must have been provided or else we are in an error-state
+            else if (shaderInputReference == null)
+            {
+                Debug.Log("ERROR: BlackboardController: Unable to complete Add Shader Input action.");
```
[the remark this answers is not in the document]
Nightmask3: Have converted it to a guarded assertion now, and also added a new file called `AssertHelpers` with two simple Assert function wrappers (`AssertHelpers.IsNotNull` and `AssertHelpers.Fail`) in order to clean up the copies of the guarded if-defs everywhere and make the code a bit less verbose

com.unity.shadergraph/Editor/Drawing/Controllers/ShaderInputViewController.cs
```diff
@@ -0,0 +1,316 @@
...
+
+        protected override void RequestModelChange(IGraphDataAction changeAction)
+        {
+            DataStore.Dispatch(changeAction);
+        }
+
+        // Called by GraphDataStore.Subscribe after the model has been changed
+        protected override void ModelChanged(GraphData graphData, IGraphDataAction changeAction)
+        {
+            if (changeAction is ChangeExposedFlagAction changeExposedFlagAction)
```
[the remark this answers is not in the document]
Nightmask3: Will change this (and the similar function in `BlackboardController` and `BlackboardSectionController`) to use a switch statement

Regarding the `DirtyNodes()` thing, one feature (or) limitation (your mileage might vary) of how the controllers currently function is that every controller receives the notifications from the data store for every action that was dispatched. 

The reason why those `DirtyNodes()` calls are specified in all those cases is because there are many more types of `IGraphDataAction`s that do not require the `DirtyNodes()` calls and I am under the impression that dirty-ing all the nodes is a pretty expensive operation, especially with larger graphs, hence I am trying to be more specific about when it is called.

Thinking about how to improve this in the future, it should be pretty straightforward to add some data like an enum to the `GraphDataActions` which specify if its the kind of action that requires dirtying the nodes or not, which in general is something I want to do anyway, i.e. have some manner of scoping for actions so that we can be more specific about our responses and how we handle them.

com.unity.shadergraph/Editor/Drawing/ViewModels/InspectorViewModel.cs
Nightmask3: Will add the other members to be cleared here, but in truth this is mostly a stub class right now as the Inspector has no controller, it's just that as a `GraphSubWindow` it now requires a view model of some sort so this is a minimal class to satisfy that requirement.

com.unity.shadergraph/Editor/Drawing/ViewModels/BlackboardViewModel.cs
Nightmask3: Yes was missing some other members here, but regarding wiping the parentView, I have renamed this function to `ResetViewModelData` in the base interface to reflect its purpose better, which is that it should only be clearing out the data that the view directly uses in the presentation, as this function is called before a controller initializes the view model with the data it needs.

Also has to do with how the view models are currently fed the parentView reference only once, at their creation time, which might be something subject to debate, it made sense to me as we don't have dockable/undockable windows right now, but might in the future with overlays.

LandonTownsendUnity (approved):
Should be good now that the VFX graph compile errors are fixed

com.unity.shadergraph/Editor/Drawing/Blackboard/SGBlackboard.cs
Nightmask3: Oh good catch yes this shouldn't be necessary anymore

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -30,6 +33,25 @@ public MaterialGraphView()
...
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                var assemblyName = assembly.GetName().ToString();
+                if (assemblyName.Contains("GraphView"))
+                {
+                    graphViewAssembly = assembly;
+                }
+            }
+
+            Type graphViewType = graphViewAssembly?.GetType("UnityEditor.Experimental.GraphView.GraphView");
```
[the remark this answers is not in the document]
Nightmask3: Absolutely, this was a best of three solutions that I tried that was the least hacky in terms of the actual fix, but relies on functionality that GV decides inexplicably should be private/internal as you say.

Maintaining a running list of stuff like this to bring up with the GTFO team as we proceed 👍

com.unity.shadergraph/Editor/Drawing/Views/GraphSubWindow.cs
Nightmask3: Converted to a guarded assert call

Nightmask3:
Also ran the Universal test projects locally, results below: 

1) UniversalGraphicsTests_Foundation:
Play mode tests:

image

Edit mode tests:

image

2) UniversalGraphicsTests_Lighting:

Edit mode tests:

image

Play mode tests:

image

esmelusina (commented):
Haven't deep dived on it as much as I would like, but looks good. Clearing the lock.

tracychen1234 (approved):
Looks good to me (:

PR #3668, opened 2021-02-26
Fixed an issue in the planar reflection probe convolution.

anisunity:
There was an issue in clamping the borders while doing the downscaling which lead to black edges on the convoluted reflection probe. This may alterate the results of our tests, we need to run them.

image

Testing status
Tested it on couple resolutions and lighting conditions, the issues doesn't repro anymore.


sebastienlagarde (approved):

sebastienlagarde:
Please update screenshots , thanks

anisunity:
I updated and pushed the new screenshots

remi-chapelain (approved):
Discussed with @anisunity , no more testing needed on this one ! ✔️

PR #3673, opened 2021-02-26
[2020.2] Fix for shader variant error message

elizabeth-legros:
fixing error message that used to attach to the master node to print a debug log until we get a better way to handle graph-wide errors
Backport of https://github.com/Unity-Technologies/Graphics/pull/3664 
fix for https://fogbugz.unity3d.com/f/cases/1304168/
Tested repro of case before and after, could not repro


LandonTownsendUnity (approved):
Dev tested, and not risky enough to need more testing. Approving.

cdxntchou (approved):

PR #3677, opened 2021-02-27
Universal/light layers

kaychang-unity:

---
### Purpose of this PR
This adds support for light layers.
UniversalRenderPipelineAsset has a new checkbox to enable light layers. In this mode, the culling mask property of lights is switched to a light layer property (bitmask). A new shadow layer also appear for the light if it casts shadows.

---
### Testing status
- Need a test scene

---
### Comments to reviewers
- There was a bug where when static batching is enabled where `unity_RenderingLayer` are not always correctly set to different meshes.  This does not happens anymore with recent trunk revision.
- Should light layer names be renamable by users? They are currently hardcoded.
- Should we add the code related to decals?
- How to convert a light culling mask into a light layer?


kaychang-unity:
UniversalRenderPipelineAsset has new checkbox:
image

kaychang-unity:
When light layer is enabled, Light "Culling Mask" option is replaced by light layer option:
image

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
```diff
@@ -105,6 +105,7 @@ float4x4 unity_ObjectToWorld;
 float4x4 unity_WorldToObject;
 float4 unity_LODFade; // x is the fade value ranging within [0,1]. y is x quantized into 16 levels
 real4 unity_WorldTransformParams; // w is usually 1.0, or -1.0 for odd-negative scale transforms
+float4 unity_RenderingLayer;
```
pbbastian, line 108: Since this is a different block, we should add an empty line and a comment before it
```suggestion

// Render Layer block feature
float4 unity_RenderingLayer;
```

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added View Vector node to mimic old behavior of View Direction node in URP.
 - Added support for the PlayStation 5 platform.
 - Enabled deferred renderer in UI.
+- Light layers interface.
```
pbbastian, line 12:
```suggestion
- Added support for light layers, which uses Rendering Layer Masks to make Lights in your Scene only light up specific Meshes.
```

com.unity.render-pipelines.universal/Shaders/Lit.shader
```diff
@@ -129,6 +130,7 @@ Shader "Universal Render Pipeline/Lit"
             //--------------------------------------
             // GPU Instancing
             #pragma multi_compile_instancing
+            #pragma instancing_options renderinglayer
```
pbbastian, line 133: What's the impact of this if light layers are turned off?

pbbastian (changes requested):
@kaychang-unity Thank you for implementing this. I have some questions:
* Can you give a technical overview of how your solution works?
* Are there any risks, performance considerations (including build times), or dependencies involved?
* Are there any other technical areas this has an impact on? (e.g. decals)
* Which platforms is this targeting?
* Does this differ from the HDRP light layers in any way?
* Is documentation planned? @oleks-k
* Which rendering layers are getting exposed as light layers? Which aren't?
* Does this mean that some of the 4/8 lights available per object might be wasted due to non matching light layer masks?
* Should this include any analytics?
* Is this feature complete, or is there more that needs to be implemented later?
* Should this be on or off by default?

Regarding your questions:
> Should light layer names be renamable by users? They are currently hardcoded.
I think that should be possible, especially since it's possible in HDRP.
> Should we add the code related to decals?
@lukaschod 
> How to convert a light culling mask into a light layer?
Hmm, I'm not sure if this is possible. Maybe the HDRP people has some experience with this? @JulienIgnace-Unity  

Also, make sure to add the test scene mentioned before this lands :)

com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree7.shader
```diff
@@ -41,7 +41,9 @@ Shader "Universal Render Pipeline/Nature/SpeedTree7"
             #pragma multi_compile _ _ADDITIONAL_LIGHT_SHADOWS
             #pragma multi_compile _ _SHADOWS_SOFT
             #pragma multi_compile _ LOD_FADE_CROSSFADE
+            #pragma multi_compile_fragment _ _LIGHT_LAYERS
             #pragma multi_compile_fragment _ _SCREEN_SPACE_OCCLUSION
+            #pragma multi_compile_fragment _ _LIGHT_LAYERS
```
hkr, line 46: duplicate

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
hkr: Please document what's in the 4 components

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -7,6 +7,10 @@
 // Keep in sync with RenderingUtils.useStructuredBuffer
 #define USE_STRUCTURED_BUFFER_FOR_LIGHT_DATA 0
 
+#define RENDERING_LIGHT_LAYERS_MASK (255)
+#define RENDERING_LIGHT_LAYERS_MASK_SHIFT (0)
```
hkr, line 11: why do we have this shift of 0?

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -878,23 +899,32 @@ half4 UniversalFragmentPBR(InputData inputData, SurfaceData surfaceData)
...
         Light light = GetAdditionalLight(lightIndex, inputData.positionWS, shadowMask);
-        #if defined(_SCREEN_SPACE_OCCLUSION)
-            light.color *= aoFactor.directAmbientOcclusion;
-        #endif
-        color += LightingPhysicallyBased(brdfData, brdfDataClearCoat,
-                                         light,
-                                         inputData.normalWS, inputData.viewDirectionWS,
-                                         surfaceData.clearCoatMask, specularHighlightsOff);
+
+        if (IsMatchingLightLayer(light.lightLayers, meshRenderingLayers))
```
hkr, line 916: Does our compiler create a branch here if light layers are enabled?

hkr:
Is this feature supported for OpenGL ES 2.0?
The additional uniform array might cause problems. Also the integer/bit ops that are used.

sebastienlagarde:
>Should light layer names be renamable by users? They are currently hardcoded.

For this you will need a default settings asset. URP don't have one yet. It come with various workflows. It would be nice to be able to rename light layers, but I will suggest to do it in a separate PR, and then to evaluate if you want to introduce a default settings asset in 21.2 (it have consequence as you need to have it in template etc...).

>Should we add the code related to decals?
I don't see the relationship? Light layers and decal layers are differents thing if this is what you refer too, you shouldn't have to worry about decal in this PR.

>How to convert a light culling mask into a light layer?
I let you judge but it can be a very tricky problem. We haven't look at it for HDRP so we don't have experience about it, but clearly, if we have done dedicated light layers it was because of issue with light culling mask, so I don't expect you can get a 1-1 matching. Users can do what they want, with naming they want for culling mask.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
sebastienlagarde: it is because it's matching HDRP definition for readibility:

in hdrp we have:

#define RENDERING_LIGHT_LAYERS_MASK (255)
#define RENDERING_LIGHT_LAYERS_MASK_SHIFT (0)
#define RENDERING_DECAL_LAYERS_MASK (65280)
#define RENDERING_DECAL_LAYERS_MASK_SHIFT (8)

sebastienlagarde: we can have various stuff in rendering layer mask (in the future)

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
sebastienlagarde: Don't know for mobile or urp but in HDRP yes this code create a dynamic branch. On AMD GCN architecture it is a scalar branch so it is efficeint, but don't know for low level platform

com.unity.render-pipelines.universal/Shaders/Lit.shader
sebastienlagarde: This is enabled with instancing. 
renderinglayer will be sent with or without lightlayer enabled. So extra CPU cost when using instancing. Either the feature is used or not. (Note: In HDRP renderinglayer  is used also for decal layers, and in the future it can be used for other purpose).

kaychang-unity: I wasn't sure myself what it did but HDRP was enabling this so I assumed it was also needed by URP rendering layers. There is no option to disable this if the project does not use light layers, is there?
Furthermore, I don't think this is documented feature, as it is not documented in [2021.1 doc](https://docs.unity3d.com/2021.1/Documentation/Manual/GPUInstancing.html) in the valid switches for `#pragma instancing options`.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
kaychang-unity: I kept this in case more bits of the rendering layer mask are used for other feature (decal?).

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
kaychang-unity: This is verbatim copy of HDRP code&layout. I will add comment and change layout.

com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree7.shader
kaychang-unity: Indeed!

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
kaychang-unity: Updated.

kaychang-unity:
- Are there any risks, performance considerations (including build times), or dependencies involved?
This adds extra shader variant _LIGHT_LAYERS but the shader preprocessor strips them when light layers feature is enabled.
 
- Are there any other technical areas this has an impact on? (e.g. decals)
What's the status for URP decals? Is URP team is planning to implement decals using HDRP implementation?
Additionally, users wants to use rendering layer masks for their own purposes (but currently they can't), and this light layer implementation also consumes 8 bits of it.

- Which platforms is this targeting?
This is working for both Forward and Deferred renderers. So for Forward mode -> all platforms, for deferred mode -> only platform supporting deferred (NOT GL, GLES2, GLES3, DX10).

- Does this differ from the HDRP light layers in any way?
No.

- Is documentation planned?
Yes, @oleks-k

- Which rendering layers are getting exposed as light layers? The first 8?
Yes, same as HDRP

- Does this mean that some of the 4/8 lights available per object might be wasted due to non matching light layer masks?
Good point, in case there is few light count, bits in the rendering layer mask are wasted. Also not that implementation of light layers in forward renderer is inefficient and requires an extra dynamic branch, which does not happen when using the legacy culling mask. The issue has been shared with @phi-lira , it is expected.

- Should this include any analytics?
It could, sure. However, this is not expected to be performant.

- Is this feature complete, or is there more that needs to be implemented later?
I hope so. Unless HDRP team has more plans in progress regarding it.

Should this be on or off by default?
- Off by default. It is never efficient to use light layers in neither URP forward nor deferred. It is only implemented for feature parity between URP forward and URP deferred (URP forward has culling mask support, URP deferred has not), as well as compatibility with HDRP.

- Should light layer names be renamable by users? They are currently hardcoded.
I am asking because I could not find a way to rename the light layers in HDRP, unless you switch the HDRP render pipeline asset to "debug" view, in which case the light layers names appear and can be renamed but have a big "Obsolete" prefix in front of them.
Note that culling masks can be renamed but appear under "layers" in the UI (top left corner) which can be confusing (this is not the rendering layer mask).

I will work on adding a proper test scene.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
kaychang-unity: In next commit https://github.com/Unity-Technologies/Graphics/pull/3677/commits/bf099f9f22df2d385cb535f2b6d69df3ce1ec439, I adjusted the code so that IsMatchingLightLayer() is compiled out when light layers is not enabled. I confirmed the code is compiled out on Nintendo Switch since we can see actual assembly, I would expect the same for desktop and consoles. I'm not sure about the fate on mobiles.

kaychang-unity:
@hkr I confirm it does not work with GLES2, I will add some safety to not allow light layers on this platform.

kaychang-unity:
Following conversation with @TheoWong-pixel , I moved around the UI elements and renamed "Link Light Layers" into "Custom Shadow Layer"
image

kaychang-unity:
@hkr @rolandk-unity I added a warning when GLES2 is in the list of platforms in the player settings. As well, light layers will be implicitly disabled when GLES2 platform is detected at runtime.

image

jenniferd-unity (approved):

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
```diff
@@ -7,7 +7,7 @@
 
 #define PREFERRED_CBUFFER_SIZE (64 * 1024)
 #define SIZEOF_VEC4_TILEDATA 1 // uint4
-#define SIZEOF_VEC4_PUNCTUALLIGHTDATA 5 // 5 * float4
+#define SIZEOF_VEC4_PUNCTUALLIGHTDATA 6 // 6 * float4
```
phi-lira, line 10: Maybe this could be auto generated from C# as well.

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
```diff
@@ -45,6 +45,7 @@ struct PunctualLightData
     float3 spotDirection;       // spotLights support
     int flags;                  // Light flags (enum kLightFlags and LightFlag in C# code)
     float4 occlusionProbeInfo;
+    uint lightLayers;           // Optional light layer mask
```
phi-lira, line 48:
```suggestion
    uint layerMask;           // Optional light layer mask
```

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
```diff
@@ -73,6 +74,9 @@ Light UnityLightFromPunctualLightDataAndWorldSpacePosition(PunctualLightData pun
     {
         light.shadowAttenuation = AdditionalLightShadow(shadowLightIndex, positionWS, lightDirection, shadowMask, punctualLightData.occlusionProbeInfo);
     }
+
+    light.lightLayers = punctualLightData.lightLayers;
```
phi-lira, line 78:
```suggestion
    light.layerMask = punctualLightData.layerMask;
```

This is what I get confused when using lightLayer vs layer mask

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
phi-lira: @kaychang-unity so is the compiler smart to strip out code when using DEFAULT_LIGHT_LAYERS?

phi-lira (changes requested):

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
kaychang-unity: It is smart enough on Switch as the referenced variables (light.lightlayers, meshRenderingLayers) becomes values known at compile time. I think it will be the same on all platforms.

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -322,13 +325,17 @@ internal GraphicsFormat GetGBufferFormat(int index)
                 return GraphicsFormat.R32_SFloat;
             else if (index == GBufferShadowMask) // Optional: shadow mask is outputed in mixed lighting subtractive mode for non-static meshes only
                 return GraphicsFormat.R8G8B8A8_UNorm;
+            else if (index == GBufferRenderingLayers) // Optional: rendering layers sis outputed when light layers are enabled (subset of rendering layers)
+                return GraphicsFormat.R8G8B8A8_UNorm;
```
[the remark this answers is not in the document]
kaychang-unity: We can, R8G8B8A8_UNorm was the format used by HDRP (although they only use 16bits).

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
kaychang-unity: I remember HDRP has something for that, do we also have something similar in URP?

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -694,43 +701,51 @@ internal void SetupLights(ScriptableRenderContext context, ref RenderingData ren
             Profiler.EndSample();
         }
 
-        public void ResolveMixedLightingMode(ref RenderingData renderingData)
+        internal void ResolveMixedLightingMode(ref RenderingData renderingData)
```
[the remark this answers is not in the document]
kaychang-unity: DeferredLights is internal class that is furthermore still unused in released packages for URP, so I assume it should be fine?

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -694,43 +701,51 @@ internal void SetupLights(ScriptableRenderContext context, ref RenderingData ren
...
+            if (this.GbufferAttachments == null || this.GbufferAttachments.Length != gbufferSliceCount)
+            {
+                this.GbufferAttachments = new RenderTargetHandle[gbufferSliceCount];
+                for (int i = 0; i < gbufferSliceCount; ++i)
+                    this.GbufferAttachments[i].Init(k_GBufferNames[i]);
+            }
         }
 
-        public bool IsRuntimeSupportedThisFrame()
+        internal bool IsRuntimeSupportedThisFrame()
```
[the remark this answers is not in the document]
kaychang-unity: DeferredLights is internal class that is furthermore still unused in released packages for URP, so I assume it should be fine?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -713,6 +730,11 @@ public bool supportsMixedLighting
             get { return m_MixedLightingSupported; }
         }
 
+        public bool supportsLightLayers
```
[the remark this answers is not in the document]
kaychang-unity: Added comment.

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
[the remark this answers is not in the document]
kaychang-unity: I moved to R8_Unorm.

thomas-zeng (approved):

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
kaychang-unity: Ok, I renamed (a lot of code) to layerMask.

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
kaychang-unity: Actually found the tool. The "size" of the structure in float4 units is needed when manually trying to read back structure from an uniform buffer (no officially supported in Unity). But the tool will not generate actual "size" of the structure, so I am leaving it as it is for now.

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
kaychang-unity: Renamed.

hkr (approved):

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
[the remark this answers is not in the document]
phi-lira: Ah that's fine then.

com.unity.render-pipelines.universal/Shaders/Utils/Deferred.hlsl
phi-lira: Ok. Not blocking issue.

phi-lira (approved):

JulienIgnace-Unity (approved):

Jonasmortensen (approved):
Overall looks good to me.
Only a couple of thoughts:
- Light Layer property can be removed from light inspector when light is baked
- Interaction with Shadowmask can give unexpected results. I'm not sure if it is possible for the mask to take light/shadow layers into account but would be nice.
- About renaming light layers: The layer names should be hardcoded as long as URP doesn't have Global Settings

ernestasKupciunas (approved):
It seems to work as expected. I was able to play around with the Light Layers in the editor, set different objects to be affected by particular light layers. Rendering on mobile devices looks good too.
Tested using [Boat attack](https://github.com/Unity-Technologies/BoatAttack) and [Winter](https://github.com/ernestasKupciunas/Winter/tree/Winter_Demo) projects.
Winter:
ResultWinterLightLayersDeferred

BoatAttack:
Boat


Unity:
Version: 2021.2.0a14.1944
Revision: trunk 9f66320aebca
Built: Fri, 09 Apr 2021 17:39:30 GMT

Devices under testing:

simon-engelbrecht-soerensen (approved):
Working as expected!

pbbastian (approved):

pbbastian (approved):

kroatoa:
I'm assuming this relates to https://unity3d.atlassian.net/browse/UR-1297. That roadmap item is missing the Product manager approval, so if you could poke the relevant one to sign off, that would be good. Otherwise we risk adding the delay for that approval to the weekly bundling.

Other than that I have no complaints.

phi-lira:
URP Update top assets is known issue already fixed on master. It's ok to merge with this one red.
 HDRP tests, this PR doesn't touch any HDRP file so ignoring.

phi-lira:
@kaychang-unity wait for test to finish before resolving conflict. Once ABV is green, resolve conflict and don't run test again. looking at conflict file we dont' need to rerun graphics test due to that.

PR #3681, opened 2021-02-28
Hd/alpha to mask frame setting fix 1315452

victorsclui:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1317950/

Cleaned up a little how the overrideable delegates are handled for frame setting UI. This enabled the FrameSettingsField.AlphaToMask field to have positive dependency on MSAA frame setting and negative dependency on LitShaderMode frame setting at the same time. Also added overrideable delegate for FrameSettingsField.AlphaToMask to better take HDRP Asset restriction into account.

---
### Testing status

Checked FrameSettingsField.AlphaToMask is properly indented and no longer overrideable when LitShaderMode is in Deferred mode.


TomasKiniulis (approved):
Looks good!
Checked all Lit Shader Modes, can confirm that in deferred mode alpha to mask is not overrideable, alpha to mask is always indented.
Checked HDRP Default Settings and Camera frame settings - Alpha To Mask works correctly in Forward mode

https://user-images.githubusercontent.com/50582134/109473783-7016ae00-7a7c-11eb-8074-6acaa121598c.mp4
https://user-images.githubusercontent.com/50582134/109473882-90466d00-7a7c-11eb-9395-383cc309a352.mp4

sebastienlagarde (approved):

RSlysz (approved):
Looks good.

The dependency mechanism was a way to quickly manage dependencies but as things got more complex, perhaps we should try to see if there is a better way to handle them in the future.

victorsclui:
@RSlysz yea currently many custom overideable delegate is needed because we need to take into account the override frame setting, the default frame setting, and the HDRP asset all together. It is conceivable to refactor so there is a dependency mechanism that handles all this so that custom overrideable delegate is needed less..

PR #3652, opened 2021-02-25
[Yamato] Add expression trigger to run package jobs on all PRs

sophiaaar:
### Purpose of this PR
Add expression trigger to run package jobs on all PRs so that we don't inadventantly break the package validation

---
### Testing status
Tested that the jobs start when the PR is un-drafted

---
### Comments to reviewers

PR #3682, opened 2021-02-28
quick regression fix

esmelusina:
### Purpose of this PR
PR #3625 introduced a regression to custom interpolator previews (they landed very close to each other).

---
### Testing status
There is no testing framework for preview shaders. Manually, we can see that for all cases of preview shaders involving custom interpolators (downstream and master) the shaders were failing to compile, and that this fix restores that.

### DIFF
--- a/com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
+++ b/com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
@@ -295,13 +295,18 @@ namespace UnityEditor.ShaderGraph.Drawing
                 }
             }
 
-            // Custom Interpolator Blocks have implied connections to their Custom Interpolator Nodes
+            // Custom Interpolator Blocks have implied connections to their Custom Interpolator Nodes...
             if (dir == PropagationDirection.Downstream && node is BlockNode bnode && bnode.isCustomBlock)
             {
                 foreach (var cin in CustomInterpolatorUtils.GetCustomBlockNodeDependents(bnode))
                 {
                     action(cin);
                 }
+            }
+            // ... Just as custom Interpolator Nodes have implied connections to their custom interpolator blocks
+            if (dir == PropagationDirection.Upstream && node is CustomInterpolatorNode ciNode)
+            {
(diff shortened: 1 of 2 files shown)


com.unity.shadergraph/Editor/Generation/Processors/CustomInterpolatorUtils.cs
```diff
@@ -122,8 +122,6 @@ internal void ProcessExistingStackData(List<AbstractMaterialNode> vertexNodes, L
                         // sorted insertion, based on dependencies already present in pixelNodes (an issue because we're faking for the preview).
                         InsertAntecedent(pixelNodes, ant);
                     }
-                    // all antecedents of cin are vertex blocks, but we can add them to Pixel to try and get reasonable preview generation.
-                    pixelNodes.InsertRange(0, anties);
```
esmelusina, line 126: This was replaced by the lines above, but I forgot to remove it.

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
```diff
@@ -295,14 +295,19 @@ static void ForeachConnectedNode(AbstractMaterialNode node, PropagationDirection
...
                 foreach (var cin in CustomInterpolatorUtils.GetCustomBlockNodeDependents(bnode))
                 {
                     action(cin);
                 }
             }
+            // ... Just as custom Interpolator Nodes have implied connections to their custom interpolator blocks
+            if (dir == PropagationDirection.Upstream && node is CustomInterpolatorNode ciNode)
+            {
+                action(ciNode.e_targetBlockNode);
+            }
```
esmelusina, line 310: #3652 solves SRP batching by gathering properties ahead of time. This case wasn't necessary for the previous execution flow, but was leaving it out was still an oversight on my part.

LandonTownsendUnity (approved):
I couldn't get the original bug to repro but it looks like Custom Interpolators still work fine on this branch, so I'll approve.

cdxntchou (approved):

PR #3689, opened 2021-03-01
[HDRP] Fix issue with physically-based DoF and transparent materials with depth writes on

pmavridis:
### Purpose of this PR
The camera depth texture (HDShaderIDs._CameraDepthTexture) was not explicitly bound by the physical DoF pass, and it was using the depth mip-chain texture (set from another pass). 

This was working fine for most cases, but the depth mip-chain is missing the transparent depth-writes, so these were missing from the physical DoF computation! The non-physical code path is already binding the correct depth texture.

---
### Testing status
I have tested with the Walkman scene

Before (note that the Walkman body is set to transparent with depth writes on) :
image

After (same setup as above):
image

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/2x_Lighting/2011_ForwardEmissiveForDeferred/UnlitSGEmissive.shadergraph
+++ b/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/2x_Lighting/2011_ForwardEmissiveForDeferred/UnlitSGEmissive.shadergraph
@@ -1,5 +1,5 @@
 {
-    "m_SGVersion": 2,
+    "m_SGVersion": 3,
     "m_Type": "UnityEditor.ShaderGraph.GraphData",
     "m_ObjectId": "1441cc241b3c4d72afec9496d6934b6c",
     "m_Properties": [],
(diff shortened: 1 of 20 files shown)


FrancescoC-unity (approved):

sebastienlagarde (approved):

RemyUnity:
Good description, and LGTM. Just one thing : Do you think that we should consider adding a test scene for this case (transparent depth prepass with physically-based DoF) ?
A solution could be to duplicate tests 1802, 1803 & 1804 and switch to physically based DoF on the new ones.

pmavridis:
> A solution could be to duplicate tests 1802, 1803 & 1804 and switch to physically based DoF on the new ones.

Yes, I agree. But I'm also doing some additional fixes for the near blur, so I think it makes sense to prepare the new test (with some transparent materials too) with the PR for the near blur, so we don't have to update the ref images multiple times.

RemyUnity:
> > A solution could be to duplicate tests 1802, 1803 & 1804 and switch to physically based DoF on the new ones.
> 
> Yes, I agree. But I'm also doing some additional fixes for the near blur, so I think it makes sense to prepare the new test (with some transparent materials too) with the PR for the near blur, so we don't have to update the ref images multiple times.

Ok, good for me !

RemyUnity (approved):

PR #3691, opened 2021-03-01
Added support for screen space shadows (directional and point, no area) for shadow matte unlit shader graph.

anisunity:
Added support for screen space shadows (direction and point) for the unlit shader graph shadow matte.

Testing status
Added two new tests to cover this use case.
This shoudn't affect anything outside of these tests.


sebastienlagarde (approved):
Please add the feature in the what's new for 12.0 page. Thanks

sebastienlagarde:
Formatting is incorrect

remi-chapelain (commented):
Expected that Unlit shadow matte (on the left) only get semi transparent ray traced shadow but don't get color like the lit on the right ? 

image

anisunity:
Color shadow fixed and formatting fixed

anisunity:
My shadow change broke a test

remi-chapelain (approved):
Thanks for adding the colored shadow setup to the 712 test ✔️

sebastienlagarde:
Ping on this PR: missing update of What's new and there is one test failing (the added one)

anisunity:
The added test is not the one failing. It is an other one that fails. I pushed a fix that should make everything green (does locally at least)

PR #3693, opened 2021-03-01
Fix SSGI frame setting not greyed out while SSGI is disabled in HDRP Asset

victorsclui:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1317954/

Very simple fix, just add some UI handling code for SSGI frame setting the same way we already have UI handling code for SSAO frame setting.

---
### Testing status
Checked SSGI frame setting is now properly greyed out if current HDRPAsset disables SSGI.


sebastienlagarde (approved):

iM0ve (approved):
Looks good

PR #3694, opened 2021-03-01
Place the scene in the correct folder to avoid template loading issues

pierre-unity:
### Purpose of this PR

**Backport to 10.x.x/release branch**

The template scene must be specifically in Assets/Scenes/SampleScene.unity, otherwise, it is unable to load automatically and an error will be fired: 

> Cannot open scene at path Assets/Scenes/SampleScene.unity. Maybe the scene hasn't been imported. This can be done with an AssetDatabase.Refresh() call.

I have therefore placed the "Scenes" folder back at the top-level inside the asset folder:

image

All the other files are in "Assets\SampleSceneAssets", so that they can be deleted very quickly.

---
### Comments to reviewers
The severity of the change is very low, however, because many asset files are moved, it looks like a scary change.

Depency: I'll notify the tutorial team that they need to update the text regarding the deletion of the template, now that the folder names are different.

In addition, the lightmaps in the first room might still look broken if you test with 2021.2.0a6 and beyond, due to this issue: https://fogbugz.unity3d.com/f/cases/edit/1317192/


sebastienlagarde:
validated by QA here:
https://github.com/Unity-Technologies/Graphics/pull/3692

PR #3695, opened 2021-03-01
Hd/fix access current hdrp asset for default frame settings 1317968

victorsclui:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1317968/

Until we stored HDRP default setting in a separate scriptableObject, Default frame setting lives only inside the default HDRPAsset. So GetDefaultFrameSettingsFor should be calling HDRenderPipelineAsset hdrpAsset = HDRenderPipeline.defaultAsset; instead of HDRenderPipelineAsset hdrpAsset = GetHDRPAssetFor(owner);.

---
### Testing status

Checked now camera Override Frame Setting is reading from the default Frame Setting from the default HDRP Asset instead of the current one now.


sebastienlagarde (approved):

sebastienlagarde:
@QA: merging to move it in the pacakge, small change. Added QA for awareness

PR #3697, opened 2021-03-01
[10.x.x] [universal] [fix] hdr toggle for sceneview

martint-unity:
---
### Purpose of this PR

#2508

backport of this PR #3716

---
### Testing status
Yamato

---
### Comments to reviewers


ellioman (approved):

phi-lira (approved):

erikabar (approved):
Verified https://fogbugz.unity3d.com/f/cases/edit/1313942/
Checked if existing functionality with HDR and scene view works

PR #3699, opened 2021-03-01
Fixed HDRP material being constantly dirty

adrien-de-tocqueville:
### Purpose of this PR

This PR fixes issues with material saving in HDRP.
Previously, opening a material inspector made the material dirty without modifying it, so saving the project would save all materials opened since last save, essentially for nothing

There was multiple reasons:
- Unlit shader used emission texture and emissive color for GI, instead of base color and base texture. Since lit shader also has properties with the same name for emissive, it caused double modifications. So changing the properties for GI in unlit fixes two things at the same time
- the doubleSidedGI setter of material always calls setdirty, even if the new doubleSidedGI value is the same as the old one
- There was some properties for distortion in lit shader that made some passes be disabled and then reenabled. But distortion isn't supported by the lit shader.. So just cleaning up code actually fix the thing

---
### Testing status
Added a Debug.Log at the end of resetkeyword callback to print the dirtyness of the material
Verified that after saving a material, the material isn't dirty and saving again doesn't do anything


com.unity.render-pipelines.high-definition/Editor/Material/Unlit/BaseUnlitGUI.cs
```diff
@@ -235,9 +235,7 @@ public static void SetupBaseUnlitKeywords(this Material material)
                 MaterialEditor.FixupEmissiveFlag(material);
             }
 
-            // Commented out for now because unfortunately we used the hard coded property names used by the GI system for our own parameters
```
adrien-de-tocqueville, line 238: this isn't the case since a long time

com.unity.render-pipelines.high-definition/Editor/Material/Unlit/BaseUnlitGUI.cs
```diff
@@ -283,7 +284,13 @@ public static void SetupMainTexForAlphaTestGI(this Material material, string col
 
         static public void SetupBaseUnlitPass(this Material material)
         {
-            if (material.HasProperty(kDistortionEnable))
+            if (material.shader.IsShaderGraph())
```
adrien-de-tocqueville, line 287: put that if before in case someone creates a shadergaph with a property names _DIstortionEnable

com.unity.render-pipelines.high-definition/Editor/Material/Unlit/BaseUnlitGUI.cs
```diff
@@ -249,12 +247,15 @@ public static void SetupBaseUnlitKeywords(this Material material)
...
-                material.doubleSidedGI = doubleSidedEnable;
+                doubleSidedGI = doubleSidedEnable;
             else if (doubleSidedGIMode == DoubleSidedGIMode.On)
-                material.doubleSidedGI = true;
+                doubleSidedGI = true;
             else if (doubleSidedGIMode == DoubleSidedGIMode.Off)
-                material.doubleSidedGI = false;
+                doubleSidedGI = false;
+            if (doubleSidedGI != material.doubleSidedGI)
+                material.doubleSidedGI = doubleSidedGI;
```
adrien-de-tocqueville, line 258: a c++ PR to fix SetDoubleSidedGI would be a good idea I suppose

sebastienlagarde: For now it is fine, just add a comment here to explain that calling material.doubleSidedGI always dirty the amterial

sebastienlagarde (approved):

RemyUnity (changes requested):
Hey, this PR is not only about the materials beeing dirty, but it also removes the distortion settings and passes from AxF, Lit and LitTessellation shader.

Either change the title+description or remove those changes to match them.

adrien-de-tocqueville:
> Hey, this PR is not only about the materials beeing dirty, but it also removes the distortion settings and passes from AxF, Lit and LitTessellation shader.
> 
> Either change the title+description or remove those changes to match them

Hey, as mentionned in the description, distortion isn't supported by the lit(tessellation) and AxF shader.
The pass was not removed when distortion was removed from lit shader, but it is not actually possible to setup distortion on a lit shader from the UI

RemyUnity (approved):

PR #3700, opened 2021-03-01
Making the FrameSettingsAttribute field more robust. Using it in othe…

kecho:
### Purpose of this PR
Fix a crash in IL2CPP introduced by a previous PR.  Doing a safe check on frame settings name attributes. Ensuring that other systems are using the cached attribute name list.

---
### Testing status
Currently testing ps4 player il2cpp

---
### Comments to reviewers
Crash occurs because il2cpp treats enums slightly differently, not as ints. Currently testing the exact test that failed to ensure this %100 fixes the issue.


victorsclui (approved):
lgtm : )

sebastienlagarde (approved):

PR #3705, opened 2021-03-01
[2021.2] [ShaderGraph] URP Uber Shader

cdxntchou:
### Purpose of this PR
This PR adds an "Allow Material Override" checkbox to Lit and Unlit subtargets, as well as controls for Cast and Receive Shadows, Depth Write and Depth Test.
imageimage

By default, "Allow Material Override" will not be checked, both to preserve the behavior of existing shadergraphs, and to simplify the initial user experience.  When checked, it allows the Material to control the following surface options (that were previously only controlled from the ShaderGraph):

For Lit subtargets:
* Workflow Mode
* Surface Type
* Render Face
* Depth Write
* Depth Test
* Alpha Clipping
* Cast Shadows
* Receive Shadows

For Unlit subtargets:
* Surface Type
* Render Face
* Depth Write
* Depth Test
* Alpha Clipping
* Cast Shadows

Example Unlit Material, when"Allow Material Override" is checked on the ShaderGraph:
image

It achieves this by modifying the subtargets to generate properties and keywords to control the surface options, in combination with a setup function that applies state updates to the Materials to control these properties and keywords: ShaderUtils.UpdateMaterial(material).  To be able to run the correct update on each material, we created a ShaderID enum to identify the various Universal shader types, extending the existing ShaderPathID to include shadergraph generated shaders.  We also attach metadata to Universal ShaderGraph assets, so we can tag them with their ShaderID.  The helper function ShaderUtils.GetShaderID() is provided to get the ShaderID for any shader, either hand-written or shadergraph generated.

This PR also creates two new Material Inspectors: ShaderGraphLitGUI (for Lit subtarget), and ShaderGraphUnlitGUI (For Unlit subtarget) to handle displaying the surface controls appropriately, and calling the update function when needed.

These inspectors are based off of the BaseShaderGUI base class, so some changes are made to BaseShaderGUI so that it works with ShaderGraph shaders in addition to the existing hand-written shaders.

This PR also makes a few associated changes to ShaderGraph:
1) Adds "ShaderGraphShader"="True" tag to all ShaderGraph-built subshaders, allowing us to easily tell when a Material is a ShaderGraph (even if the ShaderGraph is not asset based).  Added "Material.IsShaderGraph" utility function, and changed the old "Shader.IsShaderGraph" function to "Shader.IsShaderGraphAsset" to make it clear it only works on Assets.

2) Adds "Stages" dropdown to Keywords, allowing user to set up "Fragment" or "Vertex" only keywords to reduce compiled variants.
image

3) Adds linearGrey and red options as default texture mode, and changed the "bump" option to be called "Normal Map"
image

4)  Texture2D properties using Normal Map (old "bump") default will now be marked with [NormalMap] tag, so that users are reminded to assign a normal map usage texture to that property.

5) Adds CastsShadows option (for Materials that use Lit/Unlit SubTargets)
6) Adds Depth Test control (for Materials that use Lit/Unlit SubTargets)
7) Made the SubTarget and Material controls look identical (same naming, same options available)
image
image

---
### Testing status

Tested:
- [x] Tested that assigning a new Shader to a Material will reset keywords properly
- [x] Tested that newly created Materials have their keywords set up properly even if the inspector is not visible or run.
- [x] Test bump maps / normal maps work with the new subtargets
- [x] Tested combinations of different Target settings, to see that they modify the default properties in the generated shader, and will populate newly created Materials with those defaults.
- [x] Tested all cull mode settings (front/back/both) on Material
- [x] Tested setting Opaque/Transparent, and all blend modes on Material
- [x] Tested enabling/disabling alpha clipping on Material
- [x] Tested enabling/disabling receive shadows on Material
- [x] Tested enabling/disabling cast shadows on Material
- [x] Tested Depth Write and Depth Test settings on Material
- [x] Verified that keyword options all generate the correct #pragma in HLSL
- [x] Tested that exposed keyword options are controllable in the new Material Inspector
- [x] Tested that keyword stages option UI works with undo/redo, and is serialized/deserialized properly
- [x] Verified that existing KeywordDescriptors from Targets are generated in code correctly
- [x] Tested pre-existing URP graphs upgrade to an equivalent behavior with this PR -- tested alpha clipping, workflow mode, surface type, render face
- [x] Validate that old (pre-stack) graphs upgrade two sided setting correctly from PBRMaster and UnlitMaster
- [x] Verified that shadergraph master previews show up correctly with URP Lit/Unlit subtarget, and selecting various combinations of default surface settings on them
- [x] Test we didn't break URP sprite subtargets
- [x] Test that material inspectors for old Lit.shader, Unlit.shader, and any other variants still work properly.
- [x] Tested that pre-existing Universal Lit ShaderGraph Materials that are using Baked GI continue to work after upgrading to this PR.  The GI flag is preserved and the baked results are identical (both with GI disabled and GI enabled):
image
- [x] Tested adding keywords, modifying them from the Material, check they are preserved
- [x] Tested use of Clear Coat -- upgrades correctly, Target only control, renders properly
- [x] Previews work in both Material Override and non override mode, identical results
- [x] Tested that modifying ShaderGraph updates the keywords on the Materials using that ShaderGraph
- [x] Tested that Material versioning can process ShaderGraph Materials now
- [x] Tested that modifying Materials via inspector will update keywords
- [x] Tested that Target and Material Inspector surface controls look and behave consistently
- [x] Tested that ShaderGraph blocks (Alpha, Alpha Threshold, Specular, Metallic) are added or removed correctly based on the active options in the Target.
- [x] Tested that shadow culling matches the setting of alpha clip, render face settings.
- [x] Tested with both URP and HDRP Targets active

Yamato:

ShaderGraph PR Job: 🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Furp-uber/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/6559408/job/pipeline

Universal PR Job:  matches equivalent master results 🟢
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Furp-uber/.yamato%252Fall-universal_split.yml%2523PR_Universal_Split_trunk/6560960/job/pipeline

failing jobs:
URP_PostPro on Win_DX11_Standalone_XR_cache_mono_Linear on version trunk -- same 24/55 tests failed.
URP_PostPro on Win_DX11_playmode_XR_cache_mono_Linear on version trunk -- same 24/55 tests failed.
URP_PostPro on Win_DX12_Standalone_XR_cache_mono_Linear on version trunk -- same 24/55 tests failed.
URP_PostPro on OSX_Metal_playmode_cache_mono_Linear on version trunk -- success on re-run
URP_Lighting on Win_DX11_Standalone_XR_cache_mono_Linear on version trunk -- same 28/75 tests failed.
URP_Lighting on Win_DX11_playmode_XR_cache_mono_Linear on version trunk -- same 28/94 tests failed.
URP_Lighting on Win_DX12_Standalone_XR_cache_mono_Linear on version trunk -- same 28/75 tests failed.
URP_Foundation on Win_DX11_Standalone_cache_mono_Linear on version trunk -- success on re-run
URP_Foundation on Win_DX11_Standalone_XR_cache_mono_Linear on version trunk -- same 60/94 tests failed.
URP_Foundation on Win_DX11_playmode_XR_cache_mono_Linear on version trunk -- same 60/113 tests failed.
URP_Foundation on Win_DX12_Standalone_XR_cache_mono_Linear on version trunk -- same 60/94 tests failed.
URP_Foundation on OSX_Metal_playmode_cache_mono_Linear on version trunk -- success on re-run
URP_Terrain on Win_DX11_playmode_XR_cache_mono_Linear on version trunk -- same 8/68 tests failed.
URP_2D on Win_DX11_playmode_XR_cache_mono_Linear on version trunk -- same 18/74 tests failed.
URP_2D on OSX_OpenGLCore_editmode_mono_Linear on version trunk -- same 1/63 tests failed.
URP_2D on Win_DX11_editmode_mono_Linear on version trunk -- same 1/63 tests failed.

equivalent master version: ed962901fcba6e7b7ea7ac14b615d8f62bacfe03

master failing job;
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Fall-universal_split.yml%2523Nightly_Universal_Split_trunk/6518789/job/pipeline

HDRP PR Job: 🟢
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Furp-uber/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/6433493/job/pipeline

latest after merge:  lots of failures, but matches trunk  🟢
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Furp-uber/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/6567033/job/pipeline

rerunning:
Build HDRP on Win_DX11_mono_Linear_Standalone_cache_build_Player on version trunk 
HDRP on Win_DX11_editmode_mono_Linear on version trunk

---
### Comments to reviewers

Discovered Issues for emulating Lit.Shader / Unlit.Shader,
These won't necessarily be fixed in this PR, I'm just noting the issue for future work on the Lit.shadergraph:
- [x] no linearGrey default color option (this PR)
- [x] no MainTexture / MainColor tags (separate PR)
- [ ] no ToggleOff tag
- [x] no Normal tag (this PR)
- [ ] no HiddenInInspector tag exposed on blackboard properties (only internal in code) -- can hack serialized file
- [x] no Texture Scale and Offset support (separate PR)
- [x] keyword "stage" not exposed on blackboard keywords -- added drop down (this PR)
- [ ] ShaderGraph Unlit subtarget casts shadows by default, Unlit.Shader does NOT, will need to make sure to flag Unlit.ShaderGraph as non-shadowcasting by default to maintain behavior.
- [ ] ShaderGraph generated meta passes (for Global Illumination) do not do the same diffuse/specular color blend that Lit.shader does

Issues Found, that are Not Fixed / by design for now:
- [ ] Transparent Multiply mode ignores alpha (pre-existing issue -- may be intended?)
- [x] Transparent Premultiply mode -- specular ignores alpha (pre-existing issue -- this is intended)
- [ ] culling option currently also controls culling in shadow passes.  May want independent shadow pass culling control like HDRP?
- [x] DepthOnly needs to be stripped when ZWrite is force disabled
- [ ] should we strip DepthNormalOnly passes when ZWrite is force disabled?
- [ ] Baked GI for alpha-clipped shadergraphs is not correct -- will ignore the alpha clipping.  Seems to be a limitation of the meta-pass approach to gathering GI values?
- [ ] meta pass GI also doesn't seem to take into account differing values for the front and back sides of the object -- even when double-sided GI is enabled on the Material.
- [ ] Switching URP ==> HDRP does not currently update material keywords (until you open the material inspector)
- [ ] Swtiching HDRP ==> URP does not currently update material keywords (until you open the material inspector)


TestProjects/ShaderGraph/Assets/CommonAssets/Editor/KeywordTests.cs
```diff
@@ -195,11 +195,7 @@ public void CanGetKeywordDeclarationForReals()
             var sb =  new ShaderStringBuilder();
             foreach (var keyword in m_Collector.keywords)
             {
-                string declaration = keyword.GetKeywordDeclarationString();
-                if(!string.IsNullOrEmpty(declaration))
-                {
-                    sb.AppendLine(declaration);
-                }
+                keyword.AppendKeywordDeclarationStrings(sb);
```
cdxntchou, line 198: This more optimal form just appends the keyword declaration string directly into the builder...  no temp string GC allocs

TestProjects/ShaderGraph/SaveLoadGraphs/SimpleWithSubGraph.shadergraph
```diff
@@ -1124,7 +1124,7 @@
     },
     "m_Name": "A",
     "m_DefaultReferenceName": "Vector1_e29cd19dafd54a1dbd31dcf1d75c21b2",
-    "m_OverrideReferenceName": "a_3_4_d__",
```
cdxntchou, line 1127: Fixing an issue with this test on GL devices -- not allowed to use double underscore in identifiers.

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
```diff
@@ -674,19 +674,6 @@ void ProcessCompletedShaderCompilations()
                     CheckForErrors(renderData.shaderData);
 
                     previewsCompiled.Add(renderData);
-
-                    if (renderData == m_MasterRenderData)
-                    {
-                        // TODO: this may be a good thing to do BEFORE requesting shader compilation
```
cdxntchou, line 680: Applied this TODO -- it is now done before we start requesting shader compiles

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
```diff
@@ -1168,6 +1175,21 @@ void BuildKeywordFields(PropertySheet propertySheet, ShaderInput shaderInput)
                     out keywordScopeField));
             }
 
+            {
```
cdxntchou, line 1178: New UI for the keyword stage controls

com.unity.shadergraph/Editor/Data/Graphs/Texture2DShaderProperty.cs
```diff
@@ -10,7 +10,25 @@ namespace UnityEditor.ShaderGraph.Internal
     [BlackboardInputInfo(50)]
     public sealed class Texture2DShaderProperty : AbstractShaderProperty<SerializableTexture>
     {
-        public enum DefaultType { White, Black, Grey, Bump }
+        public enum DefaultType { White, Black, Grey, NormalMap, LinearGrey, Red }
```
cdxntchou, line 13: Added LinearGrey and Red options, changed the name of "Bump" to NormalMap

com.unity.shadergraph/Editor/Data/Graphs/Texture2DShaderProperty.cs
```diff
@@ -27,7 +45,8 @@ internal Texture2DShaderProperty()
 
         internal override string GetPropertyBlockString()
         {
-            return $"{hideTagString}{modifiableTagString}[NoScaleOffset]{referenceName}(\"{displayName}\", 2D) = \"{defaultType.ToString().ToLower()}\" {{}}";
+            var normalTagString = (defaultType == DefaultType.NormalMap) ? "[Normal]" : "";
+            return $"{hideTagString}{modifiableTagString}{normalTagString}[NoScaleOffset]{referenceName}(\"{displayName}\", 2D) = \"{ToShaderLabString(defaultType)}\" {{}}";
```
cdxntchou, line 49: Any Texture property tagged as "Bump/NormalMap" now adds the [Normal] tag to help ensure users give it a normal map texture.

com.unity.shadergraph/Editor/Data/Graphs/ShaderKeyword.cs
```diff
@@ -82,6 +82,15 @@ public KeywordScope keywordScope
             set => m_KeywordScope = value;
         }
 
+        [SerializeField]
+        private KeywordShaderStage m_KeywordStages = KeywordShaderStage.All;
```
cdxntchou, line 86: Adding stage controls to the Keywords

com.unity.render-pipelines.universal/Editor/ShaderGraph/UniversalProperties.cs
```diff
@@ -0,0 +1,105 @@
...
+        public static readonly string SG_AlphaClip = "_URP_AlphaClip";
+        public static readonly string SG_SrcBlend = "_URP_SrcBlend";
+        public static readonly string SG_DstBlend = "_URP_DstBlend";
+        public static readonly string SG_ZTest = "_URP_ZTest";
+        public static readonly string SG_ZWrite = "_URP_ZWrite";
+        public static readonly string SG_ZWriteControl = "_URP_ZWriteControl";
+        public static readonly string SG_Cull = "_URP_CullMode";
+        public static readonly string SG_CastShadows = "_URP_CastShadows";
+        public static readonly string SG_ReceiveShadows = "_URP_ReceiveShadows";
+        public static readonly string SG_QueueOffset = "_URP_QueueOffset";
```
phi-lira, line 52: Not having consistency between all URP shaders is a bad decision imo. It also violates the constraint 2 in your document. 
 
 This will not only make it hard for user animating property themselves as it will also break/not work with existing tooling we have in Unity like AnimationClip.  cc @amechtley

phi-lira (changes requested):
A few things:
- Is this PR breaking users? In what scenario? We need to know what is changing, have a conscious decision about that and if something cannot be autoupgraded we need to write an upgrade guide to users.

- I think at least consistency between all shaders is absolutely required here, having different shader property names on different shaders will be problematic for not only users but internal system at Unity that rely on property names.

- It's not clear from the PR which keyword change, we need to understand that as well so we can discuss impact for user.

com.unity.render-pipelines.universal/Editor/ShaderGraph/UniversalProperties.cs
cdxntchou: This PR is intended to not break any existing shader or material behavior, and we're doing extensive testing to ensure that.  

There is an edge case where if the user happens to have added a property on their shader that matches the name of one of the new target-generated properties, the shader will start issuing errors.  This makes it important to use very unique names for these properties.

Andre brought up that this PR does break an existing _workflow_ :   iterating on shader surface settings and viewing the changes in scene.  The global "Allow Material Override" checkbox that I am working on adding will fix that, by making it possible for the user to select whether they want the new Uber-shader behavior or not.   Actually, now that I think about it, since the default value there will be "false" this will also mean they won't even get the edge case I described above unless they check the "Allow Material Override" box.

cdxntchou: For the reasons described in the document, it is not possible to satisfy all of the constraints simultaneously.  Option 1 is selected as the least-bad solution.

cdxntchou: How often do users animating surface settings like alpha clipping, blend mode, transparent/opaque?
These are the only properties that will be made unique.

cdxntchou: I haven't pushed the unique keyword change yet, but the set that would be made unique are the ones driven by the properties, that can be set independently in different render pipelines:
_ALPHATEST_ON
_SURFACE_TYPE_TRANSPARENT
_ALPHAPREMULTIPLY_ON
_ALPHAMODULATE_ON

The following are using shared state between pipelines, so I think they can be kept as shared names (though it's a bit dangerous as any discrepancy between HDRP and URP setting of these may cause bugs):
_NORMALMAP
_EMISSION

esmelusina (approved):
Nothing stood out for me, maybe get another set of eyes on it if you have any concerns.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
```diff
@@ -225,20 +248,6 @@ public bool TryUpgradeFromMasterNode(IMasterNode1 masterNode, out Dictionary<Blo
                     break;
             }
 
-            // PBRMasterNode adds/removes Metallic/Specular based on settings
```
cdxntchou, line 228: Moved this logic below, somewhat simpler code.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
```diff
@@ -247,302 +256,357 @@ public bool TryUpgradeFromMasterNode(IMasterNode1 masterNode, out Dictionary<Blo
...
 
         #region SubShader
         static class SubShaders
         {
-            // Overloads to do inline PassDescriptor modifications
-            // NOTE: param order should match PassDescriptor field order for consistency
-            #region PassVariant
-            private static PassDescriptor PassVariant(in PassDescriptor source, PragmaCollection pragmas)
+            // SM 4.5, compute with dots instancing
+            public static SubShaderDescriptor LitComputeDotsSubShader(UniversalTarget target, WorkflowMode workflowMode, string renderType, string renderQueue, bool complexLit)
```
cdxntchou, line 279: Converted these static variables into static functions, as quite a bit of it is now dynamic.

com.unity.render-pipelines.universal/Editor/ShaderGraph/UniversalProperties.cs
```diff
@@ -0,0 +1,32 @@
...
+        public static readonly string ReceiveShadows = "_ReceiveShadows";
+        public static readonly string QueueOffset = "_QueueOffset";
+
+        // for ShaderGraph shaders only
+        public static readonly string ZTest = "_ZTest";
+        public static readonly string ZWriteControl = "_ZWriteControl";
+
+        // Global Illumination requires some properties to be named specifically:
+        public static readonly string EmissionMap = "_EmissionMap";
+        public static readonly string EmissionColor = "_EmissionColor";
```
jessebarker, line 30: Good that these are called out as fixed dependencies.  Is there a docs impact on this (I think Val is evaluating this as a result of another PR I did)?

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
```diff
@@ -650,26 +697,39 @@ static class LitDefines
...
+                referenceName = ShaderKeywordStrings._RECEIVE_SHADOWS_OFF,
+                type = KeywordType.Boolean,
+                definition = KeywordDefinition.ShaderFeature,
+                scope = KeywordScope.Local,
+            };
+
             public static readonly KeywordDescriptor GBufferNormalsOct = new KeywordDescriptor()
             {
-                displayName = "GBuffer normal octaedron encoding",
+                displayName = "GBuffer normal octahedron encoding",
```
jessebarker, line 732: Should this actually be "octahedral"?

jessebarker (approved):
LGTM

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
cdxntchou: Yes, as the adjective form of octahedron?   :)

com.unity.render-pipelines.universal/Editor/ShaderGraph/UniversalProperties.cs
cdxntchou: I think this is somewhat documented already -- but it's more of a lighting docs thing than a shadergraph thing.

cdxntchou:
@phi-lira @eh-unity @verasl @LandonTownsendUnity This PR is ready to go from my standpoint, all yamato tests passing, all of the feedback addressed:  the default behavior is now identical to the old behavior (Target control), and the property and keyword names match Lit.shader / Unlit.shader.  Please check if there is anything else that should be changed!

I've got tests being added in a separate branch:  sg/urp-uber-tests (all green so far there as well), but would like to get this large change in to unblock dependent work and avoid continuous master merge fixups and re-running yamato tests..

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/MaterialPostprocessor.cs
```diff
@@ -154,13 +161,23 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
...
                     else
                     {
-                        assetVersion.version = UniversalProjectSettings.materialVersionForUpgrade;
-                        debug += $" assumed to be version {UniversalProjectSettings.materialVersionForUpgrade} due to missing version.";
+                        if (shaderID.IsShaderGraph())
+                        {
+                            // ShaderGraph materials NEVER had asset versioning applied prior to version 5.
+                            // so if we see a ShaderGraph material with no assetVersion, set it to 4 to ensure we apply all necessary versions.
+                            assetVersion.version = 5;
+                            debug += $" shadergraph material assumed to be version 4 due to missing version.";
```
phi-lira, line 174: Should it be 4 or 5?

phi-lira: Why not assume `UniversalProjectSettings.materialVersionForUpgrade`?

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -214,20 +254,55 @@ public void ShaderPropertiesGUI(Material material)
         ////////////////////////////////////
         #region DrawingFunctions
 
+        public void DrawShaderGraphProperties(Material material, IEnumerable<MaterialProperty> properties)
```
phi-lira, line 257: Does this function need to be public? If so we need public API docs. We have currently some missing API docs in this class and task to fix it, but we should not introduce more public undocumented API at least.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -214,20 +254,55 @@ public void ShaderPropertiesGUI(Material material)
...
+                    continue;
+
+                float h = materialEditor.GetPropertyHeight(prop, prop.displayName);
+                Rect r = EditorGUILayout.GetControlRect(true, h, EditorStyles.layerMaskField);
+
+                materialEditor.ShaderProperty(r, prop, prop.displayName);
+            }
+        }
+
+        public static void DrawFloatToggleProperty(GUIContent styles, MaterialProperty prop)
```
phi-lira, line 274: Does it need to be public? At worst maybe protected?

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -325,19 +404,53 @@ protected static void DrawTileOffset(MaterialEditor materialEditor, MaterialProp
...
+                {
+                    // Lit.shadergraph or Unlit.shadergraph, but no material control defined
+                    // enable the pass in the material, so shader can decide...
+                    castShadows = true;
+                }
+                else
+                {
+                    // Lit.shader or Unlit.shader -- set based on transparency
+                    castShadows = Rendering.Universal.ShaderGUI.LitGUI.IsOpaque(material);
+                }
```
phi-lira, line 433: Thanks for handling this. I'll have a discussion with @Verasl I think URP team should align with ShaderGraph (we can do in follow up PR)

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -325,19 +404,53 @@ protected static void DrawTileOffset(MaterialEditor materialEditor, MaterialProp
...
+            if (material.HasProperty(Property.ReceiveShadows))
+                CoreUtils.SetKeyword(material, ShaderKeywordStrings._RECEIVE_SHADOWS_OFF, material.GetFloat(Property.ReceiveShadows) == 0.0f);
+        }
 
-            // Setup double sided GI
-            bool doubleSidedGI = (RenderFace)material.GetFloat("_Cull")  != RenderFace.Front;
-            if (doubleSidedGI != material.doubleSidedGI)
-                material.doubleSidedGI = doubleSidedGI;
+        // this is the function used by Lit.shader, Unlit.shader GUIs
+        public static void SetMaterialKeywords(Material material, Action<Material> shadingModelFunc = null, Action<Material> shaderFunc = null)
```
phi-lira, line 443: Does it need to be public/protected? If so, needs public API docs.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -350,40 +463,63 @@ public static void SetMaterialKeywords(Material material, Action<Material> shadi
...
+        public static void SetMaterialSrcDstBlendProperties(Material material, UnityEngine.Rendering.BlendMode srcBlend, UnityEngine.Rendering.BlendMode dstBlend)
+        {
+            if (material.HasProperty(Property.SrcBlend))
+                material.SetFloat(Property.SrcBlend, (float)srcBlend);
+
+            if (material.HasProperty(Property.DstBlend))
+                material.SetFloat(Property.DstBlend, (float)dstBlend);
+        }
+
+        public static void SetMaterialZWriteProperty(Material material, bool zwriteEnabled)
```
phi-lira, line 497: Same here, evaluate if needs to be public and if so add API docs.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/DepthNormalsOnlyPass.hlsl
```diff
@@ -19,7 +19,7 @@ half4 frag(PackedVaryings packedInput) : SV_TARGET
     SurfaceDescriptionInputs surfaceDescriptionInputs = BuildSurfaceDescriptionInputs(unpacked);
     SurfaceDescription surfaceDescription = SurfaceDescriptionFunction(surfaceDescriptionInputs);
 
-    #if _AlphaClip
+    #if _ALPHATEST_ON
```
phi-lira, line 22: Was this a bug?

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalTarget.cs
```diff
@@ -41,15 +61,24 @@ enum AlphaMode
         Multiply,
     }
 
-    sealed class UniversalTarget : Target, ILegacyTarget
+    public enum RenderFace
```
phi-lira, line 64: Needs public API docs.

com.unity.shadergraph/Editor/Generation/TargetResources/Fields.cs
```diff
@@ -14,13 +14,12 @@ internal static class Fields
         public static FieldDescriptor GraphVertex =           new FieldDescriptor(kFeatures, "graphVertex", "FEATURES_GRAPH_VERTEX");
         public static FieldDescriptor GraphPixel =            new FieldDescriptor(kFeatures, "graphPixel", "FEATURES_GRAPH_PIXEL");
         public static FieldDescriptor GraphColorInterp =      new FieldDescriptor(kFeatures, "graphColorInterp", "FEATURES_GRAPH_COLOR_INTERP");
-        public static FieldDescriptor AlphaClip =             new FieldDescriptor(string.Empty, "AlphaClip", "_AlphaClip 1");
```
phi-lira, line 17: Ah so this was renamed. Will this rename break users?

phi-lira (approved):
LGTM. Minor comments, missing public API docs.
 Also, `PBRMasterGUI.cs` file was deleted but I didn't see any .meta file being deleted for that?

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
```diff
@@ -120,17 +124,50 @@ public override void GetActiveBlocks(ref TargetActiveBlockContext context)
...
             context.AddBlock(BlockFields.SurfaceDescription.Emission);
             context.AddBlock(BlockFields.SurfaceDescription.Occlusion);
-            context.AddBlock(BlockFields.SurfaceDescription.Specular,           workflowMode == WorkflowMode.Specular);
-            context.AddBlock(BlockFields.SurfaceDescription.Metallic,           workflowMode == WorkflowMode.Metallic);
-            context.AddBlock(BlockFields.SurfaceDescription.Alpha,              target.surfaceType == SurfaceType.Transparent || target.alphaClip);
-            context.AddBlock(BlockFields.SurfaceDescription.AlphaClipThreshold, target.alphaClip);
+
+            // when the surface options are material controlled, we must show all of these blocks
+            // when target controlled, we can cull the unnecessary blocks
+            context.AddBlock(BlockFields.SurfaceDescription.Specular,           (workflowMode == WorkflowMode.Specular) || target.allowMaterialOverride);
```
eh-unity, line 130: Since there are these clear sections for different "kinds" of blocks, perhaps we should separate them into internal helper functions?

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
```diff
@@ -247,302 +239,361 @@ public bool TryUpgradeFromMasterNode(IMasterNode1 masterNode, out Dictionary<Blo
...
+                    result.passes.Add(LitPasses.Forward(target, workflowMode));
+
+                // cull the shadowcaster pass if we know it will never be used
+                if (target.castShadows || target.allowMaterialOverride)
+                    result.passes.Add(CorePasses.ShadowCaster(target));
+
+                if (target.mayWriteDepth)
+                    result.passes.Add(CorePasses.DepthOnly(target));
+
+                result.passes.Add(LitPasses.DepthNormalOnly(target));
```
eh-unity, line 325: FYI, I added method chaining/fluent interface support to the shadergraph collections a while back. You could write these as result.passes.Add().Add().Add()...
I think this is fine as is not necessary to change.

eh-unity (approved):
Looks pretty good to me.

erikabar (approved):
@cinight is on this PR and will test it, I am approving without testing.

com.unity.shadergraph/Editor/Generation/TargetResources/Fields.cs
cdxntchou: Nope.  Fields are an internal concept in ShaderGraph, not user accessible.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/DepthNormalsOnlyPass.hlsl
cdxntchou: _AlphaClip was the name of the #define that the old URP Target used to control alpha clipping.
Since this is upgraded to a (potential) keyword now, we rename it to match the name used by URP Lit.shader.

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/MaterialPostprocessor.cs
cdxntchou: The number should be the version at which ShaderGraph Material Upgrading was first introduced.  Prior to that version, the Material Upgrader ignored ShaderGraph based Materials entirely.   Without doing this, if we assume latest version, it will not run the upgrader on any pre-existing ShaderGraph Materials, assuming they have already been upgraded.  This was a problem when I had an upgrade pass to apply keywords to all pre-existing ShaderGraph Materials.  That pass was removed as unnecessary now that we set "Allow Material Override" to false by default -- since no keywords are needed in that mode (it's effectively the same as the old behavior), it no longer needs to update any pre-existing graphs.   I left it in there though in case we do add upgraders that affect ShaderGraph in the future, as we can't assume latest version for ShaderGraphs.

cdxntchou: It was 4, but someone added another upgrader pass which got merged in, so I bumped it to 5.  I will fix the comment to match.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
cdxntchou: Nothing "needs" to be public for it to work.   I guess the question is whether the public would like to be able to use it.   In this case, ShaderGraph is exposing a function to effectively replace this one, so I can make this one internal for now.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
cdxntchou: I'll set it internal.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
cdxntchou: The function signature is unchanged in this PR (just the diff is bad here -- it's not matching up correctly with the function declaration -- should be line ~328 in original)

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/DepthNormalsOnlyPass.hlsl
cdxntchou: This doesn't break shadergraphs because it was always a generated #define, not exposed to user in any way.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
cdxntchou: changed it to internal

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
cdxntchou: internal

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalTarget.cs
cdxntchou: internal

LandonTownsendUnity (approved):

LandonTownsendUnity (changes requested):
That render queue override we saw during the pair testing doesn't exist on master - it looks like this PR caused that. 
image

LandonTownsendUnity (approved):
Approving now that the advanced controls are fixed.

cinight (approved):
All good. Test doc: https://docs.google.com/document/d/1Td6gGYRSYRDAXeFViWiYlR3IWktUeRA_BT8gdUdSZu8/edit#

marctem (approved):

PR #3707, opened 2021-03-01
Avoid unnecessary RenderGraphBuilder.ReadTexture in the "Set Final Target" pass.

victorsclui:
### Purpose of this PR

In the "Set Final Target" render pass, sometimes it is necessary to copy the depth buffer. However we are currently calling RenderGraphBuilder.ReadTexture even if we are not coping the depth buffer. This PR is a tiny change that avoids unnecessary RenderGraphBuilder.ReadTexture in the "Set Final Target" pass.

---
### Testing status

Yamato Nightly HDRP test all green (except HDRP on OSX_Metal_playmode_cache_mono_Linear which is red anyways).

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffinal-target-pass-avoid-depth-texture-read


sebastienlagarde (approved):
Please add a changelog

victorsclui:
Changelog added

JulienIgnace-Unity (approved):

PR #3708, opened 2021-03-01
[2021.2] Fixed fogbugz case 1312513

thomas-zeng:
---
### Purpose of this PR
- Fixed case: https://fogbugz.unity3d.com/f/cases/1312513/

---
### Testing status
- ABV
- QA

---
### Comments to reviewers
- Removed TryGetDisplayRefreshRate call as openXR does not support it.
- Changed default frame target to 300 so that XR application target fps never throttle the frame cadence.


0lento:
Related to same issue: https://github.com/Unity-Technologies/Graphics/pull/3698

I see this PR didn't put platform checks in, should we really run this part on PC if the earlier fix (that caused the regression with OpenXR) only affects Quest right now?

As additional note, current OpenXR implementation doesn't support Quest natively atm so can't test how it behaves there with it.

jackpr-unity (approved):
Testing against the original bug, the packages produced by this PR look good
- Running with the MockHMD, we sustain 300 fps
- Using the Valve Index, we easily hit the target 144 hz refresh rate. The SteamVR advanced frame timer shows we hit framerate with good available overhead.

thomas-zeng:
Hi @0lento，
Thanks for the feedback.
The reason we didn't put platform ifdef here is because we believe the logic should work for all platforms.
The Quest fix you mentioned is to fix an issue where android platform sets target framerate to 30fps by default. And in XR case, XR application's frame cadence should be controlled by XR provider, not unity code.
With this PR's change, SRP sets target framerate to 300 which should be higher than all XR provider's display frequencies. This allows frame cadence to be controlled by provider(and not throttled by SRP).

erikabar (approved):

phi-lira (approved):

PR #3713, opened 2021-03-02
[WIP][URP][HDRP] Add SpeedTree8 shadergraph and update material upgraders

nerites:

### Checklist for PR maker

---
### Purpose of this PR
Adding SpeedTree8 support (including material upgrader) to HDRP.
Added the SpeedTree8 shadergraph to URP as well, though the default is still the original SpeedTree8 shader pending wind support in the shadergraph. 

Wind doesn't work in the shadergraphs for reasons I don't understand at the moment. 

For now, seeking reviews from ShaderGraph, URP, and HDRP on my current implementation, since it shouldn't be significantly affected by wind-related changes.

---
### Testing status
HDRP:
- Imported SpeedTree8 assets and confirmed shaders were correctly applied with expected wind and billboard keywords
- Clicked SpeedTree8 materials to confirm that shader keywords and properties get correctly retained through HDShaderUtils.ResetMaterialKeywords

URP:
- Imported SpeedTree8 assets and confirmed that shaders were correctly applied with expected wind and billboard keywords

---
### Comments to reviewers
I'd love feedback on the specific things by area representatives:

Shadergraph: I'm doing the following workarounds. If Shadergraph offers or plans to offer a better way of doing these things, I'd like to know how.
- For `#define KEYWORD`, I'm using Material.EnableKeyword() in asset postprocessing.
- Instead of creating a SpeedTree8 subtarget, I've kept the ST8 shadergraphs as "lit" subtargets, instead handling any keyword/property overriding in material postprocessing that gets triggered by import or material dirty. 
- (Subset of previous bullet) For setting culling mode, I'm setting _CullMode as a float property in HDRP in asset postprocessing, and waiting on a per-material culling hookup to be implemented for URP. 
- For the _WindQuality property used by previous versions of ST8, I'm adding an additional, internal _WindQuality float, since the _WINDQUALITY float generated by the keyword enum I added in Shadergraph does not get detected by native code due to differences in casing. 

HDRP:
- I've added speedtree-specific handling in MaterialPostProcessor, HDShaderUtils.ResetMaterialKeywords, and they all use a function provided by my new SpeedTree8MaterialUpgrader. Does that seem like the right way to do things?

URP:
- I've added ST8-specific handling in MaterialPostprocessor and UniversalRenderPipelineMaterialUpgrader. They shader some code that would be nice to define once instead of in two places- any recommendations for how I should better organize this?


com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree8.shader
```diff
@@ -50,7 +50,7 @@ Shader "Universal Render Pipeline/Nature/SpeedTree8"
 
             #pragma vertex SpeedTree8Vert
             #pragma fragment SpeedTree8Frag
-
+#pragma enable_d3d11_debug_symbols
```
ellioman, line 53: This should be removed :)

com.unity.render-pipelines.high-definition/Editor/Material/Nature/SpeedTree8MaterialUpgrader.cs
```diff
@@ -0,0 +1,146 @@
...
+                            material.EnableKeyword("EFFECT_BILLBOARD");
+                            material.SetFloat("EFFECT_BILLBOARD", 1);
+                        }
+                        else
+                        {
+                            material.DisableKeyword("EFFECT_BILLBOARD");
+                            material.SetFloat("EFFECT_BILLBOARD", 0);
+                        }
+                        // Billboards have double sided mode = none.
+                        material.SetVector("_DoubleSidedConstants", kDoubleSidedNone);
```
adrien-de-tocqueville, line 111: Are billboard double sided ? mode = none means that normals on the backface will be set equal to normals on the front face

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
```diff
@@ -440,10 +441,75 @@ internal SpeedTreeBillboardUpgrader(string oldShaderName)
     }
     internal class SpeedTree8Upgrader : MaterialUpgrader
     {
+        private enum WindQuality
```
adrien-de-tocqueville, line 444: there's a lot of duplicate code with HDRP, maybe some of it can be moved to render-pipelines.core ?

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
```diff
@@ -78,7 +78,7 @@ public enum UpgradeFlags
         /// </summary>
         /// <param name="material">Material to upgrade.</param>
         /// <param name="flags">Upgrade flag</param>
-        public void Upgrade(Material material, UpgradeFlags flags)
+        public virtual void Upgrade(Material material, UpgradeFlags flags)
```
adrien-de-tocqueville, line 81: is that change necessary, doesn't seem to be used

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesFunctions.hlsl
```diff
@@ -283,4 +284,33 @@ float2 GetNormalizedScreenSpaceUV(float4 positionCS)
     #define UnityStereoTransformScreenSpaceTex(uv) uv
 #endif // defined(UNITY_SINGLE_PASS_STEREO)
 
+uint2 ComputeFadeMaskSeed(float3 V, uint2 positionSS)
```
adrien-de-tocqueville, line 287: this function is already in LODCrossFade.hlsl, is there a reason to have it twice ?

com.unity.render-pipelines.high-definition/Editor/Material/Nature/SpeedTree8MaterialUpgrader.cs
nerites: Billboards get backfaces culled, so they have double sided mode disabled. Or is there another mode I should use?

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesFunctions.hlsl
nerites: It's defined in HDRP's ShaderVariableFunctions but not in Universal, so when I define it in LODCrossfade I get redefinition errors. 
So I moved the definition to Universal's ShaderVariableFunctions instead.

com.unity.render-pipelines.high-definition/Editor/Material/Nature/SpeedTree8MaterialUpgrader.cs
adrien-de-tocqueville: If backfaces are culled the mode doesn't have imoprtance, but I think then the keyword _DOUBLESIDED_ON should be disabled (currently it's always enabled at the beginning of the function)

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
nerites: Maybe I should move it to shadergraph instead? Since this upgrader is specifically for the ST8 shadergraph.

com.unity.render-pipelines.high-definition/Editor/Material/Nature/SpeedTree8MaterialUpgrader.cs
nerites: That makes sense!

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
nerites: Ah, SHadergraph doesn't have MaterialUpgrader defined because that's defined in core.

com.unity.render-pipelines.high-definition/Editor/Material/Nature/SpeedTree8MaterialUpgrader.cs
```diff
@@ -0,0 +1,146 @@
...
+                        windInt = GetWindQualityFromKeywords(material.shaderKeywords);
+                        if (!WindIntValid(windInt))
+                            windInt = material.HasFloat("_WindQuality") ? (int)material.GetFloat("_WindQuality") : 0;
+                        if (!WindIntValid(windInt))
+                            windInt = 0;
+                    }
+
+                    material.EnableKeyword(WindQualityString[windInt]);
+                    material.SetFloat("_WindQuality", windInt); // A legacy float used in native code to apply wind data
+                    material.SetFloat("_WINDQUALITY", windInt); // The actual name of the keyword enum for the shadergraph
```
adrien-de-tocqueville, line 93: I am not sure this works to set a keyword. If you are relying on the fact that the _WINDQUALITY property has a shaderlab attribute, I think you need to call
```MaterialEditor.ApplyMaterialPropertyDrawers(material)```
in order for the keywords to be regenerated
Maybe it's called somewhere else for you but I am not sure

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
adrien-de-tocqueville: shadergaph has a dependency on core rp though so you should still be able to access it

cdxntchou:
> For #define KEYWORD, I'm using Material.EnableKeyword() in asset postprocessing.

This sounds like the right way to do things.  They are shader_feature keywords you are enabling based on some decision in the importer?

> Instead of creating a SpeedTree8 subtarget, I've kept the ST8 shadergraphs as "lit" subtargets, instead handling any keyword/property overriding in material postprocessing that gets triggered by import or material dirty.

This is better than writing a new subtarget.  Only question is whether someone may want to come along and modify these decisions later on -- can they just re-run the importer?  Or can they modify the Material directly?

> (Subset of previous bullet) For setting culling mode, I'm setting _CullMode as a float property in HDRP in asset postprocessing, and waiting on a per-material culling hookup to be implemented for URP.

Only thing to be aware of is that HDRP (and future URP) material inspector will allow users to change the _CullMode themselves -- but if that's desired, then great!  You're just setting it up for them initially.

> For the _WindQuality property used by previous versions of ST8, I'm adding an additional, internal _WindQuality float, since the _WINDQUALITY float generated by the keyword enum I added in Shadergraph does not get detected by native code due to differences in casing.

This sounds like an issue to fix -- using a float instead of a keyword will result in it being a dynamic branch, which you probably do not want, since WindQuality is used to disable wind for performance.   As a dynamic branch it would not give you any performance back.

nerites:
Thanks for the reviews everyone! I've taken all the feedback and cleaned up my code considerably. Since Shadergraphs aren't backwards compatible, I'm starting from 7.x.x and moving forwards. Here's the next iteration: https://github.com/Unity-Technologies/Graphics/pull/3789

PR #3716, opened 2021-03-02
[URP] fixed regression regarding hdr toggle

martint-unity:
---
### Purpose of this PR

#2508

---
### Testing status
Yamato

---
### Comments to reviewers


ellioman (approved):

phi-lira (approved):

phi-lira:
Adding @Verasl as I know he doesn't like the SceneView checks. 
  For context, when a camera target is set to a render texture, we should create the intermediate texture with same format as camera target. However the scene view camera has a render texture as target and we still want to control the format being LDR or HDR to match game view. Hence the change. 
  
  @martint-unity and @Verasl another question, when a camera has render texture as target, does it make sense to display HDR toggle since we take the format from the texture? Maybe we should grey it out in a follow up PR?

erikabar (approved):
Verified https://fogbugz.unity3d.com/f/cases/1313942/
also checked if previous functionality works (scene view and HDR settings)

Verasl (approved):
LGTM

PR #3718, opened 2021-03-02
[URP] fixed regression regarding hdr toggle

martint-unity:
---
### Purpose of this PR

#2508

backport of this PR #3716

---
### Testing status

---
### Comments to reviewers


ellioman (approved):

phi-lira (approved):

erikabar (approved):
Verified https://fogbugz.unity3d.com/f/cases/edit/1313942/
Tested if existing functionality with HDR and Scene view works

PR #3723, opened 2021-03-02
Removed shader variants for every split project except 2d and lighting

Nzollner:
### Purpose of this PR
This PR ensures unnecessary shader variants are removed from each split project for URP except for UniversalGraphicsTest_2D and UniversalGraphicsTest_Lighting.
By doing this will we need less time to run the lighting tests.
---
### Testing status
I have tested if shader variants that should not be there are removed.
---
### Comments to reviewers
The tool I used to check shader variants were:
https://github.com/Unity-Technologies/Graphics-QA-Scripts/tree/master/Shader_Variant_Tool


cinight (approved):
Only Lighting tests are failing and this PR didn't change anything in Lighting project so this PR is good.

phi-lira (approved):

PR #3724, opened 2021-03-02
Removed shader variants from lighting

Nzollner:
### Purpose of this PR
This PR ensures unnecessary shader variants are removed from UniversalGraphicsTest_Lighting.
By doing this will we need less time to run the lighting tests.
---
### Testing status
I have tested if shader variants that should not be there are removed.
---
### Comments to reviewers
The tool I used to check shader variants were:
https://github.com/Unity-Technologies/Graphics-QA-Scripts/tree/master/Shader_Variant_Tool


cinight (approved):
The Failed Lighting tests result looks same as Graphics master.
So not this PR's problem.

phi-lira (approved):

KEngelstoft:
This PR causes a new test failure

Test Name: UniversalProjectEditorTests.CheckAllLightingSettings
Test Fixture:
Test Status: Failure
Test Execution Time: 00:00:00.062
Message:
  Lighting settings (Assets/Scenes/034_Shader_HighlightsEnvironmentGradientBakedSettings.lighting) have baked GI with filter mode enabled. It is recommended to turn of filter mode to reduce halo effect (If you still want to use it please contact URP team first).
  Expected: True
  But was:  False

PR #3726, opened 2021-03-02
Add gfx-qa-foundation to URP codeowners

pbbastian:
### Purpose of this PR
We decided in the last URP meeting to add QA as codeowners for the URP folder, so that they can decide what should happen QA-wise.


erikabar (approved):
Looks good

PR #3727, opened 2021-03-02
APV  Debug fixes

alelievr:
### Purpose of this PR
- Fix culling issues in the probe volume debug component
- Apply HDRP pre-exposure to the probes (so I renamed the exposure field to exposure offset)
image
- Fix probes debug appearing behind the fog in HDRP
- Hide debug lines when they are behind geometry instead of being rendered with a less visible color
image
- Fixed a leak in the debug code that was making it unusable after few seconds of visualizing probes. As a side effect, the probe debug is only visible in the scene view.

Note that to be able to apply the pre-exposure in the debug shader, I needed to make a specific shader to display the probes in HDRP, which means I also had to add a new system to get this shader from core.
So I added this interface: `IOverrideCoreEditorResources` which allows a RenderPipelineAsset to override the shader for core (the name is generic just in case we add more stuff in it later).

I didn't do the doc yet as I want to check if the design is okay as it is.

---
### Testing status
Played around with the Probe Reference Volume component debug modes:
- Draw Probes with fog
- Checked that the exposure of the probe debug is applied correctly in HDRP
- Checked with both scene & game view for leaks when rendering the probe debug
- Couldn't test on URP to see if the correct shader is used because APV is not supported yet :/ 
image

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoringEditor.cs
+++ b/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoringEditor.cs
@@ -211,7 +211,7 @@ namespace UnityEngine.Rendering
         void UpdateDebugMaterial()
         {
             Shader debugShader = Shader.Find("Hidden/InstancedProbeShader");
-            if (GraphicsSettings.renderPipelineAsset is IOverrideCoreEditorResources overrideResources)
+            if (QualitySettings.renderPipeline is IOverrideCoreEditorResources overrideResources)
                 debugShader = overrideResources.GetProbeVolumeProbeShader();
 
             debugMaterial = new Material(debugShader) { enableInstancing = true };


com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/Shaders/ProbeVolumeGizmo.shader
```diff
@@ -0,0 +1,124 @@
...
+    Properties
+    {
+    }
+
+    SubShader
+    {
+        Tags{ "RenderPipeline" = "HDRenderPipeline" "RenderType" = "Opaque" }
+        LOD 100
+
+        CGINCLUDE
```
sebastienlagarde, line 12: Use HLSLINCLUDE and not CGINCLUDE

com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/Shaders/ProbeVolumeGizmo.shader
```diff
@@ -0,0 +1,124 @@
...
+        #include "UnityCG.cginc"
```
sebastienlagarde, line 13: remove #include "UnityCG.cginc" (it is only for builtin)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineAsset.cs
```diff
@@ -24,7 +24,7 @@ enum LensAttenuationMode
     /// High Definition Render Pipeline asset.
     /// </summary>
     [HelpURL(Documentation.baseURL + Documentation.version + Documentation.subURL + "HDRP-Asset" + Documentation.endURL)]
-    public partial class HDRenderPipelineAsset : RenderPipelineAsset, IVirtualTexturingEnabledRenderPipeline
+    public partial class HDRenderPipelineAsset : RenderPipelineAsset, IVirtualTexturingEnabledRenderPipeline, IOverrideCoreEditorResources
```
jenniferd-unity, line 27: this is related to Erik Hakala's work. @JulienIgnace-Unity could you confirm the design?

jenniferd-unity: also Resources are moving to HD's Global Settings

jenniferd-unity (commented):
Need to sync with upcoming changes

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoringEditor.cs
```diff
@@ -125,7 +158,7 @@ public override void OnInspectorGUI()
                     EditorGUI.BeginDisabledGroup(!m_DrawProbes.boolValue);
                     m_ProbeShading.enumValueIndex = EditorGUILayout.Popup("Probe Shading Mode", m_ProbeShading.enumValueIndex, ProbeShadingModes);
                     EditorGUI.BeginDisabledGroup(m_ProbeShading.enumValueIndex != 1);
-                    m_Exposure.floatValue = EditorGUILayout.FloatField("Probe exposure", m_Exposure.floatValue);
+                    m_Exposure.floatValue = EditorGUILayout.FloatField("Probe Exposure Offset", m_Exposure.floatValue);
```
FrancescoC-unity, line 161: Better to say Exposure Compensation here to be more in line with nomenclature.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoringEditor.cs
```diff
@@ -165,12 +198,141 @@ private void Constrain()
...
+            if (GraphicsSettings.renderPipelineAsset is IOverrideCoreEditorResources overrideResources)
+                debugShader = overrideResources.GetProbeVolumeProbeShader();
+
+            debugMaterial = new Material(debugShader) { enableInstancing = true };
+        }
+
         public void OnSceneGUI()
         {
-            ProbeReferenceVolumeAuthoring pvra = target as ProbeReferenceVolumeAuthoring;
+            // if (Event.current.type == EventType.Repaint)
```
FrancescoC-unity, line 222: Why this commented out line? leftover or reminder? If the latter a comment would be nice

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoringEditor.cs
```diff
@@ -165,12 +198,141 @@ private void Constrain()
...
+                {
+                    if (actualTarget.ShouldCull(debug.cellPosition))
+                        continue;
+
+                    for (int i = 0; i < debug.probeBuffers.Count; ++i)
+                    {
+                        var probeBuffer = debug.probeBuffers[i];
+                        var props = debug.props[i];
+                        props.SetInt("_ShadingMode", m_ProbeShading.intValue);
+                        props.SetFloat("_Exposure", -m_Exposure.floatValue);
```
FrancescoC-unity, line 248: Let's change the name of the variable too to represent what it is

com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/Shaders/ProbeVolumeGizmo.shader
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        float3 evalSH(float3 normal, float4 SHAr, float4 SHAg, float4 SHAb)
+        {
+            float4 normalPadded = float4(normal, 1);
+
+            float3 x;
+
+            // Remove bound by L0 again
+            SHAr.xyz = (SHAr.xyz - 0.5f) * (SHAr.w * 4.0f);
```
FrancescoC-unity, line 67: We changed the encoding since so this needs updating

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoringEditor.cs
```diff
@@ -165,12 +198,141 @@ private void Constrain()
...
+
+                for (int batchIndex = 0; batchIndex < debugData.probeMaps.Count; batchIndex++)
+                {
+                    for (int indexInBatch = 0; indexInBatch < debugData.probeMaps[batchIndex].Length; indexInBatch++)
+                    {
+                        int probeIdx = debugData.probeMaps[batchIndex][indexInBatch];
+
+                        shBuffer[0][indexInBatch] = new Vector4(cell.sh[probeIdx][0, 3], cell.sh[probeIdx][0, 1], cell.sh[probeIdx][0, 2], cell.sh[probeIdx][0, 0]);
+                        shBuffer[1][indexInBatch] = new Vector4(cell.sh[probeIdx][1, 3], cell.sh[probeIdx][1, 1], cell.sh[probeIdx][1, 2], cell.sh[probeIdx][1, 0]);
+                        shBuffer[2][indexInBatch] = new Vector4(cell.sh[probeIdx][2, 3], cell.sh[probeIdx][2, 1], cell.sh[probeIdx][2, 2], cell.sh[probeIdx][2, 0]);
```
FrancescoC-unity, line 322: I see we still do L1 all the time, wonder if it is worth adding L2 when we are in that mode

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoringEditor.cs
alelievr: It was not supposed to be pushed, I'll revert it

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoringEditor.cs
alelievr: Considering that the L2 mode is completely broken right now, I won't be adding it in this PR. but that's an improvement to consider in the future. Also, it would be necessary to link it with the setting we have in HDRP, so the gizmo is in sync with either L1 or L2 used in the rendering.

PR #3730, opened 2021-03-02
Cleanup shader UI - 2

adrien-de-tocqueville:
### Purpose of this PR

Follow up to PR #3462. This one is moving some api to core in order to share it with URP (URP change will be done in a separate PR)

Finally I removed some incorrect [HideInInspector] attributes in shaders (doesn't do anything for now but it will be important for material variants)

~~It also does a change on HDRP to copy URP behavior: When assigning an emissive map with the color being black (the default value), the color is changed to white : this makes the emissive map visible, otherwise assigning the texture doesn't result in any visible change~~

I also updated the decal inspector to show as disabled things where the affect checkbox is untick. Also the opacity color picker turns into a slider for opacity only, which i believe is easier to understand because base color is unticked but the slider still has effect.
(Note: I disabled the UI not affected but it can also be hidden if you prefer)

decal

While working with decals I also noticed the option to disable on terrain materials was not displayed in the surface options
I also fixed a bug with the terrain header that could not be folded independently of the surface options header

image

---
### Testing status

checked for missing doc
general testing of material inspector but really nothing have changed
Tested different variants of unticking decal surface options to see if UI reacts correclty

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/TerrainLit/TerrainLitGUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/TerrainLit/TerrainLitGUI.cs
@@ -14,12 +14,12 @@ namespace UnityEditor.Rendering.HighDefinition
     /// </summary>
     class TerrainLitGUI : HDShaderGUI, ITerrainLayerCustomUI
     {
-        const SurfaceOptionUIBlock.Features surfaceOptionFeatures = SurfaceOptionUIBlock.Features.Unlit;
+        const SurfaceOptionUIBlock.Features surfaceOptionFeatures = SurfaceOptionUIBlock.Features.Unlit | SurfaceOptionUIBlock.Features.ReceiveDecal;
 
         [Flags]
         enum Expandable
         {
-            Terrain     = 1 << 0,
+            Terrain     = 1 << 1,
         }
 
         MaterialUIBlockList uiBlocks = new MaterialUIBlockList
@@ -40,7 +40,7 @@ namespace UnityEditor.Rendering.HighDefinition
                 // TODO: move the terrain UI to a MaterialUIBlock to clarify the code
                 DrawTerrainGUI(materialEditor);
 
-                uiBlocks.FetchUIBlock<SurfaceOptionUIBlock>().UpdateMaterialProperties(props);
+                uiBlocks.FetchUIBlock<AdvancedOptionsUIBlock>().UpdateMaterialProperties(props);
                 uiBlocks.FetchUIBlock<AdvancedOptionsUIBlock>().OnGUI();
 
                 ApplyKeywordsAndPassesIfNeeded(changed.changed, uiBlocks.materials);


com.unity.render-pipelines.high-definition/Editor/Material/Unlit/BaseUnlitGUI.cs
```diff
@@ -265,7 +265,11 @@ public static void SetupMainTexForAlphaTestGI(this Material material, string col
             if (material.HasProperty(colorMapPropertyName))
             {
                 var mainTex = material.GetTexture(colorMapPropertyName);
+                var mainTexScale = material.GetTextureScale(colorMapPropertyName);
+                var mainTexOffset = material.GetTextureOffset(colorMapPropertyName);
                 material.SetTexture("_MainTex", mainTex);
+                material.SetTextureScale("_MainTex", mainTexScale);
+                material.SetTextureOffset("_MainTex", mainTexOffset);
```
adrien-de-tocqueville, line 272: I also did that because URP does it, but couldn't find confirmation it was necessary. It just seems more logical

com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
```diff
@@ -277,8 +277,8 @@ Shader "HDRP/LayeredLit"
         [HideInInspector] _StencilWriteMaskMV("_StencilWriteMaskMV", Int) = 32 // StencilUsage.ObjectMotionVector
 
         // Blending state
-        [HideInInspector] _SurfaceType("__surfacetype", Float) = 0.0
-        [HideInInspector] _BlendMode ("__blendmode", Float) = 0.0
+        _SurfaceType("__surfacetype", Float) = 0.0
```
sebastienlagarde, line 280: note: HideInInspector shouldn't matter here as we have our custom editor anyway

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
adrien-de-tocqueville: actually, all the HideInInspector are useless because we have our own editor. But the tag will be used for material variants to display the number of overriden properties (it's a UX request). And we don't want to count these properties in

sebastienlagarde:
PR is ready, waiting for QA

adrien-de-tocqueville:
Since QA hasn't gone trough it yet I added some usability improvements

TomasKiniulis (approved):
Looks good! Haven't ran into any issues.

What I've checked:
- Made sure that correct properties are grayed out and displayed changing in UI
- Properties are correctly updated when modified through script
- Checked that HDRP/TerrainLit now receives or ignores decals depending on "Receive Decals" option

PR #3736, opened 2021-03-03
Avoid dynamic allocation in GBufferPass.Configure().

kaychang-unity:
---
### Purpose of this PR
Removed dynamic allocation inside GBufferPass.Configure().

---
### Testing status
Ran UniversalGraphicsTest locally.

---

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
+++ b/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
@@ -366,6 +366,7 @@ namespace UnityEngine.Rendering.Universal.Internal
         internal RenderTargetHandle TileDepthInfoTexture { get; set; }
 
         internal RenderTargetIdentifier[] GbufferAttachmentIdentifiers { get; set; }
+        internal GraphicsFormat[] GbufferFormats { get; set; }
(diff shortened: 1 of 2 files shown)


kaychang-unity:
@jonuuukas After checking the code, I realized the original logic was not completely incorrect. The Lighting buffer graphics format is actually not known by DeferredLights (it could be R11G11B10, R16G16B16A16) and GraphicsFormat.None will be returned by DeferredLights.GetGbufferFormat(3). Is that an issue?

ellioman (approved):
Looks good. Tested on my branch that has deferred scenes and no errors 👍

jonuuukas (approved):
This is not an issue at the moment - tested it on RenderPass-enabled Deferred renderer and everything is fine as there is currently a workaround for GraphicsFormat.None that is provided by the gBufferFormat array in the MRT setup of renderPass attachments. The workaround might be faulty though and we'll fix it before making the RenderPass API public. What would be the correct way to get the actual format of the gbufferLighting attachment?

kaychang-unity:
The gbufferLighting texture is the camera color target, so I suppose you can access it as `cameraData.cameraTargetDescriptor.graphicsFormat`.

PR #3737, opened 2021-03-03
forgot null check on preview parsing- causing issues with block node …

esmelusina:
### Purpose of this PR
Accidentally pushed a null ref in #3682.

---
### Testing status
Deleting a custom block node would cause a null ref in preview manager, resulting in the deletion to fail.

---
### Comments to reviewers
Whoops.


cdxntchou (approved):

jessebarker (approved):
LGTM

PR #3739, opened 2021-03-03
[Fix] case 1314613: Remove GC from Motion Blur

he-ax:
### Purpose of this PR
Fix issue ([case 1314613](https://fogbugz.unity3d.com/f/cases/1314613/)) where boxing would create GC when comparing MotionBlurModeParameter in IsActive().
Universal does not provide a motion blur solution, this happens when using Post Processing (ppv2) with Universal.

---
### Testing status
Tested manually with bug project on windows.

---
### Comments to reviewers
Should this be backported?


nigeljw-unity (approved):
LGTM

erikabar (approved):
approving, Hedvig tested it with the project from bug report

PR #3742, opened 2021-03-03
[PVV2] Remove 2018.4 support in Yamato

sebastienlagarde:
### Purpose of this PR
Remove 2018.4 support in Yamato

---
### Testing status

---
### Comments to reviewers
I guess currently PPV2 aren't generated so I touched directly the yaamto script


theo-pnv (approved):
Yes, the ppv2 yml file is not handled by the ruamel generator so we're fine by directly editing it. It won't be overridden by future PRs. 
Thanks for the PR!

PR #3743, opened 2021-03-03
[HDRP] Override Fixed Exposure Volume in Particle Shader Graph Samples Scene

TomasKiniulis:
### Purpose of this PR
Particles Shader Graph Samples scene has incorrect exposure when imported on top of HDRP Template.
Issue comes up since HDRP Template uses Default Volume "Automatic Histogram" exposure with different settings compared to project created from HDRP Wizard which has "Automatic" exposure.

Found while smoke testing 10.4.0 package release.

Previous result with samples imported on HDRP Template project:
image

HDRP wizard result on the left, HDRP Template PR result on the right:
image

---
### Testing status
Made sure that importing Particles Shader Graph samples now has correct exposure and VolumeOverrides.asset is not modified locally.


iM0ve (approved):
Looks good. With the Fixed Override 0 checked now the project will be less dependand on the default settings or them being changed

sebastienlagarde (approved):

PR #3745, opened 2021-03-03
[HDRP][Path Tracing] Make sure that objects not present  in path traced layers do not cast shadows

eturquin:
Addresses https://fogbugz.unity3d.com/f/cases/1318857
Currently, in path tracing, objects tagged as shadow casters will sill cast shadows even if they are not in the layers that have been activated for path tracing.
This PR makes sure that only objects in path traced layers (and also tagged as shadow casters) will cast shadows.
This has been tested on the scene from the corresp. bug report, and a couple others.


remi-chapelain (changes requested):
Not sure if everything was commit & pushed but still have the issue on my side. :|

eturquin:
> Not sure if everything was commit & pushed but still have the issue on my side. :|

Should be good now. ;) (was an issue with the rayTracedShadows option, that should be forced off in path tracing)

remi-chapelain (approved):
Last commit fixes the issue properly. 

Tested various combination of layers, shadow mode to see if this wasn't braking anything. 

Also, not related to this PR, but when an object in a layer is not included in pathtracer, it prevents you to see the grid in the scene view.

cc5501869a3e192d7aa38dc6f1ff97b9

is this worth fixing ?

sebastienlagarde (approved):

sebastienlagarde:
>is this worth fixing ?
guess we should check it there is other case it appear or if it is only related to path tracing. 
Could also be due to order of rendering. If it only appear with pathtracing, maybe worth checking. If we can reproduce with other stuff, let's also log a fogbugzz but will not belongs to the same team. Rather very low priority however.

eturquin:
> > is this worth fixing ?
> > guess we should check it there is other case it appear or if it is only related to path tracing.
> > Could also be due to order of rendering. If it only appear with pathtracing, maybe worth checking. If we can reproduce with other stuff, let's also log a fogbugzz but will not belongs to the same team. Rather very low priority however.

Not sure it is worth fixing: I'm assuming we would need to render the prepass with the same layer as the path tracer, so that we get correct Zbuffer?

PR #3755, opened 2021-03-04
Fix some memory leaks

FrancescoC-unity:
Fix some misc leaks in: 

- APV
- Clouds
- SSS 

when disposing an HDRP asset.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -770,6 +775,9 @@ protected override void Dispose(bool disposing)
 
             CleanupLightLoop();
 
+            ReleaseVolumetricClouds();
+            CleanupSubsurfaceScattering();
```
sebastienlagarde, line 779: can you do backport PR for 2021.1 and 2020.2 only for this CleanupSubsurfaceScattering();
 call here? (no need for changelog ,thanks)

sebastienlagarde (approved):

PR #3758, opened 2021-03-04
URP D3D11 Feature Level 10 shaders (case 1278390)

pbbastian:
### Purpose of this PR
After we moved to having two sub-shaders where one is SM4.5 and the other SM2.0, we are missing shader variants for DirectX 11 feature level 10 GPUs as they do not support SM4.5. This PR fixes that by adding `d3d11` to the SM2.0 sub-shader.

---
### Testing status
* Fix verified by QA
* Checked impact on build times and memory
  * Generally speaking increases build time for DX11 by 7-10 minutes on our foundation test project on Yamato (no cache server)
  * Run-time memory use increases from ~140 MB to ~150 MB
  * We think both of these are okay for now given the platform. In the future, we might want to introduce an option to strip either shader target entirely.

BEFORE:
ed4a789a_d3d11

AFTER:
d3d11


ellioman (approved):

phi-lira (approved):
LGTM.

erikabar (approved):
Verified the fix using -force-feature-level-10-0 argument on windows. 
all URP shaders and the template scene are rendering correctly.

erikabar (approved):

phi-lira (approved):

erikabar (approved):

PR #3764, opened 2021-03-04
Cleanup additional light data

KEngelstoft:
### Purpose of this PR
Cleanup missing behaviors, likely old LWRP Additional light data scripts since it was only assigned on the lights.
Added the auto generated Lighting Settings assets.
Scenes 145 and 146 doesn't test environment lighting, so don't include it.

### Testing status
Non-draft PR so all the graphics tests will be ran.

Gets rid of 8 warnings in the player log:
The referenced script on this Behaviour (Game Object 'Directional Light') is missing!

### Comments to reviewers
I asked Tim who originally added the scenes but he didn't remember the source of the scrips.

Should scene 145 and 146 live in the foundation project instead (no baked lighting in them)?

### DIFF
--- a/TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/145_ManyAdditionalLightShadowTypesSettings.lighting
+++ b/TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/145_ManyAdditionalLightShadowTypesSettings.lighting
@@ -48,7 +48,7 @@ LightingSettings:
   m_PVRBounces: 2
   m_PVRMinBounces: 1
   m_PVREnvironmentMIS: 1
-  m_PVRFilteringMode: 1
+  m_PVRFilteringMode: 0
   m_PVRDenoiserTypeDirect: 1
   m_PVRDenoiserTypeIndirect: 1
   m_PVRDenoiserTypeAO: 1
--- a/TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/146_AdditionalLightShadowCustomResolutionSettings.lighting
+++ b/TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/146_AdditionalLightShadowCustomResolutionSettings.lighting
@@ -48,7 +48,7 @@ LightingSettings:
   m_PVRBounces: 2
   m_PVRMinBounces: 1
   m_PVREnvironmentMIS: 1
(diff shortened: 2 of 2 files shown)


TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/031_Shader_GlossyEnvironmentSky.unity
```diff
@@ -76,21 +77,30 @@ LightmapSettings:
...
+    m_PVREnvironmentSampleCount: 500
+    m_PVREnvironmentReferencePointCount: 2048
+    m_PVRFilteringMode: 2
+    m_PVRDenoiserTypeDirect: 0
+    m_PVRDenoiserTypeIndirect: 0
+    m_PVRDenoiserTypeAO: 0
     m_PVRFilterTypeDirect: 0
     m_PVRFilterTypeIndirect: 0
     m_PVRFilterTypeAO: 0
-    m_PVRFilteringMode: 1
```
phi-lira, line 82: Not sure what this settings changes to, but we should disable filtering for our GI tests. We don't want to test filtering/denoiser in our tests and by having it enabled it just add the risk that our test fails when you make changes to filtering.

phi-lira (commented):
Not related to your PR but just a note for follow up action.
 If test 145 and 146 doesn't test GI they are wrongfully added to this test project.
 We should move them to Foundation instead.

phi-lira (approved):

TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/031_Shader_GlossyEnvironmentSky.unity
KEngelstoft: m_PVRFilteringMode is an enum describing None, Auto, Advanced for filtering and denoising.
The value isn't used if there are no lightmaps in the scene (there are no lightmaps in this case so it doesn't change anything).

KEngelstoft:
> Not related to your PR but just a note for follow up action.
> If test 145 and 146 doesn't test GI they are wrongfully added to this test project.
> We should move them to Foundation instead.

I agree.

hdb-unity (approved):
Looks good to merge once CI passes

NOTE: did not run locally, just checked CI results and PR diff.

PR #3765, opened 2021-03-04
[OnHold] Material variants for SRP

adrien-de-tocqueville:
### Purpose of this PR

SRP part of the material variants as described in this doc https://docs.google.com/document/d/1dz0KpuH5c9qyH82fiU7ajJTsbDAmEhK_W_LnbLIPS7o

goes with this c++ PR: https://github.cds.internal.unity3d.com/unity/unity/pull/3290

##### What's changed:
URP/HDRP:
- 90% of the PR is update to wrap all shader property UI in BeginProperty/EndProperty
- updated shader gui to use the new ValidateMaterial callback to setup keywords and passes automatically
- some changes to the MaterialHeaderScope to work with the space added on the left for the lock icon

SG:
- Change shadergraph import order to make sure they are imported before materials

---
### Testing status

Test document from HDRP QA: https://confluence.unity3d.com/pages/viewpage.action?pageId=121927842

---
### Comments to reviewers

What can be tested:
- Upgrade existing project (URP / HDRP)
- Verify all material property can be overriden in the inspector
- Test that there is no regression with anything related to material (upgrade scripts, or else)

### DIFF
--- a/com.unity.shadergraph/Editor/Importers/ShaderSubGraphImporter.cs
+++ b/com.unity.shadergraph/Editor/Importers/ShaderSubGraphImporter.cs
@@ -19,7 +19,7 @@ using UnityEngine.Pool;
 namespace UnityEditor.ShaderGraph
 {
     [ExcludeFromPreset]
-    [ScriptedImporter(25, Extension, -905)]
+    [ScriptedImporter(25, Extension, -2001)] // Import before ShaderGraph
     class ShaderSubGraphImporter : ScriptedImporter
     {
         public const string Extension = "shadersubgraph";
--- a/com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
+++ b/com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
@@ -24,9 +24,9 @@ namespace UnityEditor.ShaderGraph
     // sure that all shader graphs get re-imported. Re-importing is required,
     // because the shader graph codegen is different for V2.
     // This ifdef can be removed once V2 is the only option.
-    [ScriptedImporter(114, Extension, -902)]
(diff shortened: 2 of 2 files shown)


com.unity.shadergraph/Editor/Importers/ShaderSubGraphImporter.cs
```diff
@@ -19,7 +19,7 @@
 namespace UnityEditor.ShaderGraph
 {
     [ExcludeFromPreset]
-    [ScriptedImporter(25, Extension, -905)]
+    [ScriptedImporter(25, Extension, -2001)] // Import before ShaderGraph
```
cdxntchou, line 22: Should this number be smaller than the ShaderGraphImporter ?

kroatoa (commented):
Is there any reason the testing was done on DX11 only and not on any non windows platform?

com.unity.shadergraph/Editor/Importers/ShaderSubGraphImporter.cs
adrien-de-tocqueville: It could, but inside an import queue (so assets with the same import priority), assets are sorted by dependencies. In the shadergraph importer, the function GatherDependenciesFromSourceFile registers a dependency with subgraphs so they should always be imported before

adrien-de-tocqueville:
> Is there any reason the testing was done on DX11 only and not on any non windows platform?

The build was run and used by UX on mac but there hasn't been extensive testing. The code shouldn't be platform specific so I don't know if additional testing is required ?

kroatoa (approved):
Thanks for the clarification and the PR seems to otherwise well covered in reviewers.

PR #3774, opened 2021-03-04
Fix shadows in 10.x.x when rendergraph is used

FrancescoC-unity:
Because we don't have shared textures in 10.x.x the shadows have a chance (now very high) to stomp on each other if their descriptor is the same.  As a result the atlas stomped upon would be completely broken. 

This change forces the textures of shadow atlas to not be RG textures and instead imports them.

For QA: I tried the repro in the doc, but it would be nice if you could play with shadows a bit after this change.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -1163,6 +1163,7 @@ namespace UnityEngine.Rendering.HighDefinition
             m_SkyManager.Cleanup();
             CleanupVolumetricLighting();
             CleanupProbeVolumes();
+            CleanupSubsurfaceScattering();
 
             for(int bsdfIdx = 0; bsdfIdx < m_IBLFilterArray.Length; ++bsdfIdx)
             {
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -591,6 +591,8 @@ public HDRenderPipeline(HDRenderPipelineAsset asset, HDRenderPipelineAsset defau
             InitializeProbeVolumes();
             CustomPassUtils.Initialize();
 
+            m_ShadowManager.InitializeNonRenderGraphResources();
```
JulienIgnace-Unity, line 594: Maybe put a comment here to explain why the "NonRenderGraph" stuff here is called no matter what.

JulienIgnace-Unity (approved):
Let's pray it does not break something else :p

iM0ve (approved):
Looks good, fixes the issue.

**Whats tested**
* Checked that the issue I described in the 10.4 Package review document is no longer occuring
* Checked that Update on Light move shadow is still working
* Checked cached shadow atlases are now functioning
* Tried cached shadows on all types of lights
* Changing the cache shadow atlas size in HDRP asset
* Real-world use Amalienborg


https://user-images.githubusercontent.com/36502659/110104052-57c1ce80-7daf-11eb-9c29-5198c1e76d82.mp4

PR #3776, opened 2021-03-04
2021.1 Universal/fix aces filter

kaychang-unity:
---
### 
Backport from master Purpose of https://github.com/Unity-Technologies/Graphics/pull/2880/
This fixes ACES tonemapping not correctly working on Nintendo Switch.
The generated LUT had artefacts due to fp16 precision.
I tested Android Vulkan devices (Google Pixel 2, Glaxy S9 on ARM GPU), as well as iOS Metal (Iphone8), but the issue does not happens on these platforms.

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
@@ -254,11 +254,7 @@ real YCoCgCheckBoardEdgeFilter(real centerLum, real2 a0, real2 a1, real2 a2, rea
 }
 
 // Converts linear RGB to LMS
-#if defined(SHADER_API_SWITCH) // Must use full float precision on Switch to avoid artefact when using ACES tonemapping
-float3 LinearToLMS(float3 x)
-#else
-real3 LinearToLMS(real3 x)
-#endif
+float3 LinearToLMS(float3 x) // Full float precision to avoid precision artefact when using ACES tonemapping
 {
     const real3x3 LIN_2_LMS_MAT = {
         3.90405e-1, 5.49941e-1, 8.92632e-3,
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
```diff
@@ -254,7 +254,11 @@ real YCoCgCheckBoardEdgeFilter(real centerLum, real2 a0, real2 a1, real2 a2, rea
 }
 
 // Converts linear RGB to LMS
+#if defined(SHADER_API_SWITCH) // Must use full float precision on Switch to avoid artefact when using ACES tonemapping
+float3 LinearToLMS(float3 x)
+#else
 real3 LinearToLMS(real3 x)
+#endif
```
phi-lira, line 261: Imo if this function needs full precision it should be changed for all platforms, otherwise we have the same issue on mobile. Checking the original PR, @eh-unity made the same comment.

com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
```diff
@@ -408,7 +420,11 @@ real3 LinearToLogC(real3 x)
 #endif
 }
 
+#if defined(SHADER_API_SWITCH) // Must use full float precision on Switch to avoid artefact when using ACES tonemapping
+float3 LogCToLinear(float3 x)
+#else
 real LogCToLinear_Precise(real x)
```
phi-lira, line 426: This looks wrong, LogCToLinear_Precise will not be defined for SWITCH.

phi-lira (changes requested):

com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
kaychang-unity: I tested on iPhone8 metal and Galaxy S9 Mali-G72 Vulkan but could not reproduce the issue, hence why it is conservatively kept it Switch only. But I can enable it for all platforms if you prefer. There is similar floating point fix in `AcesTonemap()` and `NeutralCurve()` for Switch but done from a different PR, do you want to change them for all platforms too?

com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
kaychang-unity: Thanks, I should have checked this.
I made the fixed and confirmed this is loading on Switch.

phi-lira (approved):
LGTM. Are you going to also put a PR to master to change the behaviour to use float instead of half for all platforms?

PR #3778, opened 2021-03-05
Fixed warnings for Universal2D.hlsl

unity-cchu:
### Purpose of this PR
To fix warnings in Universal2D.hlsl

---
### Testing status
Ran graphics tests


com.unity.render-pipelines.universal/Shaders/Utils/Universal2D.hlsl
```diff
@@ -4,13 +4,13 @@
...
-    float2 uv               : TEXCOORD0;
+    half2  uv               : TEXCOORD0;
 };
 
 struct Varyings
 {
-    float2 uv        : TEXCOORD0;
-    float4 vertex : SV_POSITION;
+    float4 vertex    : SV_POSITION;
+    half2  uv        : TEXCOORD0;
```
phi-lira, line 13: Careful with this, depending on your texture resolution you won't have enough precision to store uv coord in half precision.

com.unity.render-pipelines.universal/Shaders/Utils/Universal2D.hlsl
```diff
@@ -26,8 +26,7 @@ Varyings vert(Attributes input)
 
 half4 frag(Varyings input) : SV_Target
 {
-    half2 uv = input.uv;
-    half4 texColor = SAMPLE_TEXTURE2D(_BaseMap, sampler_BaseMap, uv);
+    half4 texColor = SAMPLE_TEXTURE2D(_BaseMap, sampler_BaseMap, input.uv);
```
ellioman, line 29: This code still gives us a warning as SAMPLE_TEXTURE2D returns a float4.
So you either need to have texColor as float4 or do a cast.

ellioman (changes requested):

simon-engelbrecht-soerensen (approved):

PR #3786, opened 2021-03-05
Light2D Upgrading & Bug Fix

unity-cchu:

### Checklist for PR maker
- [X] Have you updated the changelog? Each package has a `CHANGELOG.md` file.

---
### Purpose of this PR
Added versioning and upgrading. Fixes https://fogbugz.unity3d.com/f/cases/1317755/

---
### Testing

How this was tested:
Did the instructions below.
Migrated Dragon Crashers and made sure it upgraded properly (https://github.com/Unity-Technologies/2d-rpg-demo)

Using master, create a new project using the 2D Renderer. Add two lights. 

The first should have its Shadow Strength, Volumetric Intensity, Volumetric Shadow Strength set to a non zero value.
The second should have its Shadow Strength, Volumetric Intensity, Volumetric Shadow Strength set to a zero value.

Save the scene.

Switch to 2d/Fix-for-data-defaults. Reopen the scene. The first light should have the toggles for the three properties enabled. The second light should have the toggles for the three properties disabled. Creating a new light2D (or adding a light2D component) should have all three properties disabled with their values set to a non-zero value. 

image

Additionally in the debug view all three lights should have their Component version set to Version_1

image

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/2D/Light2D.cs
+++ b/com.unity.render-pipelines.universal/Runtime/2D/Light2D.cs
@@ -331,7 +331,6 @@ namespace UnityEngine.Experimental.Rendering.Universal
             UpdateBoundingSphere();
         }
 
-#if UNITY_EDITOR
         public void OnBeforeSerialize()
         {
             m_ComponentVersion = k_CurrentComponentVersion;
@@ -349,7 +348,5 @@ namespace UnityEngine.Experimental.Rendering.Universal
                 m_ComponentVersion = ComponentVersions.Version_1;
             }
         }
-
-#endif
     }
 }


com.unity.render-pipelines.universal/Runtime/2D/Light2D.cs
```diff
@@ -318,5 +330,26 @@ private void LateUpdate()
             UpdateMesh(true);
             UpdateBoundingSphere();
         }
+
+#if UNITY_EDITOR
```
pbbastian, line 334: This is causing a compile error on build, because the implementation is missing in non editor compilations, but the interface is specified on the class regardless.

What's the reason for the implementation only being in editor?

pbbastian (changes requested):

erikabar (approved):
approving, developer tested it (see the description)

PR #3787, opened 2021-03-05
Fix Profiler errors with XR multi-pass

fabien-unity:
---
### Purpose of this PR
- Fix for https://fogbugz.unity3d.com/f/cases/1313141/
- Disable nested profiling scopes in XR

---
### Testing status
Tested locally with repro instructions from the bug report (win10 + d3d11 + Oculus + multi-pass).

---
### Comments to reviewers
This is a workaround until we can move the C++ Skybox code over to URP.
With this change, we're losing the hierarchical nested scopes for XR.


thomas-zeng (approved):

DennisDeRykeUnity (approved):
Three issues were originally reported in case 1313141: 
(1)The skybox becomes a simple color. 
(2) The hands don't even render at all. 
(3) Profiler spam. 

This PR addresses profiler spam; the shader issues have been separated into case 1316440.

Fabien has already verified that this PR fixes the profiler spam.  Per discussion, he considers this to be a low-risk change.  Without performing tests myself, I am approving contingent on Fabien verifying that automated tests are as green as the parent master commit.

eh-unity (approved):
Looks good to me.

eh-unity:
Cancelled master, Green custom revision (failed vulkan test, rerun is green).

PR #3789, opened 2021-03-06
[7.x.x] Add SpeedTree8 shadergraph to HDRP and URP

nerites:

### Checklist for PR maker

---
### Purpose of this PR
Adding SpeedTree8 support (including material upgrader) to HDRP.
Added the SpeedTree8 shadergraph to URP as well, though the default is still the original SpeedTree8 shader pending wind support in the shadergraph.
Added a graphics test in HDRP for SpeedTree8. Also added SpeedTrees to the terrainlit and terrainbaselit test scenes.

TODO: Move SpeedTree8MaterialUpgrader from Core to Shadergraph package? -> Even invoking MaterialUpgrader using Core.Editor.MaterialUpgrader does not find the class from the shadergraph package, for some reason. Why?

---
### Testing status
Tested importing and material upgrading in HDRP. It works, except for an issue with embedded materials where when the Library is nuked, the materials don't get setup properly. This is because the speedtree asset is imported before the HDRP asset is imported, so the correct shader is not grabbed, and the correct postprocessing is not run. This is a recurring issue and topic of discussion not unique to SpeedTree, so it should not be blocking for this PR.

For URP, tested that manually switching over the shader works as expected. No upgrader or importer support for the shadergraph is needed yet as it is not yet the default shader. 

---
### Comments to reviewers
@LandonTownsendUnity do you mind checking if the normal flipping is behaving as you'd expect in HDRP?
@adrien-de-tocqueville for changes to HDShaderUtils.ResetMaterialKeywords.
@cdxntchou for placement of files within shadergraph and core packages. Notably, LODCrossfade is not specific to SpeedTree. Maybe it should be placed someplace else/exposed generally somehow? 
@phi-lira for feedback on adding the shadergraph to Universal and not hooking it up since it's not intended to replace the default yet.


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDShaderUtils.cs
```diff
@@ -105,10 +105,14 @@ public static bool ResetMaterialKeywords(Material material)
...
             if (resetter != null)
             {
+                if (HDSpeedTree8MaterialUpgrader.IsHDSpeedTree8Material(material))
+                    HDSpeedTree8MaterialUpgrader.SaveHDSpeedTree8Setup(material);
                 CoreEditorUtils.RemoveMaterialKeywords(material);
                 // We need to reapply ToggleOff/Toggle keyword after reset via ApplyMaterialPropertyDrawers
                 MaterialEditor.ApplyMaterialPropertyDrawers(material);
                 resetter(material);
+                if (HDSpeedTree8MaterialUpgrader.IsHDSpeedTree8Material(material))
+                    HDSpeedTree8MaterialUpgrader.RestoreHDSpeedTree8Setup(material);
```
adrien-de-tocqueville, line 115: I am not sure you need to do that. Looking at the save/restore code, I don't see why you need to save and restore _WINDQUALITY and _CullMode, which are float properties so they don't get resetted.
About the EFFECT_BILLBOARD, it should be set or unset from inside the resetter function (the one of sg lit in your case) depending on the value of the associated property (which i don't find by the way, in the previous PR there was some code based on the name of the material but not anymore)

nerites: They get overwritten to default values- either during SynchronizeShaderGraphProperties as part of resetting, or during ApplyMaterialPropertyDrawers, since the material has been initialized by the importer, but the drawer has not. 

Let me move this code inside the resetter to see if the correct wind property gets restored after ApplyMaterialPropertyDrawers.

nerites: Yeah unfortunately ApplyMaterialPropertyDrawers resets windquality and cullmode. Is there a way to update the materialpropertydrawer so that this doesn't happen?

EFFECT_BILLBOARD would be the name of the float toggling the EFFECT_BILLBOARD keyword and would have the same problem as wind and cullmode, where it gets reset by ApplyMaterialPropertyDrawers. I can still do it by the name of the material but that feels hacky and prevents users from disabling it (which becomes relevantin later releases of Graphics, where you can override builtin keywords through the SG UI).

adrien-de-tocqueville: ApplyMaterialPropertyDrawers does not modify the property values, it only sets keywords based on the value. So it's probably something else that modifies the values. But I checked out the branch to test and for me the values were not resetted so maybe i was doing something wrong.

For EFFECT_BILLBOARD, what i was trying to say is that ther's no way (or i haven't found it) to modify the keyword value from the UI, but I believe there should be a property and the keywoard should be set depending on the property value (the keyword would actually be set by ApplyMaterialPropertyDrawers)

nerites: I confirmed that the floats seem to stick now. I must have been resetting them myself as I iterated on changes without fully understanding the system. :) Thanks so much for the help.

Cleaned up here: https://github.com/Unity-Technologies/Graphics/pull/3789/commits/6893f6902e91500e5f5b7ab3969d3b9f4d073edc

com.unity.render-pipelines.high-definition/Editor/Material/Nature/HDSpeedTree8MaterialUpgrader.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        /// <summary>
+        /// Restores SpeedTree8-specific material properties and keywords that were set during import and should not be reset.
+        /// </summary>
+        /// <param name="mat">SpeedTree8 material.</param>
+        public static void RestoreHDSpeedTree8Keywords(Material mat)
+        {
+            int wq = (int)mat.GetFloat("_WINDQUALITY");
+            mat.EnableKeyword(WindQualityString[wq]);
```
adrien-de-tocqueville, line 43: just for completeness, I don't think this is necessary because i believe it's already done by the applymaterialdrawer (quite sure but that may be worth testing).
It can be interesting to remove because if the billboard code below gets removed in next version, the function could be deleted altogether

adrien-de-tocqueville (approved):
the material keyword reset code lgtm

kroatoa:
Am I right in assuming that this PR effectively targets 19.4? 
I don't see any mention of this having landed in branches targeting the later versions, how is that tracked?

nerites:
@kroatoa Since shadergraph is not backwards compatible, I started with 19.4 and am making PRs for later versions as I get them ready. I will have a 20.2 PR up later today, but it's not tracked anywhere outside my worklog for now. :)

com.unity.render-pipelines.high-definition/Editor/Material/Nature/HDSpeedTree8MaterialUpgrader.cs
nerites: I confirmed it imports fine without it. :) Looks like streamlining the code with the upgrader solved a lot of my problems.
Removed in 7eab867

nerites:
10.x.x version here. No changes from 7.x.x. https://github.com/Unity-Technologies/Graphics/pull/3861

kroatoa (approved):
Thanks for clarifying

nerites:
@kroatoa I have some API changes that will require incrementing package versions for Core and HDRP. How do I decide whether to increment minor or major versions? Who do I need to get approval from for incrementing the versions of those packages?

phi-lira:
Did the PR to master landed? I can only find this one but it was closed: https://github.com/Unity-Technologies/Graphics/pull/3713

nerites:
@phi-lira No, I am remaking it. I can have a new PR to master up today. I started with 19.4 since Shadergraph isn't backwards compatible, so it was easier to work from 19.4 and go forward.

kroatoa:
> 
> 
> @kroatoa I have some API changes that will require incrementing package versions for Core and HDRP. How do I decide whether to increment minor or major versions? Who do I need to get approval from for incrementing the versions of those packages?

Please reach out to the RM for the branch you are targeting. if it's more than one you can also put a post in #release-management, and we'll take it from there.

nerites:
After some discussion with other folks, I'm only targeting 10.x.x and latest. PRs here: 
10.x.x: https://github.com/Unity-Technologies/Graphics/pull/3861
Latest: https://github.com/Unity-Technologies/Graphics/pull/3911

PR #3805, opened 2021-03-08
Various small improvements to render graph

JulienIgnace-Unity:
### Purpose of this PR
This PR implements several small fixes/improvements to render graph.
- RenderGraph Viewer can now capture all cameras
- Frame info log from debug window now outputs all active cameras (this required to add an "enable logging" toggle)
- Added a name to the execution parameters of render graph for better tracking
- Added commented out test code for a particular case in UpdateResourceAllocationAndSynchronization
- Render Graph resource pooled resource usage should now be stable (equivalent frame should use the same resources at the same place)
- Fixed condition for resolving Transparent motion vectors
- Removed useless dependency on the color buffer in ClearStencilbuffer and RenderGizmos functions.

---
### Testing status
Tested the Viewer/Log manually with multiple cameras (offscreen + planar). The rest is tested through automated testing.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphResourcePool.cs
+++ b/com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphResourcePool.cs
@@ -15,7 +15,7 @@ namespace UnityEngine.Experimental.Rendering.RenderGraphModule
         // Dictionary tracks resources by hash and stores resources with same hash in a List (list instead of a stack because we need to be able to remove stale allocations, potentially in the middle of the stack).
         // The list needs to be sorted otherwise you could get inconsistent resource usage from one frame to another.
         protected  Dictionary<int, SortedList<int, (Type resource, int frameIndex)>> m_ResourcePool = new Dictionary<int, SortedList<int, (Type resource, int frameIndex)>>();
-        protected List<int> m_RemoveList = new List<int>(); // Used to remove stale resources as there is no RemoveAll on SortedLists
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -1301,9 +1300,10 @@ void RenderSky(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle colorBu
                 passData.hdCamera = hdCamera;
                 passData.volumetricLighting = builder.ReadTexture(volumetricLighting);
                 passData.colorBuffer = builder.WriteTexture(colorBuffer);
-                passData.depthTexture = builder.WriteTexture(depthTexture);
+                passData.depthTexture = builder.ReadTexture(depthTexture);
```
sebastienlagarde, line 1303: isn't this a fix that should be backported?

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
JulienIgnace-Unity: We can yes but it's not super important. It only reduces the depth texture lifetime in some very specific use cases.
Can make a backport PR if you want.

FrancescoC-unity (approved):

PR #3806, opened 2021-03-08
[HDRP] Make Decal shader compatible with SRP batcher

sebastienlagarde:
### Purpose of this PR
- Fix for fogbugzz https://fogbugz.unity3d.com/f/cases/1311586/

---
### Testing status
Tested that the shader is compatible with SRP batcher, checked the both shader graph and regular shader work with Mesh decal.

---
### Comments to reviewers


sebastienlagarde:
No need for QA on this PR

arnaudcarre (approved):

PR #3809, opened 2021-03-08
[10.x][2020.3] Fix issue with blackboard scrollbars being overlapped by resize regions at window edges 

Nightmask3:
---
### Purpose of this PR

Prior to this fix, due to a recent refactor of the Shader Graph Blackboard to inherit from the GraphSubWindow class, a bug that was introduced was the scrollbars being unusable by mouse-click/drag due to the resize regions overlapping the scrollbars.

Broken behavior:

https://user-images.githubusercontent.com/13033542/110389206-0d438a80-8019-11eb-98cc-00f3aa217737.mp4

This PR changes the styling of the resizer element used in the graph sub windows and better accomodates the scrollbar and resizers without requiring padding.

Fixed behavior:

https://user-images.githubusercontent.com/13033542/110408909-3ecd4d80-803b-11eb-9de6-7e9437a5adbe.mp4

This is how the resizing elements are laid out now (for both blackboard and inspector window):

image

Note the grey boxes outside the edges of the window. 

---
### Testing status

Only styling changes made here, no code changes, so chances of any functionality regression are minimal.

### DIFF
--- a/com.unity.shadergraph/Editor/Resources/Resizable.uss
+++ b/com.unity.shadergraph/Editor/Resources/Resizable.uss
@@ -12,19 +12,23 @@ ResizableElement
 ResizableElement #right
 {
     width: 12px;
+    margin-right: -8;
     flex-direction: column;
     align-items: stretch;
 }
 
 ResizableElement #left
 {
-    width: 12px;
+    width: 14px;
+    margin-left: -8;
     flex-direction: column;
     align-items: stretch;
 }
 ResizableElement #middle
 {
     flex:1 0 auto;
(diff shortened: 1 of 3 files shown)


LandonTownsendUnity (approved):
Tested it a bit. I have only a minor complaint that if there is only one scroll bar (horizontal or vertical) the lower right most arrow isn't easily clickable. However, since you have access to the bar itself, this isn't too big of a deal.

marctem (approved):
Much better this way.

marctem (approved):

PR #3818, opened 2021-03-09
Improve probe volume profile UI

alelievr:
### Purpose of this PR

Before:
image

After:
image

- Improved a bit the UI of the Reference Volume Profile
- Replaced `Brick Size` by `Min Distance Between Probes`
- Removed the max subdivision, it is now computed from the `Min Distance Between Probes`
- Removed the Index Dimensions field, it's now computed from the position of the bricks during the GI bake and stored in the probe data asset.

Note that as a side effect of this change, after this PR, you'll need to re-bake the probe volume in your scenes to update the index dimensions in the asset.

---
### Testing status
This work is mainly about UI changes, the only dangerous change here is the Index Dimensions thing so I focused my tests on this.

I tested that after baking a scene, the light probes were correctly applied to the objects. Also tried re-loading the scene and entering play mode and I didn't see any weird behavior.

I also tested in Windows Player and it works

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
@@ -257,9 +257,12 @@ namespace UnityEngine.Rendering
         /// <param name="budget"></param>
         public void SetMemoryBudget(ProbeVolumeTextureMemoryBudget budget)
         {
-            m_MemoryBudget = budget;
-            Cleanup();
-            InitProbeReferenceVolume(s_ProbeIndexPoolAllocationSize, m_MemoryBudget, m_PendingIndexDimChange);
+            if (m_MemoryBudget != budget)
+            {
+                m_MemoryBudget = budget;
+                Cleanup();
+                InitProbeReferenceVolume(s_ProbeIndexPoolAllocationSize, m_MemoryBudget, m_PendingIndexDimChange);
+            }
         }
 
         internal void AddPendingAssetLoading(ProbeVolumeAsset asset)


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -18,23 +19,21 @@ public sealed class ProbeReferenceVolumeProfile : ScriptableObject
...
         /// </summary>
+        [Range(0.0f, 1.0f), Delayed]
         public float normalBias = 0.2f;
-        /// <summary>
-        /// Index field dimensions.
-        /// </summary>
-        public Vector3Int indexDimensions = s_DefaultIndexDimensions;
+
+        public int maxSubdivision => Mathf.CeilToInt(Mathf.Log(cellSize / minDistanceBetweenProbes, 2));
+        public int brickSize => (int)(minDistanceBetweenProbes * 4.0f);
```
FrancescoC-unity, line 36: Note that this is not correct, while it is true we have 4 probes per brick, it means that the distance between probe is BrickSize/3 not 4 (see crude visualization below)  

X ---- X ---- X ---- X

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -18,23 +19,21 @@ public sealed class ProbeReferenceVolumeProfile : ScriptableObject
...
         /// The normal bias to apply during shading.
         /// </summary>
+        [Range(0.0f, 1.0f), Delayed]
         public float normalBias = 0.2f;
-        /// <summary>
-        /// Index field dimensions.
-        /// </summary>
-        public Vector3Int indexDimensions = s_DefaultIndexDimensions;
+
+        public int maxSubdivision => Mathf.CeilToInt(Mathf.Log(cellSize / minDistanceBetweenProbes, 2));
```
FrancescoC-unity, line 35: I think we want to use brick size in this formula, not distance between probes.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
alelievr: Oops, I forgot to change it, sorry 😅

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -252,10 +253,23 @@ private static void OnAdditionalProbesBakeCompleted()
                     {
                         var asset = refVol2Asset[refVol];
                         asset.cells.Add(cell);
+
+                        foreach (var p in cell.probePositions)
```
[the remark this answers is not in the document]
FrancescoC-unity: In fact I guess we should do it only on asset load (modify the index on asset load and verify it fits the newly loaded cells)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
alelievr: Yes, thanks!

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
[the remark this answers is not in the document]
FrancescoC-unity: The init of the index is never called with maxCellIndex in any non-editor workflow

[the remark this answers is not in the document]
FrancescoC-unity: Also, if we have additive scene loading, we need to update said index whenever we load an additional scene otherwise it won't necessarily fit.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoring.cs
```diff
@@ -178,12 +177,8 @@ private void OnValidate()
...
             if (m_Profile != null)
             {
-                bool hasIndexDimensionChangedOnProfileSwitch = m_PrevProfile == null || (m_PrevProfile != null && m_PrevProfile.indexDimensions != m_Profile.indexDimensions);
-                if (hasIndexDimensionChangedOnProfileSwitch)
-                {
-                    var refVol = ProbeReferenceVolume.instance;
-                    refVol.AddPendingIndexDimensionChange(indexDimensions);
-                }
+                var refVol = ProbeReferenceVolume.instance;
+                refVol.AddPendingIndexDimensionChange(volumeAsset.maxCellIndex);
```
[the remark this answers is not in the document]
FrancescoC-unity: I would not call the pending stuff here at all unless the asset changes (or more precisely the max cell index changes), probably we don't need to call from here but can do from the ref volume alone (and check if we need changes upon asset loading)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -56,17 +55,15 @@ public bool IsEquivalent(ProbeReferenceVolumeProfile otherProfile)
...
         private SerializedProperty m_NormalBias;
         private SerializedProperty m_IndexDimensions;
 
         sealed class Styles
         {
             // TODO: Better tooltip are needed here.
             public readonly GUIContent cellSizeStyle = new GUIContent("Cell Size", "Determine the size of the cells.");
-            public readonly GUIContent brickSizeStyle = new GUIContent("Brick Size", "The smallest possible brick size.");
-            public readonly GUIContent maxSubdivStyle = new GUIContent("Max Subdivision", "The maximum number of possible subdivisions.");
+            public readonly GUIContent minDistanceBetweenProbes = new GUIContent("Min Distance Between Probes", "The minimal distance between two probes in meter.");
```
FrancescoC-unity, line 66: In meters plural

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -244,11 +251,15 @@ public static ProbeReferenceVolume instance
...
+        /// Set the memory budget for the Probe Volume System.
+        /// </summary>
+        /// <param name="budget"></param>
+        public void SetMemoryBudget(ProbeVolumeTextureMemoryBudget budget)
         {
-            m_PendingIndexDimChange = indexDimensions;
-            m_NeedsIndexDimChange = true;
-            m_NeedLoadAsset = true;
+            m_MemoryBudget = budget;
+            Cleanup();
```
FrancescoC-unity, line 261: This is a rather costly op, let's do only if there is a change

FrancescoC-unity (approved):

sebastienlagarde (approved):

PR #3821, opened 2021-03-09
Fixed the fogbugz case 1318629

thomas-zeng:
---
### Purpose of this PR
Fixed the fogbugz case https://fogbugz.unity3d.com/f/cases/1318629/

---
### Testing status
Tested locally with the repro project.

---
### Comments to reviewers
Moved base camera `BeginCameraRendering` before InitializeCameraData


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -557,6 +553,10 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
                 }
             }
 
+            using (new ProfilingScope(null, Profiling.Pipeline.endCameraRendering))
+            {
+                EndCameraRendering(context, baseCamera);
+            }
```
phi-lira, line 559: This doesn't look correct. EndCameraRendering for overlay cameras is called before the EndCameraRendering for base camera.

phi-lira (changes requested):
Can you describe in the PR more about why the EndCameraRendering for base camera has to be called after after all EndCameraRendering of overlay cameras? That seems inconsistent and most likely will introduce issues for users.

erikabar (approved):
Developer tested with repro project, approving

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
thomas-zeng: Addressed in f1a61a9

phi-lira (approved):

phi-lira:
Lots of failed tests, maybe remerge master to see the status. Currently one Hybrid tests should be failing.

thomas-zeng:
Hi @phi-lira,

I have merged with master branch and rerun yamato tests. The test results are all green besides hybrid renderer tests.

sandy-carter-unity (approved):

PR #3828, opened 2021-03-10
Improve Screen Space Reflection with WorldSpace Speed Rejection

skhiat:
### Improve SSR

Allow speed rejection algorithm use world space motion vector:
image
By selecting Motion Vector from reflecting surface or reflected surface or both:
image
With hard or soft threshold:
image

Which improve for instance we can configure to have to accumulate sample when object and camera are moving on the same direction and reject when their going on opposite direction.

https://user-images.githubusercontent.com/57442369/132487392-3a116d83-b327-4d3d-9017-be5f28cd9521.mp4

https://user-images.githubusercontent.com/57442369/132487399-f89d6d07-4dcd-494c-9f2c-dcd69a579258.mp4

Expose hard coded parameters:
test
- Roughness Bias
iMD5ErlLx2_res2

- Speed Rejection
RngcBfBAey_res

- And Allow world space motion vector etc.:
Unity_wQ5iJXUdnI

- Add FullScreen Debug View:
Unity_cWH0Q7EeVw
Unity_WDQxKDcqax


sebastienlagarde (approved):

remi-chapelain (changes requested):
While this fixes the simple camera travelling problem or simple object moving issue pretty neatly. 

It introduces some very hard to remove ghosting (needs very large speed rejection) in a classic FPS setting when moving the camera looking at a reflection on the ground, so either mapping of the slider should be updated or we need to iterate how rejection is done. 

https://user-images.githubusercontent.com/57442369/110922028-809e1280-831f-11eb-8644-9cb3981a1578.mp4

Also, those two new parameters should be located in the advanced section IMO

sebastienlagarde:
@soufianekhiat any update?

soufianekhiat:
> @soufianekhiat any update?

Back on it next week. One piece is missing for speed rejection.

remi-chapelain (changes requested):
- I tested the with the WORLD_DIFF set to 1 which gives the best results, that way we are able to have more ghosting if we want to and the "old version" by default
- I tested all camera movement (classic FPS run and cinematic movement, dolly, pan, truck.. etc). 
- I tested that it wasn't hard to find a proper threshold with a long corridor with multiple objects a different depth with a smooth floor since the threshold is in world space
- Also tested that the "approx" algo stays unchanged by this. 

Here's the changes I'd like, for me to approve the PR : 
- WORLD_DIFF is defined to 1 ✔️  (done)
- Remove multiplier of 4096 from HDRenderPipeline.LightLoop.cs (with WORLD_DIFF, between 0 and 1 seems to be a proper range now)  ✔️  (done)
- New mapping, I'd prefer precision (the power at which the factor is raised) to be set to 6 instead of 0.5 (basically change from a square root to a higher power because we need more precision for lower speed rejection value than higher. What I like is the default value to be similar to what SSR was before to prevent some unwanted ghosting, it seems to be the case at around 0.5 ✔️  (done)
- Have those 2 new parameters (roughness bias and speed rejection) in the advanced section ❌
- Also, I'm not sure we need the word "factor" in those 2 new parameters. ❌
For example, we already have speed rejection in TAA and it's clear enough this way. 
- Maybe we can mention in the documentation (around the speed rejection) what was the intent of it ? The idea that a low value is when you favors ghosting over "correct" reflection, and that the intent of exposing this is for users who'd like to make slow controlled cinematic camera movement. ❌

**Before:** 
https://user-images.githubusercontent.com/57442369/113755562-c11b6100-9710-11eb-8489-bfb3f07401d0.mp4

**After:**
https://user-images.githubusercontent.com/57442369/113755580-c678ab80-9710-11eb-9bc3-137ec1b9e4d4.mp4

remi-chapelain (commented):
Commited a few changes that includes : 
- Remapping of the userspeed so that we can reject only part of the moving reflection on screen. 
- Speed rejection controls the speed input upper value. 

In situations where the difference of speed are null, we should continue accumulate, where the difference is low we should only accumulate "a little" and if the speed is too fast, we should discard. The speed rejection factor should control where we place this threshold. 

Currently we are "blocked" because motion vectors gives flickering motion and not at all smooth speed difference even with smooth sinusoïdal movement. https://unity.gyazo.com/169d49b6a97023b6cfdb35b8e4ad0f1d

skhiat:
@remi-chapelain update.

remi-chapelain:
A lot have improved and changed since previous comment:

The approach that have been chosen is to use world space motion of vectors of surfaces.
Two modes are possible : 
- Hard rejection, which either reject of accumulate one pixel with a hard threshold. 
- Smooth rejection, which will accumulate more or less depending on the speed, allowing users to be able to control the amount of ghosting / noise they want. 

In both mode it's possible to take into account the movement of the reflected surface or/and the reflecting surface for the speed used for speed rejection. Disabling reflecting surface influence can help when the camera is moving with the same velocity as another object, to be able to reject the background but keeping what's doesn't move relative to the camera. 

What has been tested: 
- Objects moving at different speeds with a static camera
- Rotating objects not moving relative to the camera
- Moving rotating objects
- Still reject against different type of skies


https://user-images.githubusercontent.com/57442369/132487392-3a116d83-b327-4d3d-9017-be5f28cd9521.mp4

https://user-images.githubusercontent.com/57442369/132487399-f89d6d07-4dcd-494c-9f2c-dcd69a579258.mp4

https://user-images.githubusercontent.com/57442369/132487422-f7e1acd5-4df3-4601-a998-cfcb4a5383c5.mp4

In most cases this is only useful with very low camera/object movement. The scaler factor upper speed range will only work up to certain speed. With higher speed, accumulation will cause bigger problem anyway so this can't really be used. 

Also, we still get motion vectors from the skies but the sky distance is at camera far plane so if this distance is way further from the geometry it's difficult to find a common matching speed for close and far away objects because of parallax.

###
What's missing to be able to approve is: 
- Documentation
- Possibility to disable speed rejection as a whole (and let it be off by default to avoid breaking changes)
- Add ScreenSpaceReflectionSpeedRejection Debug Mode (really hard to find correct speeds threshold without) 
- Make sure the Roughness bias is "outside" the speed rejection enable bloc

remi-chapelain (changes requested):
A lot have improved and changed since previous comment:

The approach that have been chosen is to use world space motion of vectors of surfaces.
Two modes are possible : 
- Hard rejection, which either reject of accumulate one pixel with a hard threshold. 
- Smooth rejection, which will accumulate more or less depending on the speed, allowing users to be able to control the amount of ghosting / noise they want. 

In both mode it's possible to take into account the movement of the reflected surface or/and the reflecting surface for the speed used for speed rejection. Disabling reflecting surface influence can help when the camera is moving with the same velocity as another object, to be able to reject the background but keeping what's doesn't move relative to the camera. 

What has been tested: 
- Objects moving at different speeds with a static camera
- Rotating objects not moving relative to the camera
- Moving rotating objects
- Still reject against different type of skies


https://user-images.githubusercontent.com/57442369/132487392-3a116d83-b327-4d3d-9017-be5f28cd9521.mp4

https://user-images.githubusercontent.com/57442369/132487399-f89d6d07-4dcd-494c-9f2c-dcd69a579258.mp4

https://user-images.githubusercontent.com/57442369/132487422-f7e1acd5-4df3-4601-a998-cfcb4a5383c5.mp4

In most cases this is only useful with very low camera/object movement. The scaler factor upper speed range will only work up to certain speed. With higher speed, accumulation will cause bigger problem anyway so this can't really be used. 

Also, we still get motion vectors from the skies but the sky distance is at camera far plane so if this distance is way further from the geometry it's difficult to find a common matching speed for close and far away objects because of parallax.

###
What's missing to be able to approve is: 
- Documentation
- Possibility to disable speed rejection as a whole (and let it be off by default to avoid breaking changes)
- Add ScreenSpaceReflectionSpeedRejection Debug Mode (really hard to find correct speeds threshold without) 
- Make sure the Roughness bias is "outside" the speed rejection enable bloc

remi-chapelain:
Before approving just want to poke @TheoWong-pixel to make sure that this is the best approach in terms of UI position and UX wise to do this. 

Little bit of context, this PR adds 7 new properties:
image
World Space Speed Rejection which the only one not advanced, when enabled this, the speed smooth rejection is shown, otherwise, it stays hidden. 

Then there's the 4 other speed rejection parameters. 

And lastly the roughness bias which is on its own and isn't concerned by speed rejection at all. 

Here's a quick video in actions. 
https://user-images.githubusercontent.com/57442369/133079362-eec0d932-984c-445d-85f6-d8db1e3f741d.mp4

https://user-images.githubusercontent.com/57442369/133078710-b1bffb56-9fac-4e84-b0e3-5f1c95e2ce63.mp4

com.unity.render-pipelines.high-definition/Documentation~/Override-Screen-Space-Reflection.md
```diff
@@ -45,6 +45,20 @@ HDRP uses the [Volume](Volumes.md) framework to calculate SSR, so to enable and
...
+| **World Space Speed Rejection**   | When enabled, speed from will be computed in world space to reject samples.<br/>This property only appears if you set **Algorithm** to **PBR Accumulation**. |
+| **Speed Rejection**               | Controls the likelihood history will be rejected based on the previous frame motion vectors of both the surface and the hit object.<br/>This property only appears if you set **Algorithm** to **PBR Accumulation**. |
+| **Speed Rejection Scaler Factor** | Controls the upper range of speed. The faster the objects or camera are moving, the higher this number should be.<br/>This property only appears if you set **Algorithm** to **PBR Accumulation**. |
```
remi-chapelain, line 50: And World Space Speed Rejection to True

remi-chapelain (approved):
Aside from the little documentation addition I requested and the UI/UX review on the inspector, I approve the changes ! Everything that was problematic in the previous comments has been fixed and some fine tuning control have been added to the (current) screen space rejection. ✔️

com.unity.render-pipelines.high-definition/Documentation~/Override-Screen-Space-Reflection.md
skhiat: Same for smooth rejection.

TheoWong-pixel:
> Before approving just want to poke @TheoWong-pixel to make sure that this is the best approach in terms of UI position and UX wise to do this.
> 
> Little bit of context, this PR adds 7 new properties:
> image
> **World Space Speed Rejection** which the only one not advanced, when enabled this, the **Speed Smooth Rejection** is shown, otherwise, it stays hidden.
> 
> Then there's the 4 other speed rejection parameters.
> 
> And lastly the roughness bias which is on its own and isn't concerned by speed rejection at all.
> 
> Here's a quick video in actions.
> https://user-images.githubusercontent.com/57442369/133078710-b1bffb56-9fac-4e84-b0e3-5f1c95e2ce63.mp4

Yes, your thinking is correct, thank you!

Vic-Cooper:
Updated docs. Please review for accuracy.

Vic-Cooper (approved):
lgtm :)

sebastienlagarde:
@soufianekhiat looks like there is a true error remaining in the DXR test:

Unhandled log message: '[Error] Render Graph Execution error at pass Render SSR (25)'. Use UnityEngine.TestTools.LogAssert.Expect

sebastienlagarde:
@soufianekhiat dxr still failed

anisunity:
Careful with the compute shader compiled size, that is a lot of kernel...

sebastienlagarde:
Any Update, DXR job still fail?

PR #3829, opened 2021-03-10
New Mip Bias API for material rendering

kecho:
### Purpose of this PR  ###
New API inside HDAdditionalCameraData.cs, which exposes mip bias for material passes. This will serve as a tool for our temporal upsamplers to reshape our image /  shift mip bias for proper reconstruction.

Changes are :
* New API in HDAdditionalCameraData, which propagates to constant _GlobalMipBias
* Hijacking of TEXTURE_2D_SAMPLE at the HDRP level (no need to touch the core package) and exchanging it with TEXTURE_2D_SAMPLE_BIAS
* Exposes a new setting in the debug panel that allows us to test / debug override mip bias.
image

* Adds a new test, to test this api on all platforms supported for HDRP.
* The test ensures that the mip override can work on a per view basis.
* The first and second view use a lit shader to draw Ender.
* The third view uses a shader graph to draw Ender.
image

---
### Info for reviewers
In order to get the changes suggested by @sebastienlagarde , I had to touch the platform hlsl. This means that there is another PR here:
https://github.cds.internal.unity3d.com/unity/com.unity.render-pipelines.nda/pull/21

Testing for nda platforms here (both URP and HDRP):
https://yamato.cds.internal.unity3d.com/jobs/339-ScriptableRenderPipelinePrivate/tree/graphics%252Fmipbias

### Testing status
* Manual tests on dx11 and dx12 (9910_GlobalMipBias), testing on shader graphs as well
* Test ran on all platforms

### Info for QA
Can you guys verify that:
* Debug setting works as expected on a scene (only on deferred / lit materials)
* Test at least in one console (ps4 to be sure) using the runtime debugger.

Note: for console we have: https://github.cds.internal.unity3d.com/unity/com.unity.render-pipelines.nda/pull/21


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
```diff
@@ -30,6 +30,13 @@ public partial class HDRenderPipeline
 
         DebugDisplaySettings m_DebugDisplaySettings = new DebugDisplaySettings();
 
+        public static bool allowDebugMaterialSamplerOverride = UnityEngine.SystemInfo.hasMaterialSamplerOverride;
```
jenniferd-unity, line 33: @sebastienlagarde do you know if this is also available in URP? should we let them know we added this feature since they are redoing also the debug menu (@arttu-peltonen is helping out actually)

jenniferd-unity: Shouldnt this boolean be moved to DebugDisplaySettings > DebugData ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
```diff
@@ -30,6 +30,13 @@ public partial class HDRenderPipeline
...
+        private static MaterialSamplerOverride s_MaterialDebugSamplerOverride = new MaterialSamplerOverride() { mipBias = 0.0f };
```
jenniferd-unity, line 34: I am not sure why we have a different workflow than the other debug Materials above?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
arttu-peltonen: > @sebastienlagarde do you know if this is also available in URP? should we let them know we added this feature since they are redoing also the debug menu (@arttu-peltonen is helping out actually)

There's no mip bias overrides in URP (nor was it available in builtin from what I can tell). Perhaps it's a useful feature to add later on? I think our most immediate focus is to achieve URP<->builtin feature parity.

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
```diff
@@ -1288,6 +1290,58 @@ void RegisterLightingDebug()
                     });
                 }
 
+                if (HDRenderPipeline.allowDebugMaterialSamplerOverride)
```
sebastienlagarde, line 1293: You should put this control in the Material section of the debug menu and not in the lighting section

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
sebastienlagarde: The feature (mip bias) is in builtin, so available for any pipeline.
Not sure that there is really a need to display it in URP (neither in HDRP as it is only here to check that the feature works, but otherwise in practice I don't really see a usage for it - maybe check mips of texture?) 
Anyway, it could be cross platform yeah if we really want.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
sebastienlagarde: yup, it should be in Material section, not in lighting

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
sebastienlagarde: https://ono.unity3d.com/unity/unity/pull-request/122268/_/graphics/mipbias

jenniferd-unity: i think Andre was suggesting it could be nice to add it to URP too - see https://unity.slack.com/archives/G077VATKP/p1615452059011200?thread_ts=1615398479.007700&cid=G077VATKP 
@sebastienlagarde should we add a jira for this?

sebastienlagarde: Dont' be confuse between the feature itself which works for all pipeline and this debug tool which is mainly to check if implementation works. Andre refer to having the feature.
It is very little work, so we can add it easily if needed, it is just that I don't see any usage in debug menu, neither for HDRP outside of validating it works :).

jenniferd-unity: ok so lets make sure we have the feature working for URP too and not only let UX know that URP interface will not support this for now.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
kecho: Will move this to material panel

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
kecho: will do

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
kecho: * I'll move the boolean to the debug settings.  Thanks for the feedback @jenniferd-unity . Keep it coming 👍 
* This can be easily added to the URP debug settings. I actually tried to, but I quickly discovered there was no debug menu. I did tested this on a user script, and works in URP as well. It will be up to you guys if you guys want it there once that debug menu is up and running. Should be easy to add.

arttu-peltonen: Yeah URP debug menu is still in a branch and hasn't landed. If somebody needs it later in URP, feel free to direct them to me and I can add it to the debug menu.

kecho: Sounds good. We are doing now discussions on the API, it hasnt been finalized :) But once its official and in the trunk ill ping you.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
kecho: This has been addressed

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
kecho: addressed

JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -302,6 +302,12 @@ internal void SetParentCamera(HDCamera parentHdCam)
         // XR multipass and instanced views are supported (see XRSystem)
         internal XRPass xr { get; private set; }
 
+        internal float GlobalMipBias { set; get; } = 0.0f;
```
JulienIgnace-Unity, line 305: Properties should be lower pascal case.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1867,6 +1867,9 @@ AOVRequestData aovRequest
                         CustomPassVolume.Update(hdCamera);
                 }
 
+                if (m_CurrentDebugDisplaySettings != null && m_CurrentDebugDisplaySettings.data.UseDebugGlobalMipBiasOverride())
```
JulienIgnace-Unity, line 1870: It feels like this should probably be done in HDCamera.Update().
You can access the current debug with hdrp.m_CurrentDebugDisplaySettings (which should probably be renamed :p )

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -156,6 +156,12 @@ TEXTURE2D(_PrevExposureTexture);
 
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesXR.cs.hlsl"
 
+#ifdef SAMPLE_TEXTURE2D
+#undef SAMPLE_TEXTURE2D
+#define SAMPLE_TEXTURE2D(textureName, samplerName, coord2) \
```
JulienIgnace-Unity, line 161: Is this what we want?
Unless I'm mistaken, doing that will cause all samples 2D to apply the bias, including internal render loop stuff. 
Thinking of stuff like color pyramid which has mips etc. There might not be many such textures but it could break things I guess.
The alternative is not super nice though. It means that we'd need to update all our hand written + SG shaders to use SAMPLE_TEXTURE2D_BIAS with the global bias instead?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
JulienIgnace-Unity: Or maybe we can just update the global CB with a non zero Global Mip Bias only around passes where we do DrawRenderers calls? 
That's quite a few additional update to the global CB but might not be that bad.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
kecho: Yes, so what's happening is that in the CPU (C# side) i reset the bias to 0 always, after material rendering (I have to double check that).
  
So color pyramid / other passes will remain with bias == 0.
As you said, the alternative is to be explicit into every shader. We can discuss this perhaps with Seb.

JulienIgnace-Unity: From what I see we update the global CB at the beginning of the camera render once with the current global mip bias and then only resets it at the end before posts. Am I missing something?

kecho: @JulienIgnace-Unity I have updated the review with the proper changes. Could you verify? I have only done this during Prepass (that is gbuffer laydown).
I will have to rework this on the actual DLSS change and verify mip bias against hte missing passes. This time around I rather just do it on a couple of passes before going wide.
Let me know.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -302,6 +302,10 @@ internal void SetParentCamera(HDCamera parentHdCam)
         // XR multipass and instanced views are supported (see XRSystem)
         internal XRPass xr { get; private set; }
 
+        internal float globalMipBias { set; get; } = 0.0f;
+
+        internal void SetGlobalMipBiasFromCamera() { globalMipBias = m_AdditionalCameraData == null ? 0.0f : m_AdditionalCameraData.materialMipBias; }
```
jenniferd-unity, line 307: is it valid to have m_AdditionalCameraData null in HDCamera? @RSlysz

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
jenniferd-unity: it would be worth it to add that type of info to the documentation btw. what is affected by the feature?

jenniferd-unity (changes requested):
I don't see the documentation yet. Otherwise code looks good to me.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
RSlysz: Hum HDCamera is a class that aggregate all data from Camera and Additional data to speed up frame computation in the pipeline.
I am not 100% sure here. 
@JulienIgnace-Unity  Any idea ?

RSlysz: IMO it should have been deal with null before and if null we should have use the default additional data here.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
kecho: I can add some documentation. Every texture sampler is affected. The only big difference is the fact that you can now pass a bias. If the bias is 0 though, you get the exact same behaviour.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
kecho: Removed the SetGlobalMipBiasFromCamera function, and instead moved all this to Update. We do have to do the null check there, because we could be dealing with a reflection probe camera (which might have additional camera as null)

kecho:
> For the documentation, it is a good point Changelog was missing but since you added public stuff, it could be worth to add Lewis or Vic as reviewer and update the following .md files:
> HDRP-Camera.md and Render-Pipeline-Debug-Window.md.

So I added documentation into Render-Pipeline-Debug-Window.md, but I think I shouldnt add into HDRP-Camera.md, because the mip bias I added is not exposed in the UI (intentionally), only programmatically.
Thus, having a <summary> tag in the HDAdditionalCameraData might suffice? If not where exactly should I add it in HDRP-Camera.md? I only see things there that are exposed in the inspector.
Added Lews for doc review as well. Let me know @jenniferd-unity !

jenniferd-unity:
> > For the documentation, it is a good point Changelog was missing but since you added public stuff, it could be worth to add Lewis or Vic as reviewer and update the following .md files:
> > HDRP-Camera.md and Render-Pipeline-Debug-Window.md.
> 
> So I added documentation into Render-Pipeline-Debug-Window.md, but I think I shouldnt add into HDRP-Camera.md, because the mip bias I added is not exposed in the UI (intentionally), only programmatically.
> Thus, having a 'summary' tag in the HDAdditionalCameraData might suffice? If not where exactly should I add it in HDRP-Camera.md? I only see things there that are exposed in the inspector.
> Added Lews for doc review as well. Let me know @jenniferd-unity !

Oh true no UI change for the Camera. So what I dont know is where could we let users know about what mipbias affects? I am also curious about how your PR works with the "Texture MipMap Stripping" option from the Player Window?

kecho:
> > > For the documentation, it is a good point Changelog was missing but since you added public stuff, it could be worth to add Lewis or Vic as reviewer and update the following .md files:
> > > HDRP-Camera.md and Render-Pipeline-Debug-Window.md.
> > 
> > 
> > So I added documentation into Render-Pipeline-Debug-Window.md, but I think I shouldnt add into HDRP-Camera.md, because the mip bias I added is not exposed in the UI (intentionally), only programmatically.
> > Thus, having a 'summary' tag in the HDAdditionalCameraData might suffice? If not where exactly should I add it in HDRP-Camera.md? I only see things there that are exposed in the inspector.
> > Added Lews for doc review as well. Let me know @jenniferd-unity !
> 
> Oh true no UI change for the Camera. So what I dont know is where could we let users know about what mipbias affects? I am also curious about how your PR works with the "Texture MipMap Stripping" option from the Player Window?

So not sure the intention is to give the users direct acccess to this feature, but I dont see any other way. The main purpose is for DLSS to control this internally, so I might even have to change the name of the variable in a later PR.
Keep in mind too, that Id like to include a test as well, thus why a user script has to access it (and why it has to be public). If you know of  better way to achieve these two goals let me know.
As far as the mip stripping setting, I think that just has to do with mip levels on textures content, which is different from this. This is purely mip selection at sample time / runtime. Hope that makes sense!

JordanL8 (approved):

TomasKiniulis (approved):
Checked Xbox and Editor template Deferred mode, mip map bias debug works perfectly well. 
Editor:

https://user-images.githubusercontent.com/50582134/115025209-bf4c5d00-9ec9-11eb-95af-4448e4969be2.mp4

Xbox:

https://user-images.githubusercontent.com/50582134/115025426-02a6cb80-9eca-11eb-8ee2-4624eaf6578f.mp4

JMargevics (approved):
image
Checked new exposed property via Debug Mode on PS4. Works as expected ✅

kecho:
In order to get the changes suggested by @sebastienlagarde , I had to touch the platform hlsl. This means that there is another PR here:
https://github.cds.internal.unity3d.com/unity/com.unity.render-pipelines.nda/pull/21

Testing for nda platforms here (both URP and HDRP):
https://yamato.cds.internal.unity3d.com/jobs/339-ScriptableRenderPipelinePrivate/tree/graphics%252Fmipbias

com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Debug-Window.md
```diff
@@ -96,19 +96,21 @@ The **display stats** panel is only visible in play mode and can be used to debu
...
+| **- Not A Pure Metal Color**                     | Use the color picker to select the color that the debugger displays if a pixel defined as metallic has a non-zero albedo value. The debugger only highlights these pixels if you enable the **True Metals** checkbox.<br/>This property only appears when you select **Diffuse Color** or **Metal or SpecularColor** from the **Material Validator** drop-down. |
+| **- Pure Metals**                                | Enable the checkbox to make the debugger highlight any pixels which Unity defines as metallic, but which have a non-zero albedo value. The debugger uses the **Not A Pure Metal Color** to highlight these pixels.<br/>This property only appears when you select **Diffuse Color** or **Metal or SpecularColor** from the **Material Validator** drop-down. |
+| **Override Global Material Texutre Mip Bias**    | *NOTE: feature still under development, for now it just works on deferred rendering passes* Enable the checkbox to override the mip bias of texture samplers in material shaders. Use the 'Debug Global Material Texture Mip Bias Value' to control the mip bias override. |
```
sebastienlagarde, line 112: >for now it just works on deferred rendering passes*
what do you mean? it is expected to work in forward as well. You can't add "NOTE" like this. 
"NOTE: feature still under development," is usually not something we write in doc AFAIK (cc @JordanL8 ). 

However we should mention that it is not compatible with virtual texturing

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -78,8 +78,16 @@ public partial class HDRenderPipeline
             lightingBuffers.diffuseLightingBuffer = CreateDiffuseLightingBuffer(m_RenderGraph, msaa);
             lightingBuffers.sssBuffer = CreateSSSBuffer(m_RenderGraph, msaa);
 
+            if (m_CurrentDebugDisplaySettings != null && m_CurrentDebugDisplaySettings.data.UseDebugGlobalMipBiasOverride())
```
sebastienlagarde, line 81: So this isn't enough, you also need to encompass the forward pass.
Then you need to update your test. 
The test should have at least the two type of material we support (i.e forward and deferred), but I will suggest the following list to have a good coverage: 
Lit.shader, shader graph lit, unlit.shader, shadergraph unlit, Axf.shader, ShaderGraph Fabric
+ a Lit Material receiving a Decal.shader and a lit Material receiving a shaderGraph decal.

Lastly you will also need to modifiy the code for cluster decal (and add it to the test, so in addition)
+ a Lit Material **Transparent** receiving a Decal.shader

For the code of cluster decal, please go into DecalUtilities.hlsl and add 
ComputeTextureLOD(...) + _GlobalMipBias


Note that we render normal map during depth prepass to be used by SSR / SSAO. This will be affected as well but I guess it is ok to not test result for those effect in this case.

sebastienlagarde: Lastly, but this can done in a different PR, we will need to handle virtual texturing. You can add a bias test in the runtime test either by modifying one of the existing VT test or adding a new one which highlight the blur. 

Thanks

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -156,6 +156,78 @@ TEXTURE2D(_PrevExposureTexture);
 
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesXR.cs.hlsl"
 
+// In HDRP, all material samplers have the possibility of having a mip bias.
```
sebastienlagarde, line 159: Let's add a define in ShaderConfig.hlsl to control if we enable those Macro or not. 
Also good to check with our performance test if we have any difference for our console test

sebastienlagarde:
We will also need to update the ShaderGraph texture sampling node to offer an option to not use the bias (i.e generate PLATFORM_SAMPLE_TEXTURE2D  instead of SAMPLE_TEXTURE2D )

jenniferd-unity (approved):
Approving - assuming Seb's comments will be tackled in a different PR

com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Debug-Window.md
JordanL8: Yeah, we don't do it all in caps https://unity-docs.gitbook.io/style-guide/standard-notes-and-information/notes-and-info-boxes. Also we document the current state of the feature and tend not to mention things like "feature still under development". I don't remember seeing this entry originally so I'll have a look and make suggestions now

JordanL8: Enable the checkbox to override the mipmap level bias of texture samplers in material shaders. Use the **Debug Global Material Texture Mip Bias Value** to control the mipmap level bias override.<br/>When using this feature, be aware of the following:<br/>&#8226; It only works in deferred mode.<br/>&#8226; It does not support virtual texturing.<br/>&#8226; It does not affect custom passes.

Markdown:
`Enable the checkbox to override the mipmap level bias of texture samplers in material shaders. Use the **Debug Global Material Texture Mip Bias Value** to control the mipmap level bias override.<br/>When using this feature, be aware of the following:<br/>&#8226; It only works in deferred mode.<br/>&#8226; It does not support virtual texturing.<br/>&#8226; It does not affect custom passes.`

JordanL8: I put only works in deferred mode as a placeholder, can you elaborate on the 
> affect the Material pass (i.e Decal, GBuffer and forward).
bit?

kecho: No more need for NOTE, as I have supported everything that needs support. @sebastienlagarde check the Apply/Reset functions I added surrounding the interesting passes. 

The final proposal for this doc line is then this one:
| **Override Global Material Texutre Mip Bias**    | Enable the checkbox to override the mipmap level bias of texture samplers in material shaders. Use the **Debug Global Material Texture Mip Bias Value** to control the mipmap level bias override.<br/>When using this feature, be aware of the following:<br/>&#8226; It only affects gbuffer, forward opaque, transparency and decal passes. <br/>&#8226; It does not affect virtual texturing sampling.<br/>&#8226; It does not affect custom passes. |

Let me know if this all works.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
kecho: Done, check the ShaderConfig.cs and let me know if that works. I tested it locally and works well too.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
kecho: This has been fixed!

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
```diff
@@ -235,6 +235,8 @@ void BindMotionVectorPassColorBuffers(in RenderGraphBuilder builder, in PrepassO
                 // At this point in forward all objects have been rendered to the prepass (depth/normal/motion vectors) so we can resolve them
                 ResolvePrepassBuffers(renderGraph, hdCamera, ref result);
 
+                ApplyCameraMipBias(hdCamera);
```
sebastienlagarde, line 238: During the prepass we also sample normal map (this is use for SSR and SSAO). I "guess" we should then do it for depth prepass as well, what do you think?

sebastienlagarde:
don't forget to handle the cluster decal case

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
kecho: > We will also need to update the ShaderGraph texture sampling node to offer an option to not use the bias (i.e generate PLATFORM_SAMPLE_TEXTURE2D instead of SAMPLE_TEXTURE2D )

@sebastienlagarde I have added a new checkbox for texture samplers:
image

The code should be on this commit:  https://github.com/Unity-Technologies/Graphics/pull/3829/commits/6108c57bb899afde845313e1a6d4ec6f92253063

@JordanL8 let me know your thoughts on the documentation (leaving a comment might be better for iteration purposes).

Will update the tests soon to also encompass these nodes

kecho: @cdxntchou has been added to the review for approval.

kecho:
> We will also need to update the ShaderGraph texture sampling node to offer an option to not use the bias (i.e generate PLATFORM_SAMPLE_TEXTURE2D instead of SAMPLE_TEXTURE2D )

@sebastienlagarde I have added a new checkbox for texture samplers:

image

The code should be on this commit: https://github.com/Unity-Technologies/Graphics/pull/3829/commits/6108c57bb899afde845313e1a6d4ec6f92253063

@JordanL8 let me know your thoughts on the documentation (leaving a comment might be better for iteration purposes).

Will update the tests soon to also encompass these nodes
@cdxntchou has been added to the reviewers so he can approve / advice on naming etc.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
kecho: Good point. Yes lets add it to the depth prepass, since this is technically part of material sampling (due to normal sampling as you point out)

kecho: Done!

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
kecho: image
Hey seb, everything looking good, except that decals do not seam to blend with forward. Wondering if there is something im missing? Or should it work out of the box. LMK if I should dig in a bit more (this happens irrelevant of my changes)

kecho: Ive highlighted in yello were the decal should be doing something to transparent.

com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Debug-Window.md
```diff
@@ -96,19 +96,21 @@ The **display stats** panel is only visible in play mode and can be used to debu
...
+| **- Pure Metals**                                | Enable the checkbox to make the debugger highlight any pixels which Unity defines as metallic, but which have a non-zero albedo value. The debugger uses the **Not A Pure Metal Color** to highlight these pixels.<br/>This property only appears when you select **Diffuse Color** or **Metal or SpecularColor** from the **Material Validator** drop-down. |
+| **Override Global Material Texutre Mip Bias**    | Enable the checkbox to override the mipmap level bias of texture samplers in material shaders. Use the **Debug Global Material Texture Mip Bias Value** to control the mipmap level bias override.<br/>When using this feature, be aware of the following:<br/>&#8226; It only affects gbuffer, forward opaque, transparency and decal passes. <br/>&#8226; It does not affect virtual texturing sampling.<br/>&#8226; It does not affect custom passes. |
```
JordanL8, line 112: Didn't notice this first time around: typo in Texutre Mip Bias

JordanL8 (changes requested):

JordanL8:
I can update the Shader Graph documentation. How about:
Disable Global Mip Bias: Indicates whether this Shader Graph ignores the **Override Global Material Texture Mip Bias** mode in the Render Pipeline Debug window.

cdxntchou (approved):
ShaderGraph code looks good

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DArrayNode.cs
```diff
@@ -37,6 +39,23 @@ public SampleTexture2DArrayNode()
             UpdateNodeAfterDeserialization();
         }
 
+        [SerializeField]
+        private bool m_DisableGlobalMipBias = false;
+
+        [ToggleControl("Disable Global Mip Bias")]
+        public ToggleData DisableGlobalMipBias
```
cdxntchou, line 46: This should probably be moved to the node inspector, rather than a first class control on the node, as it's a rather advanced control.  Would like to keep those a little more hidden to avoid confusing new users.

cdxntchou (changes requested):

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DArrayNode.cs
kecho: That sounds good, I will look into it.

JulienIgnace-Unity (approved):

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawerUtils.cs
```diff
@@ -154,5 +154,36 @@ internal static void AddCustomInterpolatorProperties(VisualElement parentElement
...
             }
         }
+
+        internal static void AddCustomCheckboxProperty(
+            VisualElement parentElement, AbstractMaterialNode node,
+            Action setNodesAsDirtyCallback, Action updateNodeViewsCallback,
+            String label,
+            String undoLabel,
+            Func<bool> GetterFn,
+            Func<bool, bool> SetterFn)
```
cdxntchou, line 164: Might make more sense as an Action<bool> ?

cdxntchou (approved):

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawerUtils.cs
kecho: Done!

sebastienlagarde (approved):

PR #3089, opened 2021-01-13
Improved shadow fade

lukaschod:
#### Also I decided to refactor shadow cascade GUI for HDRP and URP. For several reasons:

Partition sizes inconsistant

Last cascade had extra pixel at end inconsistantly

Any resizing was causing pixel jumping

Did not wanted to increase code duplication as extending URP cascade would resulted in very similar code of HDRP one.
I could not used HDRP solution as I had to move it to renderpipeline public core. HDRP solution was not written to be public code.

#### New shadow cascade GUI

Includes:
- Pixel perfect.
- Supports hovering.
- Supports focusing and keyboard arrow based value change.
- Support hidding border handles.
- New high quality snatch handle images.
- Gradient no longer recreates textures and uses preloaded texture.

(New urp upper, old urp bottom)

(New hdrp upper, old hdrp bottom)

## More info https://docs.google.com/document/d/1MAFk4AINTloYLgw8vBzhpah031LismehPQqVxiUhSx0.

---
### Testing status
- [x] Compare URP, HDRP and builtin shadow fade.
- [x] Test if it fixes original issues.
- [x] Test with multiple lights.
- [x] Ran gfx tests locally.
- [ ] Performance comparison. There is high chance for small regression on old mobiles as I had to add additional instruction. Sadly nothing can be done. However with baked shadow, I expect it to perform better as I got rid of one inline condition.
- [ ] Test if shadowmask correctly shadow blends to realtime shadow in all cases.
- [ ] Test if shadow fade does not start to happen in near plane, because we are no longer fading from camera, but from last cascade (I have special code path to take care of it).
- [ ] Test if new shadow cascade UI performs correctly in all cases.

---
### TODO
- [ ] Update shadow cascade images in urp docs
- [ ] Update shadow cascade images in hdrp docs
- [x] Update graphics test templates

---
### Comments to reviewers
I am mostly concerned about UI side if I did not screw up there with images, paths and so on (As rarely touch UI code).

PR #3831, opened 2021-03-11
[Universal] Make sure upgrading to Universal RP 2021 does not change Spot Light shadow resolutions

Wilfrid-Unity:
### Purpose of this PR

https://github.com/Unity-Technologies/Graphics/pull/2126 added to Universal RP support for custom shadow resolution for additional lights.
When a Universal RP project is upgraded to a Universal RP package containing those changes, its UniversalRenderPipeline Assets get new "Shadow Resolution Tiers" fields added with default values 256/512/1024 ; and its spot light shadows get assigned the **lowest** tier resolution.

The lower resolution might cause the visual quality of spot light shadows in a project to look worse after the upgrade.

This pull request ensures that after upgrade, spot light shadows will keep using the same resolution.
This is done by using the highest tier resolution as default.

Taking inspiration from@Verasl suggestion, during upgrade the 3 tier resolutions are also modified according to the additional light shadow atlas resolution set in the asset (respectively: full-res, half-res, quarter res if possible).

---
### Testing status

In a test scene based on UniversalRP template scene added with shadowed spot lights, I locally checked that the changes in this pull request effectively conserve spot light shadow resolutions after the upgrade to Unity 2021.

---
### Comments to reviewers

Effort tracked by internal ticket https://fogbugz.unity3d.com/f/cases/1316540

2021.1 backport pull-request: https://github.com/Unity-Technologies/Graphics/pull/3832


Verasl (approved):
LGTM

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -927,6 +931,22 @@ public void OnAfterDeserialize()
                 k_AssetVersion = 7;
             }
 
+            if (k_AssetVersion < 8)
```
Verasl, line 934: Seems there was a merge issue at some point and we have ended up with 2 version 8 upgraders, this one needs to be 9 and moved after the version 8 below it

Verasl: and further down in static void UpgradeAsset(UniversalRenderPipelineAsset asset) we need to bump the last check to be version 9 not 8 too

Verasl (changes requested):
Some minor changes to the upgrader, seems like a merge issue, we really need to update this upgrader, a bunch of 'ifs' is not the way :(

simon-engelbrecht-soerensen (approved):
Appears to be working.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Wilfrid-Unity: Thanks for noticing this. 

I fixed it in 1b2caec, can you please check the change?

Should I also bump the version numbers in [UpgradeAsset in the 2021.1](https://github.com/Unity-Technologies/Graphics/blob/e82aa3dc79a5e852bd88952530b9b46e83031ddd/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs#L967) backport PR?

Verasl: Fix looks good, for now we have to hold off back porting to 21.1 since it seems there is another upgrader that hasn't been back ported, this would cause some severe issues with users so I'm looking to get answers on what to do here.

Verasl (approved):
LGTM

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -379,7 +379,7 @@ void DrawShadowResolutionTierSettings()
                 {
                     var fieldSlot = new Rect(contentRect.x + pixelShift, contentRect.y, num - labelWidth, contentRect.height); // Define the rectangle for the field
                     int value = EditorGUI.DelayedIntField(fieldSlot, values[index].intValue);
-                    values[index].intValue = Mathf.Max(128, Mathf.NextPowerOfTwo(value));
+                    values[index].intValue = Mathf.Max(UniversalAdditionalLightData.AdditionalLightsShadowMinimumResolution, Mathf.NextPowerOfTwo(value));
```
nigeljw-unity, line 382: This isn't from this change, but we really need to get away from defaulting to power of two sizes. Most modern GPU architectures support non power of two sizes, and it is only really relevant for mipmaps and compressed formats.

nigeljw-unity: If it is a atlas fitting limitation, then it is a bit more understandable.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -964,10 +984,10 @@ static void UpgradeAsset(UniversalRenderPipelineAsset asset)
                 asset.k_AssetPreviousVersion = 5;
             }
 
-            if (asset.k_AssetPreviousVersion < 8)
+            if (asset.k_AssetPreviousVersion < 9)
             {
                 // The added feature was reverted, we keep this version to avoid breakage in case somebody already has version 7
```
nigeljw-unity, line 989: Nit: Comment needs updating. Should the comment be generic to say previous version instead of explicitly saying 7 or 8, etc.

nigeljw-unity (approved):
LGTM

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Wilfrid-Unity: Thanks for the info about 21.1. 
It seems that it would be better if both 2021.2 (this PR) and 2021.1 (backport PR) landed together at the same time?
Please notify when you know what more we should do for 21.1.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
Wilfrid-Unity: Yes, in the case of Additional Light shadows, it is a limitation of the current (simple) atlas fitting.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -186,6 +186,10 @@ public partial class UniversalRenderPipelineAsset : RenderPipelineAsset, ISerial
         internal const int k_ShadowCascadeMinCount = 1;
         internal const int k_ShadowCascadeMaxCount = 4;
 
+        public static readonly int AdditionalLightsDefaultShadowResolutionTierLow = 256;
```
eh-unity, line 189: I'm not sure if I like constant tiers. It doesn't scale very well. IMO user should just be able to put a number or 2^N for pot. It' depends on the platform and content what's suitable.
This is not a deal breaker though.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -934,6 +938,22 @@ public void OnAfterDeserialize()
                 k_AssetVersion = 8;
             }
 
+            if (k_AssetVersion < 9)
+            {
+                if (m_AdditionalLightsShadowResolutionTierHigh == AdditionalLightsDefaultShadowResolutionTierHigh &&
```
eh-unity, line 943: Maybe add a function "isCustomShadowResolution" or a local boolean at least.

eh-unity (approved):

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Wilfrid-Unity: Thanks for the suggestion, I added an intermediate variable in https://github.com/Unity-Technologies/Graphics/commit/82f939df2d54238754f8cb91df530258f5bf9212

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Wilfrid-Unity: @lukaschod 
Will https://github.com/Unity-Technologies/Graphics/pull/3089 be back-ported to branch https://github.com/Unity-Technologies/Graphics/tree/2021.1/staging ?

It seems that `UniversalRenderPipelineAsset.k_AssetVersion = 8` means "UniversalRenderPipelineAsset version containing improved shadow fade settings".

(With the current PR, we would have `UniversalRenderPipelineAsset.k_AssetVersion = 9` meaning "UniversalRenderPipelineAsset version containing auto-upgraded spot light shadow resolutions".)


Having "auto-upgraded spot light shadow resolutions" available in 2021.2 but not in 2021.1 might be surprising for users ("when upgrading project to 2021.1 nothing happened, but it changed when upgrading again to 2021.2!?"). Therefore I would like to land both pull requests #3831 and #3832 together.

@lukaschod 
If you plan to back-port #3089 to 2021.1, then #3832 has to wait (because asset update must happen in increasing version order).
If you don't plan to back-port #3089 to 2021.1, then on 2021.1 we can directly skip `UniversalRenderPipelineAsset.k_AssetVersion` from 7 to 9, and we will be able to land #3831 and #3832.
Please tell us, so that we can decide which path to take.

lukaschod: Nop I do not plan to backport #3089. It has some core changes and I do not want to introduce that risk.

lukaschod (approved):
looks good

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Wilfrid-Unity: Thanks a lot for the info Lukas.

I updated `UniversalRenderPipelineAsset.k_AssetVersion` in my [2021.1 backport pull request](https://github.com/Unity-Technologies/Graphics/pull/3832/commits/997afaea67392ccc4edcbca3012ecc07f909ecab) to match with this pull request (2021.2).

Wilfrid-Unity:
> Lots of failed tests. Master should be failing only Hybrid Tests, maybe if you merge master and rerun it will give a better idea.

Thanks for the info @phi-lira, I merged main branch and now there are no more failures, apart from Hybrid Tests (or Bokken test infrastructure issues not related to this pull request).


What about branch 2021.1/staging, are there some failures expected on that branch?

Wilfrid-Unity:
Everything looks OK on both 2021.2 and 2021.1, I'm merging the pull requests.

PR #3837, opened 2021-03-11
[HDRP] Depth of Field near blur improvements

pmavridis:
### Purpose of this PR
This PR improves the quality of physically based DoF.

Here is a reference test image, rendered with the Path Tracer in Unity:
image

**Before** this PR, physical DoF had issues with near DoF (as you see at the edges of the walkman):
image

**After** this PR the quality of the near DoF has been improved:
image

I'm also adding some missing graphics tests for the pb DoF mode, and also add tests for transparent materials with depth writes.

---
### Testing status
By flipping the images of the path traced and the improved DoF, I have verified that the DoF/CoC computations are correct. There are still differences in lighting and in areas where occlusion is important, but these are expected. 

I have tested the above scene and also the HDRP template. I'm still testing various smaller scenes to find edge cases.

More things to test:
- I have tested mostly the "use physical camera" mode, maybe we need more testing with the "Manual" mode to set near/far blur ranges
- I'm sure there will be edge cases where the algorithm does not work very well. We need to find them and see what can be improved.

---
### Comments to reviewers
I rewrote a large part of the DoFGather shader, incorporating some elements from the presentation "A Life of a Bokeh" from SIGGRAPH 2018. In particular:
- ring-based occlusion calculation (slide 27)
- mirror-based back ground filling  (slide 76)

When not using enough samples (<=4), the algorithm switches to a fast approximation (not based on this presentation) for people that prioritize speed.


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFCoCMinMax.compute
```diff
@@ -0,0 +1,64 @@
...
+}
+
+[numthreads(GROUP_RES, GROUP_RES, 1)]
+void KMainCoCMinMax(uint3 dispatchThreadId : SV_DispatchThreadID)
+{
+    UNITY_XR_ASSIGN_VIEW_INDEX(dispatchThreadId.z);
+
+    float4 minMaxTile = InitMinMaxTile();
+
+    for (uint j = 0; j < TILE_RES; j++)
```
FrancescoC-unity, line 51: Looks good but fwiw this loop can be heavily optimized by using either LDS and/or wave intrinsics where available. 

See the motion blur GenTile shader that is doing something similar (min/max on a tile)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFGather.compute
```diff
@@ -88,106 +118,293 @@ float2 PointInCircle(float angle)
...
+}
 
-    // Note: for the far-field, we don't need to search further than than the central CoC.
-    // If there is a larger CoC that overlaps the central pixel then it will have greater depth
-    float centerCoC = GetCoCRadius(posInputs.positionSS);
-    float maxRadius = abs(centerCoC);
+void DoFGatherFast(PositionInputs posInputs, int numSamples, float maxRadius, SampleData centerSample, out float4 outColor, out float outAlpha)
+{
+    AccumData accumData;
+    InitAccumData(accumData);
```
FrancescoC-unity, line 150: Can use ZERO_INITIALIZE like we do in most places to 0-init structs, but this is fine as well (mentioning if you want to keep code leaner)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFGather.compute
```diff
@@ -88,106 +118,293 @@ float2 PointInCircle(float angle)
...
+                float2 sampleTC = posInputs.positionSS + sampleRadius * PointInCircle(_sign[j] * (i + r1) * dAng);
+#else
+                float2 sampleTC = posInputs.positionSS + sampleRadius * PointInCircle(_sign[j] * r1 * PI);
+#endif
+                sampleData[j].color = GetColorSample(sampleTC, lod);
+                sampleData[j].CoC = GetCoCRadius(sampleTC);
+            }
+
+            float bias = 1.0 + r1;
+            float2 limit = float2(bias, 9 * bias);
```
FrancescoC-unity, line 189: What's the 9 here? worth giving it a name?

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFMinMaxDilate.compute
```diff
@@ -0,0 +1,49 @@
...
+    UNITY_XR_ASSIGN_VIEW_INDEX(dispatchThreadId.z);
+
+    float4 tileInfo = _InputTexture[COORD_TEXTURE2D_X(dispatchThreadId.xy)];
+
+    int2 startIndex = max((int2)dispatchThreadId.xy - int2(1, 1), int2(0, 0));
+    int2 endIndex =  min((int2)dispatchThreadId.xy + int2(1, 1), _ScreenSize.xy / TILE_RES - int2(1, 1));
+
+    for (int i = startIndex.x; i <= endIndex.x; i++)
+    {
+        for (int j = startIndex.y; j <= endIndex.y; j++)
```
FrancescoC-unity, line 41: Looks good but fwiw this loop can be heavily optimized by using either LDS and/or wave intrinsics where available. 

See the motion blur GenTile shader that is doing something similar (min/max on a tile)

FrancescoC-unity (approved):

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFMinMaxDilate.compute
pmavridis: True, for this PR my priority was the image quality, so I kept the naive implementation to be able to experiment easily. We can change this in a future PR. The size of this RT is rather small, so I think it is not a big overhead.

pmavridis:
Some changes to address the artifacts that we found. Most notably:
-Now layers in the near field are also using ring-based occlusion (with front-to-back blending, instead of back-to-front which is used in the far field)
- The bg and fg layer classification is now done per-tile 
- Bug fixes and code refactor/clean-up.

iM0ve (commented):
Review: **In Progress**

Still testing, I see the improvements but found some issues.

**Whats tested**

* On 3 HDRP projects (Real-world usage)
  * Spotlight tunnel
  * HDRP template
  * Amalienborg 
* Different game aspect ratios
* Various materials(transparent, opaque)
* Split view Game/Scene
* Adjusting parameters related to DoF (Aperature, Focal len, Focus distance) 

**Concerns:**
1) When using split-screen + Always refresh. On the branch PR we have significantly increased noise, it wasnt there before the PR.

2) Resolution setting is affecting Physically Based DOF, when it shouldn't.

https://user-images.githubusercontent.com/36502659/111495336-f05d3480-8747-11eb-9e3b-7b5332858052.mp4



https://user-images.githubusercontent.com/36502659/111339312-567e8480-8680-11eb-8427-0e4c64dba442.mp4

pmavridis:
> 1. When using split-screen + Always refresh. On the branch PR we have significantly increased noise, it wasnt there before the 

Good catch, my last commit has fixed this issue. The temporal accumulation was not working properly in this case.

sebastienlagarde:
Waiting for @iM0ve  approval

pmavridis:
Still need to screenshots for the graphics tests on all platforms. I just launched a yamato abv, hopefully we will not need any additional code revisions.

iM0ve:
Still investigating flickering issues

https://user-images.githubusercontent.com/36502659/111478308-a0c33c80-8738-11eb-8c04-d28c971326a6.mp4

pmavridis:
> Still investigating flickering issues

This was caused by having a second game camera without TAA in the scene. My last commit should probably fix this issue.
One extra thing that I have tested is that when switching DoF modes (physical on/off) there is no flashing. 
(the code that was creating this bug was there to prevent the flashing, but was working properly with only one game camera).

pmavridis:
Switching to Draft again, because I will push a few more commits, to avoid triggering Yamato all the time.
Next commit fixes an issue where choosing a small "Resolution scale" was causing excessive noise, as noted by @iM0ve.

iM0ve (approved):
Updated the review above

pmavridis:
I'm grabbing new reference images and after that it will be ready for merge.

pmavridis:
The XR test gives a black image, need to investigate on that before we merge.

tlaedre:
It's a bit confusing that not only does the source branch for this PR seems to not be deleted, it's also out of sync with master. Kinda looks like an older version of the branch was re-pushed?

https://github.com/Unity-Technologies/Graphics/tree/HDRP/dof_near_blur

pmavridis:
> for this PR seem to not be deleted, it's also out of sync with master. Kinda looks like an older version of the branch was re-pushed?

Sorry about the confusion, I re-pushed the branch when preparing the back-ports for 10.x and 2021.1. Deleting again since everything is merged into master.

PR #3840, opened 2021-03-11
RThandleSystem no longer fixes MSAA sample count

JulienIgnace-Unity:
### Purpose of this PR
The goal of this PR is to remove the constraint that all RTHandles need to be created with the same number of MSAA samples.
Updated HDRP code to reflect the change:
- The MSAA sample count in the HDRP asset is no longer a "capability". It's now only used as a default value (see frame setting change bellow).
- Changed the "MSAA Within Forward" toggle in the frame settings to a dropdown with MSAA sample count. This allows users to choose msaa sample rate per camera now
- In the aforementioned frame setting, a new entry was added "From HDRP Asset". If chosen on a camera, this will use the value set in the HDRP asset.
- In HDRP code, HDCamera now contains "msaaEnabled" and "msaaSamples" properties. Only those should be used to query SMAA state within the code as their correct value is "resolved" here (in case it's taking the value from HDRP asset).

---
### Testing status
- Ran automated tests locally and on Yamato
- Tested the frame setting by loading a prior scene with MSAA and switched to the various modes on the camera.
- Tested the fallback to HDRP asset.
- Did not test migration of reflection probes probes.
A few things should be further tested:
- Migration of existing data (camera and probes). Formerly enabled MSAA should fallback to "From HDRP Asset"
- New frame settings: Test overriding, different values, from HDRP Asset etc to make sure the behavior is consistent (may have missed some combinations with deferred/forward etc)


JordanL8 (approved):

FrancescoC-unity (approved):

com.unity.render-pipelines.core/Documentation~/rthandle-system-using.md
```diff
@@ -29,7 +27,7 @@ The code examples in the rest of this page use the default global instance.
 Before rendering with a Camera, you need to set the resolution the RTHandle system uses as a reference size. To do so, call the `SetReferenceSize` function.
 
 ```c#
-RTHandles.SetReferenceSize(width, hight, msaaSamples);
+RTHandles.SetReferenceSize(width, hight);
```
0lento, line 30: small typo left over from original doc here ( hight vs height )

JulienIgnace-Unity: Good catch, will fix :)

com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
```diff
@@ -287,14 +287,14 @@ internal bool StripShader(HDRenderPipelineAsset hdAsset, ComputeShader shader, s
             if (inputData.shaderKeywordSet.IsEnabled(m_ScreenSpaceShadowOFFKeywords) && shadowInitParams.supportScreenSpaceShadows)
                 return true;
 
-            if (inputData.shaderKeywordSet.IsEnabled(m_ScreenSpaceShadowONKeywords) && !shadowInitParams.supportScreenSpaceShadows)
-                return true;
-
-            if (inputData.shaderKeywordSet.IsEnabled(m_MSAA) && !hdAsset.currentPlatformRenderPipelineSettings.supportMSAA)
+            if (inputData.shaderKeywordSet.IsEnabled(m_MSAA) && (hdAsset.currentPlatformRenderPipelineSettings.supportedLitShaderMode == RenderPipelineSettings.SupportedLitShaderMode.DeferredOnly))
```
FrancescoC-unity, line 290: Both this and https://github.com/Unity-Technologies/Graphics/pull/3840/files#diff-2e328001603088c74758f8d9ccc9157aae3f8c220f0c0786af92db676ebad710R80 will be stripped less often now, do we have data how many are we not stripping now with forward?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineAsset.Migration.cs
```diff
@@ -136,6 +137,12 @@ enum Version
                 FrameSettings.MigrateVirtualTexturing(ref data.m_RenderingPathDefaultCameraFrameSettings);
                 FrameSettings.MigrateVirtualTexturing(ref data.m_RenderingPathDefaultBakedOrCustomReflectionFrameSettings);
                 FrameSettings.MigrateVirtualTexturing(ref data.m_RenderingPathDefaultRealtimeReflectionFrameSettings);
+            }),
+            MigrationStep.New(Version.UpdateMSAA, (HDRenderPipelineAsset data) =>
+            {
+                FrameSettings.MigrateMSAA(ref data.m_RenderingPathDefaultCameraFrameSettings);
+                FrameSettings.MigrateMSAA(ref data.m_RenderingPathDefaultBakedOrCustomReflectionFrameSettings);
+                FrameSettings.MigrateMSAA(ref data.m_RenderingPathDefaultRealtimeReflectionFrameSettings);
```
RSlysz, line 145: So if I get it well, if it was true, now it get FromHDRPAsset. But this is the HDRPAsset. I'm confused.
Should it be MSAA2x here?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.Migration.cs
```diff
@@ -136,6 +136,7 @@ class ObsoleteFrameSettings
     public partial struct FrameSettings
     {
 #pragma warning disable 618 // Type or member is obsolete
+#pragma warning disable 612
```
RSlysz, line 139: What kind of warning it is? I think it is usefull to keep a reminder of what was disabled in comment

FrancescoC-unity:
Mostly LGTM, I would just add in doc that chosing different MSAA samples for different cameras at the moment causes a lot of extra memory cost.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -100,10 +117,13 @@ public enum FrameSettingsField
...
         ClearGBuffers = 5,
         /// <summary>When enabled, Cameras using these Frame Settings calculate MSAA when they render the Scene. Set Lit Shader Mode to Forward to access this option.</summary>
-        [FrameSettingsField(0, displayedName: "MSAA within Forward", negativeDependencies: new[] { LitShaderMode }, customOrderInGroup: 3, tooltip: "When enabled, Cameras using these Frame Settings calculate MSAA when they render the Scene. Set Lit Shader Mode to Forward to access this option.")]
+        [Obsolete]
         MSAA = 31,
+        /// <summary>Specify the level of MSAA used when rendering the Scene. Set Lit Shader Mode to Forward to access this option.</summary>
+        [FrameSettingsField(0, displayedName: "MSAA", type: FrameSettingsFieldAttribute.DisplayType.Others, targetType: typeof(MSAAMode), negativeDependencies: new[] { LitShaderMode }, customOrderInGroup: 3, tooltip: "Specifies the MSAA mode for Cameras using these Frame Settings.")]
+        MSAAMode = 4,
```
RSlysz, line 124: Can't we reuse the same bit? I don't see obvious reason to change it :/ This way we will keep after migration "ok this camera was overriding MSAA"

If it change, you must also migrate the override flag to the new bit

RSlysz (changes requested):
Throughly check if the migration is correct in hdrpAsset. It seams strange to set things to set FromHDRPAsset in them

Also either keep same bit or also migrate the override checkbox.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.Migration.cs
JulienIgnace-Unity: I think it was for obsolete property (which was not covered by the other warning)

RSlysz (approved):

TomasKiniulis (changes requested):
Tested various combinations of Forward/Deferred rendering and MSAA settings on:
- Default Settings in Camera, Realtime Reflection, Baked or Custom Reflection
- Custom Frame Settings of Camera, Planar Probes, Baked Reflection Probes
All combinations seem to work correctly.

However I can't seem to get Migration working. Checking with Julien. Enabling MSAA in HDRP asset gives msaa errors and breaks rendering while Default Settings and Custom Frame Settings Input Fields are empty.
image

TomasKiniulis (approved):
Migration issue is sorted out now. Tested with two HDRP projects one with asset set to Deferred other with Forward and MSAA enabled. Both cases are correctly set to "From HDRP Asset".
After checking with Julien note that on latest master there is an issue where MSAA gets reenabled after reopening the project so migration is still set to "From HDRP Asset", while testing from older master commits it gets set to "None" correctly.

PR #3841, opened 2021-03-11
Disable filtering and denoising in baked GI tests

KEngelstoft:
### Purpose of this PR
Disable lightmap filtering and denoising on scenes 2104, 2106, 2107, 2108, 2109, 2403. 2404, 2505 to increase test determinism. Before, the the tests would start fail the denoiser gets upgraded .

### Testing status
Yamato is green.


sebastienlagarde:
Note: I merged master in this PR because I have updated a screenshots on master, so it will avoid to get conflict when we will merge this PR. Thanks for helping with this!

sebastienlagarde (approved):

PR #3846, opened 2021-03-11
Fix lightmaps in first room and removed SSR on transparent for glass

pierre-unity:
### Purpose of this PR
This PR fixes 2 issues.

1. Rebake the lightmaps: they were broken in the first room after a change to the fbx importer in 2021.2.0a6.
image

2. Remove SSR on transparent for 2 materials, to avoid transparent objects behind glass from being culled, such as butterflies in the tree cage.
image

As long as this PR is only for 2021.2, it will not cause any problem. If we were to backport these changes in 2020.2, it will break some lightmap UVs in 2020.2.

---
### Comments to reviewers
It's a very safe change, only reflection probes, lightmaps and 2 materials are updated. Checking in a very recent 2021.2 build will be enough.

However, if you check it with 2021.1 and 2020.2, you will then encounter the same broken lightmap issues in the first room:
image


sebastienlagarde (approved):

remi-chapelain (approved):
Looks good to me ! ✔️

PR #3847, opened 2021-03-11
Fix issue with APV normal bias not being correct until profile GUI is shown

FrancescoC-unity:
Due to some faulty logic on the normal profile update, the value was never set until the Inspector GUI of the probe reference volume profile was displayed. 

This PR forces the update of constants upon initialization which fixes the issue.


FrancescoC-unity:
Potentially found a small issue, gonna fix up.

alelievr (approved):

sebastienlagarde (approved):

PR #3849, opened 2021-03-12
[10.x][2020.3][SG] Update the SG package reference to Searcher to 4.3.2

Nightmask3:
---
### Purpose of this PR

Updates the Shader Graph's package reference to the Searcher package from 4.3.1 to 4.3.2, and in doing so fixes this bug for 2020.3:

https://fogbugz.unity3d.com/f/cases/1302267/

---
### Testing status

Verified that the bug is actually fixed. Searcher no longer collapses inadvertently when a category or drop down arrow is clicked


LandonTownsendUnity:
Do we need the PR to the editor to land before this works? When we land that, will we need to change the minimum editor version in SRP?

LandonTownsendUnity (approved):

Nightmask3:
> Do we need the PR to the editor to land before this works? When we land that, will we need to change the minimum editor version in SRP?

Yes the 20.3 Ono PR would need to land but Thierry indicated that all of them have been merged now. (for master and 21.1 as well) so we're good to go. 

Regarding the minimum editor version, that's a good question, I don't believe so as the packages aren't really tied to the editor versioning, but @marctem might know better.

marctem (approved):

PR #3790, opened 2021-03-07
[2021.2] [ShaderGraph] [BugFix 1319066] Fix for master node preview after deleting blocks

cdxntchou:
### Purpose of this PR
Fix for bug:  https://fogbugz.unity3d.com/f/cases/1319066/

PR Links:
2021.2: https://github.com/Unity-Technologies/Graphics/pull/3790
2021.1: https://github.com/Unity-Technologies/Graphics/pull/3807
2020.3: https://github.com/Unity-Technologies/Graphics/pull/3801

Address regression -- changes to property gathering made it so master preview will fail to compile if a block node is deleted.
This PR puts back the code that replaces missing blocks with a temporary default block.

This also fixes a related issue (that were actually pre-existing the property gathering change):
* Opening a graph with missing property blocks results in the preview not using the proper default values for that missing block

Another discovered issue, that is similar but caused by a different change, was moved to a separate bug:
https://fogbugz.unity3d.com/f/cases/1320695/

---
### Testing status

- [x] Tested deleting a block (preview and final shader) - URP
- [ ] Tested deleting all the blocks (preview and final shader) - URP -- this seems to be a separate bug introduced by custom interpolators -- creating a new bug for Esme to fix.  https://fogbugz.unity3d.com/f/cases/1320695/
- [x] Tested undo/redo deletion of a block - URP
- [x] Tested re-population of the blocks via subtarget selection - URP
- [x] Checked that generated code for temporary block properties is correct in preview and final shader - URP
- [x] Tested that deleting a block, then re-opening the graph results in preview showing default value (fixed a pre-existing issue here)
- [x] All of above for HDRP
- [x] Tested with Targets that use dynamic active block set (i.e. sprite unlit)

Yamato:

ShaderGraph PR Job: 🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252F1319066/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/5729875/job/pipeline

HDRP PR Job: 🟢 -- failures match master failures
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252F1319066/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/5730571/job/pipeline

Failures:
HDRP on Linux_Vulkan_playmode_cache_mono_Linear on version trunk -- failed same way on same 1 test
[...]

PR #3850, opened 2021-03-12
[ShaderGraph] [BugFix 1320695] Fix for Master Preview in URP when only Custom Interpolators are present

esmelusina:
### Purpose of this PR
#3790 addresses this bug, https://fogbugz.unity3d.com/f/cases/1319066/, but Custom Interpolators are a bit of a special case here.

This PR addresses a related bug described here https://fogbugz.unity3d.com/f/cases/1320695/.

If only custom interpolators are present, certain aspects of vertex stage generation would be skipped. This prevents that from happening. It also adds preprocessors to the descriptors for custom interpolators. Note that this is also only an issue with URP.

---
### Testing status
- [x] For each URP shader subtarget, add a custom interpolator (ensuring that it has an input edge and a node linking to a fragment port). Delete all other vertex blocks. No errors should occur.

### DIFF
--- a/com.unity.shadergraph/Editor/Generation/Processors/CustomInterpolatorUtils.cs
+++ b/com.unity.shadergraph/Editor/Generation/Processors/CustomInterpolatorUtils.cs
@@ -61,7 +61,7 @@ namespace UnityEditor.ShaderGraph
             internal bool isFunc => src != null && dst != null && name != null && splice != null;
             internal bool isDefine => define != null && splice != null && src == null && dst == null & name == null;
             internal bool isValid => isDefine || isBlock || isStruct || isFunc;
-            internal bool hasPreprocessor => String.IsNullOrEmpty(preprocessor);
+            internal bool hasPreprocessor => !String.IsNullOrEmpty(preprocessor);
 
             internal static Descriptor MakeFunc(string splice, string name, string dstType, string srcType, string define = "", string preprocessor = "") => new Descriptor { splice = splice, name = name, dst = dstType, src = srcType, define = define, preprocessor = preprocessor };
             internal static Descriptor MakeStruct(string splice, string name, string define = "", string preprocessor = "") => new Descriptor { splice = splice, name = name, define = define, preprocessor = preprocessor };


com.unity.shadergraph/Editor/Generation/Processors/CustomInterpolatorUtils.cs
```diff
@@ -54,16 +54,19 @@ internal struct Descriptor
...
             internal string define;   // defined for client code to indicate we're live.
             internal string splice;   // splice location, prefer use something from the list.
+            internal string preprocessor;
 
             internal bool isBlock => src != null && dst != null && name == null && splice != null;
             internal bool isStruct => src == null && dst == null && name != null && splice != null;
             internal bool isFunc => src != null && dst != null && name != null && splice != null;
             internal bool isDefine => define != null && splice != null && src == null && dst == null & name == null;
+            internal bool isValid => isDefine || isBlock || isStruct || isFunc;
+            internal bool hasPreprocessor => preprocessor != null && preprocessor != "";
```
joshua-davis, line 64: nit: can be simplified to `string.IsNullOrEmpty`

com.unity.shadergraph/Editor/Generation/Processors/CustomInterpolatorUtils.cs
```diff
@@ -54,16 +54,19 @@ internal struct Descriptor
...
 
-            internal static Descriptor MakeFunc(string splice, string name, string dstType, string srcType, string define = "") => new Descriptor { splice = splice, name = name, dst = dstType, src = srcType, define = define };
-            internal static Descriptor MakeStruct(string splice, string name, string define = "") => new Descriptor { splice = splice, name = name, define = define };
-            internal static Descriptor MakeBlock(string splice, string dst, string src) => new Descriptor { splice = splice, dst = dst, src = src };
-            internal static Descriptor MakeDefine(string splice, string define) => new Descriptor { splice = splice, define = define };
+            internal static Descriptor MakeFunc(string splice, string name, string dstType, string srcType, string define = "", string preprocessor = "") => new Descriptor { splice = splice, name = name, dst = dstType, src = srcType, define = define, preprocessor = preprocessor };
```
joshua-davis, line 66: Is it a problem that all of these share the same defines?

cdxntchou (approved):

com.unity.shadergraph/Editor/Generation/Processors/CustomInterpolatorUtils.cs
esmelusina: Shouldn't be a proble, what do you mean?

joshua-davis: I wasn't sure if it could be problem or not. Just wanted to ask you if it could be. Basically wasn't sure if different defines might be needed for blocks, structs, defines, or functions.

esmelusina: Oh sure-- uh... yea, here's a bit more explanation than what you're probably looking for--

Each descriptor is pseudo-typed based on the static construction function used. A descriptor instance does not represent more than one of these types at a time. Even then, the define is used the same way in all cases-- it's appended at the end of the code generated for the descriptor so that subsequent template code can branch off of it (primarily for cases where nothing is generated).

Quick example would be to look at Varyings.hlsl in URP, you can see the preprocessor wrapped call for copying custom interpolators into the VertexDescription object:
```
#if defined(CUSTOMINTERPOLATOR_VARYPASSTHROUGH_FUNC)
    CustomInterpolatorPassThroughFunc(output, vertexDescription);
#endif
```

The define comes from UniversalTarget.cs, which describes how and where the above function gets generated. The define is used to allow the callsite to guard itself.
```
CustomInterpSubGen.Descriptor.MakeFunc(CustomInterpSubGen.Splice.k_splicePreSurface, "CustomInterpolatorPassThroughFunc", "Varyings", "VertexDescription", "CUSTOMINTERPOLATOR_VARYPASSTHROUGH_FUNC")
```

phi-lira (approved):

erikabar (approved):
Developer tested their change, approving

com.unity.shadergraph/Editor/Generation/Processors/CustomInterpolatorUtils.cs
joshua-davis: Oh yeah...it's effectively a static constructor, missed that. Makes sense!

joshua-davis (approved):

esmelusina:
A trunk change has left some of the tests misbehaving. The results here are consistent with current master and the actual testing impact this change should be safe. I'm going to go ahead and merge this after the next set of tests run.

PR #3862, opened 2021-03-12
[2021.2] [ShaderGraph] Adding "regenerate" button option next to the "View Generated Shader" to simplify debugging

cdxntchou:
### Purpose of this PR
Allows user to reduce iteration time when viewing the generated shader repeatedly, by not launching another copy of the editor.

---
### Testing status
- [x] Tested only the "View Generated Shader" button shows up initially, and it properly generates the shader and launches editor
- [x] Verified that after the file exists, the "Regenerate" button shows up
- [x] Verified that clicking "Regenerate" will update the file on disk, but not launch an editor
- [x] Verified that clicking "View Generated Shader" while "Regenerate" is shown will both regenerate the shader and launch the editor

Yamato tests:

ShaderGraph PR Job:  🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Fregenerate-button/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/5870416/job

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Importers/ShaderGraphImporterEditor.cs
+++ b/com.unity.shadergraph/Editor/Importers/ShaderGraphImporterEditor.cs
@@ -59,15 +59,33 @@ namespace UnityEditor.ShaderGraph
                 Debug.Assert(importer != null, "importer != null");
                 ShowGraphEditWindow(importer.assetPath);
             }
-            if (GUILayout.Button("View Generated Shader"))
+            using (var horizontalScope = new GUILayout.HorizontalScope("box"))
             {
                 AssetImporter importer = target as AssetImporter;
                 string assetName = Path.GetFileNameWithoutExtension(importer.assetPath);
                 string path = String.Format("Temp/GeneratedFromGraph-{0}.shader", assetName.Replace(" ", ""));
+                bool alreadyExists = File.Exists(path);
+                bool update = false;
+                bool open = false;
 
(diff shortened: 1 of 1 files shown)


Nightmask3 (approved):
Tested it out and verified everything listed in the description functions as described.

Also tried deleting the generated shader file out from under the editor and it switched back to only allowing to "View Generated Shader" and removed the "Regenerate" until that button was pressed again, i.e. it functions as expected and handles that case properly.

joshua-davis (approved):
Functionality and code looks good. Would be nice if it could also re-focus the editor somehow, especially if the script was in another tab in your editor, but I think that's the whole problem with the `View Generated Shader` button.

cdxntchou:
> Functionality and code looks good. Would be nice if it could also re-focus the editor somehow, especially if the script was in another tab in your editor, but I think that's the whole problem with the `View Generated Shader` button.

Yeah, I noticed that opening .cs files won't launch another copy of the editor, simply re-focuses the existing editor, but I'm not sure how that was accomplished.

jessebarker (approved):
LGTM

LandonTownsendUnity (approved):
Can't believe I never realized how much I needed this. Developer testing is sufficient.

ValGrimm-U3D:
Related doc ticket (not to block PR, just informational): https://unity3d.atlassian.net/browse/DOC-3032

PR #3864, opened 2021-03-15
Reduced the maximal number of bounces for both RTGI and RTR (case 1318876).

anisunity:
https://fogbugz.unity3d.com/f/cases/1318876/

On QA's request i reduce the maximal number of bounces as the effects become unsuable at higher bounce count. Note that this doesn't change the values on the volume component that were previously created (the max/min values are serialized in the component).

Testing status
The change is trivial, nothing to test.


remi-chapelain (approved):
LGTM ✔️ 
Quick side note : [those values are not serialized anymore](https://github.com/Unity-Technologies/Graphics/pull/3629) so the change should upgrade automatically without reseting / removing the override

sebastienlagarde (approved):

PR #3865, opened 2021-03-15
Fixed error message when having MSAA and Screen Space Shadows (case 1318698).

anisunity:
https://fogbugz.unity3d.com/f/cases/1318698/
In the current documentation Screen space shadows and MSAA are marked to be not compatible, but nothing in the code was preventing both things to run at the same time (which lead to either visual artifacts at "super sampled" pixels or errors when trying to bind MSAA targets in non MSAA UAVs in compute shaders). This PR overrides the frame settings the same way it is currently done for SSR and other similar incompatible effects.

Testings status
Tested the dataset shared in the fogbugz, no more erros and the shadow displays correctly.


remi-chapelain (approved):
All good ! ✔️ 

I'll link this [issue here](https://fogbugz.unity3d.com/f/cases/1318693/), since it's related to how do we inform users that theses feature will be disabled if MSAA is on.

sebastienlagarde (approved):

PR #3872, opened 2021-03-15
Changed the behavior of the clear coat and SSR/RTR for the stack lit to mimic the Lit's behavior (case 1320154).

anisunity:
https://fogbugz.unity3d.com/f/cases/1320154/
An inconsistency was  found between the StackLit's CC and the Lit's CC. This unifies the behavior between the two. Some tests may fail and will need to have their screenshots updates. This changes which normal/smoothness is outputed by the stacklit based on the value of the coat mask (not only the checkbox) and the way the result of the SSR is used for the lighting.

Testing status
Made sure the behavior is the same on the dataset shared on the fogbug.
Updated the screenshot on the DXR test.


remi-chapelain (commented):
Issue fixed ! ✔️ 
However, launched HDRP tests and no stack lit tests failed. 
I've modifed one to include screen space reflection to have that case covered to prevent regression. Will update references image when yamato will fail too ! 

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-1320154/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/5823373/job

sebastienlagarde (approved):

remi-chapelain:
Updated the screenshots, relaunched yamato.

PR #3873, opened 2021-03-15
Fixed Default LookDev Profile being in the package instead of the project

JulienIgnace-Unity:
### Purpose of this PR
Currently when creating a new project with HDRP, the default LookDev volume profile referenced in the default HDRP asset is located in the package folder. This is not good because it's not editable.
This PR changes that so it's copied in the asset folder like the default volume profile through the HD Wizard UI (a new check should appear regarding the default lookdev volume profile).
Also updated the hd template to fix this.

This fixes this issue: https://fogbugz.unity3d.com/f/cases/1317821/

---
### Testing status
Tested with a new project in which I imported HDRP and used the wizard to fix everything and checked that the profile was in the right place.


RSlysz (approved):

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
```diff
@@ -493,22 +510,36 @@ void FixDefaultVolumeProfileAssigned(bool fromAsyncUnused)
             if (!IsHdrpAssetUsedCorrect())
                 FixHdrpAssetUsed(fromAsync: false);
 
-            var hdrpAsset = HDRenderPipeline.currentAsset;
+            var hdrpAsset = HDRenderPipeline.defaultAsset;
```
sebastienlagarde, line 513: @jenniferd-unity  this will have an impact on your PR, rights?

jenniferd-unity: i already fixed that in my PR actually :) but i ll make sure i dont break it.

jenniferd-unity (approved):

TomasKiniulis (approved):
Looks solid! Checked on new empty project with HDRP and Wizard ran, adding any Raytracing Overrides does not cause any exceptions anymore.

PR #3874, opened 2021-03-15
Fix SSR for 4K Render Target

skhiat:
Fix SSR for 4K Render Target.

Before:
Unity_WpfxV9aobk

After:
Unity_gMyQEKJEVp

TODO check;
* edge cases
* With large or small difference between far-near plane


sebastienlagarde (approved):

remi-chapelain (approved):
Has been pair tested with @skhiat and that indeed fix 99% of the issue. ✔️ 
Increasing the epsilon can help with the remaining percentage but that will introduce more artefacts in some cases too. 

However, even if it fixes it, what worries me a bit is that we found that the probability that it occurs is greater on the right of the screen that on the left and have no idea why.. so might come back someday !

PR #3876, opened 2021-03-15
Fixed Nans happening in RTR when the history render target is bigger than the current viewport (case 1321139).

anisunity:
https://fogbugz.unity3d.com/f/cases/1321139/
Nans where happening after the split into two reflections denoisers. This PR fixes it.

Testing status
The issue doesn't repro anymore on the shared dataset.


sebastienlagarde (approved):

remi-chapelain (approved):
Fix has already been discussed and I thinking necessary testing has already been done ✔️

PR #3879, opened 2021-03-15
Added tooltips for features being disabled because of MSAA/Raytracing

JulienIgnace-Unity:
### Purpose of this PR
Added tooltips for features being disabled because of MSAA/Raytracing.
We decided to add tooltip in the frame settings because this is the central place where features are enabled for a camera and as such have better visibility. We can't add warnings/messages in the volume overrides themselves as we don't have a proper context when editing them (is MSAA enabled? etc).

This adresses two cases:
https://fogbugz.unity3d.com/f/cases/1318706/
https://fogbugz.unity3d.com/f/cases/1318693/

---
### Testing status
Checked that the tooltips were correct in the editor.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -143,6 +143,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Virtual Texturing Resolver now performs RTHandle resize logic in HDRP instead of in core Unity
 - Cached the base types of Volume Manager to improve memory and cpu usage.
 - Reduced the maximal number of bounces for both RTGI and RTR (case 1318876).
+- Added tooltips to warn when some features may be disabled depending on MSAA/raytracing usage.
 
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

remi-chapelain (commented):
I like that change and it's still better than previously but I have the feeling that users should also be informed when enabling the features as a whole in the HDRP asset. 

Since you didn't mention it in the description, has this also been considered ? If yes, what's the reason for not doing it ? Thanks ! 

here's an example of what I have in mind (in addition of what's already been done). 
image

JulienIgnace-Unity:
For the MSAA one, it's going to go away in a subsequent PR so it did not make sense to change it now (it will become just a default MSAA value instead of a capability of the pipe).
But in general I would avoid adding more info box like this as it clutters the UI in my opinion.
We actually have quite a few interactions between features like this and if we wanted to be consistent and show such info for everything I'd expect this to get out of hand quickly :)

remi-chapelain (approved):
Fair enough. ✔️

PR #3880, opened 2021-03-15
Fix GBuffer debug view when using virtual texturing

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1319492/ 

The issue was due to the fact that the debug view was sort of working just because the gbuffer was set globally during the lighting pass, however with VT the way we set it might need to change. 

Regardless, this is the correct fix long run as we want to move away from global setting of resources. 

**What did I test** The repro case as it is simple enough and not much more should be needed.


sebastienlagarde (approved):

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Checking with @JMargevics testing status is sufficient. Approving.

PR #3882, opened 2021-03-15
Added a fallback for the ray traced directional shadow in case of a transmission (case 1307870).

anisunity:
https://fogbugz.unity3d.com/f/cases/1307870/
Currently the screen space shadow was overriding the whole on screen shadow, which was an issue in the case of transmitted shadows (sun light).  

This fixes the behavior and falls back on the rasterized shadow (only in the case of ray traced, the behavior was already correct for screen space shadow.

Testing status
Added a test to cover the new behavior. 
Made sure the issue doesn't happen anymore on the original dataset

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Ray-Traced-Shadows.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Ray-Traced-Shadows.md
@@ -89,7 +89,7 @@ Ray-traced shadows offer an alternative to the shadow map that Point and Spot Li
 
 **Ray-traced Point Light shadows (radius = 0.5m)**
 
-Ray-traced shadows offer the possibility of semi-transparent shadows for Point Lights cast by tranparent and transmissive GameObjects.
+Ray-traced shadows offer the possibility of semi-transparent shadows for Point Lights cast by transparent and transmissive GameObjects.
 
 ![](Images/RayTracedShadows11.png)
 


anisunity:
Did the changes requested by @sebastienlagarde

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -202,7 +208,13 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
             DirectionalLightData light = _DirectionalLightDatas[_DirectionalShadowIndex];
 
 #if defined(SCREEN_SPACE_SHADOWS_ON) && !defined(_SURFACE_TYPE_TRANSPARENT)
-            if ((light.screenSpaceShadowIndex & SCREEN_SPACE_SHADOW_INDEX_MASK) != INVALID_SCREEN_SPACE_SHADOW)
+            // Two different options are possible here
+            // - We have a ray trace shadow in which case we have no valid signal for a transmission and we need to fallback on the rasterized shadow
+            // - We have a screen space shadow and it already contains the transmission shadow and we can use it straight away
+            bool visibleLight = dot(bsdfData.normalWS, -light.forward) > 0.0;
```
sebastienlagarde, line 214: can you factor all those conditions in a single function so we can share the code between both call?  thanks

remi-chapelain (changes requested):
Forgot to take a look last week sorry !

Quickly took a look at the test scene, thanks for adding it ! ✔️ 
Tested double sided quad / alpha clipped transmission and multi mat to be sure ✔️ 

I have some small concerns : 
- Is it expected that punctual lights don't have this behavior at all meaning we can't see the shadows through the translucent object ? 
- Can we add this info to the documentation about the new fallback behavior for dir light ? Maybe @JordanL8 can take care of it ? 
- I also don't think this is mentioned in doc too for Area light (that have the same behavior as the dir light in that case)

anisunity:
@sebastienlagarde merged hd/bugfix into this PR and did the changes you requested

sebastienlagarde:
@remi-chapelain if you can recheck, thanks

remi-chapelain (changes requested):
Did a quick re-check as requested and everything is still working as expected. ✔️ 

I'm just waiting on the quick changes to the doc to be sure theses behavior are noted somewhere : 
- Punctual lights don't have this behavior, not implemented yet
- Area light and directional will fallback to raster shadow in the case of transmission.

Vic-Cooper:
Documented this fix in the Light Types guide section of Light-Components.

remi-chapelain (commented):
Thanks for adding it to the doc, but not to be THAT guy, but does that makes more sense to have that on the [ray traced shadows side]( https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.4/manual/Ray-Traced-Shadows.html) of the doc ?
To me, people struggling to understand this will look on the raytracing side first, and people who read it on the "light" side will never remember it and/or wonder why we talk about this really specific edge case here ?

Vic-Cooper:
You're totally right, this information should be there! I'll make that change now :)

remi-chapelain (approved):
Thanks ! Looks good to me 💚

PR #3883, opened 2021-03-15
Fix Decal Normal Blending NaN (Case #1317162)

johnpars:
### Purpose of this PR
Fix for case: https://fogbugz.unity3d.com/f/cases/1317162/

This PR addresses an issue with how the decal normal is blended/applied to the surface normal.
If these two vectors are perfectly opposing (for example the decal transform is facing up, and the surface facing down), then a blended vector would produce a 0 vector at f(0.5), resulting in the NaN when normalized. 

---
### Testing status
I confirmed that the repro project case no longer produces the NaNs.

I also tested all three axes on an axis-aligned cube with three decal projectors and confirmed that the issue no longer remains in those cases. Images below are taken with the NaN Debug view and pixels in red indicate a NaN.

**Before**
Broken_Screen

**After**
Fixed_Screen

### Comment to reviewers
I decided to fix this by introducing an epsilon to the normal blend term. It adds a cost of one additional ALU to decal buffer decoding. Doing this in the decoding allows to fix the problem in one place instead of safe normalizing everywhere (since many different shaders have to have their own approach to applying decal to surface data)


com.unity.render-pipelines.high-definition/Runtime/Material/Decal/Decal.hlsl
```diff
@@ -81,7 +81,11 @@ void DecodeFromDBuffer(
     // Use (254.0 / 255.0) instead of 0.5 to allow to encode 0 perfectly (encode as 127)
     // Range goes from -0.99607 to 1.0039
     surfaceData.normalWS.xyz = inDBuffer1.xyz * 2.0 - (254.0 / 255.0);
-    surfaceData.normalWS.w = inDBuffer1.w;
+
+    // Case 1317162: Apply a NaN guard via the normal blend mask.
+    // An epsilon will prevent a situation where two opposing normals (surface & decal) produce a 0 normal from blending.
+    surfaceData.normalWS.w = inDBuffer1.w + 1e-5;
```
sebastienlagarde, line 87: I am concern by this fix. I am not confortable with touching the alpha value as we will be unable to know if we are pure 0 or not.  Maybe we should do surfaceData.normalWS.w = inDBuffer1.w > 0 ? saturate(inDBuffer1.w + 1e-5) : 0;
thought?

sebastienlagarde: ah currently guess it is the opposite and need to check pure 1

johnpars: As discussed in slack we will replace this approach by instead safe normalizing at all occurrences of `ApplyDecalToSurfaceData`. Note that in the future we will introduce surface gradient framework to decals which will fix this implicitly and remove the need for safe normalize.

sebastienlagarde (approved):

TomasKiniulis (approved):
Thanks for the fix John! Also did a quick check if it works correctly with Double Sided and different Normal Modes. Not running into any NaNs anymore.

PR #3887, opened 2021-03-16
[HDRP]Refactor MaterialScope calls on MaterialUIBlocks

alex-vazquez:
---
### Purpose of this PR
Clean MaterialUIBlock code using MaterialHeaderScope.

https://jira.unity3d.com/browse/XPIPELINE-119

Adding translation to material sections.

---
### Testing status
- Check the Material editor sections mantain their expanded/collapsed status and look in the same way as before
- Make sure tesselation is not show when it does not need to be shown
- See that Transparent Inputs, show disabled their internal properties, instead of having a empty foldout.

---
### Comments to reviewers
Cleaning the calls of the Headers Scopes, as everything inherit from MaterialHeaderScope

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/RefractionUIBlock.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/RefractionUIBlock.cs
@@ -79,7 +79,7 @@ namespace UnityEditor.Rendering.HighDefinition
         /// </summary>
         public override void OnGUI()
         {
-            using (new EditorGUI.DisabledScope(refractionModel == null))
+            if (refractionModel != null)
             {
                 materialEditor.ShaderProperty(refractionModel, Styles.refractionModelText);
                 var mode = (ScreenSpaceRefraction.RefractionModel)refractionModel.floatValue;
--- a/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/DistortionUIBlock.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/DistortionUIBlock.cs
@@ -66,7 +66,7 @@ namespace UnityEditor.Rendering.HighDefinition
         /// </summary>
         public override void OnGUI()
         {
-            using (new EditorGUI.DisabledScope(distortionEnable == null))
+            if (distortionEnable != null)
             {
                 materialEditor.ShaderProperty(distortionEnable, Styles.distortionEnableText);
 


com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayeringOptionsUIBlock.cs
```diff
@@ -76,21 +75,14 @@ public override void LoadMaterialProperties()
...
-            // We're using a subheader here because we know that layering options are only used within layers
-            using (var header = new MaterialHeaderScope(Styles.header, (uint)m_ExpandableBit, materialEditor, subHeader: true))
-            {
-                if (header.expanded)
-                {
-                    DrawLayeringOptionsGUI();
-                }
-            }
-        }
+        protected override bool isSubHeader => true;
```
jenniferd-unity, line 80: why is this class specifically a subheader?

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/MaterialUIBlock.cs
```diff
@@ -133,11 +134,23 @@ public enum ExpandableBit : uint
             User19 = 1 << 30,
         }
 
+        internal MaterialUIBlock()
+        {
```
jenniferd-unity, line 138: [question about c#] here i assume it is okay we initialize the values to their default value that dont need to be explicit?

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/MaterialUIBlock.cs
```diff
@@ -198,6 +211,34 @@ protected MaterialProperty[] FindPropertyLayered(string propertyName, int layerC
...
+        /// If the section should be shown
+        /// </summary>
+        protected virtual bool showSection => true;
+
+        /// <summary>
+        /// GUI callback when the header is open
+        /// </summary>
+        protected virtual void OnGUIOpen()
+        {
+            throw new NotImplementedException($"You must implement {nameof(OnGUIOpen)} if you are not overriding {nameof(OnGUI)}");
```
jenniferd-unity, line 241: great idea with the exception :)

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/RefractionUIBlock.cs
```diff
@@ -79,7 +79,7 @@ public override void LoadMaterialProperties()
         /// </summary>
         public override void OnGUI()
         {
-            if (refractionModel != null)
+            using (new EditorGUI.DisabledScope(refractionModel == null))
```
jenniferd-unity, line 82: i assume this changes the UI, is it on purpose?

alex-vazquez: Yeah, before we wer not showing anything, now we show everything disabled, I think it makes more sense as we end up with a empty foldout.

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayeringOptionsUIBlock.cs
```diff
@@ -76,21 +75,14 @@ public override void LoadMaterialProperties()
         }
 
         /// <summary>
-        /// Renders the properties in the block.
+        /// Property that specifies if the scope is a subheader
         /// </summary>
-        public override void OnGUI()
-        {
-            // We're using a subheader here because we know that layering options are only used within layers
-            using (var header = new MaterialHeaderScope(Styles.header, (uint)m_ExpandableBit, materialEditor, subHeader: true))
```
alex-vazquez, line 84: @jenniferd-unity , before it was also a sub header

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/LayeringOptionsUIBlock.cs
alex-vazquez: see my other comment please

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/MaterialUIBlock.cs
alex-vazquez: Yes, the default constructor on c# initializes:

- All numeric fields in the class to zero.
- All string and object fields to null.

So later on the method, `internal void Initialize(MaterialEditor materialEditor, MaterialProperty[] properties, MaterialUIBlockList parent)` everything will be handled.
If a `header` is not speficied and the `OnGUI` is not overrided, the dev will get the exception on `MaterialHeaderScope`, this along the other exception will handle both paths on any MaterialUIBlock

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/MaterialUIBlock.cs
alex-vazquez: Thank you :)

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/MaterialUIBlock.cs
RSlysz: If I get it well here you make the MaterialUIBlock non usable to user as you need to comply with the internal visibility to be able to create one of its child.

Is this something that we want? At that point, why MaterialUIBlock is even public? we can all works in package internal visibility.
Be sure you don't break up potential feature :)

------

And what you said about constructor is not 100% accurate so here is more explanation for Jenny.
The (default) constructor on c# initializes:
- All value type (numeric fields and struct) in the class to 0, 0f, default struct... if they don't have a provided initializer at the scope of the class.
- All object type (string, array and object fields) to null if they don't have a provided initializer at the scope of the class.
- Then once all is initialized, the content of the constructor (if any) runs and potentially add more value change.

Also, default constructor always exists and is public except specified differently.

Try this to be sure ;)
```
using UnityEngine;

#if UNITY_EDITOR
using UnityEditor;
#endif

public class tester : MonoBehaviour
{
    class TestClass
    {
        public int int_default;
        public int int_42 = 42;

        public class Klass
        {
            public int int_field;
            public Klass(int val) => int_field = val;
        }
        public Klass klass_default;
        public Klass klass_33 = new Klass(33);
    }
    
#if UNITY_EDITOR
    [MenuItem("CONTEXT/tester/Test")]
    static void Test()
    {
        TestClass tc = new TestClass();
        
        Debug.Log($"int_default={tc?.int_default}");                                                //expect 0
        Debug.Log($"int_42={tc?.int_42}");                                                          //expect 42
        Debug.Log($"klass_default={(tc?.klass_default?.int_field == null ? "NULL": "NOT NULL")}");  //expect null
        Debug.Log($"klass_33={tc?.klass_33?.int_field}");                                           //expect 33
    }
#endif
}
```

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/MaterialUIBlock.cs
RSlysz: Nice indeed!

RSlysz (changes requested):
Check the constructor visibility regarding feature provided to user. I feel not many write there material and the inspector that go with but now they cannot do it anymore with our tool. Check if it is intentional.

jenniferd-unity (approved):

alex-vazquez:
Awesome @Unity-Technologies/gfx-qa-hdrp  team, could you have a look at the materials editors, see that sections mantain their state(open/closed) and make and have an special look at `Transparency Inputs` and `Tesselation`? Many thanks.

RSlysz (approved):
I still don't get why we need a protected default constructor. I mean, it do nothing and the class is abstract so you already have to inherit it to use the mechanisms it contains. However, now it don't forbid user to use it so 📗

TomasKiniulis (changes requested):
Heya @alex-vazquez! Checked Lit, LitTessellation, Unlit, Decal, TerrainLit, Axf LayeredLit materials with Transparent surface type where possible. I've ran into only one issue.

Unlit material "Transparency Inputs" section expanded spams the following:
```
NullReferenceException: Object reference not set to an instance of an object
UnityEditor.MaterialEditor.GetPropertyRect (UnityEditor.MaterialProperty prop, System.String label, System.Boolean ignoreDrawer) (at <6fc837d8b04442e1b1d652756199dcd4>:0)
UnityEditor.MaterialEditor.GetPropertyRect (UnityEditor.MaterialProperty prop, UnityEngine.GUIContent label, System.Boolean ignoreDrawer) (at <6fc837d8b04442e1b1d652756199dcd4>:0)
UnityEditor.MaterialEditor.ShaderProperty (UnityEditor.MaterialProperty prop, UnityEngine.GUIContent label, System.Int32 labelIndent) (at <6fc837d8b04442e1b1d652756199dcd4>:0)
UnityEditor.MaterialEditor.ShaderProperty (UnityEditor.MaterialProperty prop, UnityEngine.GUIContent label) (at <6fc837d8b04442e1b1d652756199dcd4>:0)
UnityEditor.MaterialEditor.ShaderProperty (UnityEditor.MaterialProperty prop, System.String label) (at <6fc837d8b04442e1b1d652756199dcd4>:0)
UnityEditor.Rendering.HighDefinition.RefractionUIBlock.OnGUI () (at C:/Users/Tomas/Documents/Github_Projects/Graphics3/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/RefractionUIBlock.cs:84)
UnityEditor.Rendering.HighDefinition.MaterialUIBlockList.OnGUI (UnityEditor.MaterialEditor materialEditor, UnityEditor.MaterialProperty[] properties) (at C:/Users/Tomas/Documents/Github_Projects/Graphics3/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/MaterialUIBlockList.cs:62)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr, Boolean&)
```


https://user-images.githubusercontent.com/50582134/112028460-dbfeaa80-8b40-11eb-81df-0374c7bc5884.mp4

TomasKiniulis (approved):
Perfect! Not encountering any issues anymore.

PR #3890, opened 2021-03-16
[HDRP]Deprecating VolumeComponentDeprecated

alex-vazquez:
---
### Purpose of this PR
Using the build in attributes instead of having a custom attribute just for volumes

[XPIPELINE-117](https://jira.unity3d.com/browse/XPIPELINE-117)

---
### Testing status
Check that the Volumes that have the attribute `HideInInspector` or `Obsolete` are not being shown when pressing `Add Override`

---
### Comments to reviewers
Making a difference between when we are actually deprecating stuff(using Obsolete), and when we want to hide the component to be added (we use HideInInspector)

### DIFF
--- a/com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
+++ b/com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
@@ -9,11 +9,13 @@ namespace UnityEngine.Rendering.Tests
 {
     public class VolumeComponentEditorTests
     {
+        [HideInInspector]
         class VolumeComponentNoAdditionalAttributes : VolumeComponent
         {
             public MinFloatParameter parameter = new MinFloatParameter(0f, 0f);
         }
 
+        [HideInInspector]
         class VolumeComponentAllAdditionalAttributes : VolumeComponent
         {
             [AdditionalProperty]
@@ -23,6 +25,7 @@ namespace UnityEngine.Rendering.Tests
             public FloatParameter parameter2 = new MinFloatParameter(0f, 0f);
         }
 
+        [HideInInspector]
         class VolumeComponentMixedAdditionalAttributes : VolumeComponent
         {
             public MinFloatParameter parameter1 = new MinFloatParameter(0f, 0f);
@@ -108,6 +111,7 @@ namespace UnityEngine.Rendering.Tests
 
         #region Decorators Handling Test
 
+        [HideInInspector]
         class VolumeComponentDecorators : VolumeComponent
         {
             [Tooltip("Increase to make the noise texture appear bigger and less")]


JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/Debug/VolumeDebug.cs
```diff
@@ -131,7 +131,8 @@ static public List<Type> componentTypes
                 if (s_ComponentTypes == null)
                 {
                     s_ComponentTypes = VolumeManager.instance.baseComponentTypeArray
-                        .Where(t => !t.IsDefined(typeof(VolumeComponentDeprecated), false))
+                        .Where(t => !t.IsDefined(typeof(HideInInspector), false))
```
RSlysz, line 134: This is a runtime API. I feel it weird to react to editor attribute HideInInspector.

RSlysz (commented):
Shouldn't we have two different API, one that filter for inspector and another targeting runtime stuff?

com.unity.render-pipelines.high-definition/Runtime/Debug/VolumeDebug.cs
alex-vazquez: We were using mixing them, because we were not deprecating them actually, and we have them on code behind... About HideInInspector, do you think is better to add another attribute?

RSlysz: Well I agree it sound stupid to add another to do the "same". But the naming is kinda not intuitive. I am not sure about that.

@JulienIgnace-Unity what do you think of this?

sebastienlagarde (approved):

com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -136,6 +136,7 @@ The version number for this package has increased due to a version update of a r
 - Updated macros to be compatible with the new shader preprocessor.
 - Updated shaders to be compatible with Microsoft's DXC.
 - Changed CommandBufferPool.Get() to create an unnamed CommandBuffer. (No profiling markers)
+- Deprecating VolumeComponentDeprecad, using HideInInspector or Obsolete instead
```
jenniferd-unity, line 139: typo reported by Arttu

com.unity.render-pipelines.high-definition/Runtime/Debug/VolumeDebug.cs
jenniferd-unity: HideInInspector is definitely used by our users in shipped games sadly T_T

jenniferd-unity (approved):

com.unity.render-pipelines.high-definition/Runtime/Debug/VolumeDebug.cs
alex-vazquez: cc @JulienIgnace-Unity , are you good to go with `HideInInspector`?

JulienIgnace-Unity: I feel like it would be a bit overkill to have another one. Since here we are talking about debug stuff, I feel like HideInInspector is actually kind of relevant (inspector == debugger)

alex-vazquez:
@Unity-Technologies/gfx-qa-hdrp , could you check that the Volume Components shown on this PR on the Add Override are the same as the ones shown on master? Thank you :)

iM0ve (approved):
What's tested:
* Compared master override list with PR branch. They match
* Used a few overrides to check they still work (Fog, Indirect intensity override)
* Upgraded a volume profile in search of any unexpected results

PR #3054, opened 2021-01-11
Implementation of SRP Global Settings for HDRP

jenniferd-unity:
### Purpose of this PR
Presentation powerpoint [here](https://docs.google.com/presentation/d/1e1_FChJuNwaTEyZUsZVbYdmsAEs6ExpO2LdHJq58iww/edit?usp=sharing)
This is the first implementation of HDRP Global Settings as an asset separate from the HDRPAsset. Further development is planned.

[UI updated on Feb 26th]
image

I have moved from the HDRenderPipeline.defaultAsset various settings:
- Resources (Editor/RayTracing/Player)
- Default Volume Profile for Scenes and LookDev
- Layer names
- Shader Variant Log level
- Default Frame Settings for Camera, Reflection Probes
- Lens Attenuation
- Diffusion Profile Settings List
- Custom Post Processes Lists

No behaviors should be changed on any HDRP project with those changes except:
- the HDRPAsset no longer contains settings that were duplicated for the default asset
- there is always a HDRenderPipelineGlobalSettings asset active
- this asset can be viewed from the Project Settings > Graphics window
- individual HDRenderPipelineGlobalSettings assets have their own inspector.

When upgrading to this new version:
- a new asset should be created in the HDDefaultResources folder called HDRenderPipelineGlobalSettings .asset
- values of the HDRPAsset saved in GraphicsSettings compatible are migrated to this new asset

The create menu can be found here: 
image

When deleting the active Global Settings asset, a new one will be created and an error message displayed:
https://user-images.githubusercontent.com/71718938/108837110-145aa980-75d2-11eb-8a70-6c66236f2941.mp4

Asset creation workflow: set as active Asset if created from the HDRP Settings menu only.
https://user-images.githubusercontent.com/71718938/108840941-4faba700-75d7-11eb-853e-0cb2d87876d8.mp4

---
### Testing status
Tried the following locally - on Windows only:
- Editor: upgrading HDRP template, Spaceship demo, small local projects,
- Creating new project from Hub,
- Creating players for some projects (template HD and spaceship demo),
- Running HDRP tests, RayTracing tests.

---
### Comments to reviewers
This is the first implementation of HDRP Global Settings as an asset separate from the HDRPAsset. Further development is planned.

---
### JIRA
[https://jira.unity3d.com/browse/XPIPELINE-38](https://jira.unity3d.com/browse/XPIPELINE-38)
[https://unity3d.atlassian.net/browse/UR-1967](https://unity3d.atlassian.net/browse/UR-1967)

PR #3894, opened 2021-03-16
Fixed HDRPAsset loosing its reference to the ray tracing resources when clicking on a different quality level that doesn't have ray tracing (case 1320304).

anisunity:
https://fogbugz.unity3d.com/f/cases/1320304/
When the quality settings were introduced, the forced upgrade of the asset was taking into account the current asset to adjust the reference to the ray tracing resources for both the current asset and default asset. This PR fixes that ( takes into account the default for the default's reference).
When the player was built while just selecting the other asset in the quality menu the reference to the resource was lost and that least to chained exceptions which lead to memory leaks and finally a crash.

Testing status
The issue doesn't repro anymore.


jenniferd-unity:
Note that the HDRP Global Settings PR may obsolete this PR unless you are planning on backporting it.

TomasKiniulis (commented):
Tested on completely fresh HDRP template along case repro steps. Can confirm that Player runs correctly now. However, adding these changes to @jenniferd-unity PR https://github.com/Unity-Technologies/Graphics/pull/3054#event-4341423349 reintroduces crashes.

sebastienlagarde:
Fix for @jenniferd-unity  will be done in the PR when this one will be merge in master

PR #3895, opened 2021-03-17
Fixing multi camera rendering for DRS

kecho:
### Purpose of this PR

Fogbugz: https://fogbugz.unity3d.com/f/cases/1315571/

Fixing issues with DX12, Hardware DRS with Multicamera rendering (having for example a reflection probe camera).

There is a single singleton call to ScalableBufferManager.ResizeBuffers, which gets set only once per frame. 

The pseudo code is like this:

```
DrsHandler.ResizeBuffers();
For each camera c,
   c.Update();
For each camera c,
   c.RenderRequest();
```

This change adds the posibility of cameras to change 'context' by calling a new API into DynamicResolutionRendering.
The resulting pseudocde is now like this:

```
For each camera c
   DrsHandler.SetActiveCamera(c);
   c.Update();
   
For each camera c
   DrsHandler.SetActiveCamera(c);
   c.RenderRequest();
```

This change adds a bit of CPU code, because now calling ResizeBuffers happens every frame, however it only iterates on a few dozen render targets. Also, this change saves memory since the resulting buffers are still aliased (i.e. the render target of a planar reflection volume is on the same dx12 heap of the DRSd main buffer).

---
### Testing status
* Tested on dx12 drs hardware mode, a scene with a planar reflection volume (as indicated in the fogbugz).
image
* Requires test from qa in xb1 with DRS on hardware mode.
* Requires test in ps4 by qa with DRS on hardware mode.

---
### Comments to reviewers
* This change adds new APIs
* This change preserves the old behaviour. New APIs only change behaviour if used. If not used, old behaviour is preserved (a global default sate is used).
* In the process of backporting this change to 10.x 
* Cannot close the fogbugz until this change has been preported, and we verified that all changes required are in 10.x


com.unity.render-pipelines.core/CHANGELOG.md.preformat.bak
```diff
@@ -0,0 +1,227 @@
+# Changelog
```
FrancescoC-unity, line 1: This file looks like it shouldn't be here :D

kecho: FFFF

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -147,58 +232,84 @@ public void SetCurrentCameraRequest(bool cameraRequest)
...
-            {
-                m_PrevFraction = m_CurrentFraction;
-
-                if (!m_ForceSoftwareFallback && type == DynamicResolutionType.Hardware)
-                {
-                    ScalableBufferManager.ResizeBuffers(m_CurrentFraction, m_CurrentFraction);
-                }
-
-                if (OnResolutionChange != null)
-                    OnResolutionChange();
```
FrancescoC-unity, line 186: This was called for SW dynamic res too and it doesn't look like it is called anymore?

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -59,21 +59,84 @@ public class DynamicResolutionHandler
...
         private PerformDynamicRes m_DynamicResMethod = null;
-        private static DynamicResolutionHandler s_Instance = new DynamicResolutionHandler();
+
+        private GlobalDynamicResolutionSettings m_CachedSettings = GlobalDynamicResolutionSettings.NewDefault();
+
+        private WeakReference m_OwnerCameraWeakRef = null;
+        private static Dictionary<int, DynamicResolutionHandler> s_CameraInstances = new Dictionary<int, DynamicResolutionHandler>();
+        private static DynamicResolutionHandler s_DefaultInstance = new DynamicResolutionHandler()
+        {
+            m_DynamicResMethod = DefaultDynamicResMethod
```
FrancescoC-unity, line 72: Any specific reason why to move this out of the constructor?

kecho: Yes. Default behaviour. 
If you look below, a user must set the dynamic res method explicitely per camera. Id not, then the default one (this one) is used.
If we dont take this approach, it means that the user would require to revise their code, and ensure they bind the camera they want the method to be on.
An alternative is to take this out of the instance i guess..

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: yep. nice catch. Restored.

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: I simplified this a lot by making the method and scaler type statics.

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -59,21 +61,79 @@ public class DynamicResolutionHandler
...
+                instance = new DynamicResolutionHandler();
+                instance.m_OwnerCameraWeakRef = new WeakReference(camera);
+                s_CameraInstances.Add(key, instance);
+            }
+            return instance;
+        }
+
+        private static int g_GarbageTickId = 0;
+        private static List<int> s_GarbageHandlers = new List<int>();
+        private static void GarbageCollectCameras()
```
FrancescoC-unity, line 109: Do we need to do this at all? Realistically how many cameras we'll have during lifetime of a run (before domain reload where static stuff is reset anyway)? 

If we create the dictionary with a sensible initial capacity it should be fine? Let's chat later

FrancescoC-unity:
For QA: I would also do a somewhat thorough pass on software dynamic res in general, I don't expect problems, but that logic has changed to a bit.

TomasKiniulis (approved):
Checking repro project on Xbox also fixes the issue! Attaching xbox player master and PR results

master:
https://user-images.githubusercontent.com/50582134/111664072-16530980-881a-11eb-8774-0d17b034104e.mp4
PR:
https://user-images.githubusercontent.com/50582134/111664111-1f43db00-881a-11eb-82d9-60276736dcb7.mp4

Note that there's notcieable flickering happening in PR video. This is unrelated to the PR, and is caused by area lights. Filed a ticket here: https://fogbugz.unity3d.com/f/cases/1322629/

Requesting @JMargevics for pass on PS4(he's in dots training this week though)

Edit: also checked software dynamic res on Editor and Xbox player with various scaling filters, also forces screen percentage on or off, behaviour seems to stay the same

sebastienlagarde (approved):

sebastienlagarde:
@kecho can you retarget this bugfix on master? given it add new API we can't add it to the 2020.2 LTS

JMargevics:
Hey just tried this on PS4 with hardware mode enabled for DRS and got some incorrect rendering:
https://user-images.githubusercontent.com/50581102/112007485-5b827e80-8b2d-11eb-8c8e-8ef3168cc34b.mp4

With DRS disabled completely scene works fine
ps4nodrs

JMargevics:
In comparison on master nothing is visible and the console freezes after few seconds. 
image

JMargevics (changes requested):
PS4 doesn't freeze with this PR but rendering is still corrupted.

kecho:
Good find @JMargevics . Will tackle it asap.

kecho:
@JMargevics, to fix PS4 is going to take a lot of time, because I have to fix it in the unity editor :(
Could you make a fogbugz for ps4? we should treat it as a separate issue. At least lets provide a fix for PC for our users. We can make a note that the ps4 fix is coming.
Id like this change to be in the wild to mitigate any additional issues if they occur.
let me know. Once you create the fogbugz you can assign it to me. @sebastienlagarde what do you think of this plan?

kecho:
@FrancescoC-unity and @sebastienlagarde , GNM is having tons of issues for DRS hardware mode in the backend side:
1) I think cmask across aliased textures is busted
2) Bad mip calculations
3) bunch of other exceptions / validation issues.
Its gonna be a mega bug :) but I think its doable. Let me know @JMargevics when you create the bug, i can jump into it next bug week.

kecho:
@sebastienlagarde rebased to master. Just need the pass from QA, since the ps4 issues can't be fixed with HDRP changes. Ill have to rewrite certain pieces of the DRS system on GNM.

JMargevics (approved):

JMargevics:
> 
> 
> @JMargevics, to fix PS4 is going to take a lot of time, because I have to fix it in the unity editor :(
> Could you make a fogbugz for ps4? we should treat it as a separate issue. At least lets provide a fix for PC for our users. We can make a note that the ps4 fix is coming.
> Id like this change to be in the wild to mitigate any additional issues if they occur.
> let me know. Once you create the fogbugz you can assign it to me. @sebastienlagarde what do you think of this plan?

If it's a heavy issue, then yeah, I will make a separate report for it. Greenlighting this one then 👍

FrancescoC-unity:
> @FrancescoC-unity and @sebastienlagarde , GNM is having tons of issues for DRS hardware mode in the backend side:
> 
> 1. I think cmask across aliased textures is busted
> 2. Bad mip calculations
> 3. bunch of other exceptions / validation issues.
>    Its gonna be a mega bug :) but I think its doable. Let me know @JMargevics when you create the bug, i can jump into it next bug week.

If it is broken currently on master anyway, I'd say it is fine. 

Worth checking with platform team if they are aware.

PR #3898, opened 2021-03-17
[HDRP] Fixed custom pass custom drawer rendering with SerializedProperties

alelievr:
### Purpose of this PR
Fixed a potential issue causing the custom pass UI drawers to not render correctly when using custom properties.

### Testing status
I created a custom pass drawer with a bunch of properties with custom attributes like Range, Space, Header etc. and played with the custom pass properties, play mode, foldout in UI. No UI error found

For reference this is the issue that could happen: https://fogbugz.unity3d.com/f/cases/1304095/


RSlysz (approved):
version before was having a wrong handling in case of multiselection. This fix is way better

sebastienlagarde:
@alelievr is this bugfix only on trunk? if it is the case can you retarget to master? thanks

alelievr:
Double checked with the serialize reference feature and this can be backported up to 10.x (2020.2)

sebastienlagarde (approved):

PR #3905, opened 2021-03-17
[URP] Renderer upgrader bug fix (case 1321796)

cinight:
### Purpose of this PR
Fixed [Case 1321796](https://fogbugz.unity3d.com/f/cases/1321796/) where the upgrader doesn't work in trunk, due to the list of assets do not contain all of the RendererData when the cache exists.
Fixed infinite import loop when the project contains some corrupted RendererData assets which RendererData is a subAsset of RendererFeature.
Putting the upgrader back to correct namespace.

### Testing status

Yamato test failure are same on Graphics master. Playmode tests are green.
I tested below on Mac manually.

#### Test case 1 - upgrading with cache (repro steps from [1321796](https://fogbugz.unity3d.com/f/cases/1321796/)):
1. Clone UniversalRenderingExamples, discarded all changes and deleted Library folder
2. Open with 2021.2.0a6
3. Close Unity
4. Edit package manifest to use local Graphics master
5. Open with 2021.2.0a10
6. Open all the scenes and check
Before fix - no rendering and errors in console
After fix - rendering fine and no errors
7. In ProjectView, search "t:ForwardRendererData"
After fix - the search returns nothing, means the asset changes are saved

#### Test case 2 - upgrading with corrupted asset:
1. Delete the big UniversalGraphicsTest project's Library folder
2. Edit Assets\csc.rsp to make it an empty file
3. Open project with 2021.2.0a10
Before fix - stuck in infinite importing loop
After fix - project opens fine and RendererData are all upgraded successfully. No related warning / error in console.

#### Test case 3 - URP template (repro steps from [1322972](https://fogbugz.unity3d.com/f/cases/1322972/)):
I cannot reproduce the error with 2021.2.0a10.

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/AssetPostProcessors/ForwardToUniversalRendererPostprocessor.cs
+++ b/com.unity.render-pipelines.universal/Editor/AssetPostProcessors/ForwardToUniversalRendererPostprocessor.cs
@@ -23,9 +23,8 @@ namespace UnityEditor.Rendering.Universal
                 scriptProperty.objectReferenceValue = stdRendererScriptObj;
                 so.ApplyModifiedProperties();
 
-                //Save and re-import asset
+                //Re-import asset
                 //This prevents the "Importer(NativeFormatImporter) generated inconsistent result" warning
-                AssetDatabase.SaveAssets();
                 AssetDatabase.ImportAsset(rendererDataPath);
 
                 editedAssetsCount++;


kaychang-unity (approved):
I tested on my test project and it works (Windows editor).

kaychang-unity (changes requested):
I tried running the new code on UniversalGraphicsTests and it gets stuck in an infinite loop when the project loads.

cinight:
Looks like all the failures are same as Graphics master. Also tried opening the projects locally and they opens fine.
Now merged latest master into the branch and tests are re-running.

kaychang-unity (approved):
It works on my test projects, thanks!

erikabar (approved):
already tested (see the description), approving

hdb-unity (approved):
Checked project loading times:
**Original:**
[Project] Loading completed in 130.883 seconds

**PR:**
[Project] Loading completed in 144.475 seconds

Version: 2021.2.0a10

Seems within reason to me (the new code does more work + there will be some variance in the times).

Also appears to fix the issue in my case.

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/ForwardToUniversalRendererPostprocessor.cs
```diff
@@ -3,57 +3,91 @@
...
-                //change the script to use UniversalRendererData
+                //Change the script to use UniversalRendererData
+                so.Update();
                 scriptProperty.objectReferenceValue = stdRendererScriptObj;
                 so.ApplyModifiedProperties();
-                noOfFwdRendererData++;
+
+                //Save and re-import asset
+                //This prevents the "Importer(NativeFormatImporter) generated inconsistent result" warning
+                AssetDatabase.SaveAssets();
```
phi-lira, line 28: Minor: Do we need to call AssetDatabase.SaveAssets() for each asset? AssetDatabase.SaveAssets() is expensive operation.

phi-lira (approved):

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/ForwardToUniversalRendererPostprocessor.cs
cinight: I tested the big UniversalGraphicsTest on my Mac, doesn't look like it causes any importing time differences.

without SaveAssets():
153.282+1.569 seconds = 154.851
150.601+1.531 seconds = 152.132
with SaveAssets()
148.142+1.230 seconds = 149.372
151.260+1.629 seconds = 152.889

But anyway I tested both test cases without SaveAssets() and the upgrade still works. So I've removed SaveAssets() in latest commit. Will merge the PR when Yamato tests are good as well.

PR #3909, opened 2021-03-17
[HDRP][Path Tracing] Fabric material support

eturquin:
Before:
Objects with Fabric materials would appear all black in path tracing (like all other unsupported materials).

Now:
Objects with Fabric materials will be rendered with a look matching closely their rasterized counterparts. Note that SSS and back diffuse may look slightly different, as the computations are more accurate in PT (resp. random walk and proper BTDF). Also, like in other PT materials (for the time being), there is no texture filtering, which can lead to slight aliasing on fine weaves patterns.

Supports both Cotton/wool and Silk variants:

Cotton/wool is made of:
* Lambert BRDF + SSS
* Sheen BRDF
* Lambert BTDF

Silk is made of:
* Burley BRDF + SSS
* Colored Aniso GGX BRDF
* Lambert BTDF

The fabric material has been tested on the HDRP's Fabric material samples scene.

This PR also contains a fix for the normal used in light sampling activation (shading instead of geom), and minor refactor of the Lit material.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
@@ -68,7 +68,7 @@ HDRP path tracing in Unity 2020.2 has the following limitations:
 - Does not support tessellation.
 - Does not support Tube and Disc shaped Area Light.
 - Does not support Translucent Opaque Materials.
-- Does not support several of HDRP's Materials. This includes Fabric, Eye, StackLit, Hair, Decal.
+- Does not support several of HDRP's Materials. This includes Eye, StackLit, Hair, Decal.
 - Does not support per-pixel displacement (parallax occlusion mapping, height map, depth offset).
 - Does not support MSAA.
 - For renderers that have [LODs](https://docs.unity3d.com/Documentation/Manual/LevelOfDetail.html), the ray tracing acceleration structure only includes the highest level LOD and ignores the lower LODs.


eturquin:
fabric_pt
fabric_raster

remi-chapelain:
Not tested yet but out of my mind we need to update docs for this. We have a line that says that fabric is not supported in the path tracer [here](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.4/manual/Ray-Tracing-Getting-Started.html).
Also, had trouble findings this limitation sections a bit, I expected to find this in the path tracer section, would it make more sense to have it there ?  
@JordanL8

remi-chapelain (commented):
Not tested yet but out of my mind we need to update docs for this. We have a line that says that fabric is not supported in the path tracer [here](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.4/manual/Ray-Tracing-Getting-Started.html#unsupported-features-of-path-tracing).
Also, had trouble findings this limitation sections a bit, I expected to find this in the path tracer section, would it make more sense to have it there ? (or both)
@JordanL8

JordanL8:
Yeah, we should at least add a link from the path-tracing page to the limitations section on the ray-tracing page. I'll also remove fabric from the list of unsupported materials.

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/FabricPathTracing.hlsl
```diff
@@ -0,0 +1,280 @@
...
+        return false;
+
+    mtlData.bsdfWeight[0] = Luminance(mtlData.bsdfData.diffuseColor) * mtlData.bsdfData.ambientOcclusion;
+
+    // If N.V < 0 (can happen with normal mapping) we want to avoid spec sampling
+    float NdotV = dot(mtlData.bsdfData.normalWS, mtlData.V);
+    if (NdotV > 0.001)
+    {
+        mtlData.bsdfWeight[1] = HasFlag(mtlData.bsdfData.materialFeatures, MATERIALFEATUREFLAGS_FABRIC_COTTON_WOOL) ?
+            0.1 * Luminance(mtlData.bsdfData.fresnel0) : Luminance(F_Schlick(mtlData.bsdfData.fresnel0, NdotV));
```
sebastienlagarde, line 49: why is there the * 0.1 here? maybe add some comment?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitPathTracing.hlsl
```diff
@@ -73,12 +75,10 @@ bool CreateMaterialData(PathIntersection pathIntersection, BuiltinData builtinDa
     else // Below
     {
         float NdotV = -dot(mtlData.bsdfData.normalWS, mtlData.V);
-        float F = F_FresnelDielectric(1.0 / mtlData.bsdfData.ior, NdotV);
+        float F = F_FresnelDielectric(1.0 / mtlData.ior, NdotV);
 
         // If N.V < 0 (can happen with normal mapping) we want to avoid spec sampling
         bool consistentNormal = (NdotV > 0.001);
-        mtlData.bsdfWeight[0] = 0.0;
```
sebastienlagarde, line 80: is it expected to not fill mtlData.bsdfWeight[0] anymore? (I do'nt have the full picture of the code, so maybe it is init elsewhere)

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitPathTracing.hlsl
```diff
@@ -358,7 +358,7 @@ float3 ApplyAbsorption(MaterialData mtlData, float dist, bool isSampleBelow, flo
     #ifdef _REFRACTION_THIN
         value *= exp(-mtlData.bsdfData.absorptionCoefficient * REFRACTION_THIN_DISTANCE);
     #else
-        value *= exp(-mtlData.bsdfData.absorptionCoefficient * dist);
+        value *= exp(-mtlData.bsdfData.absorptionCoefficient * min(dist, surfaceData.atDistance * 10.0));
```
sebastienlagarde, line 361: why is there a * 10.0 here? (add comment)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingBSDF.hlsl
```diff
@@ -387,6 +461,25 @@ bool SampleDelta(MaterialData mtlData,
...
 }
 
+bool SampleLambert(MaterialData mtlData,
+                   float3 inputSample,
+               out float3 outgoingDir,
+               out float3 value,
+               out float pdf)
+{
+    bool retVal = BRDF::SampleLambert(mtlData, inputSample, outgoingDir, value, pdf);
+    outgoingDir *= -1.0;
```
sebastienlagarde, line 471: why is there a -1.0? can you comment?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
```diff
@@ -101,7 +101,7 @@ void ComputeSurfaceScattering(inout PathIntersection pathIntersection : SV_RayPa
     #ifdef _SURFACE_TYPE_TRANSPARENT
         float3 lightNormal = 0.0;
     #else
-        float3 lightNormal = mtlData.bsdfData.geomNormalWS;
+        float3 lightNormal = mtlData.bsdfData.normalWS;
```
sebastienlagarde, line 104: if this is a fix, it must be done in a separate PR and we backport the fix to other version

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitPathTracing.hlsl
sebastienlagarde: ok I see it was just moved code, but we should add a comment.

sebastienlagarde (approved):
Good for me, just add more comment into the code for various values, thanks

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitPathTracing.hlsl
eturquin: Yes it it, right above.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingBSDF.hlsl
eturquin: Well it's just a flip, I'll write it as x = -x;

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
eturquin: Ok I'll do a separate PR.

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/FabricPathTracing.hlsl
eturquin: I added some comments.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingBSDF.hlsl
```diff
@@ -387,6 +461,25 @@ bool SampleDelta(MaterialData mtlData,
...
 }
 
+bool SampleLambert(MaterialData mtlData,
+                   float3 inputSample,
+               out float3 outgoingDir,
+               out float3 value,
+               out float pdf)
+{
+    bool retVal = BRDF::SampleLambert(mtlData, inputSample, outgoingDir, value, pdf);
+    outgoingDir = -outgoingDir;
```
sebastienlagarde, line 471: sorry I was not asking for why the -1, I was asking why we are flipping the direction here?

eturquin: Oh right... XD
It's a back diffuse lobe, so you generate a Lambert BRDF direction as usual, and then you just flip it to the lower hemisphere. Because the lobe is azimuthally invariant, all that matters is that they share the same dot product with the normal, so that they have same pdf and value. If we were generating it in the tangent frame, we could just flip the z coord, but because it is directly in world space we flip the whole vector.

sebastienlagarde:
@waiting for @remi-chapelain  review

sebastienlagarde:
@eturquin could you add a test for fabric and path tracing? thanks

remi-chapelain (approved):
Aside for a very specific use cases that have been discussed with @eturquin, the path tracer fabric material version is really on par with what's happening in raster ✔️ 

Tested material sample, HDRP fabric sample and Civic center fabric material. ✔️ 

Will take a quick look at the test scene when pushed. 

693509258465d53394b3b19def83c942

sebastienlagarde:
Other question. @eturquin  we need to add in the What's new in 12.0 that we now support fabric shader, could you take the screenshots / test scene from Remi and add it to it? thanks

eturquin:
> Other question. @eturquin we need to add in the What's new in 12.0 that we now support fabric shader, could you take the screenshots / test scene from Remi and add it to it? thanks

Yes can do. I've also added a test... but there is something slightly fishy going on, the results seem non-deterministic with 1 sample (not sure it is linked to fabric in particular though). I'm double-checking with Remy.

remi-chapelain:
I think I fixed the test stability, passes everytime locally now. 
[relaunched yamato.](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Fpt_merge_fabric/.yamato%252Fall-hdrp_dxr.yml%2523PR_HDRP_DXR_trunk/6035363/job)

remi-chapelain (approved):
Test is now passing on yamato. Good for merge on my side. ✔️ 

_Side note : the fact that it was unstable before showed that there is probably an issue in how the directional light list is created. The order is probably not the same every time lighting is evaluated giving not deterministic results when there's more than one dir light. @eturquin is looking into it but not related to the PR._

PR #3915, opened 2021-03-18
Cleanup URP shader UI

adrien-de-tocqueville:
### Purpose of this PR

Follow up to PR #3730 and #3462 for URP.
This PR cleans up and removes a lot of duplicate or unecessary code in the URP shader inspectors
The aim is mainly to prepare the material variants work.

At the same time, i also did the equivalent for URP of the PR #3699 to fix materials being always dirty in the editor.
Main change is that it doesn't clear keywords at every reset anymore. Technically this could break things if the keyword reset code assumed keyword list is empty, but looking through the code and doing some test, it doesn't seem to be the case.
Keywords are still cleared when shader changes to avoid an ever growing list

It also contains minor bugfixes:
* The `Clear Coat` checkbox now works correclty with multi editing
* fix `doubleSidedGI` when changing cull mode

---
### Testing status

Did some general testing of the modified properties in the editor, with and without multi editing
Verified the list of keywords after various UI changes


com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -368,16 +332,28 @@ protected static void DrawTileOffset(MaterialEditor materialEditor, MaterialProp
...
             if (material.HasProperty("_ReceiveShadows"))
                 CoreUtils.SetKeyword(material, "_RECEIVE_SHADOWS_OFF", material.GetFloat("_ReceiveShadows") == 0.0f);
 
+            // Setup double sided GI
+            bool doubleSidedGI = (RenderFace)material.GetFloat("_Cull")  != RenderFace.Front;
+            if (doubleSidedGI != material.doubleSidedGI)
+                material.doubleSidedGI = doubleSidedGI;
+
+            // Temporary fix for lightmapping. TODO: to be replaced with attribute tag.
+            if (material.HasProperty("_MainTex"))
```
eh-unity, line 348: There is [MainTexture] and [MainColor] attributes these days.
material.mainTexture, etc.
material.color
Should work these days. If not then it's a bug. A shader missing the attributes or the lookup is done incorrectly.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -486,57 +467,39 @@ public static void SetupMaterialBlendMode(Material material)
...
-                (rect.width - preLabelWidth) * 0.5f, EditorGUIUtility.singleLineHeight);
+                (rect.width - preLabelWidth) * 0.5f - 1, EditorGUIUtility.singleLineHeight);
+            EditorGUI.BeginChangeCheck();
+            EditorGUI.showMixedValue = prop1.hasMixedValue;
             var prop1val = EditorGUI.FloatField(propRect1, prop1Label, prop1.floatValue);
+            if (EditorGUI.EndChangeCheck())
+                prop1.floatValue = prop1val;
 
-            Rect propRect2 = new Rect(propRect1.x + propRect1.width, rect.y,
+            Rect propRect2 = new Rect(propRect1.x + propRect1.width + 2, rect.y,
```
eh-unity, line 486: Should have constant for whatever 2 is and make it `+ <constantName>`

com.unity.render-pipelines.universal/Editor/ShaderGUI/ParticleGUI.cs
```diff
@@ -166,18 +166,10 @@ public static void FadingOptions(Material material, MaterialEditor materialEdito
             {
                 // Soft Particles
                 {
-                    EditorGUI.showMixedValue = properties.softParticlesEnabled.hasMixedValue;
-                    var enabled = properties.softParticlesEnabled.floatValue;
+                    materialEditor.ShaderProperty(properties.softParticlesEnabled, Styles.softParticlesEnabled);
+                    var enabled = properties.softParticlesEnabled.floatValue >= 0.5f;
```
eh-unity, line 170: Personally I don't think >= is needed. 0 == false, true otherwise is well defined.

eh-unity (commented):
Just some quick comments. Overall LGTM.

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
adrien-de-tocqueville: Actually the attributes are there no the shader but the lightmapper doesn't use them, it accesses directly _MainTex and co
So we need to duplicate the properties value
Also note that this code was already present, i just moved it to this function to make sure it's executed when needed

phi-lira (approved):

erikabar (approved):
tested all property types, multi-editing, undo, redo. everything works as expected.

PR #3919, opened 2021-03-18
Fix sss in planar reflections planar

adrien-de-tocqueville:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1319027/
https://fogbugz.unity3d.com/f/cases/1322320/

SSS materials were appearing black in planar reflection with certain camera positions.
Transform the depth buffer value to linear space before doing depth comparison

---
### Testing status

tested the 2 repro projects


sebastienlagarde:
Added @anisunity as he have look in depth (no pun intented :) ) the issue with oblique matrix and depth buffer

com.unity.render-pipelines.high-definition/Runtime/Material/SubsurfaceScattering/SubsurfaceScattering.compute
```diff
@@ -374,8 +376,8 @@ void SubsurfaceScattering(uint3 groupId          : SV_GroupID,
     // Reconstruct the view-space position corresponding to the central sample.
     float2 centerPosNDC = posInput.positionNDC;
     float2 cornerPosNDC = centerPosNDC + 0.5 * _ScreenSize.zw;
-    float3 centerPosVS  = ComputeViewSpacePosition(centerPosNDC, centerDepth, UNITY_MATRIX_I_P);
```
anisunity, line 377: If i understand correctly you are decompressing the oblique depth with the non-oblique matrix right? If yes this is incorrect

anisunity: @adrien-de-tocqueville

adrien-de-tocqueville: yes that was indeed incorrect, I think the actual fix is simpler than that

com.unity.render-pipelines.high-definition/Runtime/Material/SubsurfaceScattering/SubsurfaceScattering.compute
```diff
@@ -107,12 +107,12 @@ float4 LoadSampleFromVideoMemory(int2 pixelCoord)
 // Returns {irradiance, linearDepth}.
 float4 LoadSample(int2 pixelCoord, int2 cacheOffset)
 {
+    float4 value;
```
sebastienlagarde, line 110: just to confirm, you have try different compile option and this wasn't compiling and you ahve fix warning?

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/Material/SubsurfaceScattering/SubsurfaceScattering.compute
adrien-de-tocqueville: yes it's for the SSS_USE_LDS_CACHE define

iM0ve (approved):
Testing on these 2 repro projects + automated tests. Should be sufficient for the change

iM0ve:
Also @adrien-de-tocqueville do you think its worth to add an automated test for this specific case?

adrien-de-tocqueville:
> Also @adrien-de-tocqueville do you think its worth to add an automated test for this specific case?

@remi-chapelain wrote that in his fogbug, I think he will take care of it:
>Note to self : Update HDRP Runtime tests scene 001 to include the SSS Head in the planar reflection there to avoid future regression !

sebastienlagarde:
Yes @remi-chapelain mention me that he will modify an existing test (we should modify the runtime 001 test to see the character with skin in the mirror

PR #3928, opened 2021-03-19
[HDRP][Path Tracing] Small normal fix for light list creation

eturquin:
Before:
When creating the list of active lights, we were testing against the geometric normal.
Now:
We test against the shading normal, which will remove some hard shadow termination issues on coarsely tessellated objects, with directional lights.
There were minor changes in test ref images, which have been updated.


remi-chapelain:
Could you add what you tested and what needs testing (if it does) with this small change ? :)

sebastienlagarde (approved):

remi-chapelain (approved):
Quickly sanity checked a few project I had opened. 
Also checked with the default unity sphere which is a good candidate for a coarse tesselated object and indeed it helps a bit to reduce some artefacts. 

93293082425b6aeb267397ef84213caa

PR #3931, opened 2021-03-19
Temporarily disabled warnings in Core shader library for gles platforms.

JulienIgnace-Unity:
### Purpose of this PR
This PR temporarily disables warnings in Core shader library for gles platforms.
The proper fixes will be done later as we start using those functions for real on gles platforms.
https://fogbugz.unity3d.com/f/cases/1316887/

---
### Testing status
Tested the repro steps from the linked case.


com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -1,6 +1,8 @@
 #ifndef UNITY_COMMON_INCLUDED
 #define UNITY_COMMON_INCLUDED
 
+#pragma warning (disable : 3205) // conversion of larger type to smaller
```
sebastienlagarde, line 4: Rather than disabling for the world, I will suggest to 
#pragma warning (disable : 3205)  / #pragma warning (enable : 3205) on each required file

phi-lira (approved):

sebastienlagarde (approved):

stramit (approved):

PR #3933, opened 2021-03-19
[7.x.x] Fixed issue that caused XR HDR to not work correctly.

phi-lira:
### Purpose of this PR
 Fixed case https://issuetracker.unity3d.com/issues/xr-urp-emission-effect-does-not-work-when-in-play-mode-and-xr-is-enabled

### Testing status
 Tested locally with MockHMD

---
### Comments to reviewers
 The issue was caused because the graphics format was never changed on the stereo path. 
 It's a little bit hard to see in the diff but I change de if structured and make sure to not apply scale on stereo path.


thomas-zeng (approved):
Thanks felipe! LGTM

JasonCostanza (approved):
I verified image expectations with Felipe over slack and it checked out.

PR #3934, opened 2021-03-19
Setup the base foundation project to only run a few tests

joshua-davis:
Updated the lit shader graph and reference surface shader.
Replicated some materials that were in Universal.
Set several materials to use shader graph variants.
Reverted the universal material that was accidentally modified.
Set up the filters to only run a sub-set of tests.


jessebarker (approved):
LGTM

PR #3935, opened 2021-03-19
Sg builtin target test project

jessebarker:
### Purpose of this PR

Adding Yamato metafiles, as well as the generated yml configs for the new built-in target foundation and lighting test projects.

---
### Testing status
Triggered yamato jobs for both.
Testing results match expected (not necessarily passing, but in known state)

---
### Comments to reviewers


joshua-davis (approved):
Not sure if we need nightlies, but otherwise seems fine from what I can parse of this

jessebarker:
> Not sure if we need nightlies, but otherwise seems fine from what I can parse of this

We don't yet, but will when we merge into master.

PR #3937, opened 2021-03-19
Platform/ios/astc default texture format

hkr:
iPhone reference image updates.
The related changes to core Unity did not land yet


tomzig16 (approved):

ellioman (approved):
Looks good. Yamato failures are known and not related to this PR.

PR #3938, opened 2021-03-19
Fix return values from GetStereoProjectionMatrix() and SetStereoViewMatrix()

fabien-unity:
---
### Purpose of this PR
- fix for https://fogbugz.unity3d.com/f/cases/1312813/
- update stereo matrices before the event `BeginCameraRendering` by calling `camera.SetStereoProjectionMatrix` and `camera.SetStereoViewMatrix` with values from the universal render pipeline

---
### Testing status
Local tests with repro project on Win10 + Oculus + SPI


erikabar (approved):
Already tested by developer with user project, XR QA is added so I am approving it from URP QA side

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -540,6 +541,7 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
 
                         using (new ProfilingScope(null, Profiling.Pipeline.beginCameraRendering))
                         {
+                            UpdateCameraStereoMatrices(baseCameraData);
```
thomas-zeng, line 544: I think we should call UpdateCameraStereoMatrices after InitializeAdditionalCameraData. And for overlay camera, shouldn't we use overlay camera data here?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -506,6 +506,7 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
 #endif
             using (new ProfilingScope(null, Profiling.Pipeline.beginCameraRendering))
             {
+                UpdateCameraStereoMatrices(baseCameraData);
```
thomas-zeng, line 509: I would suggest to call UpdateCameraStereoMatrices outside of the beginCameraRendering profiling scope.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
fabien-unity: Good point, done!

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
fabien-unity: Good point, done!

phi-lira (approved):

thomas-zeng (approved):

DennisDeRykeUnity:
@fabien-unity On Slack you indicated that the failed automated tests also failed for master.  I see that you tested using the repro bug project on Oculus.  Do you think this is sufficient testing for this PR?  If not, then tell me what likely pain points I should test myself (which features, what platforms, etc.).

fabien-unity:
@DennisDeRykeUnity Could you also test on Quest to make sure there's no regressions ?

DennisDeRykeUnity (approved):
I see no regressions using the following configuration.  Based on this plus Fabien's own testing, I approve. 

- Unity (trunk 5be3997fdbde)
- URP (universal/xr/legacy-stereo-api 738c793)
- Vulkan (also OpenGLES3)
- MultiView
- URP Template scene
- Quest

PR #3943, opened 2021-03-19
Added clone of URP foundation tests for built-in

joshua-davis:
This should only add files in the BuiltInGraphicsTest_Foundation folder.
These are the URP tests with only some filters to limit the tests, few materials changes, and some shader graph shaders added


jessebarker (approved):
LGTM

PR #3950, opened 2021-03-22
[VFX] Fix compilation with builtin renderpipeline 

PaulDemeulenaere:
### Purpose of this PR
Fix compilation issue with built-in render pipeline when `UnityCG.inc` and `com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl` are both included.

- UnityCG.inc is needed for UnityWorldToClipPos, UnityApplyLinearShadowBias but also variable like UNITY_MATRIX_VP.
- Common.hlsl is always included in [VFXCommon.hlsl](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.visualeffectgraph/Shaders/VFXCommon.hlsl#L2)

---
### Testing status
Tested repro project locally from https://fogbugz.unity3d.com/f/cases/1308481/
_after_small

---
### Comments to reviewers
I'm not sure we should really do this : BRP isn't officially supported with VFXGraph.
This issue has been introduced by this change : https://github.com/Unity-Technologies/Graphics/pull/3443


PaulDemeulenaere:
Actually, we won't fix this.

PR #3954, opened 2021-03-22
Fixed up the two main sub-shaders to have the same passes (some were …

joshua-davis:
This PR was primarily created to get the two ShaderGraph tests in foundation to pass.

The root issue was that the shadow caster pass was incorrectly declared. One issue was that it was accidentally set with a pragma to only run in gl variant renderers. This caused a weird bug where the URP shadow caster was running instead of the built-in one. 

After this was fixed, a few minor bugs in the built-in shadow caster pass were addressed, mostly around pragmas and not running alpha clipping.

Additionally, the two sub-shaders were cleaned up a bit to at least have the same set of built-in passes.

With this PR, both foundation and lighting should now pass with the enabled tests (at least they did locally).

### DIFF
--- a/com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/Editor/ShaderGraph/Targets/BuiltInTarget.cs
+++ b/com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/Editor/ShaderGraph/Targets/BuiltInTarget.cs
@@ -561,14 +561,14 @@ namespace UnityEditor.Rendering.BuiltIn.ShaderGraph
     {
         public static readonly PragmaCollection Default = new PragmaCollection
         {
-            { Pragma.Target(ShaderModel.Target20) },
+            { Pragma.Target(ShaderModel.Target30) },
             { Pragma.Vertex("vert") },
             { Pragma.Fragment("frag") },
         };
 
         public static readonly PragmaCollection Instanced = new PragmaCollection
         {
-            { Pragma.Target(ShaderModel.Target20) },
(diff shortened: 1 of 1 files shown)


jessebarker (approved):
LGTM

com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/ShaderLibrary/Shim/HLSLSupportShim.hlsl
```diff
@@ -97,7 +97,13 @@
 #define UNITY_DECLARE_TEXCUBE_SHADOWMAP(tex) TEXTURECUBE_SHADOW(tex); SAMPLER_CMP(sampler##tex)
 #define UNITY_SAMPLE_SHADOW(tex,coord) SAMPLE_TEXTURE2D_SHADOW(tex, sampler##tex, coord)
 #define UNITY_SAMPLE_SHADOW_PROJ(tex,coord) SAMPLE_TEXTURE2D_SHADOW(tex, sampler##tex, (coord.xyz / coord.w))
-#define UNITY_SAMPLE_TEXCUBE_SHADOW(tex,coord) SAMPLE_TEXTURECUBE_SHADOW(tex, sampler##tex, coord)
+
+#if defined(SHADER_API_GLCORE) || defined(SHADER_API_GLES3) || defined(SHADER_API_VULKAN) || defined(SHADER_API_SWITCH)
```
joshua-davis, line 101: This seems to be what caused the vulkan compiler error

PR #3958, opened 2021-03-23
[HDRP] Decal scaling

AndrewSaraevUnity:
### Checklist for PR maker

---
### Purpose of this PR
This adds the Scale Mode option to Decal Projector. The "Inherit from Hierarchy" mode makes the decal rendering respect the transform scale allowing decals to be part of a prefab that is placed and scaled in a scene. Decal rendering doesn't currently support skewing so this option uses lossyScale and the shape behaves similarly to the shape of Box Collider.

Decal Projector Scale Mode

---
### Testing status
Tested manually in the editor. In the "Inherit from Hierarchy" mode the rendering, the inspector and handles support zero and nagative transform scale on any axis.

---
### Comments to reviewers
The scale is applied exclusively on the Decal Projector side by mimicking its effect with the rotation and the size feeded to the Decal System. Maybe it'd be better to update Decal System so it would be aware of the scale.

The change in the documentation is requred but want to confirm the look and the name of the property in the Inspector first.


sebastienlagarde:
Thansk for doing this PR

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
```diff
@@ -278,7 +280,8 @@ void OnSceneGUI()
 
         void DrawBoxTransformationHandles(DecalProjector decalProjector)
         {
-            using (new Handles.DrawingScope(fullColor, Matrix4x4.TRS(decalProjector.transform.position, decalProjector.transform.rotation, Vector3.one)))
+            Vector3 scale = decalProjector.effectiveScale;
+            using (new Handles.DrawingScope(Color.white, Matrix4x4.TRS(decalProjector.transform.position, decalProjector.transform.rotation, scale)))
```
RSlysz, line 284: Why the need for color change?

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
```diff
@@ -363,22 +398,31 @@ void DrawUVHandles(DecalProjector decalProjector)
...
+                            float decalSize = decalProjector.size[channel];
+                            float limit = k_LimitInv * decalSize;
+                            if (handleSize > limit || handleSize < -limit)
+                            {
+                                uvScale[channel] = decalSize / handleSize;
+                                uvBias[channel] = minusNewUVStart / handleSize;
+                            }
+                            else
+                            {
+                                // TODO: Decide if decalProjector.size and uvHandles.size should ever have negative values. They can't currently.
```
RSlysz, line 417: I don't think it is a good Iidea. As you said there is already a missing negative UV. I see no interest to have both.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
```diff
@@ -568,6 +616,8 @@ public override void OnInspectorGUI()
                     ReinitSavedRatioSizePivotPosition();
                 EditorGUI.EndProperty();
 
+                EditorGUILayout.PropertyField(m_ApplyScale, k_ApplyScale);
```
RSlysz, line 619: As it affect the size of decal projector, should it be first field? Size line is not directly affected but highly depend on this to know if you need to take into account the hierarchy or not.

RSlysz (changes requested):
Check the color change on the handles. I don't think this is wanted.

Otherwise, thanks for the fix when size is 0 on an axis. The remaining seams good too. There is no need for data migration either. 👍 
Naming seams good. The tooltips don't follow what we do for checkbox usually but we can change it. I'm more concerned about the place it appears in the Inspector. I'll check with UXs for this.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
AndrewSaraevUnity: Sorry, it's my mistake in porting it from a version of decals without custom gizmo colors. Thanks for noticing.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
AndrewSaraevUnity: Can you clarify: Not a good idea to ever have negative values in these properties? Or not a good idea to remove the multiplication by their signs, just in case they will become negative someday?

I can imagine the need in negative UV tiling that may require allowing negative size in these handles.

RSlysz: no I mean: if we add the negative value for UV and we will try as it can be usefull, I don't see interest to also have negative value for size. Seams kind of a useless complication to me.

Why negative UV tilling should recquire negative size in handles? I think I miss something here.

RSlysz:
Ok so after discutions with UX, as the apply scale don't change value but affect the usage of value for Size, Projection Depth, Pivot, Tilling and Offset it should appears above. And it is more like an edition mode so we arrive to this conclusion:

It should be a dropdown on the first field called **Scale Mode** which provide:
- **Scale Invariant**
- **Inherit from Hierarchy**

while **Scale Invariant** should be the default selected.

And the tooltips (correct me @JordanL8 ) should be more like this:
"Chose the scaling mode you want this Decal use for the projector. Scale Invariant ignore the transformation hierarchy and always directly use the scale values entered in this Decal inspector. Inherit from Hierarchy can produce display that won't correspond to actual hierarchy transformation when it is skewed as the projector will always be an orthogonal projection, but it will try to multiply scale value entered in this Decal inspector by the lossy scale of the Transform."

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
AndrewSaraevUnity: When UV tiling is positive and user would want to flip it with handles by moving an edge from one side to another the movement will stop at the zero rect size, won't it? Because uvHandles.size will ensure that its size after editing is non-negative.
https://github.com/Unity-Technologies/Graphics/blob/848d723429c6c35fe19c12350a44a8b55b2b757d/com.unity.render-pipelines.high-definition/Editor/Material/Decal/DisplacableRectHandles.cs#L288

RSlysz: Oh right sorry I was considering the wrong handle from the beginning. Yes totally make sens.

JordanL8:
Damn, I should have read the rest of the conversation before adding the documentation 😄

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
AndrewSaraevUnity: So I can remove `* Mathf.Sign(decalSize)` but keep `* Mathf.Sign(handleSize)`, and remove a todo, right?

JordanL8:
The tooltip should be short and we should describe the options in the documentation instead. Tooltip:
"Specifies the scaling mode to apply to Decals that use this projector."

JordanL8:
For the doc, is this alright?

```
The scaling mode to apply to decals that use this Decal Projector. The options are:<br/>&#8226; **Scale Invariant**: Ignores the transformation hierarchy and uses the scale values in this component directly.<br/>&#8226; **Inherit from Hierarchy**: Multiplies the [lossy scale](https://docs.unity3d.com/ScriptReference/Transform-lossyScale.html) of the Transform with the Decal Projector's own scale then applies this to the decal. Note that since the Decal Projector uses orthogonal projection, if the transformation hierarchy is skewed, the decal does not scale correctly.
```

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
RSlysz: Well If I am not wrong, the DisplaceableRectHandles will not be able to display negative size. It needs a little of work on this too for the handle to works correctly. So for now, I assume that keeping the comment is better and we will check this along the negative UV request. What do you think?

(If I am wrong and still work please do it)

RSlysz:
> the decal does not scale correctly

It is more that angles of projection will not correspond and you can have somewhat of a "displacement" of the projection or some kind of stretching (object stretch but not projection)

but I think it is ok.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
AndrewSaraevUnity: Yes, keeping a comment until the negative UV request makes sense. I'll only remove the decal size sign multiplication for now.

JordanL8:
Okay cool, thanks for the explanation. Although, with 'the decal does not scale correctly', the user will understand from this that 'it won't look correct' so I think its ok not to go into too much detail!

For skewed, I'll add a link to the [Transform](https://docs.unity3d.com/Manual/class-Transform.html) page since we describe it on there.

If those are the final names from UX, I'll edit the doc and add this content. @AndrewSaraevUnity, could you take another screenshot after you've changed the UI? Either let me know and I'll replace the one in the docs with it or you can overwrite Documentation~/Images/DecalProjector6.png yourself

AndrewSaraevUnity:
> If those are the final names from UX, I'll edit the doc and add this content. @AndrewSaraevUnity, could you take another screenshot after you've changed the UI? Either let me know and I'll replace the one in the docs with it or you can overwrite Documentation~/Images/DecalProjector6.png yourself

@JordanL8, thanks. Yes, I'll take a screenshot and update the image in the docs.

JordanL8:
> @JordanL8, thanks. Yes, I'll take a screenshot and update the image in the docs.

Cool! I've pushed the docs changes now. Also when you redo the UI, can you change the tooltip to what was mentioned above?
"Specifies the scaling mode to apply to decals that use this Decal Projector."

JordanL8 (approved):

RSlysz (approved):
Looks great 📗 :)

sebastienlagarde (approved):

TomasKiniulis (approved):
Thanks Andrew, looks stable!

What I've tested:
- Undo/Redo between Scale Modes
- Scale Modes with multiple selected decals
- Scales Modes with Decals parented to an object
- Various transformations with Inherit from Hierarchy mode before and after modified boundaries and pivot position

PR #3959, opened 2021-03-23
[URP] Reflection Probe Blending & Box Projection

svens-unity:
### Purpose of this PR
This PR brings support for reflection probe blending ([URP-40](https://jira.unity3d.com/browse/URP-40)) and box projection ([URP-42](https://jira.unity3d.com/browse/URP-42)) to the Universal Render Pipeline. Reflection probe blending is build upon the built-in sampler. Therefore we are still limited to a **maximum of two reflection probes per object**. However, to improve the quality of blends we have chosen to do the **blending probes per-pixel** instead of per-object. This should improve visual fidelity of the blends, especially for large objects. As the blending weight is determined for each pixel individually. Objects will only ever receive reflections of a probe inside the volume. The blend distance will **no longer extend outwards** of the probe volume. The weight to each probe is determined as follows
1. According to probe importance (desc)
2. According to probe volume (asc)
3. Any remaining weight is assigned to the skybox reflection probe. 

The weight is normalized if both volume and importance are equal and the contributed weight to both probes exceeds one.  Thus creating an equal blend between both probes.

21-03-26_reflection_probes_per_pixel_blend
URP_boxProjection

---
### Documents
- [User documentation of feature & API](https://docs.google.com/document/d/17cm-GKtp4kaRk7_SW6Gm8pPvYYHfP1ku2pyP44vrMjM/edit?usp=sharing) (WIP) 
- [Proposed blending design](https://docs.google.com/document/d/1XmXjehkdtmkywnq8sZ08Bjyyy5Ta06JhJXHz8JapwR8/edit?usp=sharing)
- [Technical design document](https://docs.google.com/document/d/12T3GPGpBdzanGUeWU-_SndkAIpdbuPjupqttL-vOf-I/edit?usp=sharing)

---
### Inspectors/UX

Both features can be enabled or disabled individually using checkboxes on the URP asset. 
image

Furthermore, it enables the blend distance field on the reflection probe inspector. 
image

---
### Testing status
- ABV on ono branch
- SRP ABV on this branch
- QA perf tests
- Added graphics tests to UniversalGraphicsTest_Lighting
 (0154_ReflectionProbes_Blending)
149_ReflectionProbes_Blending
 (0155_ReflectionProbes_BoxProjection)
150_ReflectionProbes_BoxProjection

---
### Comments to reviewers
- Check if we added the keywords for this feature to all shaders. It is possible that we missed some.
- Should we override the behavior of [Reflection Probe Node](https://docs.unity3d.com/Packages/com.unity.shadergraph@10.4/manual/Reflection-Probe-Node.html) in ShaderGraph to return the blended result? 
- For built-in Reflection Probe blending & box projection are disabled on certain platforms. Should we automatically configure this for users too. Or this is up to the user to configure with URPAssets?
- What should the default settings be for this feature on the URPAsset? Off by default will allow for smooth upgrading as it provides the same behavior that is in URP at the moment (with the exception that the Reflection Probes field is removed on the Renderer).

### DIFF
--- a/TestProjects/UniversalGraphicsTest_Lighting/Assets/CommonAssets/UniversalRPAsset.asset
+++ b/TestProjects/UniversalGraphicsTest_Lighting/Assets/CommonAssets/UniversalRPAsset.asset
@@ -42,8 +42,8 @@ MonoBehaviour:
   m_AdditionalLightsShadowResolutionTierLow: 128
   m_AdditionalLightsShadowResolutionTierMedium: 256
   m_AdditionalLightsShadowResolutionTierHigh: 512
-  m_ReflectionProbeBlending: 1
-  m_ReflectionProbeBoxProjection: 1
+  m_ReflectionProbeBlending: 0
+  m_ReflectionProbeBoxProjection: 0
   m_ShadowDistance: 56
   m_ShadowCascadeCount: 4
   m_Cascade2Split: 0.25
@@ -56,6 +56,7 @@ MonoBehaviour:
   m_UseSRPBatcher: 1
   m_SupportsDynamicBatching: 0
   m_MixedLightingSupported: 1
+  m_SupportsLightLayers: 0
   m_DebugLevel: 0
   m_UseAdaptivePerformance: 1
   m_ColorGradingMode: 0
--- a/TestProjects/UniversalGraphicsTest_Lighting/ProjectSettings/EditorBuildSettings.asset
+++ b/TestProjects/UniversalGraphicsTest_Lighting/ProjectSettings/EditorBuildSettings.asset
@@ -1,4 +1,4 @@
-﻿%YAML 1.1
+%YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
 --- !u!1045 &1
 EditorBuildSettings:
@@ -86,11 +86,11 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/Scenes/146_AdditionalLightShadowCustomResolution.unity
     guid: 5cb07294042344649a35ad91c91b539a
-  - enabled: 1
-    path: Assets/Scenes/149_ReflectionProbes_Blending.unity
+  - enabled: 0
+    path: Assets/Scenes/154_ReflectionProbes_Blending.unity
     guid: 563a251f782be254a91472c4db94e506
-  - enabled: 1
-    path: Assets/Scenes/150_ReflectionProbes_BoxProjection.unity
+  - enabled: 0
+    path: Assets/Scenes/155_ReflectionProbes_BoxProjection.unity
     guid: dbc896364427cae40affa52a983b43a2
   - enabled: 1
     path: Assets/Scenes/019_Lighting_Scene_PointLights.unity


com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalTarget.cs
```diff
@@ -766,6 +766,24 @@ static class CoreKeywordDescriptors
             scope = KeywordScope.Global,
         };
 
+        public static readonly KeywordDescriptor ReflectionProbeBlending = new KeywordDescriptor()
```
Nzollner, line 769: Shader graph currently only uses SpecCube0

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -141,6 +145,14 @@ bool StripUnusedFeatures(ShaderFeatures features, Shader shader, ShaderSnippetDa
             if (!IsFeatureEnabled(features, ShaderFeatures.AdditionalLightShadows) && isAdditionalLightShadow)
                 return true;
 
+            //bool isReflectionProbeBlending = compilerData.shaderKeywordSet.IsEnabled(m_ReflectionProbeBlending);
```
Nzollner, line 148: Need to look into how to strip shader variants with reflection probes

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -143,6 +143,10 @@ public partial class UniversalRenderPipelineAsset : RenderPipelineAsset, ISerial
         [SerializeField] int m_AdditionalLightsShadowResolutionTierMedium = 512;
         [SerializeField] int m_AdditionalLightsShadowResolutionTierHigh = 1024;
 
+        // Reflection Probes
```
Nzollner, line 146: Do we want these startup values?
What are others' opinions?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -640,7 +640,8 @@ static void SetSupportedRenderingFeatures()
                 lightProbeProxyVolumes = false,
                 motionVectors = false,
                 receiveShadows = false,
-                reflectionProbes = true,
+                reflectionProbes = false,
```
Nzollner, line 643: Note that we also need to change these values if we want to change the default values.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -1087,6 +1090,10 @@ static void SetupPerFrameShaderConstants()
             Color glossyEnvColor = CoreUtils.ConvertLinearToActiveColorSpace(linearGlossyEnvColor);
             Shader.SetGlobalVector(ShaderPropertyId.glossyEnvironmentColor, glossyEnvColor);
 
+            // Used as fallback cubemap for reflections
```
Nzollner, line 1093: Document that it is needed when 2 probes are blending

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,146 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
 #define SAMPLE_GI(lmName, shName, normalWSName) SampleSHPixel(shName, normalWSName)
 #endif
 
-half3 GlossyEnvironmentReflection(half3 reflectVector, half perceptualRoughness, half occlusion)
+half3 BoxProjectedCubemapDirection(half3 reflectionWS, half3 positionWS, float4 cubemapPositionWS, float4 boxMin, float4 boxMax)
```
Nzollner, line 630: Add documentation for how the reflection probes are handled in this implementation.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,146 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+
+half CalculateProbeVolumeSqrMagnitude(float4 probeBoxMin, float4 probeBoxMax)
+{
+    half3 maxToMin = half3(probeBoxMax.xyz - probeBoxMin.xyz);
+    return dot(maxToMin, maxToMin);
+}
+
+half3 CalculateIrradianceFromReflectionProbes(half3 reflectVector, half3 positionWS, half perceptualRoughness)
+{
+    half probe0Volume = CalculateProbeVolumeSqrMagnitude(unity_SpecCube0_BoxMin, unity_SpecCube0_BoxMax);
```
Nzollner, line 666: Start discussion if reflection probe dominance should be on gpu or cpu

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,146 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+    // A probe is dominant if its importance is higher
+    // Or have equal importance but smaller volume
+    bool probe0Dominant = importanceSign > 0 || (importanceSign == 0 && volumeSign < 0);
+    bool probe1Dominant = importanceSign < 0 || (importanceSign == 0 && volumeSign > 0);
+
+    float desiredWeightProbe0 = CalculateProbeWeight(positionWS, unity_SpecCube0_BoxMin, unity_SpecCube0_BoxMax);
+    float desiredWeightProbe1 = CalculateProbeWeight(positionWS, unity_SpecCube1_BoxMin, unity_SpecCube1_BoxMax);
+
+    // Subject the probes weight if the other probe is dominant
+    float weightProbe0 = probe1Dominant ? min(desiredWeightProbe0, 1 - desiredWeightProbe1) : desiredWeightProbe0;
```
Nzollner, line 681: can we benefit from rewriting the code as branches using [branch], since we only have one probe dominant per object which doesn't change doing a frame.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,146 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+    // If neither is dominant this is not guaranteed - only normalize weights if totalweight exceeds 1.
+    weightProbe0 /= max(totalWeight, 1);
+    weightProbe1 /= max(totalWeight, 1);
+
+    half3 irradiance = half3(0, 0, 0);
+    half3 originalReflectVector = reflectVector;
+    half mip = PerceptualRoughnessToMipmapLevel(perceptualRoughness);
+
+    // Sample the first reflection probe
+    if (weightProbe0 > 0.01)
```
Nzollner, line 696: Should the handling of reflection probes be specified as a function?

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -675,20 +800,20 @@ half3 SubtractDirectMainLightFromLightmap(Light mainLight, half3 normalWS, half3
 }
 
 half3 GlobalIllumination(BRDFData brdfData, BRDFData brdfDataClearCoat, float clearCoatMask,
-    half3 bakedGI, half occlusion,
+    half3 bakedGI, half occlusion, half3 positionWS,
```
Nzollner, line 803: Mention how the addition of posistionWS in functions can lead to breaking users code

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
svens-unity: `[SupportedRenderingFeatures.reflectionProbes](https://docs.unity3d.com/ScriptReference/Rendering.SupportedRenderingFeatures-reflectionProbes.html)` only decides whether the reflection probe option on Renderers should be displayed in the inspector. As we do not support all the modes here we want to opt out on having this displayed in URP. 
`reflectionProbesBlendDistance` on the other hand we need to enable the blend distance field in the ReflectionProbe inspector.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -640,7 +640,8 @@ static void SetSupportedRenderingFeatures()
                 lightProbeProxyVolumes = false,
                 motionVectors = false,
                 receiveShadows = false,
-                reflectionProbes = true,
+                reflectionProbes = false,
+                reflectionProbesBlendDistance = true,
```
svens-unity, line 644: Move this behind a define based on the trunk PR. This should allow the changes to the package work independently from the trunk version.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
Nzollner: I will implement this

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: `probe0Dominant` and `probe1Dominant` could be assigned directly to `unity_SpecCube1_BoxMin.w`.  Thereby avoiding the need to calculate volumes and determine the sign between the two volumes. Using the Mali offline compiler this would solve 3 arithmetic instructions. The trade-off here is that it'd bake the volume importance behavior into the component. It won't be possible for anyone to build blending that is only based on importance alone.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,146 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+    half probe0Volume = CalculateProbeVolumeSqrMagnitude(unity_SpecCube0_BoxMin, unity_SpecCube0_BoxMax);
+    half probe1Volume = CalculateProbeVolumeSqrMagnitude(unity_SpecCube1_BoxMin, unity_SpecCube1_BoxMax);
+
+    float volumeSign = sign(probe0Volume - probe1Volume);
+    float importanceSign = unity_SpecCube1_BoxMin.w;
+
+    // A probe is dominant if its importance is higher
+    // Or have equal importance but smaller volume
+    bool probe0Dominant = importanceSign > 0 || (importanceSign == 0 && volumeSign < 0);
+    bool probe1Dominant = importanceSign < 0 || (importanceSign == 0 && volumeSign > 0);
```
svens-unity, line 675: `probe0Dominant` and `probe1Dominant` could be assigned directly to `unity_SpecCube1_BoxMin.w`.  Thereby avoiding the need to calculate volumes and determine the sign between the two volumes. Using the Mali offline compiler this would solve 3 arithmetic instructions. The trade-off here is that it'd bake the volume importance behavior into the component. It won't be possible for anyone to build blending that is only based on importance alone.

What is the preferred trade-off here?

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
Nzollner: This is not a nice solution because we need to get texture and sampler as arguments.
That is why we have chosen not to do it.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
svens-unity: Keeping the default as false will provide a clean upgrade process - it will look the same as it does in current URP. It does make it harder to discover the feature as users will have to actively enable it first.

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
Nzollner: The shader stripping is now implemented

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLit.shader
```diff
@@ -68,6 +68,8 @@ Shader "Universal Render Pipeline/Terrain/Lit"
             #pragma multi_compile _ _MAIN_LIGHT_SHADOWS _MAIN_LIGHT_SHADOWS_CASCADE _MAIN_LIGHT_SHADOWS_SCREEN
             #pragma multi_compile _ _ADDITIONAL_LIGHTS_VERTEX _ADDITIONAL_LIGHTS
             #pragma multi_compile_fragment _ _ADDITIONAL_LIGHT_SHADOWS
+            #pragma multi_compile_fragment _ _REFLECTION_PROBE_BLENDING
+            #pragma multi_compile_fragment _ _REFLECTION_PROBE_BOX_PROJECTION
```
phi-lira, line 72: Does per-object data work for terrain patches? I had the idea it doesn't.

phi-lira (commented):
I didn't see reflection probe added to SimpleLit? We don't support on that one?
I guess we also should not support on BakedLit, we just need to make sure docs state that.

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLit.shader
Nzollner: It works per pixel. But it will not use box projection because of the uneven floor will give unwanted results with the current implementation of box projection

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -675,20 +800,20 @@ half3 SubtractDirectMainLightFromLightmap(Light mainLight, half3 normalWS, half3
...
+    half3 bakedGI, half occlusion, half3 positionWS,
     half3 normalWS, half3 viewDirectionWS)
 {
     half3 reflectVector = reflect(-viewDirectionWS, normalWS);
     half NoV = saturate(dot(normalWS, viewDirectionWS));
     half fresnelTerm = Pow4(1.0 - NoV);
 
     half3 indirectDiffuse = bakedGI;
-    half3 indirectSpecular = GlossyEnvironmentReflection(reflectVector, brdfData.perceptualRoughness, 1.0h);
+    half3 indirectSpecular = GlossyEnvironmentReflection(reflectVector, positionWS, brdfData.perceptualRoughness, 1.0h);
```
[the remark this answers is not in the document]
Nzollner: Fixed

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -675,20 +800,20 @@ half3 SubtractDirectMainLightFromLightmap(Light mainLight, half3 normalWS, half3
 }
 
 half3 GlobalIllumination(BRDFData brdfData, BRDFData brdfDataClearCoat, float clearCoatMask,
```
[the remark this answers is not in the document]
Nzollner: Now have a override

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
Nzollner: No longer the case with backward compatible overrides

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,146 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
 #define SAMPLE_GI(lmName, shName, normalWSName) SampleSHPixel(shName, normalWSName)
 #endif
 
-half3 GlossyEnvironmentReflection(half3 reflectVector, half perceptualRoughness, half occlusion)
+half3 BoxProjectedCubemapDirection(half3 reflectionWS, half3 positionWS, float4 cubemapPositionWS, float4 boxMin, float4 boxMax)
+{
+    // Is this probe using box projection?
+    [branch] if (cubemapPositionWS.w > 0.0)
```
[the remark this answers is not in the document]
Nzollner: I used PIX to gpu capture using DX12 with and without the branch tag and there was no real performance difference.
Without the branch tag was the branch removed from the compiled shader.

Because there was no measurable performance benefit, we have chosen to remove the tag so the compiler can optimize the code and take advantage of future compiler optimizations which would otherwise be more restricted.

TheoWong-pixel:
Two things to think about. ...
1. If we want move this to the renderer instead of the render pipeline asset as we are trying to keep that agnostic, or should we just move all these settings over together at once later on. 
2. i can see why we would want box projection and probe blending to be off when you convert a project over from built-in but would we want to have these on for any new project that are being made in URP or when a new render pipeline asset / Renderer is made?  seems like both are pretty handy and intuitive to use and features you would just want to work without having to enable in a menu somewhere.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -611,6 +615,16 @@ internal int GetAdditionalLightsShadowResolution(int additionalLightsShadowResol
             return additionalLightsShadowResolutionTierMedium;
         }
 
+        public bool reflectionProbeBlending
```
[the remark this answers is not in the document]
svens-unity: tracked in [docs](https://docs.google.com/document/d/17cm-GKtp4kaRk7_SW6Gm8pPvYYHfP1ku2pyP44vrMjM/edit?usp=sharing). I will push documentation all at once after approval. Resolving comment to keep PR tidy.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -611,6 +615,16 @@ internal int GetAdditionalLightsShadowResolution(int additionalLightsShadowResol
...
+        {
+            get { return m_ReflectionProbeBlending; }
+        }
+
+        public bool reflectionProbeBoxProjection
```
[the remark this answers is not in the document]
svens-unity: tracked in [docs](https://docs.google.com/document/d/17cm-GKtp4kaRk7_SW6Gm8pPvYYHfP1ku2pyP44vrMjM/edit?usp=sharing). I will push documentation all at once after approval. Resolving comment to keep PR tidy.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -40,6 +40,8 @@ namespace UnityEngine.Rendering.Universal
         public NativeArray<VisibleLight> visibleLights;
         public bool shadeAdditionalLightsPerVertex;
         public bool supportsMixedLighting;
+        public bool reflectionProbeBoxProjection;
```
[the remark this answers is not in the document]
svens-unity: tracked in [docs](https://docs.google.com/document/d/17cm-GKtp4kaRk7_SW6Gm8pPvYYHfP1ku2pyP44vrMjM/edit?usp=sharing). I will push documentation all at once after approval. Resolving comment to keep PR tidy.

svens-unity:
> I didn't see reflection probe added to SimpleLit? We don't support on that one?
> I guess we also should not support on BakedLit, we just need to make sure docs state that.
Neither of these were sampling the reflection probes before this PR. Should they?

The best place to document this would be on the pages for each material individually [here](https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@10.4/manual/simple-lit-shader.html)?

com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitAdd.shader
```diff
@@ -62,6 +62,8 @@ Shader "Hidden/Universal Render Pipeline/Terrain/Lit (Add Pass)"
             #pragma multi_compile _ _MAIN_LIGHT_SHADOWS _MAIN_LIGHT_SHADOWS_CASCADE _MAIN_LIGHT_SHADOWS_SCREEN
             #pragma multi_compile _ _ADDITIONAL_LIGHTS_VERTEX _ADDITIONAL_LIGHTS
             #pragma multi_compile_fragment _ _ADDITIONAL_LIGHT_SHADOWS
+            #pragma multi_compile_fragment _ _REFLECTION_PROBE_BLENDING
+            #pragma multi_compile_fragment _ _REFLECTION_PROBE_BOX_PROJECTION
```
svens-unity, line 66: We removed box projection for the TerrainLit shader. Shouldn't we remove it for this shader too?

svens-unity:
> 1. If we want move this to the renderer instead of the render pipeline asset as we are trying to keep that agnostic, or should we just move all these settings over together at once later on.

Spoke with Andre and we agreed it would be best to group them together. Once we move all the settings it won't be too much extra work to move too. This keeps the settings logically grouped in one place for users to find.

> 2. i can see why we would want box projection and probe blending to be off when you convert a project over from built-in but would we want to have these on for any new project that are being made in URP or when a new render pipeline asset / Renderer is made?  seems like both are pretty handy and intuitive to use and features you would just want to work without having to enable in a menu somewhere.

I propose to keep it off by default at first. This provides a smooth upgrade path from older URP versions. The documentation will clearly state how to enable it. Once this has landed and proven to be stable we can considering enabling it by default. We can also consider enabling it by default in the URP template.

ernestasKupciunas (commented):
On mobile devices, it seems to work as intended. Blending works on all tested devices.
Although 1 tiny issue found on Mali-400 GPU once the blending is enabled. Seems like reflected gameObject on the reflective gameObject is a little bit transparent:
Mali400

Full testing doc:
Testing doc: https://docs.google.com/document/d/1NG26QZu47m0B4dtpNuBGc7x-zHAYcyYDdJGtw7f8GQ4/edit#

Tested with:
VLNQA00108, LGE LG Leon 4G LTE, 5.0.1, Snapdragon 410 MSM8916, Adreno 306
VLNQA00126, Samsung Galaxy S5 Mini, 6.0.1, Exynos 3 Quad 3470, Mali-400 MP
AppleTV 4k 64GB, tvOS: 14.3

phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
pbbastian: You could do it as an upgrade, such that new assets have the feature enabled, and old assets get upgraded such that the feature is disabled.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -38,6 +38,10 @@ struct InputData
 half4 _GlossyEnvironmentColor;
 half4 _SubtractiveShadowColor;
 
+half4 _GlossyEnvironmentCubeMap_HDR;
+TEXTURECUBE(_GlossyEnvironmentCubeMap);
```
pbbastian, line 42: Not sure if/how we want to re-arrange this, but the texture and sampler are not part of the constant buffer (this is under the header "Constant Buffers")

com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree8.shader
```diff
@@ -54,6 +54,7 @@ Shader "Universal Render Pipeline/Nature/SpeedTree8"
             #pragma multi_compile _ _MAIN_LIGHT_SHADOWS _MAIN_LIGHT_SHADOWS_CASCADE _MAIN_LIGHT_SHADOWS_SCREEN
             #pragma multi_compile _ _ADDITIONAL_LIGHTS_VERTEX _ADDITIONAL_LIGHTS
             #pragma multi_compile _ _ADDITIONAL_LIGHT_SHADOWS
+            #pragma multi_compile_fragment _ _REFLECTION_PROBE_BLENDING
```
pbbastian, line 57: Is it intentional that there is no box projection for this?

pbbastian (changes requested):
Generally looks good! I only have a few questions :)

com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree8.shader
svens-unity: Yes, the idea is that terrain and speed tree shaders are generally used outdoors. For which box projection isn't very useful. By removing box projection here we might save a few cycles and just blend as if it were a normal probe.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
phi-lira: Question is if we want all new assets to have feature enabled or not? For box projection definitely not, for blending depends on user decision, quality or project. That's why I think we really should have something to ask user what is the quality intended when he is creating a new asset and create settings based on those quality presets.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: Did a performance test on mobile with Ernestas. Difference is not significant and not baking `w` gives more freedom. We will leave it as is.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -342,6 +353,16 @@ void DrawLightingSettings()
...
 
+                EditorGUI.indentLevel--;
+                EditorGUILayout.Space();
+
+                // Reflection Probes
+                EditorGUILayout.LabelField(Styles.reflectionProbesSettingsText);
+                EditorGUI.indentLevel++;
+                EditorGUILayout.PropertyField(m_ReflectionProbeBlendingProp, Styles.reflectionProbeBlendingText);
+                EditorGUILayout.PropertyField(m_ReflectionProbeBoxProjectionProp, Styles.reflectionProbeBoxProjectionText);
+                EditorGUI.indentLevel++;
```
nigeljw-unity, line 364: This ident increment seems redundant. Just remove the decrement below?

com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree8.shader
pbbastian: Is this something that's done elsewhere? I could imagine having terrain and trees indoors.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
svens-unity: Any suggestions on where it should go? Above the comment? I do think it should be in this file.  Also it would be preferred to keep the `_GlossyEnvironmentCubeMap_HDR` close to the texture.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
svens-unity: fixed in 3952967

com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree8.shader
Nzollner: We tried to implement the box projection for the terrain, but the formula to calculate the box projection can't handle uneven terrain.
image
The lines in the terrain are pillars projected from the left side of the scene. And the terrain goes uphill from the right to the left.

The same formula is used for box projection in built-in so I expect the same issue to appear, but it seems like they do allow box projection.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
pbbastian: @phi-lira Constant buffers can only contain values, no textures nor samplers. So they're not in any constant buffer, which is why I think it's weird to have under a "Constant Buffers" heading.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,165 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+    half3 maxToMin = half3(probeBoxMax.xyz - probeBoxMin.xyz);
+    return dot(maxToMin, maxToMin);
+}
+
+half3 CalculateIrradianceFromReflectionProbes(half3 reflectVector, float3 positionWS, half perceptualRoughness)
+{
+    half probe0Volume = CalculateProbeVolumeSqrMagnitude(unity_SpecCube0_BoxMin, unity_SpecCube0_BoxMax);
+    half probe1Volume = CalculateProbeVolumeSqrMagnitude(unity_SpecCube1_BoxMin, unity_SpecCube1_BoxMax);
+
+    float volumeSign = sign(probe0Volume - probe1Volume);
```
nigeljw-unity, line 669: sign shader intrinsic returns an int and we compare to int below. Might also want to cast to int for importance as well.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,165 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+    half3 originalReflectVector = reflectVector;
+    half mip = PerceptualRoughnessToMipmapLevel(perceptualRoughness);
+
+    // Sample the first reflection probe
+    if (weightProbe0 > 0.01)
+    {
+#ifdef _REFLECTION_PROBE_BOX_PROJECTION
+        reflectVector = BoxProjectedCubemapDirection(originalReflectVector, positionWS, unity_SpecCube0_ProbePosition, unity_SpecCube0_BoxMin, unity_SpecCube0_BoxMax);
+#else
+        reflectVector = originalReflectVector;
```
nigeljw-unity, line 701: This assignment isn't necessary

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,165 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+#endif // UNITY_USE_NATIVE_HDR || UNITY_DOTS_INSTANCING_ENABLED
+    }
+
+    // Sample the second reflection probe
+    if (weightProbe1 > 0.01)
+    {
+#ifdef _REFLECTION_PROBE_BOX_PROJECTION
+        reflectVector = BoxProjectedCubemapDirection(originalReflectVector, positionWS, unity_SpecCube1_ProbePosition, unity_SpecCube1_BoxMin, unity_SpecCube1_BoxMax);
+#else
+        reflectVector = originalReflectVector;
```
nigeljw-unity, line 719: This assignment isn't necessary

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,165 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+        irradiance += weightProbe1 * encodedIrradiance.rbg;
+#else
+        irradiance += weightProbe1 * DecodeHDREnvironment(encodedIrradiance, unity_SpecCube1_HDR);
+#endif // UNITY_USE_NATIVE_HDR || UNITY_DOTS_INSTANCING_ENABLED
+    }
+
+    // Use any remaining weight to blend to environment reflection cube map
+    if (totalWeight < 0.99)
+    {
+        reflectVector = originalReflectVector;
```
nigeljw-unity, line 733: Instead of doing assignment, just pass in original copy.

nigeljw-unity (changes requested):

svens-unity:
Shader graph node will be addressed in a follow up PR.

com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree8.shader
nigeljw-unity: I think my take on this is that it should be on the tech artist to know not to use box projection in this case, but I agree that for terrain it makes some sense not to support it for now (until we have tiling), but for speedtree, I think we should add it for now.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,165 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+    irradiance = encodedIrradiance.rgb;
+#else
+    irradiance = DecodeHDREnvironment(encodedIrradiance, unity_SpecCube0_HDR);
+#endif // UNITY_USE_NATIVE_HDR || UNITY_DOTS_INSTANCING_ENABLED
+#endif // _REFLECTION_PROBE_BLENDING
+
+    return irradiance * occlusion;
+#endif // _ENVIRONMENTREFLECTIONS_OFF
+
+    return _GlossyEnvironmentColor.rgb * occlusion;
```
nigeljw-unity, line 771: This return statement needs to be in an else condition for _ENVIRONMENTREFLECTIONS_OFF otherwise we will end up with two return statements.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,165 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+    irradiance = encodedIrradiance.rgb;
 #else
-    half3 irradiance = DecodeHDREnvironment(encodedIrradiance, unity_SpecCube0_HDR);
-#endif
+    irradiance = DecodeHDREnvironment(encodedIrradiance, unity_SpecCube0_HDR);
+#endif // UNITY_USE_NATIVE_HDR || UNITY_DOTS_INSTANCING_ENABLED
 
     return irradiance * occlusion;
-#endif // GLOSSY_REFLECTIONS
+#endif // _ENVIRONMENTREFLECTIONS_OFF
```
nigeljw-unity, line 789: Same is true here but this preexisted your changes, but probably good to fix.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -627,21 +627,165 @@ half3 SampleLightmap(float2 lightmapUV, half3 normalWS)
...
+
+    return irradiance * occlusion;
+#endif // _ENVIRONMENTREFLECTIONS_OFF
+
+    return _GlossyEnvironmentColor.rgb * occlusion;
+}
 half3 GlossyEnvironmentReflection(half3 reflectVector, half perceptualRoughness, half occlusion)
 {
 #if !defined(_ENVIRONMENTREFLECTIONS_OFF)
+    half3 irradiance = half3(0, 0, 0);
```
nigeljw-unity, line 776: Is the zero assignment here necessary? I think you can just declare it since we always assign below. These type of redundant assignments/memsets can build up.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
svens-unity: For the constants buffer are you ok with leaving as is? To me this feels like the right place. There is other stuff under there that isnt in the constant buffer either (the structured buffer for lightdata as an example). I can change the name of the comment if you have any suggestions.

com.unity.render-pipelines.universal/Shaders/Nature/SpeedTree8.shader
svens-unity: Enabled box projection for SpeedTree8 in 742303f. Terrain still has it disabled.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: Resolved in 203240f

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: resolved in 203240f

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: resolved in 203240f

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: resolved in 203240f

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: Compiler should handle this fine. But I do agree this creates much cleaner code. I have added an `else ` in 203240f

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: resolved in 203240f

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
svens-unity: Having both disabled by default poses the least amount of risk right now. A future PR will set up suitable defaults for every project. For the time being users will have to opt-in.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
svens-unity: resolved in 203240f

ernestasKupciunas (approved):
Besides one minor bug on Mali-400 device, I have no issues with this PR

Verasl (approved):
LGTM
Can me make sure to make a task somewhere for the Shader Graph node work, so it is not forgotten?

VitaSkruibyte (approved):
Looks good, couldn't find any outstanding issues related to PR. Here's the main [test document](https://docs.google.com/document/d/1vP5TpNG1B8cRoE-Dm6ntsCgn6Kr_m0U28kLo2rjTFis/edit?usp=sharing) (in comment section added another test doc that covers some untested areas that are mentioned in the main doc + some typical usage testing)

VitaSkruibyte:
Approving on behalf of the @gfx-qa-urp team, don't need to wait for the team tag's review ;)

pbbastian (approved):

erikabar (approved):

nigeljw-unity (approved):
LGTM!

kroatoa:
Hi,
I'm assuming this relates to https://unity3d.atlassian.net/browse/UR-2016. That roadmap item is missing the Product manager approval, so if you could poke the relevant one to sign off, that would be good. Otherwise we risk adding the delay for that approval to the weekly bundling.

svens-unity:
Correct, the related Jira ticket on the release board is https://unity3d.atlassian.net/browse/UR-2016. Approval should not be in order. @kroatoa

kroatoa (approved):
Thanks!

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
svens-unity: Define won't be necessary because we are in core now.

PR #3964, opened 2021-03-23
Dirtying an asset that was upgraded to a new Version

jenniferd-unity:
---
### Purpose of this PR
Currently, when HDRP assets go through migration after upgrading, they are not automatically dirtied.
This causes assets being changed in memory but not saved to disk even after a "Save Project" action.
This PR now makes sure that when an asset is migrated to a newer version, it is dirty so the "Save Project" action saves those changes to file.

---
### Testing status
On Windows, I took the Spaceship demo project (master branch) and upgraded to 2021.2.0a9 with my changes.
When opening the project and then saving it, we can now see ON DISK that both the HDRP Global Settings asset is created with up-to-date values and the HDRP Asset assigned to Graphics is also changed to latest version.

---
### Comments to reviewers
This PR may be needed asap to fix issues when people don't manually save the changes.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -111,6 +111,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed preview camera updating the skybox material triggering GI baking (case 1314361/1314373).
 - The default LookDev volume profile is now copied and referened in the Asset folder instead of the package folder.
 - Fixed SSS on console platforms.
+- Assets going through the migration system are now dirtied.
 
 ### Changed
 - Changed Window/Render Pipeline/HD Render Pipeline Wizard to Window/Rendering/HDRP Wizard


com.unity.render-pipelines.high-definition/Runtime/Core/Migration/MigrationDescription.cs
```diff
@@ -104,10 +104,16 @@ public bool Migrate(TTarget target)
             }
 
 #if UNITY_EDITOR
-            // Special in prefab case
-            if (target is UnityEngine.Object && UnityEditor.PrefabUtility.IsPartOfNonAssetPrefabInstance(target as UnityEngine.Object))
+            UnityEngine.Object targetObject = target as UnityEngine.Object;
+            if (targetObject != null)
```
RSlysz, line 108: I believe this should never happens but yes better check this. 👍

RSlysz (approved):

com.unity.render-pipelines.high-definition/Runtime/Core/Migration/MigrationDescription.cs
jenniferd-unity: same, the code was already checking for it and i was on the safe side :)

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

JMargevics (approved):
Tested on MacOS:
Made a template in 2021.2.0a9/HDRP 12 and migrated to a10/This branch. Then i changed HDRP global settings, saved the project and verified that HDRenderPipelineGlobalSetttings.asset has been modified ✅

PR #3966, opened 2021-03-23
[XPipeline]Consolidation for Material's Priority in URP and HDRP

alex-vazquez:
### Purpose of this PR

[JIRA](https://jira.unity3d.com/browse/XPIPELINE-89)

Unification of the sorting priority property for both pipelines

HDRP
image

URP:
image

---
### Testing status

Check old material saved with old range, that the range is not modified even the inspector is opened

image

It is only modified when the priority is changed.

HDRP:
- Check that the UI clamps the values between the new ranges
- Check old material with value outside the range
- Check that the property is only shown with the surface type is advanced
- Check the new location on shadergraph

URP:
- Check the field name

---
### Comments to reviewers
The RenderQueues mantain the same ranges, we are only limiting on UI. This makes us avoid remapping materials render queues if the sorting priority was higher than 50 or lower than -50

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2021.1-to-2021.2.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2021.1-to-2021.2.md
@@ -22,4 +22,4 @@ From 2021.2, the HDRP Asset assigned in the Graphics Settings no longer acts as 
 
 ### Transparent Surface Type
 
-If you had materials with **Transparent Surface Type**, with a sorting priority that is excceding the new range of -50/50, you will need to remap them to the new range values. This will not affect you until you edit the sorting priority property on the material inspector.
+From 2021.2, the range for **Sorting Priority** values has decreased from between -100 and 100 to between -50 and 50. If you used transparent materials (**Surface Type** set to **Transparent**) with a sorting priority lower than -50 or greater than 50, you must remap them to within the new range. HDRP does not clamp the Sorting Priority to the new range until you edit the Sorting Priority property.


jenniferd-unity (approved):

alex-vazquez:
@sebastienlagarde . I've double check with @TheoWong-pixel  and we decided to leave the sorting priority where it is.

I've checked the sorting priorty saved with the old ranges, and even the inspector is opened, the value is not changed until the user modifies it. Do I still need to update the Upgrade Guide?

sebastienlagarde:
> @sebastienlagarde . I've double check with @TheoWong-pixel and we decided to leave the sorting priority where it was. Under the surface options and no in Advanced.
> 
> I've checked the sorting priorty saved with the old ranges, and even the inspector is opened, the value is not changed until the user modifies it. Do I still need to update the Upgrade Guide?

Ah good to heard. So yes, please add information in the upgrade guide to explain this behavior. But good to heard it will not break anything.

sebastienlagarde (approved):
Wiating for upgrade guide

phi-lira:
Btw since you are looking at this I think we should consolidate the tooltip as well. I recently found that the tooltip for URP contained wrong information. Elements with lower priority are renderer first (priority is somewhat confusing term imo)
 
Screenshot 2021-03-18 at 16 00 34

com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -91,7 +91,7 @@ protected class Styles
             public static readonly GUIContent fixNormalNow = EditorGUIUtility.TrTextContent("Fix now",
                 "Converts the assigned texture to be a normal map format.");
 
-            public static readonly GUIContent queueSlider = EditorGUIUtility.TrTextContent("Priority",
+            public static readonly GUIContent queueSlider = EditorGUIUtility.TrTextContent("Sorting Priority",
                 "Determines the chronological rendering order for a Material. High values are rendered first.");
```
phi-lira, line 95: While you are looking at it we should fix this tooltip. Materials with lower value are rendered first.

alex-vazquez:
@JordanL8 could you check the upgrade guide please? Thank you :)

alex-vazquez:
@Unity-Technologies/gfx-qa-hdrp team, could you do a quick check over a material with a transparent sorting priority saved with higher values and see that the values are not modified until you edit the property on the inspector?

@Unity-Technologies/gfx-qa-urp team, could you have a look at the name and the new tooltip?

Thank you awesome QA teams!

RemyUnity:
As well as consolidating the tooltip, wouldn't it be a good idea to consolidate the setting position in both pipelines ? It is in surface options for HDRP, but advanced for URP.

alex-vazquez:
> n surface options for HDRP, but advanced for URP.

On HDRP is really tight to the Transparent Surface Type, on URP is for all the materials, we thought to put it on advanced, as it is an advanced property but as the whole HDRP pipeline is advanced It make sense to leave it there with the change of the surface. 

Besides that, on URP we leave it on Advanced as I said it is an advanced option, that not every user will modify.

If you want more information, @TheoWong-pixel, can give you a UX point of view. Thank you for your heads up :)

iM0ve (approved):
I think the testing is sufficient for the change. Especially when RenderQueues ranges are the same and this is mostly a limitation on UI

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2021.1-to-2021.2.md
```diff
@@ -17,3 +17,9 @@ ClearFlag.Depth does not implicitely clear stencil anymore. ClearFlag.Stencil ad
 ## HDRP Global Settings
 
 From 2021.2, the HDRP Asset assigned in the Graphics Settings no longer acts as the default Asset for HDRP. A new HDRP Global Settings Asset now exists to save settings unrelated to which HDRP Asset is active.
+
+## Materials
+
+### Transparent Surface Type
+
+If you had materials with **Transparent Surface Type**, with a sorting priority that is excceding the new range of -50/50, you will need to remap them to the new range values. This will not affect you until you edit the sorting priority property on the material inspector.
```
JordanL8, line 25: `From 2021.2, the range for **Sorting Priority** values has decreased from between -100 and 100 to between -50 and 50. If you used transparent materials (**Surface Type** set to **Transparent**) with a sorting priority lower than -50 or greater than 50, you must remap them to within the new range. HDRP does not clamp the Sorting Priority to the new range until you edit the Sorting Priority property.`

JordanL8 (changes requested):

JordanL8 (approved):

phi-lira (approved):

sebastienlagarde (approved):

simon-engelbrecht-soerensen (approved):
LGTM on URP

PR #3969, opened 2021-03-23
Set metallic material to double sided for transparent samples

remi-chapelain:
### Purpose of this PR
Currently, in the HDRP material samples, we can't see the backside of the aluminium in the transparent Unity Ball setup. 

What this PR does, is duplicating the "Aluminium Satin" material with a new double sided one. This new material is now assigned to the transparent Unity Ball setup (the fives in the material scene and the hair ball).

Also, when [this issue](https://fogbugz.unity3d.com/f/cases/1323451/) will be taken care of, the transparent setup for ray tracing (pathtracer included) will be more correct. With one of the multi mat object being double sided, the whole object will be considered double sided and results will look more like expectations.

**Materials scene** _(only transparents to see the change better, not actual setup)_
b5f8c43c8b657fd5ef9a5500c22378a9

**Hair Scene**
578ac44f8c7d11673611c6eec8eaed71

### Testing status
Simply tested that importing samples had the new material and that the scenes includes the good setup. 

---
### Comments to reviewers
I'm not 100% sure we need to duplicate this material and assign it only for transparent Unity Ball or simple set Double Sided in the current "Aluminum Satin" material.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Samples~/MaterialSamples/Hair.unity
+++ b/com.unity.render-pipelines.high-definition/Samples~/MaterialSamples/Hair.unity
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 0}
-  m_IndirectSpecularColor: {r: 299.8381, g: 336.92505, b: 537.1908, a: 1}
+  m_IndirectSpecularColor: {r: 299.87256, g: 336.97363, b: 537.2549, a: 1}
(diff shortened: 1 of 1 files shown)


JMargevics (approved):
Change makes perfect sense. Not much to check - auto approving 👍

sebastienlagarde (approved):

SeanPuller:
I think the duplication of the material is better. If someone wants to just grab the aluminum and use it (probably not going to happen very often) they don't have to uncheck double sided and it specifically highlights this use case by having a specific double sided material used with transparent objects.

SeanPuller (approved):

PR #3972, opened 2021-03-23
Re-enabling the basic spot light test for the built-in target

joshua-davis:
---
### Purpose of this PR
Adding a simple test that shows the expected lighting of built-in with spot lights. Spot lights don't match URP at all though, seemingly due to the falloff of spot lights in built-in.

---
### Testing status

From a local run:
| URP        | BuiltIn-ShaderGraph           | BuiltIn-SurfaceShader |
| ------------- | ------------- | ----- |
| 022_Lighting_BasicSpot expected      | 022_Lighting_BasicSpot_ShaderGraph | 022_Lighting_BasicSpot_SurfaceShader |

The shader graph built-in seems to match surface shaders within a reasonable tolerance so I believe it's fine to accept the results of this run as the new reference images. The main difference between surface shaders and shader graph is at the border of the shadow.

---
### Comments to reviewers
As a small note, URP had two different shaders. Left was blinn-phong and right was metallic PBR. Shader graph is not planning to support blinn-phong so the left was replaced with the same metallic shader.


jessebarker (approved):
LGTM

LandonTownsendUnity (approved):

PR #3976, opened 2021-03-23
Fix for tests failures caused by missing files in the test framework

joshua-davis:
Tests started failing after a recent change. This is because some test infrastructure files were accidentally removed, in particular stuff that setup tests (lighting, etc...).

For the other files re-enabled:
CopyImageToReferenceFolders: Adds a right-click menu to allow copying assets to the reference location. Seemed like a good utility to leave in.
BuiltInProjectEditorTests: Adds an editor check to make sure the render pipeline is built-in.


jessebarker (approved):
LGTM.  Too bad some of this can't be shared across projects as it's all the same boilerplate.

PR #3997, opened 2021-03-25
Urp/depth priming

NicoLeyman:

### Checklist for PR maker

---
### Purpose of this PR

This PR adds support for depth priming to the forward render mode.
Depth priming involves re-using the depth buffer generated during a prepass (originally for use in post processing) at a later stage when rendering opaques. The opaque pass then uses no z-writing and checks for equal depth before a fragment is shaded. This should reduce overdraw.

Potential issues lie in the fact that the depth prepass must generate MSAA depth that is kept around for use in the opaque pass and resolved for use during post processing.

Because of this we've already disabled the feature on hardware with hidden surface removal. (TBDR hardware does not like it when you store/load an MSAA target between passes)

Here is the new Depth Priming Mode setting added by the PR.
Tooltip
InfoBox

---
### Testing status
Some manual functional testing on DX11/12/Vulkan. +  URP GraphicsTestProject.

Functional QA pass uncovered several issues, most of which have been fixed.
One thing we couldn't fix cleanly was Depth Priming + Text Mesh Rendering. This should be fixed as part of a push to overhaul SRP UI/Text drawing. Text is drawn as part of the opaque pass regardless of transparency and overrides the depth state for just the text drawcalls while Depth Priming overrides the depth state for the entire renderpass.
As long as this is an issue it's recommended to leave Depth Priming disabled by default.

Performance Pass (mobile & desktop) pending.

---
### Comments to reviewers

This PR is not expected to be fully finished atm. I'm looking for some feedback regarding the viability of the changes in context of Mobile platforms, URP and SRP as a whole (in case of the changes to ScriptableRenderer).
Once the code has been sanity checked we can look at refining it and performing performance tests.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
@@ -605,7 +605,7 @@ namespace UnityEngine.Rendering.Universal
             }
 
             // Depth priming requires a manual resolve of MSAA depth right after the depth prepass. If autoresolve is supported but MSAA is 1x then a copy is still required.
-            if (useDepthPriming && (!SystemInfo.supportsMultisampleAutoResolve || cameraTargetDescriptor.msaaSamples == 1))
+            if (useDepthPriming && (SystemInfo.graphicsDeviceType != GraphicsDeviceType.Vulkan || cameraTargetDescriptor.msaaSamples == 1))
             {
                 m_PrimedDepthCopyPass.Setup(m_ActiveCameraDepthAttachment, m_DepthTexture);
                 m_PrimedDepthCopyPass.AllocateRT = false;


com.unity.render-pipelines.universal/Shaders/Particles/ParticlesUnlitForwardPass.hlsl
```diff
@@ -111,7 +111,7 @@ half4 fragParticleUnlit(VaryingsParticle input) : SV_Target
 
     half3 result = albedo.rgb + emission;
 
-    #if defined(_SCREEN_SPACE_OCCLUSION)
+    #if defined(_SCREEN_SPACE_OCCLUSION) && !defined(_SURFACE_TYPE_TRANSPARENT)
```
ellioman, line 114: Merging latest master will remove this change :)

com.unity.render-pipelines.universal/Shaders/Unlit.shader
```diff
@@ -121,7 +121,7 @@ Shader "Universal Render Pipeline/Unlit"
                 fogFactor = input.fogCoord;
             #endif
 
-                #if defined(_SCREEN_SPACE_OCCLUSION)
+                #if defined(_SCREEN_SPACE_OCCLUSION) && !defined(_SURFACE_TYPE_TRANSPARENT)
```
ellioman, line 124: Same. Merging latest master will remove this change.

com.unity.render-pipelines.universal/Shaders/Unlit.shader
```diff
@@ -327,7 +327,7 @@ Shader "Universal Render Pipeline/Unlit"
                     color *= alpha;
                 #endif
 
-                #if defined(_SCREEN_SPACE_OCCLUSION)
+                #if defined(_SCREEN_SPACE_OCCLUSION) && !defined(_SURFACE_TYPE_TRANSPARENT)
```
ellioman, line 330: And again, merging latest master will remove this change.

com.unity.render-pipelines.universal/Runtime/Passes/DepthNormalOnlyPass.cs
```diff
@@ -14,9 +14,12 @@ public class DepthNormalOnlyPass : ScriptableRenderPass
         internal ShaderTagId shaderTagId { get; set; } = k_ShaderTagId;
 
         private RenderTargetHandle depthHandle { get; set; }
+        private RenderTargetHandle cameraDepthHandle { get; set; }
         private RenderTargetHandle normalHandle { get; set; }
         private FilteringSettings m_FilteringSettings;
 
+        private bool useDepthPriming;
```
ellioman, line 21: -> m_UseDepthPriming

com.unity.render-pipelines.universal/Runtime/Passes/DepthNormalOnlyPass.cs
```diff
@@ -33,7 +36,7 @@ public DepthNormalOnlyPass(RenderPassEvent evt, RenderQueueRange renderQueueRang
         /// <summary>
         /// Configure the pass
         /// </summary>
-        public void Setup(RenderTextureDescriptor baseDescriptor, RenderTargetHandle depthHandle, RenderTargetHandle normalHandle)
+        public void Setup(RenderTextureDescriptor baseDescriptor, RenderTargetHandle depthHandle, RenderTargetHandle normalHandle, RenderTargetHandle cameraDepthHandle, bool useDepthPriming)
```
ellioman, line 39: We need to make sure we do not break users. This is a public API change so could we default the new parameters to `null` and `false`?

com.unity.render-pipelines.universal/Runtime/Passes/DepthNormalOnlyPass.cs
ellioman: Or change it to a private property

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -38,7 +39,8 @@ public DepthOnlyPass(RenderPassEvent evt, RenderQueueRange renderQueueRange, Lay
         /// </summary>
         public void Setup(
             RenderTextureDescriptor baseDescriptor,
-            RenderTargetHandle depthAttachmentHandle)
+            RenderTargetHandle depthAttachmentHandle,
+            bool useDepthPriming)
```
ellioman, line 43: Needs a fallback here, default it to `false`?

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -50,15 +52,26 @@ public DepthOnlyPass(RenderPassEvent evt, RenderQueueRange renderQueueRange, Lay
...
         }
 
         public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderingData)
         {
             if (this.allocateDepth)
                 cmd.GetTemporaryRT(depthAttachmentHandle.id, descriptor, FilterMode.Point);
             var desc = renderingData.cameraData.cameraTargetDescriptor;
-            ConfigureTarget(new RenderTargetIdentifier(depthAttachmentHandle.Identifier(), 0, CubemapFace.Unknown, -1), GraphicsFormat.DepthAuto, desc.width, desc.height, 1, true);
-            ConfigureClear(ClearFlag.All, Color.black);
+
```
ellioman, line 63: Can you add a short comment about the logic here?

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -50,15 +52,26 @@ public DepthOnlyPass(RenderPassEvent evt, RenderQueueRange renderQueueRange, Lay
...
+            if (this.useDepthPriming)
+            {
+                overrideCameraTarget = false;
+            }
+            else
+            {
+                ConfigureTarget(new RenderTargetIdentifier(depthAttachmentHandle.Identifier(), 0, CubemapFace.Unknown, -1), GraphicsFormat.DepthAuto, desc.width, desc.height, 1, true);
+            }
```
ellioman, line 71: Can you add a short comment about the logic here?

com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs
```diff
@@ -54,6 +57,29 @@ internal DrawObjectsPass(URPProfileId profileId, bool opaque, RenderPassEvent ev
             m_ProfilingSampler = ProfilingSampler.Get(profileId);
         }
 
+        public void SetUseDepthPriming(bool useDepthPriming)
```
ellioman, line 60: Add some documentation about this new function

com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs
```diff
@@ -54,6 +57,29 @@ internal DrawObjectsPass(URPProfileId profileId, bool opaque, RenderPassEvent ev
...
+        {
+            m_UseDepthPriming = useDepthPriming;
+        }
+
+        public override void Configure(CommandBuffer cmd, RenderTextureDescriptor cameraTextureDescriptor)
```
ellioman, line 65: Does this make more sense in the `OnCameraSetup()` callback?

com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs
```diff
@@ -54,6 +57,29 @@ internal DrawObjectsPass(URPProfileId profileId, bool opaque, RenderPassEvent ev
...
+        {
+            if (m_UseDepthPriming)
+            {
+                m_RenderStateBlock.depthState = new DepthState(false, CompareFunction.Equal);
+                m_RenderStateBlock.mask |= RenderStateMask.Depth;
+                m_PreviouslyUsedDepthPriming = true;
+            }
+            // If the previous execution of the pass was using depth priming, only then reset the depth state to default.
+            // Always resetting to default would break the custom state set by certain DrawObjectPass instances such as particle rendering.
+            else if (m_PreviouslyUsedDepthPriming)
```
pbbastian, line 75: Hmm, it feels like this is hiding an underlying issue. What's the actual problem going on here? I'd like to avoid this statefulness.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -374,7 +410,13 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             }
 #endif
 
-            if (usesRenderPass)
+#if !UNITY_ANDROID && !UNITY_IOS
```
pbbastian, line 413: What about desktop TBDR such as Apple Silicon and Windows ARM devices?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -374,7 +410,13 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+            bool useDepthPriming = m_DepthPrimingMode == DepthPrimingMode.Auto && requiresDepthPrepass && m_RenderingMode == RenderingMode.Forward && (cameraData.renderType == CameraRenderType.Base || cameraData.clearDepth) /*&& SystemInfo.graphicsDeviceType != GraphicsDeviceType.Direct3D11 && SystemInfo.graphicsDeviceType != GraphicsDeviceType.Direct3D12*/;
```
pbbastian, line 414: Remove the commented out bit

pbbastian (changes requested):
This is generally looking really solid! I have some requests :)

ellioman (changes requested):
Looking good. Needs some minor changes before approving.
I also added QA and @oleks-k for reviewing.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: I hadn't considered TBDR hardware on non-Android and non-iOS devices. I'm expanding this check and moving it to the constructor. Not sure if I have the vender ID's right however.

com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs
NicoLeyman: Luckily it looks like this is no longer relevant. Adding a DepthOnly pass to the custom shader of the particles test scene seems to resolve this. Wasn't very fond of this state management at all so glad it's gone.

com.unity.render-pipelines.universal/Runtime/Data/ForwardRendererData.asset
```diff
@@ -39,4 +40,5 @@ MonoBehaviour:
     zFailOperation: 0
   m_ShadowTransparentReceive: 1
   m_RenderingMode: 0
+  m_DepthPrepassMode: 0
```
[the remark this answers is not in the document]
NicoLeyman: I'd expect that as well. For some reason it must not be opening that asset anymore. I'll commit the change if it ever pops up in my working directory.

com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs
pbbastian: Ahh okay, great to hear! I did have a hard time figuring out when it actually hit the second state :p

NicoLeyman:
Pretty much all PR feedback has been addressed. Did a pass over the Graphics Tests and fixed some issues uncovered as a result.
TODO:
- More docs
- Perf testing (w/ depth priming checks updated correctly per platform)
- Automated test coverage.

com.unity.render-pipelines.universal/Editor/UniversalRendererDataEditor.cs
```diff
@@ -18,6 +18,8 @@ private static class Styles
             public static readonly GUIContent TransparentMask = new GUIContent("Transparent Layer Mask", "Controls which transparent layers this renderer draws.");
             public static readonly GUIContent LightingLabel = new GUIContent("Lighting", "Settings related to lighting and rendering paths.");
             public static readonly GUIContent RenderingModeLabel = new GUIContent("Rendering Path", "Select a rendering path.");
+            public static readonly GUIContent DepthPrepassModeLabel = new GUIContent("Depth Prepass Mode", "How this renderer should decide to use a depth prepass. Auto means a depth prepass will be used if needed by a later pass. Forced will always use a depth prepass.");
```
oleks-k, line 21: Does Unity execute the depth prepass regardless of this setting?
What does "use" mean here? Does the Renderer include something somewhere when it "uses" a depth prepass?
We can have a quick call if the question doesn't make much sense :)

pbbastian (approved):

ellioman (approved):

com.unity.render-pipelines.universal/Editor/UniversalRendererDataEditor.cs
NicoLeyman: I guess "use" should be "execute" here. So it determines whether it's executed as needed or always.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -15,6 +16,28 @@ public enum RenderingMode
...
+    /// <summary>
+    /// When the Universal Renderer should use a depth prepass in Forward mode.
+    /// </summary>
+    public enum DepthPrimingMode
+    {
+        /// <summary>A depth prepass will only be used if the rendering mode is set to forward and there is a depth prepass used.</summary>
+        Auto,
+        /// <summary>A depth prepass will never be used.</summary>
+        Disabled
+    }
```
phi-lira, line 39: What's use case that we have depth prepass and want priming mode disabled? Is there any downside to it?

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalTarget.cs
```diff
@@ -369,6 +369,36 @@ static class CorePasses
             customInterpolators = CoreCustomInterpDescriptors.Common
         };
 
+        public static readonly PassDescriptor DepthNormalOnly = new PassDescriptor()
```
phi-lira, line 372: Is there any changes here  other than moving the code around?

com.unity.render-pipelines.universal/Runtime/Passes/DepthNormalOnlyPass.cs
```diff
@@ -14,9 +14,12 @@ public class DepthNormalOnlyPass : ScriptableRenderPass
         internal ShaderTagId shaderTagId { get; set; } = k_ShaderTagId;
 
         private RenderTargetHandle depthHandle { get; set; }
+        private RenderTargetHandle cameraDepthHandle { get; set; }
```
phi-lira, line 17: camera depth can be accessed from within pass by calling renderingData.cameraData.renderer.cameraDepthTarget.

com.unity.render-pipelines.universal/Runtime/Passes/DepthNormalOnlyPass.cs
```diff
@@ -14,9 +14,12 @@ public class DepthNormalOnlyPass : ScriptableRenderPass
...
         private RenderTargetHandle normalHandle { get; set; }
         private FilteringSettings m_FilteringSettings;
 
+        private bool m_UseDepthPriming;
```
phi-lira, line 21: This seems to be a pattern that has to be copied to several passes. The passes have access to the ScriptableRenderer, I wonder if we should expose a property on the renderer that tells if a depth prepass/priming is used.

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -50,15 +55,28 @@ public DepthOnlyPass(RenderPassEvent evt, RenderQueueRange renderQueueRange, Lay
...
+                ConfigureTarget(new RenderTargetIdentifier(depthAttachmentHandleMSAA.Identifier(), 0, CubemapFace.Unknown, -1), GraphicsFormat.DepthAuto, desc.width, desc.height, 1, true);
+            }
+            // When not using depth priming the camera target should be set to our non MSAA depth target.
+            else
+            {
+                ConfigureTarget(new RenderTargetIdentifier(depthAttachmentHandle.Identifier(), 0, CubemapFace.Unknown, -1), GraphicsFormat.DepthAuto, desc.width, desc.height, 1, true);
+            }
+
+            // Only clear depth here so we don't clear any bound color target. It might be unused by this pass but that doesn't mean we can just clear it. (e.g. in case of overlay cameras + depth priming)
+            ConfigureClear(ClearFlag.Depth, Color.black);
```
phi-lira, line 79: @manuele-bonanno since you are working on MSSA and depth "prepass" we have today.

com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs
```diff
@@ -54,6 +56,30 @@ internal DrawObjectsPass(URPProfileId profileId, bool opaque, RenderPassEvent ev
...
+        /// If this is true then the depth state will be overridden to only allow pixels with a depth value matching the one in the depth generated by the depth prepass while Depth writing will be turned off.
+        /// </summary>
+        public void SetUseDepthPriming(bool useDepthPriming)
+        {
+            m_UseDepthPriming = useDepthPriming;
+        }
+
+        public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderingData)
+        {
+            if (m_UseDepthPriming)
```
phi-lira, line 71: if you would expose property in renderer to access depth priming here you could do `renderingData.cameraData.renderer.useDepthPriming`.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -1078,11 +1091,12 @@ void SetRenderPassAttachments(CommandBuffer cmd, ScriptableRenderPass renderPass
 
                 // When render pass doesn't call ConfigureTarget we assume it's expected to render to camera target
                 // which might be backbuffer or the framebuffer render textures.
+
                 if (!renderPass.overrideCameraTarget)
                 {
                     // Default render pass attachment for passes before main rendering is current active
                     // early return so we don't change current render target setup.
-                    if (renderPass.renderPassEvent < RenderPassEvent.BeforeRenderingOpaques)
+                    if (renderPass.renderPassEvent < RenderPassEvent.BeforeRenderingPrePasses)
```
phi-lira, line 1099: We need to check if this causes a behaviour change regarding clearing. I think @Verasl mentioned this PR caused a different in clear in boat attack. It could be caused due to this.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -37,8 +60,11 @@ private static class Profiling
         internal RenderingMode actualRenderingMode { get { return GL.wireframe || m_DeferredLights == null || !m_DeferredLights.IsRuntimeSupportedThisFrame() || m_DeferredLights.IsOverlay ? RenderingMode.Forward : this.renderingMode; } }
         internal bool accurateGbufferNormals { get { return m_DeferredLights != null ? m_DeferredLights.AccurateGbufferNormals : false; } }
         internal bool usesRenderPass;
+        internal DepthPrepassMode depthPrepassMode { get { return m_DepthPrepassMode; } }
+        internal DepthPrimingMode depthPrimingMode { get { return m_DepthPrimingMode; } }
```
phi-lira, line 64: Do we think users would want to change this at runtime? Maybe if so we could expose it.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -123,8 +152,29 @@ public UniversalRenderer(UniversalRendererData data) : base(data)
...
+#elif UNITY_EDITOR_OSX || UNITY_STANDALONE_OSX
+            if (CultureInfo.InvariantCulture.CompareInfo.IndexOf(SystemInfo.graphicsDeviceVendor, "SILICON", CompareOptions.IgnoreCase) >= 0)
+            {
+                this.m_CanUseDepthPriming = false;
+            }
+#endif
+
+#else
+            this.m_CanUseDepthPriming = false;
+#endif
```
phi-lira, line 176: I particularly don't like implicit logic like this disabling things under the hood for users. What's the motivation to disable depth priming if we already have a depth prepass?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: I don't think there is a valid use case for it atm. If there is a need for a depth pass it should be derived from the render passes. If there's a reason to have a way to force a depth prepass @Verasl might know. Could be that we both forgot about the prepass being auto injected when we went over this previously.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
phi-lira: For platforms that don't support https://docs.unity3d.com/ScriptReference/SystemInfo-hasHiddenSurfaceRemovalOnGPU.html it would be improvement to not sort opaques from front-to-back and do more aggresive batching.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -123,8 +152,29 @@ public UniversalRenderer(UniversalRendererData data) : base(data)
...
             this.usesRenderPass = data.useNativeRenderPass;
 
+#if !UNITY_ANDROID && !UNITY_IOS
+            this.m_CanUseDepthPriming = true;
+
+#if UNITY_EDITOR_WIN || UNITY_STANDALONE_WIN
+            if (CultureInfo.InvariantCulture.CompareInfo.IndexOf(SystemInfo.graphicsDeviceVendor, "ARM", CompareOptions.IgnoreCase) >= 0)
+            {
+                this.m_CanUseDepthPriming = false;
+            }
```
phi-lira, line 166: TODO: check if https://docs.unity3d.com/ScriptReference/SystemInfo-hasHiddenSurfaceRemovalOnGPU.html is enabled on Mali. If it's not at least on CPU my understanding is that we could improve it.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -123,8 +152,29 @@ public UniversalRenderer(UniversalRendererData data) : base(data)
...
+#elif UNITY_EDITOR_OSX || UNITY_STANDALONE_OSX
+            if (CultureInfo.InvariantCulture.CompareInfo.IndexOf(SystemInfo.graphicsDeviceVendor, "SILICON", CompareOptions.IgnoreCase) >= 0)
+            {
+                this.m_CanUseDepthPriming = false;
+            }
```
phi-lira, line 171: Why is it disabled in OSX?

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
```diff
@@ -178,7 +178,10 @@ internal bool Setup(ScreenSpaceAmbientOcclusionSettings featureSettings, Scripta
                 }
                 else
                 {
-                    renderPassEvent = featureSettings.AfterOpaque ? RenderPassEvent.AfterRenderingOpaques : RenderPassEvent.AfterRenderingPrePasses;
+                    // Rendering after PrePasses is usually correct except when depth priming is in play:
+                    // then we rely on a depth resolve taking place after the PrePasses in order to have it ready for SSAO.
+                    // Hence we set the event to BeforeRenderingOpaques at the earliest.
+                    renderPassEvent = featureSettings.AfterOpaque ? RenderPassEvent.AfterRenderingOpaques : RenderPassEvent.BeforeRenderingOpaques;
```
phi-lira, line 184: I'm worried we might introduce a regression if there's a custom pass that injects also in BeforeRenderingOpaques but run before SSAO. 
 
 Safer approach can be to do do `RenderPassEvent.AfterRenderingPrePasses + 1` (this is just a prioriity queue).

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -463,12 +528,21 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
             }
 
+            // Depth priming requires a manual resolve of MSAA depth right after the depth prepass.
+            if (useDepthPriming)
+            {
+                m_PrimedDepthCopyPass.Setup(m_ActiveCameraDepthAttachment, m_DepthTexture);
+                m_PrimedDepthCopyPass.AllocateRT = false;
+
+                EnqueuePass(m_PrimedDepthCopyPass);
+            }
```
phi-lira, line 544: @manuele-bonanno for awareness as it seems this could be optimized in a follow up PR with your work?

phi-lira (changes requested):

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalTarget.cs
NicoLeyman: No, the code was moved around from Lit to Core so we could have Unlit ShaderGraph shaders w/ a depthnormals pass.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
manuele-bonanno: yes this could be removed on platforms with depth resolve support (once my PR lands)

manuele-bonanno: Something that might be an issue here.
If you look at CanCopyDepth(), there is this TODO comment:
// TODO:  We don't have support to highp Texture2DMS currently and this breaks depth precision.
 
 because of this, using MSAA'd textures as source of the copy is currently disabled by default.
 So if that is still the case (unless that comment is outdated), aren't we breaking potentially depth precision here by copying a multisampled surface?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -852,10 +899,8 @@ bool CanCopyDepth(ref CameraData cameraData)
             bool supportsDepthTarget = RenderingUtils.SupportsRenderTextureFormat(RenderTextureFormat.Depth);
             bool supportsDepthCopy = !msaaEnabledForCamera && (supportsDepthTarget || supportsTextureCopy);
 
-            // TODO:  We don't have support to highp Texture2DMS currently and this breaks depth precision.
-            // currently disabling it until shader changes kick in.
-            //bool msaaDepthResolve = msaaEnabledForCamera && SystemInfo.supportsMultisampledTextures != 0;
-            bool msaaDepthResolve = false;
+            bool msaaDepthResolve = msaaEnabledForCamera && SystemInfo.supportsMultisampledTextures != 0;
+
```
manuele-bonanno, line 903: this will now introduce an MSAA depth copy on mobile/TBDR platforms when MSAA > 1 (instead of the current non-AA depth prepass) which is quite a big hit performance wise. Looks like a regression

to reproduce the issue:
enable MSAA, tick "Depth Texture". The depth prepass should now be replaced by a MSAA'd copy

manuele-bonanno (changes requested):

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
manuele-bonanno: @pbbastian confirmed that we now have highp support for Texture2DMS so all good :)

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: It wouldn't be super useful but there's no conceptual problems with it. Exposing it only required a conditional copy pass in the constructor to become independent of the depth priming mode at construction time.

oleks-k (approved):
Looks good!

com.unity.render-pipelines.universal/Runtime/Passes/DepthNormalOnlyPass.cs
NicoLeyman: Fixed

com.unity.render-pipelines.universal/Runtime/Passes/DepthNormalOnlyPass.cs
NicoLeyman: Fixed

com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs
NicoLeyman: Fixed

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceAmbientOcclusion.cs
NicoLeyman: Fixed

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: This was a crude and ugly way to check if a system might have TBDR hardware.

I've replaced this whole section with `m_CanUseDepthPriming = !SystemInfo.hasHiddenSurfaceRemovalOnGPU;`
This still leaves us with Depth Priming on Tiled GPU's without HSR. To detect those we should expose an internal property.
The assumption is that for Tiled GPUs:

- If HSR is also supported then Depth Priming will have no benefits.
- If some form of automated sorting is present we can potentially still gain a small benefit from Depth Priming.
- If MSAA is in play then we would have to perform prohibitively expensive MSAA depth load/stores unless we can merge the depth and opaque passes together.

For now I would proceed as follows:
- Disable Depth Priming on TBDR hardware. (current state of the PR)
- Expose the required info in Unity to allow us to detect Tile GPUs and then disable it for those as well. (so in a later PR)
- Add in a special case for Tiled Hardware without HSR: allow if no MSAA or if the Depth and Opaque passes can be merged. (another later PR)

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: I have added an additional condition that hasHiddenSurfaceRemovalOnGPU must be false however that will still let non HSR tiled GPU's slip through the cracks. This is another reason to expose hasTiledGPU to SystemInfo.
I think I'll just set this back to false until we can guarantee there's no perf regression for tiled hardware.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
phi-lira: Ok. We can expose if they ask. Resolving conversation.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
phi-lira: Makes sense to me.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -461,6 +461,8 @@ static class RenderPassBlock
             m_ActiveDepthAttachment = depthAttachment;
         }
 
+        public bool useDepthPriming { get; set; } = false;
```
phi-lira, line 464: Needs public API docs.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -37,8 +51,10 @@ private static class Profiling
         internal RenderingMode actualRenderingMode { get { return GL.wireframe || m_DeferredLights == null || !m_DeferredLights.IsRuntimeSupportedThisFrame() || m_DeferredLights.IsOverlay ? RenderingMode.Forward : this.renderingMode; } }
         internal bool accurateGbufferNormals { get { return m_DeferredLights != null ? m_DeferredLights.AccurateGbufferNormals : false; } }
         internal bool usesRenderPass;
+        public DepthPrimingMode depthPrimingMode { get { return m_DepthPrimingMode; } set { m_DepthPrimingMode = value; } }
```
phi-lira, line 54: Needs public API docs.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -1,6 +1,7 @@
 using UnityEngine.Rendering.Universal.Internal;
 using UnityEngine.Experimental.Rendering;
 using System.Reflection;
+using System.Globalization;
```
phi-lira, line 4: I guess this is not needed anymore?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -122,8 +140,11 @@ public UniversalRenderer(UniversalRendererData data) : base(data)
             m_ForwardLights = new ForwardLights();
             //m_DeferredLights.LightCulling = data.lightCulling;
             this.m_RenderingMode = data.renderingMode;
+            this.m_DepthPrimingMode = data.depthPrimingMode;
             this.usesRenderPass = data.useNativeRenderPass;
 
+            this.m_CanUseDepthPriming = !SystemInfo.hasHiddenSurfaceRemovalOnGPU;
```
phi-lira, line 146: We need to check if ARM has this flag and possibly disable for ARM. @manuele-bonanno ?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -442,7 +477,7 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
                     }
                     else
                     {
-                        m_DepthNormalPrepass.Setup(cameraTargetDescriptor, m_DepthTexture, m_NormalsTexture);
+                        m_DepthNormalPrepass.Setup(cameraTargetDescriptor, m_DepthTexture, m_NormalsTexture, useDepthPriming);
```
phi-lira, line 480: No need to pass this as argument as the pass already has access to the renderer.

phi-lira (approved):
LGTM. Only thing I'd like to discuss for this PR is the HSR and ARM.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
manuele-bonanno: Sounds good thanks!
In general I think it is good to enable the copy and remove depth prepass when MSAA > 1 on mobile platforms.

But since the change is not directly related to the depth priming, I agree it's safer to do it in a separate PR so all platforms can be tested for potential regressions. I am planning to look into that once some engine PRs (in progress) with new depth resolve API will land

manuele-bonanno (approved):
looks good!

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
manuele-bonanno: yes storing MSAA sounds very expensive on ARM because of the lack of compression, according to them. 
Not sure about that flag for ARM GPUs. @hkr?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: Indeed. Initially kept this in because I setup the normal descriptors in the setup based on depth priming being enabled or not but it's possible to alter the description as needed during OnCameraSetup(). Fixing.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: Indeed. removing

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
NicoLeyman: Should have been internal. Fixing

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: Added through a <summary> tag. Is this sufficient ?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: hasHiddenSurfaceRemovalOnGPU is true for:
- Qualcomm Adreno 5 and up
- PowerVR
- Apple Silicon

whereas the internal hasTiledGPU is true for:
- Qualcomm Adreno
- PowerVR
- ARM Mali
- Vivante
- Apple Silicon

So it appears we're missing a few platforms by simply just checking for hasHiddenSurfaceRemovalOnGPU .

hkr: Mali's HSR (FPK) is not as efficient as e.g. PowerVR.
Main use of `hasHiddenSurfaceRemovalOnGPU` so far was to decide if we should do front-to-back sort of opaque objects or not.
For Mali it's usually worth sorting, also because sorting is relatively cheap for many games.
For most games depth priming is probably significantly more expensive than sorting. So the trade-off will be different.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -334,6 +361,8 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             if (requiresDepthPrepass && this.actualRenderingMode == RenderingMode.Deferred && !renderPassInputs.requiresNormalsTexture)
                 requiresDepthPrepass = false;
 
+            requiresDepthPrepass |= m_DepthPrimingMode == DepthPrimingMode.Forced && m_CanUseDepthPriming;
```
hkr, line 364: Maybe we should allow depth priming for "Forced" even on mobile GPUs with HSR.
There are probably some games where depth priming may still be beneficial in some situations.

TheoWong-pixel (approved):

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
NicoLeyman: There are some situations where we could avoid the perf cost on Tiled Hardware. 
- When MSAA is not in use.
- When we can merge the depth prepass and opaque pass into one to avoid having to store/load MSAA data between passes.

If a GPU actually has HSR I'm not sure where we could gain perf. Wouldn't we be doing the same thing that already happens in the hardware ?  If we have  depth priming on (so depth write off and compare to equals) could that further speed up the HSR rendering ?
ARM has tiled hardware w/ automatic sorting so there we could gain the benefits of depth priming still IF we navigate the situations that we can actually use it.

Right now the plan is to add depth priming support for these situations in a later PR.

hdb-unity (approved):
QA testing complete.

All PR-related issues resolved. Some non-PR issues still to report+discuss.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
hkr: At least on Mali and Adreno the HSR implementations cannot avoid all redundant fragment shader executions.
For example Adreno only tests against a low resolution depth buffer. There are also slow paths where it cannot do HSR at all. It's kind of a black box.
Based on heuristics the Adreno driver it may also switch to full immediate rendering without any binning or HSR.

Even PowerVR and similar will have to disable HSR if the fragment shaders writes to depth, for alpha-test or if there are other side-effects such as UAV writes.

thomas-zeng (approved):
LGTM. 
To XR QA, Please test this feature on Quest with shared depth enabled.

aleksandrasdzikia (approved):
The PR looks good. tested different scenarios on mobile platform: MSAA_ON and SSAO_OFF, MSAA_OFF and SSAO_ON, Both ON and OFF. So, no regressions without priming and no issues found with priming. A known issue is with UI rendering, but no other issues found related to this PR. Used URP Examples test project. Devices used: Huawei P20 (Mali-G72), Galaxy Note 10 (Adreno (TM) 640), Meizu Pro 6 Plus (Mali-T880), One Plus Nord (Adreno 620), iPad Pro 12.9 (A12x) and iPhone SE 1st gen (A9).

ryanhy-unity (approved):
Took a look at this with several XR configs and saw no regressions

Quest with Vulkan and OpenGLES3
Rift with DX11
Mock HMD in editor play mode DX11

Tested with the XR URP end to end test project

hkr (approved):

Jonasmortensen (commented):
Looks mostly good. Only found two issues that have been discussed with author

PR #3999, opened 2021-03-25
Reduce clear calls in TextureCache ClearCache functions

christine-johnson:
---
### Purpose of this PR
Reduces clear calls in TextureCache ClearCache functions by clearing texture arrays in a single call rather than clearing each slice separately.


FrancescoC-unity (approved):
Discussed with @christine-johnson and the tests she did sounded good to me.  (Checked via renderdoc that clears do happen) 

Let's see  how yamato goes.

PR #4002, opened 2021-03-25
[2021.2][URP] Update URP Lit SubTarget to generate emission color property

jessebarker:
### Checklist for PR maker

---
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1319637/
It should also fix https://fogbugz.unity3d.com/f/cases/1324684/

The Universal Lit Shader Graph SubTarget was not adding the _EmissionColor property to its generated shaders, which caused the registered material editor to generate errors when a material was created with the Lit Shader Graph.  This PR simply updates the Lit SubTarget to add that property.

It might be good in the future if the PBRMasterGUI lived with the Universal Target code, rather than in Shader Graph, so these pieces could be more tightly coupled.

---
### Testing status
Manual testing using the repro steps in the bug.
Yamato results to come

---
### Comments to reviewers
I would like to confirm that the default value for this property is correct (easily changed)

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
+++ b/com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
@@ -141,7 +141,6 @@ namespace UnityEditor.Rendering.Universal.ShaderGraph
                 colorMode = ColorMode.HDR,
                 value = new Color(1.0f, 1.0f, 1.0f, 1.0f)
             });
-
         }
 
         public override void GetPropertiesGUI(ref TargetPropertyGUIContext context, Action onChange, Action<String> registerUndo)


com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalLitSubTarget.cs
```diff
@@ -128,6 +129,20 @@ public override void GetActiveBlocks(ref TargetActiveBlockContext context)
...
+        {
+            base.CollectShaderProperties(collector, generationMode);
+            collector.AddShaderProperty(new ColorShaderProperty()
+            {
+                overrideReferenceName = "_EmissionColor",
+                hidden = true,
+                overrideHLSLDeclaration = true,
+                hlslDeclarationOverride = HLSLDeclaration.UnityPerMaterial,
+                value = new Color(1.0f, 1.0f, 1.0f, 1.0f)
+            });
```
cdxntchou, line 142: What does the _EmissionColor property do?

cdxntchou: Aside from getting rid of the warning spew?  :)

jessebarker: No idea.  It's not really in Shader Graph.  It's really a pipeline thing (even though the material editor GUI is in SG).

jessebarker: So, there is a custom ShaderGUI registered by the URP Lit SubTarget which "asserts" this property on the material editor.  If it's not there, we get the error.  It's always declared by the hand-written Lit.shader.  This adds it in the same fashion. The difference is the default value, which is the one thing I'm still not sure about (Lit.shader has it as a 3-component 0).

cdxntchou: Right, but is it the correct thing to add this property, or remove the warning?   Need to find someone who can answer that question..  :)

cdxntchou: ah, nevermind, I read the bug -- Kasper says to do it, he should know..   :)

cdxntchou: he should know the correct default as well.

cdxntchou: @KEngelstoft mind enlightening us as to what this value does?   :)

jessebarker: Yeah, I'd put him on this PR as reviewer for precisely that reason :-)

cdxntchou: ok, so it's kind of equivalent to MainTex/MainColor -- a fixed function definition of a GI input.
MainTex/MainColor define the bounce color for GI,
and _EmissionMap/_EmissionColor define the emission color for GI.

KEngelstoft: In the built in Standard shader, there is a checkbox for Emission. Once checked, the emission HDR color property is shown and the default color is black.
There is an enum dropdown letting the user choose if emission should control realtime, baked or no GI.
image

KEngelstoft (commented):
Please change default color to black.

PR #4007, opened 2021-03-25
Created some shader graph built-in target tests for vertex blocks

joshua-davis:
---
### Purpose of this PR
Three tests were add to try and show that each vertex block field is being properly used for the built-in target. As much as possible, the goal was to show that the target uses these fields correctly in the vertex stage (not relying on fragment stage interpretation of the data).

---
### Comments to reviewers
For the position block test, the position is offset by the normal and then rendered. the lighting looks reasonably accurate and the shadows correctly reflect the positional offsets.
063_Lighting_BasicVertexPosition

For the normal block test, the shader negates the normal in the vertex block and then negates it again in the pixel. This means that for a normal pixel lit sphere everything should look roughly normal (left sphere) but for a vertex lit sphere the lighting should be inverted (right sphere, taking attenuation into account).
064_Lighting_BasicVertexNormal

I couldn't come up with a good test for tangent that didn't involve sampling in the pixel shader. The attached test negates the tangent vector in the vertex block and uses that to sample in the pixel stage. The left sphere is the negated tangent and the right is normal.
065_Lighting_BasicVertexTangent

No comparison images were made with respect to surface shaders as I didn't think they were necessary due to how simple the lighting is. Hopefully other tests cover the usages of these fields in the pixel blocks. Also some things are harder to deal with in surface shaders due to differing spaces (e.g. normal/tangent).

### DIFF
--- a/TestProjects/BuiltInGraphicsTest_Foundation/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/063_Lighting_BasicVertexPosition.png
+++ b/TestProjects/BuiltInGraphicsTest_Foundation/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/063_Lighting_BasicVertexPosition.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:2004db12ce39bda801505d8f05e64272201101c459136ccbc181e973dce13554
+size 20845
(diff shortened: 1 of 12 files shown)


jessebarker (approved):
LGTM

LandonTownsendUnity (approved):
These tests look good!

PR #4008, opened 2021-03-26
Bump testframework.build to preview.14 for all test projects

theopnv:
### Purpose of this PR

Currently, the `com.unity.test-framework.build` package is pinned at version 13 for the smoke tests project. Version 13 was released on the candidates (so yamato can find it as it implicitly references it) but not publicly (so opening the project locally and outside of Unity doesn't work).
In the meantime version 14 was released, so the utf team suggests publicly releasing this version and skipping version 13.
So this PR is bumping all the test project manifests to version 14.

---
### Testing status
ABV (matching master) + Smoke tests


theo-pnv:
Failures match master

sophiaaar (approved):

theo-pnv:
Confirmation from the utf team that preview.14 is now released publicly, and I'm able to open a project locally without any update to the manifest.json, so merging!

PR #4010, opened 2021-03-26
[URP]Camera Inspector Fixes

alex-vazquez:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1324574/

Fix the update of the output properties for overlay that are on the stack cameras of a Base Camera

- XR rendering was accessed even it was not enabled( missing `#if ENABLE_VR && ENABLE_XR_MODULE` )
- Fixing the update of AllowDynamic Resolution
- Fixing the update of the allowXR rendering (Was a TODO)
- Not updating cameras that are on the stack but they are of type Base
- Fixes removing cameras from the stack

---
### Testing status
- Check that the output properties are correctly modified

---
### Comments to reviewers
No changelog need as this was pushed recently by @lukaschod

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
@@ -78,7 +78,7 @@ namespace UnityEditor.Rendering.Universal
             base.OnEnable();
             settings.OnEnable();
 
-            m_SerializedCamera = new UniversalRenderPipelineSerializedCamera(serializedObject, settings);
+            m_SerializedCamera = new UniversalRenderPipelineSerializedCamera(serializedObject);
 
             m_CommonCameraSettingsFoldout = new SavedBool($"{target.GetType()}.CommonCameraSettingsFoldout", false);
             m_EnvironmentSettingsFoldout = new SavedBool($"{target.GetType()}.EnvironmentSettingsFoldout", false);
@@ -106,18 +106,30 @@ namespace UnityEditor.Rendering.Universal
                 drawHeaderCallback = rect => EditorGUI.LabelField(rect, Styles.cameras),
                 drawElementCallback = DrawElementCallback,
                 onSelectCallback = SelectElement,
-                onRemoveCallback = list =>
-                {
-                    m_SerializedCamera.cameras.DeleteArrayElementAtIndex(list.index);
-                    ReorderableList.defaultBehaviours.DoRemoveButton(list);
-                    m_SerializedCamera.Apply();
-
-                    // Force update the list as removed camera could been there
-                    m_TypeErrorCameras.Clear();
-                    m_OutputWarningCameras.Clear();
-                },
+                onRemoveCallback = RemoveCamera,
+                onCanRemoveCallback = CanRemoveCamera,
                 onAddDropdownCallback = AddCameraToCameraList
             };
+        }
+
+        bool CanRemoveCamera(ReorderableList list) => m_SerializedCamera.numCameras > 0;
+
+        void RemoveCamera(ReorderableList list)
+        {
+            // As multi selection is disabled, selectedIndices will only return 1 element, remove that element from the list
+            if (list.selectedIndices.Any())
+            {
+                m_SerializedCamera.cameras.DeleteArrayElementAtIndex(list.selectedIndices.First());
+            }
+            else
+            {
+                // Nothing selected, remove the last item on the list
+                ReorderableList.defaultBehaviours.DoRemoveButton(list);
+            }
+
+            // Force update the list as removed camera could been there
+            m_TypeErrorCameras.Clear();
+            m_OutputWarningCameras.Clear();
         }
 
         void SelectElement(ReorderableList list)
@@ -137,9 +149,8 @@ namespace UnityEditor.Rendering.Universal
             rect.height = EditorGUIUtility.singleLineHeight;
             rect.y += 1;
 
-            var element = m_SerializedCamera.cameras.GetArrayElementAtIndex(index);
-
-            var cam = element.objectReferenceValue as Camera;
+            (Camera camera, UniversalRenderPipelineSerializedCamera serializedCamera)overlayCamera = m_SerializedCamera[index];
+            Camera cam = overlayCamera.camera;
             if (cam != null)
             {
                 bool typeError = false;
@@ -158,7 +169,7 @@ namespace UnityEditor.Rendering.Universal
                 }
 
                 bool outputWarning = false;
-                if (IsStackCameraOutputDirty(cam))
+                if (IsStackCameraOutputDirty(cam, overlayCamera.serializedCamera))
                 {
                     outputWarning = true;
                     if (!m_OutputWarningCameras.Contains(cam))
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -396,115 +397,131 @@ private void UpdateStackCamerasOutput()
...
             {
-                camera.targetDisplay = targetDisplay;
+                cam.targetDisplay = targetDisplay;
                 isChanged = true;
             }
 
-            // Force same target display todo
-            StereoTargetEyeMask stereoTargetEye = (StereoTargetEyeMask)settings.targetEye.intValue;
-            if (camera.stereoTargetEye != stereoTargetEye)
+#if ENABLE_VR && ENABLE_XR_MODULE
```
lukaschod, line 466: I assume we do not use stereoTargetEye?

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineSerializedCamera.cs
```diff
@@ -33,14 +35,33 @@ internal class UniversalRenderPipelineSerializedCamera : ISerializedCamera
         public SerializedProperty allowXRRendering { get; }
 #endif
 
+        public (Camera camera, UniversalRenderPipelineSerializedCamera serializedCamera) this[int index]
```
lukaschod, line 38: TBH I am not really fan of anonymous type. As it simulates somewhat encapsulation, but does not give advantage of tieing data under ceceptual name. Is there a reason why we do not use struct here or saperate methods for accessing?

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
alex-vazquez: We are modifying directly the `allowXRRendering`:

```
void DrawXRRendering()
        {
            Rect controlRect = EditorGUILayout.GetControlRect(true);
            EditorGUI.BeginProperty(controlRect, Styles.xrTargetEye, m_SerializedCamera.allowXRRendering);
            int selectedValue = !m_SerializedCamera.allowXRRendering.boolValue ? 0 : 1;
            m_SerializedCamera.allowXRRendering.boolValue = EditorGUI.IntPopup(controlRect, Styles.xrTargetEye, selectedValue, Styles.xrTargetEyeOptions, Styles.xrTargetEyeValues) == 1;
            EditorGUI.EndProperty();
        }
```

That's why I decided to mantain and check the same property here, but let me know if you think is better to go to the `steroTargetEye`

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineSerializedCamera.cs
alex-vazquez: There is no deep reason why I decided to go with the anonymous type here, I thought that creating a class for just two members was little overkill. Let me know if you disagree.

phi-lira:
Adding @thomas-zeng for XR awareness

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -396,115 +397,131 @@ private void UpdateStackCamerasOutput()
...
+            // Force same target display
+            int selectedValue = !m_SerializedCamera.allowXRRendering.boolValue ? 0 : 1;
```
jenniferd-unity, line 468: shouldnt it be 
!boolValue? 0 : boolValue
?

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineSerializedCamera.cs
jenniferd-unity: I dont know enough here. but maybe we should compare with how we do things somewhere else in our codebase?
Martin may be able to help?

jenniferd-unity (approved):
Approving - supposing my question about boolValue was not relevant. if not, let us change it first

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineSerializedCamera.cs
alex-vazquez: @martin-unity3d  do you have some insight over this?

thomas-zeng (approved):

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -137,51 +137,49 @@ void DrawElementCallback(Rect rect, int index, bool isActive, bool isFocused)
             rect.height = EditorGUIUtility.singleLineHeight;
             rect.y += 1;
 
-            var element = m_SerializedCamera.cameras.GetArrayElementAtIndex(index);
-
-            var cam = element.objectReferenceValue as Camera;
```
martint-unity, line 142: Maybe keep this var so you don't have to do this overlayCamera.camer all the time

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineSerializedCamera.cs
martint-unity: no real insight but I think it looks probably more confusing then it is. Which means maybe simplify it like @lukaschod suggested.

martint-unity (commented):
lgtm, just have a few comments. No deal-breakers :)

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
alex-vazquez: we are casting the  bool value in an inverse way. So it is correct. In the other usages is done in the same way

alex-vazquez:
@lukaschod, @martint-unity could you have a look at this PR please?

martint-unity (approved):
LGTM

alex-vazquez:
@Unity-Technologies/gfx-qa-urp  could you have a look at this PR? This is blocking me on other PR's :)

Thank you

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineSerializedCamera.cs
alex-vazquez: I've seen this on our codebase, and HDRP is using it a lot and URP is not using at all xD. Maybe we can discuss it usages and see if we should move to proper structs

alex-vazquez:
@martint-unity could you do a quick pass on my last changes from errors found from @erikabar

martint-unity (approved):
LGTM

erikabar (approved):
Verified fixes for Camera Inspector; checked if there's a warning thrown when I am modifying Base camera's properties which is still in the stack of the other Base camera. Tested if "Fix" button works

lukaschod (approved):

PR #4013, opened 2021-03-26
Fixed error when disabling opaques

JulienIgnace-Unity:
### Purpose of this PR
Fixed error when disabling opaques
The error was caused by a missing implicit conversion operator from render graph TextureHandle to Texture.
Also moved implicit operators in the code to go from "most likely" to "less likely" for readability (all RTHandle can be converted to RTIs but not to RenderTexture, so RTI first then Texture then RenderTexture)

Steps to repro:
- Load any scene with master
- Override frame setting for camera and disable opaque objects
- Notice errors in the console

---
### Testing status
Tested the repro steps before and after the fix.


FrancescoC-unity (approved):

sebastienlagarde (approved):

JMargevics (approved):
Auto approving, Julien's testing coverage is sufficient 👍

PR #4014, opened 2021-03-26
Add virtual texturing to the path tracing limitations

JordanL8:
### Purpose of this PR
Added virtual texturing to the list of path tracing limitations. Also moved the path tracing limitations to the path tracing document.


JordanL8:
@remi-chapelain for awareness, since we have updated this section a few times in the past.

sebastienlagarde (approved):

PR #4016, opened 2021-03-26
[HDRP] Add Planar reflections coverage for SubSurfaceScattering in HDRP Runtime Tests

remi-chapelain:
### Purpose of this PR
Following [this PR](https://github.com/Unity-Technologies/Graphics/pull/3919), it was needed to prevent another regression so coverage has been added. 

Test scene BEFORE
001-HDTemplate

Test scene NOW
image

---
### Testing status
Nothing yet, launched yamato to get new screenshots. 

---
### Comments to reviewers
Fairly easy change, doesn't need any review.


sebastienlagarde:
Note: I will need to update the console test as well after the change land

PR #4018, opened 2021-03-26
Adapt URP to use RTHandles

sandy-carter-unity:
## Purpose of this PR
This PR aims to replace usage of RenderTargetHandle to use RTHandles instead in order to support setting `UnityEngine.XR.XRSettings.renderViewportScale` in URP.

RTHandle is a SRP core utility class that helps manage unity RenderTexture resources and has DynamicResolutionHandler support. It is currently being used by HDRP and RenderGraph modules. The class also has built-in support for render target dynamic scaling. This will add dynamic resolution support to URP and unifies URP’s render texture management. It also opens the way for URP to integrate with RenderGraph in the future.

---
## Testing status
We are currently testing these changes and adapting the tests to work on all thoses cases.
At the moment, 59/103 of the playmode tests pass.

---
## Comments to reviewers
This PR is still a work in progress. See **Left to do**.

## Color and Depth in the same RenderTarget

The nature of RTHandles prevents the combination of color and depth/stencil in the same RTHandle.
Therefore, and RTHandle is equivalent to one GPU render texture

## UniversalRenderer's intermediate color and intermediate depth

Previously we have this logic that dictates whether to create intermediate color or depth/stencil. There is a case where a color attachment is created with depth and the depth attachment is not.
```c#
/* ... */
if (cameraData.renderType == CameraRenderType.Base)
{
    RenderTargetHandle cameraTargetHandle = RenderTargetHandle.GetCameraTarget(cameraData.xr);

    m_ActiveCameraColorAttachment = (createColorTexture) ? m_CameraColorAttachment : cameraTargetHandle;
    m_ActiveCameraDepthAttachment = (createDepthTexture) ? m_CameraDepthAttachment : cameraTargetHandle;

    bool intermediateRenderTexture = createColorTexture || createDepthTexture;
   
    // Doesn't create texture for Overlay cameras as they are already overlaying on top of created textures.
    if (intermediateRenderTexture)
        CreateCameraRenderTarget(context, ref cameraTargetDescriptor, createColorTexture, createDepthTexture);
}
/* ... */
void CreateCameraRenderTarget(ScriptableRenderContext context, ref RenderTextureDescriptor descriptor, bool createColor, bool createDepth)
{
    /* ... */
    if (createColor)
    {
        /* ... */
        bool useDepthRenderBuffer = m_ActiveCameraDepthAttachment == RenderTargetHandle.CameraTarget;
        colorDescriptor.depthBufferBits = (useDepthRenderBuffer) ? k_DepthStencilBufferBits : 0;
        cmd.GetTemporaryRT(m_ActiveCameraColorAttachment.id, colorDescriptor, FilterMode.Bilinear);
    }
    if (createDepth)
    {
        /* ... */
        cmd.GetTemporaryRT(m_ActiveCameraDepthAttachment.id, depthDescriptor, FilterMode.Point);
    }
    /* ... */
}
```

Converting this to RTHandles requires removing all GetTemporaryRTs and replacing them with RTHandles which are allocated to last longer than one frame.
When they are (re)allocated, the reference changes, this is why `CreateCameraRenderTarget` is moved to before the assignment.
Finally and more importantly, the case of `createColorTexture=true, createDepthTexture=false` is replaced so that if an intermediate color RTHandle is used, an intermediate depth RTHandle is also used.
In terms of allocated GPU resourced, it should be the same due to the above use of `depthBufferBits` on the color attachment.

---
## Left to do
- [ ] Fix changes in order to pass remaining tests.
- [ ] Compatibility with RenderTargetHandle: Adapt Scriptable render pass to support RenderTargetHandle as well as RTHandle and reintroduce RenderTargetHandle which was removed during conversion.

### DIFF
--- a/com.unity.render-pipelines.universal/Shaders/PostProcessing/PaniniProjection.shader
+++ b/com.unity.render-pipelines.universal/Shaders/PostProcessing/PaniniProjection.shader
@@ -100,6 +100,7 @@ Shader "Hidden/Universal Render Pipeline/PaniniProjection"
         half4 Frag(Varyings input) : SV_Target
         {
             UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
+            input.uv = input.uv / _RTHandleScale.xy;
 
             #if _GENERIC
             float2 proj_pos = Panini_Generic((2.0 * input.uv - 1.0) * _Params.xy * _Params.w, _Params.z);
@@ -109,8 +110,9 @@ Shader "Hidden/Universal Render Pipeline/PaniniProjection"
 
             float2 proj_ndc = proj_pos / _Params.xy;
             float2 coords = proj_ndc * 0.5 + 0.5;
+            coords = clamp(coords, float2(0, 0), float2(1, 1));
 
-            return SAMPLE_TEXTURE2D_X(_SourceTex, sampler_LinearClamp, coords);
+            return SAMPLE_TEXTURE2D_X(_SourceTex, sampler_LinearClamp, coords * _RTHandleScale.xy);
         }
 
     ENDHLSL


com.unity.render-pipelines.universal/Runtime/Passes/CopyColorPass.cs
```diff
@@ -77,30 +65,33 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
-                ScriptableRenderer.SetRenderTarget(cmd, opaqueColorRT, BuiltinRenderTextureType.CameraTarget, clearFlag,
+                ScriptableRenderer.SetRenderTarget(cmd, destination ?? k_CameraTarget, k_CameraTarget, clearFlag,
                     clearColor);
 
                 bool useDrawProceduleBlit = renderingData.cameraData.xr.enabled;
                 switch (m_DownsamplingMethod)
                 {
                     case Downsampling.None:
-                        RenderingUtils.Blit(cmd, source, opaqueColorRT, m_CopyColorMaterial, 0, useDrawProceduleBlit);
+                        RenderingUtils.Blit(cmd, source, destination ?? k_CameraTarget, m_CopyColorMaterial, 0, useDrawProceduleBlit);
```
phi-lira, line 75: Not sure why destination would be null here. But it seems the logic is to bind the camera target in case destination is null, you can get camera color target by using renderingData.cameraData.renderer.cameraColorTexture.

com.unity.render-pipelines.universal/Runtime/Passes/CopyColorPass.cs
```diff
@@ -11,13 +11,15 @@ namespace UnityEngine.Rendering.Universal.Internal
     /// </summary>
     public class CopyColorPass : ScriptableRenderPass
     {
+        static readonly RTHandle k_CameraTarget = RTHandles.Alloc(BuiltinRenderTextureType.CameraTarget);
```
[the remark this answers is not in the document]
phi-lira: This will also break the work from @robin-demoor that exposes a swapable camera color buffer.

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -14,15 +14,15 @@ namespace UnityEngine.Rendering.Universal.Internal
     /// </summary>
     public class CopyDepthPass : ScriptableRenderPass
     {
-        private RenderTargetHandle source { get; set; }
-        private RenderTargetHandle destination { get; set; }
-        internal bool AllocateRT  { get; set; }
+        static readonly RTHandle k_CameraTarget = RTHandles.Alloc(BuiltinRenderTextureType.CameraTarget);
```
phi-lira, line 17: Same problem with CopyColorPass I mentioned above.

com.unity.render-pipelines.universal/Runtime/Passes/DepthNormalOnlyPass.cs
```diff
@@ -67,13 +65,8 @@ public void Setup(RenderTextureDescriptor baseDescriptor, RenderTargetHandle dep
...
         {
             if (this.allocateNormal)
-                cmd.GetTemporaryRT(normalHandle.id, normalDescriptor, FilterMode.Point);
-            if (this.allocateDepth)
-                cmd.GetTemporaryRT(depthHandle.id, depthDescriptor, FilterMode.Point);
-            ConfigureTarget(
-                new RenderTargetIdentifier(normalHandle.Identifier(), 0, CubemapFace.Unknown, -1),
-                new RenderTargetIdentifier(depthHandle.Identifier(), 0, CubemapFace.Unknown, -1)
-            );
+                cmd.GetTemporaryRT(Shader.PropertyToID(normal.name), normalDescriptor, FilterMode.Point);
```
phi-lira, line 68: This is a screenspace camera texture, so I think it has to be converted to RTHandle otherwise there will be a clash of two dynamic scaling systems.

phi-lira: What happened to `allocateDepth` and `depthRT` creation?

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -12,10 +12,10 @@ namespace UnityEngine.Rendering.Universal.Internal
     public class DepthOnlyPass : ScriptableRenderPass
     {
         private static readonly ShaderTagId k_ShaderTagId = new ShaderTagId("DepthOnly");
+        static readonly RTHandle k_CameraTarget = RTHandles.Alloc(BuiltinRenderTextureType.CameraTarget);
```
phi-lira, line 15: Same problem as CopyColor and CopyDepth.

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -26,7 +26,7 @@ public FinalBlitPass(RenderPassEvent evt, Material blitMaterial)
         /// </summary>
         /// <param name="baseDescriptor"></param>
         /// <param name="colorHandle"></param>
-        public void Setup(RenderTextureDescriptor baseDescriptor, RenderTargetHandle colorHandle)
+        public void Setup(RenderTextureDescriptor baseDescriptor, RTHandle colorHandle)
```
phi-lira, line 29: Can't change public API. Make override or consider making obsolete.

com.unity.render-pipelines.universal/Runtime/Passes/GBufferPass.cs
```diff
@@ -65,16 +63,12 @@ public override void Configure(CommandBuffer cmd, RenderTextureDescriptor camera
                 if (i == m_DeferredLights.GBufferNormalSmoothnessIndex && m_DeferredLights.HasNormalPrepass)
                     continue;
 
-                RenderTextureDescriptor gbufferSlice = cameraTextureDescriptor;
-                gbufferSlice.depthBufferBits = 0; // make sure no depth surface is actually created
-                gbufferSlice.stencilFormat = GraphicsFormat.None;
-                gbufferSlice.graphicsFormat = m_DeferredLights.GetGBufferFormat(i);
-                cmd.GetTemporaryRT(m_DeferredLights.GbufferAttachments[i].id, gbufferSlice);
```
phi-lira, line 72: How are you allocating / deallocating gbuffer RTs?

com.unity.render-pipelines.universal/Runtime/Passes/MainLightShadowCasterPass.cs
```diff
@@ -121,9 +119,15 @@ public bool Setup(ref RenderingData renderingData)
...
-            m_MainLightShadowmapTexture = ShadowUtils.GetTemporaryShadowTexture(m_ShadowmapWidth,
-                m_ShadowmapHeight, k_ShadowmapBufferBits);
-            ConfigureTarget(new RenderTargetIdentifier(m_MainLightShadowmapTexture), GraphicsFormat.ShadowAuto, m_ShadowmapWidth, m_ShadowmapHeight, 1, true);
+            if (m_MainLightShadowmap == null || m_MainLightShadowmap.rt.width != m_ShadowmapWidth ||
+                m_MainLightShadowmap.rt.height != m_ShadowmapHeight)
+            {
+                m_MainLightShadowmap?.Release();
+                m_MainLightShadowmap = ShadowUtils.AllocShadowRTHandle(m_ShadowmapWidth, m_ShadowmapHeight,
+                    k_ShadowmapBufferBits, "_MainLightShadowmapTexture");
+            }
```
phi-lira, line 128: Textures should be created and disposes per-camera, otherwise you keep resources around for no reason.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -20,11 +20,18 @@ namespace UnityEngine.Rendering.Universal.Internal
     /// </summary>
     public class PostProcessPass : ScriptableRenderPass
     {
+        static readonly RTHandle k_CameraTarget = RTHandles.Alloc(BuiltinRenderTextureType.CameraTarget);
+        static readonly RTHandle k_CurrentActive = RTHandles.Alloc(BuiltinRenderTextureType.CurrentActive);
```
phi-lira, line 24: No need to alloc textures for these. 
 You can get camera target from renderer and current active from BuiltinRenderTextureType.CurrentActive

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -114,42 +123,112 @@ public PostProcessPass(RenderPassEvent evt, PostProcessData data, Material blitM
...
-            {
-                ShaderConstants._BloomMipUp[i] = Shader.PropertyToID("_BloomMipUp" + i);
-                ShaderConstants._BloomMipDown[i] = Shader.PropertyToID("_BloomMipDown" + i);
-            }
-
             m_MRT2 = new RenderTargetIdentifier[2];
             m_ResetHistory = true;
             base.useNativeRenderPass = false;
+
+            m_TempTarget = RTHandles.Alloc(
```
phi-lira, line 130: You should allocate texture upon usage, otherwise you keep resources allocated even though not used.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -133,19 +133,21 @@ public enum ScriptableRenderPassInput
     /// </summary>
     [MovedFrom("UnityEngine.Rendering.LWRP")] public abstract partial class ScriptableRenderPass
     {
+        static readonly RTHandle k_CameraTarget = RTHandles.Alloc(BuiltinRenderTextureType.CameraTarget);
+
         public RenderPassEvent renderPassEvent { get; set; }
 
-        public RenderTargetIdentifier[] colorAttachments
+        public RTHandle[] colorAttachments
```
phi-lira, line 140: Can't change public API, add overload.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -190,17 +192,17 @@ public Color clearColor
 
 
         internal GraphicsFormat[] renderTargetFormat { get; set; }
-        RenderTargetIdentifier[] m_ColorAttachments = new RenderTargetIdentifier[] {BuiltinRenderTextureType.CameraTarget};
-        RenderTargetIdentifier m_DepthAttachment = BuiltinRenderTextureType.CameraTarget;
+        RTHandle[] m_ColorAttachments = new RTHandle[] {k_CameraTarget};
```
phi-lira, line 195: Not sure how to solve this but you should be able to get RTHandle reference to camera target without allocating or maybe use `null`?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -236,13 +238,13 @@ public void ConfigureInput(ScriptableRenderPassInput passInput)
         /// <param name="colorAttachment">Color attachment identifier.</param>
         /// <param name="depthAttachment">Depth attachment identifier.</param>
         /// <seealso cref="Configure"/>
-        public void ConfigureTarget(RenderTargetIdentifier colorAttachment, RenderTargetIdentifier depthAttachment)
+        public void ConfigureTarget(RTHandle colorAttachment, RTHandle depthAttachment)
```
phi-lira, line 241: Can't change public API

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -191,23 +190,89 @@ public UniversalRenderer(UniversalRendererData data) : base(data)
 
             // RenderTexture format depends on camera and pipeline (HDR, non HDR, etc)
             // Samples (MSAA) depend on camera and pipeline
-            m_CameraColorAttachment.Init("_CameraColorTexture");
-            m_CameraDepthAttachment.Init("_CameraDepthAttachment");
-            m_DepthTexture.Init("_CameraDepthTexture");
-            m_NormalsTexture.Init("_CameraNormalsTexture");
-            if (this.renderingMode == RenderingMode.Deferred)
+            m_CameraAttachments.color = RTHandles.Alloc(
```
phi-lira, line 193: Should create upon demand

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -191,23 +190,89 @@ public UniversalRenderer(UniversalRendererData data) : base(data)
...
+            m_DepthTexture = RTHandles.Alloc(
+                Vector2.one,
+                depthBufferBits: DepthBits.Depth32,
+                colorFormat: GraphicsFormat.DepthAuto,
+                filterMode: FilterMode.Point,
+                dimension: TextureDimension.Tex2D,
+                enableMSAA: false,
+                name: "_CameraDepthTexture");
+            m_NormalsTexture = RTHandles.Alloc(URPShaderIDs._CameraNormalsTexture, "_CameraNormalsTexture");
+            if (renderingMode == RenderingMode.Deferred)
```
phi-lira, line 229: Should create upon demand, ideally in the GBuffer pass like it was with GetTemporaryRT.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -191,23 +190,89 @@ public UniversalRenderer(UniversalRendererData data) : base(data)
...
+                m_GBufferHandles[(int)DeferredLights.GBufferHandles.Lighting] = null; // Set to be m_ActiveCameraAttachments during runtime
+                m_GBufferHandles[(int)DeferredLights.GBufferHandles.ShadowMask] = RTHandles.Alloc(
+                    Vector2.one,
+                    colorFormat: DeferredLights.GetGBufferFormat(DeferredLights.GBufferHandles.ShadowMask),
+                    name: "_GBuffer4");
             }
-            m_OpaqueColor.Init("_CameraOpaqueTexture");
-            m_DepthInfoTexture.Init("_DepthInfoTexture");
-            m_TileDepthInfoTexture.Init("_TileDepthInfoTexture");
+            m_OpaqueColor = RTHandles.Alloc(size =>
```
phi-lira, line 254: Should create upon demand.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -379,41 +457,27 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
                 createDepthTexture |= createColorTexture;
                 createColorTexture = createDepthTexture;
             }
+            RTHandle cameraTarget = k_CameraTarget;
+#if ENABLE_VR && ENABLE_XR_MODULE
+            if (cameraData.xr.enabled)
+                cameraTarget = RTHandles.Alloc(cameraData.xr.renderTarget);
```
phi-lira, line 463: It's not always in XR that you need to create intermediate color texture.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -435,14 +499,12 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
                         int gbufferNormalIndex = m_DeferredLights.GBufferNormalSmoothnessIndex;
-                        m_DepthNormalPrepass.Setup(cameraTargetDescriptor, m_ActiveCameraDepthAttachment, m_GBufferHandles[(int)DeferredLights.GBufferHandles.NormalSmoothness]);
+                        m_DepthNormalPrepass.Setup(cameraTargetDescriptor, m_ActiveCameraAttachments.depth, m_GBufferHandles[(int)DeferredLights.GBufferHandles.NormalSmoothness]);
 
                         // Change the normal format to the one used by the gbuffer.
                         RenderTextureDescriptor normalDescriptor = m_DepthNormalPrepass.normalDescriptor;
                         normalDescriptor.graphicsFormat = m_DeferredLights.GetGBufferFormat(gbufferNormalIndex);
                         m_DepthNormalPrepass.normalDescriptor = normalDescriptor;
-                        // Depth is allocated by this renderer.
-                        m_DepthNormalPrepass.allocateDepth = false;
```
phi-lira, line 445: ?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -492,26 +554,48 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+                m_CopyDepthPass.Setup(m_ActiveCameraAttachments.depth, m_DepthTexture);
 
                 EnqueuePass(m_CopyDepthPass);
             }
 
             // For Base Cameras: Set the depth texture to the far Z if we do not have a depth prepass or copy depth
             if (cameraData.renderType == CameraRenderType.Base && !requiresDepthPrepass && !requiresDepthCopyPass)
             {
-                Shader.SetGlobalTexture(m_DepthTexture.id, SystemInfo.usesReversedZBuffer ? Texture2D.blackTexture : Texture2D.whiteTexture);
+                Shader.SetGlobalTexture(Shader.PropertyToID(m_DepthTexture.name), SystemInfo.usesReversedZBuffer ? Texture2D.blackTexture : Texture2D.whiteTexture);
```
phi-lira, line 565: This Shader.PropertyToID could be cached

phi-lira (changes requested):

com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
```diff
@@ -86,7 +44,7 @@ public static void CreateNormalMapRenderTexture(this IRenderPass2D pass, Renderi
             {
                 if (pass.rendererData.isNormalsRenderTargetValid)
                 {
-                    cmd.ReleaseTemporaryRT(pass.rendererData.normalsRenderTarget.id);
+                    cmd.ReleaseTemporaryRT(Shader.PropertyToID(pass.rendererData.normalsRenderTarget.name));
```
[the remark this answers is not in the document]
sandy-carter-unity: You are correct. This is an intermediate step in the conversion to RTHandles. I still need to finish the 2D renderer.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
sandy-carter-unity: What's the proper strategy here? Should I use a property with get to lazy alloc such as:

```c#
private RTHandle m_OpaqueColor {
    get {
        if (m_OpaqueColor_ == null) {
            m_OpaqueColor_ = RTHandles.Alloc(...);
        }
        return m_OpaqueColor_ ;
    }
}
```

com.unity.render-pipelines.universal/Runtime/2D/Passes/Render2DLightingPass.cs
```diff
@@ -8,20 +8,6 @@ namespace UnityEngine.Experimental.Rendering.Universal
...
-        private static readonly int k_RendererColorID = Shader.PropertyToID("_RendererColor");
-        private static readonly int k_CameraSortingLayerTextureID = Shader.PropertyToID("_CameraSortingLayerTexture");
-
-        private static readonly int[] k_ShapeLightTextureIDs =
-        {
-            Shader.PropertyToID("_ShapeLightTexture0"),
-            Shader.PropertyToID("_ShapeLightTexture1"),
-            Shader.PropertyToID("_ShapeLightTexture2"),
-            Shader.PropertyToID("_ShapeLightTexture3")
-        };
```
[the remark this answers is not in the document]
sandy-carter-unity: I can revert this. I was following the style of HDRP's `com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDStringConstants.cs`

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
sandy-carter-unity: We can no longer allocate within a pass because this would change the reference in this class from null to an RTHandle.
Previously we would have the same `RenderTargetHandle` with an id that would `GetTemporaryRT` but otherwise wouldn't change the object.
Now It's trickier because you have times where would use the Handle before it is allocated, so I moved to allocate within `UniversalRenderer`.

kaychang-unity (approved):

com.unity.render-pipelines.universal/Runtime/2D/Passes/Render2DLightingPass.cs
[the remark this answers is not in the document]
phi-lira: It's ok to change the style if needed, but I definitely don't recommend to do that in a big PR like this. It makes code review more likely to miss issues and introduces risk. I prefer any change like this to go into separate PR, but let's discuss first if that should be done.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
phi-lira: The above can work if you still make sure to dealloc / cleanup per-camera.
  You want to avoid having resources lingering between multiple frames if they are not needed.
  
  Another approach is to delegate to the pass to alloc/dealloc as it was with GetTemRT/ReleaseTempRT

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
phi-lira: If you would try to use Handle before the resource is allocated it would also be a problem in GetTemporaryRT, unless the handle was something special like BuiltinRenderTextureType. We should probably discuss further this. Let's setup a call so we can go through the code together and look at some cases.

sandy-carter-unity:
Removed the commit which introduces `URPShaderIDs`, this should make the diff less noisy.

com.unity.render-pipelines.universal/Runtime/2D/Passes/Render2DLightingPass.cs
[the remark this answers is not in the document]
sandy-carter-unity: I removed this changeset

com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
```diff
@@ -171,19 +174,11 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
                 }
             }
 
-            RenderTargetHandle colorTargetHandle;
-            RenderTargetHandle depthTargetHandle;
-
-            CommandBuffer cmd = CommandBufferPool.Get();
-            using (new ProfilingScope(cmd, m_ProfilingSampler))
```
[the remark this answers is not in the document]
sandy-carter-unity: Because `cmd` was no longer used in this scope. Perhaps I should re-introduce it with null?

com.unity.render-pipelines.universal/Runtime/Passes/ColorGradingLutPass.cs
```diff
@@ -85,7 +85,7 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
                 var material = hdr ? m_LutBuilderHdr : m_LutBuilderLdr;
                 var desc = new RenderTextureDescriptor(lutWidth, lutHeight, format, 0);
                 desc.vrUsage = VRTextureUsage.None; // We only need one for both eyes in VR
-                cmd.GetTemporaryRT(m_InternalLut.id, desc, FilterMode.Bilinear);
+                cmd.GetTemporaryRT(Shader.PropertyToID(m_InternalLut.name), desc, FilterMode.Bilinear);
```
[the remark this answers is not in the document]
sandy-carter-unity: Correct. I haven't converted this one yet.

com.unity.render-pipelines.universal/Runtime/Passes/GBufferPass.cs
sandy-carter-unity: It is done in UniversalRenderer

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
sandy-carter-unity: > Another approach is to delegate to the pass to alloc/dealloc as it was with GetTemRT/ReleaseTempRT

Is the goal to Allocate and Free on each frame? This kind of goes against the dynamic resizing feature of RTHandles that do linger.

com.unity.render-pipelines.universal/Shaders/PostProcessing/FinalPost.shader
```diff
@@ -115,7 +115,7 @@ Shader "Hidden/Universal Render Pipeline/FinalPost"
 
             #if _FILM_GRAIN
```
robin-demoor, line 116: Is it possible that the uv at line 57 should also be *_RTHandleScale?

robin-demoor:
I think SSAO.hlsl/ScreenSpaceAmbientOcclusion.cs also needs the * _RTHandleScale adjustment

com.unity.render-pipelines.universal/Runtime/2D/Passes/Utility/RendererLighting.cs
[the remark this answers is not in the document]
sandy-carter-unity: Don't think I will end up converting these to RTHandles since they don't need to use dynamic scaling.

sandy-carter-unity:
Splitting this PR with P1 in https://github.com/Unity-Technologies/Graphics/pull/5346

PR #4019, opened 2021-03-26
Removing the duplicate legacy cginc files from the built-in target

joshua-davis:
---
### Purpose of this PR
For iteration, the legacy built-in target shaders (cginc) were copied into the built-in target for shader graph. A few changes were necessary to these files to avoid duplicate symbols. These definitions were instead moved to the SRP core shader files (mostly just ifdef guards) so as to prevent any changes required to trunc.

---
### Testing status
Ran the tests for the built-in target and verified they all still worked.

---
### Comments to reviewers


com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
```diff
@@ -1,5 +1,5 @@
-#ifndef UNITY_IMAGE_BASED_LIGHTING_INCLUDED
```
joshua-davis, line 1: This include guard is identical to one in the cginc files we have to include.

com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/Editor/ShaderGraph/Targets/BuiltInTarget.cs
```diff
@@ -874,6 +880,13 @@ static class CoreKeywordDescriptors
             referenceName = "USELEGACYSPRITEBLOCKS",
             type = KeywordType.Boolean,
         };
+
+        public static readonly KeywordDescriptor BuiltInTargetAPI = new KeywordDescriptor()
```
joshua-davis, line 884: Add a define for the built-in target passes so we can ifdef on things

com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/ShaderLibrary/Shim/HLSLSupportShim.hlsl
```diff
@@ -3,6 +3,13 @@
 
 #define HLSL_SUPPORT_INCLUDED
 
+#if !defined(SHADER_API_GLES)
```
joshua-davis, line 6: All changes in this file were copied out of HLSLSupport.cginc and were necessary defines to remove the duplicate files.

jessebarker (approved):
So we're now just including the cginc files directly from Unity/trunk (effectively).  That's great.

joshua-davis:
> So we're now just including the cginc files directly from Unity/trunk (effectively). That's great.

Yep, should just be linked to trunc now

PR #4020, opened 2021-03-26
Use UnityEngine.Assertions for assert functionality

jessebarker:
---
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1322069/

The addition of AssertHelpers put an implicit dependency on the test framework package (to use NUnit's assert functionality).  That causes a problem if someone removes that package from a project.  Rather than make this dependency explicit, this PR uses the assert functionality in UnityEngine.Assertions.

---
### Testing status
Manual testing following the repro steps in the bug
Yamato details to follow

---
### Comments to reviewers
Weirdly, directly calling Assert.Fail() came up with that function being undefined, so I just used the actual internals of what the (possibly obsolete) Fail function does (the comments hint this is the right direction moving forward).


Nightmask3 (approved):
looks good to me! 
Yeah its odd that the UnityEngine assertions library chooses not to implement Assert.Fail given that the intent seems to be a general purpose replacement for NUnit (if I understand the intent correctly) but 🤷‍♂️

com.unity.shadergraph/Editor/Util/AssertHelpers.cs
```diff
@@ -1,20 +1,20 @@
-using NUnit.Framework;
+using UnityEngine.Assertions;
 
 namespace UnityEditor.ShaderGraph
 {
     static class AssertHelpers
     {
-        public static void IsNotNull(object anObject, string message, params object[] args)
```
joshua-davis, line 7: It seems no one was using the args, but you could keep this in as it can be handy (looks to be for string formatting). It looks like NUnit defines this as:
```
static public void Fail(string? message, params object?[]? args)
        {
            if (message == null) message = string.Empty;
            else if (args != null && args.Length > 0)
                message = string.Format(message, args);

            ReportFailure(message);
        }
```

If you want to leave it alone that's fine too as string interpolation is an easy enough replacement

jessebarker: I wasn't sure if that would cause problems (function signature mismatch) as our internal stuff doesn't define them the same way NUnit does.  I don't feel strongly one way or the other, but none of the actual calls to these actually provided that final argument, so I dropped it.

marctem: We could keep the params arg on our Fail call, but the Unity version of IsNotNull doesn't offer it so we'd have to implement it ourselves. I think consistency of not including it is fine; we can always add it later if absolutely need it.

marctem (approved):

PR #4025, opened 2021-03-29
[2021.2] Adding _SURFACE_TYPE_TRANSPARENT keyword to URP and URP ShaderGraph shaders

ellioman:
# Purpose of this PR
Currently we have a bug in SSAO where it being applied to transparent objects.
In order to fix that I've added a new keyword (`_SURFACE_TYPE_TRANSPARENT`) to URP shaders. This is identical to the Keyword HDRP uses. A Material Upgrader was also added to properly set the keyword on materials.

# Example of the bug
### Original scene
Screen Shot 2021-03-26 at 11 33 28

### Same Scene with a opaque plane above it
Screen Shot 2021-03-26 at 11 33 08

### Same Scene with a transparent plane (alpha = 1.0) above it
Screen Shot 2021-03-26 at 11 32 01

# Testing
### What has been tested
- Tested manually in the editor (OS X). 

### What needs testing
- Make sure the fix works on all platforms

# Automated testing
- A new Test scene will be added once #3174 is merged in to master.

# Yamato 
-> when I take the PR out of draft mode


com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalTarget.cs
```diff
@@ -865,6 +865,15 @@ static class CoreKeywordDescriptors
...
             type = KeywordType.Boolean,
         };
+
+        public static KeywordDescriptor SurfaceTypeTransparent = new KeywordDescriptor()
+        {
+            displayName = "Surface Type Transparent",
+            referenceName = "_SURFACE_TYPE_TRANSPARENT",
+            type = KeywordType.Boolean,
+            definition = KeywordDefinition.ShaderFeature,
+            scope = KeywordScope.Global,
```
phi-lira, line 875: Shouldn't it be local keyword?

phi-lira (approved):
LGTM. Missing changelog.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Targets/UniversalTarget.cs
ellioman: Yes, and fragment only. Fixed.

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/MaterialPostprocessor.cs
```diff
@@ -256,6 +256,23 @@ static void UpgradeV3(Material material, ShaderPathID shaderID)
...
         static void UpgradeV4(Material material, ShaderPathID shaderID)
         {}
+
+        static void UpgradeV5(Material material, ShaderPathID shaderID)
+        {
+            var propertyID = Shader.PropertyToID("_Surface");
+            if (material.HasProperty(propertyID))
+            {
+                float surfaceType = material.GetFloat(propertyID);
+                if (surfaceType >= 1.0f)
```
eh-unity, line 266: Should this be an enum instead of 1? If it's enable/disable i.e. 1/0 case, then I'd probably test against 0.

eh-unity (approved):
Looks good.

aleksandrasdzikia (approved):
Looks good, the issue is fixed. Ran on OnePlus Nord, Galaxy Note10, Huawei P20 and p40 Lite, iPhone SE.

Verasl (approved):
LGTM

PR #4026, opened 2021-03-29
Fix spot light near plane gizmo

robcupisz:
Spot light's (cone shape) gizmo was drawing the shadow's near plane at the wrong distance - not matching where the near plane actually is placed for the shadow projection.

The distance was incorrect for cone-shaped spot lights, but correct for pyramid-shaped spot lights, so here's a before and after comparison. Also the vertical quad is placed at the minimum distance to the light where it still casts shadow - as another sanity check for where the near plane lies.

Before, gizmos for cone-shaped and pyramid-shaped spot lights overlayed:

spot-near-plane-before

After, gizmos for cone-shaped and pyramid-shaped spot lights overlayed, near planes now matching:

spot-near-plane-after

After, cone-shaped spot light and its near plane matching the closes position for a caster to still not be clipped by the near plane:

spot-near-plane-after2

---
### Testing status
Manually tested, since this doesn't change any other behaviour than the gizmo.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.Handles.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.Handles.cs
@@ -187,9 +187,8 @@ namespace UnityEditor.Rendering.HighDefinition
 
             if (shadowPlaneDistance > 0)
             {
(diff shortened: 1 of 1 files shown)


sebastienlagarde:
Thanks for this fix!

sebastienlagarde:
could you add a changelog? thanks

adrien-de-tocqueville (approved):

remi-chapelain (approved):
Looking good, no need for futher testing ! ✔️

PR #4031, opened 2021-03-29
SceneObjectIDMap: Fix a nasty bug where game object to entry index dictionary would get out of sync from the entries list

pastasfuture:
SceneObjectIDMapBug

Fix a bug where game object to entry index dictionary would get out of sync from the entries list after an insertion occurred to fill a hole. This resulted in a bug where reflection probes would occasionally generate the same bake texture filename as each other, would overwrite each others textures, and would point to the same texture after the bake completed. Issue can be reproduced by baking, removing reflection probes, adding new reflection probes, and baking again.


sebastienlagarde:
Recreate the PR here: https://github.com/Unity-Technologies/Graphics/pull/4031

PR #4037, opened 2021-03-30
Soft shadow improvements

robcupisz:
DRAFT PR - WORK IN PROGRESS

PCSS sampling improvements - instead of sampling in a flat disk pattern, we now use a visibility cone (or: pyramid), with its edges aimed at the edges of the shadow's near plane. The idea is that anything outside of that cone has no business contributing to the shadow of the currently shaded pixel, both in blocker search and final coverage.

Also makes area lights use PCSS when the shadow filtering quality is set to Very High.

Before
pcss-before0
pcss-before1

After
pcss-after0
pcss-after1


com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAlgorithms.hlsl
```diff
@@ -154,6 +154,11 @@ float EvalShadow_PunctualDepth(HDShadowData sd, Texture2D tex, SamplerComparison
 //
 float EvalShadow_AreaDepth(HDShadowData sd, Texture2D tex, float2 positionSS, float3 positionWS, float3 normalWS, float3 L, float L_dist, bool perspective)
 {
+    // TODO: implement less messily. Also probably shouldn't even rely on a global quality setting, but on a per-light shadow algorithm switch.
+#ifdef SHADOW_HIGH // Do PCSS
+    return EvalShadow_PunctualDepth(sd, tex, s_linear_clamp_compare_sampler, positionSS, positionWS, normalWS, L, L_dist, perspective);
```
sebastienlagarde, line 159: So if we do this, we should move the definition of the algorithm used in the HDShadowAlgorithms.hlsl and do same kind of thing that we do for the other algorithm, so it is configurable

#ifdef SHADOW_ULTRA_LOW
#define PUNCTUAL_FILTER_ALGORITHM(sd, posSS, posTC, tex, samp, bias) SampleShadow_PCF_Tent_3x3(sd.isInCachedAtlas ? _CachedShadowAtlasSize.zwxy : _ShadowAtlasSize.zwxy, posTC, tex, samp, bias)
#define DIRECTIONAL_FILTER_ALGORITHM(sd, posSS, posTC, tex, samp, bias) SampleShadow_Gather_PCF(_CascadeShadowAtlasSize.zwxy, posTC, tex, samp, bias)
#elif defined(SHADOW_LOW)
#define PUNCTUAL_FILTER_ALGORITHM(sd, posSS, posTC, tex, samp, bias) SampleShadow_PCF_Tent_3x3(sd.isInCachedAtlas ? _CachedShadowAtlasSize.zwxy : _ShadowAtlasSize.zwxy, posTC, tex, samp, bias)
...
#endif

sebastienlagarde: >but on a per-light shadow algorithm switch.
That's will be an issue.
For this, either you add new variant (and combination will explode) or you do a dynamic branch (and VGPR will explode for area lights (PCSS used way more VGPR than EVSM). 

So for now we will keep it as global.

robertcupisz: Yup, it just needs a bit of a refactor, since the entire EvalShadow_AreaDepth() needs to reuse what happens inside EvalShadow_PunctualDepth(). In other words it's not sufficient to define the algo for area lights at the same level as PUNCTUAL_FILTER_ALGORITHM

robertcupisz: > So if we do this, we should move the definition of the algorithm used in the HDShadowAlgorithms.hlsl and do same kind of thing that we do for the other algorithm, so it is configurable

Yup, it just needs a bit of a refactor, since the entire EvalShadow_AreaDepth() needs to reuse what happens inside EvalShadow_PunctualDepth(). In other words it's not sufficient to define the algo for area lights at the same level as PUNCTUAL_FILTER_ALGORITHM

FrancescoC-unity:
So the sampling improvements is something we do want provided the fitting for the softness is verified again so ping me when ready to be out of draft. 

Regarding the PCSS for area light, we should add a new tier; I will do a PR with a new "Very High" setting that will do that for area lights and link the PR here.

sebastienlagarde:
Hey, any update on this one?

Passeridae:
Is it going to be merged soon? Our team really needs this shadow improvement;)

robertcupisz:
Will be done as soon as I'm back from a break!  @sebastienlagarde: I'll DM you

sebastienlagarde:
Landed in trunk

PR #4043, opened 2021-03-30
Add control of the subdivision of the APV structure

alelievr:
### Purpose of this PR

- Added a field to control how much we can subdivision inside a probe volume
image
This field can be used to either set a minimum subdivision (at least subdivide of x) or a maximum subdivision (don't subdivide above x).
Internally this field is made of 2 float values between 0 and 1, then these two values are remapped using the settings of the probe profile asset to give the min and max subdiv levels.

https://user-images.githubusercontent.com/6877923/113013690-340a6200-917c-11eb-8682-17b793e8466f.mp4

- Optimized a lot the C# position baking process went from >3h (had to kill the process) to ~15s for the HDRP template with a probe min distance of 1m.
Specifically, I optimized the probe deduplication code (replaced the linear search with a dictionary, this is the biggest win because in some cases it looked for duplication in lists of ~900k positions). Then I optimized the OBB culling code (with a sphere check), once again it's a big win because all the bricks are tested against this culling function before being subdivided.

- Optimized the Gizmo rendering, for a scene with a lot of bricks like the template and a pretty high density of probes (1 every meter), the Gizmo rendering was taking up to 800ms. I converted all the brick gizmo code by a DrawMesh() which completely removed the cost of rendering these gizmos (>10ms now, so something like 80x improvement).

- Replaced the Cell Size field of the probe volume profile with an enum, this may not be the final design but it avoids the issue with mismatching sizes between Cell Size and Brick size, explanation here: https://unity.slack.com/archives/C0108P9UH6H/p1616775586133000

https://user-images.githubusercontent.com/6877923/113013614-248b1900-917c-11eb-8128-a0d57ebfe3fc.mp4

---
### Testing status
Everything was manually tested because we don't have any automated test yet, so it's possible there are a few issue left.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffeature%252Fprobe-hint-volume/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/6271942/job

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
@@ -30,6 +30,8 @@ namespace UnityEngine.Rendering
             CellSize243 = 243,
         }
 
+        // This field will be replaced by something else (probably a distance based setting in meter) when the artists decide
+        // what they want. So, we shouldn't rely on this information too much.
         /// <summary>
         /// The size of a Cell.
         /// </summary>


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -3,6 +3,7 @@
 using System.Collections.Generic;
 using Unity.Collections;
 using System;
+using System.Linq;
```
FrancescoC-unity, line 6: Is this leftover, I might be wrong, but I don't see any LINQ usage here?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
```diff
@@ -57,19 +57,18 @@ static protected int RenderersToVolumes(ref Renderer[] renderers, ref ProbeRefer
...
+        static protected int NavPathsToVolumes(ref ProbeReferenceVolume.Volume cellVolume, ref List<ProbeReferenceVolume.Volume> volumes, ref Dictionary<Scene, int> sceneRefs)
         {
             // TODO
             return 0;
         }
 
-        static protected int ImportanceVolumesToVolumes(ref ProbeReferenceVolume.Volume cellVolume, ref List<ProbeReferenceVolume.Volume> volumes, ref Dictionary<Scene, int> sceneRef)
+        static protected int ImportanceVolumesToVolumes(ref ProbeReferenceVolume.Volume cellVolume, ref List<ProbeReferenceVolume.Volume> volumes, ref Dictionary<Scene, int> sceneRefs)
         {
-            // TODO
```
FrancescoC-unity, line 68: Why removing the todo?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
alelievr: Line 468, `uniquePositions.Keys.ToArray();` requires linq

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Skin.cs
```diff
@@ -12,6 +12,7 @@ internal static class Styles
 
             internal static readonly GUIContent s_Size = new GUIContent("Size", "Modify the size of this Probe Volume. This is independent of the Transform's Scale.");
             internal static readonly GUIContent s_DebugColorLabel = new GUIContent("Debug Color", "This color is used to visualize per-pixel probe volume weights in the render pipeline debugger.");
+            internal static readonly GUIContent s_MinMaxSubdivSlider = new GUIContent("Subdivision Controller", "TODO");
```
FrancescoC-unity, line 15: Can we give a tooltip or are we waiting for artists feedback?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
```diff
@@ -11,19 +11,19 @@ namespace UnityEngine.Rendering
     internal class ProbeBrickIndex
     {
         // a few constants
-        internal const int kMaxSubdivisionLevels = 15; // 4 bits
+        internal const int kMaxSubdivisionLevels = 7; // 4 bits
```
FrancescoC-unity, line 14: This can now be 3 bits, so at the very least the comment is wrong.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
FrancescoC-unity: 👍

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -768,7 +794,7 @@ private void Rasterize(Volume volume, List<Brick> outBricks)
 
             // Calculate smallest brick size capable of covering shortest AABB dimension
             float minVolumeSize = Mathf.Min(AABB.size.x, Mathf.Min(AABB.size.y, AABB.size.z));
-            int brickSubDivLevel = Mathf.Min(Mathf.CeilToInt(Mathf.Log(minVolumeSize, 3)), m_MaxSubdivision);
+            int brickSubDivLevel = Mathf.Min(Mathf.CeilToInt(Mathf.Log(minVolumeSize, 3)), m_MaxSubdivision) + 1;
```
FrancescoC-unity, line 797: What was the impact of this?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -11,29 +11,55 @@ namespace UnityEngine.Rendering
...
         /// </summary>
-        [Min(1)]
-        public int cellSize = 64;
+        public CellSize cellSizeInBricks = CellSize.CellSize81;
+
         /// <summary>
         /// The size of a Brick.
         /// </summary>
-        [Min(0.3333f)]
+        [Min(0.01f)]
```
FrancescoC-unity, line 41: Ouch that's low :D

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -11,29 +11,55 @@ namespace UnityEngine.Rendering
...
+
+        public int maxSubdivision => Mathf.CeilToInt(Mathf.Log((float)cellSize / brickSize, 3));
+        public float brickSize => Mathf.Max(0.01f, minDistanceBetweenProbes * 3.0f);
+        public int cellSize => Mathf.CeilToInt((float)cellSizeInBricks * brickSize);
+
+        void OnEnable()
+        {
+            if (version != CoreUtils.GetLastEnumValue<Version>())
+            {
+                // Migration code
```
FrancescoC-unity, line 60: Planned TODO or is something forgot?

alelievr: It's in case we need to migrate some stuff one day (and I'm pretty sure it will happen :) )

FrancescoC-unity: I mean we already have to with the cell size stuff, so probably should add something here if  we can.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
alelievr: I think it's an error, I'll revert it

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
alelievr: Yep, though I'm not sure how is this comment useful 😅

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Skin.cs
alelievr: I'll add one, but it'll probably change since artists didn't validated this approach

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
alelievr: We can't, the problem is that there was no version field before, so it's not possible to detect that the asset was at a previous version, we had this issue several times already in HDRP.
So for this PR it will discard the value that was in cell size (breaking change) but it's fine since the feature isn't released yet.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
FrancescoC-unity: It is relevant because the subdivision level is packed in the index, so we need to know how many bits are used for that

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -11,29 +11,55 @@ namespace UnityEngine.Rendering
...
+            [InspectorName("243")]
+            CellSize243 = 243,
+        }
 
         /// <summary>
         /// The size of a Cell.
         /// </summary>
-        [Min(1)]
-        public int cellSize = 64;
+        public CellSize cellSizeInBricks = CellSize.CellSize81;
```
FrancescoC-unity, line 36: I assume we didn't get any more solid info from artists about this right? We can go ahead with this, but given that I think is fairly sure that we'll change it, I'd add a comment so we don't rely on this being the representation for whatever reason.

FrancescoC-unity: I am wondering already if we should not mention brick here at all (I'd like to keep that as hidden as possible from user), can we replace this with "cell size in probes" count for user facing stuff?

alelievr: Yeah, artists haven't decided yet, though I think they will decide to do something with a distance instead of this enum.
I'll add a comment, I can also change this to be a cell size in probes but that will be changed again when artists decide what they want

FrancescoC-unity: Yeah let's sync on those after artists have their idea done, so I guess just a comment will do for now.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoring.cs
```diff
@@ -113,6 +113,12 @@ public enum ProbeShadingMode
         [SerializeField]
         private bool m_GreedyDilation = false;
 
+        Dictionary<ProbeReferenceVolume.Cell, MeshGizmo> brickGizmos = new Dictionary<ProbeReferenceVolume.Cell, MeshGizmo>();
+        MeshGizmo cellGizmo;
+
+        // In some cases Unity will magically popuplate this private field with a correct value even though it should not be serialized.
+        // The [NonSerialized] attribute allows to force the asset to be null in case a domain reload happens.
+        [System.NonSerialized]
```
JulienIgnace-Unity, line 121: I think we landed a fix for this a couple weeks ago (basically just returning the field to null in OnDestroy iirc).
Might be worth keeping this anyway for safety I suppose.

JulienIgnace-Unity (approved):

FrancescoC-unity (approved):

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoring.cs
alelievr: Do you have the link of the fix? This issue has been here for years that would be really cool if it was fixed :D

JulienIgnace-Unity: I meant, a local fix in the APV code :)

JulienIgnace-Unity: https://github.com/Unity-Technologies/Graphics/pull/4035

alelievr: Ah okay

sebastienlagarde (approved):

PR #4049, opened 2021-03-30
Fixed an unhelpful error message when custom function nodes don't have a valid file

joshua-davis:
Fix for https://fogbugz.unity3d.com/f/cases/1323493/

### Checklist for PR maker

---
### Purpose of this PR
When a custom function node in a shader graph had a guid to a file that didn't exist it displayed an incomprehensible error. This fix clarifies that error by saying the file doesn't exist.

---
### Testing status
- [x] Tested the attached shader graph in the bug (file guid set but to an invalid file).

| Old | New |
| image | image |

Validated old behavior as well:
- [x] Valid file has no errors
- [x] File guid is valid, but the extension is wrong displays an error (same behavior as before):

| Old | New |
| image | image |

- [x] Default custom function node (File guid is empty, no outputs). No error but a warning about no output slot. This isn't an error because it's not used in the graph so the node gets removed. This preserves old behavior.

| Old | New |
| image | image |

- [x] File guid is empty and there is an output slot. This is the same as the bug and used to display an incomprehensible message. New error: 

| Old | New |
| image | image |

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Nodes/Utility/CustomFunctionNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Utility/CustomFunctionNode.cs
@@ -400,7 +400,7 @@ namespace UnityEditor.ShaderGraph
                 if (!string.IsNullOrEmpty(functionSource))
                 {
                     string path = AssetDatabase.GUIDToAssetPath(functionSource);
-                    if (!string.IsNullOrEmpty(path))
+                    if (!string.IsNullOrEmpty(path) && AssetDatabase.LoadAssetAtPath<UnityEngine.Object>(path) != null)
                     {
                         string extension = path.Substring(path.LastIndexOf('.'));
                         if (!s_ValidExtensions.Contains(extension))
--- a/com.unity.shadergraph/Editor/Data/Nodes/Utility/CustomFunctionNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Utility/CustomFunctionNode.cs
@@ -51,8 +51,16 @@ namespace UnityEditor.ShaderGraph
             }
         }
 
+        enum SourceFileStatus
+        {
+            Empty,        // No File specified
+            DoesNotExist, // Either file doesn't exist (empty name) or guid points to a non-existant file
+            Invalid,      // File exists but isn't of a valid type (such as wrong extension)
+            Valid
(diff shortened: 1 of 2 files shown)


com.unity.shadergraph/Editor/Data/Nodes/Utility/CustomFunctionNode.cs
```diff
@@ -385,26 +393,33 @@ void ValidateBareTextureSlots()
...
-                        {
-                            owner.AddValidationError(objectId, k_InvalidFileType, ShaderCompilerMessageSeverity.Error);
-                        }
+                            fileStatus = SourceFileStatus.Invalid;
                         else
-                        {
-                            owner.ClearErrorsForNode(this);
-                        }
+                            fileStatus = SourceFileStatus.Valid;
                     }
```
jessebarker, line 410: Any reason to remove the curly braces here?  (I'm not asking to put them back, but have a mild preference for them)

jessebarker (approved):
LGTM

com.unity.shadergraph/Editor/Data/Nodes/Utility/CustomFunctionNode.cs
joshua-davis: No reason, it's just how it ended up after refactoring the logic a few times

ghost (changes requested):
CFN searches for the "half" variant of the function when the node is set to "inherit" and the graph is set to "single," you can see through the blue bar color that the node is "single" precision but it's still searching for the half precision function
image

ghost (changes requested):
Renaming the source file and deleting its meta file and reloading the graph does not seem to give the intended error message for a missing file 
image

ghost (commented):
Other than the two problems I listed, I tested every state and everything else seems to work fine, additionally I made sure that having a valid file doesn't accidentally crush any other errors (errors from nodes plugged into inputs, errors resulting in mismatched number of inputs/outputs from the function definition, and syntax errors in the used function)

joshua-davis:
> Renaming the source file and deleting its meta file and reloading the graph does not seem to give the intended error message for a missing file
> image

Added a fix for this

ghost (approved):
All problems I found were either not caused by this PR or fixed.

PR #4050, opened 2021-03-30
[Fogbugz 1325700] Fixing volumetric screen zmask & depth pyramid for hw drs

kecho:
### Purpose of this PR
Fogbugz: https://fogbugz.unity3d.com/f/cases/1325700/
Fixing issues with Hardware DRS:
* ZMask textures for volumetric lighting were not tagged for hardware DRS and they should, since they are masks of the current mip depth buffer.  This causes Volumetric lighting to be missing / wrong due to the occlusion computations in VolumetricLighting.compute sample the wrong depth.
* The Depth Pyramid, which is tagged to be hardware scaled, is not considering that it could hold mip sizes scaled by the DRS system. This causes the depth pyramid pass to fail mip creation (since its using the wrong offsets / sizes).  An alternative fix is to not tag the depth pyramid buffer as DRS but this would cause an extra cost in memory.

### Testing status
Tested locally on HDRP tempalte scene, inside the with the plan in the glass. Detailed instructions entered in the fogbugz.
* Added a fog volume and a very intesne light. Toggled DRS on / off
* Checked the contact shadow buffer in the analysis -> HDRP debugger and observed that occlusion is there after my changes.
* Doing full CI run for regression.

### Comment for QA or Reviewer
* Have not tested these changes on xbox 1. In fact, not sure DRS hardware mode has been tested on xb1 recently (and I lack of one)
* DRS hardware for PS4 is very broken (in many other ways) there is a separate bug tracking this. So no need to test ps4 yet

ChangesForDRSDPForGithub

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
@@ -840,6 +840,8 @@ namespace UnityEngine.Rendering.HighDefinition
 
                 using (var builder = renderGraph.AddRenderPass<GenerateMaxZMaskPassData>("Generate Max Z Mask for Volumetric", out var passData))
                 {
+                    //TODO: move the entire vbuffer to hardware DRS mode. When Hardware DRS is enabled we will save performance
+                    // on these buffers, however the final vbuffer will be wasting resolution. This requires a bit of more work to optimize.
                     passData.parameters = PrepareGenerateMaxZParameters(hdCamera, depthMipInfo);
                     passData.depthTexture = builder.ReadTexture(depthTexture);
                     passData.maxZ8xBuffer = builder.CreateTransientTexture(new TextureDesc(Vector2.one * 0.125f, true, true)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
```diff
@@ -842,11 +842,11 @@ TextureHandle GenerateMaxZPass(RenderGraph renderGraph, HDCamera hdCamera, Textu
                 {
                     passData.parameters = PrepareGenerateMaxZParameters(hdCamera, depthMipInfo);
                     passData.depthTexture = builder.ReadTexture(depthTexture);
-                    passData.maxZ8xBuffer = builder.CreateTransientTexture(new TextureDesc(Vector2.one * 0.125f, false, true)
+                    passData.maxZ8xBuffer = builder.CreateTransientTexture(new TextureDesc(Vector2.one * 0.125f, true, true)
```
FrancescoC-unity, line 845: So this is a tad trickier than this actually given that the VBuffer is not scaled; this change is fine but it will over estimate and potentially lead to worse performance. 

So for the bug fix sake this is fine, but I'll keep it in mind as we might need to change it a bit.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
```diff
@@ -625,11 +629,13 @@ public void ComputePackedMipChainInfo(Vector2Int viewportSize)
...
                     mipLevelOffsets[mipLevel] = mipBegin;
 
-                    textureSize.x = Math.Max(textureSize.x, mipBegin.x + mipSize.x);
-                    textureSize.y = Math.Max(textureSize.y, mipBegin.y + mipSize.y);
+                    hardwareTextureSize.x = Math.Max(hardwareTextureSize.x, mipBegin.x + mipSize.x);
+                    hardwareTextureSize.y = Math.Max(hardwareTextureSize.y, mipBegin.y + mipSize.y);
                 }
                 while ((mipSize.x > 1) || (mipSize.y > 1));
 
+                textureSize = new Vector2Int((int)((float)hardwareTextureSize.x * textureScale.x), (int)((float)hardwareTextureSize.y * textureScale.y));
```
FrancescoC-unity, line 637: Is this rounding via cast matching what the actual texture size ends up being?  I think it should be fine, but I remember seeing some non-obvious rounding going on in the past hw side, so worth maybe check via renderdoc if you haven't? 

(Worth checking several steps of percentage, so that we check both odd and even texture sizes )

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
sebastienlagarde: @kecho can you add a comment there (TODO) so we don't forget, thanks

TomasKiniulis (changes requested):
Looks good on Desktop, but I do get some issues on Xbox.

Attaching test results:
**No Dynamic Resolution**
noDynamic
**Hardware Dynamic Res** - contact shadows seem to be gone and instead noticeable flickering occurs
https://user-images.githubusercontent.com/50582134/113292198-d81b1700-92fc-11eb-8323-a9c9714637db.mp4

**Master** 
https://user-images.githubusercontent.com/50582134/113292329-07ca1f00-92fd-11eb-9861-dd336048714c.mp4

TomasKiniulis (approved):
Checking with Kleber, the listed issue will likely require C++ change. Opened up a ticket here: https://fogbugz.unity3d.com/f/cases/1326106/. Approving the PR, since it's an improvement from master result.

sebastienlagarde (approved):

PR #4059, opened 2021-03-31
[2021.2][ShaderGraph] Fix PBRMasterGUI not to require _EmissionColor property

jessebarker:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1319637/

A recent change to PBRMasterGUI put a requirement on the _EmissionColor property, which is not added to a Lit ShaderGraph by default.  This makes that conditional again.

---
### Testing status
Manual testing according to the bug.
Yamato results to follow

---
### Comments to reviewers
This should reflect the Slack thread on this issue.


cdxntchou (approved):

Verasl (approved):
This fixes the warning but will also require users to add a property called '_EmissionColor' to be able to access this GI setting, might be good to explain this in the Shader Graph docs.

@KEngelstoft thoughts?

KEngelstoft:
> This fixes the warning but will also require users to add a property called '_EmissionColor' to be able to access this GI setting, might be good to explain this in the Shader Graph docs.
> 
> @KEngelstoft thoughts?

Yes, makes sense to add a note in the docs.

KEngelstoft (approved):
Looks good, please consider adding a note in the docs as discussed in this PR.

PR #4067, opened 2021-04-01
Fixed the 'disconnect all' option on stack block items being grayed out. Fixes fogbugz 1313201

joshua-davis:

### Checklist for PR maker

---
### Purpose of this PR
Bugfix for https://fogbugz.unity3d.com/f/cases/1313201/
The root issue is that the base node's `BuildContextualMenu` always adds a "disconnect all" option. When visiting nodes it looks like all nodes in the stack are visited, starting with the most depth first item. So the block node adds a "disconnect all" option that will pass the `DisconnectAllStatus` check, but the parent stack node will fail this check (since it has no edges) and as both options have the same name, the last one wins.

To solve this I simply disabled the base check on the stack node (our context view) as I've seen on some other nodes.

---
### Testing status
- [x] Verify that 'Disconnect all' is grayed out on a stack block with no connections
- [x] Verify that 'Disconnect all' is not grayed out on a stack block with connections
- [x] Verify that 'Disconnect all' on a stack block works
- [x] Do the same for a non stack block (both connected to a non-stack block and connected to a stack block).

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a bug where synchronously compiling an unencountered shader variant for preview was causing long delays in graph updates [1323744]
 - Fixed a regression where custom function node file-included functions could not access shadergraph properties [1322467]
 - Fixed an unhelpful error message when custom function nodes didn't have a valid file [1323493].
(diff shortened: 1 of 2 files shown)


jessebarker (approved):
LGTM

LandonTownsendUnity (approved):
Tested:
Bug repro case
Nodes have all the correct right click options
All right click options on nodes work as they should
Stack nodes have all the correct right click options
All right click options on stack nodes work as they should

PR #4078, opened 2021-04-06
Hd/update shadow layer editor

RSlysz:
### Purpose of this PR
Fix the display of "link light layer" in the Shadow part of the light. (UX required renaming to **Custom Shadow Layer** and checkbox state inversion ; checked become not linked anymore)
Also, While doing this, I found that the way the Light Layer and the shadow layer sync is a little odd. When clarifying this in the code, I also added better support for multi-edition. When the former "link light layer" had multiple different state, prior we do nothing up to the time the inspector is opened again on a light. Such light was updated. Other not. This is especially strange as we can have modified the Light Layer in General for a bunch of light and the one that are linked don't update there shadow mask on the new selected one. I fixed this behaviour in this PR. The sync is now done when modifying either the **Light Layer** in General or modifying the checkbox **Custom Shadow Layer**. And when this happens we don't discard anymore if there is multiple different state but sync every Lights that have **Custom Shadow Layer** checked in the selection.
image

---
### Testing status
Tested Inspector part: (display is UX requirement)
- Custom Shadow Layer checked: show Layer below
- Custom Shadow Layer unchecked: hide Layer below
- Custom Shadow Layer multiple different value: show Layer below in disabled state
- Undo works on Custom Shadow Layer
- Multiple edition works on Custom Shadow Layer
- Tested with the following setup that changing the layer in Light Layer when Custom Shadow Layer unchecked update the shadows correctly and also, when checked the Layer field update the shadow correctly
image

---
### Comments to reviewers
@JordanL8 there is no image of the component in the documentation, only texts, AFAIK


JulienIgnace-Unity (approved):

RemyUnity:
Did you also test upgrading from previous version ? Maybe I'm wrong but I don't see upgrading code here ?

RSlysz:
There is no data layout change, only the way Editor us it changed so there is nothing to migrate

iM0ve (approved):
Good test coverage. I also wanted to ask about upgrading, but I see its already discussed in the comments.

Looks good to go

PR #4081, opened 2021-04-06
Fix for NaN that can happen on area lights at a certain angle. 

FrancescoC-unity:
If V1oV2 was exactly -1, rsqrt would end up being rsqrt(0) which will result in a NaN, at least on console. 

This fixes the issue. 

Fix for https://fogbugz.unity3d.com/f/cases/1322629/

For QA:  I tested the repro case, and now it is not black anymore.


com.unity.render-pipelines.core/ShaderLibrary/AreaLighting.hlsl
```diff
@@ -21,7 +21,8 @@ real3 ComputeEdgeFactor(real3 V1, real3 V2)
     if (V1oV2 < 0)
     {
         // Undo range reduction.
-        y = PI * rsqrt(saturate(1 - V1oV2 * V1oV2)) - y;
+        float f = saturate(1 - V1oV2 * V1oV2);
```
sebastienlagarde, line 24: should we do something like y = PI * rsqrt(min(Epsilon, saturate(1 - V1oV2 * V1oV2))) - y;
instead?

FrancescoC-unity: We could, though that will provide an exploding value there at a specific viewing edge that I am afraid will introduce problems in bloom, however I guess we can select a decent epsilon so that doesn't happen. 

I'll do that

RemyUnity (commented):
LGTM if the bug steps don't repro anymore.
Beeing a pretty specific issues, would a new test scene, or a test scene modification be worth it ?

RemyUnity (approved):

FrancescoC-unity:
> LGTM if the bug steps don't repro anymore.
> Beeing a pretty specific issues, would a new test scene, or a test scene modification be worth it ?

Unfortunately this happens only on Xbox and with an _extremely_ specific combination of view and light angle, so probably not worth adding modifications (it would probably hinder the runtime tests and non runtime tests don't run on consoles)

sebastienlagarde:
@FrancescoC-unity reverted currently due to multiple yamato error

PR #4082, opened 2021-04-06
[HDRP][Ray Tracing] Meshes added to the accel struct now default to dbl-sided when they mix single and dbl-sided materials

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1323451

Before:
When adding meshes to the ray-tracing acceleration structure, meshes with a mix of single and double-sided materials default to single-sided. This can lead to missing parts, or incorrect refractive behaviour (see the material samples scene for examples of this, on the glass and the alpha-clipping widgets).

Now:
When adding meshes to the ray-tracing acceleration structure, meshes with a mix of single and double-sided materials default to double-sided. This fixes the cases mentioned right above.

To test: HDRP material samples scene. Speaking of, in the end it'd be more correct to change the Aluminium material of this scene to "double-sided" "flip", it would make more sense for the glass and transparent spheres.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Getting-Started.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Getting-Started.md
@@ -1,8 +1,6 @@
 # Getting started with ray tracing
 
-The High Definition Render Pipeline (HDRP) includes preview ray tracing support from Unity 2019.3. Ray tracing is a feature that allows you to access data that is not on screen. For example, you can use it to request position data, normal data, or lighting data, and then use this data to compute quantities that are hard to approximate using classic rasterization techniques.
-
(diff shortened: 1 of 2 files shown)


sebastienlagarde (approved):

remi-chapelain:
Haven't tested yet, but it could be great to add this information in the documentation in the meantime :) 
@Vic-Cooper @JordanL8

remi-chapelain (changes requested):
Waiting on the documentation change because I think it's important to have this behavior noted somewhere. Especially the fact that in a multi mat scenario, we can't have both, single sided for one and double sided for the other. 

Other than that, I tested the behavior with other multi mat meshes, raytraced shadow, everything is working as expected. ✔️

sebastienlagarde:
@eturquin can you do a follow up on this PR? thanks

Vic-Cooper:
Docs updated.

PR #4083, opened 2021-04-06
HD/Decals: allows negative UV

RSlysz:
### Purpose of this PR
Fix Decal issue with UV Edit mode when tilling is negative ( https://fogbugz.unity3d.com/f/cases/1322552/ )

Note: At first I felt it strange to have the UV be not on the projector gizmo when negative but it makes perfect sens when attempting to manipulate the handles. In order for it to be negative, it must cross the opposite edge

---
### Testing status
Manually tested bugs step to reproduce and manipulation on both axes.
WpeyzP66fC

---
### Comments to reviewers
@AndrewSaraevUnity for awareness


sebastienlagarde (approved):

RemyUnity (approved):
Thanks for description and video, LGTM 👍

PR #4085, opened 2021-04-06
Fixed camera preview with multi selection

adrien-de-tocqueville:
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1324126/

Issue what that the same camera was used for each preview, so the previous projection matrix and similar variables were wrong

---
### Testing status
tested the repro steps


sebastienlagarde (approved):

iM0ve (approved):
I reproduced the issue in a more complex project in an attempt to find unexpected issues. No new issues found and the PR resolves the case. Also did various small tests for example checked that Motion blur still works on Preview and Game cameras.


https://user-images.githubusercontent.com/36502659/114388606-957af980-9b9c-11eb-850b-72331168fa92.mp4

PR #4087, opened 2021-04-06
[HDRP] Use a separate set of history buffers for every AOV request

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1323684/
Previously all AOVs were sharing the same history buffers with the main camera. This was creating problems, because each AOV might render the scene with completely different settings (different set of lights, material overrides, etc).

---
### Testing status
I have tested by recording the beauty pass with TAA enabled  and at the same time outputing the Normal AOV (as described in the bug ticket)

**Before:**
beauty_0040_buggy

**After:**
beauty_0040_fixed


sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVRequestData.cs
```diff
@@ -59,6 +59,7 @@ public struct AOVRequestData
         private readonly AOVRequestBufferAllocator m_BufferAllocator;
         private readonly AOVRequestCustomPassBufferAllocator m_CustomPassBufferAllocator;
         private List<GameObject> m_LightFilter;
+        private BufferedRTHandleSystem m_HistoryRTSystem; // Each AOV render request gets a separate set of history buffers (to avoid mixing history data between different AOVs)
```
JulienIgnace-Unity, line 62: I think you need to manually call dispose on the system like we do in HDCamera otherwise it can leak render textures.

pmavridis: Yes you are right, I'm looking now how to properly do this (I will probably move the AOV m_HistoryRTSystem to a static dictionary).

sebastienlagarde:
@QA: this PR have been validated by Jean Blouin with recorder.

RemyUnity (approved):
LGTM. Fix confirmed by the bug reporter and clear PR description, thank you.

PR #1799, opened 2020-09-09
Universal/debug views/merged

andrewt-unity:
Relevant UR Jira: https://unity3d.atlassian.net/browse/UR-2019

# **Please read;**
## **PR Workflow for the Graphics repository:**
* **All PRs must be opened as draft initially**
* Reviewers can be added while the PR is still in draft
* The PR can be marked as “Ready for Review” once the reviewers have confirmed that **no more changes are needed**
* Tests will start automatically after the PR is marked as “Ready for Review”
* **Do not use [skip ci]** - this can break some of our tooling
* Read the [Graphics repository & Yamato FAQ](http://go/graphics-yamato-faq).

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [ ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ]  Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Port the work done at Hackweek 2019 to implement debug-views on URP using the same system as those already implemented in HDRP.

---
### Testing status

**Manual Tests**: What did you do?
- [x] Opened test project + Run graphic tests locally
- [x] Built a player
- [x] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in developer mode, you have a button at end of resources that check the paths)
- Other: 

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics

Any test projects to go with this to help reviewers?

---
### Comments to reviewers
[...]

PR #4089, opened 2021-04-06
URP Debug views polish

arttu-peltonen:
### Checklist for PR maker
  - _New feature for 2021.2, no backports_
  - _API docs added, separate user documentation to be added later_
  - _No graphics test needed_
---
### Purpose of this PR

This is a follow-up PR to the earlier [initial URP Debug Views implementation](https://github.com/Unity-Technologies/Graphics/pull/1799), mostly focusing on UI improvements, general polish and bugfixes.

### Visual comparison

The UI looked like this before:
3rendering-old

With this PR:
0common
1material
2lighting
3rendering

#### UI / General
- Added Common tab for easy access to most common debug modes.
- Everything arranged under collapsible foldouts, added window + foldout context menus
- Removed Validation tab, rearranged its contents under Rendering and Lighting
- General UI margin & layout polish (both editor & runtime) according to UX design
  - While this is mostly affecting URP/RP Core, some UI changes will slightly affect HDRP too. 
- Implement conditional UI to reduce unnecessary options from UI unless specific mode is selected.
  - e.g. Pixel Validation Mode, Material Validation Mode
- Fix mouse & touch input for runtime UI, proper InputSystem handling
#### New debug view functionality
- Added channel mask to "Outside Range" Pixel Validation Mode:  
  channelmask
- Map Size slider for Map Overlays
  - Depth buffer overlay mode also now respects specified size, not always full-screen  
 mapsize
- Prettier number-tile display (now using HDRP implementation), currently only used for LightingComplexity mode:
Old:
lighting_complexity_old
New:
lighting_complexity_new_2

#### Removed functionality
- Disabled mipmap debug mode from UI. Shader Graph doesn't support it, so a decision was made to drop it for now.
- Removed LOD DebugMode, currently the info required by this is not available in the shaders.
#### Bugfixes
- Fix logged errors on assembly reload
- Various runtime UI bugfixes
- Fix broken DebugLightingFeatureFlagsMask.Everything
- Fix pixel validation mode not working in game view
- Fixed debug UI outside mobile safe area
- Fixed XR single-pass 

---
### Testing status
- Debug views tested manually using BoatAttack
- Tested both Forward and Deferred
- Tested on mobile with Unity Remote (actual mobile builds not made)
- Checked that Rendering Debugger UI didn't break for HDRP
- ~**Not tested** on XR~ Tested XR with MockHMD in Editor Playmode. Both single-pass and multi-pass should work.

---


com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
```diff
@@ -0,0 +1,379 @@
...
+                    widget.isHidden = true;
+                }
+                else if (mode == DebugMaterialValidationMode.Metallic && widget.displayName == k_AlbedoSettingsContainerName)
+                {
+                    widget.isHidden = true;
+                }
+                else
+                {
+                    widget.isHidden = false;
+                }
```
alex-vazquez, line 205: Can we find a way to extract the mode conditions outside the loop?

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsRendering.cs
```diff
@@ -1,41 +1,286 @@
...
+                debugSceneOverrideMode = DebugSceneOverrideMode.ShadedWireframe;
+            }
+            else if (overdraw)
+            {
+                debugSceneOverrideMode = DebugSceneOverrideMode.Overdraw;
+            }
+            else
+            {
+                debugSceneOverrideMode = DebugSceneOverrideMode.None;
+            }
```
alex-vazquez, line 63: I think it would be more clear to have a `switch`

alex-vazquez (commented):
Good work, some minor comments :)

com.unity.render-pipelines.core/Editor/Debugging/DebugUIDrawer.cs
```diff
@@ -90,18 +90,29 @@ protected void Apply(DebugUI.IValueField widget, DebugState state, object value)
...
+        protected Rect PrepareControlRect(float height = -1, bool fullWidth = false)
         {
             if (height < 0)
                 height = EditorGUIUtility.singleLineHeight;
             var rect = GUILayoutUtility.GetRect(1f, 1f, height, height);
-            rect.width -= 2f;
-            rect.xMin += 2f;
-            EditorGUIUtility.labelWidth = rect.width / 2f;
+
+            rect.width -= 4f;
```
eh-unity, line 104: Add a named constant for these and maybe even a "shrink" method or commetn?

com.unity.render-pipelines.core/Editor/Debugging/DebugWindow.cs
```diff
@@ -423,18 +433,16 @@ void OnGUI()
                 Rect splitterRect = new Rect(splitterPos - 3, 0, 6, Screen.height);
                 GUI.Box(splitterRect, "", s_SplitterLeft);
 
-                GUILayout.Space(10f);
+                GUILayout.Space(2f);
```
eh-unity, line 436: Maybe these indentation constants should be a group of named constants, a sort of "theme" maybe.

eh-unity (approved):
LGTM.

com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed normal bias field of reference volume being wrong until the profile UI was displayed.
 - Fixed L2 for Probe Volumes.
 - Fixed assertion on compression of L1 coefficients for Probe Volume.
+- Fixed missing increment/decrement controls from DebugUIIntField & DebugUIUIntField widget prefabs.
```
phi-lira, line 22: Was this issue something that is reproduced in previous versions or only in current version?

com.unity.render-pipelines.core/Runtime/Debugging/Prefabs/Resources/DebugUICanvas.prefab
```diff
@@ -127,7 +127,7 @@ MonoBehaviour:
     prefab: {fileID: 224720214277421396, guid: f22bcc84a5f4a1944b075a2c4ac71493, type: 3}
   - type: UnityEngine.Rendering.DebugUI+FloatField, Unity.RenderPipelines.Core.Runtime,
       Version=0.0.0.0, Culture=neutral, PublicKeyToken=null
-    prefab: {fileID: 224720214277421396, guid: 7d4fd3415ea7dd64bbcfe13fb48a730b, type: 3}
+    prefab: {fileID: 224720214277421396, guid: d8c744701b43c864b88e7f8144e19bc5, type: 3}
```
phi-lira, line 130: Is this intentional change?

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -109,11 +109,7 @@ bool StripUnusedPass(ShaderFeatures features, ShaderSnippetData snippetData)
 
         bool StripUnusedFeatures(ShaderFeatures features, Shader shader, ShaderSnippetData snippetData, ShaderCompilerData compilerData)
         {
-#if URP_ENABLE_DEBUG_DISPLAY
             bool stripDebugDisplayShaders = !Debug.isDebugBuild;
```
phi-lira, line 112: We need a follow up PR that introduces option in URP to force strip / include debug in build.

phi-lira: Maybe something for when global settings land?

com.unity.render-pipelines.universal/Runtime/Passes/GBufferPass.cs
```diff
@@ -106,7 +106,32 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+                        else
+                        {
+                            context.DrawRenderers(renderingData.cullResults, ref debugDrawingSettings, ref m_FilteringSettings, universalMaterialTypeTag, false, tagValues, stateBlocks);
+                        }
+                    }
+                }
+                else
+                {
+                    context.DrawRenderers(renderingData.cullResults, ref drawingSettings, ref m_FilteringSettings, universalMaterialTypeTag, false, tagValues, stateBlocks);
+                }
```
phi-lira, line 133: This code is repeated in some places, also users will have to write this if they want to write custom pass that supports debug mode. Maybe we should add helper function to encapsulate this logic?

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -435,11 +435,7 @@ int GetDestination()
                     m_Materials.uber.EnableKeyword(ShaderKeywordStrings.UseFastSRGBLinearConversion);
                 }
 
-                if (isSceneViewCamera)
-                {
-                    // Scene-view doesn't necessarily use the final-blit, so ensure we set the debug properties here (if necessary)...
-                    DebugHandler?.UpdateShaderGlobalPropertiesFinalBlitPass(cmd, ref cameraData);
-                }
+                GetActiveDebugHandler(renderingData)?.UpdateShaderGlobalPropertiesForFinalValidationPass(cmd, ref cameraData, !m_HasFinalPass);
```
phi-lira, line 438: Why is the scene view camera check not needed anymore?

com.unity.render-pipelines.universal/ShaderLibrary/Debug/DebugViewEnums.cs
```diff
@@ -3,110 +3,232 @@
 namespace UnityEngine.Rendering.Universal
 {
     #region Material Settings
+
+    /// <summary>
+    /// Debug material modes.
+    /// </summary>
     [GenerateHLSL]
     public enum DebugMaterialMode
```
phi-lira, line 11: If feels weird that the Debug API is internal but we expose this as public. Is there any usage for user for this to be public without debug API?

com.unity.render-pipelines.universal/ShaderLibrary/Debug/DebugViewEnums.cs
```diff
@@ -3,110 +3,232 @@
...
     {
+        /// <summary>No mipmap debug.</summary>
         None,
+        /// <summary>Display the mipmap level sampled.</summary>
         Level,
+        /// <summary>Display the amount of mip levels available.</summary>
         Count,
+        /// <summary>Display the mip ratio.</summary>
+        Ratio
     }
```
phi-lira, line 130: Should we expose this since we don't support Mip level debug?

phi-lira (approved):
LGTM. Minor questions.

phi-lira (commented):
DebugMaterialSettings.cs doesn't have a deleted .meta file?

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
arttu-peltonen: Yes, that's my plan - to move the "debug shaders enabled" option from HDRP asset into HDRP global asset, and at the same time, introduce the same option in URP global asset. 

The question is, is this okay until then? I don't believe I can wait for URP global settings to land and still make the change before cutoff. Do you think this type of change can happen after cutoff?

com.unity.render-pipelines.core/Runtime/Debugging/DebugUI.Fields.cs
```diff
@@ -241,6 +241,39 @@ public override float ValidateValue(float value)
...
+                //enumValues = Enum.GetValues(value).Cast<int>().ToArray();
+
+                var values = Enum.GetValues(enumType);
+                var enumValues = new int[values.Length];
+                for (int i = 0; i < values.Length; i++)
+                    enumValues[i] = (int)values.GetValue(i);
+
+                return enumValues;
+            }
+        }
```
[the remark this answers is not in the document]
arttu-peltonen: I think this code is pretty specific to debug views. I'm skipping `ObsoleteAttribute` because debug modes are never serialized, so there's no need to mark anything obsolete - it will just be removed. For general-purpose inspector enums, there is already `EnumDataUtility`.

com.unity.render-pipelines.core/Runtime/Debugging/DebugUI.Fields.cs
```diff
@@ -328,8 +345,7 @@ internal void InitQuickSeparators()
 
             internal void InitIndexes()
             {
-                if (enumNames == null)
-                    enumNames = new GUIContent[0];
+                enumNames ??= new GUIContent[0];
```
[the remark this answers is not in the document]
arttu-peltonen: Actually I had to switch to the beta branch of the formatter to get this in properly :(

com.unity.render-pipelines.core/Runtime/Debugging/Prefabs/Scripts/DebugUIHandlerCanvas.cs
```diff
@@ -144,7 +161,7 @@ void Traverse(DebugUI.IContainer container, Transform parentTransform, DebugUIHa
 
                 var childContainer = go.GetComponent<DebugUIHandlerContainer>();
                 if (childContainer != null && child is DebugUI.IContainer)
-                    Traverse(child as DebugUI.IContainer, childContainer.contentHolder, uiHandler);
+                    Traverse(child as DebugUI.IContainer, childContainer.contentHolder, uiHandler, ref selectedHandler);
```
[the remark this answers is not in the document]
arttu-peltonen: Oh wow, cool syntax, thanks!

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
arttu-peltonen: I'm not sure if that's possible here, as each condition also depends on the loop variable `widget`.

com.unity.render-pipelines.core/CHANGELOG.md
arttu-peltonen: This was broken in the branch I started on and I assumed it was broken, but you're right - I should verify this was broken in a shipped version as well, and only include this in release notes if that's the case.

com.unity.render-pipelines.core/Editor/Debugging/DebugWindow.cs
arttu-peltonen: I added a name for these so they are easier to understand and change. For a broader theme, I think it might just end up looking like a garbled mess of arcane names given to every integer between 0 and 20 :) after all, to be consistent it should be done for the numbers 3 and 6 visible above as well, but many of these numbers just come down to trying to match the UI design so it looks right...

com.unity.render-pipelines.core/Runtime/Debugging/Prefabs/Resources/DebugUICanvas.prefab
arttu-peltonen: Yes and no. I think what happened is I made a local copy under Assets, of a prefab that existed under under Packages/RP Core. When I reopened the project, Unity found duplicate GUIDs and assigned a new GUID for the prefab under Packages. I accidentally submitted that change, which lead to some other prefabs having their prefab links broken, so this line is fixing that prefab link.

Do you think I should try and revert back to the original GUIDs?

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
arttu-peltonen: This fixes an issue where DEBUG_DISPLAY keyword was enabled for parts of the pipeline that it shouldn't be. There are multiple passes that can potentially be the "actual last pass rendered". We need to hook the pixel debug validation logic into exactly this last pass (enabling DEBUG_DISPLAY keyword), and make sure this keyword is disabled from all other "potentially final passes", to avoid having the logic active in multiple passes.

Now all of these potentially last passes (FinalBlitPass, UberPost, "Final Pass") call this function, which can either enable or disable the keyword based on whether the pass is actually final. The previous check was trying to do the same thing, but it didn't work right.

com.unity.render-pipelines.universal/ShaderLibrary/Debug/DebugViewEnums.cs
arttu-peltonen: The mode was not removed completely, just the UI was disabled. I can remove this and all related shader code.

com.unity.render-pipelines.universal/ShaderLibrary/Debug/DebugViewEnums.cs
arttu-peltonen: I don't think there's ways for users to actually use these. These are also public in [the equivalent HDRP file](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.high-definition/Runtime/Debug/LightingDebug.cs), so I thought I'd document them rather than make them internal. I guess it would be useful to generate documentation from this file and link to it from the feature docs, but I can also mark them internal if preferred.

arttu-peltonen:
> DebugMaterialSettings.cs doesn't have a deleted .meta file?

Looks like git didn't realize I renamed DebugMaterialSettings.cs -> DebugDisplaySettingsMaterial.cs, but _did_ realize it for its meta file:
> ...ntime/Debug/DebugMaterialSettings.cs.meta → ...ebug/DebugDisplaySettingsMaterial.cs.meta

com.unity.render-pipelines.core/CHANGELOG.md
arttu-peltonen: Checked and this was missing from HDRP in 2020.2.3, so I'd say let's keep it. It wasn't a problem before because a) there's only a few int/uint fields to begin with, and b) there was no mouse support and you could still increment/decrement with keyboard left/right.

com.unity.render-pipelines.core/Runtime/Debugging/DebugUI.Fields.cs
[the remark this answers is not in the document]
alex-vazquez: what happens if I launch the formatter and I am not in the beta branch?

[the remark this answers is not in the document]
arttu-peltonen: Hmm, in fact I see now Yamato formatting has failed because of this. I've opened a slack thread for advice on what to do:https://unity.slack.com/archives/CU41BKSJF/p1619512249177000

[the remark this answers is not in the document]
arttu-peltonen: I've reverted the use of ??=.

com.unity.render-pipelines.universal/Runtime/2D/Passes/Render2DLightingPass.cs
```diff
@@ -157,19 +157,12 @@ private void Render(ScriptableRenderContext context, CommandBuffer cmd, ref Rend
...
-                    else
+                RenderStateBlock renderStateBlock = new RenderStateBlock();
+                activeDebugHandler.DrawWithDebugRenderState(context, cmd, ref renderingData, ref drawSettings, ref filterSettings, ref renderStateBlock,
+                    (ref RenderingData data, ref DrawingSettings ds, ref FilteringSettings fs, ref RenderStateBlock rsb) =>
                     {
-                        context.DrawRenderers(renderingData.cullResults, ref debugDrawSettings, ref filterSettings);
-                    }
-                }
+                        context.DrawRenderers(data.cullResults, ref ds, ref fs, ref rsb);
+                    });
```
arttu-peltonen, line 165: I think this is nicer and more concise now, but I'm not totally happy with it... there are restrictions how ref paramters can be used with lambdas:
1) seems like ref parameters in the must be named explicitly as ref (so I have to say `ref DrawingSettings ds` instead of just `ds`)
2) more annoyingly, ref parameters cannot be captured from outer scope and used inside the lambda. For this reason, I have to pass renderingData and filterSettings to `DrawWithDebugRenderState` even though they are not modified. If I'm missing how this could be done more cleanly, let me know!

com.unity.render-pipelines.universal/Runtime/Passes/GBufferPass.cs
arttu-peltonen: This code is now removed from GBufferPass (since I've opted to use forward renderer for majority of debug drawing). The point about encapsulating is still valid, and I've now addressed it here: https://github.com/Unity-Technologies/Graphics/pull/4089/commits/a3abbc41d8e699ffc895b5b6fa883bf40cadfc98

DennisDeRykeUnity:
@arttu-peltonen I have held off from testing this because I saw that you were in discussion with Thomas about how to make it work for XR.  After you have committed what you think is needed to make this feature viable for XR, please ping me here.  Also please post known limitations (i.e. does it work for Standalone players or only Playmode, is it intended to work in the headset or only on the monitor, etc.).  Thanks!

arttu-peltonen:
> @arttu-peltonen I have held off from testing this because I saw that you were in discussion with Thomas about how to make it work for XR. After you have committed what you think is needed to make this feature viable for XR, please ping me here. Also please post known limitations (i.e. does it work for Standalone players or only Playmode, is it intended to work in the headset or only on the monitor, etc.). Thanks!

@DennisDeRykeUnity thanks! It's now ready for testing. It should work in Editor Playmode (both single-pass instanced and multi-pass). _Right now_ it is also available in player builds (needs Development Mode), but I expect it's not usable there (mainly due to UI) and will most likely be disabled from standalone for now. I'm not sure if you can use a headset together with Editor Playmode, but if you can, feel free to test that too.

alex-vazquez (approved):
Nice job.

TheoWong-pixel (approved):

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsCommon.cs
```diff
@@ -0,0 +1,97 @@
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace UnityEngine.Rendering.Universal
+{
+    public class DebugDisplaySettingsCommon : IDebugDisplaySettingsData
```
phi-lira, line 6: Is this meant to be public? if so needs public API docs.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -1150,12 +1144,11 @@ static void CheckAndApplyDebugSettings(ref RenderingData renderingData)
...
-                    cameraData.renderScale,
-                    cameraData.isHdrEnabled,
-                    msaaSamples,
-                    true,
-                    cameraData.requiresOpaqueTexture);
+                GraphicsFormat hdrFmt = RenderingUtils.SupportsGraphicsFormat(GraphicsFormat.R16G16B16A16_SFloat, FormatUsage.Linear | FormatUsage.Render) ?
+                    GraphicsFormat.R16G16B16A16_SFloat
+                    : SystemInfo.GetGraphicsFormat(DefaultFormat.HDR);
+                cameraData.cameraTargetDescriptor.graphicsFormat = cameraData.isHdrEnabled ? hdrFmt : SystemInfo.GetGraphicsFormat(DefaultFormat.LDR);
+                cameraData.cameraTargetDescriptor.msaaSamples = msaaSamples;
```
phi-lira, line 1151: Why not use same logic as in https://sourcegraph.com/github.com/Unity-Technologies/Graphics/-/blob/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs#L486

com.unity.render-pipelines.universal/ShaderLibrary/Debug/DebuggingFullscreen.hlsl
```diff
@@ -3,25 +3,21 @@
...
 #if defined(DEBUG_DISPLAY)
 
+int _ValidationChannels;
 float _RangeMinimum;
 float _RangeMaximum;
-int _HighlightOutOfRangeAlpha;
 
-TEXTURE2D(_DebugTexture);        SAMPLER(sampler_DebugTexture);
+TEXTURE2D_X(_DebugTexture);
+TEXTURE2D(_DebugTextureNoStereo);
```
phi-lira, line 11: You don't need to declare two textures. TEXTURE2D_X should work on both stereo and non-stereo. Then when sampling, the _X macro should work as well in both cases.

phi-lira (approved):
LGTM. Minor comments.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
arttu-peltonen: I don't want to create a new descriptor, I just want to override a few key values on the descriptor. The original code which creates a new descriptor breaks XR because in cameraTargetDescriptor has been set up to use a texture array etc. 

However I think your point is valid, and I shouldn't duplicate this logic - I should probably refactor `CreateRenderTextureDescriptor` so that there are specific utility functions that are responsible for determining `graphicsFormat`.

com.unity.render-pipelines.universal/ShaderLibrary/Debug/DebuggingFullscreen.hlsl
arttu-peltonen: Normally that's true, but in this case I need "NoStereo" variant for textures that are Texture2D, even when XR is on. This is the case for shadow maps -  they are not texture arrays even in XR.

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsCommon.cs
arttu-peltonen: Shouldn't be public, removing.

com.unity.render-pipelines.core/Runtime/Debugging/DebugManager.cs
```diff
@@ -243,6 +258,25 @@ void OnPanelDirty(DebugUI.Panel panel)
             onSetDirty();
         }
 
+        /// <summary>
+        /// Request DebugWindow to open the specified panel.
+        /// </summary>
+        /// <param name="index"></param>
```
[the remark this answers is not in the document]
arttu-peltonen: Thanks, added.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
arttu-peltonen: Fixed in this commit https://github.com/Unity-Technologies/Graphics/pull/4089/commits/2a5e271e9bb38bc6bae66ffc33cd451d029ef824

com.unity.render-pipelines.universal/ShaderLibrary/Debug/DebuggingFullscreen.hlsl
phi-lira: Ah I see.

com.unity.render-pipelines.core/Editor/Debugging/DebugWindow.cs
```diff
@@ -136,6 +137,13 @@ static void Init()
             open = true;
         }
 
+        [MenuItem("Window/Analysis/Rendering Debugger", validate = true)]
+        static bool ValidateMenuItem()
+        {
+            bool srpActive = RenderPipelineManager.currentPipeline != null;
+            return srpActive;
```
ellioman, line 144: Do we need the `srpActive` variable? Could it simply return `RenderPipelineManager.currentPipeline != null` ?

com.unity.render-pipelines.core/Editor/Debugging/DebugWindow.cs
```diff
@@ -542,6 +557,11 @@ public class Styles
 
             public Styles()
             {
+                Color textColorDarkSkin = new Color32(210, 210, 210, 255);
+                Color textColorLightSkin = new Color32(102, 102, 102, 255);
+                Color backgroundColorDarkSkin = new Color32(38, 38, 38, 128);
+                Color backgroundColorLightSkin = new Color32(128, 128, 128, 96);
```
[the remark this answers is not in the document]
ellioman: We do have for the background color with a slightly different color. 
You can call it like this, which checks the isProSkin: `CoreEditorStyles.backgroundColor`

But we do not have one for text color. Do we need that? If yes, then I guess it makes so add this to the `CoreEditorStyles` class.

com.unity.render-pipelines.core/Runtime/Debugging/DebugManager.cs
```diff
@@ -243,6 +258,25 @@ void OnPanelDirty(DebugUI.Panel panel)
...
+        }
+
+        internal int? GetRequestedEditorWindowPanelIndex()
+        {
+            int? requestedIndex = m_RequestedPanelIndex;
+            m_RequestedPanelIndex = null;
+            return requestedIndex;
+        }
+
         // TODO: Optimally we should use a query path here instead of a display name
```
ellioman, line 280: Can we remove this TODO and create a task in Jira for this?

com.unity.render-pipelines.core/Runtime/Debugging/Prefabs/Scripts/DebugUIHandlerPanel.cs
```diff
@@ -56,13 +57,26 @@ public void SelectPreviousItem()
...
+        {
+            DebugManager.instance.SetScrollTarget(null); // Release scroll target
+        }
+
+        internal void SetScrollTarget(DebugUIHandlerWidget target)
+        {
+            m_ScrollTarget = target;
+        }
+
         // TODO: Jumps around with foldouts and the likes, fix me
```
ellioman, line 73: Can we remove this TODO and create a Jira task for this?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -478,9 +477,7 @@ static class RenderPassBlock
 
         public ScriptableRenderer(ScriptableRendererData data)
         {
-#if URP_ENABLE_DEBUG_DISPLAY
             DebugHandler = new DebugHandler(data);
```
ellioman, line 480: Is the DebugHandler and resources tied to that, stripped from builds when doing release builds?

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -33,15 +33,9 @@ struct InputData
...
     half4   shadowMask;
-
-    #if defined(LIGHTMAP_ON)
     half2   lightmapUV;
-    #else
     float3  vertexSH;
-    #endif
-    #if defined(_NORMALMAP)
     half3x3 tangentToWorld;
-    #endif
```
ellioman, line 44: Is this change intentional?

ellioman (changes requested):
Looks good. 
Mostly some minor comments but I need clarification about the change in Input.hlsl.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
arttu-peltonen: Yes, @eh-unity and @lukaschod thought that since this is just an intermediate struct, not varying struct, the convention has been to leave out the ifdefs from the struct definition and let compiler optimize out unused stuff.

evelinastaniulyte (changes requested):
The issues I noticed on mobile:
- [Can’t open Debugger UI with new Input System](https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit#bookmark=id.lal1go3djkr0)
-[ Lighting Features - “None” can’t be marked](https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit#bookmark=id.1mp6n78wzbiq)
- [Lighting Features - missing “Everything” checkbox (editor panel has it)](https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit#bookmark=id.bvse0ihez2zi)
- [Scrollbar is glitching when Lighting features unfolded](https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit#bookmark=id.542wfi3h1st5)
- [Unable to scroll down - the view always goes back to top](https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit#bookmark=id.z9ggbq2uhgze)
- [[UX] Placeholder of the  “<” arrow UI is very small. Hard to click the right place](https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit#bookmark=id.c3gdkn2ykjz0)
- [[UX] Hard to close the Debug window & to not click any UI button since the window is large and covers the biggest part of the screen. Maybe need to think about “Hide” button](https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit#bookmark=id.c3gdkn2ykjz0)
Full testing doc: 
https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
arttu-peltonen: This define was a temporary one meant for landing the first PR, and it's now removed. I think this is a mistake - creating DebugHandler is unnecessary in release builds so I'll remove it (it only contains a single shader in terms of resources, but still). Shader stripping happens here: https://github.com/Unity-Technologies/Graphics/blob/a597ff26b25d7a3628d021760c930bb3779f99cb/com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs#L164

com.unity.render-pipelines.core/Runtime/Debugging/Prefabs/Scripts/DebugUIHandlerPanel.cs
arttu-peltonen: Yes still valid. I have a jira here: https://jira.unity3d.com/browse/XPIPELINE-129

com.unity.render-pipelines.core/Runtime/Debugging/DebugManager.cs
arttu-peltonen: I don't know what this is referring to, @JulienIgnace-Unity do you happen to have a jira for this?

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
arttu-peltonen: And just to clarify, these 3 fields that the ifdefs are protecting were all added in the first debug views PR: https://github.com/Unity-Technologies/Graphics/pull/1799/files#diff-cb824df074d9e71d1b765eee08de28a09813837fc84b3aef29c0b7184dcacabe

ellioman: Great! After some discussion and testing the ifdef checks are unnecessary as the variables will be stripped if not used.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
ellioman: Great great great!

com.unity.render-pipelines.core/Runtime/Debugging/Prefabs/Scripts/DebugUIHandlerPanel.cs
ellioman: Okay great 👍   
I would personally prefer that we would remove the TODO since the task is tracked elsewhere but this is a minor thing.

ellioman (approved):

DennisDeRykeUnity:
This is not a huge problem, but entering Play mode resets Render Debug settings.  For example, I set Material Override to Albedo.  Then I entered Play mode and it toggled back to None.  This behaviour is the same regardless of whether the XR Management package is installed or not.

- Unity (trunk a2a2b9196a1f)
- URP (universal/debug-views/uipolish c036da5)

DennisDeRykeUnity:
I tested the following configuration with each of the following Rendering Debugger settings. 

- Unity (trunk a2a2b9196a1f)
- URP (universal/debug-views/uipolish c036da5)
- D3D11
- Quest Link
- Stereo Instancing 
- URP Template scene
- Build Settings > Development Build
- Play mode and Standalone players

Material Override = Albedo
Vertex Attribute = Normal 
Material Validation Mode = Metallic
Lighting Debug Mode = Reflections; Lighting Features = Additional Lights
Additional Wireframe Modes = Shaded Wireframe


In addition to the quirks noted in my previous comment, I saw two bugs: 

1) In Play mode, setting Vertex Attribute = Normal works for the left eye but the right eye shows only the skybox (no geometry); same for Game view and in the headset.  

2) None of these Render Debug settings have any effect on built Standalone XR players.  However, I tried two of these settings in built Standalone non-XR players and they had no effect there either.

arttu-peltonen:
@DennisDeRykeUnity  thanks for the feedback :) 

> This is not a huge problem, but entering Play mode resets Render Debug settings. For example, I set Material Override to Albedo. Then I entered Play mode and it toggled back to None. This behaviour is the same regardless of whether the XR Management package is installed or not.

I think this is known behavior, also present in HDRP. I agree that ideally the debug state should survive entering play mode.
 
> Also, the Material Override which I set while in Play mode persist upon exiting Play mode. This is backwards from other settings in the Editor. Most settings that were set prior to entering Play mode persevere upon entering Play mode; most settings that are set within Play mode revert back to their pre-Play-mode values upon exiting Play mode.
 
This happens I guess because there's a domain reload when entering play mode, but not when exiting play mode. Unlike, say, anything in the inspector, debug modes are not "scene data" and thus I believe they should not revert upon exiting play mode (in the same way as e.g. the "Mute Audio" button state does not revert) - debug modes are more akin to editor tools in that sense. I think the only issue here is the inconsistency between entering and exiting play mode - entering play mode shouldn't reset debug options either.

> Also, the Lighting Debug setting does not persist between Editor sessions. Is this the intended behaviour?

None of the debug settings should persist when you close down editor and restart it. Are there some that do?

erikabar (approved):
Tested Debugger Window on Editor and Runtime. Some issues are remaining that will be fixed in a follow up PR. 
Test doc: https://docs.google.com/document/d/13ncezQcHEKxoGPwLaySLVMtuMOPeUhjmbaxbIwHqNGg/edit?ts=60814397#

DennisDeRykeUnity:
URP Rendering Debugger is useless in Standalone mode with Rift (actually tested via Quest Link).  Using the following configuration I saw the five problems below.

- Unity (trunk a2a2b9196a1f)
- URP (universal/debug-views/uipolish c036da5)
- D3D11
- Quest Link
- Stereo Instancing
- URP Template scene
- Build Settings > Development Build
- Built Standalone player

1) CTRL+Backspace brings up the Render Debug menu in mirror mode (i.e. on the monitor) but not in the HMD. 

2) Setting Material Override to anything other than None makes the scene look as if Albedo was chosen, or makes the scene entirely grey.  Same behavior on monitor and in HMD.  

3) With any Lighting Debug Mode setting, trying to enable Lighting Features > None invokes Development Console with “NullReferenceException: Object reference not set to instance of an object”. 

4) For each individual Lighting Features toggle I cycled through every Lighting Debug Mode setting.  Other than item 3 above, the only noticeable effect was that some combinations disabled post processing.  Same behavior on monitor and in HMD.  

5) The only Additional Wireframe Modes that work are None and Wireframe.

arttu-peltonen:
@evelinastaniulyte thanks! Following issues are now fixed (hopefully - they work in device simulator but fixes are speculative as I don't have a device):

- Can’t open Debugger UI with new Input System
- Scrollbar is glitching when Lighting features unfolded
- Unable to scroll down - the view always goes back to top
- [UX] Placeholder of the “<” arrow UI is very small. Hard to click the right place

Hopefully that covers the most urgent ones, could you re-test?

DennisDeRykeUnity (approved):
Today I tested the latest commit with the config below, touching the Render Debug settings listed below.  After this PR lands I'll probably file Fogbugz cases for the four items below.  Some of these were noted in previous comments but I am summarizing here.  

I am not testing on Quest since Standalone is not supported for XR, per Slack discussion.  

I approve.  This looks like a valuable feature for our cusotmers.  Thanks to all who have made this happen, especially Arttu for making this viable for XR!  


Configuration:
- Unity (trunk a2a2b9196a1f)
- URP (universal/debug-views/uipolish 5015807)
- D3D11
- Quest Link
- Stereo Instancing
- Play mode
- URP Template scene, plus customizations for specific debug modes


Settings Tested:
Vertex Attribute = Normal
Material Override = Albedo
Material Override = Specular
Material Override = Smoothness
Material Override = Ambient Occlusion
Material Override = Emission
Material Override = World Normal Space
Material Override = World Tangent Space
Material Override = Lighting Complexity
Material Override = Metallic
Lighting Features = Everything, Lighting Debug Mode = Shadow Cascades
Lighting Features = Everything, Lighting Debug Mode = Lighting Without Normal Maps
Lighting Features = Everything, Lighting Debug Mode = Lighting With Normal Maps
Lighting Features = Everything, Lighting Debug Mode = Reflections
Lighting Features = Everything, Lighting Debug Mode = Reflections With Smoothness
Additional Wireframe Modes = Wireframe
Additional Wireframe Modes = Solid Wireframe
Additional Wireframe Modes = Shaded Wireframe
Overdraw = Enabled
Map Overlays = Depth
Pixel Validation Mode = Highest Values Outside Range

Issues: 
1) I am not convinved that the Map Overlays for either Shadow Map setting works, for XR or non-XR.  I just get gray viewports for these settings, despite setting both lights to static and baked, toggling off Global illumination, and rebaking.  

2) Entering Play mode resets Render Debug settings.  It would be nicer if these settings persisted upon entering Play mode, especially for consistency because the settings persist upon exiting Play mode.  

3) Setting Vertex Attribute = Normal works for the left eye but the right eye shows only the skybox (no geometry); same for Game view and in the headset.

4) URP Rendering Debugger is useless in Standalone mode for XR.  This was expected per Slack discussion with devs.  At minimum Release notes should indicate lack of support, but it would be better to disable the Render Debug menu altogether for XR Standalone.

arttu-peltonen:
@DennisDeRykeUnity thanks! :) And yeah let's follow-up on the points you raised as bugs after landing, I'm especially curious about 1.

Jonasmortensen (approved):
Looks very good all in all. A few quality of life improvements and bugs (already discussed with author) that can be addressed in later PR.

evelinastaniulyte:
Looks good on mobile side now. The crucial bugs are fixed already.
Some minor bugs still persist, but they already submitted in Fogbugz & Jira and will be fixed later.
Full testing document: 
https://docs.google.com/document/d/1Z3X_bhxl8yQIm4FcbzM2rD1asYwDKca0GIEsjCx5hzQ/edit#

evelinastaniulyte (approved):

PR #4090, opened 2021-04-06
Added an error message when not using "Full Res" Texture Quality in HDRP.

JulienIgnace-Unity:
### Purpose of this PR
Added an error message when not using "Full Res" Texture Quality in HDRP.

The Texture Quality settings found in the Quality Settings is problematic with HDRP. Indeed it reduces the size of ALL textures, not just regular texture assets used for materials. This means that it can break internal textures used by HDRP for rendering (like blue noise textures for example). Until we have a proper way to achieve reducing texture size for groups of textures we can only display an error to the users.

This issue was found investigating https://fogbugz.unity3d.com/f/cases/1324278/

---
### Testing status
Tested with two different quality level, one with Half res, one with Full Res. Switching from one quality to the other will display the message.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added Area Light support for Hair and Fabric master nodes.
 - Added a fallback for the ray traced directional shadow in case of a transmission (case 1307870).
 - Added support for Fabric material in Path Tracing.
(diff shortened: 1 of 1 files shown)


RemyUnity:
Is displaying an error message in the console enough in that case ?
The console can be hidden and the message will only display in the small bottom bar of the editor.
Maybe having a popup will make it more obvious, or even forcing the setting to avoid rendering errors ?

Also, did you try in cross pipeline context ? Does the error also display if you switch to a quality level with URP and half res for example ?

JulienIgnace-Unity:
Found another way to handle that with a small C++ PR.

PR #4091, opened 2021-04-06
[HDRP][Ray Tracing] Only meshes with HDRP materials are added to the acceleration structure

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1322365

Before:
Meshes with non-HDRP materials would be added to the ray tracing (and path tracing) acceleration structure no matter what, resulting in black shapes, as described in the ticket above.

Now:
Meshes with non-HDRP materials are not added to the acceleration structure anymore, so it's as if they don't exist for ray and path tracing.

To test: in addition to confirmation that TextMeshPro are not producing black meshes anymore, it'd be good to double-check that it doesn't introduce any undesirable effects in ray tracing in particular (maybe AO or opaque shadows for instance, where materials are irrelevant). If this is the case, we can separate the ray tracing and path tracing filters.


RemyUnity:
Can you describe what was tested ?

remi-chapelain (approved):
Here's what I checked
- Multi mat if one of the multi mat was not HDRP ✔️ (properly supported) 
- Performance with 100,000 instances of mesh to see if the new string comparison has a substantial impact ✔️ (nothing really measurable)
- Recursive rendering ✔️ , (before you were able to see unsupported materials through transparent recursive rendered objects, ex TMP pro)
- Text Mesh Pro disappearing ✔️ 

As for AO, you indeed lose the RTAO contribution of meshes with "not HDRP" materials but I wouldn't say that's an issue per se, I can't really think of a case where you'd want to have an unsupported material in your AS to generate RTAO, there's better way to do something around those lines with layer masks. 

I also launched DXR test on yamato to be sure we do not need to update a capture here. 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Frt_hdrp_mtls_only/.yamato%252Fall-hdrp_dxr.yml%2523PR_HDRP_DXR_trunk/6222395/job

RemyUnity (changes requested):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
```diff
@@ -114,6 +114,19 @@ internal void ReleaseRayTracingManager()
...
 
+        static bool IsValidRayTracedMaterial(Material currentMaterial)
+        {
+            if (currentMaterial == null || currentMaterial.shader == null)
+                return false;
+
+            // For the time being, we only consider non-decal HDRP materials as valid
+            if (currentMaterial.shader.name.Substring(0, 5) != "HDRP/" ||
+                currentMaterial.shader.name.Substring(5) == "Decal")
+                return false;
```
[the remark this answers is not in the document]
eturquin: We can use the tag indeed, but that's another string comparison basically, I was trying to limit them to a minimum (although won't make a huge difference in the end). But if it's a more robust way to perform the test, we can switch to that.

remi-chapelain (changes requested):
Thanks to @RemyUnity, indeed, this excluded all default HDRP shader graph since they are named Shader Graphs/Lit. It will return the first 5 chars and will be skipped for the AS. 

The graphs in the materials samples have been renamed to keep the convention "HDRP/" so that's why it was working on my tests. 

Since Shader name can be renamed in SG, testing via shader name string comparison is not adequate here.

eturquin:
> The graphs in the materials samples have been renamed to keep the convention "HDRP/" so that's why it was working on my tests.
> 
> Since Shader name can be renamed in SG, testing via shader name string comparison is not adequate here.

Ok I'll change the tests accordingly... it would be worth re-running your scene with 100,000 instances after that.

RemyUnity:
If the cost is the test is a concern, it might be an option to cache it, in something like a dictionary.

remi-chapelain (commented):
Indeed, this version works properly with shader but performs worse. Here's the frame time I got from the perf test framework with 1000 instance of a multi mat mesh with 4 materials

- No test (function IsValidRayTracedMaterial returns true) : 150ms
- Null materials and decal tests : 170ms (before the PR)
- Null materials and substring : 155ms (performs better but don't work with SG)
- Null materials, decals and shader tag : 207ms

Performance wise, that means with this specific scnenario we have a 20%-ish increase in frame compared to before the PR.

remi-chapelain (commented):
I tried the latest changes and this works properly in classic ray tracing scene (I tested small office raytracing the church and apartment) and the performance is not really worse and results is the same. 

However, in some cases, especially when you have multiple instance of the same material in the scene, it starts to become an issue. I tried with a scene with 10k mesh, each with 4 materials (2 SG, 2 Lit). 
What I also tried is, as @RemyUnity suggested, caching the materials instead of re-doing the test for each material and this can make a huge difference. Here's the frame times
- Before the PR : 150ms 
- Latest commit : 190/200ms
- Latest commit with caching : 135ms

I also checked that caching materials do not make a difference in other "classic" scenes. 

image

remi-chapelain (approved):
To recap : 
- TextMeshPro is now out of the AS (as every non HDRP materials)
- The method used now use first the firty string method (because it's faster) then if it fails, it check the shader tag
- Also caching has been implemented which for scenes where a lot of the same material is used on mulitple mesh, saves around 20% time compared to before the PR 🥳 (126ms instead of 155ms before the PR in the 10k lantern scene). So, more tests but less time. 
- The cache is cleared every frame because materials can change between frames though. (although rare)

There's still room for improvement in the future because we discard the cache even thought there's very little chance that a material that was HDRP is not going to be next frame. So we could still detect in a way to clear the cache only when there is material change to save even more cost on these specific use case scenes.

RemyUnity (approved):
Nice and clear thread and reports here :)
It's better than before, so I'd say we are good.

An idea to explore if we want to maybe extend the performance further would be to base the cache on the shader instead of the materials, so it can be re-used between frames, maybe at the cost of the Material.shader accessor ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
```diff
@@ -114,6 +115,32 @@ internal void ReleaseRayTracingManager()
...
 
+        bool IsValidRayTracedMaterial(Material currentMaterial)
+        {
+            if (currentMaterial == null || currentMaterial.shader == null)
+                return false;
+
+            bool isValid;
+
+            // We use a cache, to speed up the case where materials are reused many times
+            int matId = currentMaterial.GetInstanceID();
```
sebastienlagarde, line 126: Hey, so I don't really like the fact that we have a cache that can grow indefinitely (or which are reset every frame). I am agree with @RemyUnity suggestion that it should used shader instead.

What about using this instead:

int shaderId = currentMaterial.shader.GetInstanceID();
if (m_ShaderValidityCache.TryGetValue(shaderId , out isValid))
                return isValid;
                
if (currentMaterial..FindPass("IndirectDXR", false)
   isValid = true;

return false;

if we found the pass "IndirectDXR" it mean the shaders IS HDRP AND support DXR :)
                
thougth?

sebastienlagarde: And we never clear the cache in this case (we shouldn't have a huge amount of shader)

eturquin: After further tests and discussion, I have followed the suggestion to move to a persistent shader-based cache (no clear at each frame anymore), however the actual test on the pass name didn't work, so we keep it as it was before. But since the cache hit rate is now extremely high, I removed the faster, dirtier test based on substrings.

sebastienlagarde (approved):

PR #4092, opened 2021-04-06
Fixed duplicate symbols in variants where real is defined as half. Th…

joshua-davis:
…is just adds guards in SRP core for functions that are also defined in built-in

---
### Purpose of this PR
Some variants weren't compiling for the Lit.shadergraph used in the testing projects. After some investigation, it turns out there were duplicate symbols only when real is defined as half (such as SHADER_API_MOBILE ). Currently there are 6 extra functions that needed guards.

---
### Testing status
Compiled hlsl, vulkan, and metal with all variants and checked the shader compiler log for any errors (actual errors, not warnings).

---
### Comments to reviewers
I half wonder if using this ifdef is appropriate? Mainly should this replacement only be done when real is half or always? This causes the half version to always get used even if we have single precision. Arguably this is more correct though since it's always falling back to built-in's behavior.


jessebarker (approved):
LGTM

PR #4093, opened 2021-04-06
Fix constant buffer being stomped on by shadows during async tasks

FrancescoC-unity:
This is a tad complex in that the global constant buffer is used by some tasks that run in the async compute pipe, but the shadow task that is running on the graphics pipe is updating the content of said global constant buffer. 

The CBuffer share the same memory so stomping happens; this PR fixes the issue by allowing a new instance of a constant buffer to be used by shadows. 

This required changing CBuffer API, also I changed how shadow atlas are init as it started to be very hard to read/follow. 

Fix for https://fogbugz.unity3d.com/f/cases/1324551/

NOTE: This _needs_ backport, but the backport is likely to be complex. I'll do a manual backport. I still targeted hd/bugfix, but let me know if you want me to target master instead. 

I'll wait for an answer to ^^ before doing the backport.

### DIFF
--- a/.yamato/_green_job_revisions_trunk.metafile
+++ b/.yamato/_green_job_revisions_trunk.metafile
@@ -53,15 +53,15 @@ shadergraph_pr_job_-_trunk:
       updated_at: '2021-04-25 13:33:54.600865'
       version: ''
 vfx_urp_pr_job_-_trunk:
-  updated_at: '2021-04-28 10:30:19.474513'
+  updated_at: '2021-04-29 07:30:45.029832'
   last_green_revisions:
     Android:
-      revision: b5c81da9dae099c1577789936e19e2a6c3fb961e
-      updated_at: '2021-04-25 13:33:54.600865'
+      revision: 576b9143c951f07dc76261e61c4ae7b5a1b64491
+      updated_at: '2021-04-29 00:02:48.083173'
       version: ''
     Linux:
-      revision: b5c81da9dae099c1577789936e19e2a6c3fb961e
-      updated_at: '2021-04-25 18:08:03.332305'
+      revision: 576b9143c951f07dc76261e61c4ae7b5a1b64491
+      updated_at: '2021-04-29 00:02:48.083173'
       version: ''
     OSX:
       revision: b5c81da9dae099c1577789936e19e2a6c3fb961e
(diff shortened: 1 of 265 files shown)


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
No QA requried, the Yamato test are enough

com.unity.render-pipelines.core/Runtime/Common/ConstantBuffer.cs
```diff
@@ -122,86 +122,134 @@ public static void ReleaseAll()
...
+
+        ComputeBuffer m_GPUConstantBuffer = null;
+
+        /// <summary>
+        /// Constant Buffer constructor.
+        /// </summary>
+        public ConstantBuffer()
+        {
+            m_GPUConstantBuffer = new ComputeBuffer(1, UnsafeUtility.SizeOf<CBType>(), ComputeBufferType.Constant);
+            ConstantBuffer.Register(this);
```
JulienIgnace-Unity, line 164: Register() should only be called for Singleton constant buffers (as it's necessary for ReleaseAll to work).
Explicitly allocated CBs must be released explicitly as well.

com.unity.render-pipelines.core/Runtime/Common/ConstantBuffer.cs
```diff
@@ -122,86 +122,134 @@ public static void ReleaseAll()
...
-
-                CoreUtils.SafeRelease(m_GPUConstantBuffer);
-                s_Instance = null;
+                s_Instance = value;
             }
         }
+
+        public override void Release()
+        {
+            s_Instance = null;
```
JulienIgnace-Unity, line 252: Need to call base.Release() here

com.unity.render-pipelines.core/Runtime/Common/ConstantBuffer.cs
```diff
@@ -105,14 +105,14 @@ public class ConstantBuffer
...
-            var cb = TypedConstantBuffer<CBType>.instance;
+            var cb = ConstantBufferSingleton<CBType>.instance;
 
             cb.Set(mat, shaderId);
         }
 
         /// <summary>
         /// Release all currently allocated constant buffers.
-        /// This needs to be called before shutting down the application.
+        /// This needs to be called before shutting down the application.
```
JulienIgnace-Unity, line 115: Maybe Update the doc to explain it's only for CBs used through the static API? (not sure of the right wording here)

com.unity.render-pipelines.core/Runtime/Common/ConstantBuffer.cs
```diff
@@ -19,7 +19,7 @@ public class ConstantBuffer
         /// <param name="shaderId">Shader porperty id to bind the constant buffer to.</param>
         public static void PushGlobal<CBType>(CommandBuffer cmd, in CBType data, int shaderId) where CBType : struct
```
JulienIgnace-Unity, line 20: Might be worth adding the PushGlobal version to regular CBs as well for consistency.

PR #4094, opened 2021-04-06
[ShaderGraph] [2021.2] Add upgrade buttons to ShaderGraph inspector

cdxntchou:

---
### Purpose of this PR
'Fixes' bug https://fogbugz.unity3d.com/f/cases/1304743/

We don't ever want to automatically modify the shadergraph files on disk, but this PR makes it easy for the user to invoke manually.

This adds buttons to the ShaderGraph inspector to Upgrade them to latest, and Upgrade All ShaderGraphs:
image

When you click one of the buttons, it asks for confirmation via dialog box:
image
image

It displays a progress bar when updating the whole project:
image

The buttons do not appear if the ShaderGraph selected is up to date:
image

---
### Testing status
Tested that upgrading results in the same graph, but no longer shows up as modified when loaded.
Verified that the modified files get reimported immediately.
Tested that the dialog for "Upgrade All" can be cancelled with no effect.
Tested that the inspector updates itself after upgrade, to not display the upgrade buttons anymore.
Tested that we properly check out files before modifying them.
Tested that yamato tests still pass after upgrade.

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Nodes/Input/Gradient/GradientNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Input/Gradient/GradientNode.cs
@@ -78,12 +78,17 @@ namespace UnityEditor.ShaderGraph
 
         public override void OnBeforeSerialize()
         {
+            // call property getter to ensure we've fully de-serialized, before we serialize.
+            // We must store the property value out to a variable, because we don't want to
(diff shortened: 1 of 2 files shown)


marctem:
The saving code is good but the UX isn't. We need to rethink how upgrading and asset management work together in a future version. Closing for now.

PR #4095, opened 2021-04-06
[ShaderGraph] Replace TextAsset usage with ShaderInclude

esmelusina:
---
### Purpose of this PR
Trunk is moving Text files with shader related extensions to a separate asset type, ShaderInclude. This change originally caused a regression for us (which has been resolved), but we need to move to the new asset type. This change only effects Custom Function Node's asset selector for File sources, improving the filtering to exclude unsupported extensions by default.

Related issues/bugs:
https://jira.unity3d.com/browse/GSG-387
https://fogbugz.unity3d.com/f/cases/1324528/
https://fogbugz.unity3d.com/f/cases/1322466/

Note that our previous soft-checking for file extensions remains (as it impacts some ui/error msgs), as the goal of this PR is just to comply with Trunk.

---
### Testing status
Manually checked asset type filtering on the ObjectField for all supported types. QA will do some additional testing. Extensions are further tested (to an extent) in the Custom Function Node related graphics tests.

---
### Notes for Reviewers
Min version needed to be bumped for ShaderGraph, for which requires a min version bump for all of graphics.

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Nodes/Utility/CustomFunctionNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Utility/CustomFunctionNode.cs
@@ -60,9 +60,9 @@ namespace UnityEditor.ShaderGraph
         };
 
         // With ShaderInclude asset type, it should no longer be necessary to soft-check the extension.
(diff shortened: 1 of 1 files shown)


joshua-davis (approved):

phi-lira (approved):

LandonTownsendUnity (approved):
Should be good now!

PR #4097, opened 2021-04-07
Fix to render depth or depth/normal of waving grass

Seongdae:
### Checklist for PR maker

---
### Purpose of this PR
Fix to render depth of waving grass in depth prepass
 - In opaque pass billboard offset was applied
 - But, those offset wasn't added to depth and depth/normal prepass
 - In case of billboard grass, the vertex result was not the same between opaque and prepasses

---
### Testing status
Add new terrain
Add detail billboard grass
Bush detail where you want to check

---
### Comments to reviewers
It wasn't working to draw depth of waving billboard grass to depth texture


ellioman (changes requested):
I'm not 100% sure what the bug is that this PR is trying to fix. Would be nice to get that info in the description.

With a PR that got introduced today, we now have a `WavingGrassDepthNormalsPass.hlsl` file with the DepthNormals passes, which also need to be looked at.

In the same PR, four test scenes were added in the Terrain Project to test depth & normals (Forward & Deferred). Can we expand on that and include some testing for this in order to avoid any breakages in the future?

Seongdae:
@ellioman Thanks for letting me know about `WavingGrassDepthNormalsPass.hlsl` added, I updated and merged from master to my branch.

In case of detail checked as billboard, depth prepass of waving grass didn't work and I tried frame debugger to check if the prepass worked. the prepass was called but depth didn't appear in depth texture.(depth normal is also the same)

`TerrainBillboardGrass(v.vertex, v.tangent.xy)` was applied in opaque pass before but depth prepasses didn't have it.
So I added `TerrainBillboardGrass(v.vertex, v.tangent.xy)` to vertex stage of depth and depth/normal pass then the result appeared in the frame debugger and the those work well in screen space features(ambient occlusion and shadow)

here are the comparison

[before adding `TerrainBillboardGrass(v.vertex, v.tangent.xy)`]
image

[after adding `TerrainBillboardGrass(v.vertex, v.tangent.xy)`]
image

phi-lira:
Adding @russoh for terrain awareness

russoh:
I'll loop in terrain team for a review, would be good to have somebody from the team, look over this.

spaceburp:
would it be possible to cover what Elvar brought up regarding tests?

> In the same PR, four test scenes were added in the Terrain Project to test depth & normals (Forward & Deferred). Can we expand on that and include some testing for this in order to avoid any breakages in the future?

russoh (commented):
The changes logically look sane to me, but I agree with other sentiments that we should probably add a test that would fail an image comparison if those changes were subsequently stripped out again. Also, do we have any larger terrain + grass demos that can be used to validate against?

phi-lira:
@Seongdae please ping reviewers for another look.

Seongdae:
Should I handle the tests of terrain and does it allow me to add scenes for comparing with detail shadows?

phi-lira:
@Seongdae yes, please add tests, ping @ellioman or @russoh for help.

ellioman:
What's the status on this PR @Seongdae ?
Regarding the test discussion, there is now two tests scenes for the DepthNormals that you can check and verify:
- 201_DepthNormalsDeferred
- 201_DepthNormalsForward

Seongdae:
@ellioman this is to fix depth/normals of billboard grass, but there is no billboard grass in the test scenes you told me. Should I add billboard grass to the scenes?

ellioman:
> @ellioman this is to fix depth/normals of billboard grass, but there is no billboard grass in the test scenes you told me. Should I add billboard grass to the scenes?

@Seongdae 
Yes, I think that's a good idea. Those scenes should be testing as much as possible Terrain related. So if you can structure the scene with the billboards as well, then that's perfect. 👍

Seongdae:
Checked it in the terrain test scenes @ellioman told me, and also update the reference images now. I'm waiting for Yamato, but there seem to be an issues in the current test environment.

Seongdae:
The tests related to Terrain were passed now.

ellioman (approved):

ellioman (approved):

PR #4098, opened 2021-04-07
Urp/feature/before transparent blocks

Seongdae:
### Checklist for PR maker

---
### Purpose of this PR
To process the render state before transparent passes.
While I was working on a renderer feature of screen space shadow, I added a specific render pass of restoring keywords from screen space shadow. It doesn't seems to be efficient so I would like to propose the better way

---
### Testing status
Tested on a example project of universal render pipeline

---
### Comments to reviewers
If this PR was approved, the 'TransparentSettingsPass' and 'RestoreShadowKeywordsPass' would be needed anymore.
It will provide a function that care about the states before transparent passes to each renderer

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
@@ -445,7 +445,7 @@ namespace UnityEngine.Rendering.Universal
         // To enable it - override SupportsNativeRenderPass method in the feature and return true
         internal bool disableNativeRenderPassInFeatures = false;
 
-        internal bool useRenderPassEnabled { get; private set; }
+        internal bool useRenderPassEnabled = false;
         static RenderTargetIdentifier[] m_ActiveColorAttachments = new RenderTargetIdentifier[] {0, 0, 0, 0, 0, 0, 0, 0 };
         static RenderTargetIdentifier m_ActiveDepthAttachment;
 


kaychang-unity (approved):

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -555,6 +556,14 @@ public virtual void SetupLights(ScriptableRenderContext context, ref RenderingDa
         {
         }
 
+        /// <summary>
+        /// Override this method to change the render state or prepare for transparent objects when transparent passes were enqueued.
+        /// </summary>
+        /// <param name="cmd"></param>
+        public virtual void BeforeTransparent(CommandBuffer cmd, ref RenderingData renderingData)
+        {
+        }
```
phi-lira, line 565: I'm reluctant to accept adding new public API, users can inject custom code before transparent with a custom pass, so I think this would be more ways of doing the same thing which leads to confusion.

phi-lira (changes requested):
Requesting changes as I'm not convinced we need to add a new public API since we could execute custom code before transparent with existing API today. 
 
 For this particular issue you are trying to solve, we recently added _SURFACE_TYPE_TRANSPARENT keyword you could use to sample shadows differently for transparents instead of disabling screenspace shadow keyword. https://github.com/Unity-Technologies/Graphics/pull/4025

Seongdae (commented):
@phi-lira Thanks, I didn't know about '_SURFACE_TYPE_TRANSPARENT', It seems to help me and be able to find a good solution for me.

Seongdae:
@phi-lira Apply '_SURFACE_TYPE_TRANSPARENT' to shadowing on screen space. But I can't avoid using restoring keywords of main light shadow after screen space shadows.

In frame debugger transparent objects didn't use screen space shadow', but use shadow map directly with '_MAIN_LIGHT_SHADOWS_SCREEN' enabled. I think it makes users confused also the transparent objects couldn't recognize whether the shadow was cascades or no-cascade. So I used a pass that restores keywords of main light shadow for transparent pass again.

It can make sure that screen space shadows is not be able to be used in shader code for transparent objects. It looks good.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Seongdae: I removed it and restored them to the original codes

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -693,6 +702,9 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
                     ExecuteBlock(RenderPassBlock.MainRenderingOpaque, in renderBlocks, context, ref renderingData);
                 }
 
+                // Before transparent blocks...
+                InternalBeforeTransparent(context, ref renderingData);
+
```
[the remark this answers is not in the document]
Seongdae: Removed

phi-lira (approved):
LGTM.

phi-lira (approved):

erikabar (approved):
no regressions found, did a quick smoke test with UniversalRenderingExamples project

phi-lira:
Getting back to look at PRs now that we are past cutoff. @Seongdae can you let me know when merge conflict are solved, then I will merge PR.

phi-lira:
Missing ABV. Kicking.

Seongdae:
@phi-lira ABV was succeeded except Top Asset Store tests

PR #4099, opened 2021-04-07
[Universal] RenderPass merging

jonuuukas:
### Purpose of this PR
This is a follow up PR for the initial RenderPass PR: https://github.com/Unity-Technologies/Graphics/pull/3380.
This PR introduces the RenderPass merging mechanism, that should allow for eligible passes to be merged.

This also includes:
-  moving most of the native RenderPass logic to a separate partial class in NativeRenderPass.cs.
-  disabling RendererFeatures to use NativeRenderPass unless specific override is called to enable
-  all the technical intricacies of merging passes and subpasses (please read through the TDD linked below for the implementation details)

TDD and more in-depth explanation: https://docs.google.com/document/d/1ddJS70UnbyUT21FxvSgeVy6XFoZXScrMWn5u7J4l790

Entry point for the whole RenderPass Conversion:
https://docs.google.com/document/d/174wpA7AvycbY65wg_Nd3FHt2zJK7sAnOx3qiDUynSro

---
### Testing status
Manually ran the URP editor tests, will run them on iOS and compare non-RenderPass and RenderPass rendering paths.
---
### Comments to reviewers

NOTE: This PR won't enable RenderPasses. The setting will still be disabled, and the selector hidden to users. This is an incremental step to enable further work and optimizations.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
+++ b/com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
@@ -508,11 +508,10 @@ namespace UnityEngine.Rendering.Universal
 
         private static GraphicsFormat GetDefaultGraphicsFormat(CameraData cameraData)
         {
-            bool isHdrEnabled = cameraData.isHdrEnabled;
-
-            GraphicsFormat hdrFormat = GraphicsFormat.None;
-            if (isHdrEnabled)
+            if (cameraData.isHdrEnabled)
             {
+                GraphicsFormat hdrFormat = GraphicsFormat.None;
+
                 if (!Graphics.preserveFramebufferAlpha &&
                     RenderingUtils.SupportsGraphicsFormat(GraphicsFormat.B10G11R11_UFloatPack32,
                         FormatUsage.Linear | FormatUsage.Render))
@@ -522,9 +521,11 @@ namespace UnityEngine.Rendering.Universal
                     hdrFormat = GraphicsFormat.R16G16B16A16_SFloat;
                 else
                     hdrFormat = SystemInfo.GetGraphicsFormat(DefaultFormat.HDR);
+
+                return hdrFormat;
             }
 
-            return isHdrEnabled ? hdrFormat : SystemInfo.GetGraphicsFormat(DefaultFormat.LDR);
+            return SystemInfo.GetGraphicsFormat(DefaultFormat.LDR);
         }
     }
 }


CrzDog:
hi, this is a good step for mobile platform, any plan to move urp to render graph/ frame graph based rendering architecture? I think render graph is more suitable for native render pass merging

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -187,7 +188,10 @@ public Color clearColor
         internal bool depthOnly { get; set; }
         // this flag is updated each frame to keep track of which pass is the last for the current camera
         internal bool isLastPass { get; set; }
+        // index to track the position in the current frame
+        internal int renderPassQueueIndex { get; set; }
 
+        internal NativeArray<int> m_InputAttachmentIndices;
```
lukaschod, line 194: I dont see where it is used

jonuuukas: @lukaschod - yeah, it's not used yet. we can remove it for now, but it will still be added in next PR that will introduce input attachments/fb fetch and etc.

sandy-carter-unity (approved):

thomas-zeng (approved):

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -0,0 +1,528 @@
...
+                    var rpDesc = InitializeRenderPassDescriptor(cameraData, renderPass);
+
+                    renderPass.isLastPass = false;
+                    renderPass.renderPassQueueIndex = i;
+
+                    Hash128 hash = CreateRenderPassHash(rpDesc, currentHashIndex);
+
+                    m_PassIndexToPassHash[i] = hash;
+
+                    bool RPEnabled = renderPass.useNativeRenderPass && isRenderPassEnabled;
```
nigeljw-unity, line 103: Potential to move this before CreateRenderPassHash

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -0,0 +1,528 @@
...
+        {
+            for (int i = 0; i < array.Length; ++i)
+                if (array[i] == -1)
+                    return i;
+            return array.Length - 1;
+        }
+
+        internal static Hash128 CreateRenderPassHash(int width, int height, int depthID, int sample, uint hashIndex)
+        {
+            return new Hash128((uint)width * 10000 + (uint)height, (uint)depthID, (uint)sample, hashIndex);
```
nigeljw-unity, line 492: Let's use something a bit more explicit than the multiply by 10000 as we discussed in the call. Probably a bit shift would be more intuitive.

nigeljw-unity (commented):
Just added the comments from our call. I'll do a deeper dive tomorrow.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -178,7 +179,7 @@ public Color clearColor
         internal bool overrideCameraTarget { get; set; }
         internal bool isBlitRenderPass { get; set; }
 
-        internal bool useNativeRenderPass { get; set; }
+        public bool useNativeRenderPass { get; set; }
```
lukaschod, line 182: As this is public it needs at least some api documentation

lukaschod (approved):
Looks good

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
phi-lira: Actually we should not make this public. Why is that? @jonuuukas ?

jonuuukas: initially this was needed to disable the API for injected passes, but as we did change how that works this, i think, is no longer necessary. Let me quickly test this and if all is good, i'll revert it

ernestasKupciunas (approved):
PR looks good. I was running projects using Deferred and Forward renderers. Tested Metal, Vulkan and Gles3 Graphics APIs. Visual correctness looks good. No suspicious activity in the logs. No bugs related to this PR. Good job!
Winter:
Screen Shot 2021-04-26 at 13 06 58

BoatAttack:
Screen Shot 2021-04-26 at 17 13 38-min
PolyFarm:
Screen Shot 2021-04-26 at 14 59 34
Projects used for testing:

- [Winter](https://github.com/ernestasKupciunas/Winter.git)
- [BoatAttack](https://github.com/Unity-Technologies/BoatAttack)
- [PolyFarm](https://github.com/Unity-Technologies/PolygonFarmBenchmark)

Devices under testing:

Unity used for testing:
Version: 2021.2.0a16.2217
Revision: trunk b5c81da9dae0
Built: Sun, 25 Apr 2021 10:56:38 GMT

nigeljw-unity (commented):
Still got some other things I want to go over, but just posting these comments in advance of the meeting.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -916,7 +909,11 @@ private bool IsRenderPassEnabled(ScriptableRenderPass renderPass)
             // Track CPU only as GPU markers for this scope were "too noisy".
             using (new ProfilingScope(null, Profiling.RenderPass.configure))
             {
-                renderPass.Configure(cmd, cameraData.cameraTargetDescriptor);
+                if (IsRenderPassEnabled(renderPass) && cameraData.cameraType == CameraType.Game)
```
phi-lira, line 912: `CameraType.VR` is also "game" camera.

com.unity.render-pipelines.universal/Runtime/ScriptableRendererFeature.cs
```diff
@@ -39,6 +39,14 @@ void OnValidate()
             Create();
         }
 
+        /// <summary>
+        /// Override this method and return true if the feature should use the Native RenderPass API
+        /// </summary>
+        public virtual bool SupportsNativeRenderPass()
```
phi-lira, line 45: Can we make this internal until we actually enable the feature / discuss if this is final API?

phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -0,0 +1,528 @@
...
+        internal struct RenderPassDescriptor
+        {
+            internal int w, h, samples, depthID;
+
+            internal RenderPassDescriptor(int width, int height, int sampleCount, int rtID)
+            {
+                w = width;
+                h = height;
+                samples = sampleCount;
+                depthID = rtID;
```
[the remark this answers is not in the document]
manuele-bonanno: as discussed in the call: this will require Vulkan work to be added to the backlog. Currently Unity doesn't support this

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -0,0 +1,528 @@
...
+            return new RenderPassDescriptor(w, h, samples, depthID);
+        }
+
+        private static GraphicsFormat GetDefaultGraphicsFormat(CameraData cameraData)
+        {
+            GraphicsFormat hdrFormat = GraphicsFormat.None;
+            if (cameraData.isHdrEnabled)
+            {
+                if (!Graphics.preserveFramebufferAlpha &&
+                    RenderingUtils.SupportsGraphicsFormat(GraphicsFormat.B10G11R11_UFloatPack32,
```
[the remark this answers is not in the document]
manuele-bonanno: making all the requested changes except this because of coding convention requirement for single line if statements (as discussed in the call)

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
manuele-bonanno: as discussed with the VR team, before enabling VR support a VR pass is needed to fix trunk issues and add URP support. It will be enabled only on mobile first (we have no means to test XR/VR ourselves)

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -0,0 +1,528 @@
...
+                    var depthAttachmentTarget = (passDepthAttachment != BuiltinRenderTextureType.CameraTarget) ?
+                        passDepthAttachment : (usesTargetTexture
+                            ? new RenderTargetIdentifier(cameraData.targetTexture.depthBuffer)
+                            : BuiltinRenderTextureType.Depth);
+
+                    // TODO: review the lastPassToBB logic to mak it work with merged passes
+                    // keep track if this is the current camera's last pass and the RT is the backbuffer (BuiltinRenderTextureType.CameraTarget)
+                    // knowing isLastPassToBB can help decide the optimal store action as it gives us additional information about the current frame
+                    bool isLastPassToBB = pass.isLastPass && (colorAttachmentTarget == BuiltinRenderTextureType.CameraTarget);
+                    currentAttachmentDescriptor.ConfigureTarget(colorAttachmentTarget, ((uint)finalClearFlag & (uint)ClearFlag.Color) == 0, !(samples > 1 && isLastPassToBB));
```
nigeljw-unity, line 269: I know this proceeded this change, but is this actually the right logic? !(samples > 1 && isLastPassToBB)
Can you clarify it for me, so we want to store when no msaa or when this is not the last pass for the back buffer?

manuele-bonanno: this is old logic from before the merge. For now we always store but the next PR should optimize that

manuele-bonanno: in general though yes, if the pass wouldn't be merged (which was the case before this PR) this would try to discard if MSAA > 1 and is the last pass.
Now with the merging all the store/discard logic will have to take into account the mergeable pass list

hdb-unity (approved):
No PR issues found - checked for regressions in 2x Universal RP projects.

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
manuele-bonanno: just to clarify: "why when MSAA > 1 we requested to store?"

this require an explanation of how Unity implicitly converts to resolve actions in the device implementations:

when MSAA is on, internally the device code converts DontCare actions in Resolve, and Store actions in StoreAndResolve. If the pass is the last pass to BB then we want to NOT store, so the action will become Resolve. Resolve still stores the resolved surface but will discards the MSAA'd surface, while StoreAndResolve would also store the MSAA'd surface which wastes a lot of bandwidth.

If MSAA is OFF: then you want to store the non-AA RT, otherwise the final result of the pass wouldn't make it to the BB memory as it would be discarded

Hope it makes sense now

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
phi-lira: Ok. Sounds good.

manuele-bonanno (approved):

nigeljw-unity (approved):

phi-lira (approved):

PR #4105, opened 2021-04-07
Sample distortion mip chain with different safe guard to avoid NaNs

FrancescoC-unity:
Change the way we clamp the sampling of the color pyramid in distortion, previous way was not solid under non even sizes. 

Fix for https://fogbugz.unity3d.com/f/cases/1324548/

For QA: I was not able to test much as I couldn't repro :D But Adrian did test (he could repro) and it worked for him.


FrancescoC-unity:
So this is not enough, gonna look into it.

FrancescoC-unity:
NVM, need to think more about this.

TomasKiniulis (approved):
I do not reproduce this issue on my Xbox, since Adrien was able to validate I think this is good to go.

sebastienlagarde (approved):

PR #4110, opened 2021-04-07
Re eneabling de exposure. It was originally disabled for vulkan since…

kecho:
### Purpose of this PR
Re-enabling de-exposure on unlit (which was fixed for all platforms except vulkan).  Original PR with the fix was this one:
https://github.com/Unity-Technologies/Graphics/pull/3627
We had to avoid this fix for vulkan and shader graphs due to a compiler issue in vulkan (compiler was using wrong instruction).

This fix re-enables the fix. Keeping the define just in case we have to turn de-exposure off.
The compiler fix is found here:
https://ono.unity3d.com/unity/unity/pull-request/123549/_/graphics/vulkan/shader-compiler-ubfe

Information on the compiler issue: https://fogbugz.unity3d.com/f/cases/1323529/

---
### Testing status
Testing on all platforms, but specifically interested on the results of vulkan 

---
### Comments to reviewers

---
### Comments to qa
The bug  https://fogbugz.unity3d.com/f/cases/1323529/ has the fix which has landed. We are only pending this PR to close it completely. Can you guys assign it to the right person (once this PR passes review). Not sure what to do with the bug :)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
@@ -104,14 +104,6 @@ Pass
     #endif
     #endif
 
-    // NOTE: disabling de exposure usage on unlit shader graphs (caused by virtual texture constant buffer binding).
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

sebastienlagarde:
For QA, no need for QA here, just added for awareness, autoamtion is enough

PR #4118, opened 2021-04-07
[HDRP] Fix issue with the color space of AOVs

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1324759/
We were getting incorrect results in AOVs because in the debug output, we consider that the final target will apply an sRGB curve, which isn't the case when we export to AOV.

---
### Testing status
- After following the repro steps, we now get the correct/expected values for the normal AOV.
- I have updated the compositor test (the AOV layer was affected)

---
### Comments to reviewers
The existing AOV API provides enough flexibility to use both sRGB and standard buffers as output.  On the shader side we always output linear values to AOVs, and any sRGB conversion will happen based on the buffer type that the user has allocated. 
So the user for example should not allocate sRGB buffers to output normals, unless he wants to do so.


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
@QA: this PR have been validated by Jean Blouin with recorder. Merging

PR #4119, opened 2021-04-07
Fix for graph settings inspector switching 

elizabeth-legros:
We needed a way to trigger the inspector update without forcing the inspector view to the Node Settings tab. In order to not break any interfaces, decided to add a static flag that we toggle before triggering the inspector update in this way. 
https://fogbugz.unity3d.com/f/cases/1327208/
Was not able to reproduce the bug after fix.
Green run https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix-for-inspector-selection/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/6302631/job/pipeline


Nightmask3 (approved):
Approving!

ghost (approved):
Tested with properties, nodes, and stack nodes

PR #4120, opened 2021-04-08
Fix issue with OnDemand shadows assert getting incorrectly triggered

FrancescoC-unity:
The issue came from a missing check whether an element was already put in the Shadow has valid data dictionary; this probably came as an issue on my end not pushing all the local changes I had the time. 

In any case this fixes the issue.  There is no fogbugz as it was a report from forum. 

**What did I test**: Had a moving spotlight with OnDemand and trigger RequestShadowRendering multiple times. Before this fix an assert would show and the shadow wouldn't get updated; with this fix the shadow updates as expected.

**NOTE**: This should not be backported to 10.x.x, I'll retarget master.


FrancescoC-unity:
Gah need to reopen targeting master.

PR #4121, opened 2021-04-08
Fix issue with OnDemand shadows assert getting incorrectly triggered

FrancescoC-unity:
The issue came from a missing check whether an element was already put in the Shadow has valid data dictionary; this probably came as an issue on my end not pushing all the local changes I had the time.

In any case this fixes the issue. There is no fogbugz as it was a report from forum.

**What did I test**: Had a moving spotlight with OnDemand and trigger RequestShadowRendering multiple times. Before this fix an assert would show and the shadow wouldn't get updated; with this fix the shadow updates as expected. 
I can provide scripts to QA if they wanted to test more.

NOTE: This should not be backported to 10.x.x, hence why it is targeting master.


TomasKiniulis (approved):
RequestShadowMapRendering() does not throw any errors now. Tested Point, Spot and Area On Demand lights with Update On Light Movement toggled on and off. Works as expected!

master:

https://user-images.githubusercontent.com/50582134/114404301-4047e380-9bae-11eb-9e15-42af23e5bfe2.mp4

PR:

https://user-images.githubusercontent.com/50582134/114404347-4938b500-9bae-11eb-81f3-d3bc96fb1834.mp4

sebastienlagarde (approved):

PR #4124, opened 2021-04-08
[HDRP] Fix camera controller in template with old input system

alelievr:
### Purpose of this PR
Fix the camera controller in the template with the old input system (https://fogbugz.unity3d.com/f/cases/1326816/)
I also adjusted the sensitivity of the the new input system so it feels more like the old one

---
### Testing status
Tested in both old and new input system with mouse & gamepad (the old system doesn't support gamepads)

https://user-images.githubusercontent.com/6877923/114010130-14ea8f00-9864-11eb-859a-453085a70c46.mp4

https://user-images.githubusercontent.com/6877923/114010135-15832580-9864-11eb-9db7-2d068dca7a66.mp4

https://user-images.githubusercontent.com/6877923/114010139-161bbc00-9864-11eb-8c01-5ed721c33b63.mp4


com.unity.template-hd/Assets/SampleSceneAssets/Scripts/SimpleCameraController.cs
```diff
@@ -188,7 +191,7 @@ void Update()
             // Rotation
             if (IsCameraRotationAllowed())
             {
-                var mouseMovement = GetInputLookRotation() * Time.deltaTime * 5;
+                var mouseMovement = GetInputLookRotation() * Time.deltaTime * 6 * mouseSensitivity;
```
RSlysz, line 194: I think if you already have an exposed sensitivity factor, the 6 should be part of it.

RSlysz (approved):

com.unity.template-hd/Assets/SampleSceneAssets/Scripts/SimpleCameraController.cs
alelievr: Ok

JulienIgnace-Unity (approved):

RemyUnity (approved):
Kudos for the videos for comparison. It seems good to me.

PR #4126, opened 2021-04-08
updated icons

martint-unity:
---
### Purpose of this PR
Updated with new icons, Theses are used at the moment in the converter framework

### Tests
Check the icon help is not broken
image

### DIFF
--- a/com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
+++ b/com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
@@ -122,7 +122,7 @@ namespace UnityEditor.Rendering
             iconFail = EditorGUIUtility.FindTexture("console.erroricon");
             iconSuccess = EditorGUIUtility.FindTexture("TestPassed");
             iconPending = EditorGUIUtility.FindTexture("Toolbar Minus");
-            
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/Editor/EditorWindowWithHelpButton.cs
```diff
@@ -5,11 +5,13 @@ namespace UnityEditor.Rendering
     /// <summary>Editor window that adds a button to browse the help url, specify <see cref="HelpURLAttribute"/> when defining your inherited class</summary>
     public class EditorWindowWithHelpButton : EditorWindow
     {
+        private static GUIContent iconHelpGUIContent = new GUIContent(CoreEditorStyles.iconHelp);
```
alex-vazquez, line 8: static GUIContent m_IconHelpGUIContent = ...

alex-vazquez (approved):

com.unity.render-pipelines.core/Editor/EditorWindowWithHelpButton.cs
martint-unity: done

PR #4130, opened 2021-04-08
Fixed Clear GBuffer frame setting.

JulienIgnace-Unity:
### Purpose of this PR
This PR fixes the "Clear GBuffer" frame settings.
The fix can be checked by taking a render doc capture while the frame setting is enabled and checking that the GBuffer is indeed cleared before rendering.


RemyUnity (approved):
LGTM

FrancescoC-unity (approved):

sebastienlagarde (approved):

PR #4131, opened 2021-04-08
[12.0.0] Create URP Graphics Test Package

phi-lira:
### Purpose of this PR
 This PR creates a URP graphics test package. This is useful for two reasons:
  - we don't duplicate scripts and assets between test projects
  - the package has the graphics test framework dependency so other teams / packages can integrate test with URP by adding the URP graphics test package.

### Testing status
 - Run tests locally.

### Comments to reviewers
 - Common assets and scripts were moved to the new graphics test package.
 - Each split project references the new graphics test package.
 - I could see some issues / warning in the test projects but they didn't seem related to my changes so I plan to fix them in follow PR.


phi-lira:
Closing. Next iteration https://github.com/Unity-Technologies/Graphics/pull/4182

PR #3859, opened 2021-03-12
Added support for multi volumetric cloud shadows.

anisunity:
Due to the mechanic of using the sun cookie to get volumetric cloud shadows, various "validity" limitations were occuring with multiple cameras.

This required a bit of texture management to have everything stilll work when we have multiple cameras. The problem also happens with cloud layers (@adrien-de-tocqueville for visibility) and must be fixed.

Testing status
The issue doesn't happen anymore with multiple of cameras.
The cloud shadows tests are still green.

@remi-chapelain if you can check the cloud shadow visual debug view

PR #4132, opened 2021-04-08
Fix for multi cameras using cloud layers shadows

adrien-de-tocqueville:
### Purpose of this PR

equivalent of https://github.com/Unity-Technologies/Graphics/pull/3859 for cloud layer
Fix cloud layer shadows when multiple cameras are part of diferent volumes.
Also now correctly put back original cookie on light component when cloud layer is disabled

This adds a new api on the cloud system but I don't see any other way to do it

gif showing the issue. Each time a camera gets in focus, it replace the shadow cookie for the other one.
The correct behavior here should be no shadow for the scene view camera (because it's outisde of any volume), and the shadow shown in each camera preview should be different but always the same for each camera

---
### Testing status
Tested that the issue doesn't happen anymore and shadow still works
Also tested with volumetric clouds just to make sure


com.unity.render-pipelines.high-definition/Runtime/Sky/CloudSystem/CloudLayer/CloudLayer.cs
```diff
@@ -173,7 +176,7 @@ public class CloudMap
             public ClampedFloatParameter thickness = new ClampedFloatParameter(0.5f, 0, 1);
 
             /// <summary>Enable to cast shadows.</summary>
-            [Tooltip("Projects a portion of the clouds around the sun light to simulate cloud shadows.")]
+            [Tooltip("Projects a portion of the clouds around the sun light to simulate cloud shadows. This will override the cookie of your directional light.")]
```
sebastienlagarde, line 179: Let's be sure this information is also in the documentation

sebastienlagarde (approved):
let's talk

sebastienlagarde:
@Unity-Technologies/gfx-qa-hdrp  let's be sure that volumetric aren't affected

iM0ve (approved):
* Verified that each camera can now have different light layer shadows
* That volumetric clouds still work
* Read-only properties are no longer on the override (tiling, directional light)
* Newly added property shadow size works as expected
* Cookies on dir and spot light are still working as before


https://user-images.githubusercontent.com/36502659/115742586-7e58ba80-a399-11eb-9e6f-8dc81f0249a7.mp4

adrien-de-tocqueville:
I moved the place where the sun light is determined in the light loop so that shadow now shows during the first frame

PR #4133, opened 2021-04-08
Fix Panini Projection for floating point color buffers (Case 1325845)

johnpars:
### Purpose of this PR
Fix case [1325845](https://fogbugz.unity3d.com/f/cases/1325845/). 
When a user uses Panini Projection with floating point buffers configured for HDRP and Post Processing (at the same time; the issue does not reproduce if only one or the other is set), then they will get a blown out result like this:
image

The issue was that the panini pass was assuming a float3 buffer and not accounting for the case of an alpha, which is necessary for the floating point buffers. Why exactly this only mattered when _both_ HDRP + PP buffers are configured this way, I am still not sure. I assume something about this particular setup upset the compiler. Accounting for the alpha case solves the issue in this scenario:
image

---
### Testing status
I manually walked through the repro steps of the FB case and confirmed that the issue is no longer present. 

---
### Comments to reviewers
Introduces new variants for panini projection to compile for with/without alpha.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -151,6 +151,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed the camera controller in the template with the old input system (case 1326816).
 - Fixed broken Lanczos filter artifacts on ps4, caused by a very aggressive epsilon (case 1328904)
 - Fixed global Settings ignore the path set via Fix All in HDRP wizard (case 1327978)
+- Fixed GBuffer clear option in FrameSettings not working
 - Fixed usage of Panini Projection with floating point HDRP and Post Processing color buffers.
 
 ### Changed
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
@@ -56,10 +56,9 @@ namespace UnityEngine.Rendering.HighDefinition
             CoreUtils.Destroy(m_DownsampleDepthMaterial);
         }
 
-        bool NeedClearGBuffer()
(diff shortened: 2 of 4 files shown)


remi-chapelain:
Adding HDRP QA for awareness :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -3019,6 +3019,9 @@ TextureHandle PaniniProjectionPass(RenderGraph renderGraph, HDCamera hdCamera, T
                     else
                         passData.paniniProjectionCS.EnableKeyword("UNITDISTANCE");
 
+                    if (m_EnableAlpha)
```
JulienIgnace-Unity, line 3022: You may want to add the else statement to disable the keyword. 
It will avoid issues when doing two cameras back to back with one using alpha and the other not. The second camera would then not use the proper keyword.
(It's not currently possible to have different color buffer types per camera but with render graph it should be soon a possibility so might as well do it right now)

johnpars: Will do - also there are a few other instances in the post stack that only enable that keyword, like Stop NaNs pass: 
https://github.com/Unity-Technologies/Graphics/blob/a85221a204200b45175d26a4ea95195b5ed2e51f/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs#L546-L547

I can also fix that up for the other instances of it in this PR.

RemyUnity:
We do have a graphic test for panini project, but not for this buffer format.
Should we consider adding graphics tests for post processes with both color buffer settings ?

remi-chapelain (approved):
Pretty simple fix, accepted on behalf on QA. ✔️ 
We will add a test in another PR that will cover this case along with other Post process that makes sense. (needs to be discussed which one)

sebastienlagarde (approved):

PR #4134, opened 2021-04-08
[HDRP] Add support for alpha channel in FXAA

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1323941/
FXAA was not taking alpha into account (we got the request from a compositor user)

After this PR, FXAA will:
- properly antialias the alpha channel
- affect only on pixels with alpha > 0

---
### Testing status
I have tested this with the graphics compositor:

**Before:** (Alpha channel of a sphere with FAA on)
image

**After:**
image

**Before:** (The sphere is on a separate compositor layer. FXAA is enabled only for the sphere but affects the entire image)
image

**After:** (FXAA affects only the sphere)
image

---
### Comments to reviewers
This does not have any impact when alpha post-processing is disabled.
For SMAA we can do another PR in the future.


FrancescoC-unity:
Looks fine, but probably need to update FXAA.compute too.

FrancescoC-unity:
Also, I guess is worth mentioning in  the doc that it will lead to approx 2x the cost given we need to run FXAA twice (though I 100% agree that's the way to go)

pmavridis:
I have tested the FXAA.compute using fixed dynamic resolution and it was working fine. I did not manage to test with a scaler function, but I don't expect that it will change anything.

> Also, I guess is worth mentioning in the doc that it will lead to approx 2x the cost
The docs [1] already mention that "enabling this option incurs a performance overhead." so I think we are OK.
Btw, the shader goes from 8 to 12 memory fetches, so I guess the cost should be closer to 1.5x, given that ALU for alpha is also a bit more "lite".

[1] https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@10.4/manual/Alpha-Output.html

sebastienlagarde (approved):

TomasKiniulis:
@pmavridis should there be any graphics tests added for this? Any specific cases QA should have a look here?

pmavridis:
> @pmavridis should there be any graphics tests added for this? Any specific cases QA should have a look here?

Yes, I will update the PR with an extra test for the alpha support. I don't think there are any special cases to test.

sebastienlagarde:
@GFX QA: merging for pacakge 10.5.0 preparation

PR #4135, opened 2021-04-08
AnimationClip upgrade 

sim-bz:
### Purpose of this PR
When users migrate from legacy to URP, any AnimationClips that have curves bound to material properties are likely to break as a consequence of shader property name changes. (E.g., a curve bound to `_Color` will have no effect, if the shader used by the material on the target object now uses the property name `_BaseColor`.) Moreover, because of how the `Renderer` component exposes animatable bindings, users get no feedback in the Animation window that a curve will no longer have any effect. **This PR ensures that AnimationClips are upgraded along with materials.**

Full details of the problem and the reason for the solution adopted in this PR can be found in [this doc](https://docs.google.com/document/d/115HmJqpvohkHVXcGlyCBRzrWzlURxl8W1Zc_hkLTqNU/edit#).

---
### Testing status
Package `com.unity.testing.urp-upgrade` contains tests covering the new workflows.

New automated unit tests:
- `AnimationClipUpgrader_UnitTests.cs`
- `UpgradeUtility_UnitTests.cs`

New integration tests with actual Unity assets (AnimationClip, Timeline, Prefab, etc.)
- `AnimationClipUpgrader_IntegrationTests.cs`

**Animation team will assign one of our QA for a manual test pass once graphics engineers and QA have signed off on this PR.** 

---
### Comments to reviewers

**General:**
- The majority of files in this PR are assets in the new test package used for integration testing. They have been audited by the animation team and should generally be ignorable. **You should be able to filter out everything that is not `.asmdef`, `.cs`, or `.md`**
- All new code is editor-only.

**URP Devs:**
- Per @phi-lira 's comments on the draft PR, we have put most of the new code in the `com.unity.render-pipelines.universal` package. This was done to minimize changes to internals visibility across assemblies, and with sensitivity to the fact that current product requirements are constrained to the upgrade experience going from Legacy -> URP.
- Per a discussion with graphics devs and PM, we have made the existing material upgrade menu items call the animation clip upgrade code. (See changes in `UniversalRenderPipelineMaterialUpgrader.cs`.) This was done to avoid adding new menu items, and in light of the implicit assumption that materials have already been upgraded in order for us to be able to update AnimationClips. **It currently opens a dialog after materials have been upgraded, before upgrading clips. We need URP owners to provide feedback on whether they would prefer this dialog be omitted.**
- All code in `AnimationClipUpgrader.cs` and `AnimationClipUpgrader_Types.cs` is animation-specific, and has been internally reviewed by the animation team. These files contain domain-specific code for how we gather dependencies between clips and prefabs that use them, as well as domain-specific code for how we actually upgrade AnimationClips, when necessary. **It adds no new public API.**
- `UpgradeUtility.cs` contains domain-agnostic code that we believe should be considered for adoption in the upcoming general-purpose upgrade framework at some point, as it contains utilities for assets that may have serialized shader property names which have changed. (See [this doc](https://docs.google.com/document/d/115HmJqpvohkHVXcGlyCBRzrWzlURxl8W1Zc_hkLTqNU/edit#) for details.) **It adds no new public API.** Its tests are in `UpgradeUtility_UnitTests.cs`.
- `MaterialUpgrader.cs` **adds new internal members** to the `MaterialUpgrader` class. These members provide the upgrade utility information about an upgrader's property rename tables, and the new shader it upgrades to. In order to avoid committing any new public API, `com.unity.render-pipelines.core/Editor/AssemblyInfo.cs` adds internals visibility for `Unity.RenderPipelines.Universal.Editor` and `Unity.RenderPipelines.Universal.Editor.Tests`.

**Graphics QA:**
- All new tests were added in a dedicated test package `com.unity.testing.urp-upgrade`. Most of these new files are animation-specific tests and related test code. **We need graphics QA to provide feedback on the naming of this package and/or location of these integration tests.** We believe it will be easiest to maintain if kept separate from other upgrade projects that are currently landing, but also want to ensure it does not add substantially to the overhead from graphics CI. **We also need graphics QA to advise on appropriate triggers to run tests in this project.**
- `Unity.Testing.SRP.Universal-Upgrade.Editor.asmdef` (in com.unity.testing.urp-upgrade) adds an assembly reference to Moq, the mocking framework that is approved for use at Unity. It is used by the new unit tests added to that assembly in order to create actual unit-level tests for the new code in this PR. From what we can see, the render pipeline packages do not generally express any explicit dependencies on test packages, even though their test assemblies use them (e.g., `com.unity.test-framework`). **We need graphics QA to advise on one of the following approaches:**
    ~~1. Add a dependency on `nuget.moq` to `com.unity.render-pipelines.universal`~~
    2. Move the new unit tests _and_ the dependency on `nuget.moq` to the recently created `com.unity.testing.urp-upgrade` package.
- `com.unity.render-pipelines.universal/Editor/AssemblyInfo.cs` adds internals access for Moq (see above) as well as for the test project. **We need graphics QA to provide feedback on this approach.** Internals access for Moq is standard practice for mocking interfaces that are not part of the public API. However, in the case of the test project, while we would normally just use reflection to call and test non-public methods, there are a number of internal types and interfaces involved that make it insurmountably difficult.

### DIFF
--- a/.yamato/config/universal_upgrade.metafile
+++ b/.yamato/config/universal_upgrade.metafile
@@ -5,7 +5,6 @@ test_platforms:
   - type: editmode
     commands:
       - [all]: npm install upm-ci-utils@stable -g --registry https://artifactory.prd.cds.internal.unity3d.com/artifactory/api/npm/upm-npm
-      # - [all]: upm-ci package pack --package-path com.unity.testing.urp-upgrade
       - [all]: upm-ci project pack --project-path TestProjects/UniversalUpgradeTest
       - [all]: upm-ci package test -u trunk --package-path com.unity.testing.urp-upgrade --type package-tests --extra-create-project-arg=-upmNoDefaultPackages
 platforms:


martint-unity:
@pbbastian for awareness

com.unity.render-pipelines.core/Tests/Editor/UpgradeUtility_UnitTests.cs
```diff
@@ -0,0 +1,232 @@
+using System.Collections.Generic;
+using System.Linq;
+using Moq;
+using NUnit.Framework;
+using IMaterial = UnityEditor.Rendering.UpgradeUtility.IMaterial;
+using RenameType = UnityEditor.Rendering.MaterialUpgrader.RenameType;
+using static UnityEditor.Rendering.Tests.UpgraderTestUtility;
+
+namespace UnityEditor.Rendering.Tests
```
amechtley, line 9: Graphics: These tests cover the internal UpgradeUtility class that I propose you should add/adjust into public API for your upgrader tool

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
```diff
@@ -54,7 +54,7 @@ private static void GetShaderNamesToIgnore(ref HashSet<string> shadersToIgnore)
             shadersToIgnore.Add("Sprites/Default");
         }
 
-        private static void GetUpgraders(ref List<MaterialUpgrader> upgraders)
+        internal static void GetUpgraders(ref List<MaterialUpgrader> upgraders)
```
amechtley, line 57: Graphics: Please note private API made internal here

amechtley (commented):
@phi-lira @martint-unity Because this PR mostly consists of test project assets, I have added comments to parts we would particularly like you to evaluate, to make it easier to find them. Other .cs files are pretty strictly animation-related.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAnimationClipUpgrader.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    class UniversalRenderPipelineAnimationClipUpgrader
+    {
+        [MenuItem("Edit/Rendering/AnimationClips/Convert All AnimationClips to URP Properties")]
+        static void UpgradeAllAnimationClipsMenuItem()
+        {
+            var upgraders = new List<MaterialUpgrader>();
+            UniversalRenderPipelineMaterialUpgrader.GetUpgraders(ref upgraders);
+            AnimationClipUpgrader.DoUpgradeAllClipsMenuItem(upgraders);
+        }
```
phi-lira, line 13: Instead of adding a menu we should integrate this to the material upgrade as discussed.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
phi-lira: If I understand correctly this change will not be needed anymore as the plan is to integrate the AnimationClipUpgrader with the existing material upgrader.

com.unity.render-pipelines.core/Editor/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")] // permit using internal interfaces with Moq
+[assembly: InternalsVisibleTo("Unity.RenderPipelines.Core.Editor.Tests")]
+[assembly: InternalsVisibleTo("Unity.RenderPipelines.Universal.Editor")] // remove as needed when API can be made public
+[assembly: InternalsVisibleTo("Unity.RenderPipelines.UniversalUpgradeProjectTests")]
```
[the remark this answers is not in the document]
phi-lira: Imo we should avoid doing this. If AnimationClip API is not public and for now the scope is only to be used for URP, one possibility could be to make it internal  to `Universal.Editor` assembly and once by the time we look at it for HDRP we move to core and make API public.

Also, regarding the UniversalUpgradeProjectTest, how does that change based on the fact that we decided to integrate the animation clip upgrader with existing material upgrader?

phi-lira (changes requested):
It would be nice to get an update to integrate the work we discussed to move the AnimationClipUpgrader to the existing MaterialUpgrader. I'm not sure how much that will change /solve existing discussion points. 
 
 I'm a little bit worried about the fact that we add internals of core visible to several other assemblies. Would like to discuss possible approaches to that. 
 
 I think also it would be nice to involve SDET to get some review on this. I particularly don't know where we draw the line of tests in a package and tests in a project. We are adding a test package for URP, so I'm also interested to know how get some guidelines regarding testing and the package https://github.com/Unity-Technologies/Graphics/pull/4182

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAnimationClipUpgrader.cs
amechtley: Done in https://github.com/Unity-Technologies/Graphics/pull/4135/commits/5aa6524419a08dd468f9507f56f10ce9c2aea806. To keep changes minimal for now, it is still a second dialog prompt that appears, but as part of the existing menu command.

com.unity.render-pipelines.core/Editor/AssemblyInfo.cs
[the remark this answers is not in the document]
amechtley: Changed in https://github.com/Unity-Technologies/Graphics/pull/4135/commits/5aa6524419a08dd468f9507f56f10ce9c2aea806

For now, I have given the test project visibility of internals for the `Universal.Editor` assembly. What is in there now is not a "true" end-to-end test since we have merged everything into the existing menu commands, but it is tests for the whole AnimationClip stage of the upgrade using actual assets, instead of mocks.

TestProjects/UniversalUpgradeTest/Packages/manifest.json
```diff
@@ -0,0 +1,52 @@
+{
+  "dependencies": {
+    "nuget.moq": "1.0.0",
+    "com.unity.ide.rider": "3.0.4",
+    "com.unity.ide.visualstudio": "2.0.7",
+    "com.unity.ide.vscode": "1.2.3",
+    "com.unity.testing.urp-upgrade": "file:../../../com.unity.testing.urp-upgrade",
```
simonbz, line 7: We should be able to safely remove this test project now if we can adjust the yamato job to to pack all package dependencies. @Unity-Technologies/gfx-sdets Please let us know how this can be done.

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
```diff
@@ -35,6 +35,7 @@ public class MaterialUpgrader
         public delegate void MaterialFinalizer(Material mat);
 
         string m_OldShader;
+        internal string NewShader => m_NewShader;
```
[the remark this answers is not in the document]
phi-lira: This can be made public imo. Please add public API docs.

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
```diff
@@ -57,6 +58,24 @@ class KeywordFloatRename
         }
         List<KeywordFloatRename> m_KeywordFloatRename = new List<KeywordFloatRename>();
 
+        internal enum RenameType
```
phi-lira, line 61: Can be made public imo.

com.unity.render-pipelines.universal/Editor/AssemblyInfo.cs
```diff
@@ -1,5 +1,7 @@
 using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")] // permit using internal interfaces with Moq
```
phi-lira, line 3: What internal API moq needs from URP?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
phi-lira, line 3: Is this needed?

com.unity.render-pipelines.universal/Editor/UpgradeUtility.cs
```diff
@@ -0,0 +1,187 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+using UnityObject = UnityEngine.Object;
+
+namespace UnityEditor.Rendering
```
phi-lira, line 7: Seems like this class would be useful to have in Core, but I guess the reason it's added here is because it's internal. So that's ok for me for now. We can discuss moving this late when we can allocate more time to discuss.

com.unity.render-pipelines.universal/Tests/Editor/Unity.RenderPipelines.Universal.Editor.Tests.asmdef
```diff
@@ -1,9 +1,11 @@
 {
     "name": "Unity.RenderPipelines.Universal.Editor.Tests",
+    "rootNamespace": "",
     "references": [
         "GUID:15fc0a57446b3144c949da3e2b9737a9",
         "GUID:c579267770062bf448e75eb160330b7f",
         "GUID:df380645f10b7bc4b97d4f5eb6303d95",
+        "GUID:3eae0364be2026648bf74846acb8a731",
```
phi-lira, line 8: which assembly reference is being added?

phi-lira (approved):
LGTM. I didn't cover in my review the following:
 - AnimtionClipUpgrader.cs.
 - yml and test package files.
 
 I made minor comments, from my perspective it's ok for PR to land as it is given that other reviewers approve and CI jobs are running correctly.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
simonbz: Removed in 25554645a928746093a54734dbfc3f3143cd0208.

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
simonbz: Changed in 25554645a928746093a54734dbfc3f3143cd0208.

com.unity.render-pipelines.universal/Tests/Editor/Unity.RenderPipelines.Universal.Editor.Tests.asmdef
simonbz: Removed in b139c5ffb592c82cc87c38a42b622816ade62209.

com.unity.render-pipelines.universal/Editor/AssemblyInfo.cs
simonbz: `AnimationClipUpgrader.IAnimationClip` and `UpgradeUtility.IMaterial`. There might be more if we wish to add new mockable interfaces to the test package in the future.

amechtley (approved):

Verasl (changes requested):
Tried a fairly basic test scene, was the first setup and the system failed to upgrade anything, seems to not work if any of the materials affected by the animation clip are not upgraded by URP. So for example if you have a full character using the standard shader, but there is a particle effect that uses unlit(so not upgraded) then none of the materials upgrade even though they might need to.
Here is the testing project: 
[AnimationClipUpgrader.zip](https://github.com/Unity-Technologies/Graphics/files/6415775/AnimationClipUpgrader.zip)

simonbz:
> Tried a fairly basic test scene, was the first setup and the system failed to upgrade anything, seems to not work if any of the materials affected by the animation clip are not upgraded by URP. So for example if you have a full character using the standard shader, but there is a particle effect that uses unlit(so not upgraded) then none of the materials upgrade even though they might need to.
> Here is the testing project:
> [AnimationClipUpgrader.zip](https://github.com/Unity-Technologies/Graphics/files/6415775/AnimationClipUpgrader.zip)

I had a look at the project. This is a limitation of the AnimationClipUpgrader. We are expecting a single upgrade path for materials when we are converting animation clips, therefore, we could not upgrade this clip without breaking legacy materials. There is a warning in the console related to that:

`No clips were upgraded. Did you remember to upgrade materials first?`

Maybe this should be a bit more verbose though.

Will particle materials also be upgraded in the future, or do we need to keep a fallback to builtin materials for certain features? Is this a use case that should be handled by the animation clip upgrader?

simonbz:
@Verasl I updated AnimationClipUpgrader to take into account non-upgradable properties in the animation clip. I think this should fix your use case :)

com.unity.render-pipelines.universal/Editor/AnimationClipUpgrader.cs
```diff
@@ -447,15 +447,29 @@ IEnumerable<ClipPath> clipPaths
...
+                    if (usage == SerializedShaderPropertyUsage.UsedByUpgraded)
+                    {
+                        if (data.PropertyRenames.TryGetValue(binding, out var propertyRename))
+                        {
+                            if (propertyRename != newPropertyName)
+                            {
+                                usage |= SerializedShaderPropertyUsage.UsedByAmbiguouslyUpgraded;
+                            }
+                        }
+                    }
```
amechtley, line 469: Minor nit, but I suppose this could be folded into a single conditional, right?
```suggestion
                    if (
                        usage == SerializedShaderPropertyUsage.UsedByUpgraded
                        && data.PropertyRenames.TryGetValue(binding, out var propertyRename)
                        && propertyRename != newPropertyName
                    )
                        usage |= SerializedShaderPropertyUsage.UsedByAmbiguouslyUpgraded;
```

amechtley (approved):
Latest changes lgtm, thanks!

com.unity.render-pipelines.universal/Editor/AnimationClipUpgrader.cs
simonbz: Works for me 👍

com.unity.render-pipelines.core/Editor/AssemblyInfo.cs
```diff
@@ -0,0 +1,5 @@
+using System.Runtime.CompilerServices;
+
+// remove as needed when/if MaterialUpgrader.RenameType can be made public
+[assembly: InternalsVisibleTo("Unity.RenderPipelines.Universal.Editor")]
+[assembly: InternalsVisibleTo("Unity.Testing.SRP.Universal-Upgrade.Editor")]
```
phi-lira, line 5: Could this be remove now since `RenameType` and `NewShader` are public?

phi-lira (approved):

Verasl:
> Will particle materials also be upgraded in the future, or do we need to keep a fallback to builtin materials for certain features? Is this a use case that should be handled by the animation clip upgrader?

@simonbz eventually everything will be converted, but this will come later in 2022, but I guess we can revisit this when we get there, also we will keep you updated when we decide to add any new upgraders.

Verasl (approved):
LGTM

New changes solved the mixed material upgrade path which I think is very valuable and a likely use-case in unity projects.

com.unity.render-pipelines.universal/Editor/AnimationClipUpgrader.cs
```diff
@@ -0,0 +1,645 @@
...
+                .Select(p => (PrefabPath)AssetDatabase.GUIDToAssetPath(p))
+                .ToArray();
+            var scenePaths = selection == k_PrefabsOnly
+                ? Array.Empty<ScenePath>()
+                : AssetDatabase.FindAssets("t:Scene")
+                    .Select(p => (ScenePath)AssetDatabase.GUIDToAssetPath(p))
+                    .ToArray();
+
+            // TODO: Make following display progress bar?
+
```
martint-unity, line 590: YEs I think we would need a progress bar here.

martint-unity (commented):
I have a few comments. Otherwise lgtm

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
```diff
@@ -37,6 +37,14 @@ public class MaterialUpgrader
         string m_OldShader;
         string m_NewShader;
 
+        /// <summary>
+        /// Retrieves path to new shader.
+        /// </summary>
+        public string NewShader
```
phi-lira, line 43:
```suggestion
        public string NewShaderPath
``` 
?

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
```diff
@@ -57,6 +65,36 @@ class KeywordFloatRename
         }
         List<KeywordFloatRename> m_KeywordFloatRename = new List<KeywordFloatRename>();
 
+        /// <summary>
+        /// Type of property to rename.
+        /// </summary>
+        public enum RenameType
```
phi-lira, line 71:
```suggestion
        public enum MaterialPropertyType
```
?

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
```diff
@@ -57,6 +65,36 @@ class KeywordFloatRename
...
+            Color
+        }
+
+        /// <summary>
+        /// Retrieves a collection of renamed parameters of a specific RenameType.
+        /// </summary>
+        /// <param name="type">Rename Type</param>
+        /// <returns>Dictionary of property names to their renamed values.</returns>
+        /// <exception cref="ArgumentException">type is not valid.</exception>
+        public IReadOnlyDictionary<string, string> GetRename(RenameType type)
```
phi-lira, line 87:
```suggestion
        public IReadOnlyDictionary<string, string> GetPropertyNameMap(RenameType type)
```

com.unity.render-pipelines.universal/Editor/AnimationClipUpgrader.cs
simonbz: done in fc2fded335cbb3628d6276c97752f231af3e6166.

com.unity.render-pipelines.universal/Editor/AnimationClipUpgrader.cs
```diff
@@ -0,0 +1,645 @@
...
+            );
+            const int k_PrefabsOnly = 2;
+            const int k_Cancel = 1;
+            var selection = EditorUtility.DisplayDialogComplex(
+                L10n.Tr("Upgrade AnimationClips"),
+                dialogMessage,
+                L10n.Tr("Prefabs and Scenes"),
+                L10n.Tr("Do Not Upgrade AnimationClips"),
+                L10n.Tr("Prefabs Only")
+            );
```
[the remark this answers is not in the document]
simonbz: done in 6a293ccabf681d14d4822416c1401475c8649903

com.unity.render-pipelines.universal/Editor/AnimationClipUpgrader.cs
```diff
@@ -0,0 +1,645 @@
...
+        {
+            // exit early if no clips
+            if (clipPaths?.Length == 0)
+                return;
+
+            // display dialog box
+            var dialogMessage = L10n.Tr(
+                "Upgrading Material curves in AnimationClips assumes you have already upgraded Materials and shaders as needed. " +
+                "It also requires loading assets that use clips to inspect their usage, which can be a slow process. " +
+                "Which asset types do you want to check for usage?"
```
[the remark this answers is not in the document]
simonbz: done in 6a293ccabf681d14d4822416c1401475c8649903.

com.unity.render-pipelines.core/Editor/AssemblyInfo.cs
simonbz: indeed! Removed in c12129437be685e5764c53f50ec86f1ead708bec

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
simonbz: Done in 3fa71b7caa76812686402193099bf796e231609f.

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
simonbz: Done in 3fa71b7caa76812686402193099bf796e231609f.

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
simonbz: I'm not sure I really like that suggested name though. I think `GetRename` is closer to the truth in that instance.
`GetPropertyRenameMap` maybe? Ultimately, I'm not the major stakeholder here, so I will defer to your better judgment :)

phi-lira: I like `GetPropertyRenameMap` better, seems more descriptive.

phi-lira:
PR looks almost ready to be merged. @JessAtUnity @CookieSalad can you give a review pass?

erikabar (approved):
all issues found were addressed and verified, looks good. tested with internal and asset store projects, also tested with shaders that can or cannot be converted to URP, meshes with multiple materials. 

test doc: https://docs.google.com/document/d/1tUMpUBc7Fr5pfh-P2_uQXRko05o2IOv9vq1DYQ1Bf-o/edit#heading=h.wmstw3clirck

com.unity.render-pipelines.core/Editor/MaterialUpgrader.cs
simonbz: Fixed in 3250edaf413e50e2d32a9a32a3f5eae82e4f9c29.

.yamato/config/universal_upgrade.metafile
```diff
@@ -0,0 +1,33 @@
+project:
+  name: urp-upgrade
+  folder: UniversalUpgradeTest
+test_platforms:
+  - type: editmode
+    commands:
+      - [all]: npm install upm-ci-utils@stable -g --registry https://artifactory.prd.cds.internal.unity3d.com/artifactory/api/npm/upm-npm
+      # - [all]: upm-ci package pack --package-path com.unity.testing.urp-upgrade
+      - [all]: upm-ci project pack --project-path TestProjects/UniversalUpgradeTest
```
JessAtUnity, line 9: It might be worth moving this out to a separate pack pipeline so it can be re-used for the publish and promote jobs, but this should be fine for testing

.yamato/config/universal_upgrade.metafile
```diff
@@ -0,0 +1,33 @@
+project:
+  name: urp-upgrade
+  folder: UniversalUpgradeTest
+test_platforms:
+  - type: editmode
+    commands:
+      - [all]: npm install upm-ci-utils@stable -g --registry https://artifactory.prd.cds.internal.unity3d.com/artifactory/api/npm/upm-npm
+      # - [all]: upm-ci package pack --package-path com.unity.testing.urp-upgrade
```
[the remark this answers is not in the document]
simonbz: Yep, we can remove it. Fixed in 0cfc00f4c46048b0d58535c771837bb1bf230ae1.

CookieSalad (approved):
Animation Clips get upgraded just fine, even the ones in Timeline. Should be good to go.

PR #4137, opened 2021-04-08
New functionality that supports compile time branching based on user created dropdowns.

russoh:

### Checklist for PR maker

---
### Purpose of this PR
This PR adds support for branching based on user created dropdowns.

Motivation for work: https://docs.google.com/document/d/1XeUfc2MauivBYx82V68c_u2aQBZJxt2h9VRVu9yemGY/edit

The workflow, and UI for creating dropdowns is very similar to the behavior for working with keyword enums. Dropdowns are a new type of blackboard property, where the user can specify a list of options, and a default. All of the scoping, and keyword specific controls are stripped down, so it is simpler than keywords. The user can drag a dropdown property out onto the graph canvas, which will create a corresponding branch node, with slots for each option.

The key distinction for dropdowns - dropdowns are exposed on the UI of subgraph node instances, where those subgraphs have defined dropdowns. This allows the user to pick the desired branch per subgraph instance from the parent graph. This is all resolved at compile time.

https://jira.unity3d.com/browse/GSG-354

---
### Testing status

Manual testing:

Utilized in Surface Gradient example "Sphere Dropdown And Connection Testing":
https://github.cds.internal.unity3d.com/unity/surfgrad-examples

Automated tests:

Graphics/TestProjects/ShaderGraph/UtilityNodes.unity : Dropdowns
https://jira.unity3d.com/browse/GSG-396

---
### Comments to reviewers

Basic video of this in action:
[SubgraphDropdown.zip](https://github.com/Unity-Technologies/Graphics/files/6282326/SubgraphDropdown.zip)

For testing, use instructions here for getting the compatible build: https://github.cds.internal.unity3d.com/unity/surfgrad-examples/blob/master/README.md

IMPORTANT - you cannot create more than dropdown using the conventional + via the blackboard. If you do this, any secondarily created properties will not be visible, and the graph will break. You MUST make new properties by duplicating the first dropdown using Ctrl-D, until this bug has been addressed. The issue also affects keyword enums, and will be addressed by a separate PR as it is common to both.

I will merge in latest of master periodically to get ready for final merge.

Version numbers of serialized data will need bumping between landing of PRs.

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
@@ -1464,6 +1464,7 @@ namespace UnityEditor.ShaderGraph
                 copyProp.precision = sourceProp.precision;
                 copyProp.overrideHLSLDeclaration = sourceProp.overrideHLSLDeclaration;
                 copyProp.hlslDeclarationOverride = sourceProp.hlslDeclarationOverride;
+                copyProp.useCustomSlotLabel = sourceProp.useCustomSlotLabel;
             }
 
             // sanitize the display name (we let the .Copy() function actually copy the display name over)
--- a/com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
@@ -224,6 +224,9 @@ namespace UnityEditor.ShaderGraph
                 prop.SetupConcretePrecision(this.concretePrecision);
                 var inSlotId = m_PropertyIds[m_PropertyGuids.IndexOf(prop.guid.ToString())];
                 arguments.Add(GetSlotValue(inSlotId, generationMode, prop.concretePrecision));
+
+                if (prop.isConnectionTestable)
+                    arguments.Add(IsSlotConnected(inSlotId) ? "true" : "false");
             }
 
             var dropdowns = asset.dropdowns;
@@ -575,6 +578,23 @@ namespace UnityEditor.ShaderGraph
             {
                 visitor.AddShaderProperty(property);
             }
+        }
+
+        public AbstractShaderProperty GetShaderProperty(int id)
+        {
+            var index = m_PropertyIds.IndexOf(id);
+            if (index >= 0)
+            {
+                var guid = m_PropertyGuids[index];
+                var properties = m_SubGraph.inputs.Where(x => x.guid.ToString().Equals(guid));
+                var count = properties.Count();
+                if (properties.Count() > 0)
+                {
+                    return properties.First();
+                }
+            }
+
+            return null;
         }
 
         public void CollectShaderKeywords(KeywordCollector keywords, GenerationMode generationMode)
--- a/com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
+++ b/com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
@@ -48,6 +48,8 @@ namespace UnityEditor.ShaderGraph.Drawing.Inspector.PropertyDrawers
         // Display Name
         TextField m_DisplayNameField;
 
+        TextField m_CustomSlotLabelField;
+
         // Reference Name
         TextPropertyDrawer m_ReferenceNameDrawer;
         TextField m_ReferenceNameField;
@@ -209,6 +211,57 @@ namespace UnityEditor.ShaderGraph.Drawing.Inspector.PropertyDrawers
                     "Exposed",
                     out var exposedToggleVisualElement));
                 exposedToggle = exposedToggleVisualElement as Toggle;
+            }
+        }
+
+        void BuildCustomBindingField(PropertySheet propertySheet, AbstractShaderProperty property)
+        {
+            if (isSubGraph && property.isCustomSlotAllowed)
+            {
+                var toggleDataPropertyDrawer = new ToggleDataPropertyDrawer();
+                propertySheet.Add(toggleDataPropertyDrawer.CreateGUI(
+                    newValue =>
+                    {
+                        if (property.useCustomSlotLabel == newValue.isOn)
+                            return;
+                        this._preChangeValueCallback("Change Custom Binding");
+                        property.useCustomSlotLabel = newValue.isOn;
+                        graphData.ValidateGraph();
+                        this._postChangeValueCallback(true, ModificationScope.Topological);
+                    },
+                    new ToggleData(property.isConnectionTestable),
+                    "Use Custom Binding",
+                    out var exposedToggleVisualElement));
+                exposedToggleVisualElement.SetEnabled(true);
+
+                if (property.useCustomSlotLabel)
+                {
+                    var textPropertyDrawer = new TextPropertyDrawer();
+                    var guiElement = textPropertyDrawer.CreateGUI(
+                        null,
+                        (string)shaderInput.customSlotLabel,
+                        "Label",
+                        1);
+
+                    m_CustomSlotLabelField = textPropertyDrawer.textField;
+                    m_CustomSlotLabelField.RegisterValueChangedCallback(
+                        evt =>
+                        {
+                            if (evt.newValue != shaderInput.customSlotLabel)
+                            {
+                                this._preChangeValueCallback("Change Custom Binding Label");
+                                shaderInput.customSlotLabel = evt.newValue;
+                                m_CustomSlotLabelField.AddToClassList("modified");
+                                this._postChangeValueCallback(true, ModificationScope.Topological);
+                            }
+                        });
+
+                    if (!string.IsNullOrEmpty(shaderInput.customSlotLabel))
+                        m_CustomSlotLabelField.AddToClassList("modified");
+                    m_CustomSlotLabelField.styleSheets.Add(Resources.Load<StyleSheet>("Styles/CustomSlotLabelField"));
+
+                    propertySheet.Add(guiElement);
+                }
             }
         }
 
@@ -408,6 +461,8 @@ namespace UnityEditor.ShaderGraph.Drawing.Inspector.PropertyDrawers
                     HandleGradientPropertyField(propertySheet, gradientProperty);
                     break;
             }
+
+            BuildCustomBindingField(propertySheet, property);
 
             BuildPrecisionField(propertySheet, property);
 
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/Android/Vulkan/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/Android/Vulkan/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:1748be28c32cf6947b4472fc0a1dc06e9593a2af5157ce15022a5818b6ba6011
-size 6216
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/IPhonePlayer/Metal/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/IPhonePlayer/Metal/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:205146432f93b8afaf4f9e3385da2146ed243e91db3a3dbfbe04ed2cbbb84c6e
-size 6913
+oid sha256:81d37406ab880ae9971b88e9d06b34943f4fafb422b249fca95cd75d18ae106c
+size 6579
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/LinuxEditor/OpenGLCore/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/LinuxEditor/OpenGLCore/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:133d42616a314183f8caea56680abe9a8a8c2523cd26ee701d5f9c40e6a26ae4
-size 6741
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:133d42616a314183f8caea56680abe9a8a8c2523cd26ee701d5f9c40e6a26ae4
-size 6741
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/LinuxPlayer/OpenGLCore/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/LinuxPlayer/OpenGLCore/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:133d42616a314183f8caea56680abe9a8a8c2523cd26ee701d5f9c40e6a26ae4
-size 6741
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/LinuxPlayer/Vulkan/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/LinuxPlayer/Vulkan/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:133d42616a314183f8caea56680abe9a8a8c2523cd26ee701d5f9c40e6a26ae4
-size 6741
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:733a3c6465c6d9bad59008b76393c82375a49c1a1302ceba2b399234b91a23d6
-size 6673
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/OSXEditor/OpenGLCore/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/OSXEditor/OpenGLCore/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:21d92a6c0c5467295b3fa529272bb21a10fdcbc3ad14edf9dcb3e8e605500e3d
-size 7217
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/OSXPlayer/OpenGLCore/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/OSXPlayer/OpenGLCore/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:21d92a6c0c5467295b3fa529272bb21a10fdcbc3ad14edf9dcb3e8e605500e3d
-size 7217
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/LogicNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/LogicNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:133d42616a314183f8caea56680abe9a8a8c2523cd26ee701d5f9c40e6a26ae4
-size 6741
+oid sha256:9429d725d8d33a8a164c1cd8eac83155ba13321bc6df8531a030cfc174d0fdac
+size 6454
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/LogicNodes.png.meta
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/LogicNodes.png.meta
@@ -1,9 +1,9 @@
 fileFormatVersion: 2
-guid: f17cc90256b0bab47b5fef6fbb0ceb38
+guid: 5de580cb8fbcec84592d83cecce9122f
 TextureImporter:
   internalIDToNameTable: []
   externalObjects: {}
-  serializedVersion: 10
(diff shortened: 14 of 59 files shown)


LandonTownsendUnity:
Test doc: https://docs.google.com/document/d/1-tZ5T0KH_u2ZZE4Fxd8B7bs7QghTkHlj3tAa3e9eSb0/edit?usp=sharing

joshua-davis (commented):
One other small note: When you re-order the items in the keywords the Default value seems to point to whatever's at the index, not the item by name. That is if you have:
{ItemA, ItemB, ItemC} and the default is currently ItemB and you then reorder to {ItemA, ItemC, ItemB} then the default will now be ItemC (both index 1). Is this intended? Not a big deal either way, just something I noticed.

com.unity.shadergraph/Editor/Data/Graphs/ShaderDropdown.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public override bool allowedInMainGraph { get => false; }
+
+        [SerializeField]
+        private int m_Value;
+
+        public int value
+        {
+            get => m_Value;
+            set => m_Value = value;
```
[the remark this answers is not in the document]
russoh: I noticed the reordering thing, same thing happens for keyword enums, so I just decided to leave it as-as, since it's super easy to fix the default right there when you're changing the ordering. The one thing I **did** specifically address is reordering items affecting the selected dropdown in parent graphs. I don't store by index like enums do, so dependent graphs should not be affected by a simple reordering of the items, like they do for materials.

Regarding m_Value being out of range, I'll have another look today and see if that's a scenario that could occur when indexing into the list, and fix it as needed.

com.unity.shadergraph/Editor/Data/Nodes/Utility/DropdownNode.cs
```diff
@@ -0,0 +1,171 @@
...
+                sb.AppendLine(string.Format($"{GetVariableNameForSlot(OutputSlotId)} = {value};"));
+            }
+            else
+            {
+                // Iterate all entries in the dropdown
+                for (int i = 0; i < dropdown.entries.Count; i++)
+                {
+                    if (i == 0)
+                    {
+                        sb.AppendLine(string.Format($"{outputSlot.concreteValueType.ToShaderString()} {GetVariableNameForSlot(OutputSlotId)};"));
```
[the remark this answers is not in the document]
russoh: It's probably a good idea to have an else block with a default, I'll do this, good suggestion.

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -220,6 +226,16 @@ public void GenerateNodeCode(ShaderStringBuilder sb, GenerationMode generationMo
...
             }
 
+            var dropdowns = asset.dropdowns;
+            foreach (var dropdown in dropdowns)
+            {
+                var name = GetDropdownEntryName(dropdown.referenceName);
+                if (dropdown.ContainsEntry(name))
+                    arguments.Add(dropdown.IndexOf(name).ToString());
+                else
+                    arguments.Add(dropdown.value.ToString());
```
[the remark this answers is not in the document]
russoh: This would happen if you had a dropdown selection in a parent graph, and then deleted the entry from the subgraph.

E.g. 
Subgraph dropdown
A, B, C

Parent graph picks B, and serialize.

Now delete B from subgraph, and save asset.

Now parent graph selection now longer exists, so just goes back to default value of dropdown.

com.unity.shadergraph/Editor/Data/Graphs/ShaderDropdown.cs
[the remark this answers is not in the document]
russoh: Where do you think this might need clamping? I'm not sure I see a failure point for this.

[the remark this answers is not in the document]
russoh: By the way, I added a fix for reordering thing, so prevent the default value changing.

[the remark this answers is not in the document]
joshua-davis:
```
public string entryName
{
    get => entries[value].displayName;
}
```
uses value unguarded so if someone ever sets value to something invalid it will break. I'm not saying this does happen currently, but this is an api that allows things to break.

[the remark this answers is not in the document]
russoh: Alright, I'll put a clamp in there to be sure.

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
```diff
@@ -1450,10 +1457,203 @@ string GetSanitizedReferenceName(string name)
...
+            });
+
+            AddPropertyRowToSheet(propertySheet, field, "Default");
+
+            var container = new IMGUIContainer(() => OnDropdownGUIHandler()) { name = "ListContainer" };
+            AddPropertyRowToSheet(propertySheet, container, "Entries");
+            container.SetEnabled(true);
+        }
+
+        void OnDropdownGUIHandler()
```
Nightmask3, line 1500: image

The name of the re-orderable list seems to be uninitialized for some reason 🤔

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
```diff
@@ -1450,10 +1457,203 @@ string GetSanitizedReferenceName(string name)
...
+        }
+
+        void DropdownSelectEntry(ReorderableList list)
+        {
+            m_DropdownSelectedIndex = list.index;
+        }
+
+        int GetFirstUnusedDropdownID()
+        {
+            if (!(shaderInput is ShaderDropdown dropdown))
```
Nightmask3, line 1589: This check might be redundant given that you also check for if this input is a dropdown prior to the IMGUI code that calls this and also in the other place you call this function (DropdownAddEntry)

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
```diff
@@ -1450,10 +1457,203 @@ string GetSanitizedReferenceName(string name)
...
+
+            // Update GUI
+            this._postChangeValueCallback(true);
+            this._dropdownChangedCallback();
+            m_DropdownSelectedIndex = list.list.Count - 1;
+        }
+
+        void DropdownRemoveEntry(ReorderableList list)
+        {
+            if (!(shaderInput is ShaderDropdown dropdown))
```
Nightmask3, line 1628: Hmm in fact given that you check for whether the input is a dropdown at the entry point to all of this code (i.e. BuildDropdownFields) all these checks after that point might be redundant? unless they are also meant to function as null checks in case a dropdown is somehow invalidated suddenly

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -721,14 +738,18 @@ public void ToSubGraph()
 
             // Collect the keyword nodes and get the corresponding keywords
             var keywordNodes = graphView.selection.OfType<IShaderNodeView>().Where(x => (x.node is KeywordNode)).Select(x => ((KeywordNode)x.node).keyword);
+            var dropdownNodes = graphView.selection.OfType<IShaderNodeView>().Where(x => (x.node is DropdownNode)).Select(x => ((DropdownNode)x.node).dropdown);
```
Nightmask3, line 741: Given that dropdowns can only ever be present in a sub graph, could we skip this statement if we're in a parent graph? could save a little bit of execution time on the LINQ expression

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -721,14 +738,18 @@ public void ToSubGraph()
...
+
             var metaKeywords = graphView.graph.keywords.Where(x => keywordNodes.Contains(x));
+            var metaDropdowns = graphView.graph.dropdowns.Where(x => dropdownNodes.Contains(x));
```
Nightmask3, line 744: Same comment as above

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
Nightmask3: Might be a premature optimization if the branch costs more than the search expression itself, so might be better to profile it before saying one way or the other

Nightmask3 (changes requested):

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -2039,6 +2145,21 @@ public override void OnAfterDeserialize(string json)
                         keywordGuidMap[input0.m_Guid.m_GuidSerialized] = keyword;
                     }
 
+                    // Do we even need to worry about this? It is possible that somebody way be using an old version, and later upgrade?
```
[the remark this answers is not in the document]
russoh: Sounds good, will do.

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
russoh: I think this is inside the call to "ToSubGraph". How does that get invoked? I could not find a call to it

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
russoh: Possibly, I just followed the pattern used by keywords, so I assumed it might be needed?

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
russoh: True, unless someone calls it in the wrong context, since there multiple types handled in here. I would prefer to leave this matching the pattern used by keywords to be honest.

russoh: Generally the extra checks are probably overly conservative, but it's trivial code, so I would not be concerned about performance implications.

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
russoh: Are you referring to the "IList"?

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
```diff
@@ -1450,10 +1457,203 @@ string GetSanitizedReferenceName(string name)
...
+        void BuildDropdownField(PropertySheet propertySheet, ShaderDropdown dropdown)
+        {
+            // Clamp value between entry list
+            int value = Mathf.Clamp(dropdown.value, 0, dropdown.entries.Count - 1);
+
+            // Default field
+            var field = new PopupField<string>(dropdown.entries.Select(x => x.displayName).ToList(), value);
+            field.RegisterValueChangedCallback(evt =>
+            {
+                this._preChangeValueCallback("Change Dropdown Value");
```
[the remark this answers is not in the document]
russoh: Yep, you're right, that's missing, I'll fix that.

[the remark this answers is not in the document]
russoh: This should work correctly now.

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
russoh: I fixed this, added the header, and it now says "Entry Name" instead of "IList".

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
russoh: Would prefer to leave this in place out of caution.

com.unity.shadergraph/Editor/Drawing/ViewModels/BlackboardViewModel.cs
```diff
@@ -17,21 +17,25 @@ class BlackboardViewModel : ISGViewModel
         public Dictionary<string, IGraphDataAction> propertyNameToAddActionMap { get; set; }
         public Dictionary<string, IGraphDataAction> defaultKeywordNameToAddActionMap { get; set; }
         public Dictionary<string, IGraphDataAction> builtInKeywordNameToAddActionMap { get; set; }
+        public Dictionary<string, IGraphDataAction> defaultDropdownNameToAddActionMap { get; set; }
```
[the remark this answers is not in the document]
russoh: I changed this to a Tuple.

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
Nightmask3: If you select a group of nodes, right click and then in the "Convert To" menu, you'll find the "To SubGraph" option

com.unity.shadergraph/Editor/Drawing/Inspector/PropertyDrawers/ShaderInputPropertyDrawer.cs
Nightmask3: Yup that was it

Nightmask3 (approved):
One last minor thing to possibly look into is the possibility of skipping a LINQ search expression for dropdowns and meta-dropdowns in the `ToSubGraph` function if we're in a non sub-graph, but apart from that looks good, approving!

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
russoh: Got it, I see now. IMO it makes most sense to leave this generic, and use the logic that already exists to allow/prevent invalid nodes being pasted into the corresponding target graph. I don't think there's value in adding extra complexity to the code here. It's a one time action, and I don't see any performance motivation for doing this.

com.unity.shadergraph/Editor/Data/Nodes/Utility/DropdownNode.cs
```diff
@@ -0,0 +1,185 @@
...
+                        sb.AppendLine($"if ({m_Dropdown.value.referenceName} == {i})");
+                    }
+                    else
+                    {
+                        sb.AppendLine($"else if ({m_Dropdown.value.referenceName} == {i})");
+                    }
+
+                    {
+                        sb.AppendLine("{");
```
joshua-davis, line 144: nit: you could use sb.BlockScope here to clean this up a bit. In particular this becomes:
```
using(sb.BlockScope())
{
    var value = GetSlotValue(GetSlotIdForPermutation(new KeyValuePair<ShaderDropdown, int>(dropdown, i)), generationMode);
    sb.AppendLine(string.Format($"{GetVariableNameForSlot(OutputSlotId)} = {value};"));
}
```
Something I just noticed that I didn't before.

joshua-davis (approved):
Small nit, but otherwise looks good

LandonTownsendUnity (approved):
All issues found during my testing were addressed.

russoh:
This one has already had a full green test panel on Shader Graph ABV with the changes specific to this branch:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Fsurfgrad%252Fdropdowns/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/6477815/job

russoh:
ScriptableRenderPipelinePrivate console passes for this PR: https://yamato.cds.internal.unity3d.com/jobs/339-ScriptableRenderPipelinePrivate/tree/sg%252Fsurfgrad%252Fdropdowns/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/6557856/job/pipeline

PR to land changes to ScriptableRenderPipelinePrivate: https://github.cds.internal.unity3d.com/unity/ScriptableRenderPipelinePrivate/pull/168

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -1367,16 +1398,23 @@ public string SanitizeGraphInputReferenceName(ShaderInput input, string desiredN
             {
                 case AbstractShaderProperty property:
                 {
-                    // must deduplicate ref names against both keyword and properties, as they occupy the same name space
-                    var existingNames = properties.Where(p => p != property).Select(p => p.referenceName).Union(keywords.Select(p => p.referenceName));
+                    // must deduplicate ref names against keywords, dropdowns, and properties, as they occupy the same name space
+                    var existingNames = properties.Where(p => p != property).Select(p => p.referenceName).Union(keywords.Select(p => p.referenceName)).Union(dropdowns.Select(p => p.referenceName));
```
cdxntchou, line 1402: yikes.  we need a better way to do unique names in the future.  :)

com.unity.shadergraph/Editor/Data/Legacy/AbstractMaterialNode0.cs
```diff
@@ -15,6 +15,8 @@ class AbstractMaterialNode0
 
         public string m_KeywordGuidSerialized;
 
+        public string m_DropdownGuidSerialized;
```
cdxntchou, line 18: Why does this need to be added to the legacy struct?  This should only be for upgrading from old graphs, which should never have the dropdown...

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -2063,6 +2184,11 @@ public override void OnAfterDeserialize(string json)
                             keywordField.SetValue(node, (JsonRef<ShaderKeyword>)keyword);
                         }
 
+                        if (!string.IsNullOrEmpty(node0.m_DropdownGuidSerialized) && dropdownGuidMap.TryGetValue(node0.m_DropdownGuidSerialized, out var dropdown))
```
cdxntchou, line 2187: shouldn't need the changes in this file -- legacy nodes should never have dropdowns to upgrade?

cdxntchou (changes requested):
Should probably remove the unnecessary versioning changes.   The rest looks good!

com.unity.shadergraph/Editor/Data/Graphs/DynamicVectorMaterialSlot.cs
```diff
@@ -21,7 +21,7 @@ class DynamicVectorMaterialSlot : MaterialSlot, IMaterialSlotHasValue<Vector4>
 
         static readonly string[] k_Labels = {"X", "Y", "Z", "W"};
 
-        private ConcreteSlotValueType m_ConcreteValueType = ConcreteSlotValueType.Vector4;
```
elizabeth-legros, line 24: Could you explain why this change was needed?

elizabeth-legros: Also, will this give any weird behavior change to our math nodes and the like?

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -1238,6 +1244,16 @@ public void CollectShaderKeywords(KeywordCollector collector, GenerationMode gen
             collector.CalculateKeywordPermutations();
         }
 
+        public bool IsInputAllowedInGraph(ShaderInput input)
+        {
+            return (isSubGraph && input.allowedInSubGraph) || (!isSubGraph && input.allowedInMainGraph);
+        }
+
+        public bool IsInputAllowedInGraph(AbstractMaterialNode node)
```
elizabeth-legros, line 1252: Is this input based or node based? Confused by function signature

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -1562,6 +1624,22 @@ public void OnKeywordChangedNoValidate()
...
 
+        public void OnDropdownChanged()
+        {
+            OnDropdownChangedNoValidate();
+            ValidateGraph();
+        }
+
+        public void OnDropdownChangedNoValidate()
+        {
+            var allNodes = GetNodes<AbstractMaterialNode>();
```
elizabeth-legros, line 1635: Validate graph should do this already, may be redundant

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -2039,6 +2146,20 @@ public override void OnAfterDeserialize(string json)
                         keywordGuidMap[input0.m_Guid.m_GuidSerialized] = keyword;
                     }
 
+                    foreach (var serializedDropdown in graphData0.m_SerializedDropdowns)
```
elizabeth-legros, line 2149: Do we need to handle dropdowns in old master node graphs coming in? Are we backporting dropdowns to master nodes?

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
russoh: Actually there's a function for both - there are guards on properties (for stuff copied from the blackboard between graphs), and for nodes (e.g. the dropdown that is related to the property).

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
russoh: I think in most of those cases, I probably just followed the template used by keywords, but this is possible.

com.unity.shadergraph/Editor/Data/Legacy/AbstractMaterialNode0.cs
elizabeth-legros, line 18: again, will dropdowns be ported to old master nodes? If no, not needed

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
russoh: This is a good question, should definitely get clarity on that flow.

com.unity.shadergraph/Editor/Data/Legacy/GraphData0.cs
```diff
@@ -17,6 +17,8 @@ class GraphData0
 
         public List<SerializationHelper.JSONSerializedElement> m_SerializedKeywords;
 
+        public List<SerializationHelper.JSONSerializedElement> m_SerializedDropdowns;
```
elizabeth-legros, line 20: if not backporting, unnecessary

com.unity.shadergraph/Editor/Data/Graphs/DynamicVectorMaterialSlot.cs
russoh: The slot was behaving inconsistently. Initially it would show a 4 channel initializer (0,0,0,0). If you then connected to it (any node), and then disconnected again, the slot would turn into a single channel initializer. With this change it always shows a default single channel initializer, so you don't have that disparity.
Also worth noting that other nodes that accept a dynamic vector, do already use a single channel initializer, so this was notably the odd one out.

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
elizabeth-legros: If we are not (which I would hope at this point tbh), then all the graphdata0 or abstractmaterialnode0 stuff should be unnecessary

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
elizabeth-legros: eh, Id say leave it. We need to fix our validation/concretization/etc to be smarter anyways.

elizabeth-legros (approved):

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
russoh: Yes, this is a bit rough! A helper function would simplify this for sure.

com.unity.shadergraph/Editor/Data/Legacy/AbstractMaterialNode0.cs
russoh: I've removed this one.

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
russoh: I've removed this one.

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
russoh: Cheers, I have removed this now.

com.unity.shadergraph/Editor/Data/Legacy/AbstractMaterialNode0.cs
russoh: Also removed, cheers.

com.unity.shadergraph/Editor/Data/Legacy/GraphData0.cs
russoh: Removed this one.

russoh:
All test have now passed. I will only pull to resolve the Changelog conflict.

PR #4141, opened 2021-04-09
Disabling alpha clipping for transparent materials in HDRP Material Samples

remi-chapelain:
### Purpose of this PR
This PR is needed because alpha clipping in those materials prevent raytraced shadows from being colored (for dir light) and semi transparent. Also, this doesn't serve any purpose and doesn't change anything on the rasterization side.

**Before with directional light ray traced shadow**
AlphaClipping_Before

**After with directional light ray traced shadow**
AlphaClipping_After

---
### Testing status
The usual : deleting the sample, and re-importing them to see that the change is here, no new warnings and nothing broken elsewhere

---
### Comments to reviewers
In the changes, you can see the tiling / offset changes for 3 of these materials, this was probably due to a mistake in some PR that is now reverted, not that it changes the render by much. (I checked in 2021.1 and copy pasted the proper tiling/offset from there)


RemyUnity (approved):
I see nice colored shadows, I approve.

SeanPuller (approved):

sebastienlagarde (approved):

PR #4142, opened 2021-04-09
[HDRP][Path Tracing] Force consistent normals to avoid black specular results

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1323455/

Before:
Smooth normals + normal or bump mapping could result in shading normals that end up being inconsistent with the tessellated geometry (for instance, 'N dot V' is negative whereas the polygon is front facing, or specular direction ending up on the wrong side of the primitive). This results in undefined BSDF computations, which we avoid by simply returning null values, and can appear as black edges on coarsely tessellated (yet supposedly smooth) meshes, or when using strong bump/normal mapping.

Now:
The shading normal is now guaranteed to produce a reflection main direction in the right geometric hemisphere (the same as the incoming view vector's). This means that we avoid the undefined situations described above.

To test:
In addition to the test scene in the associated bug report, any scene with coarse tessellations, or strong bump/normal mapping are good candidates to detect any potential new issue.
Whereas this change is needed for specular (or more generally, view-dependant) BSDFs, it shouldn't be required for diffuse BSDFs. For the time being, the same consistent normal is used everywhere, but if needs be, we could use the modified normal for spec, and the original one for diffuse lobes.

EDIT:
Following further tests from @remi-chapelain and myself, the approach has been changed to the separate diffuse/spec normals mentioned previously. This seems to keep the result closer to the original look, while fixing the issues reported in the associated ticket.


TomasKiniulis:
Hey @eturquin, thanks for the PR and nice description. To add on that, would be good if you could list the things you have tested yourself as well, I assume the repro project? Anything else or any other specific cases that you could list?

remi-chapelain (commented):
Did multiple tests on different materials and geometry.

While this fix the black outline issue with the specular contribution, it can reduce substantially the effect of the normal making meshes appear way flater than they look compared to raster since the shading normal is pushed back "closer" to its geometric normal.
This effect is amplified the more the shading normal is facing away from the view vector (grazing angle on a plane for example). 

From here there's 2 options : 
- Push this changes by default (that still corrects a few behaviors) and allow users to disable pathtracer from tampering the normal. 
- Since the issue comes mostly from specular, the idea would be to keep two shading normal, one for diffuse, and one from specular. That way we can only tamper the one we want and have results closer to what someone would expect with the normal map.  

Here's some examples : 
**Materials**
f03186d32ce1c9aa00f837d049eae94a
54f4d8dfc8f93504d95de15322b9e9c1

**Grazing**
c8dbb3fd430e1086bb74387eae07e37a

**Outline**
f3e35f54173fcbcd705ea0c93c3615b0

eturquin:
> Hey @eturquin, thanks for the PR and nice description. To add on that, would be good if you could list the things you have tested yourself as well, I assume the repro project? Anything else or any other specific cases that you could list?

Hey, sorry if the description wasn't more verbose here, it's been discussed in detail with RemyC offline. :)
I tested the 2 main situations where this issue arises:
- poorly tessellated geom supposed to represent a smooth object (like the sphere in the ticket scene)
- various bump/normal mappings applied to Lit and Fabric materials, under different light sources (directional, area, point)

I think that, with Remy many additional tests, this covers well all challenging scenarios.

remi-chapelain (commented):
Did another round of tests. Much better than the previous versions, most of the tests results are expected except one case with white spots with fabric.
- Coarse geometry (when shading normal can  be very different than geometric normal). Can't seem to repro the black outline ✔️
- Normal mapping on plane ✔️ 
9b2399eed6dd46a3b40b6793a4fecc97

- High normal map intensity with a rough normal map. We seem to recover some lost specularity ✔️ 
399852d12575a044dd07c27cc2afb5d1

- Real world example on amalienborg pavved ground ✔️ (it seems to flatten a little bit the normal effect 
205c8ce7ca43e668c9ce2aca2249f932

- Fabric materials ⚠️
23c6aa45789bd93a2413a2ff5ed1f475

remi-chapelain:
Also note for self to update the path tracer normal 5008 test to increase the coverage of it with more cases

remi-chapelain:
Given the number of tests that are already breaking, I didn't spend too much time increasing the coverage for the normal mapping test. Just added a bunch of coarse sphere with different smoothness (test case repro) and also alpha clipped double sided normal. 

5008_PathTracing_NormalMapping

remi-chapelain (approved):

remi-chapelain:
Test is 🟢 on [yamato](https://yamato.cds.internal.unity3d.com/job/6294891) 🥳

sebastienlagarde (approved):

sebastienlagarde:
Note: we will not do backport for this one, too many change

PR #4155, opened 2021-04-12
Fix compilation errors with XR automated tests on some platforms

fabien-unity:
---
### Purpose of this PR
Add missing define guards.

---
### Testing status
Local tests with `xr-reuse-tests`


phi-lira (approved):

phi-lira:
It would be nice to have automated test catching these compilation errors.

theo-pnv (approved):
Tested on my development branch that was blocked by this and now it is resuming the compilation (other unrelated errors). So this PR fixed the problem it intended to. 
Thanks Fabien!

PR #4161, opened 2021-04-13
[Universal] MSAA store actions optimizations for forward renderer Drawobjects pass

manuele-bonanno:
---
### Purpose of this PR
URP bandwidth usage is currently un-optimal on TBDR architectures.
This PR introduces a new "Store Actions" setting that allows us to improve bandwidth usage on certain passes.

When copying depth or color is enabled, the DrawObjects transparent pass will run as a separate pass after the DrawObjects opaque pass. With MSAA enabled, both opaque and transparent passes will StoreAndResolve color and store depth. These store actions are correct for the opaque pass (since its results will be used by copy color/copy depth and by the transparent pass) but a waste of bandwidth in the transparent pass case since, unless some extra passes which use its results were injected, the results of the transparent pass will not be used anymore in the frame.
With this optimization the transparent pass will now Resolve color and discard (DontCare) depth.

Applying this optimization to all possible scenarios, could potentially break any scenes were new passes are injected, as those could potentially try to use the results of the transparent pass, so we are giving multiple "Store Actions" options (in the URP Asset Advanced options).

image

**Discard**: discard results of passes which are not used later in the frame (safe in case no new passes that use those results are injected)
**Store**: always store, safest but unoptimal behaviour (same as current)
**Auto**: use the Discard behaviour if no external injected passes are detected, otherwise fallback to Store

On the low level side, we are adding Color store actions and depth store actions properties to ScriptableRenderPass: this will allow more explicit control of store actions for future optimizations (both using the RenderPass API path and the SetRenderTarget one)

**Note**:
to get fully optimized results on Metal, in terms of resolve store actions being applied correctly, this PR relies on some trunk fixes I made in [this PR](https://ono.unity3d.com/unity/unity/pull-request/123909/_/graphics/metal/deferred_store_actions_fix). 

---
### Testing status

Results when applying the **Discard** policy on iOS:

Opaque pass:
image

Transparent Pass:
image


com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -31,6 +31,7 @@ internal class Styles
             public static GUIContent requireOpaqueTextureText = EditorGUIUtility.TrTextContent("Opaque Texture", "If enabled the pipeline will copy the screen to texture after opaque objects are drawn. For transparent objects this can be bound in shaders as _CameraOpaqueTexture.");
             public static GUIContent opaqueDownsamplingText = EditorGUIUtility.TrTextContent("Opaque Downsampling", "The downsampling method that is used for the opaque texture");
             public static GUIContent supportsTerrainHolesText = EditorGUIUtility.TrTextContent("Terrain Holes", "When disabled, Universal Rendering Pipeline removes all Terrain hole Shader variants when you build for the Unity Player. This decreases build time.");
+            public static GUIContent StoreActionsOptimizationText = EditorGUIUtility.TrTextContent("Store Actions", "Sets the store actions policy on mobile GPUs. 'Discard': discard as much as possible, 'Store': always store, 'Auto': Discard if no custom passes are injected, otherwise Store.");
```
jonuuukas, line 34: should start in lower case as other GUIContent vars

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -229,6 +244,24 @@ public void ConfigureInput(ScriptableRenderPassInput passInput)
             m_Input = passInput;
         }
 
+        public void ConfigureColorStoreAction(uint attachmentIndex, RenderBufferStoreAction storeAction)
+        {
+            m_ColorStoreActions[attachmentIndex] = storeAction;
+        }
+
+        public void ConfigureColorStoreActions(RenderBufferStoreAction[] storeActions, uint attachmentsCount)
```
jonuuukas, line 252: Wouldn't it make more sense to just pass a shorter array in this call? as in, remove attachmentsCount var and just check storeActions.Length?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -229,6 +244,24 @@ public void ConfigureInput(ScriptableRenderPassInput passInput)
             m_Input = passInput;
         }
 
+        public void ConfigureColorStoreAction(uint attachmentIndex, RenderBufferStoreAction storeAction)
```
jonuuukas, line 247: i'd suggest to change this to ConfigureColorStoreAction(RenderBufferStoreAction storeAction, uint attachmentIndex = 0), as most of the cases are going to be single target

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -1241,6 +1265,32 @@ internal static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier c
...
+
+            m_ActiveDepthAttachment = depthAttachment;
+
+            RenderBufferLoadAction colorLoadAction = ((uint)clearFlag & (uint)ClearFlag.Color) != 0 ?
+                RenderBufferLoadAction.DontCare : RenderBufferLoadAction.Load;
+
+            RenderBufferLoadAction depthLoadAction = ((uint)clearFlag & (uint)ClearFlag.Depth) != 0 ?
+                RenderBufferLoadAction.DontCare : RenderBufferLoadAction.Load;
+
+            colorStoreAction = m_UseOptimizedStoreActions ? colorStoreAction : RenderBufferStoreAction.Store;
```
jonuuukas, line 1287: maybe just do `if (m_UseOptimizedStoreActions)`? would be more readable i guess

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -475,21 +475,40 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+            bool requiresDepthCopyPass = !requiresDepthPrepass
+                && renderingData.cameraData.requiresDepthTexture
+                && createDepthTexture;
+
             if (this.actualRenderingMode == RenderingMode.Deferred)
                 EnqueueDeferred(ref renderingData, requiresDepthPrepass, renderPassInputs.requiresNormalsTexture, mainLightShadows, additionalLightShadows);
             else
+            {
+                RenderBufferStoreAction opaquePassColorStoreAction = RenderBufferStoreAction.Store;
+                RenderBufferStoreAction opaquePassDepthStoreAction = RenderBufferStoreAction.Store;
```
jonuuukas, line 489: i'd encapsulate this whole part with `if cameraTargetDescriptor.msaaSamples > 1`, as otherwise nothing changes and that would reduce some clutter.

or even do something like `RenderBufferStoreAction opaquePassColorStoreAction =  ameraTargetDescriptor.msaaSamples > 1 ? ((requiresCopyColorPass || requiresDepthCopyPass) ? RenderBufferStoreAction.StoreAndResolve : RenderBufferStoreAction.Resolve) : RenderBufferStoreAction.Store` which isn't as readable, but reduces boilerplate code

jonuuukas (commented):
overall looks good, wrote down few suggestions for overall readability/usability

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
manuele-bonanno: we discussed to keep this interface as I think more explicit is less error prone

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -76,6 +76,7 @@ internal class Styles
             public static GUIContent mixedLightingSupportLabel = EditorGUIUtility.TrTextContent("Mixed Lighting", "Makes the render pipeline include mixed-lighting Shader Variants in the build.");
             public static GUIContent debugLevel = EditorGUIUtility.TrTextContent("Debug Level", "Controls the level of debug information generated by the render pipeline. When Profiling is selected, the pipeline provides detailed profiling tags.");
             public static GUIContent shaderVariantLogLevel = EditorGUIUtility.TrTextContent("Shader Variant Log Level", "Controls the level logging in of shader variants information is outputted when a build is performed. Information will appear in the Unity console when the build finishes.");
+            public static GUIContent storeActionsOptimizationText = EditorGUIUtility.TrTextContent("Store Actions", "Sets the store actions policy on mobile GPUs. 'Discard': discard as much as possible, 'Store': always store, 'Auto': Discard if no custom passes are injected, otherwise Store.");
```
phi-lira, line 79: should we say tile based GPUs instead of mobile?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -97,6 +98,13 @@ public enum ColorGradingMode
         HighDynamicRange
     }
 
+    public enum StoreActionsOptimization
```
phi-lira, line 101: Needs public API docs.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -523,6 +532,11 @@ public bool supportsTerrainHoles
             get { return m_SupportsTerrainHoles; }
         }
 
+        public StoreActionsOptimization storeActionsOptimization
+        {
+            get { return m_StoreActionsOptimization; }
+        }
```
phi-lira, line 538: Needs public API. Is there any reason to not provide setter as well?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -229,6 +244,25 @@ public void ConfigureInput(ScriptableRenderPassInput passInput)
...
```
phi-lira, line 247:
```suggestion
        public void ConfigureColorStoreAction(RenderBufferStoreAction storeAction, uint attachmentIndex = 0)
```

Suggestion to flip parameter order and then provide default parameter with 0. This would be more robust and allow easily to configure non-MRT case without caring about index. I think this API is also good enough that you don't need the get/setter properties above.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -150,6 +150,16 @@ public RenderTargetIdentifier depthAttachment
             get => m_DepthAttachment;
         }
 
+        public RenderBufferStoreAction colorStoreAction
+        {
+            get => m_ColorStoreActions[0];
+        }
```
phi-lira, line 156: It seems limited API that you can't get the MRT store action.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -229,6 +244,25 @@ public void ConfigureInput(ScriptableRenderPassInput passInput)
             m_Input = passInput;
         }
 
+        public void ConfigureColorStoreAction(uint attachmentIndex, RenderBufferStoreAction storeAction)
+        {
+            m_ColorStoreActions[attachmentIndex] = storeAction;
+        }
+
+        public void ConfigureColorStoreActions(RenderBufferStoreAction[] storeActions)
```
phi-lira, line 252: Missing public API

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -169,6 +179,9 @@ public Color clearColor
             get => m_ClearColor;
         }
 
+        RenderBufferStoreAction[] m_ColorStoreActions = new RenderBufferStoreAction[] { RenderBufferStoreAction.Store };
+        RenderBufferStoreAction m_DepthStoreAction = RenderBufferStoreAction.Store;
```
phi-lira, line 183: I wonder why not backing the entire attachemnt setup with something like https://docs.unity3d.com/ScriptReference/Rendering.RenderTargetBinding.html
 
 I also find it weird that we provide StoreActions but don't expose LoadAction. If we don't think in terms of API completeness I feel like very soon someone might ask for Load actions as well.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -1193,9 +1211,10 @@ void SetRenderPassAttachments(CommandBuffer cmd, ScriptableRenderPass renderPass
                 else
                 {
                     // Only setup render target if current render pass attachments are different from the active ones
-                    if (passColorAttachment != m_ActiveColorAttachments[0] || passDepthAttachment != m_ActiveDepthAttachment || finalClearFlag != ClearFlag.None)
+                    if (passColorAttachment != m_ActiveColorAttachments[0] || passDepthAttachment != m_ActiveDepthAttachment || finalClearFlag != ClearFlag.None ||
+                        renderPass.colorStoreAction != m_ActiveColorStoreActions[0] || renderPass.depthStoreAction != m_ActiveDepthStoreAction)
```
phi-lira, line 1215: Not for this PR, but maybe for the future we could store the current attachment setup with something like https://docs.unity3d.com/ScriptReference/Rendering.RenderTargetBinding.html then it would make code more readable to compare pass binding with active binding.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -1255,6 +1279,37 @@ internal static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier c
...
+        internal static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier colorAttachment, RenderTargetIdentifier depthAttachment, ClearFlag clearFlag, Color clearColor, RenderBufferStoreAction colorStoreAction, RenderBufferStoreAction depthStoreAction)
+        {
+            m_ActiveColorAttachments[0] = colorAttachment;
+            for (int i = 1; i < m_ActiveColorAttachments.Length; ++i)
+                m_ActiveColorAttachments[i] = 0;
+
+            m_ActiveColorStoreActions[0] = colorStoreAction;
+            m_ActiveDepthStoreAction = depthStoreAction;
+            for (int i = 1; i < m_ActiveColorStoreActions.Length; ++i)
+                m_ActiveColorStoreActions[i] = RenderBufferStoreAction.Store;
```
phi-lira, line 1291: Hooking up on the previous idea, if we used RenderTargetBinding we could configure it here once.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -480,21 +480,34 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
             else
+            {
+                RenderBufferStoreAction opaquePassColorStoreAction = RenderBufferStoreAction.Store;
+                if (cameraTargetDescriptor.msaaSamples > 1)
+                    opaquePassColorStoreAction = (requiresCopyColorPass || requiresDepthCopyPass) ? RenderBufferStoreAction.StoreAndResolve : RenderBufferStoreAction.Resolve;
+
+                RenderBufferStoreAction opaquePassDepthStoreAction = (requiresCopyColorPass || requiresDepthCopyPass) ? RenderBufferStoreAction.Store : RenderBufferStoreAction.DontCare;
+
+                m_RenderOpaqueForwardPass.ConfigureColorStoreAction(0, opaquePassColorStoreAction);
+                m_RenderOpaqueForwardPass.ConfigureDepthStoreAction(opaquePassDepthStoreAction);
```
phi-lira, line 500: Some comments here explaining the motivation would be helpful for future maintanance.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -477,6 +488,9 @@ public ScriptableRenderer(ScriptableRendererData data)
             useRenderPassEnabled = data.useNativeRenderPass;
             Clear(CameraRenderType.Base);
             m_ActiveRenderPassQueue.Clear();
+
+            m_StoreActionsOptimizationSetting = UniversalRenderPipeline.asset.storeActionsOptimization;
+            m_UseOptimizedStoreActions = m_StoreActionsOptimizationSetting != StoreActionsOptimization.Store;
```
phi-lira, line 493: StoreActionsOptimization.Store should also remove store optimization from other passes like post-processing imo.

phi-lira (changes requested):
I'm not sure if this setting should be on the renderer or in the pipeline asset.
 @TheoWong-pixel what do you think?
 
 I think from API perspective I get concerned that we expose public API to configure store actions but not to configure load. It feels the API is imcomplete/only being added for this specific case, but my worry is that soon users might find use cases they want to optimize load actions as well. @hkr for awareness/comments.
 
 Otherwise, PR seems mostly good to me, requesting changes and adding additional stakeholders.

phi-lira:
Adding @TheoWong-pixel for awareness of UX.
 Adding @hkr as he worked exposing load/store actions on Unity so he might have some feedback.
 Adding @UnityPaul for discussion about how this affects 2D renderer.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
manuele-bonanno: usually load actions are implicitly configured based on what store actions we have. Maybe I could name this setting differently? "Bandwidth optimization policy" or something like that?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
manuele-bonanno: we are only setting it in the pipeline settings, same as opaqueDownsampling or supportsTerrainHoles which don't have setters as well. So I just copied those. Since it is coming from pipeline settings seems safer to me to have it read only. wdyt?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
manuele-bonanno: sure, @jonuuukas raised the same suggestion. I preferred the explicit interface, but seems like I am outnumbered ;)
I'll change it to use 0 as default

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
manuele-bonanno: agree, will fix

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
manuele-bonanno: in engine load actions are inferred based on the store actions setup, and this change or every API implementation. The only proper way right now to set load/store actions without having to fight built-in logic is to use the RenderPass API.

So while exposing the load actions here is easy, I am not sure they will have much effect in terms of determining the final store action used by the device. So there would be a lot of work needed to allow that. If it is important can be done as a future task, but from my experience the engine does a good job in figuring out the load actions and the main problems were introduced by wrong stores, so not sure the effort to have an explicit Load action API is worth it (also considering RenderPasses will do that)

I agree having only stores exposed in the API makes it feel limited, but if we want load actions too then it might involve lot of extra work. Future PRs could address it if really needed. WDYT?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
manuele-bonanno: not sure I understand what you mean? 
Currently StoreActionsOptimization.Store will remove optimization from all the passes going through the new SetRenderTarget API that explicitely sets the store actions. So any passes using that new interface will be affected. Right now that is only the DrawOpaque and DrawTransparent passes, but it is extensible to other passes in the future including post processing

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
hkr: just for the record, I agree with @manuele-bonanno ;)

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
hkr: This seems to be the only place where `m_ActiveColorStoreActions` is read, but we only read `m_ActiveColorStoreActions[0]`. Is that on purpose?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -1193,9 +1211,10 @@ void SetRenderPassAttachments(CommandBuffer cmd, ScriptableRenderPass renderPass
                 else
                 {
                     // Only setup render target if current render pass attachments are different from the active ones
```
hkr, line 1213: Why are the store actions only used in this else branch?

hkr:
If this only affects Depth and MSAA of opaque and transparent pass then I think it would be better to give it a more explicit name in the UI

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
manuele-bonanno: this is the non MRT path so there should be only one color attachment (the first check that was already there only checks for color[0]:  passColorAttachment != m_ActiveColorAttachments[0])

In general this optimization PR is only for the forward renderer at the moment, so I didn't do anything in MRT/deferred branches

com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
```diff
@@ -421,6 +421,24 @@ public static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier buf
...
+        /// </summary>
+        /// <param name="cmd">CommandBuffer used for rendering commands.</param>
+        /// <param name="colorBuffer">Color buffer RenderTargetIdentifier.</param>
+        /// <param name="colorLoadAction">Color buffer load action.</param>
+        /// <param name="colorStoreAction">Color buffer store action.</param>
+        /// <param name="depthLoadAction">Depth buffer load action.</param>
+        /// <param name="depthStoreAction">Depth buffer store action.</param>
+        /// <param name="clearFlag">If not set to ClearFlag.None, specifies how to clear the render target after setup.</param>
+        /// <param name="clearColor">If applicable, color with which to clear the render texture after setup.</param>
+        public static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier colorBuffer, RenderBufferLoadAction colorLoadAction, RenderBufferStoreAction colorStoreAction,
```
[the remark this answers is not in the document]
manuele-bonanno: this just exposes an existing CoreUtils binding to SRP, because the specific use case that we were missing was having the possibility of setting depth store actions without setting the depth attachment. Before this change any depth actions you would set in URP would be ignored by the engine because the API was just to passing those parameters down to the engine API.
In the only place where we are currently calling this only the clearColor parameter is needed. I can add a clear depth value too just for completeness? But wouldn't be used right now. WDYT?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
manuele-bonanno: same as the previous comment. The other branch is for MRT/deferred

In general this optimization PR is only for the forward renderer at the moment, so I didn't do anything in MRT/deferred branches

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
manuele-bonanno: regarding redundant render target switches: I suppose that could happen potentially. It doesn't happen with the current optimizations I did, but if users want to customize further their store actions in some cases it could lead to unoptimal switches I guess. But not sure there is much we can do about it. If we expose more low level customization options this is one of the risks. more power->more responsibility?

manuele-bonanno: I also noticed that in some cases the engine would still do extra target switches compared to URP SetRenderTarget calls. i.e. is opaque would Store depth, and transparent would discard.
Without the store actions comparison check: 

- Opaque SetRenderTarget is called: Opaque will store all the results
- Transparent SetRenderTarget is called: will load from Opaque, will store the all results (waste of bandwidth)

With this new check introduced the behaviour will be the same in terms of loads but now the transparent pass will discard/resolve everything, no stores

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
phi-lira: Some properties in the pipeline settings have setters, usually the ones that don't is because they affect keyword stripping so we don't allow users to change at runtime.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
phi-lira: Sure we can make it explicit if you prefer. No strong opinion. Can we then flip the order to the index is last?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
phi-lira: Ok. sounds good to me and fair justification.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
manuele-bonanno: Added docs for StoreActionsOptimization and storeActionsOptimization

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
manuele-bonanno: changed it to use the default value + stripped order, no strong preference regarding the default value

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
manuele-bonanno: added more comments explaining the logic

manuele-bonanno:
@hkr regarding this affecting only Opaque and Transparent: right now yes, but the plan is to potentially further optimize more passes in the future if possible. So having a generic UI that doesn't mention Opaque and Transparent explicitly is more future proof.
Basically any pass where we will add any explicit Store action setting using the new API, will be affected by this "Store Actions" option (since the optmization switch is global and done not on a per pass basis)

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
manuele-bonanno: fixed in the latest commits

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
manuele-bonanno: thanks for the clarification, added the setter in the latest commit

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
manuele-bonanno: @phi-lira all code changes are done as discussed (except the ones we agreed shouldn't be changed).
I wait for documentation and UX reviewers feedback for changing UI/docs related stuff

com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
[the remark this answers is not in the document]
hkr: Why do you want to set depth store action when there is no depth attachment?
This only makes sense if colorBuffer is a RenderTargetIdentifier that refers to color and depth. But In that case it should have a different name.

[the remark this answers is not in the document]
manuele-bonanno: the depth attachment is used but being the backbuffer you cannot set it explicitly in URP, if you do it will crash because most devices handle this setup under the hood. You can look at the code calling this API to understand the limitation (SetRenderTarget in ScriptableRenderer.cs). The CoreUtils binding in C++ was already there, I just exposed it so it can be called in ScriptableRenderer.cs

So the only way for you to be able to change depth actions when dealing with the backbuffer is this (before my change depth actions were just ignored and never passed down to the engine)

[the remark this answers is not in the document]
hkr: Then please don't call it `colorBuffer`.
https://docs.unity3d.com/ScriptReference/Rendering.CommandBuffer.SetRenderTarget.html calls it `rt` for this case and in CoreUtils it's `buffer`.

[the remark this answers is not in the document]
manuele-bonanno: makes sense, changed it to buffer

TheoWong-pixel:
Just wanted to menu that this seems to be a very advanced feature that can be quite delicate for newer users if they dont know what they are doing. it will be better once we have the new "advance menus" in URP so we can better non-advanced users away from this settings. I have noted this in the Quality Audit.

TheoWong-pixel (approved):

hkr (approved):

jonuuukas (approved):

phi-lira:
@TheoWong-pixel Qualit Audit will only happen after cutoff and from what I understand this was in the hope to land before. Advice?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
phi-lira: PostProcessingPass.cs has some optimizations to load and store with SetRenderTarget. F.ex, I think we never store depth after post but some users asked for it. I was asking if these settings should affect those optimizations we already have there.

phi-lira:
@TheoWong-pixel what is the advance menu?

phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
manuele-bonanno: yes that sounds like a good idea. This PR is focusing on the opaque/transparent passes on top of introducing the new API. But extending the optimization options to other passes in future PRs sounds trivial (should be a matter of making sure that the new explicit actions API is called by the postproc pass)

TheoWong-pixel:
The work on the advance menu is outside of this PR. It is more structured around the UI and how we show advanced settings to users. This should handled in another PR later.

The location that @manuele-bonanno has placed this setting is fine. :)

aleksandrasdzikia (approved):
Looks good on mobile side. Tested URP Examples and Boat attack. Played around with settings, couldn't find any issues on visual correctness as well as performance. It looks that the feature is disabled by default which is expected. Ran on Mali and Adreno devices, had no issues.

erikabar (approved):
approving, testing was done by mobile QA

oleks-k (approved):
Looks good with the editions!

unity-cchu (approved):

phi-lira:
Failures also reproduced in master / XR related, known issue being fixed.

PR #4168, opened 2021-04-13
Improve probe volume debug

JulienIgnace-Unity:
### Purpose of this PR
The goal of this PR is to improve the Probe Volume debug:
- Remove debug toggles from the Probe Volume Reference component and put them in their own debug panel (no more need to select the ref volume for it to work)
- Debug shader now samples the same structure as the light loop for consistency (data was broken before)
- Debug probe is now available at runtime
- Remove dependence on gizmo rendering for probe debug.

To do that a few things were done:
- Moved debug parameters from the authoring component to the ProbeReferenceVolume instance itself
- Moved debugging code to ProbeRefereneVolume.Debug.cs (partial class)
- Debug resources (mesh/shader) are now pass at init time to the new ProbeReferenceVolume.Initialize functions (because we lack proper Core Resource Asset)
- Remove "default" probe volume gizmo shader (would not work anyway). We'll need to make a new one for URP

Known issue:
- Probe at the boundaries of bricks currently display the global ambient probe (need changes from @FrancescoC-unity from another PR)
- Debug Cells/Bricks is still editor only for now (depends on changes from @antoinel )

---
### Testing status
Tested the new debug panel in various setup to check it was sitll working as intended.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/RenderPipeline.meta
+++ b/com.unity.render-pipelines.core/Runtime/RenderPipeline.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 18ccf971905222545a09cfda33385c7c
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
```diff
@@ -84,18 +84,32 @@ APVConstants LoadAPVConstants( StructuredBuffer<int> index )
 #define APV_USE_BASE_OFFSET
 
 // We split the evaluation in several steps to make variants with different bands easier.
-float3 EvaluateAPVL0(APVResources apvRes, float3 uvw, out float L1Rx)
+float3 EvaluateAPVL0(APVResources apvRes, float3 uvw, out float L1Rx, bool usePointSampling = false)
```
FrancescoC-unity, line 87: This might get compiled to something bad perf wise; I'd write a function just for the debug since there is no situation in runtime in which using point sampling is good. 

This is also because once I do the change with probe position/location in tex, this won't be needed anymore and reverting a self-contained function is easier.

FrancescoC-unity: Putting comment here but of course referring to the whole file changes

alelievr (approved):

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
JulienIgnace-Unity: Are you sure about that? Last time I checked (which was a long time ago I admit :p ) this worked fine since it was always a compiled time constant.
Regardless, as you said we're going to change that again so I can make a specific function instead.

sebastienlagarde (approved):

FrancescoC-unity (approved):

PR #4169, opened 2021-04-13
Fixed a warning about implicit truncation of vector type

joshua-davis:
---
### Purpose of this PR
Fixed a warning on most shader graphs caused by an implicit conversion from vector4 to vector3.

---
### Testing status
Ran the tests locally and verified the shader graph errors disappeared.

---
### Comments to reviewers
I believe this is the only warning we have right now (at least in our tests) that isn't from a cginc file.


jessebarker (approved):
LGTM

PR #4171, opened 2021-04-13
Fixed another edge case where stencil materials are accessed before they are completely loaded.

kaychang-unity:
---
### Purpose of this PR
Fixed https://fogbugz.unity3d.com/f/cases/1327945/.

This is another edge case where stencil material is accessed at initialization time before it has finished loading. When it happens, only the fallback shaders are available, so the pass indices in the shader cannot be resolved (they will all be -1).
The workaround is to test if the pass indices are still -1 when the shaders are actually used and resolve the pass indices again.

Original fix did this only when lights were present in the scene, but the actual test case (012_PBS_EnvironmentBRDF_Spheres_deferred) does not have any lights, so the pass indices were never correctly resolved. This was an issue because SSAO effect requires a custom full-screen pass in some situations, and the shader pass used is located in the stencil material.

I am now running the workaround code earlier in the code.

---
### Testing status
Ran 012_PBS_EnvironmentBRDF_Spheres_deferred on MacOS Editor and checked it fixed the bug.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
+++ b/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
@@ -1078,6 +1078,13 @@ namespace UnityEngine.Rendering.Universal.Internal
 
         internal void ExecuteDeferredPass(ScriptableRenderContext context, ref RenderingData renderingData)
         {
+            // Workaround for bug.
+            // When changing the URP asset settings (ex: shadow cascade resolution), all ScriptableRenderers are recreated but
+            // materials passed in have not finished initializing at that point if they have fallback shader defined. In particular deferred shaders only have 1 pass available,
+            // which prevents from resolving correct pass indices.
+            if (m_StencilDeferredPasses[0] < 0)
+                InitStencilDeferredMaterial();
+
             CommandBuffer cmd = CommandBufferPool.Get();
             using (new ProfilingScope(cmd, m_ProfilingDeferredPass))
             {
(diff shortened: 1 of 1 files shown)


ellioman (approved):

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -1078,6 +1078,13 @@ internal void ClearStencilPartial(CommandBuffer cmd)
 
         internal void ExecuteDeferredPass(ScriptableRenderContext context, ref RenderingData renderingData)
         {
+            // Workaround for bug.
```
phi-lira, line 1081: Do we have this bug tracked somewhere?

phi-lira:
@kaychang-unity do you have an idea of why this issue would only fail locally and not be catched by automation?

simon-engelbrecht-soerensen (approved):

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
kaychang-unity: Not sure if this bug is not by design, as fixing it would require the game to stall in order to finish loading assets?

kaychang-unity:
> @kaychang-unity do you have an idea of why this issue would only fail locally and not be catched by automation?

@ellioman I suppose this is a timing issue.

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
kaychang-unity: I added FB case: https://fogbugz.unity3d.com/f/cases/1329200/

PR #4175, opened 2021-04-14
Fix compression asserts and L2 in Probe volume

FrancescoC-unity:
This PR adds two fixes: 

- Fixes the asserts we constantly get while baking (the compression on L1 is too tight) 
- Fixes L2, before this PR it looks all funky (see pic), now this is fixed. The issue was due to the fact that we were indeed compressing L2 but then pushing forward uncompressed values through the texture. 

image


alelievr (approved):

sebastienlagarde (approved):

sebastienlagarde:
@FrancescoC-unity there is formatting issue in the PR

FrancescoC-unity:
> @FrancescoC-unity there is formatting issue in the PR

Gonna fix up, weird the hook didn't catch those.

FrancescoC-unity:
> > @FrancescoC-unity there is formatting issue in the PR
> 
> Gonna fix up, weird the hook didn't catch those.

I fixed them manually, but the hooks don't work anymore on my end weirdly enough; I am trying to find out why.

PR #4178, opened 2021-04-14
Modify old Probe volume info box + add check on transform with multiple ref volume

FrancescoC-unity:
Couple of small things: 

- Several info boxes still referenced shader config and other obsolete steps required to activate the feature. 

Old:
image

New: 
image

[There is another similar info box changed for the probe volumes themselves] 

- Added a check on the reference volume transform when multiple assets are loaded at the same time. 

image


FrancescoC-unity:
I have same formatter issue here because my hooks got broken, gonna fixup.

alelievr (approved):

sebastienlagarde (approved):

PR #4180, opened 2021-04-14
Hd/update resources assets

RSlysz:
### Purpose of this PR
- Fix issue where the ResourceReloader don't reload null value at render pipeline construction anymore in HDRP.
- Prepared work to have a common ancestor for all Resource asset accross all SRP.
  - Currently it is used to have a shared inspector (all 3 assets kind in HDRP but URP can use it too later)
  - Also, it allow to have one unique EnsureAsset in HDRenderPipelineGlobalSettings in order to be sure all resource asset will always be loaded and checked the same way. (code factorization is done in this PR)
- Fix a wrong path in Reload attribute for **volumetricCloudMapGeneratorCS**
- Renamed all resource class for coherency.
- Fixed documentation link on the 3 resource asset that where either not set or set on wrong documentation page.
- Updated documentation that was not totally accurate on the default resource. It was letting think that resources asset were created in the project.

So the hierarchy is now done this way
image
Note for HDRenderPipelineEditorResources: class is define in Runtime folder but is in #if UNITY_EDITOR scope

---
### Testing status
- Tested that Reload All button appears for all 3 resource asset in HDRP when in devloper mode only
- Tested that Reload All button manage to reload all resource for the 3 asset (just to remember: it nullify everything then call the ResourceReloader)
- Tested that nullifying a resource field, it is not reloaded instantly
- Tested that a null field is reloaded when render pipeline is reconstructed (forced a reconstruction by causing a domain reload)

---
### Comments to reviewers
- For Editor resource class visibility change, it was not documented prior (it is a corner case where package validation tool fail). After discution, it is turned back as an internal class ( https://unity.slack.com/archives/GHD5LADU7/p1618397862288100?thread_ts=1618385826.279900&cid=GHD5LADU7 )
- FB is created for the miss detection in Package Validation https://fogbugz.unity3d.com/f/cases/1329149/


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineRuntimeResources.cs
```diff
@@ -160,7 +160,7 @@ public sealed class ShaderResources
             // Volumetric Clouds
             [Reload("Runtime/Lighting/VolumetricLighting/VolumetricClouds.compute")]
             public ComputeShader volumetricCloudsCS;
-            [Reload("Editor/Lighting/VolumetricLighting/CloudMapGenerator.compute")]
+            [Reload("Editor/Lighting/VolumetricClouds/CloudMapGenerator.compute")]
```
jenniferd-unity, line 163: great catch!

jenniferd-unity (approved):
Just a note about the error message but otherwise great job!

JordanL8 (approved):

JulienIgnace-Unity (approved):

TomasKiniulis (changes requested):
This seems to cause editor to break rendering with this:
`Could not load HDRenderPipelineRuntimeResources.
UnityEngine.StackTraceUtility:ExtractStackTrace () (at C:/buildslave/unity/build/Runtime/Export/Scripting/StackTrace.cs:37)
UnityEngine.DebugLogHandler:LogFormat (UnityEngine.LogType,UnityEngine.Object,string,object[])
UnityEngine.Logger:Log (UnityEngine.LogType,object)
UnityEngine.Debug:Assert (bool,string)
UnityEngine.Rendering.HighDefinition.HDRenderPipelineGlobalSettings:EnsureResources<UnityEngine.Rendering.HighDefinition.HDRenderPipelineRuntimeResources>...`

Also Player Resources and Volume profiles get set to None in HDRP settings. Reproduces after project is reopened.

RSlysz:
Above issue only appears when Runtime resource asset need to be reloaded. This happens on reimporting package like updating to a new version of HDRP.
This because I renamed it and forgot to update the path where we try reloading it. It should be good now.

TomasKiniulis (approved):
Works well now! Upgraded Spaceship, Amalienborg projects, did not run into any issues. Resources are correctly relinked when changed and editor is reopened.

sebastienlagarde (approved):

sebastienlagarde:
@RSlysz the PR cause a lot of GC alloc: https://yamato-artifactviewer.prd.cds.internal.unity3d.com/af6d3fd4-acde-4676-a483-bec17ac0b7fd%2Flogs%2FTestProjects%2FHDRP_Tests%2Ftest-results/TestReportV1.html

PR #4186, opened 2021-04-15
[ShaderGraph] Fix for ObjectField focus and Node Selection "delete" ambiguity

esmelusina:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1313943/

ObjectField focus and Node selections were both capturing the delete hotkey, which should unambiguously favor the ObjectField.

---
### Testing status
Manual testing. We have ObjectFields on various Node property drawers (eg. for blackboard properties or custom function node), and we have some ObjectFields inlined onto a node itself (eg. Texture 2D node). Selecting these Fields and using the "delete" hotkey should no longer delete the corresponding node.


com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -898,6 +898,15 @@ void UnserializeAndPasteImplementation(string operationName, string serializedDa
 
         void DeleteSelectionImplementation(string operationName, GraphView.AskUser askUser)
         {
+            // Selection state of Graph elements and the Focus state of UIElements are not mutually exclusive.
+            // For Hotkeys, askUser should be AskUser mode, which should early out so that the focused Element can win.
+            if (this.focusController.focusedElement != null
+                && focusController.focusedElement is UIElements.ObjectField
```
joshua-davis, line 904: Are there any other elements that could have this same issue? Could the logic perhaps be inverted to "if this is a graph node then delete it" or something to be more failsafe?

esmelusina: The quick; This is a regression hotfix. The real solution comes in 22.1.

The long;
Some context-- GraphView selection state and UIElement focus aren't very aware of each other. In ShaderGraph we also coerce selection state for GraphView when it already has issues with heterogenous selections. The result is that all hotkeys are dual captured by the focused element and the graph selection. This particular command gets triggered internally by GraphView (eg. we don't bind it to deletion directly). Fixing the relationship between GV selection, SG's usage of GV selection, and UITK focus is a bigger problem that we're addressing for 22.1.

Yes. There are others. In all cases for graphElement selections (which this command is for), GraphView should be yielding command capture to the focused UIElement (focus clears on single-selection already). There isn't a circumstance where this deletion command would not be yielding focus.

joshua-davis: I more explicitly meant can the focused element be something other than `UiElements.ObjectField` that we should be yielding focus to, for instance are there other UiElements types? It sounds like you've thought this through already though so I'll be approving.

joshua-davis (approved):

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
esmelusina: Oh!!! Yea-- I see what you mean. I thought you were talking about GraphElements. Wow, my bad.

I kept it scoped to the regression fix because it's a bit of a black hole. Though technically Toggles and EnumFields have captures for delete that could be meaningful when they are focused, but also it's not a very salient command for those in the way that space is. EnumFields also don't even have a UI state for focus and the only toggles that have lingering focus in SG are those on blackboard property node settings. 'Delete' is pretty uncommon for this.

Yea-- I could've probably added those, but I'm gonna stick to the regression scope for now.

LandonTownsendUnity (approved):
Seems to work good!

PR #4191, opened 2021-04-15
Optimize asset loading for Probe Volume (close to 90% cost reduction)

FrancescoC-unity:
This PR optimizes the asset loading for APV, avoiding to doing excessive readbacks and avoiding unnecessary work on very large loops (at least the hottest paths as guided by profiler). 

The win is fairly significant, measured a 87% cost reduction (close to 8x times faster) in the template scene. 

So this is I think as far we can go without having to change algorithmically or drastically how the loading works; the changes here are safe in that we don't change how things actually work. 

The profiling capture done after show that pain point now are just accessing data so there is not much we can do there without changing *what* we load

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
@@ -309,8 +309,9 @@ namespace UnityEngine.Rendering
                     if (hr.min == -1)
                         continue;
 
+                    int indexTrans = TranslateIndex(mx, 0, mz);
 
-                    GetIndexData(ref m_TmpUpdater, 0, base_offset + TranslateIndex(new Vector3Int(mx, 0, mz)), hr.cnt);
+                    GetIndexData(ref m_TmpUpdater, 0, base_offset + indexTrans, hr.cnt);
                     int start = volMin.y - hr.min;
                     int end = Mathf.Min(start + volCellSize, m_IndexDim.y);
                     start = Mathf.Max(start, 0);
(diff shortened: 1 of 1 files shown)


alelievr (approved):

alelievr:
FYI I also found out that a big chunk of loading time is also spent in ProbeGIBaking.DilateInvalidProbes. If think it's possible to greatly increase performances if we change the algorithm of this function.
Btw we probably should also disable the dilation if the option is not enabled in the settings

FrancescoC-unity:
> FYI I also found out that a big chunk of loading time is also spent in ProbeGIBaking.DilateInvalidProbes. If think it's possible to greatly increase performances if we change the algorithm of this function.
> Btw we probably should also disable the dilation if the option is not enabled in the settings

The dilation doesn't happen at loading time though, but when baking is complete. 

It is a problem for sure, but is not part of the loading workflow so I'd like to keep changes separated.

alelievr:
Yeah, I tried adding a bit of code to disable it if it's disabled in the UI and it's an astonishing gain, 
For only 1200 bricks, the OnAdditionalProbesBakeCompleted function went from ~48000ms to ~593ms

image

I'll include this change with my PR

FrancescoC-unity:
> Yeah, I tried adding a bit of code to disable it if it's disabled in the UI and it's an astonishing gain,
> For only 1200 bricks, the OnAdditionalProbesBakeCompleted function went from ~48000ms to ~593ms
> 
> image
> 
> I'll include this change with my PR

Yeah the code is fairly naive from a performance stand of point, it can be optimized.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #4199, opened 2021-04-16
First pass on URP Global Settings with Light Layer Names

jenniferd-unity:
---
### Purpose of this PR
This PR introduces Global Settings to URP. Global Settings are project-wide settings for your Render Pipeline. 
image
They are saved as assets and assigned in the Graphics Settings submenu for URP. They work in a similar fashion as HDRP's.

They currently contain light layer names and allows renaming of light layers as well as Rendering Layers.

Note that a warning message will be displayed if URP is not the currently active pipeline (from Quality Level and/or Graphics Settings' default render pipeline).
image

---
### Testing status
Tried small repro scenes. Editor / Windows.
light-layer-renaming-urpimage

- URP settings menu:
          - New button - resets light layer names + creates asset in the Asset folder
          - Clone button - copies light layer names + creates asset in the folder of the cloned asset
          - Changed Light Layer names and noticed changed in:
          - Light's inspector - Layer Name property
          - Visual Effect component - Rendering Layer Mask properties (full rendering layer names list)
- Editor Workflows:
          - Create > Rendering > URP Global Settings Asset
          - Tested search in Project window: t:UniversalRenderPipelineGlobalSettings
          - Closed and reopen editor - URP Global Settings asset is the right one and changes to light layer names are saved
          - Delete active global settings asset while URP is active:
                      - loads the default path one first
                      - loads the next available one then
                      - create one if none found + issues warning
                      - automatic fail safe only works if:
                                - URP Settings menu is open + active SRP is URP
                                - URP is active SRP and Scene/Game/SG Preview View requires rendering
          - Inspector: multi editing + single editing (with changes reflected in URP Settings menu)
- Build workflow:
            - with URP Global Settings set: builds properly and scene seems to work
            - without: cannot issue build

---
### Comments to reviewers
Jira CrossPipeline: https://jira.unity3d.com/browse/XPIPELINE-95
Follow up from Light Layers PR: https://github.com/Unity-Technologies/Graphics/pull/3677

This should also affect VFX assets, Light Editors.
image


com.unity.render-pipelines.universal/Runtime/UniversalGlobalSettings.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            if (UniversalGlobalSettings.instance)
+                return UniversalGlobalSettings.instance;
+
+            UniversalGlobalSettings assetCreated = null;
+            string path = "Assets/" + folderPath + "/UniversalGlobalSettings.asset";
+            assetCreated = AssetDatabase.LoadAssetAtPath<UniversalGlobalSettings>(path);
+            if (assetCreated == null)
+            {
+                var guidHDGlobalAssets = AssetDatabase.FindAssets("t:UniversalGlobalSettings");
```
phi-lira, line 43:
```suggestion
                var guidURPGlobalAssets = AssetDatabase.FindAssets("t:UniversalGlobalSettings");
```

Nitpicking maybe a typo :P

com.unity.render-pipelines.universal/Runtime/UniversalGlobalSettings.cs
```diff
@@ -0,0 +1,179 @@
...
+                    assetCreated.lightLayerName2 = System.String.Copy(src.lightLayerName2);
+                    assetCreated.lightLayerName3 = System.String.Copy(src.lightLayerName3);
+                    assetCreated.lightLayerName4 = System.String.Copy(src.lightLayerName4);
+                    assetCreated.lightLayerName5 = System.String.Copy(src.lightLayerName5);
+                    assetCreated.lightLayerName6 = System.String.Copy(src.lightLayerName6);
+                    assetCreated.lightLayerName7 = System.String.Copy(src.lightLayerName7);
+                }
+                else
+                {
+                }
```
phi-lira, line 102: How do you initialize layer names in this case? Shouldn't you call ResetLayerNames here?

com.unity.render-pipelines.universal/Runtime/UniversalGlobalSettings.cs
```diff
@@ -0,0 +1,179 @@
...
+                m_LightLayerNames[1] = lightLayerName1;
+                m_LightLayerNames[2] = lightLayerName2;
+                m_LightLayerNames[3] = lightLayerName3;
+                m_LightLayerNames[4] = lightLayerName4;
+                m_LightLayerNames[5] = lightLayerName5;
+                m_LightLayerNames[6] = lightLayerName6;
+                m_LightLayerNames[7] = lightLayerName7;
+
+                return m_LightLayerNames;
+            }
```
phi-lira, line 162: It's not ideal for a getter to mutate state of an object. A better solution imo would be to serialize the light name array and change the editor to display the array property nice.

phi-lira (commented):
LGTM. Minor comments, I added some other URP folks that might be interested in reviewing this.

com.unity.render-pipelines.universal/Runtime/UniversalGlobalSettings.cs
jenniferd-unity: the reason it was done this way is also for when people want to change the layer names by script. they can do it by calling "lightLayerName3" for example but internally we serialize an array. i am open to suggestion though, this was implemented that way in HDRP.

jenniferd-unity (commented):
Also @phi-lira is there any Migration/versionning system in URP?
The one in HDRP is using the migration system

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineGlobalSettings.cs
```diff
@@ -0,0 +1,240 @@
...
+
+                return m_RenderingLayerNames;
+            }
+        }
+        public string[] renderingLayerMaskNames => renderingLayerNames;
+
+        void UpdateRenderingLayerNames()
+        {
+            if (m_RenderingLayerNames == null)
+                m_RenderingLayerNames = new string[32];
```
jenniferd-unity, line 160: Initially they were 24 for URP but that was inconsistent with HDRP and Default convention. Let me know if I should revert back to 24.

jenniferd-unity: @kaychang-unity we agreed to change it to 32 for URP.

kaychang-unity (approved):
Ok!

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineGlobalSettings.cs
```diff
@@ -0,0 +1,240 @@
...
+
+#if UNITY_EDITOR
+        //Making sure there is at least one UniversalGlobalSettings instance in the project
+        internal static UniversalRenderPipelineGlobalSettings Ensure(string folderPath = "", bool canCreateNewAsset = true)
+        {
+            if (UniversalRenderPipelineGlobalSettings.instance)
+                return UniversalRenderPipelineGlobalSettings.instance;
+
+            UniversalRenderPipelineGlobalSettings assetCreated = null;
+            string path = "Assets/" + folderPath + "/UniversalRenderPipelineGlobalSettings.asset";
```
martint-unity, line 62: This is a hard coded string that is sued twice. Maybe save it somewhere?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineGlobalSettings.cs
```diff
@@ -0,0 +1,240 @@
...
+                    assetCreated.lightLayerName0 = System.String.Copy(src.lightLayerName0);
+                    assetCreated.lightLayerName1 = System.String.Copy(src.lightLayerName1);
+                    assetCreated.lightLayerName2 = System.String.Copy(src.lightLayerName2);
+                    assetCreated.lightLayerName3 = System.String.Copy(src.lightLayerName3);
+                    assetCreated.lightLayerName4 = System.String.Copy(src.lightLayerName4);
+                    assetCreated.lightLayerName5 = System.String.Copy(src.lightLayerName5);
+                    assetCreated.lightLayerName6 = System.String.Copy(src.lightLayerName6);
+                    assetCreated.lightLayerName7 = System.String.Copy(src.lightLayerName7);
+                }
+                else
```
martint-unity, line 122: Why is there an empty else here?

martint-unity (changes requested):

com.unity.render-pipelines.universal/Editor/GlobalSettings/UniversalGlobalSettingsCreator.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        static UniversalRenderPipelineGlobalSettings settings;
+        static bool updateGraphicsSettings = false;
+        public static void Clone(UniversalRenderPipelineGlobalSettings src, bool activateAsset)
+        {
+            settings = src;
+            updateGraphicsSettings = activateAsset;
+            var path = "Assets/" + src.name + ".asset";
+
```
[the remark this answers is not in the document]
jenniferd-unity: thanks! will change it :)

com.unity.render-pipelines.universal/Editor/GlobalSettings/UniversalGlobalSettingsEditor.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace UnityEditor.Rendering.Universal
+{
+    [CustomEditor(typeof(UniversalRenderPipelineGlobalSettings))]
+    [CanEditMultipleObjects]
+    sealed class UniversalGlobalSettingsEditor : Editor
+    {
+        SerializedUniversalRenderPipelineGlobalSettings m_SerializedGlobalSettings;
+
+        internal bool largeLabelWidth = true;
```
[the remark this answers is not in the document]
jenniferd-unity: true, it is no longer needed :)

com.unity.render-pipelines.universal/Editor/GlobalSettings/UniversalGlobalSettingsWindow.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        public static readonly CED.IDrawer Inspector;
+
+        static UniversalGlobalSettingsPanelIMGUI()
+        {
+            Inspector = CED.Group(
+                LightLayerNamesSection
+                //,CED.Group((serialized,owner) => EditorGUILayout.Space())
+                //,AnotherSection
```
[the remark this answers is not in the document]
jenniferd-unity: ok!

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineGlobalSettings.cs
jenniferd-unity: i am not sure to understand your comment. what string is used twice exactly?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineGlobalSettings.cs
jenniferd-unity: removed :) it was added and not cleaned up

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineGlobalSettings.cs
martint-unity: In UniversalGlobalSettingsCreator file, there is also this
/UniversalRenderPipelineGlobalSettings.asset
Maybe that name could come from a file instead? So only from one source. I don't know if we have a utils file for these kind of things. (I checked and I couldnt find) But maybe we should? :) 
Maybe in UniversalRenderPipelineAsset?

TheoWong-pixel:
Sorry, we generally have the warnings come after the input. Can we have the warning for "No Universal Render pipe ...." after the input please? 

Thanks

jenniferd-unity:
@TheoWong-pixel could you confirm this is what you want?
image
image

(will do the menu change in a sec)

jenniferd-unity:
@TheoWong-pixel could you confirm this is what you want?
image

TheoWong-pixel:
Can we have 
"URP Global Settings Asset 
--------------------------
HDRP Global Settings Asset "

and both come before "HDRP Diffusion Profile"
image

TheoWong-pixel (approved):

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -135,8 +135,16 @@ public static int maxVisibleAdditionalLights
         private readonly DebugDisplaySettingsUI m_DebugDisplaySettingsUI = new DebugDisplaySettingsUI();
 #endif
 
+        private UniversalRenderPipelineGlobalSettings m_GlobalSettings;
+        public override RenderPipelineGlobalSettings defaultSettings => m_GlobalSettings;
```
phi-lira, line 139: Needs public API docs? maybe inherit docs?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineGlobalSettings.cs
phi-lira: A pattern I see in SRP is to load resource asset by GUId instead of path string. This is safer and don't break on renaming.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -135,8 +135,16 @@ public static int maxVisibleAdditionalLights
...
+
         public UniversalRenderPipeline(UniversalRenderPipelineAsset asset)
         {
+#if UNITY_EDITOR
+            m_GlobalSettings = UniversalRenderPipelineGlobalSettings.Ensure();
+#else
+            m_GlobalSettings = UniversalRenderPipelineGlobalSettings.instance;
+#endif
```
phi-lira, line 147: Do you see in future we could possibly call `Ensure` in other places? Maybe a safer approach is to also declare Ensure in standalone and mode this if UNITY_EDITOR there?

phi-lira (approved):
LGMT. Minor comments.

Jonasmortensen (commented):
Interacting with the Global Settings asset can give unexpected results:
- Removing last global settings asset automatically creates new one
- Setting the URP Settings reference in Project Settings to `None` does nothing
- Removing the asset in use automatically updates the project settings reference to an existing one

Maybe this workflow could be changed to work similarly to lighting settings:
If no asset present, properties obtain default values and are not editable. If users want custom settings, they create an asset. If not, no asset is needed:
image

@TheoWong-pixel any thoughts? Would be nice to only force a settings asset if users actually need to change it.

Jonasmortensen (approved):
Looks good! Only concern is regarding asset workflow mentioned in previous comment but since this PR is URP only it should just be made to work like HDRP which it does

TheoWong-pixel:
> Interacting with the Global Settings asset can give unexpected results:
> 
> * Removing last global settings asset automatically creates new one
> * Setting the URP Settings reference in Project Settings to `None` does nothing
> * Removing the asset in use automatically updates the project settings reference to an existing one

Ah yes I would agree with Jonas on this one, maybe we can grey out the Project settings and implement a warning that there is no global settings asset assigned, i know why this happens since we do not have an area to define the "backup / default" for this input. 

But allowing the Global Settings to be blank (though a better option) might also affect other dependent areas that will also need to be grayed. So this might be an interaction for the next iteration since it requires more consideration.

martint-unity (approved):
LGTM

com.unity.render-pipelines.universal/Editor/GlobalSettings/UniversalGlobalSettingsWindow.cs
```diff
@@ -0,0 +1,211 @@
...
+
+    internal class UniversalGlobalSettingsPanelIMGUI
+    {
+        public class Styles
+        {
+            public const int labelWidth = 220;
+            internal static GUIStyle sectionHeaderStyle = new GUIStyle(EditorStyles.largeLabel) { richText = true, fontSize = 18, fixedHeight = 42 };
+            internal static GUIStyle subSectionHeaderStyle = new GUIStyle(EditorStyles.boldLabel);
+
+            internal static readonly GUIContent lightLayersLabel = EditorGUIUtility.TrTextContent("Light Layers Names (3D)", "When enabled, URP allocates memory for processing Light Layers. For deferred rendering, this allocation includes an extra render target in memory and extra cost.");
```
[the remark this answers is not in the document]
jenniferd-unity: yes i will change that

oleks-k (approved):
Looks good!

phi-lira (approved):

PR #4205, opened 2021-04-16
[XPipeline][URP]Unify the Color Temperature propery with HDRP

alex-vazquez:
---
### Purpose of this PR
[JIRA](https://jira.unity3d.com/browse/XPIPELINE-144)

Unification of the prioperty **Color Temperature** with HDRP

#### Now on URP:
image

#### Before on URP:
image

---
### Testing status
Checking that the color temperature is properly modified

---
### Comments to reviewers
Most of the code was on HDRP side, and was not being used.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
@@ -65,8 +65,6 @@ namespace UnityEditor.Rendering.HighDefinition
 
         public static readonly CED.IDrawer Inspector;
 
-        static Action<GUIContent, SerializedProperty, LightEditor.Settings> SliderWithTexture;
-
         internal static void RegisterEditor(HDLightEditor editor)
         {
             k_AdditionalPropertiesState.RegisterEditor(editor);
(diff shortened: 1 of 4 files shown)


arttu-peltonen (commented):
I think the code looks good, but I was wondering why the HDRP code you deleted was unused? Does it not have the equivalent temperature slider, and if it does, can URP not reuse that same code?

alex-vazquez:
@arttu-peltonen, they are using now `com.unity.render-pipelines.high-definition\Editor\Lighting\LightUnit\LightUnitSlider.cs`. The code there was unused, I think they forgot to delete it when they removed.

We can not use exactly the same code as they have other light units, and they have some icons, and it is custom for HDRP.

For now I think is a good aproach and we can see in the future if we want to include the full feature from HDRP on URP

sebastienlagarde:
I guess Alex is right :)

arttu-peltonen (approved):

TheoWong-pixel (approved):

TheoWong-pixel (approved):

pierre-unity:
Just in case, make sure the default "Filter" color is white. Otherwise, we're letting users shoot themselves in the foot and they'll start fighting the yellowish color with a colder color temperature potentially.

simon-engelbrecht-soerensen (approved):
Appears to be working

TheoWong-pixel (approved):

ellioman (approved):

Nzollner (approved):

alex-vazquez:
@phi-lira , I added what we discussed about setting that URP is using the light temperature.
@simon-engelbrecht-soerensen , check from an empty project that the light temperature can be changed, and that when you create a new light the LIght Appearance is to use the color.

CertifiedRice (approved):

phi-lira (approved):

simon-engelbrecht-soerensen (approved):
Re-approved

PR #4210, opened 2021-04-16
[HDRP] Film grain should not affect the alpha channel

pmavridis:
### Purpose of this PR
Reported in our iternal slack channel:
https://unity.slack.com/archives/C6Y79CZM0/p1618580491031800

The film grain was applied to the alpha channel, creating errors in composition workflows.
We already apply the dithering only in RGB, so it makes sense to do the same for alpha too.

---
### Testing status
I have tested this fix using the graphics compositor in HDRP.
Alpha channel before:
image

Alpha channel after:
image


FrancescoC-unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
At QA: the change is very minimal and behavior well define, merging for package 10.5.0.

PR #4213, opened 2021-04-16
Update default texture compression, lightmap and normalmap encoding

hkr:
---
### Purpose of this PR

Android, iOS, tvOS use ASTC as default texture compression format. The default lightmap encoding is now "Normal" (RGBM) instead of dLDR. The default normal map encoding is now DXT5nm-style, ASTC LA compressed.

---
### Testing status

Only changed the settings in the editor


phi-lira (approved):

phi-lira:
Nice! Feel free to merge yourself after roland approves.

hkr:
@phi-lira Do I need to increase the package version of the template package or does that happen automatically when it's released?

rolandk-unity (approved):
Looks great, thank you!

PR #4214, opened 2021-04-16
Fixed a bug where a node could be both vertex and fragment exclusive …

joshua-davis:
…but the editor let you use the node anyways, causing shader compiler errors. Fixes fogbugz 1316128

### Checklist for PR maker

---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1316128/

---
### Testing status
Made an add node, a sampled texture, and a vertex location node. Checked:
 - [x] Both plugged into add -> can't hook up to any block node
 - [x] Both plugged into add -> can plug into another add node (not hooked up to anything)
 - [x] Only texture plugged into add -> can't hook up to a vertex block node but can to a fragment
 - [x] Only vertex id plugged into add -> can't hook up to a fragment block node but can to a vertex
 - [x] Hook up output to a vertex block node -> can only connect an input to vertex id
 - [x] Hook up an output to a fragment block node -> can only connect an input to the texture sample

---
### Comments to reviewers
The issue was caused by the stage capability check terminating as soon as it found a node that wasn't "All", but this meant it wouldn't handle the none case.

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Implementation/NodeUtils.cs
+++ b/com.unity.shadergraph/Editor/Data/Implementation/NodeUtils.cs
@@ -486,15 +486,12 @@ namespace UnityEditor.Graphing
             while (s_SlotStack.Any())
             {
                 var slot = s_SlotStack.Pop();
-                ShaderStage stage;
-                if (slot.stageCapability.TryGetShaderStage(out stage))
-                {
-                    // Clear any stages from the total capabilities that this slot doesn't support (e.g. if this is vertex, clear pixel)
-                    capabilities &= slot.stageCapability;
-                    // Can early out if we know nothing is compatible, otherwise we have to keep checking everything we can reach.
-                    if (capabilities == ShaderStageCapability.None)
-                        return capabilities;
-                }
+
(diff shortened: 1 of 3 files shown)


com.unity.shadergraph/Editor/Data/Implementation/NodeUtils.cs
```diff
@@ -482,12 +482,16 @@ public static ShaderStageCapability GetEffectiveShaderStageCapability(MaterialSl
             var graph = initialSlot.owner.owner;
             s_SlotStack.Clear();
             s_SlotStack.Push(initialSlot);
+            ShaderStageCapability capabilities = ShaderStageCapability.All;
             while (s_SlotStack.Any())
             {
                 var slot = s_SlotStack.Pop();
-                ShaderStage stage;
-                if (slot.stageCapability.TryGetShaderStage(out stage))
```
joshua-davis, line 489: `TryGetShaderStage` was used to find if the capability was exactly one stage type. Turns out this isn't necessary with the right flag masking which would also handle new future stages.

jessebarker (approved):
LGTM

ghost (approved):
Looks good!

PR #4223, opened 2021-04-19
[HDRP] Renderer list culling

pmavridis:
[Requires merging C++ PR first]

### Purpose of this PR
With this PR the rendering passes that are used when creating a frame in HDRP are automatically adjusted based on the culling results, instead of using a fixed configuration based on user settings.

You can enable the feature in the HDRP default settings:
image

For now the culling is optional, in case we get any regressions that were not found during testing.
The rendering is always using the new RendererList API (even without culling).

---
### Testing status
On the default outdoors scene (no geometry), these are the passes that are automatically culled:
- Forward Depth Prepass (Deferred ForwardOnly)
- Forward (+ Emissive) Opaque
- Transparent Depth Prepass 
- Forward PreRefraction 
- Forward Transparent 
- Forward Error 
- Transparent Depth Postpass
- Color Gaussian MIP Chain  (this was for distortion)
- Accumulate Distortion 
- Apply Distortion 
- After Post-Process Objects

**Performance**
In the above case we save (at least) ~1ms of CPU time (and we also save GPU from the mip-map creation that we skip), but I need to gather more data (using player builds outside of the editor). 

For testing, the render graph viewer also cacn help visualize the culled passes (thanks @JulienIgnace-Unity for the tip):
image
(Taken in the HDRP template when looking at the sky)

---
### Comments to reviewers
Some passes like GBuffer, decals, have side-effects (setting a global constant etc) and I could not automatically cull them. 
For now, we have to manually opt-out of culling with RenderGraphBuilder.AllowRendererListCulling(false).

We can add a function to call for passes that are culled, it is not much work, but also means that we do have to create some of the resources of the pass that we cull (now we don't create anything).

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/8x_ShaderGraph/8101_Opaque/SG_IrideBentNormal.mat
+++ b/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/8x_ShaderGraph/8101_Opaque/SG_IrideBentNormal.mat
@@ -13,6 +13,9 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   version: 12
+  hdPluginSubTargetMaterialVersions:
+    m_Keys: []
+    m_Values: 
 --- !u!21 &2100000
 Material:
   serializedVersion: 6
@@ -39,8 +42,16 @@ Material:
   m_SavedProperties:
     serializedVersion: 3
     m_TexEnvs:
+    - Texture2D_6B0247CB:
+        m_Texture: {fileID: 2800000, guid: 1321f8c7675bef1449d88d773fbee0fa, type: 3}
+        m_Scale: {x: 1, y: 1}
+        m_Offset: {x: 0, y: 0}
     - Texture2D_76EF7348:
         m_Texture: {fileID: 2800000, guid: 1321f8c7675bef1449d88d773fbee0fa, type: 3}
+        m_Scale: {x: 1, y: 1}
+        m_Offset: {x: 0, y: 0}
+    - Texture2D_8323AE77:
+        m_Texture: {fileID: 2800000, guid: 11f3a6ad0f4d5f241b56e0f647f9d275, type: 3}
         m_Scale: {x: 1, y: 1}
         m_Offset: {x: 0, y: 0}
     - Texture2D_BF3EAE1A:
@@ -61,8 +72,11 @@ Material:
         m_Offset: {x: 0, y: 0}
     m_Ints: []
     m_Floats:
+    - Vector1_3FBDE535: 0.95
+    - Vector1_436A8D50: 0.95
     - Vector1_5DAD63CC: 0.95
     - Vector1_9F784B29: 0.35
+    - Vector1_D8A6BDBA: 0.35
     - Vector1_F50A7CEC: 0.95
     - _AddPrecomputedVelocity: 0
     - _AlphaCutoff: 0.5
@@ -72,10 +86,12 @@ Material:
     - _AlphaToMask: 0
     - _AlphaToMaskInspectorValue: 0
     - _BlendMode: 0
+    - _ConservativeDepthOffsetEnable: 0
     - _CullMode: 2
     - _CullModeForward: 2
     - _DepthOffsetEnable: 0
     - _DoubleSidedEnable: 0
+    - _DoubleSidedGIMode: 0
     - _DoubleSidedNormalMode: 0
     - _DstBlend: 0
     - _EnableBlendModePreserveSpecularLighting: 1
--- a/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/8x_ShaderGraph/8104_Unlit/SG_UnlitAdditive.mat
+++ b/TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/8x_ShaderGraph/8104_Unlit/SG_UnlitAdditive.mat
@@ -9,7 +9,7 @@ Material:
   m_PrefabAsset: {fileID: 0}
   m_Name: SG_UnlitAdditive
   m_Shader: {fileID: 4800000, guid: 54af59311deef214dbc5ce76da9bd9c8, type: 3}
-  m_ShaderKeywords: _ALPHATEST_ON _BLENDMODE_ADD _ENABLE_FOG_ON_TRANSPARENT _SURFACE_TYPE_TRANSPARENT
+  m_ShaderKeywords: _ALPHATEST_ON _ENABLE_FOG_ON_TRANSPARENT _SURFACE_TYPE_TRANSPARENT
   m_LightmapFlags: 4
   m_EnableInstancingVariants: 0
   m_DoubleSidedGI: 0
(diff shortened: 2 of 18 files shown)


com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraph.cs
```diff
@@ -1089,9 +1108,50 @@ void UpdateResourceAllocationAndSynchronization()
...
+        {
+            foreach (RendererListHandle handle in rendererLists)
+            {
+                var rendererList = m_Resources.GetRendererList(handle);
+                if (m_RenderGraphContext.renderContext.QueryRendererList(rendererList) == RendererListResult.kRendererListPopulated)
+                {
+                    return false;
+                }
+            }
+            return rendererLists.Count > 0 ? true : false;;
```
sebastienlagarde, line 1123: typo, double ";"

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphDefaultResources.cs
```diff
@@ -32,6 +32,9 @@ public class RenderGraphDefaultResources
         /// <summary>Default white XR 2D texture.</summary>
         public TextureHandle whiteTextureXR { get; private set; }
 
+        /// <summary>Default opaue black XR 2D texture.</summary>
```
sebastienlagarde, line 35: typo opaue

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphDefaultResources.cs
```diff
@@ -56,6 +59,7 @@ internal void InitializeForRendering(RenderGraph renderGraph)
             blackUIntTextureXR = renderGraph.ImportTexture(TextureXR.GetBlackUIntTexture());
             blackTexture3DXR = renderGraph.ImportTexture(TextureXR.GetBlackTexture3D());
             whiteTextureXR = renderGraph.ImportTexture(TextureXR.GetWhiteTexture());
+            opaqueBlackTextureXR = renderGraph.ImportTexture(TextureXR.GetOpaqueBlackTexture());
```
sebastienlagarde, line 62: why "opaqueBlackTextureXR" ? can this be BlackTexture2DXR ?

sebastienlagarde: ah I see that there is blackTextureXR, guess I need to read the whole PR to understand why opaqueBlackTextureXR

pmavridis: In FinalPass.shader, this is how we combine the after post-process RT with the output:
outColor.xyz = afterPostColor.a * outColor.xyz + afterPostColor.xyz;

This means that the "neutral" output for the pass is [0,0,0,1] and not [0,0,0,0].

I would expect blackTextureXR to be [0,0,0,1], since we also have a "clearTextureXR", but it's not :( that's why I introduced the "opaqueBlackTextureXR".

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphDefaultResources.cs
```diff
@@ -32,6 +32,9 @@ public class RenderGraphDefaultResources
         /// <summary>Default white XR 2D texture.</summary>
         public TextureHandle whiteTextureXR { get; private set; }
 
+        /// <summary>Default opaue black XR 2D texture.</summary>
+        public TextureHandle opaqueBlackTextureXR { get; private set; }
```
sebastienlagarde, line 36: coud you comment why you need to create a "opaqueBlackTextureXR" ?

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphDefaultResources.cs
pmavridis: Note that, in other places in Unity black means opaque-black, so I thought about change blackTextureXR to have alpha 1, but I expect this will break many things...

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphDefaultResources.cs
pmavridis: I have added a comment in the code so it will be more clear to people working with the codebase.

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphDefaultResources.cs
sebastienlagarde: > but I expect this will break many things...
ok, got it now. 

currently I don't think changing alpha to 1 for black texture will be harmful.  From what I see, maybe AO or SSGI could be impacted when disabled but I even don't think it have impact. I will suggest let's try it, could you make a separate PR with the change and trigger a yamato just to see what will be the impact? 

Thanks

sebastienlagarde (approved):
just wondering about the "opaque" texture stuff but otherwise change looks good

sebastienlagarde:
please add a changelog

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphDefaultResources.cs
pmavridis: Indeed, changing the alpha to 1 does not look harmful, here is the PR:
https://github.com/Unity-Technologies/Graphics/pull/4253
There are some failed tests on some platforms, but look unrelated (yamato noise).

So I will remove the opaque stuff from here and we should merge the other PR first.

sebastienlagarde (approved):

pmavridis:
Note: Yamato will fail with this PR on Trunk because we need to merge the C++ API first.

sebastienlagarde:
> Note: Yamato will fail with this PR on Trunk because we need to merge the C++ API first.

good point, however in the mean time you can launch with custom revision

JulienIgnace-Unity (approved):
Lgtm. Just make sure that resource lifetime is still correct using the Render Graph Viewer.
Doing various captures with various feature enabled/disabled and with or without pass renderer list culling enabled should be enough to make sure it's ok.

kroatoa (approved):

TomasKiniulis (commented):
Tested HDRP Template, Amalienborg palace and Spaceship demo in Editor and Standalone. Checked projects with Dynamic Render Pass Culling On/Off toggled in Edit mode and Playmode. Looks stable right now, gray view does not occur anymore for low res transparency sorting. No crashes occur from C++ change either. Waiting for Mac Editor build to succeed for final pass.

TomasKiniulis (approved):
Checked Mac Editor and MacStandalone as well with renderer list culling enabled/disabled. Works as expected.

pmavridis:
HDRP tests are green locally (apart from known IES issues), but I see many yamato failures that are not related to this PR.

The branch is a bit old, so I will merge latest master and Trunk on the C++ side and run the tests again.

CertifiedRice (approved):

PR #4224, opened 2021-04-19
Move APV Normal Bias on the volume systems + have probe volume CB and eliminate constants from index buffer.

FrancescoC-unity:
This PR moves normal bias from the profile to the traditional volume system. This has two advantages: 

 - Allow for granularity of control if so desired (there might be some jump in look, but is up to the user to decide) 
 - Moves normal bias from profile where it is the only "runtime tunable" parameter to a place that makes more sense to tune from. This way the profile is now strictly a definition of the shape of the reference volume and doesn't contain a mish mash of things. 

Moreover, with this PR I added a constant buffer for probe volume stuff. This allows us to simplify the code of the index buffer a fair bit and simplifies a lot the update of constant data. As a result the index buffer is now containing just the index and no constants like it used to. 

_Unfortunately_ this means that the user will need to re-author the normal bias previously set in the profile, but it is easy to reset to old value as it can be centralized.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
@@ -122,11 +122,11 @@ bool TryToGetPoolUVW(APVResources apvRes, float3 posWS, float3 normalWS, out flo
 
     // convert to index
     int3 index = centerIS + floor(posRS);
-    index = index % (int3)indexDim;
+    index = index % indexDim;
 
 #ifdef APV_USE_BASE_OFFSET
     // get the y-offset
-    int  yoffset = apvRes.index[index.z * (int)indexDim.x + index.x];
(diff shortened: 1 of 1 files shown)


alelievr (approved):

sebastienlagarde (approved):

sebastienlagarde (approved):

sebastienlagarde:
@FrancescoC-unity I have done two sucessive merge and solve conflict (with other probe volume Pr), I guess I have solved it correctly but better to double check

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ShaderVariablesProbeVolumes.cs
```diff
@@ -0,0 +1,24 @@
+using UnityEngine.Rendering;
+
+namespace UnityEngine.Rendering
```
JulienIgnace-Unity, line 3: Need Experimental here

PR #4231, opened 2021-04-19
Move area light's shadow frustum to match the light's rectangle

robcupisz:
### Purpose of this PR
PR moves area light's shadow frustum to match the rectangle of the area light. It's a much more sensible and intuitive behaviour, doesn't leave a big space near the area light not covered by shadows and is pretty much necessary for area light shadowed volumetrics, which would make those issues even more pronounced.

HDRP behaviour before this fix is to place the apex of the shadowmap projection at the center of the area light. That behaviour might have been accidental even, since there was even a function called GetAreaLightOffsetForShadows(), but its output was effectively discarded (due to a refactor?) and it was somewhat incorrect anyway.

PR also adds the shadow frustum wireframe gizmo. There previously wasn't any, making things doubly confusing.

Before _(frustum wireframe just for illustrative purposes, it wasn't there before this change)_
arealightshadowfrustum_old

After
arealightshadowfrustum

Caveats
- The shadow frustum (with a spherical cap) doesn't exactly match the spherical range gizmo shown when shadows are disabled, since they don't share the center. Neither accurately describes the shape of area light's influence, so I would say that's not a regression and ok for this PR.
- Range can be set below near plane, which inverts the gizmo. That's not very graceful, but the spherical range gizmo shown when shadows are disabled doesn't handle low range values either (starts intersecting with the light itself). The issue is related to the above (refining the volume of influence), and probably beyond this PR?
- Near plane should be able to go to 0, instead of k_MinShadowNearPlane=0.0001, but that would complicate the logic around switching between light types, and would cause the value to reserialize from 0 to 0.0001 (but not back). Same issue already exists for box lights (they're ortho, so near plane should also go to 0), so better to address it a separate PR (and in general is a low priority issue).

---
### Testing status
Lots of graphics tests feature shadows - images needed updating, since the shadow moved somewhat. Performed manual testing and the changes have also been in use in a demo production for a while.

---
### Documentation
No documentation needed updating. There are no area light gizmo screenshots in the docs, and shape/parameter descriptions weren't specific enough to become outdated.

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/1550_3dsMax.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/LinuxEditor/Vulkan/None/1550_3dsMax.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:4429ebbe22f536d4c60383718e7f6eb84ce584bf89b2cdbff6b83da3c848cf25
-size 289
+oid sha256:58a0259a26ee766bb7dd0e9afc7aed97d80bed7185d1d935eafb8628030c4960
+size 171917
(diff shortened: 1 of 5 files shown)


FrancescoC-unity:
I am good with the change for when ready, I think it is something we do want; just give me a ping when you think is ready for final review.

CertifiedRice (approved):

sebastienlagarde:
Any update?

robertcupisz:
Should be ready this week!

sebastienlagarde:
@JarkkoUnity  poking you on this PR for awareness (related to your current work)

JarkkoUnity:
Looks good to me. Have you tested it works as expected with other than 1:1 aspect ratio?

robertcupisz:
Yeap, of course. There was some incorrect dead code related to that. See changes to GetAreaLightOffsetForShadows()

PR #4244, opened 2021-04-20
Sg builtin/unlit alpha

joshua-davis:
---
### Purpose of this PR
Unlit shaders would fail to compile if they used Alpha but alpha clipping wasn't enabled on the target. This was because the Alpha property was only marked as a dependency when alpha clipping was on, but now that this is controlled via the material it may always be needed.

Also added a new test that shows two unlit materials, one with alpha clipping on and one without (on the material, both are off in the shader's target). One of these would fail to compile without the bulit-in target fix.

---
### Testing status
Ran tests and created a new test.

---
### Comments to reviewers


jessebarker (approved):
LGTM

PR #4249, opened 2021-04-20
Allow for auto fit a probe volume around scene contents or current selection. 

FrancescoC-unity:
Add a way to short-cut an auto fit a probe volume around the whole content of the scene or a specific selection. This is a good speed up in terms of workflow so that artist can place quick volumes around specific areas or covering the whole scene. 

Quick and dirty example 

https://user-images.githubusercontent.com/43168857/115438862-335e6c00-a20e-11eb-8366-769e126b8cc6.mp4

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
+++ b/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
@@ -54,7 +54,6 @@ namespace UnityEditor.Experimental.Rendering
                 }
             }
 
-            GUILayout.BeginHorizontal();
             if (GUILayout.Button(EditorGUIUtility.TrTextContent("Fit to Scene"), EditorStyles.miniButton))
             {
                 performFitting = true;
@@ -112,8 +111,6 @@ namespace UnityEditor.Experimental.Rendering
                 tmpClamp.z = Mathf.Max(0f, tmpClamp.z);
                 serialized.size.vector3Value = tmpClamp;
             }
-
-            GUILayout.EndHorizontal();
         }
 
         static void Drawer_ToolBar(SerializedProbeVolume serialized, Editor owner)


JulienIgnace-Unity:
Looks useful but isn't the fact that the volume fits exactly to the geometry a problem at borders since there won't be any probe around geometry there? Should we add some kind of margin or just let the user deal with it?
Tbh I expect that it's not really a problem for larger scenes since you probably won't really care about the edges but just wondering...

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
```diff
@@ -60,6 +61,90 @@ static void Drawer_FeatureEnableInfo(SerializedProbeVolume serialized, Editor ow
...
         static void Drawer_BakeToolBar(SerializedProbeVolume serialized, Editor owner)
         {
+            Bounds bounds = new Bounds();
+            bool foundABound = false;
+            bool performFitting = false;
+            bool performFittingOnlyOnSelection = false;
+
+            bool ContributesToGI(Renderer renderer)
+            {
+                var flags = GameObjectUtility.GetStaticEditorFlags(renderer.gameObject) & StaticEditorFlags.ContributeGI;
```
alelievr, line 71: also, check that the render is enabled

FrancescoC-unity:
> Looks useful but isn't the fact that the volume fits exactly to the geometry a problem at borders since there won't be any probe around geometry there? Should we add some kind of margin or just let the user deal with it?
> Tbh I expect that it's not really a problem for larger scenes since you probably won't really care about the edges but just wondering...

I did add an inflation just for that (there is a 1.05 on the size)

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
FrancescoC-unity: Good point, will update

alelievr (commented):
Looks good to me, but I agree with Julien. In case the volume is near the border of a cell, there won't be any bricks generated. I think we should at least add the min size of a brick in padding for the bounds

FrancescoC-unity:
> Looks good to me, but I agree with Julien. In case the volume is near the border of a cell, there won't be any bricks generated. I think we should at least add the min size of a brick in padding for the bounds

I can add the padding, but the brick generation will be the same at the current moment as we check against the same bounding boxes of the scene for the placement.

JulienIgnace-Unity (approved):

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
```diff
@@ -60,6 +61,90 @@ static void Drawer_FeatureEnableInfo(SerializedProbeVolume serialized, Editor ow
...
+                    var transforms = Selection.transforms;
+                    foreach (var transform in transforms)
+                    {
+                        var childrens = transform.gameObject.GetComponentsInChildren<Transform>();
+                        foreach (var children in childrens)
+                        {
+                            Renderer childRenderer;
+                            if (children.gameObject.TryGetComponent<Renderer>(out childRenderer))
+                            {
+                                bool childContributeGI = ContributesToGI(childRenderer) && childRenderer.gameObject.activeInHierarchy;
```
alelievr, line 112: Checking for activeInHierarchy is not enough, you also need to check if the renderer is enabled because TryGetComponent<> can return disabled components.

https://docs.unity3d.com/ScriptReference/Renderer-enabled.html

alelievr (approved):

PR #4250, opened 2021-04-20
Clamp SRGBToLinear inputs to avoid artifacts when using HDR and Gamma…

hkr:
### Purpose of this PR
For some reason Unity supports rendering in "Gamma Space" in HDR.
Some URP post-processing effects such as DoF convert the input to linear in the fragment shader when in Gamma Space rendering is used.
The result of `SRGBToLinear` can easily exceed `HALF_MAX`, so as a workaround I'm clamping the inputs.

(case 1327069)

---
### Testing status
manual with repro project of case 1327069.

---
### Comments to reviewers
The intention of this change is to have a simple workaround because hopefully not many people use Gamma Space rendering with HDR. Can anyone see any problems with this solution?

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
@@ -76,6 +76,9 @@ real4 LinearToGamma22(real4 c)
 // sRGB
 real SRGBToLinear(real c)
 {
+#if defined(UNITY_COLORSPACE_GAMMA) && REAL_IS_HALF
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

sebastienlagarde:
>The intention of this change is to have a simple workaround because hopefully not many people use Gamma Space rendering >with HDR. Can anyone see any problems with this solution?

I already don't see any reasons for this "Gamma Space", so I will not complain :)

Chman (approved):

phi-lira (approved):

PR #4253, opened 2021-04-20
Default black texture XR should be opaque (alpha = 1)

pmavridis:
### Purpose of this PR
The default black texture XR is now opaque (alpha = 1).
For alpha = 0, the clear texture XR can be used.


pmavridis:
This is in core, so I'm launching a full ABV to see if it affects URP or VFX.

sebastienlagarde (approved):

PR #4255, opened 2021-04-21
fixed SMAA not working for GL

matas-tunkevicius:

### Checklist for PR maker

---
### Purpose of this PR
This is a fix for SMAA not working when using OpenGL because the SMAA algorithm required SearchTex and EdgeTex to be sampled with linear sampling.

GL does not support sampler states or inline sampler states on graphics shaders, so the filtering needs to be set in the URP code.
Issue: https://fogbugz.unity3d.com/f/cases/1318214/

---
### Testing status
Compared images with Vulkan and GLES using SMAA - no more differences
Changed the reference images for the SMAA test, as it was incorrect for GL - the SMAA tests pass.
Some tests fail but they seem to be failing too on nightly.

---
### Comments to reviewers
This did not affect other APIs because they use sampler states in shaders.
SearchTex is precalculated so only changing import settings for it is enough.


com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -570,8 +570,8 @@ void DoSubpixelMorphologicalAntialiasing(ref CameraData cameraData, CommandBuffe
                 stencil = m_Depth.Identifier();
                 tempDepthBits = 0;
             }
-            cmd.GetTemporaryRT(ShaderConstants._EdgeTexture, GetCompatibleDescriptor(m_Descriptor.width, m_Descriptor.height, m_SMAAEdgeFormat, tempDepthBits), FilterMode.Point);
-            cmd.GetTemporaryRT(ShaderConstants._BlendTexture, GetCompatibleDescriptor(m_Descriptor.width, m_Descriptor.height, GraphicsFormat.R8G8B8A8_UNorm), FilterMode.Point);
+            cmd.GetTemporaryRT(ShaderConstants._EdgeTexture, GetCompatibleDescriptor(m_Descriptor.width, m_Descriptor.height, m_SMAAEdgeFormat, tempDepthBits), FilterMode.Bilinear);
+            cmd.GetTemporaryRT(ShaderConstants._BlendTexture, GetCompatibleDescriptor(m_Descriptor.width, m_Descriptor.height, GraphicsFormat.R8G8B8A8_UNorm), FilterMode.Bilinear);
```
phi-lira, line 574: Why does it fail specifically in GL? Not sure if filtering was wrong or used point for optimization. This will possibly cause a performance regression in all platforms?

matas-tunkevicius: GL does not support sampler states or inline sampler states on graphics shaders, so the filtering needs to be set in the URP code. That's why other APIs were not failing. And about the performance regression, it shouldn't have an impact because the textures we're being sampled linear before too, but it was being done in the shader.

erikabar (approved):
Approving on behalf of URP QA, mobile team should do the test pass

phi-lira (approved):

aleksandrasdzikia:
Looks good on mobile side too.

PR #4260, opened 2021-04-21
Move probe volumes option in HDRP

FrancescoC-unity:
Was previously defined in core (where it should be, but it will appear in URP already which we don't want), but moving to HDRP to avoid having the option displayed on pipelines not supporting APV.


alelievr (approved):

JulienIgnace-Unity:
Do we really want to do that?
Sure, at some point URP will use probe volume as well and we can move it back but what about custom SRPs? They will all have the ProbeVolumeOption components anyway.
Feels like we should have a proper way to hide components based on "condition" (a bit like we do with the lookdev where we use MenuItem Attribute with a condition). We could probably integrate that in the VolumeComponentMenu attribute somehow.
Same question goes when not enabling probe volumes on the project. In this case we should also hide the component anyway.

I think now that we have a proper initialization method for the probe volume system we could just check if it's initialized on the singleton (so basically all we need to do is implement an optional condition functor on VolumeComponentMenu)

PR #4254, opened 2021-04-20
Fixed Selection/Picking passes for URP

joshua-davis:
# **Please read the [Contributing guide](CONTRIBUTING.md) before making a PR.**

* Read the [Graphics repository & Yamato FAQ](http://go/graphics-yamato-faq).

### Checklist for PR maker
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.

---
### Purpose of this PR

Bug: https://fogbugz.unity3d.com/f/cases/1323618/
Shaders made with Shader Graph in URP don't react the same way to selection and picking as hand-written shaders. The cause was not having the SceneSelection and ScenePicking passes defined. URP hand-written shaders are fine because there's a fallback shader that happens to work given the field names, but with shader graph it didn't work.

---
### Testing status
Created a ShaderGraph in URP with alpha clipping.
Verified that picking worked where the mesh was visible and didn't select where alpha clipping was.
Verified that when selected, the mesh highlight would show a highlight around the clipped region.
Also verified that this worked for all 5 subtarget types in URP (Lit, Unlit, LitSprite, UnlitSprite, CustomLitSprite).

---
### Comments to reviewers
The basics of the implementation were taken from how HDRP handles this.

PR #4264, opened 2021-04-21
Adding the selection and picking passes

joshua-davis:
---
### Purpose of this PR
There are two somewhat hidden passes that can be implemented:
 - SceneSelection: Draws the outline of the selected object
 - ScenePicking: Controls what objects get selected when you click
Built-in has some hard-coded shaders to handle this if none are provided, but it relies on some very special names. In particular, alpha clipping doesn't properly work in these two passes with SG. This is actually related to https://fogbugz.unity3d.com/f/cases/1323618/ as well.

---
### Testing status
Used the project provided in the bug and verified that selection and picking work as expected, with both the lit and unlit subtarget.

---
### Comments to reviewers
This is roughly the same as https://github.com/Unity-Technologies/Graphics/pull/4254 which is also taken roughly from HDRP


jessebarker (approved):
LGTM

PR #4266, opened 2021-04-21
Hd/fix migration hdrpasset to globalsettings

RSlysz:
### Purpose of this PR

Initially this was to fix this FB https://fogbugz.unity3d.com/f/cases/1328664/ It needed a whole cleaning on the way we construct HDRenderPipelineGlobalSettings from a HDRenderPipelineAsset. (see below)

Migration from HDRPAsset to HDRPGlobalSettings fail in some case. This PR ensure en double migration pattern (from both HDRPAsset and HDRPGlobalSettings side) with a sync point:
- If migrating from HDRPAsset, a version lower than AddedHDRenderPipelineGlobalSettings, it will migrate data to HDRPGlobalSettings (create it if needed) right away and not wait anymore next call of Ensure.
- ~~If migrating from HDRPGlobalSettings, a version lower than MigratedFromHDRPAssetOrCreated, it will check if there is an HDRPAsset in graphics settings with version under AddedHDRenderPipelineGlobalSettings. If it's the case, it will trigger this asset migration right away - which will migrate datas from HDRPAsset to HDRPGlobalSettings - and then will continue other migration added in future.~~ This cannot be maintened. Meanwhile this PR was done, a public release was done with a first migration step on global settings. The migration of the HDRPGlobalSettingsAsset don't contains any step to say it was created (with or without data in HDRPAsset). Instead, Ensure that was creating the HDRPGlobalAsset if needed and not here, now make used HDRPAsset in GraphicsSettings to migrate and create one first. Then it create one if no one was created.

Now asset that need migration must implement IMigratableAsset.
The migration is attempted in several location:
- ~~OnEnable() of asset. We know that this is not called when using cache server. But this is called at import. Other migration point will attempt to limitate problematic cases but this should be addressed in another PR.~~ OnEnable caused infinite migration loop. Ensure behave somewhat the same as it is called whenever we need data in the global settings asset.
- when HDRP package upgrade, we load every asset that is IMigratableAsset. This will be time consuming for big project but the function try to load less possible things (See introduced LoadAllAsset<T>). This way, we prevent using assets that are not at last version.
- Wizard will check if used assets (HDRPAssets in quality and in graphics, used Resources that are migratable and current global settings) are at last version. And there is a Fixer to upgrade them.
- To ensure migration is done in Editor now, build exception have been added when GlobalSettings asset is missing or is note at last version or one of the used HDRPAsset is not at last version.

Side fixes done:
- the folder existence before asset creation is done at several places. I made an EnsureFolderTreeInAssetFilePath for this specific case. It support more than one folder at a time and will parse the path to grab the folder part.
- copy of data when cloning a GlobalSettings asset is now done through serialization. It will be simpler to ensure we don't forget to amend here as we add new fields.
- migration stuff have been moved to .migration dedicated files.
- HDProjectSettings have moved in runtime assembly in UNITY_EDITOR in order to be able to gather resource folder from asset creator

---
### Testing status
Manually tested following case:
- no HDRPGlobalSettings , no HDRPAsset in graphics 
  -> do nothing, 
- no HDRPGlobalSettings , adding a HDRPAsset in graphics at version former than AddedHDRenderPipelineGlobalSettings  
  -> create HDRPGlobalSettings, migrate data from HDRPAsset to HDRPGlobalSettings and clean data in HDRPAsset
- no HDRPGlobalSettings , adding a HDRPAsset in graphics at version bigger or equals to AddedHDRenderPipelineGlobalSettings    
  -> create HDRPGlobalSettings, don't migrate any data, version of HDRPGlobalSettings is the last available
- Build:
  - attempting to build with no HDRPGlobalSettings while having a HDRPAsset in graphics should raise a build error but I was **`not able to test it`** as HDRPGlobalSettings is created as soon as HDRPAsset is in GraphicsSettings. But I prefer being paranoid here.
  - attempted to build with an asset at a non last version failed (asset list tested: HDRPasset in GraphicsSettings, HDRPAsset in QualitySeetings, HDRPGlobalSettings)
image
  - attempt to build with all above asset at last version is a success (in HDRP)
  - attempt to build with all above asset at last version is a success (in HDRP + DXR)
  - verified that RayTracingResources are only in global settings if DXR is enabled.
- Wizard behaviour:
  - the error is raised as soon as one of the listed asset is not at last version (assets are HDRPAsset in quality settings, in graphics settings, current HDRPGlobalSettings' RuntimeResources asset,  and finally HDRPGlobalSettings).
  - missing asset cause no crash when computing the list
  - fix button actuallyupgrade each asset not up to date
image

- Tested EnsureFolderTreeInAssetFilePath with complexe path where no folder was existing. Folder tree was created correctly. Code used:
```
using UnityEditor;
using UnityEngine;
using UnityEngine.Rendering;

public class ScriptableTest : ScriptableObject
{
    [MenuItem("Assets/Create/ScriptableTest", priority = 0)]
    static void CreateHDRenderPipelineGlobalSettings()
    {
        const string path = "Assets/F0/F1/F2/F3/ST.asset";
        CoreUtils.EnsureFolderTreeInAssetFilePath(path);
        var assetCreated = ScriptableObject.CreateInstance<ScriptableTest>();
        AssetDatabase.CreateAsset(assetCreated, path);
        Debug.Assert(assetCreated);
        assetCreated.name = System.IO.Path.GetFileName(path);
    }
}
```

- Tested that LoadAllAsset actually load all asset of the given type. Though it rely on trunk function that is costly. So it is only used when we are sure we upgraded package HDRP.

- Migrated from HDTemplate (2020.3) to this branch (2021.2.0a13). Data was correctly copied in the created global settings asset
- Migrated from SpaceshipDemo (last public master) to this branch (2021.2.0a13). Data was correctly copied in the created global settings asset. Run seamed ok (I'm not expert here).
- Migrated from SpaceshipDemo (last internal master) to this branch (2021.2.0a13). Data was correctly copied in the created global settings asset. Run seamed ok (I'm not expert here). Checked with Thomas I that image take by benchmark seams ok.

- Checked that issue in https://fogbugz.unity3d.com/f/cases/1328664/ is fixed by merged the lens flare branche inside this branche and tested it.

---
### Comments to reviewers
Introduced change will make migration work better as it will Ensure migration is done in Editor. It implies that **Runtime_Test and Preformance_Test now need to have their HDRPAsset and GlobalSettings asset upgraded when we introduce a migration step** as the build is done directly in command line.
For Performance_Test, if there is a replacement mechanisme between the Forward and the Deffered HDRPAsset, I bet it is currently broken on master since introduction of GlobalSettings. I created two different globalSettings from the two HDRPAsset and linked the one corresponding to Forward (the one in use in project by default)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.Migration.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.Migration.cs
@@ -84,7 +84,6 @@ namespace UnityEngine.Rendering.HighDefinition
             assetToUpgrade.lightLayerName6 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName6;
             assetToUpgrade.lightLayerName7 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName7;
 
-            // Decal layer names were added in 2021 cycle
             if (oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteDecalLayerName0 != null)
             {
                 assetToUpgrade.decalLayerName0 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteDecalLayerName0;


jenniferd-unity (changes requested):
I also wonder how the Template HD projects upgraded - we probably need to check the diffusion profiles and the layer names were properly migrated.

TomasKiniulis (changes requested):
I tried upgrading HDRP Template from 2020.3 to 2021.2.0a14 PR. Still keep getting LookDev Profile, Layer names, Diffusion Profiles lost.

2020.3 result:
image (2)
image (1)

2021.2.0a14 PR branch result:
image

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/DefaultSettingsPanel.cs
```diff
@@ -125,7 +125,7 @@ public void DoGUI(string searchContext)
             {
                 if (HDRenderPipeline.currentAsset != null || HDRenderPipelineGlobalSettings.instance != null)
                 {
-                    settingsSerialized = HDRenderPipelineGlobalSettings.Ensure(folderPath: HDProjectSettings.projectSettingsFolderPath);
+                    settingsSerialized = HDRenderPipelineGlobalSettings.Ensure();
```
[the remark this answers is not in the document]
RSlysz: Folder path given was always the one in HDProjectSettings. It was simpler to make it use it directly.

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
```diff
@@ -407,7 +407,7 @@ bool IsHdrpGlobalSettingsUsedCorrect()
             => HDRenderPipelineGlobalSettings.instance != null;
 
         void FixHdrpGlobalSettingsUsed(bool fromAsync)
-            => HDRenderPipelineGlobalSettings.Ensure(folderPath: HDProjectSettings.projectSettingsFolderPath);
+            => HDRenderPipelineGlobalSettings.Ensure();
```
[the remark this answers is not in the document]
RSlysz: same: Folder path given was always the one in HDProjectSettings. It was simpler to make it use it directly.

RSlysz:
I found why it is broken. Last merge with Master introduce a migration point in OnEnable for global settings. I found it to be a good idea and kept it. But it cause an infinit migration loop which will fail in the end due to a stack overflow error.

I'm fixing it but It will take time to redo all manual tests that I had done before that 8cea854 merge.

RSlysz:
I updated code:
- Introduced OnEnable in last merge before laste review is no replaced by Ensure
- Updated Resources Ensure to deal with ADB not ready
- Remove the version in HDRenderPipelineGlobalSettings. In all case, we prevent data copy if it is not at first version.

I done more manual test:
- HDTemplate and Spaceship seams to work as expected.

RSlysz:
There is still two issues:
- asset in test project should be migrated now that we cannot have it on OnEnable (Still searching another way to do this)
- migration step on the created asset, though beeing done after, are still not registered. (tested with lens flare PR)

TomasKiniulis (approved):
From my side this looks good, didn't run into any more PR related issues. 

What I checked:
- Tested HDRP Template upgrade from 2020.3 to 2021.2 PR - Diffusion Profiles, Layers, Volumes are saved. Also made sure that Custom Post Process Orders stay.
- Upgraded Amalienborg palace - same Default Settings persist
- Upgraded URP Template from 2020.3 to 2021.2 PR with HDRP in project as well. While HDRP asset is active/inactive, also ran player build.

JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.Migration.cs
```diff
@@ -20,11 +36,95 @@ enum Version
...
+            assetToUpgrade.lightLayerName0 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName0;
+            assetToUpgrade.lightLayerName1 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName1;
+            assetToUpgrade.lightLayerName2 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName2;
+            assetToUpgrade.lightLayerName3 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName3;
+            assetToUpgrade.lightLayerName4 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName4;
+            assetToUpgrade.lightLayerName5 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName5;
+            assetToUpgrade.lightLayerName6 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName6;
+            assetToUpgrade.lightLayerName7 = oldAsset.currentPlatformRenderPipelineSettings.m_ObsoleteLightLayerName7;
+
+            // Decal layer names were added in 2021 cycle
```
sebastienlagarde, line 87: nope they are in 2020.2 (decal layer)

RSlysz: @jenniferd-unity :trollface:

jenniferd-unity: ok we can just remove the comment anyway :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.cs
```diff
@@ -502,18 +333,32 @@ void EnsureResources<T>(bool forceReload, ref T resources, string resourcePath,
                 if (forceReload)
                     ResourceReloader.ReloadAllNullIn(resources, HDUtils.GetHDRenderPipelinePath());
             }
-            else // Asset database may not be ready
+            else
             {
+                // Asset database may not be ready
                 var objs = InternalEditorUtility.LoadSerializedFileAndForget(resourcePath);
                 resources = (objs != null && objs.Length > 0) ? objs[0] as T : null;
                 if (forceReload)
```
jenniferd-unity, line 341: you cannot go through that code path since you are already in !forceReload

jenniferd-unity (changes requested):
There is some weird code left in EnsureResources again.
Otherwise a few english grammar issues here and there we can solve later :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.cs
RSlysz: Well no we are not o_o

jenniferd-unity (approved):
Sorry for the last comment :)

PR #4268, opened 2021-04-21
[10.x.x HDRP] Fix issue with raytracing resources not properly initialize

sebastienlagarde:
### Purpose of this PR

This PR correctly fix the issue attempted to be fixed in this PR:
https://github.com/Unity-Technologies/Graphics/pull/3894
and this fogbugzz:
https://fogbugz.unity3d.com/f/cases/1320304/

We are changing the behavior.

Previously, the raytracing resources was re-loaded and init only if the option  Real-time raytracing was enabled on the current use asset and reset to null otherwise. This is causing issue in case of multiple configuration with mismatch on this property (i.e some true, some false).
The PR https://github.com/Unity-Technologies/Graphics/pull/3894 is an attempt to fix this problem but wasn't the correct solution and the issue was remaining.

The correct solution is to always initialize the raytracing resources, as they are located in the Default settings and must be embed in the build if any of the configuration required it. So the users could switch at runtime between raytracing and non raytracing.

This is what this PR is doing. It now always init and reload the raytracing resources like the regular one if any of the quality level of graphicssettings RP have raytracing - this is what is done on master - BUT it use the stripper to remove the unused compute shader from it. However there is not stripper yet for raytrace shader so they can't be removed. Fortunately, on platform which don't support raytrace shader (so all of them except DX12), those shader will not be compile anyway.

The downside is that for platform which support DX12 the raytrace resources will always been embed even if raytracing is not used. As of today the usage of DX12 is related to raytracing as otherwise it is less performant it is not really a problem.

---
### Testing status
I have tested the template on windows with a RTX card. I was able to build a player with and without raytracing and checked in the compute shader compilation list with verbose mode that compute shader used for raytracing were correctly used.

---
### Comments to reviewers

For QA, can you please test this PR on metal and console to see if the player build works correctly? I want to check that a non null raytrace resource have no impact on those platform. Thanks

Note : I don't add a changelog as I am reusing the one in https://github.com/Unity-Technologies/Graphics/pull/3894 as it have been backported "Fixed HDRPAsset loosing its reference to the ray tracing resources when clicking on a different quality level that doesn't have ray tracing (case 1320304)."

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -612,18 +612,35 @@ namespace UnityEngine.Rendering.HighDefinition
             ResourceReloader.ReloadAllNullIn(asset.renderPipelineResources, HDUtils.GetHDRenderPipelinePath());
 #endif
 
-            // We always allocate raytracing resources as there is no way currently to know if they will be used or not.
-            // When we have multiple RP Asset, some could have the raytracing enabled, the other not.
-            // To do thing properly we should strip the resources inside the stripper based on the platform and all the
-            // Render pipeline available there.
-            if (asset.renderPipelineRayTracingResources == null)
-                asset.renderPipelineRayTracingResources
-                    = UnityEditor.AssetDatabase.LoadAssetAtPath<HDRenderPipelineRayTracingResources>(HDUtils.GetHDRenderPipelinePath() + "Runtime/RenderPipelineResources/HDRenderPipelineRayTracingResources.asset");
+            bool requiresRayTracingResources = false;
+            // Make sure to include ray-tracing resources if at least one of the defaultAsset or quality levels needs it
+            if (defaultAsset.currentPlatformRenderPipelineSettings.supportRayTracing)
+                requiresRayTracingResources = true;
+
+            int qualityLevelCount = QualitySettings.names.Length;
+            for (int i = 0; i < qualityLevelCount && !requiresRayTracingResources; ++i)
+            {
+                var hdrpAsset = QualitySettings.GetRenderPipelineAssetAt(i) as HDRenderPipelineAsset;
+                if (hdrpAsset != null && hdrpAsset.currentPlatformRenderPipelineSettings.supportRayTracing)
+                    requiresRayTracingResources = true;
+            }
+
+            if (requiresRayTracingResources)
+            {
+                if (asset.renderPipelineRayTracingResources == null)
+                    asset.renderPipelineRayTracingResources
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
```diff
@@ -535,23 +541,73 @@ public void OnProcessShader(Shader shader, ShaderSnippetData snippet, IList<Shad
...
+            else
+                s_ComputeShaderCache = new Dictionary<int, ComputeShader>();
+
+            if (HDRenderPipeline.defaultAsset == null)
+                return;
+
+            if (HDRenderPipeline.defaultAsset.renderPipelineRayTracingResources == null)
+                return;
+
+            foreach (var fieldInfo in HDRenderPipeline.defaultAsset.renderPipelineRayTracingResources.GetType().GetFields(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static))
```
RSlysz, line 588: You should not need BindingFlags.Static. There will never have static field in resources as they are not serialized.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -612,7 +612,20 @@ void UpgradeResourcesInAssetIfNeeded(HDRenderPipelineAsset asset)
             ResourceReloader.ReloadAllNullIn(asset.renderPipelineResources, HDUtils.GetHDRenderPipelinePath());
 #endif
 
-            if (GatherRayTracingSupport(asset.currentPlatformRenderPipelineSettings))
+            bool requiresRayTracingResources = false;
+            // Make sure to include ray-tracing resources if at least one of the defaultAsset or quality levels needs it
+            if (defaultAsset.currentPlatformRenderPipelineSettings.supportRayTracing)
+                requiresRayTracingResources = true;
```
RSlysz, line 618: Can be simplified as
`bool requiresRayTracingResources = defaultAsset.currentPlatformRenderPipelineSettings.supportRayTracing;`

Does former GatherRayTracingSupport also check if hardware is compatible? Should we check it here?

RSlysz (approved):
Still, check if we also need to verify hardware compatibility or if `supportRayTracing` can only be true if this is the case.

sebastienlagarde:
After quick check with Remi we will merge the PR so it can be tested in 10.5.0

remi-chapelain:
already tested on the branch with 2020.3.6f1. 
- Render is fine again in editor if there's mismatch for realtime raytracing flags between default HDRP and current HDRP assets. ✔️ 
- Built a player with low quality (no raytracing) by default and tried to change quality setting at runtime to be sure that raytracing was still working. ✔️ 


Will make another quick pass on new 10.5.0 package version tomorow.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
sebastienlagarde: indeed, will make the change on master. We must not check hardware support as you can build with a GPU not supporting it

PR #4269, opened 2021-04-21
Enable test 039 for Android/GLES

hkr:
Enable test 039 for Android

There were to problems in the test scene:
I found two issues in the test scene.
The red tint happens because a normal map is used as emission texture. So the normal map encoding affects the color.

Another problem is the custom vertex layout used for the particle system 8 unlit distortion. I think it’s also wrong but only causes problems when instancing is disabled such as on Mali with GLES.

I don't know why I didn't have to update the reference image

### DIFF
--- a/TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/037_Particles/Textures/metal_E.psd.meta
+++ b/TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/037_Particles/Textures/metal_E.psd.meta
@@ -50,7 +50,7 @@ TextureImporter:
   alphaUsage: 1
   alphaIsTransparency: 0
   spriteTessellationDetail: -1
-  textureType: 1
+  textureType: 0
   textureShape: 1
   maxTextureSizeSet: 0
   compressionQualitySet: 0


phi-lira:
OpenGL seems like legit failure?

hkr:
So far I found two issues i

> OpenGL seems like legit failure?

I found two issues in the test scene.
The red tint happens because a normal map is used as emission texture. So the normal map encoding affects the color.

Another problem is the custom vertex layout used for the particle system 8 unlit distortion. I think it’s also wrong but only causes problems when instancing is disabled such as on Mali with GLES.

svens-unity (approved):
Linking to case for reference: https://fogbugz.unity3d.com/f/cases/1315494

[039 is running and passing](https://hoarder.hq.unity3d.com/TestExecutionHistory?Branch=universal%2Fmobile%2Ffix-1315494&BuildConfigId=9999&SubmitterId=71&TestName=UniversalGraphicsTest_Foundation.UniversalGraphicsTests.039_Particles_Instanced&TestId=1334347&TestSessionStartDate=2021-03-29T22:00:00.000Z&TestSessionEndDate=2021-04-29T21:59:59.000Z&BuildConfigUniqueName=.yamato%2Furp_foundation-android-opengles3.yml%23URP_Foundation_Android_OpenGLES3_Standalone_cache_il2cpp_Linear_trunk).

PR #4271, opened 2021-04-21
Refactor AddShaderInputAction to make it more extensible 

tracychen1234:
---
### Purpose of this PR
Refactor AddShaderInputAction.

blackboardrefactor
---
### Testing status
Failed on AddInputTests&RemoveInputTests which the staging branch also failed.
Cherry pick changes to the current master(45611f1495e924546c0ac1eef6829bc68fc20429) and passed the two tests.

---

### DIFF
--- a/com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
+++ b/com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
@@ -390,6 +390,7 @@ namespace UnityEditor.ShaderGraph.Drawing
                 // QUICK FIX TO DEAL WITH DEPRECATED COLOR PROPERTY
                 if (name.Equals("Color", StringComparison.InvariantCultureIgnoreCase) && ShaderGraphPreferences.allowDeprecatedBehaviors)
                 {
+                    order.isKeyword = false;
                     order.deprecatedPropertyName = name;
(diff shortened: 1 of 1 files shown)


Nightmask3 (changes requested):
I actually quite like this, makes it so that the `AddShaderInputAction` itself now doesn't need to know about how to instantiate different property types, instead we have a separate `BlackboardShaderInputFactory` class to handle that, using data from `BlackboardShaderInputOrder`.

Very neat, very straightforward. Looks good to me! Only thing would be to fix the typos in the function names (adddKeywordAction etc)

com.unity.shadergraph/Editor/Drawing/Controllers/BlackboardController.cs
```diff
@@ -14,6 +14,47 @@
...
+            else
+            {
+                switch (order.deprecatedPropertyName)
+                {
+                    case "Color":
+                        output = new ColorShaderProperty(order.version);
+                        break;
+                    default:
+                        //TODO: Not sure what should be the default return property
+                        output = null;
```
tracychen1234, line 50: Thank you @Nightmask3 ! Fixed the typoes haha. I also got another question here, what should we return as default in the switch statement?

Nightmask3 (approved):

PR #4187, opened 2021-04-15
[XPipeline]Cleaning Light Inspector Code and using SerializedLight

alex-vazquez:
### Purpose of this PR

Prepare URP Light Inspector for the Unification with HDRP

Creating the Interface `ISerializedCamera` and implement it on both pipelines

On URP using the `UniversalRenderPipelineSerializedCamera`

[JIRA](https://jira.unity3d.com/browse/XPIPELINE-144)

---
### Testing status
- Check the animations are deleted
- Check that the values are changed properly(Multiple inspectors opened)

---
### Comments to reviewers
- UX do not want animations when show/hide elements, I've removed them
- First step towards the unification of the Light Inspectors
- The Additional data is always initialized, no need to check if it null on the shadow properties, this simplifies a lot the code

PR #4275, opened 2021-04-22
[URP]Fix condition to show Mode field on the Light Inspector

alex-vazquez:
---
### Purpose of this PR
Fix issue introduced on https://github.com/Unity-Technologies/Graphics/pull/4187

The condition was not showing the Mode ( Baked, Realtime, Mixed ) for lights that are not Area light.

---
### Testing status
Check that the field is being shown, and is being hidden on area lights

---
### Comments to reviewers
The condition, was wrong, as I removed the AnimBools values. Also the second condition that the type is diferent from Disc it makes no sense for me.


phi-lira (approved):

phi-lira (approved):

erikabar (approved):
looks good - the behaviour remains the same as in the previous URP versions in 2019.4 and 2020.2

martint-unity (approved):
LGTM

PR #4279, opened 2021-04-22
Updated ray tracing settings and added it to the table of contents 

JordanL8:
### Purpose of this PR
The ray tracing settings page was missing from the table of contents and was missing a property.


remi-chapelain (approved):
Thanks for adding this ! ✔️

sebastienlagarde (approved):

PR #4281, opened 2021-04-22
Per-cell loading of assets for APV

FrancescoC-unity:
With this PR the loading of assets happens at a cell granularity,  x times per frame where X is determined by user (for now only via API until we finalize the user-facing settings).  The order is based on distance to camera. 

This should allow spreading the time of loading instead of a single huge frame hitch.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
@@ -440,6 +440,8 @@ namespace UnityEngine.Experimental.Rendering
 
         void PerformPendingLoading()
         {
+            LoadPendingCells();
+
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -55,6 +55,26 @@ internal struct Cell
             public float[] validity;
         }
 
+        private class CellSortInfo : IComparable
```
JulienIgnace-Unity, line 58: No need for private here as it's the default

JulienIgnace-Unity: And in this case all its members can be public since the class is private.

alelievr (approved):

JulienIgnace-Unity (approved):

PR #4286, opened 2021-04-22
[HDRP] Timeline animation of physical camera / focus distance mode

pmavridis:
### Purpose of this PR
With this PR, we give the option to read the lens focus distance either from the physical camera or the volume component with a new dropdown.  
Also includes some minor UI tweaks that were discussed with UX and our artists.
image

This change will also allow the focus distance to be animated (when it's read from the camera).
Existing projects should be fully compatible with this change.

---
### Testing status
I have tested this manually by switching the dropdown and making sure that we read the focus distance from the right place (either the volume or the camera). 

We need a graphics test for this, but I do not include one in this PR because @TomasKiniulis is currently moving the DoF tests in a single multi-camera one and we can test this new mode in this new multi-camera test.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
```diff
@@ -37,13 +37,24 @@ public class HDPhysicalCamera
         // Lens
         // Note: focalLength is already defined in the regular camera component
         [SerializeField][Range(kMinAperture, kMaxAperture)] float m_Aperture = 16f;
+        [SerializeField][Min(0f)] float m_FocusDistance = 10f;
```
FrancescoC-unity, line 40: This is bounded differently vs https://github.com/Unity-Technologies/Graphics/pull/4286/files#diff-9ee958b0bde47c58ac148fcbc5dd458acb9a4e0fa124b2ebc3676cc515eae2a4R90 . I am not sure if there was a reason why the one in the volume was not 0, but maybe worth homogenize based on whatever of the two make sense?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
```diff
@@ -37,13 +37,24 @@ public class HDPhysicalCamera
...
 
         // Aperture shape
         [SerializeField][Range(kMinBladeCount, kMaxBladeCount)] int m_BladeCount = 5;
         [SerializeField] Vector2 m_Curvature = new Vector2(2f, 11f);
         [SerializeField][Range(0f, 1f)] float m_BarrelClipping = 0.25f;
         [SerializeField][Range(-1f, 1f)] float m_Anamorphism = 0f;
 
+
+        /// <summary>
+        /// The focus distance of the lens, used in Depth of Field Computations when the DoF volume is configured to read this value from the camera.
```
FrancescoC-unity, line 50: Computations shouldn't be capitlized here

FrancescoC-unity (approved):
LGTM, only very minor comments

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
pmavridis: Good catch, in the end I kept the same minimum as the volume.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/DepthOfField.cs
```diff
@@ -47,6 +49,22 @@ public enum DepthOfFieldResolution : int
         Full = 1
     }
 
+    /// <summary>
+    /// The source of the focus distance used in the Depth of Field Computations
```
JordanL8, line 53: Options for the source of the focus distance HDRP uses in the depth of field calculations.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/DepthOfField.cs
```diff
@@ -47,6 +49,22 @@ public enum DepthOfFieldResolution : int
...
+    /// </summary>
+    public enum FocusDistanceMode
+    {
+        /// <summary>
+        /// The focus distance is set on the volume component
```
JordanL8, line 58: Uses the focus distance from the Volume component.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/DepthOfField.cs
```diff
@@ -47,6 +49,22 @@ public enum DepthOfFieldResolution : int
...
+        /// </summary>
+        Volume,
+
+        /// <summary>
+        /// The focus distance is set on the physical camera properties
```
JordanL8, line 63: Uses the focus distance from the physical camera.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/DepthOfField.cs
```diff
@@ -66,11 +84,17 @@ public sealed class DepthOfField : VolumeComponentWithQuality, IPostProcessCompo
         //
 
         /// <summary>
-        /// Sets the distance to the focus point from the Camera.
+        /// Sets the distance to the focus plane from the Camera.
```
JordanL8, line 87: The distance to the focus plane from the Camera.

The API docs guidance says to avoids gets/sets - https://confluence.unity3d.com/pages/viewpage.action?spaceKey=DOCS&title=API+docs+content+and+formatting#heading-Propertiesandfields

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/DepthOfField.cs
```diff
@@ -66,11 +84,17 @@ public sealed class DepthOfField : VolumeComponentWithQuality, IPostProcessCompo
...
         /// </summary>
-        [Tooltip("Sets the distance to the focus point from the Camera.")]
+        [Tooltip("Sets the distance to the focus plane from the Camera.")]
         public MinFloatParameter focusDistance = new MinFloatParameter(10f, 0.1f);
 
+        /// <summary>
+        /// Controls whether the focus distance is read from the physical camera or the volume itself
```
JordanL8, line 93: Specifies where to read the focus distance from.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/DepthOfField.cs
```diff
@@ -66,11 +84,17 @@ public sealed class DepthOfField : VolumeComponentWithQuality, IPostProcessCompo
...
+        /// </summary>
+        [Tooltip("Controls whether the focus distance is read from the physical camera or the volume itself.")]
```
JordanL8, line 95: Specifies where to read the focus distance from.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
```diff
@@ -37,13 +37,24 @@ public class HDPhysicalCamera
...
 
         // Aperture shape
         [SerializeField][Range(kMinBladeCount, kMaxBladeCount)] int m_BladeCount = 5;
         [SerializeField] Vector2 m_Curvature = new Vector2(2f, 11f);
         [SerializeField][Range(0f, 1f)] float m_BarrelClipping = 0.25f;
         [SerializeField][Range(-1f, 1f)] float m_Anamorphism = 0f;
 
+
+        /// <summary>
+        /// The focus distance of the lens, used in Depth of Field computations when the DoF volume is configured to read this value from the camera.
```
JordanL8, line 50: The focus distance of the lens. The Depth of Field Volume override uses this value if you set `focusDistanceMode` to `FocusDistanceMode.Camera`.

com.unity.render-pipelines.high-definition/Documentation~/Post-Processing-Depth-of-Field.md
```diff
@@ -20,8 +20,9 @@ Depth Of Field includes [additional properties](More-Options.md) that you must m
...
-| **Focus Distance** | Set the distance to the focus point from the Camera.<br />This property only appears when you select **Use Physical Camera** from the **Focus Mode** drop-down. |
+| **Focus Mode**     | Use the drop-down to select the mode that HDRP uses to set the focus for the depth of field effect.<br />&#8226; **Off**: Select this option to disable depth of field.<br />&#8226; **Physical Camera**: Select this option to use the physical [Camera](HDRP-Camera.md) to set focusing properties for the depth of field effect. For information on what Camera properties affect depth of field, see [Physical Camera settings](#PhysicalCameraSettings).<br />&#8226; **Manual Ranges**: Select this option to use custom values to set the near and far range of the depth of field effect. |
+| **Focus Distance Mode** | Use the drop-down to select where the focus distance is specified.<br /> &#8226; **Volume**: The focus distance is read from the volume. <br /> &#8226; **Camera**: The focus distance is read from the physical camera. <br /> This property only appears when you select **Physical Camera** from the **Focus Mode** drop-down. |
```
JordanL8, line 24: ```Use the drop-down to select where the focus distance is specified.<br /> &#8226; **Volume**: Reads the focus distance from the Volume. <br /> &#8226; **Camera**: Reads the focus distance from the physical camera. <br /> This property only appears when you select **Physical Camera** from the **Focus Mode** drop-down.```

JordanL8 (changes requested):

sebastienlagarde (approved):

sebastienlagarde:
@pmavridis  you need to merge master on this one, thanks

iM0ve (commented):
**Conclusion:** The additional property on the camera is present and usable. However I was not able to animate it.

**Whats tested:**
* Upgrading Amalienborg project that uses Physical cameras
* New property Focus Distance on the camera exists, only has effect when Camera mode is chosen in the DOF override.
* Already existing Focus Distance property found on the DOF override only works when the mode is set to Volume
* Animating Focus Distance of the camera
* Using other Physical Camera properties to see that they are still working

**Conserns** 
Can not animate Focus Distance property using timeline. In the video you can see that I am able to animate FOV, it becomes red when changed and adds a keyframe. Focus Distance however does not behave in the same way, as a matter of fact, none of the Physical Camera properties can be animated with timeline.


https://user-images.githubusercontent.com/36502659/116886443-c3021280-ac31-11eb-85dd-f79d0215d393.mp4


**Project upgrade:**
No issues found here

Master
Master

PR
Unity_sg3HWuJ5cm

pmavridis:
Regarding Timeline animations, it looks that animation is not working because the phyiscal camera fields are stored in a separate sub-class (HDPhysicalCamera) and not directly in the HDAdditionalCameraData component.
After storing the fields directly in HDAdditionalCameraData aniimation was working in my test, but I need to investigate if we can do that without breaking old projects...

pmavridis:
After discussing the animation issue with @RSlysz, we have found that a solution is to convert the class to a struct.
My test shows that this does not require any explicit migration step:
- I changed some physical camera values in an HDRP project before this update.
- Updated HDRP to the version of this PR
- Opened the project and physical camera had the values I had set.

Also animation with Timeline seems to work now:
image
(it does not work for aperture and curvature because they use some custom UI drawing. We can address those later)
 
[Adding @RSlysz in the reviewers]

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
```diff
@@ -121,11 +131,30 @@ public void CopyTo(HDPhysicalCamera c)
             c.iso = iso;
             c.shutterSpeed = shutterSpeed;
             c.aperture = aperture;
+            c.focusDistance = focusDistance;
             c.bladeCount = bladeCount;
             c.curvature = curvature;
             c.barrelClipping = barrelClipping;
             c.anamorphism = anamorphism;
         }
```
RSlysz, line 139: As it is a struct now and as we copy all field, this method is obsolete.
Basically if assign the struct to another one, it will copy everything inside.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -292,7 +291,7 @@ void BeginPostProcessFrame(CommandBuffer cmd, HDCamera camera, HDRenderPipeline
 
             // Grab physical camera settings or a default instance if it's null (should only happen
             // in rare occasions due to how HDAdditionalCameraData is added to the camera)
-            m_PhysicalCamera = camera.physicalParameters ?? m_DefaultPhysicalCamera;
+            m_PhysicalCamera = camera.physicalParameters;
```
RSlysz, line 294: just to be 100% sure:
We don't modify it right ?

Before it was shared value by reference. Now it is just a copy.

RSlysz (approved):
Seams good. Raised some question just to be sure.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
pmavridis: Good point. However this is a public API (and a "blame" shows it has been arround for ~2 years), so I guess it makes sense to keep it for backwards compatibility, right?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
pmavridis: Yes we don't modify it, it's a copy. 
I kept it like this, in case a post-process pass was modifying the physicalParameters.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
RSlysz: Right. But perhaps we can start a deprecation plan like putting it as obsolete (warning, explaining why) then later putting it as obsolete (error) then later removing it. (Can be done later)

@sebastienlagarde do we have any file to keep track of deprecated functionality? And do we have any official deprecation plan? Should we invest little time on having a clear deprecation plan in packages in the next roadmap?

pmavridis: Ok, I marked the method as obsolete, so no need to keep track of this. When we say we will do somethign later, we never do it in the end...

iM0ve (approved):
With the latest changes I rechecked a few things:
* Physical camera properties still function
* After changes are made to said properties, they persist after editor restart
* Rechecked some DOF override properties, including Physically Based DOF and Quality levels for DOF
* Checked that in the Exposure override, Physical camera property can still be used
* Most importantly, successfully animated the new parameter focus distance on the camera

No more issues on my side

https://user-images.githubusercontent.com/36502659/117966096-34c21680-b32c-11eb-8b67-20e8f993c16b.mp4

PR #4291, opened 2021-04-23
Correct sub shader order in Unlit Shader Graph

pbbastian:
### Purpose of this PR
Hybrid Renderer relies on shader target 4.5 to work, so for that reason URP has a sub shader for 4.5 and one for 2.0. Unity chooses the first compatible one. The Unlit Shader Graph was putting the 2.0 subshader before the 4.5 one. This was not an issue when the platforms for each sub shader were exclusive. But since #3758, D3D11 is now in both of them. So that caused Unity to always pick shader target 2.0 for D3D11.

---
### Testing status
Verified shader output locally.


phi-lira (approved):

phi-lira (approved):

cinight (approved):
Verified this fixes the bug (case 1328636)

erikabar (approved):

PR #4293, opened 2021-04-23
[HDRP] Fix undo of some properties on light editor

adrien-de-tocqueville:
### Purpose of this PR

Fixing undo/redo on some properties of the light inspector.

The bug is very strange as it doesn't happen with every light, see the gif below. I didn't manage to understand the reason why :(
Therefore I am not 100% sure of the fix but testing seem to prove it works ok

UPDATE About the behavior on master: (noticed after the PR has been merged)
- If the "Reflector" property is unchecked, undo will work
- If the box is checked, undo will not work
Not idea why though

---
### Testing status

For spot box light
- tested undoing redoing the size

For rectangle area light
- tested that emissive mesh size gets correctly updated after undo/redo
- tested undo/redo of enable emissive mesh


iM0ve (approved):
Verified on my project:
1) Had lights with not working Undo in my scene
2) Updated the project to this PR
3) Undo now is working on previously bugged lights

Additionally checked that light values are still saved, by closing and reopening the project. Checked undo on other light properties (range, affect diffuse etc).


https://user-images.githubusercontent.com/36502659/115873677-e1a32500-a44b-11eb-8deb-9c6d96810d29.mp4

sebastienlagarde (approved):

PR #4302, opened 2021-04-23
Hide redundant checkboxes in Emissive Materials tab.

pema99:
---
### Purpose of this PR
[This PR](https://ono.unity3d.com/unity/unity/pull-request/123226/_/lighting/light-explorer-static-emissive) hides the "Show Inactive Objects" and "Isolate Selection" checkboxes from the "Static Emissives" tab of the Light Explorer, as these settings are redundant for this tab, since it shows materials, which are not scene objects. As HDRP uses a separate extension for Light Explorer tabs, these checkboxes should be hidden on the equivalent "Emissive Materials" tab, as well. This PR simply hides said checkboxes on this tab by setting a flag.

---
### Testing status
The mechanism for hiding this part of the UI is already covered by tests in the trunk PR. I don't think additional testing is warranted or practical since tests in the Graphics repo don't have access to internal members of the Light Explorer class, which would be required for testing in the same manner as the tests in the trunk PR. This change is very small, and just makes use of the newly added ability to hide the checkboxes.

---
### Comments to reviewers
This PR depends on the aforementioned trunk PR, and can't land before it.


sebastienlagarde (approved):
Please add a changelog

PR #396, opened 2020-05-08
Universal/ Deferred render pass PR

jonuuukas:
**Guide** : https://github.com/Unity-Technologies/Graphics/blob/pr-template-correction/.github/pr-read.png.md

**Display Addon** : https://userstyles.org/styles/182991/unity-graphics-pr-readme

# Purpose of this PR

Adding render pass and framebuffer fetch support to URP and modifying the deferred renderer to use it. 
Works natively on Vulkan and Metal (not OSX though). Fallback path is used for other platforms

# Testing status
Some tests might be failing due to ShaderGraph not recompiling the shaders with added positionCS. Tests 10, 15 are failing by design (same as deferred without render passes). also test 127 fails due to no support for dynamic scaling (yet, will be done in follow up PRs)

Early QA check done by Ernestas, we still need a second one
## Manual Tests
iOS metal; Android Vulkan; Metal Editor

## Links
**Yamato**: (Select your branch) https://yamato.prd.cds.internal.unity3d.com/jobs/902-Graphics
> Any test projects or documents to go with this to help reviewers?
To test you need to create a Deferred Renderer asset (same process here as in deferred renderer). Render Passes will work by default now

# Comments to reviewers
> Notes for the reviewers you have assigned.
trunk branch needed for this: graphics/render-pass/test1 ( with two changes that are still in flight)

PR #4307, opened 2021-04-26
adding FB Fetch macros and PLATFORM_SUPPORTS_NATIVE_RENDERPASS

jonuuukas:
---
### Purpose of this PR
This PR introduces fb fetch macros that will be needed for implementing input attachments for URP RenderPass work. However, it's more of a core lib thing, so adding it to core package. The macros itself are salvaged from previous Renderpass work PR (https://github.com/Unity-Technologies/Graphics/pull/396) with an addition of DXC compiler branches, that was recently introduced
---
### Testing status
Manually hacked the shaders to check if it works. But most proper testing will be done with input attachment PR
---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
@@ -205,7 +205,7 @@
 
     #if defined(UNITY_COMPILER_DXC)
 
-        //Subpass inputs are disallowed in non-fragment shader stages with DXC so we need some dummy value to use in the fragment function while it's not bing compiled
+        //Subpass inputs are disallowed in non-fragment shader stages with DXC so we need some dummy value to use in the fragment function while it's not being compiled
         #if defined(SHADER_STAGE_FRAGMENT)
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -201,6 +201,77 @@
 #define PLATFORM_SUPPORTS_PRIMITIVE_ID_IN_PIXEL_SHADER
 #endif
 
+#if defined(PLATFORM_SUPPORTS_NATIVE_RENDERPASS)
+
+    #if defined(UNITY_COMPILER_DXC)
+
+        //Subpass inputs are disallowed in non-fragment shader stages with DXC so we need some dummy value to use in the fragment function while it's not bing compiled
```
phi-lira, line 208:
```suggestion
        //Subpass inputs are disallowed in non-fragment shader stages with DXC so we need some dummy value to use in the fragment function while it's not being compiled
```

phi-lira (approved):

sebastienlagarde (approved):

phi-lira:
ABV is green except Update job that was known issue and already fix in master.

PR #4308, opened 2021-04-26
[URP]Fix missing Range Field

alex-vazquez:
---
### Purpose of this PR
Range field was being shown only on Directional lights, is just the oposite.

Error in the property check to show the field.

---
### Testing status
Check that the Range is not shown when the type of the light is directional, and is shown on all the other types.

---
### Comments to reviewers
Issue by my previous PR


alex-vazquez:
@Unity-Technologies/gfx-qa-urp, could you do a full test on the light inspector if we missed something else? Check against a version without the last changes

martint-unity (approved):

simon-engelbrecht-soerensen (approved):
Looking good!
[Doc](https://docs.google.com/document/d/1vfXqzUHQeUXDrdn4p6P0eE3GL48eq6la9JdMRn1TX4o/edit#)

PR #4309, opened 2021-04-26
Added missing .sample.json file to URP samples

Jonasmortensen:
Purpose:
A required file was missing from the Universal Render Pipeline package samples. This PR adds that file.


phi-lira (approved):

phi-lira:
QA is not needed in this PR due to only adding json for package sample.

PR #4311, opened 2021-04-26
Fix apv structure seams

alelievr:
### Purpose of this PR
Fix the cell size and other mistakes in the code, the name cellSize was ambiguous because it was either in a number of bricks or in meter, so now there are two fields: `cellSizeInMeter` and `cellSizeInBricks`

Changed the default number of bricks per cells per axis to 27 instead of 81

Updated the HelpBox in the probe profile asset to show distance between probes instead of subdivision levels: 
image

---
### Testing status
Baked the repro scene with the fix, works like expected
image

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fapv-structure-seams/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/6498835/job

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoring.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoring.cs
@@ -75,7 +75,7 @@ namespace UnityEngine.Experimental.Rendering
 
         internal ProbeReferenceVolumeProfile profile { get { return m_Profile; } }
         internal float brickSize { get { return m_Profile.brickSize; } }
-        internal float cellSizeInMeter { get { return m_Profile.cellSizeInMeters; } }
+        internal float cellSizeInMeters { get { return m_Profile.cellSizeInMeters; } }
         internal int maxSubdivision { get { return m_Profile.maxSubdivision; } }
         internal float normalBias { get { return m_Profile.normalBias; } }
 
--- a/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
+++ b/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
@@ -471,7 +471,7 @@ namespace UnityEngine.Experimental.Rendering
             UnityEditor.Experimental.Lightmapping.additionalBakedProbesCompleted += OnAdditionalProbesBakeCompleted;
 
             var volumeScale = bakingReferenceVolumeAuthoring.transform.localScale;
-            var CellSize = bakingReferenceVolumeAuthoring.cellSizeInMeter;
+            var CellSize = bakingReferenceVolumeAuthoring.cellSizeInMeters;
             var xCells = (int)Mathf.Ceil(volumeScale.x / CellSize);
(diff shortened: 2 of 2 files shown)


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -50,9 +48,10 @@ public enum CellSize
         [SerializeField]
         Version version = CoreUtils.GetLastEnumValue<Version>();
 
-        public int maxSubdivision => Mathf.CeilToInt(Mathf.Log((float)cellSize / brickSize, 3));
+        // We don't count the first subdivision because it have the same size as the cell so we add one.
```
FrancescoC-unity, line 51: Is this already valid before your subdiv changes?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -50,9 +48,10 @@ public enum CellSize
...
+        public int maxSubdivision => Mathf.CeilToInt(Mathf.Log((float)cellSizeInBricks, 3)) + 1;
         public float brickSize => Mathf.Max(0.01f, minDistanceBetweenProbes * 3.0f);
-        public int cellSize => Mathf.CeilToInt((float)cellSizeInBricks * brickSize);
+        public float cellSizeInMeter => (float)cellSizeInBricks * brickSize;
```
FrancescoC-unity, line 54: in meter_s_ plural here

JulienIgnace-Unity (approved):

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
alelievr: With this PR yes, adding one subdiv level will let the structure subdivide up to the cell level, and because the cell size is compatible with brick size (brickSize * 3^subdivLevel) now it's safe to do that (bricks won't be generated outside of the cell)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoring.cs
```diff
@@ -75,7 +75,7 @@ internal static ProbeReferenceVolumeProfile CreateReferenceVolumeProfile(Scene s
 
         internal ProbeReferenceVolumeProfile profile { get { return m_Profile; } }
         internal float brickSize { get { return m_Profile.brickSize; } }
-        internal float cellSizeInMeter { get { return m_Profile.cellSizeInMeter; } }
+        internal float cellSizeInMeter { get { return m_Profile.cellSizeInMeters; } }
```
FrancescoC-unity, line 78: Sorry for being a bit anal about it, but this is still singular :)

FrancescoC-unity (approved):

JulienIgnace-Unity:
As discussed with @alelievr , closing this PR as it's not complete (some problems are still present) and all this will also be fixed in a later PR refactoring the whole thing.

PR #4313, opened 2021-04-26
Skip adding the LOD's renderer to the checked renderer if it's null

remi-chapelain:
### Purpose of this PR
This PR fixes a [very specific use case with LOD and building the AS](https://fogbugz.unity3d.com/f/cases/1331881/). 
When adding LOD group to the Acceleration Structure, ray tracing manager will only add the top level LOD (0). 
However, for some reasons, a reference of each renderer is kept and if multiple LOD are using the same mesh, they have different renderer instance ID so they also need to be added to this reference list. 

If one of the LOD is missing a renderer, there's no check for null renderer so it make the whole building AS operation fails.

A check has been added to the "check content for raytracing" 
image. 

Also, after [this PR](https://github.com/Unity-Technologies/Graphics/pull/4082), we forgot to update the checks here and some were obsolete. Those have been updated with the new correct behavior about double sided-ness in multimat mesh having priority in the acceleration structure. 

This has been found in 2021.1 so it probably needs a backport in 11 and 10 

---
### Testing status
- Tested on the use case with one LOD missing a mesh renderer and made sure that it does not repro anymore. 
- Tested that a warning was being thrown with one or more missing renderer in an LODGroup
- Tested that no warnings were being thrown with no missing renderer in a LODGroup

---
### Comments to reviewers
N/A (purpose of the PR have been updated)

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -159,6 +159,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed CustomPassUtils scaling issues when used with RTHandles allocated from a RenderTexture.
 - Fixed ResourceReloader that was not call anymore at pipeline construction
 - Fixed undo of some properties on light editor.
+- Fixed null reference exception in Raytracing SSS volume component.
+- Fixed artifact appearing when diffuse and specular normal differ too much for eye shader with area lights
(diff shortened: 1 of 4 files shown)


sebastienlagarde (approved):
I am fine with the change, but it have indent issue and space (will fail yamato) and I am fine with adding this to the scene checker, good idea.

iM0ve (approved):
Looks good, minor change.

sebastienlagarde (approved):

PR #4318, opened 2021-04-26
Added an updater for shader graph with the builtin target that will u…

joshua-davis:
…dpate all materials using that shader graph when it's saved. This is a heavy-handed solution that should get updated when import dependencies are finished.

---
### Purpose of this PR
If a target was newly added to SG and a material was already using it, the material didn't get its keywords reset causing it to occasionally look wrong until you view the material in the inspector.

The easiest way to repro this is:
 - create a new SG with alpha clipping set to true that reads from a texture and sets the base color and alpha
 - create a material and set it on an object using some texture where you can verify alpha clipping works
 - go back to SG (with the inspector not visible) and add the built-in target (with alpha clipping on) and save
 - the object should properly work with alpha clipping (it won't in the old version)
additionally it can be more visible if you do half of this again, but with a new material:
 - remove the built-in target
 - create a new material and setup the texture on it and assign it to an object
 - go back to the SG editor and add the built-in target
This last version is important to test due to how SG works with targets. In particular, SG caches all targets and re-uses them when you activate/deactivate them, meaning any state stored on the target doesn't work (which is how URP/HDRP works and they have the same issue).

---
### Testing status
Tested the above.

---
### Comments to reviewers
Decided to make this always work for built-in by always running the material reset logic and not trying to filter based upon certain SG changes. This will catch all cases and only be a bit slower for larger projects (if using built-in) and changes are coming for import dependencies that should make this fast.

### DIFF
--- a/com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/Editor/ShaderGraph/Targets/BuiltInTarget.cs
+++ b/com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/Editor/ShaderGraph/Targets/BuiltInTarget.cs
@@ -265,7 +265,7 @@ namespace UnityEditor.Rendering.BuiltIn.ShaderGraph
         {
             m_ActiveSubTarget.value.ProcessPreviewMaterial(material);
         }
-        
+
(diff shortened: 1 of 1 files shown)


jessebarker (approved):
LGTM

PR #4320, opened 2021-04-27
Fix RTHandle scale bias [1329977]

skhiat:
Fix https://fogbugz.unity3d.com/f/cases/1329977/

Reprostep:
On Camera:
image

* Add an offset X of viewport rect.
* On SSR Enable 'PBR Accumulation'
* Increase the 'Max Ray Steps'
* On Game view increase the Accumulation factor

Before:
amWsfNAZ8H

After:
4qOvTMrHKS


JulienIgnace-Unity (changes requested):
This fix looks weird to me.
_RTHandleScaleHistory.zw / _RTHandleScaleHistory.xy does not really mean anything so I'm not sure how that works.
In theory there should be very little to do here:
- For current frame buffers, since we are using pixel coordinates, no scale should be applied as we render in the proper viewport already.
- For previous frame buffers, they may be a different size so pixel coordinates should be multiplied by _RTHandleScaleHistory.zw if they are history buffers or _RTHandleScale.zw if they are shared buffers.

skhiat:
> This fix looks weird to me.
> _RTHandleScaleHistory.zw / _RTHandleScaleHistory.xy does not really mean anything so I'm not sure how that works.
> In theory there should be very little to do here:
> 
> * For current frame buffers, since we are using pixel coordinates, no scale should be applied as we render in the proper viewport already.
> * For previous frame buffers, they may be a different size so pixel coordinates should be multiplied by _RTHandleScaleHistory.zw if they are history buffers or _RTHandleScale.zw if they are shared buffers.

[Fixed]

JulienIgnace-Unity (approved):

iM0ve (approved):
Looks good. Checking the repro case should be sufficient since it only affects SSR. I

PR #4182, opened 2021-04-14
[12.0.0] Create URP Graphics Test Package

phi-lira:
### Purpose of this PR
 This PR creates a URP graphics test package. This is useful for two reasons:
  - we don't duplicate scripts and assets between test projects
  - the package has the graphics test framework dependency so other teams / packages can integrate test with URP by adding the URP graphics test package.

### Testing status
 - Run tests locally.

### Comments to reviewers
 - Common assets and scripts were moved to the new graphics test package.
 - Each split project references the new graphics test package.
 - I could see some issues / warning in the test projects but they didn't seem related to my changes so I plan to fix them in follow PR.

PR #4321, opened 2021-04-27
Revert script changes to lighting test done in (#4182)

phi-lira:
### Purpose of this PR
 Revert scripting changes made to lighting test. in PR #4182 
 This PR should add back scripting files to Editor and Runtime folders of project and remove reference to the testing package.

---
### Testing status
 Tested locally 

---
### Comments to reviewers
 The reason to revert this files is because there's another PR that wants to add tests to the lighting project. When adding those for some reason an assembly with a different name from the one that was in the project before cause access issue to internal editor rendering assembly. I suspect this is related to some code giving access to internals to that specific assembly file that was on the project before and I couldn't figure out where. 
 
  Reverting so I unblock the other PR, this should be investigated imo later.


pigselated:
LGTM

pigselated (approved):
LGTM

PR #4322, opened 2021-04-27
Fix null exception on Raytracing SSS volume component

FrancescoC-unity:
There was missing a null check as on domain reload the asset might not be already be ready to be queried when the inspector GUI drawer is called. 

Fix for https://fogbugz.unity3d.com/f/cases/1330648

**What did I test**: the repro scene and a couple of other domain reload triggers (recompile code for example)


sebastienlagarde (approved):

iM0ve (approved):
Sufficient testing for the scope of the issue

PR #4323, opened 2021-04-27
Added define for new reflection probe API update (case 1290521)

fredericv-unity3d:
### Purpose of this PR
Fixes 1290521.

GC memory allocation was performed because of a previous workaround.
Appropriate code is now enabled with a define starting from 2021.2.0a16 and there are no more GC memory allocated.

---
### Testing status
1. Reproduced the issue without the code enabled.
  1. With an empty project, play and profile
  2. Look at  `PlayerLoop/ScriptableRuntimeReflectionSystemWrapper.Internal_ScriptableRuntimeReflectionSystemWrapper_TickRealtimeProbe()`, it should allocate 81B of GC memory.
2. Activate the appropriate code
  1. In the `Unity.RenderPipelines.HighDefinition.Runtime` `asmdef` the define `REFLECTION_PROBE_UPDATE_CACHED_DATA_AVAILABLE` must be applied for `Unity` starting from `2021.2.0a16` (this is what this PR is about)
3. Play and profile an empty project
  1. Observe that the GC memory allocation is 0 for that function call.

---


sebastienlagarde (approved):

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Checked HDRP Template project on latest trunk calling multiple On Demand probes. Seems good. No GC Allocs occur, probes remain to be updated correctly.

sebastienlagarde:
Note to self: retarget to master once hd/bugfix is merge

PR #4324, opened 2021-04-27
Opt-out of builtin auto baking of ambient/reflection probe.

JulienIgnace-Unity:
### Purpose of this PR
Opt-out of builtin auto baking of ambient/reflection probe.
This caused some issue when baking in HDRP.
https://fogbugz.unity3d.com/f/cases/1313990/

---
### Testing status
Tested the repro step in the fogbugz case

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -525,6 +525,7 @@ namespace UnityEngine.Rendering.HighDefinition
                 , overridesRealtimeReflectionProbes = true // Don't display the real time reflection probes checkbox UI in Quality Settings
                 , autoAmbientProbeBaking = false
                 , autoDefaultReflectionProbeBaking = false
+                , enlightenLightmapper = false
             };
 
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

iM0ve:
I tried to take a look at this and compare Master vs Branch. Not sure what it is fixing, on master I get exactly the same result before and after the fix. It's likely I'm doing something wrong, so maybe you could share screenshots with before and after, I could use them to evaluate and approve

Master
Unity_JQFnFquSeB

PR
Unity_5a9O0b7RTt

JulienIgnace-Unity:
Before
image

After
image

So what we fix here is that it's not blue. Basically it just shows us that we don't run useless systems meant for builtin renderer.
The fact that it's black is still a bug but will be fixed in a C++ PR that should land soon.

iM0ve (approved):
Thank you for the additional info. Looks good

sebastienlagarde:
green custom yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Fskymanager-optout/.yamato%252Fall-hdrp.yml%2523PR_HDRP_CUSTOM-REVISION/6561500/job/pipeline

PR #4326, opened 2021-04-27
Disable Enlighten lightmapper for HDRP.

JulienIgnace-Unity:
### Purpose of this PR
As enlighten has been revived in 21.2 it appears in the lightmapper choice. We don't want to allow its usage for baking so we needed to hide it in HDRP
https://jira.unity3d.com/browse/GFXGI-472

---
### Testing status
Looked at the lighting settings and confirmed Enlighten was unavailable.


sebastienlagarde (approved):

iM0ve (approved):
I additionally tested what would happen if Enlighten is already selected before project upgrade.
No warnings or errors. If Enlighten is already selected then it will remain so. After switching to another lightmpapper you are not allowed to switch back to Enlighten. New projects wont be able to choose Enlighten. The change should be safe.

kroatoa (approved):

PR #4327, opened 2021-04-27
Fix Ray Tracing Light Cluster debug mode

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1329129/

The issue stems from the fact that the Input layout included a POSITION that was not used anywhere in the shader. As a result the position was likely stripped out and a validation of the input layout failed in engine due to the mismatch what was declared and what was actually used. 

**What did I test**: the repro case


iM0ve (approved):
Since the change only affects the light cluster debug shader, testing the repro steps should be sufficient.

sebastienlagarde (approved):

PR #4328, opened 2021-04-27
Add Standalone tests for all graphics packages on Apple Silicon devices

theopnv:
### Purpose of this PR

- Added Standalone tests for all graphics packages. Several notes:
    - We have a limited amount of m1 devices so we can't run the tests on the PRs, and for all areas. Let's cover only HDRP, URP, Shadergraph and VFX Graph for now.
    - In order to spend the minimum amount of time on the m1 devices, we are building the player on an intel based machine and only running the tests on `Unity::metal::devkit` (m1 devices).
    - It's not yet possible to have editmode tests, so we'll have to wait before running some (DM if you want more details).
    - I commented the lines adding the jobs to the nightly because the tests need fixes from the development teams to be green. Infra (yamato, bokken) works fine but the tests are broken. Let's ask all graphics teams to fix them before adding them to the nightly.

---
### Testing status
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/yamato%252Fadd-apple-silicon-tests-refactor

### DIFF
--- a/.yamato/all-hdrp.yml
+++ b/.yamato/all-hdrp.yml
@@ -60,8 +60,6 @@ Nightly_HDRP_trunk:
          rerun: on_new_revision
       -  path: .yamato/hdrp-osx-metal.yml#HDRP_OSX_Metal_Standalone_cache_mono_Linear_trunk
          rerun: on_new_revision
-      -  path: .yamato/hdrp-osx_apple_silicon-metal.yml#HDRP_OSX_Apple_Silicon_Metal_Standalone_cache_mono_Linear_trunk
-         rerun: on_new_revision
     variables:
         CUSTOM_REVISION: '{{trunk.changeset.id}}'
         UTR_VERSION: "current"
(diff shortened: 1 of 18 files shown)


seanstolberg-unity:
Hi Theo, is our intention with this PR to add Silicon support/coverage, or replace Intel with Silicon? I ask because I believe we still want to have configurations for Mac/Intel and also add Silicon coverage. Some of the code changes I saw here made me think we were replacing Mac/Intel with Mac/Silicon, which I don't think is what we want to do because our customers will expect we test both.

theo-pnv:
Hi @seanstolberg-unity, thanks for following up on this one. The intent is indeed to keep both. Could you point me to the changes you think would replace the intel tests? If there are some, they're not intended.
To add more context, this PR's goal is to have apple silicon standalone tests running nightly for our 4 graphics packages (hdrp, urp, vfx, shadergraph).
And now about the PR status: The yamato definitions are in good shape and I'm able to run the tests, but we need new reference images because they are different from the ones needed for intel based tests. A current limitation of the graphics test framework is that we can't add this new folder yet (named after the processor type), so I need to make a PR there to add this new category/folder. Then we can add new reference images in the "silicon" folder and merge this PR.

PR #4331, opened 2021-04-27
Fixed Look Dev breaking when docked and rendered at zero size.

JulienIgnace-Unity:
### Purpose of this PR
Fixed Look Dev breaking when docked and rendered at zero size.
To do so, we just added a minimum size to the rendering panel of the look dev.

https://fogbugz.unity3d.com/f/cases/1327721/

---
### Testing status
Tested the repro described in the case.


RSlysz (approved):

sebastienlagarde (approved):

iM0ve (approved):
Minor change, testing the repro case should be enough.

PR #4335, opened 2021-04-27
Fix erroneous tile coordinate calculations in contact shadows

FrancescoC-unity:
Wrong parenthesis strikes again :) 

Because both values being divided where constant integer and the division was in brackets, the tile coordinate was always computed as 0, leading to unexpected behaviour when looping through the lights. 

This fixes an issue reported by Alix on slack ( https://unity.slack.com/archives/GHD5LADU7/p1619513674480600 ) 

**What did I test**: the repro scene Alix provided me slightly modified. Essentially to repro is sufficient to have a light that has contact shadows and that as the camera moves such light hits or not the first "lightlist" tile.


sebastienlagarde:
@FrancescoC-unity can we add a contact shadow test that cover this cases? Hard to figure out how this can't be catch by our automated test. Thanks

FrancescoC-unity:
> @FrancescoC-unity can we add a contact shadow test that cover this cases? Hard to figure out how this can't be catch by our automated test. Thanks

Yes, we can indeed, will add one soon.

iM0ve (approved):
Looks good

sebastienlagarde (approved):
Waiting for test

FrancescoC-unity:
I managed to change a test to cover this test, gonna run yamato and update screenshots as soon is done.

PR #4340, opened 2021-04-27
[Fogbugz 1324000] Fixing unlit emissive reflections

kecho:
### Purpose of this PR
Fixing fogbugs: https://fogbugz.unity3d.com/f/cases/1324000/
Unlit emissive was too dark, because there was a missing pre exposure multiplication.

Repro steps : 
- Open attached project
- Open SampleScene_Exposure0
- Notice that both lit and unlit ray traced reflection looked similar
- Now open SampleScene_Exposure_12
- It's the same scene with fixed exposure of 12, HDRI sky exposed to 12 and higher intensity dir light

Expected : Same result as the non exposed scene
Result : Unlit emissive is black. 

Before:
image

After:
image

---
### Testing status
* Tested with many levels of exposure in the unlit material, and now the material behaves correctly.
* Tested on PC with auto exposure and several fixed exposure values.
* Running all platforms tests (DXR and standard). Likely we might need screenshot update.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -166,6 +166,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed issue with RAS build fail when LOD was missing a renderer
 - Fixed an issue where sometime a docked lookdev could be rendered at zero size and break.
 - Fixed an issue where runtime debug window UI would leak game objects.
+- Fixed NaNs when denoising pixels where the dot product between normal and view direction is near zero (case 1329624).
 - Fixed ray traced reflections that were too dark for unlit materials. Reflections are now more consistent with the material emissiveness.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


sebastienlagarde:
@remi-chapelain can you ensure this is cover in one of our test. Also guess we need to check for both unlit with emissive and lit with emissive.

sebastienlagarde:
@kecho in your test, is your mesh emissive only or does it have a color + emissive contribution? (we should check the later if not check)

kecho:
> @kecho in your test, is your mesh emissive only or does it have a color + emissive contribution? (we should check the later if not check)

My test is emissive only, black color. I did test with color as well and looked correct.

remi-chapelain:
I'll add a test in DXR Test project in this PR as stated in the bug and take care of screenshots if some needs updating.

sebastienlagarde (approved):

remi-chapelain (approved):
Verified the fix quickly. 

Added 4 tests to cover reflections of emissive
- Low exposure perf mode
- high exposure perf mode
- Low exposure quality mode
- high exposure quality mode

108_ReflectionEmissive_Perf_Exposure_0

108_ReflectionEmissive_Quality_Exposure_0

High and low exposure look the same because the value of emissive and sun are adjusted accordingly. Only noticeable difference is the reflection resolution in performance vs quality; 

Ran all DXR Tests locally (All 🟢, which proved we didn't have proper coverage for this) since yamato is overloaded, I'd say we're good to merge without testing on yamato

PR #4341, opened 2021-04-27
Fixed property order for built-in target to match URP

joshua-davis:
---
### Purpose of this PR
Fixed the property order on the built-in target to match the order of properties on URP. Also renamed a few fields to match (e.g. Surface -> Surface Type)

---
### Testing status
Verified the properties were in the right order afterwards

---
### Comments to reviewers
This is basically a cosmetic fix, no real behavioral changes.


jessebarker (approved):
LGTM

PR #4342, opened 2021-04-27
[HDRP] Fix NaNs when denoising pixels with near zero dot(normal, view) 

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1329624/
The ticket mentions ortographic camera, but given the nature of the computations, I think this can occur with perspective projection too.

---
### Testing status
The issue does not repro anymore with the test project in the bug ticket.

---
### Comments to reviewers
An alternative could be to cap the return value of the ComputeMaxReprojectionWorldRadius function, but I think avoiding the division by zero is a proper solution.


sebastienlagarde (approved):

FrancescoC-unity (approved):

iM0ve (approved):
Change is specific to DXR denoising, testing of the repro case should be sufficient.

PR #4343, opened 2021-04-27
Added menu items for creating lit and unlit shader graphs with the bu…

joshua-davis:
…ilt-in target

---
### Purpose of this PR
Currently the only way to create a shader graph for built-in is either through another target's menu option or through a blank shader graph. This adds a menu option to create both lit and unlit built-in target shader graph.

---
### Testing status
Tested the create menu. Had to fiddle with the priorities a bit to get no separator to show up. Currently the menu looks like:
image
No idea if someone cares about the relative ordering with URP or other menu options.

---
### Comments to reviewers


jessebarker (approved):
LGTM

PR #4344, opened 2021-04-27
Add skeleton for new shadow setting

FrancescoC-unity:
Adding a new shadow quality tier to allow the work for PCSS in area lights. 

Also added abstraction for AREA_LIGHT_FILTERING so that it can be changed to slot in a new method.

I did test the change by running all tests locally and they are all green. 

For the time being Very High and High are exactly the same (waiting for @robertcupisz PR)


com.unity.render-pipelines.high-definition/Editor/VFXGraph/Shaders/Templates/PlanarPrimitive/PassForward.template
```diff
@@ -10,7 +10,7 @@ Pass
 	#define LIGHTLOOP_TILE_PASS
 	${VFXHDRPForwardDefines}
 	#pragma multi_compile USE_FPTL_LIGHTLIST USE_CLUSTERED_LIGHTLIST 
-	#pragma multi_compile SHADOW_LOW SHADOW_MEDIUM SHADOW_HIGH
+    #pragma multi_compile SHADOW_LOW SHADOW_MEDIUM SHADOW_HIGH SHADOW_VERY_HIGH
```
FrancescoC-unity, line 13: Not sure why it looks differently indented here, can't see to get this to change locally with fork :/

sebastienlagarde: Maybe running the formatting tool can work?

sebastienlagarde (approved):
Note, we will need to update documentation as well

com.unity.render-pipelines.high-definition/Editor/VFXGraph/Shaders/Templates/PlanarPrimitive/PassForward.template
FrancescoC-unity: I did try but did not work :-(

FrancescoC-unity:
> Note, we will need to update documentation as well

Will add soon

FrancescoC-unity:
> Note, we will need to update documentation as well

I was wondering, shall we update now or when there will be an actual difference? As of now it'd be a bit silly to say "Very High: it is the same as High"

sebastienlagarde:
good poitn, ok let's merge this one

sebastienlagarde:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fvery-high-shadow-setting/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/6622085/job

PR #4345, opened 2021-04-27
Add Volumetric Cloud Compatibility with MSAA

johnpars:
### Purpose of this PR
Fix case: https://fogbugz.unity3d.com/f/cases/1327436/

Volumetric Clouds will fail in the case of MSAA color target. This is due to the in-place blending operation that occurs on the upscaling + combine pass (which can't be done on an MSAA target). This PR will move that combine to a subsequent pass after the upscaling, and instead do the blending via the hardware on the MSAA target. 

Volumetric clouds now work in both scenarios:

No MSAA:
image

MSAA:
image

---
### Testing status
- I locally tested switching the number of MSAA samples on different graphics APIs. 
- I also ensured that the hardware blending operation produces the exact same result as when it is done in the CS pass. 
- [Yamato](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-1327436)

---
### Comments to reviewers
~


RemyUnity (approved):
Verified and fixed confirmed, thanks !

sebastienlagarde:
Something bad have happen with hd/bugfix, I have delete and recreated the branch but then we have thing that should be there.

those files aren't part of this PR:
Eye/Eye.hlsl
HDRenderPipelineMenuItems.cs
Motion-Vectors.md

We will need to recreate this PR against a proper master, sorry about this .Will also allow us to do a proper review (this fix is for master only, so let's recreate pr for master)

PR #4356, opened 2021-04-28
Fix for volumetric clouds performing Read and Write on same texture [Unsupported on some platforms]

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1316069/

As the title mentioned, clouds where not supported on some platforms (Metal and I assume Intel GPUs) as they were performing Read/Write on the same resource which format is 11-11-10. 

This change will perform an extra copy of the color buffer to use in those cases and sample that instead. 

I did try the following alternatives: 
 - Write full res clouds in the upsample without combining and then blending separately. [Slower]
 - Try to rewrite the shader from compute to pixel [Stop soon enough given that there are _a lot_ of samples so, LDS will be a big benefit] 

This option turned out to be the cheapest. It adds on PS4-base [where it won't run anyway, just used to profile] around 0.15ms of cost. 

**What did I test** forced the option on and off on both PC (NVIDIA card) and PS4 and check that same result is there. 

**NOTE FOR QA**: I don't have a dev-ready Metal machine to test, so we need to test on metal.

### DIFF
--- a/TestProjects/HDRP_DXR_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D12/None/108_ReflectionEmissive_Perf_Exposure_0.png
+++ b/TestProjects/HDRP_DXR_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D12/None/108_ReflectionEmissive_Perf_Exposure_0.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:0cbde868376b0cf742e0f09df78cbadb6f9d3a02bf10661ff38b08aa9176ef15
+size 130093
--- a/TestProjects/HDRP_DXR_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D12/None/108_ReflectionEmissive_Perf_Exposure_0.png.meta
+++ b/TestProjects/HDRP_DXR_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D12/None/108_ReflectionEmissive_Perf_Exposure_0.png.meta
@@ -1,0 +1,109 @@
+fileFormatVersion: 2
+guid: beb77fa7727d7334ba5404829e36f871
+TextureImporter:
+  internalIDToNameTable: []
+  externalObjects: {}
+  serializedVersion: 11
+  mipmaps:
+    mipMapMode: 0
+    enableMipMap: 0
+    sRGBTexture: 1
+    linearTexture: 0
+    fadeOut: 0
+    borderMipMap: 0
+    mipMapsPreserveCoverage: 0
+    alphaTestReferenceValue: 0.5
+    mipMapFadeDistanceStart: 1
+    mipMapFadeDistanceEnd: 3
+  bumpmap:
+    convertToNormalMap: 0
+    externalNormalMap: 0
+    heightScale: 0.25
+    normalMapFilter: 0
+  isReadable: 1
+  streamingMipmaps: 0
+  streamingMipmapsPriority: 0
+  vTOnly: 0
+  grayScaleToAlpha: 0
+  generateCubemap: 6
+  cubemapConvolution: 0
+  seamlessCubemap: 0
+  textureFormat: 1
+  maxTextureSize: 2048
+  textureSettings:
+    serializedVersion: 2
+    filterMode: 1
+    aniso: 1
+    mipBias: 0
+    wrapU: 0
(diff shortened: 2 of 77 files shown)


iM0ve (commented):
**TL:DR**
Its now working but there are still some issues. Not sure if I should approve and we address them later or we want to fix them within this PR

**Whats tested:**
* Checked that every Volumetric Clouds parameter works on Windows
* Checked that every Volumetric Clouds parameter works on Mac
* Built player on mac to see that the clouds are present

Concerns:
* Only on Scene view, only on mac. There is this weird ring under the clouds
https://user-images.githubusercontent.com/36502659/116417194-b6ea0f80-a843-11eb-8621-88461618fd08.mov


* On my mac the framerate is really slow on a build. Its way faster in the editor
Build
https://user-images.githubusercontent.com/36502659/116417697-32e45780-a844-11eb-94cb-f0f0a6e2b37a.mov

Editor

https://user-images.githubusercontent.com/36502659/116418057-776ff300-a844-11eb-84dc-ce2420adc090.mov

FrancescoC-unity:
> **TL:DR**
> Its now working but there are still some issues. Not sure if I should approve and we address them later or we want to fix them within this PR
> 
> **Whats tested:**
> 
> * Checked that every Volumetric Clouds parameter works on Windows
> * Checked that every Volumetric Clouds parameter works on Mac
> * Built player on mac to see that the clouds are present
> 
> Concerns:
> 
> * Only on Scene view, only on mac. There is this weird ring under the clouds
>   https://user-images.githubusercontent.com/36502659/116417194-b6ea0f80-a843-11eb-8621-88461618fd08.mov
> * On my mac the framerate is really slow on a build. Its way faster in the editor
>   Build
>   https://user-images.githubusercontent.com/36502659/116417697-32e45780-a844-11eb-94cb-f0f0a6e2b37a.mov
> 
> Editor
> 
>  Screen.Recording.2021-04-28.at.17.08.53.mov

I'd say those are likely different issues I'd open new tickets for those.

johnpars:
Hi! I just realized that this PR potentially fixes the same issue as this one: #4345 (however with the focus on MSAA). 
The solution in this PR is better (I perform hardware blending from an intermediate target into the MSAA target), and maybe we can expand it to handle the MSAA case and I can close the other one?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricClouds.cs
```diff
@@ -444,13 +439,21 @@ VolumetricCloudsParameters PrepareVolumetricCloudsParameters(HDCamera hdCamera,
             parameters.historyValidity = historyValidity;
             parameters.planarReflection = (hdCamera.camera.cameraType == CameraType.Reflection);
 
+            parameters.needExtraColorBufferCopy = (GetColorBufferFormat() == GraphicsFormat.B10G11R11_UFloatPack32 &&
+                // On PC and Metal, but not on console.
+                (SystemInfo.graphicsDeviceType == GraphicsDeviceType.Direct3D11 ||
+                    SystemInfo.graphicsDeviceType == GraphicsDeviceType.Direct3D12 ||
+                    SystemInfo.graphicsDeviceType == GraphicsDeviceType.Metal ||
+                    SystemInfo.graphicsDeviceType == GraphicsDeviceType.Vulkan));
```
johnpars, line 447: If we can add one more check here for MSAA, I think we can close #4345

sebastienlagarde: yes good point, MSAA fix should be integrate in this PR

sebastienlagarde:
Note to self: redirect the PR to master once hd/bugfix is merge

sebastienlagarde (approved):

FrancescoC-unity:
Actually @johnpars the problem is slightly different here as blending in place still is happening. I just avoid the Read+Write from same texture, but here the issue is still that with MSAA we'd need to change the UAV declaration to be multi-sampled and do per-sample blending. 

I think for MSAA your solution is probably still more efficient.

iM0ve (approved):
Today I opened my project again to file the fogbugz for concerns. The artefact is just gone and the framerate stabilized after I waited for ~20 seconds. So no more outstanding issues from my part. All looks good

sebastienlagarde:
will wait for discussion with John

PR #4359, opened 2021-04-28
Global Camera constants are now pushed when doing a custom render callback.

JulienIgnace-Unity:
### Purpose of this PR
When users register a custom render callback to a camera, they want to be able to use regular HDRP shaders. For this to work more easily, we need to update and bind the global constant buffer for the camera properties. This is what this PR does.
To achieve that, we needed to add some command buffer-less API to the constant buffer class.

---
### Testing status
Pending

---
### Comments to reviewers
@FrancescoC-unity for awarness on CB API change.


FrancescoC-unity (approved):

etienne-p (approved):

etienne-p:
Tried it out with our raycaster with our reflection code removed and it passed the tests!

sebastienlagarde (approved):

PR #4361, opened 2021-04-28
[HDRP]Fix history buffer allocation for AOVs when the request does not come in first frame

pmavridis:
### Purpose of this PR
Problem reported in internal slack thread:
https://unity.slack.com/archives/C6Y79CZM0/p1619541999268500

When the AOV request does not come in the first frame (there are good reasons to do that), there was a problem with allocating the per-AOV history buffers.

---
### Testing status
- I have tested Recorder 3.0.0-pre.1 from pacman and the problem does not repro anymore.
- AOV tests in HDRP are green.


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

TomasKiniulis (approved):
Seems good. Checked on AOV recorder as well with various Aov Export and Sequence settings. Works as expected now.

PR #4364, opened 2021-04-28
Fix fog precision

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1329603/

fog has precision issue when camera rotation is 0 on the X axis
before: notice the blue line that shouldn't be there
image

---
### Testing status

tested the repro project


sebastienlagarde (approved):

sebastienlagarde:
@Unity-Technologies/gfx-qa-hdrp try to check if you can reproduce error on volumetric fog at 0 axis with multiple view. thanks

iM0ve (approved):
Looks good, minor change. 
Please if you could next time add before and after screenshots. From just having this one its hard to understand what is being shown(fixed or before fix) and what is the problem. Thanks :)

PR #4366, opened 2021-04-28
Fixed emission material properties not being animable

alelievr:
### Purpose of this PR
Fix the emission properties of material not being animable from the inspector (https://fogbugz.unity3d.com/f/cases/1328108/)

Note that even though now we can animate these properties, the feature is STILL broken because of another issue:

https://user-images.githubusercontent.com/6877923/116433392-be141c00-a849-11eb-989c-6f89e4413d0b.mp4

As you can see in the video, the intensity of the emission is completely different when animated. But this has nothing to do with animation, in fact is seems to come from the material property block of the renderer. I managed to repro the issue with this simple script:
```csharp
[ExecuteAlways]
public class SetEmission : MonoBehaviour
{
    [ColorUsage(false, true)]
    public Color emissionColor;

    public bool useMaterialPropertyBlock = false;

    new MeshRenderer renderer;

    void Start()
    {
        renderer = GetComponent<MeshRenderer>();
    }

    void Update()
    {
        if (useMaterialPropertyBlock)
        {
            var p = new MaterialPropertyBlock();
            p.SetColor("_EmissiveColor", emissionColor);
            renderer.SetPropertyBlock(p);
        }
        else
        {
            var p = new MaterialPropertyBlock();
            renderer.SetPropertyBlock(p);
            renderer.sharedMaterial.SetColor("_EmissiveColor", emissionColor);
        }
    }
}
```

new fogbugz about the material property block issue: https://fogbugz.unity3d.com/f/cases/1332471/

---
### Testing status
manual tests (see video)

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -169,6 +169,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed NaNs when denoising pixels where the dot product between normal and view direction is near zero (case 1329624).
 - Fixed ray traced reflections that were too dark for unlit materials. Reflections are now more consistent with the material emissiveness.
 - Fixed pyramid color being incorrect when hardware dynamic resolution is enabled.
+- Fixed SSR Accumulation with Offset with Viewport Rect Offset on Camera
 - Fixed material Emission properties not begin animated when recording an animation (case 1328108).
 
 ### Changed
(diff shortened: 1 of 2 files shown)


adrien-de-tocqueville (commented):
I believe the decal material editor has the same issue because they also use UpdateEmissiveColorLDRFromIntensityAndEmissiveColor

alelievr:
It's not possible to animate material on the decal projector component, this feature only works with Renderers (which is a shame)

adrien-de-tocqueville (approved):

sebastienlagarde (approved):

iM0ve (approved):
Thanks for providing the video of testing. I can see the emission still works and now we can animate the property. Also new bug filled about the intensities, nice!

PR #4370, opened 2021-04-28
[Fogbugz 1330769] Fixing bloom bicubic filtering for DRS

kecho:
### Purpose of this PR
Bicubic filtering of bloom was using the wrong dimensions / coordinates for bloom. This PR applies the correct scaling to such parameters.
Fogbugz: https://fogbugz.unity3d.com/f/cases/1330769/
Repro steps in fogbugs

Tested, hardware drs, percentage of 10 to exaggerate artifact:

Before:
image

After:
image

---
### Testing status
* Tested on hardware DRS, vulkan and dx12 PC
* Should be tested in consoles (one is enough, i.e. PS4)


sebastienlagarde (approved):

FrancescoC-unity:
Did you try with SW dyn res too? I am a bit confused as HW dynamic res should have those constants you multiply with should be too far off 1 if a single view is there.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -3165,6 +3165,13 @@ void PrepareBloomData(RenderGraph renderGraph, in RenderGraphBuilder builder, Bl
             // the mip up 0 will be used by uber, so not allocated as transient.
             m_BloomBicubicParams = new Vector4(passData.bloomMipInfo[0].x, passData.bloomMipInfo[0].y, 1.0f / passData.bloomMipInfo[0].x, 1.0f / passData.bloomMipInfo[0].y);
             var mip0Scale = new Vector2(passData.bloomMipInfo[0].z, passData.bloomMipInfo[0].w);
+
+            //apply DRS on bloom bicubic params
+            m_BloomBicubicParams.x /= RTHandles.rtHandleProperties.rtHandleScale.x;
```
FrancescoC-unity, line 3170: So these parameters  _should_ be baked in the actualWidth/actualHeight used here https://github.com/Unity-Technologies/Graphics/pull/4370/files#diff-617165acfec9a5b4c1b26809291f30938d7a152b8e9fec0e5ae2c661078201dbR3148 , by dividing we are undoing that and assuming full resolution source?

kecho: yes! we want exactly that, the actual full resolution of the texture. To understand why, you need to look at bicubic sampling. Thats were these params are used. Bicubic sampling function requires texel size to compute the offsets, but the texel size must be that of the physical texture (so the uv can be offset correctly).
If we use instead say texture.width here, then we have to write special code for hardware drs, because hardware drs hides the render - time resource size. So the easiest way is to just 'undo' the texture scale.

kecho:
> Did you try with SW dyn res too? I am a bit confused as HW dynamic res should have those constants you multiply with should be too far off 1 if a single view is there.
> 
> Changes look good I am just a tad confused on why the big difference, I guess we have another camera active? Or maybe depending on starting res the rounding is relevant as going down the mip chain?

Yes so the answer is that hardware res can also have a possible rtHandle scale that is not 1, due to for example planar reflections being present and sharing the same large core textures.
The problem is also aggravated by bicubic sampling, the offsets you see are actually for bicubic sampling, and the offsets are in iv space, so something like 1/hardware width. If that doesnt happen, you can get huge jumps and stair stepping.
I exagerated the problem in the screen shots, used a %10 scale to show the insanely bad offsets

FrancescoC-unity:
> > Did you try with SW dyn res too? I am a bit confused as HW dynamic res should have those constants you multiply with should be too far off 1 if a single view is there.
> > Changes look good I am just a tad confused on why the big difference, I guess we have another camera active? Or maybe depending on starting res the rounding is relevant as going down the mip chain?
> 
> Yes so the answer is that hardware res can also have a possible rtHandle scale that is not 1, due to for example planar reflections being present and sharing the same large core textures.
> The problem is also aggravated by bicubic sampling, the offsets you see are actually for bicubic sampling, and the offsets are in iv space, so something like 1/hardware width. If that doesnt happen, you can get huge jumps and stair stepping.
> I exagerated the problem in the screen shots, used a %10 scale to show the insanely bad offsets

Yes I got that, but that implies that the problem is not there without extra cameras without this PR no? Does it go away?

The SW should show the issue even more for the same rationale (texture width differes from operating width a fair bit)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
kecho: @FrancescoC-unity I have added a comment on this snippet to explain it

FrancescoC-unity: Yup it does make sense 👍

FrancescoC-unity (approved):

RemyUnity (approved):
Tested and fixe confirmed on PC.
@TomasKiniulis  will check on console for final approval.

TomasKiniulis (approved):
Testing repro project on Xbox looks all good as well

PR #4371, opened 2021-04-28
[2021.2][ShaderGraph] Make sure to generate Vector1 property block string as an integer

jessebarker:
### Checklist for PR maker

---
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1330302/.

We are generating integer mode Vector1 properties as floats in the serialized shadergraph, which exposes a float to the material inspector.  We previously (7.x of Shader Graph) actually serialized the truncated value, so the material inspector looked correct, but it also lost the original default value (so, if the user set a default of '0.5' before switching to integer mode and then switched back later, the default would now be '0.0').  This fix preserves the original default value as well as generating the correct property block string.

---
### Testing status
Manual testing according to the repro steps in the bug.
Yamato tests to follow.

---
### Comments to reviewers


jessebarker:
This disappeared when I resolved the conflict in the changelog, but there have been no code changes:
[ShaderGraph PR Job on Yamato](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg-fix-1330302/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/6575335/job)

joshua-davis (approved):

PR #4378, opened 2021-04-29
Find APV index dimension across scenes instead of having per-asset one

FrancescoC-unity:
This PR automagically computes the index buffer size needed for all the scenes. In the current stage this is looking at all the scenes in the build settings.

Eventually we will have "bake groups" so we can have multiple subset of scenes to be considered. This is because we might have independent scenes groups that are never loaded together and so can have different settings. But that's for a future phase. 

This is a first step towards making the reference volume component irrelevant. 
I originally planned to make the PR larger, but some problems are trickier than they appear so I prefer to do more smaller PR.

What this is mostly going to be useful for [and what I am working on finding more difficulties than I thought] avoiding to reload all assets at each additive scene addition like we need now.

### DIFF
--- a/TestProjects/HDRP_Tests/ProjectSettings/EditorBuildSettings.asset
+++ b/TestProjects/HDRP_Tests/ProjectSettings/EditorBuildSettings.asset
@@ -251,7 +251,7 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/GraphicTests/Scenes/2x_Lighting/2009_MultipleSkies.unity
     guid: bdc9e712cf2e77a49a64859ccb7ba8d7
-  - enabled: 1
+  - enabled: 0
     path: Assets/GraphicTests/Scenes/2x_Lighting/2010_IES_Cookies.unity
     guid: ad7e365098dfb004aba8f7e35924b5ce
   - enabled: 1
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -125,6 +172,8 @@ private static ProbeReferenceVolumeAuthoring GetCardinalAuthoringComponent(Probe
 
         private static void OnBakeStarted()
         {
+            FindWorldBounds();
```
alelievr, line 175: Hum, so it means that every time we bake, we have to load all the scenes of the project one by one?
I'm afraid this will slow down a lot the baking, especially on big projects :/

FrancescoC-unity: Couple of points: 

 - This is the only way to avoid re-loading all assets every time a single scene is loaded. That is a potentially huge _runtime_ cost. 
 - This will need to be done on real projects anyway most likely given that any 2 or more scenes that at least share a single cell will need to be baked together to avoid issues. 
 -  We'll  add "bake groups" later on so we don't load _all_ scenes, but only the ones in  the bake group.

FrancescoC-unity: If this becomes a sensible overhead, we could bake a small asset containing those world bounds info into a separate file (do it once only and update upon scene saved) and use that instead. 

That is a bit overkill though and we have bigger issues now :-) However there are ways when this will become the problem.

JulienIgnace-Unity (approved):

FrancescoC-unity:
Apparently this had problems on yamato, so I created a new PR starting from new master. https://github.com/Unity-Technologies/Graphics/pull/4440

PR #4030, opened 2021-03-29
[2021.2][ShaderGraph] Add Shader Graph Target for the built-in render pipeline (in progress)

jessebarker:
### Checklist for PR maker

- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [x] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [x] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR

This PR adds a Shader Graph Target and related SubTargets for the built-in render pipeline.

We realize that our asset publishing ecosystem currently suffers from the inability to generate shaders for each of the render pipelines that we officially support.  Adding this built-in target allows us to generate shaders for any of the render pipelines from a single Shader Graph (with the caveat that the pipelines do not share a common feature set so not all features can be expected to work on all pipelines).

The sub-targets currently supported are a simple Unlit (Create->Unlit Shader Graph), and a simple Lit (Create->Lit Shader Graph) with metallic workflow (specular workflow is a future task).  These currently work with the forward and deferred passes of the built-in pipeline.

For reference:
[Roadmap Epic](https://unity3d.atlassian.net/browse/UR-1990)
[Shader Graph Epic](https://jira.unity3d.com/browse/GSG-175)

---
### Testing status

Tech Art has made a couple of passes with this branch around the following workflows:
- Basic 3D (built-in only) project, adding Shader Graph package and targeting built-in alone.
- URP 3D project, adding the built-in target to graphs and targeting both pipelines.

We have added 2 small test projects to support this work (all play mode only):
- [x] Clone of the URP foundation tests adapted and configured for the built-in target
- [x] Clone of the URP lighting tests adapted and configured for the built-in target

[...]

PR #4379, opened 2021-04-29
[VFX] reimport on SRP changed + remove built-in support and notify

julienf-unity:
### Purpose of this PR
This PR uses the new SRP changed callback to reimport automatically VFX on SRP switch
It also removes built-in support (and all related code) and issue a warning if current pipeline is not supported by VFX 

**Limitation**
There are a couple of cases where some VFX data can be lost when switching pipeline. Mostly slots value that are spawned based on some SRP caps (alphathreshold value for transparent outputs with motion vectors or shadows for instance)
Switiching SRP should not alter VFX asset in any way and this should be adress in another PR.

**Side note**
A callback should be added on SRP switch to inform user that some reimports will occur and this might take time

---
### Testing status
Tested locally by switching pipeline asset

---
### Comments to reviewers
This PR requires a C++ custom branch: https://ono.unity3d.com/unity/unity/pull-request/125169
At the moment this branch fails to compile with laltest master. I asked for a trunk merge
But if you encouter the compilation, just commenting out those in HDRenderPipeline.cs does the trick:
` // , autoAmbientProbeBaking = false` 
` // , autoDefaultReflectionProbeBaking = false` 
` // , enlightenLightmapper = false`

### DIFF
--- a/com.unity.visualeffectgraph/Editor/Compiler/VFXGraphCompiledData.cs
+++ b/com.unity.visualeffectgraph/Editor/Compiler/VFXGraphCompiledData.cs
@@ -991,15 +991,14 @@ namespace UnityEditor.VFX
                 if (contextToSystemIndex.TryGetValue(data, out var index))
                     yield return index;
         }
-      
+
         private void CleanRuntimeData()
         {
             if (m_Graph.visualEffectResource != null)
                 m_Graph.visualEffectResource.ClearRuntimeData();
 
             m_ExpressionGraph = new VFXExpressionGraph();
-            m_ExpressionValues = new VFXExpressionValueContainerDesc[] { };
-            return;
+            m_ExpressionValues = new VFXExpressionValueContainerDesc[] {};
         }
 
         public void Compile(VFXCompilationMode compilationMode, bool forceShaderValidation)
--- a/com.unity.visualeffectgraph/Editor/Data/VFXDataMesh.cs
+++ b/com.unity.visualeffectgraph/Editor/Data/VFXDataMesh.cs
@@ -106,7 +106,7 @@ namespace UnityEditor.VFX
             {
                 m_CachedMaterial = new Material(shader);
                 m_CachedMaterial.hideFlags = HideFlags.HideAndDontSave;
-                
+
                 VFXLibrary.currentSRPBinder?.SetupMaterial(m_CachedMaterial);
             }
 


landonth:
> This PR uses the new SRP changed callback to reimport automatically VFX on SRP switch
> It also removes built-in support (and all related code) and issue a warning if current pipeline is not supported by VFX

@julienf-unity I don't see any problem with removing BRP (Built-in RP) support as long as #4030 is merged before this PR (or before a final release.) As I understand, it will maintain VFX Graph compatibility with BRP (in a new, render pipeline agnostic way,) now that #3467 has been merged?

I suggest to maintain continuous compatibility if at all possible, for to avoid a similar situation Unity had with Enlighten real-time GI support for HDRP:

- x version 1.1 has legacy support for feature y
- x version 1.2 removes legacy support for feature y
- x version 1.3 adds new support for feature y

Ideally, the 2nd and 3rd step are swapped, avoiding final releases where the feature support is effectively missing. This could also be mitigated with backporting, but it seems it could be avoided entirely.

julienf-unity:
@landonth SG supporting built-in does not imply VFX supporting builtin. Even with the new SG integration in VFX, we still have simulation shaders to generate that relies on SRP shader library. So this PR means no builtin support in VFX.

VladNeykov:
Hi @julienf-unity , got some things jotted down on the [test doc](https://docs.google.com/document/d/18AM79g_OEHxVN-Xg0wNtuRucfMZJqCBLycSpHPUJdXM/edit#) to take a look at when you get a chance.

VladNeykov (approved):
Looking good, one small (possibly unrelated) occasional console message case can be investigated down the road (noted on the [test doc](https://docs.google.com/document/d/18AM79g_OEHxVN-Xg0wNtuRucfMZJqCBLycSpHPUJdXM/edit#)).
Otherwise tried switching between HDRP-URP-Built-in in various orders with many VFXs and it works as expected.
Attempting to use VFX in built-in provides an informative console message in the appropriate most common scenarios (SRP switch, project opening, creating new VFX, opening a VFX, and compiling a VFX).
No outstanding issues.

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
```diff
@@ -489,21 +482,66 @@ private static void LoadSRPBindersIfNeeded()
             }
         }
 
+        private static bool unsupportedSRPWarningIssued = false;
+
+        private static void LogUnsupportedSRP(VFXSRPBinder binder, bool forceLog)
+        {
+            if (binder == null && (forceLog || !unsupportedSRPWarningIssued))
+            {
+                Debug.LogWarning("The Visual Effect Graph is supported in the High Definition Render Pipeline (HDRP) and the Universal Render Pipeline (URP). Please assign your chosen Render Pipeline Asset in the Graphics Settings to use it.");
```
jenniferd-unity, line 491: it should also mention the Quality Settings (A valid Render Pipeline asset needs to be assigned in the Graphics Settings or the Quality Settings.)

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
```diff
@@ -489,21 +482,66 @@ private static void LoadSRPBindersIfNeeded()
...
+                unsupportedSRPWarningIssued = true;
+            }
+        }
+
+        public static void LogUnsupportedSRP(bool forceLog = true)
+        {
+            bool logIssued = unsupportedSRPWarningIssued;
+            var binder = currentSRPBinder;
+
+            if (logIssued || !unsupportedSRPWarningIssued) // Don't reissue warning if inner currentSRPBinder call has already logged it
```
jenniferd-unity, line 501: that if is a bit weird - it can probably be removed 
logIssued == unsupportedSRPWarningIssued value according to line 498
and you are checking the value of unsupportedSRPWarningIssued in the other function

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
```diff
@@ -489,21 +482,66 @@ private static void LoadSRPBindersIfNeeded()
...
+
         public static VFXSRPBinder currentSRPBinder
         {
             get
             {
                 LoadSRPBindersIfNeeded();
+
                 VFXSRPBinder binder = null;
-                srpBinders.TryGetValue(GraphicsSettings.currentRenderPipeline == null ? "None" : GraphicsSettings.currentRenderPipeline.GetType().Name, out binder);
+                if (GraphicsSettings.currentRenderPipeline != null)
```
jenniferd-unity, line 512: you should also check against QualitySettings.currentPipelineAsset

if you only access this value at runtime, there are simpler ways to check this with RenderPipelineManager.currentPipeline

jenniferd-unity (changes requested):
Be careful with the fact an SRP can be active from the QualityLevel while GraphicsSettings are for a different SRP. (those workflow should not happen on future versions once the deprecation of GfxSettings' defaultRP lands)

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
PaulDemeulenaere: Actually, this code is purely editor, so, we should use this code :
```
var srpAsset = QualitySettings.renderPipeline ?? GraphicsSettings.renderPipelineAsset;
```
Is it correct @jenniferd-unity ?

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
julienf-unity: This is because we have 2 paths. LogUnsupportedSRP can be called from lazily initializing the currentSRPBinder (in that case we want to log only the first time) or explicitely upon some users action (opening the graph view, explicitely trying to compile a VFX...) and in this case we want to log always.

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
julienf-unity: I'm not sure I understand. How would we test the currentPipeline ? GraphicsSettings or QualitySettings ?

jenniferd-unity: what Paul suggested is the correct version. basically if we have an SRP in the QualitySettings it takes precedence over the one assigned in GraphicsSettings.

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
jenniferd-unity: let us imagine unsupportedSRPWarningIssued equals to true. then 
if(logIssue || !unsupportedSRPWarningIssued ) <=> if(true || false) > then TRUE
the opposite is:
if(logIssue || !unsupportedSRPWarningIssued ) <=> if (false|| true) > then TRUE

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
PaulDemeulenaere: Correct me if I'm wrong :
QualitySettings.renderPipeline returns the current active renderpipeline but it can be null.
If null, GraphicsSettings.renderPipelineAsset (default fallback) will be the active renderpipeline.

The problem of accessing `RenderPipelineManager.currentPipeline` during loading/importing/domain reload is it can returns null if it's done too early.

PaulDemeulenaere (approved):
Beside [the remark](https://github.com/Unity-Technologies/Graphics/pull/4379#discussion_r631706191) from @jenniferd-unity 
It looks good to me.
However, you should also double check your empty folder after master merge : the universal template folder has been moved to urp package, you could have an unexpected remaining metafile for the folder "RenderPipeline".

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
jenniferd-unity: actually no :) almost that

QualitySettings.renderPipelineAsset returns the serialized reference to an SRP Asset of the ACTIVE Quality Level
GraphicsSettings.currentRenderPipeline returns the serialized reference to an SRP Asset of the Graphics Settings
RenderPipelineManager.currentPipeline returns the instance of a Render Pipeline used by the Render Loop (this is the result of the combination of Quality + Graphics). this one is only initialized during the first rendered frame

jenniferd-unity: active == in use by the renderer

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
julienf-unity: logIssued is just a cache to check whether the inner call to currentSRPBinder has issued the log already (logIssued is false and unsupportedSRPWarningIssued is true after the call) and in that case we don't reissue (else it will be logged twice in a row). true to false is not possible.

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
julienf-unity: Ok I will check like Paul suggested then

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
jenniferd-unity: shouldnt it be `forceLog` in the if statement instead of `logIssue`?

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
jenniferd-unity: you also have `HDRenderPipeline.currentAsset` i forgot about it

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
julienf-unity: So what would be the text? "Please assign your chosen Render Pipeline Asset in the Graphics or Quality Settings to use it" ?

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
julienf-unity: If the test "if (logIssued || !unsupportedSRPWarningIssued)" is false, it means the message has already been log during the inner call to currentSRPBinder just above (unsupportedSRPWarningIssued was false before currentSRPBinder call, and true after), regardless of forceLog. This just prevents logging twice from the same explicit LogUnsupportedSRP call.

jenniferd-unity: okayyyyyy.... currentSRPBinder is not a getter but a function call! now all of it makes sense :)
I was going crazy
so now... how do we prevent others from going crazy? :)

jenniferd-unity (approved):

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
jenniferd-unity: This is what message the tech writing team gave to me for my feature "Project graphics settings do not refer to a URP Asset. Check the settings: Graphics > Scriptable Render Pipeline Settings, Quality > Render Pipeline Asset."
maybe it could be tweaked for your case to:
"The Visual Effect Graph is supported in the High Definition Render Pipeline (HDRP) and the Universal Render Pipeline (URP). Check the settings: Graphics > Scriptable Render Pipeline Settings, Quality > Render Pipeline Asset."

com.unity.visualeffectgraph/Editor/Core/VFXLibrary.cs
julienf-unity: It's a getter but lazily initialized. We could move the initialization directly in the callback probably. But it was safer and less intrusive to do it this way now.

julienf-unity:
We'll do a follow up PR to remove the reflection and polish things after C++ has landed

PR #4380, opened 2021-04-29
Fogbugz-1324451:

ramang-unity:
You can not depend on any particular contents of the RenderTexture you get from GetTemporary function. It might be garbage, or it might be cleared to some color, depending on the platform.
Clear the render texture because it was not black and bloom was breaking.

### Checklist for PR maker

---
### Purpose of this PR
Fix bug where the temp RT needs to be cleared because docs say it can contain garbage or set to a specific color depending on platform.

---
### Testing status
Manually tested with WMR device.

---
### Comments to reviewers


maloyer-unity (approved):

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -868,6 +868,10 @@ void CreateCameraRenderTarget(ScriptableRenderContext context, ref RenderTexture
                     colorDescriptor.autoGenerateMips = false;
                     colorDescriptor.depthBufferBits = (useDepthRenderBuffer) ? k_DepthStencilBufferBits : 0;
                     cmd.GetTemporaryRT(m_ActiveCameraColorAttachment.id, colorDescriptor, FilterMode.Bilinear);
+#if ENABLE_VR && ENABLE_XR_MODULE
+                    cmd.SetRenderTarget(m_ActiveCameraColorAttachment.id, 0, CubemapFace.Unknown, -1);
```
thomas-zeng, line 872: Is this logic XR specific?
VR/XR macro will be set to true even when XR is not enabled(they detect XR module presence, not runtime). 
cameraData.xr.enabled should be checked instead.

ramang-unity: It only breaks in XR (UWP + WinMR) because of the contents for that platform is not black. Technically it could be for any it just depends. It does not break standalone windows so thus the preproc.
I will add cameraData.xr.enabled

ramang-unity: You're right. It could be for any platform. Removed the preprocs

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -868,6 +868,13 @@ void CreateCameraRenderTarget(ScriptableRenderContext context, ref RenderTexture
                     colorDescriptor.autoGenerateMips = false;
                     colorDescriptor.depthBufferBits = (useDepthRenderBuffer) ? k_DepthStencilBufferBits : 0;
                     cmd.GetTemporaryRT(m_ActiveCameraColorAttachment.id, colorDescriptor, FilterMode.Bilinear);
+#if ENABLE_VR && ENABLE_XR_MODULE && UNITY_WSA_10_0 && UNITY_64
```
ramang-unity, line 871: Need these preprocs to make sure this does not happen on tiled renderers. It ruins performance on quest and hololens. Also this bug does not occur on those platforms.

fabien-unity: any idea why only this platform has the bug? could it be a deeper issue on the C++ side for this API ?

ramang-unity: In the PIX capture we see C++ calls DiscardResource. The call after is ClearRenderTargetView clears to colour yellow which C++/C# does not call. It seems the driver adds that call.

Also Unity docs state:
You can not depend on any particular contents of the RenderTexture you get from GetTemporary function. It might be garbage, or it might be cleared to some color, depending on the platform.

ryanhy-unity (approved):
Since Raman checked the fix on WMR and Hololens and I did a quick halo test on Oculus Rift and Quest and I didn't see any regressions from master so seems good to go to me.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
thomas-zeng: Is it possible to fit in this XR specific clear logic into the URP clear flag function here: https://github.com/Unity-Technologies/Graphics/blob/6bcd723e6833becd1d68457a5b7bc991f23dba7e/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L763 ?
I think it would be great we can put all clear related logic in one place for easier maintiance.

thomas-zeng (approved):
LGTM.

com.unity.render-pipelines.universal/Shaders/XR/XROcclusionMesh.shader
```diff
@@ -51,7 +51,6 @@ Shader "Hidden/Universal Render Pipeline/XR/XROcclusionMesh"
         Pass
         {
             ZWrite On ZTest Always Blend Off Cull Off
-            ColorMask 0
```
mikesfbay, line 54: Removing colorMask 0 from occlusion shade can be expensive since the shader used to be full-screen depth-only now writes color as well. Again, we should defer to the effect (bloom) to decide what color should be cleared (i.e. in XR mode, clear the first bloom buffer).

mikesfbay (changes requested):

maloyer-unity (approved):

thomas-zeng (approved):

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -797,6 +797,10 @@ protected static ClearFlag GetCameraClearFlag(ref CameraData cameraData)
             if (debugHandler != null && debugHandler.IsActiveForCamera(ref cameraData) && debugHandler.IsScreenClearNeeded)
                 return ClearFlag.All;
 
+            // XR: TODO, remove once we have visible area of occlusion mesh available
```
fabien-unity, line 800: One last minor change, can you use `// XRTODO` here ? we used that for most of SRP.

fabien-unity (approved):

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
ramang-unity: Sure

ryanhy-unity (approved):
re-checked this after the changes and didn't see any issues

erikabar (approved):
LGTM

phi-lira (approved):

PR #4389, opened 2021-04-30
Hd/split settings

RSlysz:
### Purpose of this PR
Fix issue where wizard changing is working booleans in a file that would likely be versioned. (https://fogbugz.unity3d.com/f/cases/1330640/)

Now HDProjectSettings (saved in ProjectSettings) will only contains the folder for hdrp resource inside project and material version.
A created HDUserProject (saved in UserSettings) is created and will handle all wizard parameters like "should I pop at start?" or "Have I already poped this editor run?".
While not mandatory, data have been migrated from former HDProjectSettings to new HDUserProject.

---
### Testing status
Manually tested that, when opening the editor: 
- HDProjectSettings  is saved again with a version equal to the migration.
- HDUserSettings is created with data that was contained into HDProjectSettings.
- former fields in HDProjectSettings are renamed Obsolete something.

---
### Comments to reviewers
Below commits include ones in the branch that should be merged before. actual change are all in comit d6c8ebc 
@peeweek for awarness


TomasKiniulis (approved):
Looks good.
- Tested upgrading new HDRP project with wizard open - no errors in console, wizard popup values are correctly migrated to UserSettings/HDRPUserSettings.asset
- Checked new project creation with PR already switched - HDRPProjectSettings.asset does not update popup values. Popup values are correctly updated to UserSettings/HDRPUserSettings.asset

JulienIgnace-Unity (approved):

PR #4395, opened 2021-04-30
Fixed copy/paste for stickynotes in ShaderGraph. Fixes fogbugz 1221042.

joshua-davis:
Also fixed how pasting works to clamp withing the graph view. This required creating a new interface to get/set a rect so nodes and stickynotes could be treated equally.

### Checklist for PR maker

---
### Purpose of this PR
Fixing https://fogbugz.unity3d.com/f/cases/1221042/
Sticky notes cannot be copied/cut/pasted right now.
Also fixed sticky notes not being able to be dragged into a group which would cause some odd behaviors since you could right-click to group a sticky note with other nodes, but you could never drag it in.

---
### Testing status

- Within a graph:
  - [x] Simply copy paste of a sticky note. Note's contents/title are the same and it's selected
  - [x] Copy two sticky notes. Both get pasted and selected
  - [x] Copy a sticky note and a node. Both are pasted and selected
  - [x] Cut/Paste works.
  - [x] Right click copy/paste/cut/duplicate works (menu option is in an odd location, see notes below)
  - [x] Copy of groups works (stickynote in a group, group with a sticky note, etc...)
  - [x] Copy a sticky note and a node and pan the graph view way to the side so the original location is very far away. Paste and verify that both nodes are within the current viewport (see comments for some details).
  - [x] Validate the size and theme is correct when copying sticky nodes (change to dark theme or change text size then copy/cut + paste)
- Do all of the above for:
  - [x] Between two different graphs
  - [x] From a graph to a sub-graph
  - [x] From a sub-graph to a graph
  - [x] Between two sub-graphs

- [x] Additionally verify you can drag sticky notes into a group.
 
---
### Comments to reviewers

There are a lot of cases with sticky notes, I hope I tried them all, but any additional testing is appreciated. 

Known issues:
- The copy/cut/duplicate commands on right-click are not in the same location as for normal nodes. This is controlled in GraphView so there's no way to fix this without changing the package.
- When notes/nodes are pasted they aren't always perfectly within the viewport, but they should be mostly inside (you can at least somewhat see them) as opposed to way off where they were copied. The core "framing" logic wasn't changed to be better, only to include sticky notes.


com.unity.shadergraph/Editor/Resources/StickyNote.uss
```diff
@@ -10,7 +10,6 @@
     position:absolute;
     flex-direction:column;
     align-items:stretch;
-    --layer:-100;
```
joshua-davis, line 13: This caused a really nasty selection bug. When pasting a sticky note it would always lose focus on the pasted node even when manually added to the selection. This was because the node would be newly created and then the layout would change the layer which would update it in the hierarchy and remove it from the selection (deep inside a layout callstack under Repaint). After talking with @Nightmask3 this seemed valid to move into the sticky note constructor which also fixed the problem.

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -1308,22 +1317,21 @@ internal static void InsertCopyPasteGraph(this MaterialGraphView graphView, Copy
             }
         }
 
-        private static void ClampNodesWithinView(MaterialGraphView graphView, IEnumerable<AbstractMaterialNode> nodeList)
+        private static void ClampNodesWithinView(MaterialGraphView graphView, IEnumerable<IRectInterface> rectList)
```
joshua-davis, line 1320: This change just makes the algorithm work on an interface to get/set the rect of an object so the same code can run on nodes and sticky notes.

jessebarker (approved):
LGTM

ghost (approved):
Dev manual testing looks sufficient

Nightmask3 (approved):
Looks good! Approving! 

The one thing I might change is the location of `IRectInterface` to be inside of `Editor/Drawing/Interfaces` instead, as it's an interface mostly concerned with UI drawing functionality, and that location seems more apt IMO

joshua-davis:
> Looks good! Approving!
> 
> The one thing I might change is the location of `IRectInterface` to be inside of `Editor/Drawing/Interfaces` instead, as it's an interface mostly concerned with UI drawing functionality, and that location seems more apt IMO

Sounds good, wasn't really sure where to put it

PR #4417, opened 2021-05-03
Added a fix for _TimeParameters not being defined in built-in

joshua-davis:
---
### Purpose of this PR
Using the time nodes in SG on built-in was not working because built-in doesn't define the used variable `_TimeParameters`. We recently added the variable but forgot this wouldn't do anything if built-in doesn't use it. Conveniently, this vector is just a sub-set of variables available from the other time vectors defined in built-in. 

---
### Testing status
Locally created a test using the time node and animated the uv used to sample a texture using the time, sintime, and costime values of the node and they all seemed reasonable (couldn't think of a good way to validate the sin and cos were distinct since they're so similar to each other).

---
### Comments to reviewers
To avoid changing more code, a define is used to map the `_TimeParameters` variable into a function call. This ideally would be changed, but maybe at a later point in time.


jessebarker (approved):
LGTM

PR #4418, opened 2021-05-03
Add Tessellation support to HD master node

sebastienlagarde:
### Purpose of this PR
This PR add support for shader graph tessellation on all HDRP master node.

Jira RM: https://unity3d.atlassian.net/browse/UR-1892
Jira team: https://unity3d.atlassian.net/browse/UR-1892

Initial tech document: https://docs.google.com/document/d/1y3fbGXEK1o-YXhxRAHAA1tzHAYIvesmxv2zYB4nZmlg/edit#heading=h.x2isyr8nsfkg

This PR offer similar tessellation present on LitTessellation master node but for all HDRP master node.

image

There is a new option Tessellation on all HDRP master node. Enabling this option allow to add two new block on the vertex block in shader graph:
- Tessellation factor: Allow to control the tessellation level, higher number mean more tessellated
- Tessellation displacement (world space): Allow to provite a displacement in world space which will apply on top of the tessellated world position.
Note: We differ from the pattern of Vertex modification (which provide position in object space) as we can cumulate both vertex animation and tessellation displacement. In this context
having to provide a world space tessellation position can bloat the graph, be more complex to the users and is more expensive in the generated code (as we need to re-evalaute the vertex animation 
but in tessellation stage).

Also enabling Tessellation option allow to access to all control from LitTessellation.shader and the Material can override everything.
A new property is introduce compare to LitTessellation.shader, it is the MaxDisplacement which represent the max size in meter of the intented displacement. 
This is used by the culling code to not cull triangle which shouldn't. This property in LitTessellation.shader was autoamtically calculated from Max value in displacement heightmap and scale/bias pair,
but it is not possible to apply this pattern with shadergraph automatically.

image

image

tessellation-sg-2

Note that custom interpolators are supported as well.

The PR also come with various cleanup and better code for Raytracing and VFX integration into SG. The detailed list of change:
- Rename VertexAnimation.template.hlsl to Vertex.template.hlsl
- Rename SharedCode.template.hlsl to Pixel.template.hlsl
- Factor the code which generate the fieldDependencies, the struct, the pragma in HDSubTarget so there is more code sharing and it can deal more easily with both tessellation
- Add 2 new properties to passDescription: useTessellation and useRaytracing to allow to identify what Collections should be added when generating the pass
- Move _TESSELLATION_PHONG keyword setup from BaseLitGUI to BaseUnlitGUI to apply to all shader type (incldue unlit)
- Ensure that TimeParameter can be use in any circonstance in both vertex and tessellation stage
- Refactor tessellation motion vector management for tessellation stage. New method is more accurate. However it will still be incorrect if tessellation factor is different for same vertices between two frames.
- Add all tessellation properties on a shader generate by shader graph if tessellation is enabled
- Cleanup required interpolators for the various pass as it was brute force including most of them. This result in less interpoaltor used in most passes
- Refactor a bit the code of custom interpolator  and add support in VertTessellation
- Generate the correct pragma for Tessellation shader. Note that Tessellation can't be control from Material due to those required pragma.
- Added a MakeMacroBlock method in the custom interpolator code.
- Switch shader_feature_local _TESSELLATION_PHONG to shader_feature_local_domain _TESSELLATION_PHONG

Documentation  have been added and highlight the following limit:
- Tessellation isn't supported by raytracing. With raytracing effect tessellation will appear as non tessellated and not displaced
- Tessellation isn't supported with Visual Effect Graph
- Tessellation isn't supported by decal (in this PR)
- Tessellation can't be control from Material, it need to be define from ShaderGraph  Settings
- When using World space Displacement, the object can expand out of its original bounds, users need to adjust it manually if they want correct rendering
- Motion vector will not be correct if the tessellation factor differ for a vertices between two frames

New inputs for Master Stack have been documented
---
### Testing status

What have been tested:
- Tested LitTessellation and LayeredLitTessellation shader to be sure they haven't change
- Tested shader graph tessellation with Lit and Unlit Master node with time animated displacement map and time animated tessellation factor. Checked all the properties as well (screen adapatation, distance, phong...)
- Checked the output code from many shadergraph, checked the generated interpolator and generated code to be sure only what is required is included and all the code cleanup happened correctly.
Note: Automation handle a large number of cases
- Tested custom interpolator with 2 fixed interpolator with SG tessellation and without it
- Tested motion vector mode, both opaque and transparent motion vector on Lit master node with tessellation
- Tested SG tessellation with deferred, forward, unlit and fabric
- Tested Tessellation with raytracing / path tracing (it should fallback to no tessellation automatically)
- Tested that it don't have impact when using the shader graph in VFX (i.e tessellation is disable for VFX case)
- Tested with Keyword - tested keyword for tess factor, for displacement and on pixel shader

Yamato test are green.

For QA: **what I haven't tested. (yet)**
- Test very complex shader graph (a material with multiple texture, vertex color with anisotropy, clear coat with multiple node usage (world position, world normal, world tangent) with custom interpolator that have dependency on uv and thing like that. i.e chaining dependency and having everything at the same time.
- Haven't tested with XR other than Yamayo auto test (Auto test pass I expect it is good)
- Baking of tessellated mesh (i.e for Meta pass chacking)
- Test that it works correctly with scene selection and outline. 
- Test SG tessellation with instancing

Add following automated test

Update of 1219_Lit_Light_On_Tessellation (add displace unlit, lit and fabric + check reflection and shadow casting)

image

Update 8201_VertexAnimation with SG tess on top on vertex deformation

image

Update 8207_CustomInterpolators with Unlit SG Tess and Lit SG Tess 

image

Update 4013 Transparent motion vector with Tess version (Test motoin vector on lit and unlit SG + opaque + opaque nomotion and opque camera motion)

image

Update 002-HDMaterials in runtime test with SGlit and SGUnlit (allow to check _domain qualifier as well)

image

Update 101_ReflectionsPerfFullRes with unlit and lit SG tess to check that there is a correct fallback (i.e disabled)

image

Update  8101_Opaque with Keyword Lit Tess SG  (Add test with Keyword)

image

---
### Comments to reviewers

know issue with this PR:
- Enabling Tessellation and adding tessellation factor and displacement block cause overlap of the vertex and pixel block.


com.unity.template-hd/Packages/manifest.json
```diff
@@ -5,7 +5,11 @@
     "com.unity.ide.visualstudio": "2.0.3",
     "com.unity.ide.vscode": "1.2.2",
     "com.unity.inputsystem": "1.0.2",
-    "com.unity.render-pipelines.high-definition": "12.0.0",
+    "com.unity.render-pipelines.core": "file:d:/newSRP/Graphics/com.unity.render-pipelines.core",
+    "com.unity.render-pipelines.high-definition": "file:d:/newSRP/Graphics/com.unity.render-pipelines.high-definition",
+    "com.unity.render-pipelines.high-definition-config": "file:d:/newSRP/Graphics/com.unity.render-pipelines.high-definition-config",
+    "com.unity.shadergraph": "file:d:/newSRP/Graphics/com.unity.shadergraph",
+    "com.unity.visualeffectgraph": "file:d:/newSRP/Graphics/com.unity.visualeffectgraph",
```
phi-lira, line 12: Pushed by mistake :)

com.unity.shadergraph/Editor/Generation/TargetResources/BlockFields.cs
```diff
@@ -15,6 +15,12 @@ public struct VertexDescription
                 new NormalControl(CoordinateSpace.Object), ShaderStage.Vertex);
             public static BlockFieldDescriptor Tangent       = new BlockFieldDescriptor(VertexDescription.name, "Tangent", "VERTEXDESCRIPTION_TANGENT",
                 new TangentControl(CoordinateSpace.Object), ShaderStage.Vertex);
+
+            // TODO: In the future we may have proper ShaderStage for tessellation, this will need to be revisit then
+            public static BlockFieldDescriptor TessellationFactor = new BlockFieldDescriptor(SurfaceDescription.name, "TessellationFactor", "Tessellation Factor", "VERTEXDESCRIPTION_TESSELLATIONFACTOR",
+                new FloatControl(1.0f), ShaderStage.Vertex);
+            public static BlockFieldDescriptor TessellationDisplacement = new BlockFieldDescriptor(SurfaceDescription.name, "TessellationDisplacement", "Tessellation Displacement", "VERTEXDESCRIPTION_TESSELLATIONDISPLACEMENT",
+                new PositionControl(CoordinateSpace.World), ShaderStage.Vertex);
```
esmelusina, line 23: Should probably be VertexDescription.name instead of SurfaceDescription.name here.

com.unity.shadergraph/Editor/Generation/Processors/CustomInterpolatorUtils.cs
```diff
@@ -50,22 +50,25 @@ internal static class Splice
...
+            internal bool isValid => isDefine || isBlock || isStruct || isFunc || isMacroBlock;
             internal bool hasPreprocessor => !String.IsNullOrEmpty(preprocessor);
 
             internal static Descriptor MakeFunc(string splice, string name, string dstType, string srcType, string define = "", string preprocessor = "") => new Descriptor { splice = splice, name = name, dst = dstType, src = srcType, define = define, preprocessor = preprocessor };
             internal static Descriptor MakeStruct(string splice, string name, string define = "", string preprocessor = "") => new Descriptor { splice = splice, name = name, define = define, preprocessor = preprocessor };
             internal static Descriptor MakeBlock(string splice, string dst, string src, string preprocessor = "") => new Descriptor { splice = splice, dst = dst, src = src, preprocessor = preprocessor };
+            internal static Descriptor MakeMacroBlock(string splice, string startMacro, string endMacro, string preprocessor = "") => new Descriptor { splice = splice, dst = endMacro, src = startMacro, preprocessor = preprocessor, hasMacro = true };
```
esmelusina, line 71: This doesn't seem to be used in the PR?

sebastienlagarde: It is used in HD Targets, this is use for the Macro to Interpolate in the domain shader:
CustomInterpSubGen.Descriptor.MakeMacroBlock("CustomInterpolatorInterpolateWithBaryCoordsMeshToDS", "TESSELLATION_INTERPOLATE_BARY(", ", baryCoords)"),

com.unity.shadergraph/Editor/Generation/Descriptors/PassDescriptor.cs
```diff
@@ -30,6 +30,10 @@ internal struct PassDescriptor
         public AdditionalCommandCollection additionalCommands;
         public CustomInterpSubGen.Collection customInterpolators;
 
+        // Properties
+        public bool useTessellation;
+        public bool useRaytracing;
```
esmelusina, line 35: Could we hack these in via the Targets interface instead? Ie. allowing passes to be composed with respect to the active context.

sebastienlagarde: Not sure to understand what you mean? I am open to suggestion.
If I get it you want it to be contain fully in RenderPipeline ? Note that those parameters really need to be per pass (i.e in a list of pass we can have a mix of tessellatoin, raytracing and regular)

esmelusina: Preferably-- perhaps some static branches in the targets could handle this instead via the fields system?

esmelusina: Oh-- I mean, it's possible for us to have an entry point where the target asks for the pass information providing active context information. So the branching on which pass information to include can be handled by the target instead of embedding it into the static representation of pass structs we have now.

com.unity.shadergraph/Editor/Generation/TargetResources/BlockFields.cs
sebastienlagarde: alright, will do the change.

com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
```diff
@@ -260,10 +260,12 @@ internal static void GenerateInterpolatorFunctions(StructDescriptor shaderStruct
             packBuilder.AppendLine("return output;");
             packBuilder.DecreaseIndent();
             packBuilder.AppendLine("}");
+            packBuilder.AppendNewLine();
```
sebastienlagarde, line 263: This one is just to have more readibility in generated code

com.unity.shadergraph/Editor/Generation/Enumerations/Platform.cs
```diff
@@ -63,5 +63,11 @@ internal static Platform[] GetHighEndPlatformArray()
         {
             return new Platform[] { Platform.D3D11, Platform.Playstation, Platform.XboxOne, Platform.GameCoreXboxSeries, Platform.Vulkan, Platform.Metal, Platform.Switch };
         }
+
+        // Return platform list not compatible with DXC (The list use by HDRP)
+        internal static Platform[] GetNeverUseDXCPlatformArray()
```
sebastienlagarde, line 68: This is required for now, but there is a PR coming which will prevent to require it

com.unity.shadergraph/Editor/Generation/Descriptors/PassDescriptor.cs
sebastienlagarde: ok figured out what I will do. Preparing change

sebastienlagarde: Change done and commited

esmelusina (commented):
Thank you for minimizing halo on SG, the remaining changes to SG are fine. There is some deliberation still going on within my team- I'm waiting to approve until we have some consensus.

sebastienlagarde:
Note: QA found that there is issue with keyword, so I will fix that, otherwise the PR should be good.

JulienIgnace-Unity (approved):

alelievr (approved):

JordanL8 (approved):

VladNeykov (approved):
[[Test Doc](https://docs.google.com/document/d/1EKJ9ynP9SDmtckQCLbw_oWNxsct_JohcC-7OpAnxwv4/edit#) ] Looking good, ran over a number of common scenarios, confidence-checked tessellation with VFX's new SG integration (not present yet, but no issues), and checked DX11,DX12, Vulkan, and Metal.

TomasKiniulis (approved):
Looks good. Tried out various tessellation options with various master stack targets and sg settings. Checked baking, raytracing fallback instancing and keywords. Findings are resolved.

Test doc: https://confluence.unity3d.com/display/HDRP/HD+Master+Node+Tessellation+Support

ghost (approved):
Did a bit of testing to get a feel for the feature, looks good! Worked with the features I tested it with (custom interpolators, subgraphs, keywords, sampler states, time)

sebastienlagarde:
Note: Yamato is green:

HDRP:  https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fsg-tessellation/.yamato%252Fall-hdrp.yml%2523Nightly_HDRP_trunk/6789705/job/pipeline (the DXR XR failure come from master after I merged master)

ShaderGraph: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fsg-tessellation/.yamato%252Fall-shadergraph.yml%2523Nightly_ShaderGraph_trunk/6786862/job/pipeline

PR #4422, opened 2021-05-03
Fixed deferred renderer issue where soft shadow keyword was always di…

kaychang-unity:
…sabled after the deferred pass, causing incorrect shadow setting in the forward-only and transparent passes.

### Checklist for PR maker
No, because this feature is not enabled in prior packages.
No, because the feature is not enabled in prior packages.

---
### Purpose of this PR
This PR fixes a bug where transparent or forward-only materials would always render without soft shadows if deferred rendering was enabled. The reason was that each deferred lights independently toggled the related shader keyword on/off and left the keyword in the "off" state. Deferred light soft shadows is now controlled by a new shader keyword, so that the original shader keyword is not overridden anymore.

ShaderPreprocessor was updated to handle the new _DEFERRED_SHADOW_SOFT keyword.

---
### Testing status
Ran gfx-tests.

---


simon-engelbrecht-soerensen (approved):

phi-lira (approved):
Approving to fix issue, but will follow up on discussion about these deferred keywords in channel.

PR #4423, opened 2021-05-04
Fix shaderGraph shaders to run with SSAO in deferred

kaychang-unity:

### Checklist for PR maker
This should not need backport because deferred rendering mode is a new feature from SRP 12.
Does this need a changelog since the feature is not officially released yet?

---
### Purpose of this PR
This fixes an issue where ShaderGraph unlit shader and lit shaders did not render into depthNormal prepass with deferred renderer.
Unlit shader was missing DepthNormalsOnly pass.
Lit shader declared a DepthNormals pass but was missing DepthNormalsOnly pass when ClearCoat option was selected (ComplexLit shader).

As the difference between DepthNormals and DepthNormalsOnly passes are very suble and confusing, I simplified their usage.
If DepthNormalsOnly pass is declared, DepthNormals pass does not need to be declared anymore (less redundancy, less shaders need to be built).

DepthNormals pass: this pass is selected when the forward renderer executes its DepthNormal prepass. It is ignored by the deferred renderer.
DepthNormalsOnly pass: this pass is selected when the forward renderer executes its DepthNormal prepass. It is also selected when the deferred renderer executes its DepthNormal prepass

Forward-only materials must declare DepthNormalsOnly pass, but must not declare DepthNormals pass. We use this mechanism to find forward-only materials that need to be rendered in deferred renderer prepass, but not in forward renderer prepass.

https://fogbugz.unity3d.com/f/cases/1335954/

---
### Testing status
Ran gfx-tests.

---


kaychang-unity:
This is causing error in 2D renderer gfx-tests.

phi-lira (approved):

hdb-unity (approved):
Tested for regressions by A/B test with boat attack project (forward rendering path) on MacOS.

2 variants reduction in built player after PR. This is verified using **_shader variant log level = "all shaders"_** on URPAsset.

No regressions found. 
Also, verified in one scenario that the bug is fixed here.

PR #4428, opened 2021-05-04
Handle asymmetry in projection matrix

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1330290/

**Note**: The PR has been reverted on master due to issues with the path tracer that would get garbage directions when tracing rays with the matrix (except for the sky which)

After reconsideration, it is possible to fix the case by computing the PixelCoordToWorldSpaceViewDirectionMatrix from the view projection matrix instead of building it manually from the parameters.

The computation now requires one more matrix multiplication.
The modifications in HDUtils.cs look big, but i just moved things in separate functions and simplify the final transpose by transposing before the matrix multiplication

Note: It looks like the asymmetry done in the fogbug case repro can currently be done using the lens shift parameter when ticking the pysical camera checkbox, but this solution is more flexible as it allows to arbitrarily modify the matrix form a script

---
### Testing status

Tested the repro project
Tested matrix is still working with lense shift and vertical fov
Relying on yamato for XR tests


FrancescoC-unity (approved):

remi-chapelain (approved):
As discussed with @adrien-de-tocqueville, not much left to test that would allow us to be more confident about that. 👍 
The rest will be taken care of by automation.

sebastienlagarde (approved):

PR #4429, opened 2021-05-04
Fix APV seams between cells.

JulienIgnace-Unity:
### Purpose of this PR
Currently, APV bake probes in one batch per cell. The baking backend being non deterministic, it means that shared probes at cells boundaries may have different values, causing seams in the lighting.
To avoid that, we now bake all cells within a "baking batch" together. The "baking batch" concept introduced in this PR is just an introduction to a more fleshed out concept that we'll implement later on.
This PR also fixes debug data not being cleared when clearing GI cache.


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -486,6 +500,10 @@ public static void RunPlacement()
                         cellPositions.Add(new Vector3Int(x, y, z));
 
             int index = 0;
+            // For now we just have one baking batch. Later we'll have more than one for a set of scenes.
+            // All probes need to be baked only once for the whole batch and not once per cell
+            // The reason is that the baker is not deterministic so the same probe position baked in two different cells may have different values causing seams artefacts.
+            m_BakingBatch = new BakingBatch(0);
```
FrancescoC-unity, line 506: Even tho for now we have a single batch, I'd avoid passing an hardcoded 0 here, maybe have a static counter to grab next bakingBatch ID? (so that when we split that up we don't need to track this 0 down)

JulienIgnace-Unity: Can do that yes

JulienIgnace-Unity: Done in latest commit.

FrancescoC-unity (approved):

alelievr (approved):

PR #4430, opened 2021-05-04
Hdrp/hd lens flare samples

SeanPuller:
# **Lens Flare Samples**
LensFlareSamples

This Sample provides:
- Lens Flare Assets
Unity_a04OmhLU6u

- Lens Flare Textures
Unity_5np1EAXHRz

- A scene to easily preview and author Lens Flare Assets
https://user-images.githubusercontent.com/43151199/117018555-e346d580-acf4-11eb-9c8c-fbbcb4ac74d5.mp4

---
### Comments to reviewers
I have supported both old input and new input for LensFlareSamplesInputAndControl.cs


com.unity.render-pipelines.high-definition/Documentation~/HDRP-Sample-Content.md
```diff
@@ -35,3 +35,12 @@ This Sample includes various examples of Materials. It includes Materials that u
...
 In the **Eye** Scene, the eye examples use a carefully designed mesh with a particular UV setup. If you want to produce eyes of similar quality, open the eye mesh in 3D modelling software to see how the mesh is contructed and the UVs are setup.
+
+## Lens Flare Samples
+
+![Lens Flare Samples](Images/LensFlareSamples.png)
+
+This Sample provides Lens Flare content that can be used in your project. It provides:
+- Lens Flare Assets
+- Lens Flare Textures
+- A scene to easily preview and author Lens Flare Assets
```
Vic-Cooper, line 46: A scene you can use to create and preview Lens Flare Assets

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Sample-Content.md
```diff
@@ -35,3 +35,12 @@ This Sample includes various examples of Materials. It includes Materials that u
 The Fabric, Hair and Eye Master Nodes usually require various work from artists inside the Shader Graph and the Samples are a good head start.
 
 In the **Eye** Scene, the eye examples use a carefully designed mesh with a particular UV setup. If you want to produce eyes of similar quality, open the eye mesh in 3D modelling software to see how the mesh is contructed and the UVs are setup.
+
+## Lens Flare Samples
+
+![Lens Flare Samples](Images/LensFlareSamples.png)
+
+This Sample provides Lens Flare content that can be used in your project. It provides:
```
Vic-Cooper, line 43: This Sample includes the following Lens Flare examples that you can use in your project:

Vic-Cooper (commented):
Made a couple of small suggestions to the Samples page.

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Sample-Content.md
SeanPuller: updated

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Sample-Content.md
SeanPuller: updated

iM0ve (commented):
I briefly checked the Sample project. The flares are good but we are investigating if we could present them in a different way, for example in a simple 3D scene with a lamp. Example:


https://user-images.githubusercontent.com/36502659/117288985-c6d1a700-ae74-11eb-9c6a-a129a8df71ec.mp4



If possible, the benefits of this approach would be:
* Simpler. No scripts. Whenever a script is involved, users have to look at the code and figure out what exactly it does for you. Figure out when you won't have that script in your project, how would you set it up?
* Would not include HDRI, save storage space.
* Would have 3D scene instead of 2D. This would users to check flares in more simple manner, arguably. But also it would be very easy to replace the lamp with your own asset and see how flares fit.
* Possibility to showcase occlusion with objects, if we want to.

sebastienlagarde (approved):

iM0ve (changes requested):
**Review status:** Done

**Whats being reviewed:**
The project contains 3 scenes, the goal is to show beginners where to find and how to use lens flares.

axDL0ik2D2
Unity_QZBgUmkFda
9sjjyq6QOW

**List of concerns:**
1. Complexity. 
  * 1st scene is fine. 
  * 2nd scene could be improved, out of the box, flares are nested(hidden) under CeilingLamp object. Could be optimized for first-time users so they don't have to go search how/where flares are implemented. Keep flat hierarchy
  * 3rd scene contains a script, with its help using A and D its possible to swipe through all flares. The scene setup is complex, not suited for beginners since you are required to read the script to understand what it does and doesn't set up for you. However, it can be still useful for those who want to see how to change/place.
2. Naming. These are the names of the scenes
  * ExampleDirectionalLight
  * ExampleInInteriorLight
  * LensFlareSamples
Which scene do we guide the user to open first? Naming is a bit misguiding, the 3rd sounds like it has the samples users will be looking for, but at the same time its the most complex. Shouldn't we guide users for the most primitive setup first? Maybe also make the naming more consistent?
3. Convenience of use.
To see all the flares you must load scene 3rd. The other two scenes only have 1 each. 
To see how to add flares into your scene in the simplest manner you must open scene 1 or scene 2. Due to this, it's not very convenient to use. Would it be a bad idea just to add ALL the flares in one simple scene so the user can quickly toggle them one by one without the need of switching or browsing local files or scripts.
4. Size.
The project is quite hefty, 31.45 mb. Out of those, 6 megabytes are used for an HDRI sky. I'm not sure if size is an issue, but also maybe the HDRI is not crucial to showcase flares. Can we still optimize here, worth asking because anything we add directly increases the size of the Unity editor.
5. Issue with the naming of the folders
image


**Suggestions**
6. To address issues 1st and 3rd, we could add all flares into the second scene as shown in the image below(see hierarchy). Nest them all under one game object that contains nothing else but flares.
Capture


**Cosmetic suggestions**
These Im less certain about compared to what's mentioned above but maybe will prove to be helpful.
6. Add a reflection probe for the lamps
7. Change the background sky color to not be pure black

Current
Unity_W8Wv5qIVQv


Suggestion1 If we want an Occluder instead of plane
oIR6gi0o45

Suggestion2 no occluder.
L5KvK5d41y

SeanPuller:
> **Review status:** Done
> 
> **Whats being reviewed:**
> The project contains 3 scenes, the goal is to show beginners where to find and how to use lens flares.

The goal is 
1. to provide users with some lens flare presets
2. provide examples on how to use lens flares in their own scenes
3. provide a fast preview of the flares included and can be utilized by the user to develop their own flares (i made this to help create the flares)

> 
> axDL0ik2D2
> Unity_QZBgUmkFda
> 9sjjyq6QOW
> 
> **List of concerns:**
> 
> 1. Complexity.
> 
> * 1st scene is fine.
> * 2nd scene could be improved, out of the box, flares are nested(hidden) under CeilingLamp object. Could be optimized for first-time users so they don't have to go search how/where flares are implemented. Keep flat hierarchy

Updated:
54_Unity

> * 3rd scene contains a script, with its help using A and D its possible to swipe through all flares. The scene setup is complex, not suited for beginners since you are required to read the script to understand what it does and doesn't set up for you. However, it can be still useful for those who want to see how to change/place.
> 

The scene is meant to help you cycle through the flares and move them using the mouse. Not too complex as the instructions for what you can do are always exposed. This scene adds a ton of value as it's the fastest way to cycle through the flares and I personally used this to aid in the development of the flares.
55_Unity


> 1. Naming. These are the names of the scenes
> 
> * ExampleDirectionalLight
> * ExampleInInteriorLight
> * LensFlareSamples
>   Which scene do we guide the user to open first? Naming is a bit misguiding, the 3rd sounds like it has the samples users will be looking for, but at the same time its the most complex. Shouldn't we guide users for the most primitive setup first? Maybe also make the naming more consistent?
> 

Do they have to open them in an order? The naming I think is pretty obvious as to what they contain

> 1. Convenience of use.
>    To see all the flares you must load scene 3rd. The other two scenes only have 1 each.
>    To see how to add flares into your scene in the simplest manner you must open scene 1 or scene 2. Due to this, it's not very convenient to use. Would it be a bad idea just to add ALL the flares in one simple scene so the user can quickly toggle them one by one without the need of switching or browsing local files or scripts.

I considered this when you mentioned it before but instead of combining the functionality of multiple scene I think it just does multiple things poorly, it doesn't showcase how to use the flares and is a much slower viewing the flares and the scene I provide adds more value for lens flare authoring

> 2. Size.
>    The project is quite hefty, 31.45 mb. Out of those, 6 megabytes are used for an HDRI sky. I'm not sure if size is an issue, but also maybe the HDRI is not crucial to showcase flares. Can we still optimize here, worth asking because anything we add directly increases the size of the Unity editor.

I didn't think about it adding to the size of Unity Editor, this must be because of some new changes. We might want to rethink about how we distribute samples in general.
We could exclude it, but I found it adds a great deal of value. To "recreate" something similar would be much larger as every object in the scene would have around 3 textures and even with 1 or 2 objects would dwarf the one HDRi.

> 3. Issue with the naming of the folders
>    image
> 

Good catch, pushed an update to the package.json that should fix this

> **Suggestions**
> 6. To address issues 1st and 3rd, we could add all flares into the second scene as shown in the image below(see hierarchy). Nest them all under one game object that contains nothing else but flares.

Addressed this above

> Capture
> 
> **Cosmetic suggestions**
> These Im less certain about compared to what's mentioned above but maybe will prove to be helpful.
> 6. Add a reflection probe for the lamps

Had a look at this, you can see the reflections of 2 lights on the lamp where there would only be one. if there was a more complete scene I would include something like that but I don't think it's necessary for this scene

> 7. Change the background sky color to not be pure black
> 

Just using the default template for indoor, I think it is fine

> Current
> Unity_W8Wv5qIVQv
> 
> Suggestion1 If we want an Occluder instead of plane
> oIR6gi0o45
> 
> Suggestion2 no occluder.
> L5KvK5d41y

PR #4432, opened 2021-05-04
[2021.2] SSAO: Fixing an issue with AfterOpaque option

ellioman:
# Purpose of this PR
Due to recent changes, the AfterOpaque option did not output SSAO to the final image.

The cause was that the `_SCREEN_SPACE_OCCLUSION` keyword doesn't get enabled when AfterOpaque is on which is required when calling the `GetScreenSpaceAmbientOcclusion()` function. 

The solution is to simply define the keyword in the AfterOpaque pass in the `ScreenSpaceAmbientOcclusion.shader` as we know that SSAO texture is available when that pass is executed.

# Testing status
Tested manually in the editor.


erikabar (approved):
lgtm

phi-lira (approved):

ellioman:
Yamato issues not related to this PR.

PR #4435, opened 2021-05-04
Fixed a namespace to be inside UnityEditor to avoid a test failure

joshua-davis:
---
### Purpose of this PR
After the material target PR a shader graph editor test was failing because a class was in the in a namespace beyond the expected ones. To fix this the one class was moved from `Unity.` to `UnityEditor.`.

---
### Testing status
Ran the SG tests locally to verify.


jessebarker (approved):
LGTM

PR #4445, opened 2021-05-05
[Universal] RenderPass Input Attachments for Deferred Renderer and exposing the RenderPass option

jonuuukas:
### Purpose of this PR
This PR adds few things:
1) Input Attachment API for the RenderPass implementation. This allows for subpasses to use previous subpass attachments via framebuffer fetch
2) Implements the beforementioned API for Deferred renderer and uses it in the Deferred pass. This utilises additional render target which writes depth into a color target (which is a limitation as some devices cannot do framebuffer fetch on depth buffer). Also adds _RENDER_PASS_ENABLED keyword to GBuffer related shaders as we do support opting out from RenderPass rendering path in the runtime (whether it's fallback as something is illegible to be done in a RenderPass or some other runtime changes). Though the variants should be stripped if RenderPass is not enabled. So lots of shaders touched due to this, but nothing drastic
3) Removes ENABLE_RENDER_PASS_UI define requirement to see the checkbox for enabling RenderPass API for the URP. Simple change

Initially this was supposed to be three separate PRs, but due to time constraints and the upcoming code freeze, this was merged into one bigger PR (also for QA to not do redundant testing passes and test the whole thing)

---
### Testing status
Full QA pass is scheduled. Currently manually testing on various devices
---
### Comments to reviewers


com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
```diff
@@ -121,12 +121,27 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
     TEXTURE2D_X_HALF(_GBuffer0);
     TEXTURE2D_X_HALF(_GBuffer1);
     TEXTURE2D_X_HALF(_GBuffer2);
+
+#if _RENDER_PASS_ENABLED
+    FRAMEBUFFER_INPUT_HALF(0);
+    FRAMEBUFFER_INPUT_HALF(1);
+    FRAMEBUFFER_INPUT_HALF(2);
+    FRAMEBUFFER_INPUT_FLOAT(3);
```
phi-lira, line 129: Consider adding defined for the GBuffer FB Attachment indices, this is more meaninful and easier to understand.

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -198,6 +200,10 @@ bool StripUnusedFeatures(ShaderFeatures features, Shader shader, ShaderSnippetDa
                 !IsFeatureEnabled(features, ShaderFeatures.LightLayers))
                 return true;
 
+            if (compilerData.shaderKeywordSet.IsEnabled(m_RenderPassEnabled) &&
+                !IsFeatureEnabled(features, ShaderFeatures.RenderPassEnabled))
+                return true;
```
phi-lira, line 205: Some things to consider. 
 1) Not all platforms support this keyword. For platforms that don't we shouldn't add more variants.
 2) Is there value for this keyword to be changed at runtime? We have plan to rediscuss shader variants stripping and this one seems a good candidate to always be stripped either the enabled or disabled variant. Essentially this will avoid adding more variants to build. cc @pbbastian 
 3) What happens when a shader doesn't define the multi_compile for this?

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -368,12 +381,15 @@ internal void ExecuteNativeRenderPass(ScriptableRenderContext context, Scriptabl
...
                     {
-                        attachmentIndices[i] = renderPass.m_InputAttachmentIndices[i];
+                        attachmentIndices[i] = renderPass.m_ColorAttachmentIndices[i];
                     }
                 }
 
                 if (validPassCount == 1 || currentMergeablePasses[0] == currentPassIndex) // Check if it's the first pass
                 {
+                    if (PassHasInputAttachments(renderPass)) //First ScriptableRenderPass cannot have input attachments - something is wrong
+                        return;
```
phi-lira, line 391: Should we have an assert instead of failing silently?

com.unity.render-pipelines.universal/Runtime/Passes/DeferredPass.cs
```diff
@@ -28,6 +28,15 @@ public override void Configure(CommandBuffer cmd, RenderTextureDescriptor camera
         {
             RenderTargetIdentifier lightingAttachmentId = m_DeferredLights.GbufferAttachmentIdentifiers[m_DeferredLights.GBufferLightingIndex];
             RenderTargetIdentifier depthAttachmentId = m_DeferredLights.DepthAttachmentIdentifier;
+            if (m_DeferredLights.UseRenderPass)
```
phi-lira, line 31: I still quite don't follow why we need this settting per-pass instead of checking the useRenderPass in global asset setting?

com.unity.render-pipelines.universal/Runtime/Passes/DeferredPass.cs
```diff
@@ -28,6 +28,15 @@ public override void Configure(CommandBuffer cmd, RenderTextureDescriptor camera
...
+            {
+                var inputAttachments = new RenderTargetIdentifier[]
```
phi-lira: This will cause allocation on heap and increase GC pressure.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -406,6 +406,7 @@ static void RenderSingleCamera(ScriptableRenderContext context, CameraData camer
                 if (renderer.useRenderPassEnabled && !context.SubmitForRenderPassValidation())
                 {
                     renderer.useRenderPassEnabled = false;
+                    CoreUtils.SetKeyword(cmd, ShaderKeywordStrings.RenderPassEnabled, false);
```
phi-lira, line 409: How do users use the feature?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -572,7 +574,13 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
 #endif
 
             if (this.actualRenderingMode == RenderingMode.Deferred)
+            {
+                var cmd = CommandBufferPool.Get();
+                CoreUtils.SetKeyword(cmd, ShaderKeywordStrings.RenderPassEnabled, useRenderPassEnabled && cameraData.cameraType == CameraType.Game);
+                context.ExecuteCommandBuffer(cmd);
+                cmd.Release();
```
phi-lira: Ah I see how it's used now, I'm not sure why the other one is necessary though, you could just enable once per-camera. Also there's a small overhead to execute command buffer, I'd recommend combining this execution of other command buffers we already have.

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
```diff
@@ -114,6 +140,9 @@ FragmentOutput SurfaceDataToGbuffer(SurfaceData surfaceData, InputData inputData
     output.GBuffer1 = half4(surfaceData.specular.rgb, surfaceData.occlusion);            // specular        specular        specular        occlusion
     output.GBuffer2 = half4(packedNormalWS, surfaceData.smoothness);                     // encoded-normal  encoded-normal  encoded-normal  smoothness
     output.GBuffer3 = half4(globalIllumination, 1);                                      // GI              GI              GI              [optional: see OutputAlpha()] (lighting buffer)
+    #if _RENDER_PASS_ENABLED
+    output.GBuffer4 = inputData.positionCS.z;
+    #endif
```
phi-lira, line 145: Seems too overboard to add an attachment for this and we can get that from depth texture?

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
jonuuukas: that definitely makes sense, will fix tomorrow

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
jonuuukas: Yeah, at least a warning should be thrown. Will fix tomorrow

com.unity.render-pipelines.universal/Runtime/Passes/DeferredPass.cs
jonuuukas: yes, already fixed this

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
jonuuukas: In this case this is a transient/memoryless attachment, so it's not very costly. To use depth texture we would require a depth prepass (as depth copy between GBuffer and Deferred passes would break renderpass making fb fetch impossible). Whereas depth attachment read is not supported on some platforms

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
jonuuukas: These two calls are required for the runtime fallback in case something changes with the settings, the previous call being - illegal command is found and RenderPass option is turned off in a hard manner (shouldn't be reached unless something goes drastically wrong, f.e. Blit in a RenderPass is found and etc.). I'll check whether it falls back properly when enabled in per-camera manner as it's possible this is a bit of an overkill. And I also agree about the cmd buffer, will try to find a better place to squeeze this in tomorrow

com.unity.render-pipelines.universal/Runtime/Passes/DeferredPass.cs
jonuuukas: By this setting per-pass you mean why DeferredLights has this variable? In this case deferred lights technically is not a Pass but a class that just does a lot of logic for GBuffer and Deferred passes. It actually copies the global asset setting and in this case it's safer here to check for the global setting (as the ScriptableRenderPass.useNativeRenderPass could be true even if falling back). I agree this is a bit confusing and messy and will try to make it more understandable

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
jonuuukas: > Not all platforms support this keyword
you mean platforms that do not support RenderPass API natively? As it will still use the fb fetch macros to fallback

Regarding 2. - if we are playing it safe and want to have a proper fallback mechanism - runtime change might happen and would fallback from fb fetch to the old LOAD_TEXTURE2D_X calls. This is definitely not necessary, but i did think we want to have both paths available in case something wrong happens and we want to fallback from RenderPass path. 
I'll admit I'm not fully familiar with the shader stripping part in URP, but my goal was to always strip if RenderPass API option is not enabled. Let me know if i made a mistake in doing that.

Regarding 3. - if shader doesn't multicompile it (let's it uses our UniversalGBuffer pass), the object won't be written into the GBuffer4 (fake depth) and DeferredFog will be rendered over it during DeferredPass (or artifacts, but so far saw only deferred fog over it, but it's definitely undefined behaviour)

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -387,7 +399,17 @@ internal void ExecuteNativeRenderPass(ScriptableRenderContext context, Scriptabl
                     if (!AreAttachmentIndicesCompatible(m_ActiveRenderPassQueue[m_LastBeginSubpassPassIndex], m_ActiveRenderPassQueue[currentPassIndex]))
```
manuele-bonanno, line 399: this check should consider input attachments too, that way we could merge more passes potentially, and the else branch won't be needed

manuele-bonanno: since we are not sure would be practical to merge multiple subpasses with compatible input attachments for now makes sense to leave a comment explaining why the else branch is happening, and potential room for future optimization if needed (will need investigqtion)

manuele-bonanno (approved):

phi-lira (approved):
We have some things to discuss but can be done after cutoff. Approving this PR so we can land.

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -564,6 +566,8 @@ internal void SetupLights(ScriptableRenderContext context, ref RenderingData ren
                     CoreUtils.SetKeyword(cmd, ShaderKeywordStrings.LightmapShadowMixing, isSubtractive || isShadowMaskAlways);
                     CoreUtils.SetKeyword(cmd, ShaderKeywordStrings.ShadowsShadowMask, isShadowMask);
                     CoreUtils.SetKeyword(cmd, ShaderKeywordStrings.MixedLightingSubtractive, isSubtractive); // Backward compatibility
+                    // This should be moved to a more global scope when framebuffer fetch is introduced to more passes
+                    CoreUtils.SetKeyword(cmd, ShaderKeywordStrings.RenderPassEnabled, this.UseRenderPass && renderingData.cameraData.cameraType == CameraType.Game);
```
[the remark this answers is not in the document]
jonuuukas: Enabling the API for XR devices is a little bit tricky and requires a lot of extra effort to verify whether the API itself is stable enough and etc. So currently we do not support VR cameras and etc XR team is aware of that). 

For the time being, we only enable this for Game camera with plans to expand the support in the future.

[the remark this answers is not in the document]
nigeljw-unity: I completely understand about disabling this for XR, but for reflection camera, the only real difference between a normal camera is the disabling of gizmo rendering. Disabling this for reflection camera rendering will kill this optimization for any view with visible probes, which should be a pretty common case.

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
```diff
@@ -294,7 +321,11 @@ Shader "Hidden/Universal Render Pipeline/StencilDeferred"
 
     half4 FragFog(Varyings input) : SV_Target
     {
-        float d = LOAD_TEXTURE2D_X(_CameraDepthTexture, input.positionCS.xy).x;
+        #if _RENDER_PASS_ENABLED
+            float d = LOAD_FRAMEBUFFER_INPUT(GBUFFER3, input.positionCS.xy);
```
nigeljw-unity, line 325: Should this be reference the channel like above:
float d = LOAD_FRAMEBUFFER_INPUT(GBUFFER3, input.positionCS.xy)**.x**;

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
```diff
@@ -16,6 +16,20 @@
 #else
 #endif
 
+#if _RENDER_PASS_ENABLED
+#if OUTPUT_SHADOWMASK && defined(_LIGHT_LAYERS)
+    #define GBUFFER_OPTIONAL_SLOT_1 GBuffer5
+    #define GBUFFER_OPTIONAL_SLOT_2 GBuffer6
+    #define FER_LIGHT_LAYERS GBuffer5
```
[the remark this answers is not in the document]
jonuuukas: fixed, thanks for the catch!

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
jonuuukas: works both ways, but definitely should be using the same way. fixing this!

com.unity.render-pipelines.universal/Shaders/Terrain/WavingGrassPasses.hlsl
```diff
@@ -43,6 +43,7 @@ void InitializeInputData(GrassVertexOutput input, out InputData inputData)
     inputData = (InputData)0;
 
     inputData.positionWS = input.posWSShininess.xyz;
+    inputData.positionCS = input.clipPos;
```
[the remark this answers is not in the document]
jonuuukas: seems like WavingGrass shader is rendered in Forward-only mode which is done after GBuffer+deferred, so fb fetch isn't utilised in this case. will remove as this is not-necessary in this case

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
kaychang-unity: Yes, it is mandatory to have depth stored in a color texture with RenderPass. Depth attachment cannot be read back on iOS and for Vulkan we were told it is "faster" to have depth as color.

kaychang-unity (approved):

nigeljw-unity (approved):

ernestasKupciunas (approved):
Testing Doc: https://docs.google.com/document/d/19m22fJKdxinaTAIrqkehOCq_As5uxpZsEJ7FSqNRbsU/edit

In general, I feel pretty comfortable with the quality. Tested iOS(Metal), Android (Gles2, Gles3, Vulkan Graphics), and AppleTV(Metal). 
I was able to get the rendering correctly and see the passes in the Frame Debugger as expected on all tested devices.
A couple of issues were found during the testing, but removing RP support for Gles2 solved both issues.
Good job!!

Projects used for testing:
[BoatAttack](https://github.com/Unity-Technologies/BoatAttack)
[Winter](https://github.com/ernestasKupciunas/Winter.git)
[URPRenderingExamples](https://github.com/Unity-Technologies/UniversalRenderingExamples)
[DeferredLaserCombat](https://github.cds.internal.unity3d.com/jonas-mortensen/deferred-laser-combat)

Devices under testing:
VLNQA00126, Samsung Galaxy S5 Mini, 6.0.1, Exynos 3 Quad 3470, Mali-400 MP
AppleTV 4k 64GB, tvOS: 14.5

Unity version used for testing:
Version: 2021.2.0a18.2443
Revision: trunk 65cb2e8a5814
Built: Mon, 10 May 2021 13:28:10 GMT

phi-lira:
I had to merge master to resolve conflicts. Conflicts were trivial to solve - diff of conflicted files https://github.com/Unity-Technologies/Graphics/pull/4445/commits/3b6edcdabc7b8f8099f0109e25fcb49acb897363

PR #4446, opened 2021-05-05
Display on Probe Volume Inspector if lighting data needs to be rebaked. 

FrancescoC-unity:
Following yesterday meeting, this is doing the change of informing the user that a baking is needed to get updated data due to changes in the probe volume characteristics or because the lighting data has been cleared. 

image

This will trigger when: 

 - Probe volume transform changes.
 - Probe volume parameters (size or subdiv levels) change
 - Data is cleared.

If the probe volume is reset to the properties for which we have valid data, it will not display the error anymore.


JulienIgnace-Unity (approved):

alelievr (approved):

TheoWong-pixel:
Is the sub division controller missing Min and Max input values inputs?

FrancescoC-unity:
> Is the sub division controller missing Min and Max input values inputs?

It is dynamic, we cannot statically display in UI (hence why it is displayed in the info box). 

It is however not part of this PR anyway :-)

TheoWong-pixel (approved):

PR #4453, opened 2021-05-05
[10.x] Forward port hotfix for Hololens Remote Rendering

thomas-zeng:
---
### Purpose of this PR
- This PR forward ports a temporary fix for hololens2 extra blit issue from 19.4 branch.
- Original commit: https://github.com/Unity-Technologies/Graphics/commit/fab1ca7622cf065a70c309a3f2808cec63b9d9da

---
### Testing status
- ABV
- QA

---
### Comments to reviewers
To QA:
- Please test this PR on hololens with unity 20.3 and ensure there is no regressions.
- Please check ARMv7 platform as well as emulator and ensure no extra blit is introduced when scriptable render feature is in use.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
@@ -420,7 +420,7 @@ namespace UnityEngine.Rendering.Universal
 
 #if ENABLE_VR && ENABLE_VR_MODULE
         static List<XR.XRDisplaySubsystem> displaySubsystemList = new List<XR.XRDisplaySubsystem>();
-        static XR.XRDisplaySubsystem GetXRDisplaySubsystem()
+        static XR.XRDisplaySubsystem GetFirstXRDisplaySubsystem()
         {
             XR.XRDisplaySubsystem display = null;
             SubsystemManager.GetInstances(displaySubsystemList);
@@ -443,9 +443,9 @@ namespace UnityEngine.Rendering.Universal
                 var platform = Application.platform;
                 if (platform == RuntimePlatform.WSAPlayerX86 || platform == RuntimePlatform.WSAPlayerARM || platform == RuntimePlatform.WSAPlayerX64)
                 {
-                    var displaySubsystem = GetXRDisplaySubsystem();
+                    var displaySubsystem = GetFirstXRDisplaySubsystem();
                     
-                    if (!displaySubsystem.displayOpaque)
+                    if (displaySubsystem != null && !displaySubsystem.displayOpaque)
                         return true;
                 }
             }


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -418,7 +418,48 @@ static bool IsMultiPassStereoEnabled(Camera camera)
...
+                if (platform == RuntimePlatform.WSAPlayerX86 || platform == RuntimePlatform.WSAPlayerARM)
+                {
+                    var displaySubsystem = GetXRDisplaySubsystem();
+                    var subsystemDescriptor = displaySubsystem?.SubsystemDescriptor ?? null;
+                    string id = subsystemDescriptor?.id ?? "";
+
+                    if (id.Contains("Windows Mixed Reality Display"))
+                        return true;
+
+                    if (!XR.WSA.HolographicSettings.IsDisplayOpaque)
```
ramang-unity, line 453: Will this work with OpenXR + HL2? We should discuss this.

yl-msft: This "WSA" API won't work for OpenXR, neither will the "string" test above.  However, this [`XRDisplaySubsystem.displayOpaque`](https://docs.unity3d.com/ScriptReference/XR.XRDisplaySubsystem-displayOpaque.html) should work across all XR SDK (both windowsmr, and openxr plugin).

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -418,7 +418,48 @@ static bool IsMultiPassStereoEnabled(Camera camera)
...
+        // with custom render pass.
+        // TODO: Remove this method and usages when we have proper dependency tracking in the pipeline to know
+        // when a render pass requires camera color as input.
+        internal static bool IsRunningHololens(CameraData cameraData)
+        {
+#if PLATFORM_WINRT
+            if (cameraData.xr.enabled)
+            {
+                var platform = Application.platform;
+                if (platform == RuntimePlatform.WSAPlayerX86 || platform == RuntimePlatform.WSAPlayerARM)
```
yl-msft, line 444: HoloLens 2 also support WSAPlayerARMv7 platform.

phi-lira (approved):

phi-lira:
Will this be needed for 11.x as well? 
 From 12.x we can provide option.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
thomas-zeng: Addressed in 
cd10234

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
thomas-zeng: ARMv7 is not part of the RuntimePlatform enum: https://docs.unity3d.com/ScriptReference/RuntimePlatform.html
I think it is likely covered by ARM platform enum. I mentioned ARMv7 for QA so that we could cover that.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -418,7 +418,43 @@ static bool IsMultiPassStereoEnabled(Camera camera)
...
+        // when a render pass requires camera color as input.
+        internal static bool IsRunningHololens(CameraData cameraData)
+        {
+#if PLATFORM_WINRT
+            if (cameraData.xr.enabled)
+            {
+                var platform = Application.platform;
+                if (platform == RuntimePlatform.WSAPlayerX86 || platform == RuntimePlatform.WSAPlayerARM || platform == RuntimePlatform.WSAPlayerX64)
+                {
+                    var displaySubsystem = GetXRDisplaySubsystem();
```
yl-msft, line 446: Noticed a error/crash condition.  Not sure how to trigger, but better to be defensive since this is a code path rarely get tested.

1. There's a change `GetXRDisplaySubsystem` return null, so the if condition below should be `if (displaySubsystem != null && !displaySubsystem.displayOpaque) return true;
2. Better rename the helper function as "GetFirstXRDisplaySubsystem" since it's hardcoded to index 0.

yl-msft (approved):

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
thomas-zeng: Thanks and and agree on the suggested changes. 
Addressed in b17c607.

phi-lira:
Let me know when this is ready and I'll merge it and discuss package release.

JasonCostanza (approved):
Version 2020.3.8f1 (507919d4fff5)
Branch: 2020.3/staging

SRP:
commit cd102344ce82803a565ee87f613feeeb6c82f289 (HEAD -> 10.x.x/xr/hl2-remove-blit, origin/10.x.x/xr/hl2-remove-blit)

I ran a blank scene on both HMD's and also ran the WMR HMD through a local run of Gfx automation which checked out as well. I think this looks good to go with no major egregious regressions that I could find in my testing.

phi-lira (approved):

phi-lira:
Merging as requested by @thomas-zeng

PR #4459, opened 2021-05-05
[2021.2][ShaderGraph] Use 3-component transform  for view to object direction

jessebarker:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1333781/

When transforming direction from view space to object space, we were adding a w==1.0 component and multiplying by the full inverse view matrix, which yields incorrect results.  We want to transform the raw direction by the 3x3 inverse view matrix.

---
### Testing status
Repro steps from the bug
Yamato [ShaderGraph PR job](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg-fix-1333781/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/6683916/job)

---
### Comments to reviewers
If the checks below don't show the above job, it is because I had to force-push an updated commit message on the change (the original was inaccurate).  However, the actual code did not change so the results of the job are still valid.


joshua-davis (approved):

jessebarker:
I'm seeing the same URP_Foundation failure as appears on the master commit this branch is based off of, so I'm merging this.  The only tests that actually use a transform node are passing.

PR #4463, opened 2021-05-06
Surface gradient decal blending

adrien-de-tocqueville:
### Purpose of this PR

This PR adds a new optional way for HDRP to blend normal maps for decals: it uses surface gradient to apply decal normal while preserving the underlying object normal.

Before:
surf-master

After:
surf-branch

This is enabled via an option in the HDRP Asset.
It has an other option to enable high precision buffer: If disabled, the normal will have max angle of 45°:

image

---
### Testing status

Tested with HDRP/Decal and ShaderGraph Decal on all shader types: Lit (layered/tessellated), AxF, Terrain, ShaderGraph
Tested that disabling the option gives the same result as before

Added graphic test
1714_DecalSurfaceGradient

---
### Comments to reviewers

This introduce a change in the way decal data is stored in the DBuffer, which will break backward compatibility with existing custom decal shaders. We should mention it the the upgrade guide and try to output a compilation error for shaders using the old technique

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -866,7 +866,7 @@ namespace UnityEngine.Rendering.HighDefinition
                 // Check if recursive rendering is enabled or not. This will control the cull of primitive
                 // during the gbuffer and forward pass
                 ScreenSpaceReflection settings = hdCamera.volumeStack.GetComponent<ScreenSpaceReflection>();
-                bool enableRaytracedReflections = hdCamera.frameSettings.IsEnabled(FrameSettingsField.RayTracing) && settings.rayTracing.value;
+                bool enableRaytracedReflections = hdCamera.frameSettings.IsEnabled(FrameSettingsField.RayTracing) && ScreenSpaceReflection.RayTracingActive(settings);
                 m_ShaderVariablesGlobalCB._EnableRayTracedReflections = enableRaytracedReflections ? 1 : 0;
                 RecursiveRendering recursiveSettings = hdCamera.volumeStack.GetComponent<RecursiveRendering>();
                 m_ShaderVariablesGlobalCB._EnableRecursiveRayTracing = recursiveSettings.enable.value ? 1u : 0u;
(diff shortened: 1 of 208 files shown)


JMargevics (approved):
enabled
switch

Feature looks good to me. Things I've tried: 

- Upgrading decals from current master
- Enabling HDRP asset option
- With or without high precision buffer
- Checked tooltips
- Inside LookDev
- Undo/Redo

@adrien-de-tocqueville said he works on a test to cover the automation. 

All good ✅

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1088,7 +1088,7 @@ protected override void Render(ScriptableRenderContext renderContext, Camera[] c
 #if ENABLE_NVIDIA && ENABLE_NVIDIA_MODULE
             m_DebugDisplaySettings.nvidiaDebugView.Update();
 #endif
-            Terrain.GetActiveTerrains(m_ActiveTerrains);
+            //Terrain.GetActiveTerrains(m_ActiveTerrains);
```
sebastienlagarde, line 1091: is it expected? I dot' think so

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1463,8 +1463,8 @@ ref renderDatas
                         settingsCopy.forceResolution = true;
                         DynamicResolutionHandler.UpdateAndUseCamera(camera, settingsCopy);
 
-                        foreach (var terrain in m_ActiveTerrains)
-                            terrain.SetKeepUnusedCameraRenderingResources(camera.GetInstanceID(), true);
+                        //foreach (var terrain in m_ActiveTerrains)
```
sebastienlagarde, line 1466: is it expected? I dot' think so

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
```diff
@@ -94,6 +94,8 @@ internal static RenderPipelineSettings NewDefault()
                 supportedLitShaderMode = SupportedLitShaderMode.DeferredOnly,
                 supportDecals = true,
                 supportDecalLayers = false,
+                supportSurfaceGradient = false,
```
sebastienlagarde, line 97: should be true by default. We want it for new project but keep it false for old project (can be done wit the migration step)

sebastienlagarde (commented):
You forget to strip the extra variant in the stripper

sebastienlagarde (approved):

PR #4464, opened 2021-05-06
Fix issue with index calculations across scenes after master merge

FrancescoC-unity:
This fixes some issues that I apparently mis-merged when getting up to speed with master.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
+++ b/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
@@ -195,6 +195,9 @@ namespace UnityEngine.Experimental.Rendering
             if (refVolAuthList.Length == 0)
                 return;
 
+            FindWorldBounds();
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -195,6 +195,9 @@ private static void OnBakeStarted()
             if (refVolAuthList.Length == 0)
                 return;
 
+            FindWorldBounds();
+            refVolAuthList = GameObject.FindObjectsOfType<ProbeReferenceVolumeAuthoring>();
```
JulienIgnace-Unity, line 199: Why do we need to get the list twice? It can change between calls?

FrancescoC-unity: Yeah it can get invalidated when loading other scenes

JulienIgnace-Unity (approved):

PR #4465, opened 2021-05-06
Made the ui bind to properties earlier in the setup

martint-unity:
### Purpose of this PR
We need to bind these properties earlier during CreateGUI call. Otherwise we had an issue in where we are using the Search API and the UI didn't update properly when the Search was done.
So added an Action to be called after initialization is done for the converters.

image

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/Converter/converter_editor.uxml
+++ b/com.unity.render-pipelines.universal/Editor/Converter/converter_editor.uxml
@@ -9,7 +9,7 @@
             <ui:DropdownField index="-1" name="conversionsDropDown" style="height: 27px; width: 208px; display: none;" />
             <ui:VisualElement style="flex-grow: 1;" />
         </ui:VisualElement>
-        <ui:ScrollView scroll-deceleration-rate="0,135" elasticity="0,1" name="convertersScrollView" style="flex-grow: 1; flex-shrink: 1; max-width: none; width: 627px; padding-right: 0;" />
+        <ui:ScrollView scroll-deceleration-rate="0,135" elasticity="0,1" name="convertersScrollView" style="flex-grow: 1; flex-shrink: 1; max-width: none; width: 628px; padding-right: 0;" />
     </ui:VisualElement>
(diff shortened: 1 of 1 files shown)


erikabar (approved):
LGTM

com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
```diff
@@ -210,144 +283,67 @@ void GetAndSetData(int i)
...
-                Status status;
-                bool active = true;
-                // If this data hasn't been filled in from the init phase then we can assume that there are no issues / warnings
-                if (string.IsNullOrEmpty(converterItemInfos[j].warningMessage))
-                {
-                    status = Status.Pending;
-                }
-                else
+            conv.OnInitialize(initCtx,
+                () =>
```
svens-unity, line 291: pref we would move the code out to function.

com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
```diff
@@ -210,144 +283,67 @@ void GetAndSetData(int i)
...
-                });
-            }
+                    // Default all the entries to true
+                    for (var j = 0; j < converterItemInfos.Count; j++)
+                    {
+                        string message = string.Empty;
+                        Status status;
+                        bool active = true;
+                        // If this data hasn't been filled in from the init phase then we can assume that there are no issues / warnings
+                        if (string.IsNullOrEmpty(converterItemInfos[j].warningMessage))
```
svens-unity, line 304: I dont fully understand the logic here. Having no warning message means the item is still pending? 
We can look into this for a future PR.

svens-unity (approved):
It would be great to clean up the callbacks and lambdas before expanding usage to other teams or promoting this API to public.

TheoWong-pixel (approved):

phi-lira (approved):

PR #4466, opened 2021-05-06
Fix reflections intensity multiplier being ignored with URP and HRV2 (Graphics part)

vincent-breysse:
### Purpose of this PR
Part of the implementation on the Hybrid renderer side was missing to handle `unity_SpecCube0`. So the shader code fetching the cubemap was hidden under a define when using DOTS packages. 

Now that the hybrid code is done, this ifdef should be modified. See https://github.com/Unity-Technologies/dots/pull/7364

---
### Testing status (with 21.2.0a16 + hybrid 0.13.0-preview.16)
* Manual testing : Validated TestURPLitProperties scene (HybridURPSamples) with this Graphics PR and with/without the DOTS PR


svens-unity (approved):
Thanks for implementing this 👍 It'd be good to support spec1 and fallback skybox too in the future :)

sebbbi (approved):

erikabar (approved):
LGTM

PR #4470, opened 2021-05-06
[APV] Distance field based subdivision & placement

alelievr:
#### GPU probe placement

Replace the current bounding box subdivision system with a GPU-based approach using a signed distance field of the scene.

Rough descriptions of the algo:
- Split the world into cells
- For each cell, voxelize the mesh renderers into a 3D texture
- Compute the distance field (unsigned) of the 3D texture
- Voxelize the probe volumes (store the min and max subdivision inside a 3D texture)
- Generate a list of brick position based on the probe volume min/max data and the distance field
- Readback all the positions and convert them to bricks for the probe volume  system

Nested Probe Volumes example: 

https://user-images.githubusercontent.com/6877923/117336478-86861f00-ae9c-11eb-80b8-53d5b5129061.mp4

#### Realtime subdivision visualization:

Added a bool in the debug window to view the subdivision in real-time, each cell takes between ~0.5 and ~4 ms to compute, so it's possible to work with this, but of course, it doesn't scale well with the number of cells (number captured with a simplification level of 3 and min probe distance of 1).

Note that the culling settings in the debug menu work for the real-time subdivision, so if you're having performance issues, it can help to reduce the culling distance.

image

#### Simplification level in probe volume profile

image

#### Probe Volume new fields

Added two new fields in the Probe Volume component:
- An object layer mask that filters the game objects used for the subdivision 
- Added an offset on the distance for geometry subdivision (see video)

https://user-images.githubusercontent.com/6877923/117459769-0e773200-af4c-11eb-8f0a-8c0c1075513c.mp4

---
### Testing status

~I know there is a bug with the real-time subdivision not working when we change something in the profile.~
~There is a scaling issue when baking with simplification levels 1 and 3.~

Tested with template scene and all the other simple APV test scene we have.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffeature%252Fapv-gpu-subdivision/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/6864755/job


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -396,6 +399,10 @@ static float CalculateSurfaceArea(Matrix4x4 transform, Mesh mesh)
         static void DilateInvalidProbes(Vector3[] probePositions,
             List<Brick> bricks, SphericalHarmonicsL2[] sh, float[] validity, ProbeDilationSettings dilationSettings)
         {
+            if (!dilationSettings.dilate)
```
FrancescoC-unity, line 402: While this would be the right thing to do I observed that _without_ dilation at all the result are hideous with current subdiv to the point that we shouldn't allow that. Did you try with the new subdiv? (the template_1 from the APV_Test repo is a good one)

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -215,6 +215,7 @@ static void OnBakeStarted()
         static void OnAdditionalProbesBakeCompleted()
         {
             UnityEditor.Experimental.Lightmapping.additionalBakedProbesCompleted -= OnAdditionalProbesBakeCompleted;
+            UnityEngine.Profiling.Profiler.BeginSample("OnAdditionalProbesBakeCompleted");
```
JulienIgnace-Unity, line 218: Do we want to keep those or was it just for testing purpose during developpement?

FrancescoC-unity: I think it wouldn't hurt keeping this here.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeSubdivisionContext.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                var probeVolume = pv.component;
+                var halfSize = probeVolume.size / 2.0f;
+                var minCellPosition = (probeVolume.transform.position - halfSize) / cellSize;
+                var maxCellPosition = (probeVolume.transform.position + halfSize) / cellSize;
+
+                Vector3Int min = new Vector3Int(Mathf.FloorToInt(minCellPosition.x), Mathf.FloorToInt(minCellPosition.y), Mathf.FloorToInt(minCellPosition.z));
+                Vector3Int max = new Vector3Int(Mathf.CeilToInt(maxCellPosition.x), Mathf.CeilToInt(maxCellPosition.y), Mathf.CeilToInt(maxCellPosition.z));
+
+                for (int x = min.x; x < max.x; x++)
```
FrancescoC-unity, line 63: I might very well be missing something, but given we do this cellPositions additions per probes, if we have two probe volumes touching the same cell, won't we be adding two cell positions for the same cell here?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs.hlsl
```diff
@@ -0,0 +1,52 @@
...
+    float3 Z;
+    int minSubdivisionLevel;
+    int maxSubdivisionLevel;
+    float geometryDistanceOffset;
+};
+
+//
+// Accessors for UnityEngine.Experimental.Rendering.ProbePlacement+GPUProbeVolumeOBB
+//
+float3 GetCorner(GPUProbeVolumeOBB value)
```
FrancescoC-unity, line 23: Do we need these accessors? I don't see them being used?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
```diff
@@ -3,236 +3,443 @@
...
+                    return Matrix4x4.Rotate(Quaternion.Euler((Time.realtimeSinceStartup * 10f) % 360, 0, 0));
                 }
+
+                // We need to bind at least something for rendering
+                cmd.SetRenderTarget(dummyRenderTarget);
+                cmd.SetViewport(new Rect(0, 0, dummyRenderTarget.width, dummyRenderTarget.height));
+                var props = new MaterialPropertyBlock();
+                foreach (var kp in renderers)
+                {
+                    // Only mesh renderers are supported for the voxelization.
```
FrancescoC-unity, line 298: This might be a problem for terrain.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeSubdivide.compute
```diff
@@ -0,0 +1,261 @@
...
+Texture3D _ProbeVolumeData;
+RWTexture3D<float4> _Output;
+RWTexture3D<float4> _FinalOutput;
+
+AppendStructuredBuffer<float3> _Bricks;
+RWStructuredBuffer<float3> _BricksToClear;
+
+StructuredBuffer<GPUProbeVolumeOBB> _ProbeVolumes;
+
+float3 _Size;
```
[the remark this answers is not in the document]
FrancescoC-unity: Generally I see a bunch of those are float3 and can be float (float3 * float is well defined in HLSL)

Also prefer to pass float4 since these are going to go into a cbuffer and padded singularly, what I usually do is send a bunch of 

float4 SDFParams1...N

and then do defines with proper naming in shader of the granularity you need so for example 

#define  _Offset SDFParams1.w

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeSubdivide.compute
```diff
@@ -0,0 +1,261 @@
...
+
+    if (nearest.w < 0.5)
+        nearest = float4(-10, -10, -10, 0);
+
+    int j = 0;
+    for (int i = 0; i < 27; i++)
+    {
+        float3 o = id + offset3D[i] * _Offset;
+
+        o %= _Size.x;
```
FrancescoC-unity, line 115: If this is unsigned, casting with (uint) the size before the modulo will help perf.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeSubdivide.compute
```diff
@@ -0,0 +1,261 @@
...
+#pragma kernel Subdivide
+
+#pragma enable_d3d11_debug_symbols
+
+#include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl"
+#include "Packages/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs.hlsl"
+
+Texture3D _Input;
+Texture3D _ProbeVolumeData;
+RWTexture3D<float4> _Output;
```
FrancescoC-unity, line 17: I see this is always written in its first 3 channels, can it be float3?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeSubdivide.compute
```diff
@@ -0,0 +1,261 @@
...
+            }
+        }
+    }
+
+    return float2(pMin, pMax);
+}
+
+bool PointInsideOBB(GPUProbeVolumeOBB obbA, GPUProbeVolumeOBB obbB)
+{
+    float3 axes[] =
```
FrancescoC-unity, line 172: It is fine because this is a bake time shader, but worth pointing that generally this pattern (declaring local arrays in hlsl) leads to awful code generation

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/VoxelizeScene.shader
```diff
@@ -0,0 +1,85 @@
+Shader "Hidden/ProbeVolume/VoxelizeScene"
+{
+    Properties
```
FrancescoC-unity, line 3: Do we need this even if empty?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
alelievr: I think it got removed at some point and a merge conflict included it back, I can keep it :)

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeSubdivide.compute
```diff
@@ -0,0 +1,261 @@
+// Each #kernel tells which function to compile; you can have many kernels
+#pragma kernel Clear
+#pragma kernel ClearBuffer
+#pragma kernel FillUVMap
+#pragma kernel JumpFlooding
+#pragma kernel FinalPass
+#pragma kernel VoxelizeProbeVolumeData 
+#pragma kernel Subdivide
+
+#pragma enable_d3d11_debug_symbols
```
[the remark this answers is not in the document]
alelievr: done

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeSubdivisionContext.cs
alelievr: It's fine because the add is protected by the HashSet `cellPositions`, so it's not possible to have duplication in the final list

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
```diff
@@ -3,236 +3,443 @@
...
-                foreach (ProbeReferenceVolume.Volume c in cullees)
+                using (new ProfilingScope(cmd, new ProfilingSampler("Clear")))
                 {
-                    if (result.Contains(c))
-                        continue;
+                    cmd.SetComputeTextureParam(subdivideSceneCS, s_ClearKernel, _Output, sceneSDF);
+                    DispatchCompute(cmd, s_ClearKernel, sceneSDF.width, sceneSDF.height, sceneSDF.volumeDepth);
+                }
+
+                cmd.SetRandomWriteTarget(4, sceneSDF);
```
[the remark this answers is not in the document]
alelievr: 0 doesn't work, 1 and 2 crash the editor (Dx11), and 4 isn't used by HDRP or URP IIRC.
This binding thing is really weird, I don't think I can do something better than that, I'll put a name instead of this magic number though

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
```diff
@@ -3,236 +3,443 @@
...
-            LightsToVolumes(ref cellVolume, ref influenceVolumes, ref sceneRefs);
-
-            // Extract all ProbeVolumes inside the cell
-            List<ProbeReferenceVolume.Volume> indicatorVolumes = new List<ProbeReferenceVolume.Volume>();
-            ProbeVolumesToVolumes(ref probeVolumes, ref cellVolume, ref indicatorVolumes, ref sceneRefs);
-
-            // Cull all influencers against ProbeVolumes
-            culledVolumes = new List<ProbeReferenceVolume.Volume>();
-            CullVolumes(ref influenceVolumes, ref indicatorVolumes, ref culledVolumes);
+            subdivideSceneCS.GetKernelThreadGroupSizes(kernel, out uint x, out uint y, out uint z);
```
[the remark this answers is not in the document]
alelievr: I hope it's fixed, we already have a couple of these functions in our code :)
But I can comment it if you think it's safer

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/VoxelizeScene.shader
```diff
@@ -0,0 +1,85 @@
...
+            Cull Off
+            ColorMask 0
+            ZWrite Off
+            ZClip Off
+
+            HLSLPROGRAM
+            #pragma vertex vert
+            #pragma fragment frag
+            #pragma target 4.5
+            #pragma enable_d3d11_debug_symbols
```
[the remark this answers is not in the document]
alelievr: fixed

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/VoxelizeScene.shader
alelievr: no, I've removed it

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs.hlsl
FrancescoC-unity: Yup, there is a needAccessor in the attribute GenerateHLSL

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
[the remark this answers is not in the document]
FrancescoC-unity: Nah it's fine, just leaving a comment in case we have problems later on at least we remember this is here

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs.hlsl
alelievr: Thanks

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
alelievr: Will do something for terrain in the next PR

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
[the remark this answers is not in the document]
alelievr: 👍

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeAuthoring.cs
```diff
@@ -91,47 +92,54 @@ internal static ProbeReferenceVolumeProfile CreateReferenceVolumeProfile(Scene s
...
-        [System.NonSerialized]
+        [NonSerialized]
         ProbeVolumeAsset m_PrevAsset = null;
 #endif
         public ProbeVolumeAsset volumeAsset = null;
 
         internal void QueueAssetLoading()
         {
-            if (volumeAsset == null || m_Profile == null)
+            if (m_Profile == null)
```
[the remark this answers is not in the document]
alelievr: I've split the function in two

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
FrancescoC-unity: Ping on this. If it does look as hideous as it did before, we might need to _not_ do this change and let the dilation happen all the time (or at least make it default to ON)

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeSubdivide.compute
FrancescoC-unity: I brainfarted and saw it as buffer for whatever reason :D

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolumeProfile.cs
```diff
@@ -15,39 +15,40 @@ internal enum Version
...
         /// </summary>
-        public CellSize cellSizeInBricks = CellSize.CellSize81;
+        public int cellSizeInBricks => (int)Mathf.Pow(3, simplificationLevels);
 
         /// <summary>
-        /// The size of a Brick.
+        /// The minimum distance between two probes in meters.
         /// </summary>
-        [Min(0.1f)]
+        [Min(0.01f)]
```
FrancescoC-unity, line 35: That's way too small :-( I don't think we should allow such a  low value

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
alelievr: Yes, I removed it

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
```diff
@@ -3,236 +3,598 @@
...
+
+        static int s_ClearBufferKernel;
+        static int s_ClearKernel;
+        static int s_JumpFloodingKernel;
+        static int s_FillUVKernel;
+        static int s_FinalPassKernel;
+        static int s_VoxelizeProbeVolumesKernel;
+        static int s_SubdivideKernel;
+
+        static ComputeShader _subdivideSceneCS;
```
FrancescoC-unity, line 154: Is this ever released? I don't think it is auto released upon destruction.

alelievr: We don't need to release it (there is no method for it)

FrancescoC-unity (approved):
We have some issues (possibly and probably unrelated?) that we need to tackle, but let's not block this big PR and let's look at the issues later.

JulienIgnace-Unity (approved):

PR #4471, opened 2021-05-06
Added a toggle to render the volumetric clouds locally or in the skybox.

anisunity:
Most of the users will use the volumetric clouds without having the camera go into/above them or have the geometry intersect with the cloud volume. This PR adds a toggle in the volume component that allows them to have the clouds simply as a background visual on top of the current Sky. When in background mode, the user doesn't have to override his far clip plane which can be useful in many user cases.


anisunity:
I tested this on the sample scene and for me the shadows and the planar are also working properly with the feature. The  non far-plane adjustment also works properly. I still think it is worth a quick smoke test @remi-chapelain

remi-chapelain (approved):
Tested to see with the new "skybox" clouds that : 
- Planar reflection were still working
- Skies were still working
- Cloud layer was still displayed between the sky and the clouds
- Shadows were still castings

Checked tooltips and documentation to inform the change

Also modified automated tests because "skybox" clouds is the new default to make sure that the tests still passes
Added a new tests to test "skybox" clouds.

Approving the feature but screenshots needs to be updated so 
!!!**NOT READY FOR MERGE**!!!

anisunity:
Found a small issue in the code that i need to fix

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricClouds.compute
```diff
@@ -188,10 +194,18 @@ CloudRay BuildRay(uint2 halfResCoord)
     ray.direction = normalize(pos);
 
     // Compute the max cloud ray length
+#ifdef LOCAL_VOLUMETRIC_CLOUDS
     ray.maxRayLength = lerp(_ProjectionParams.z, length(pos), isOccluded);
+#else
+    ray.maxRayLength = 200000.0f;
```
sebastienlagarde, line 200: it looks like a bit an arbitrary large number, can we at least setup it under a nameful define? / also add a comment about why this value, thanks

anisunity: Will do

anisunity: @sebastienlagarde done

sebastienlagarde:
there is discrepency on the volumetric cloud screenshots, not sure it is expected, it is darker?

remi-chapelain:
Screenshots are up to date ! 👍

PR #4474, opened 2021-05-06
[ShaderGraph] [2021.2] BranchOnInputConnection uses "Not Connected" in Preview

cdxntchou:
### Purpose of this PR
Currently the node will default to "Connected" in the subgraph previews, which makes it less useful when using the feature to provide custom default values, like accessing mesh uv.

This PR changes the behavior so in preview mode within the subgraph, it always displays the "Not Connected" option.

It also removes a ShaderStage restriction on the original node (no reason it can't be used in Vertex stage).

It also fixes https://fogbugz.unity3d.com/f/cases/1334621/.

Before:  Preview does not use the uv0 default.
image

After:  Preview uses the uv0 default:
image

The behavior of the sub-graph outside of preview is unchanged, this only affects previewing when editing the subgraph using the branch node.

There are other use cases in which the user might want to have control over which value to use in previews (Connected or Not Connected).

In particular, if the user is using the branch node to enable or disable some part of the functionality based on a connection, they may wish to preview both possibilities.  In this case, we should probably add some way for the user to directly control what it does in Preview mode.

I'm leaving that part out of this PR, as it probably requires a design spike to look at the best control.

---
### Testing status

image
- [x] Tested two subgraphs using the branch on input connection node.
- [x] Verified the connected mode generates the proper behavior, both in outer graph preview and in final shader in scene.
- [x]  Verified the not connected mode generates the proper behavior, both in outer graph preview and in final shader in scene.
- [x] Verified that subgraphs using the BOIC node can now be connected to the vertex stage (as long as they don't contain any other stage contrained nodes).
- [x] Tested subgraph-within-a-subgraph with both subgraphs using similar branch nodes.
- [x] Tested deleting a subgraph with a property marked 'Use Custom Binding' -- parent graphs open and can be edited.
- [x] Tested four levels of subgraph nesting, all using BOIC nodes.
image

Yamato:

ShaderGraph PR Job: 🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Fconnection-preview-default/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_trunk/6768724/job/pipeline

---
### Comments to reviewers
Note that I changed the BOIC node to be an AbstractMaterialNode instead of a CodeFunctionNode, as the CodeFunctionNode does not allow customizing behavior in Preview mode.


russoh (approved):
This makes sense to me. The "Not Connected" subgraph is the one that contains the (complex) default value, and so always using that in preview seems most useful.

russoh (changes requested):
(Removed comment, red herring).

russoh (approved):

jessebarker (approved):
LGTM

LandonTownsendUnity (approved):

PR #4476, opened 2021-05-07
Universal Render Settings Upgrader

Verasl:
### Purpose of this PR
This PR adds a Converter to the Render Pipeline Converter that will generate and setup multiple URP Assets based off settings found in both Quality Settings and Graphics Tiers.
This should allow users to more easily transition to using URP in a builtin project by matching equivalent settings to get a similar experience off the bat rather than living with the defaults of a new URP Asset.
This also creates Renderer Assets that match the current rendering path in the Graphics Tiers, a follow up PR will add support for checking Cameras throughout the project and creating an additional Renderer if the project uses both Forward and Deferred rendering throughout.

Currently Settings that convert 1:1 are:

- HDR
- MSAA
- Shadows Off/Hard/Soft
- Reflection Probe Blending
- Reflection Probe Box Projection
- Shadow Distance
- Shadow Cascades
- Shadow Cascade Splits

We make best guesses on these settings due to them not working the same under the hood in URP:

- Soft Particles
  - Currently if enabled, I enable the depth texture
- Pixel Light Count
  - Currently if 0, lights are off but should I look at setting them to vertex instead?
  - If set to 1, I count this as the main light rather than 1 additional light
  - If > 1 then I enable Additional lights up to max URP light count(currently 8 ) which differs from a seemingly infinite builtin amount
- Shadow Resolution
  - Builtin had Low/Med/High/etc.. which dont map 1:1 to any resolution as it is based on many things
  - I based this off 1080p view on OSX editor to get the shadow map resolution at this point
  - Both main light and additional light atlas shadows are set the same
  - I do not touch the Scalability settings for the additional light shadows
- Render Path
  - Right now this is based off the setting in graphic tier 3 for the current build platform for all pipeline quality levels

---
### Testing status
Simple testing of converting the Foundation Test project quality levels.
This PR needs some real world testing projects, with a varied quality/graphics tier setup to truly figure out if it does the job.

---
### Comments to reviewers
Very early WIP, code needs full review.
This does not upgrade to projects using multiple render paths in the same project.
As Graphics Tiers have been removed from SRP there is not easy alternative so only the 3rd Tier and current build platform are taken into consideration when upgrading.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
@@ -713,10 +713,8 @@ namespace UnityEngine.Rendering.Universal
             // Descriptor settings                                            /
             ///////////////////////////////////////////////////////////////////
 
-            // TODO: We should guarantee that additional camera data and renderer always exists for URP.
-            // When additional camera data and renderer are not present we assume forward renderer by default which supports MSAA.
             var renderer = additionalCameraData?.scriptableRenderer;
-            bool rendererSupportsMSAA = renderer == null || renderer.supportedRenderingFeatures.msaa;
+            bool rendererSupportsMSAA = renderer != null && renderer.supportedRenderingFeatures.msaa;
 
             int msaaSamples = 1;
             if (camera.allowMSAA && asset.msaaSampleCount > 1 && rendererSupportsMSAA)


com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        #endregion
+
+        private enum Parent
+        {
+            Prefab,
+            PrefabVariant,
+            Scene,
+        }
```
svens-unity, line 299: I dont believe this is being used?

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+            public bool ReflectionProbeBoxProjection;
+            public bool ReflectionProbeBlending;
+            public bool CascadeShadows;
+            public bool HDR;
+            public RenderingPath RenderingPath;
+        }
+
+        private class SettingsItem { }
+
+        private class RenderSettingItem : SettingsItem
```
svens-unity, line 273: Do we need both classes?

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEditor.Search;
```
svens-unity, line 4:
```suggestion
```
I don't believe these are being used.

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+using UnityEngine;
+using UnityEngine.Rendering;
+using UnityEngine.Rendering.Universal;
+using ShadowQuality = UnityEngine.ShadowQuality;
+using ShadowResolution = UnityEngine.ShadowResolution;
+
+namespace UnityEditor.Rendering.Universal
```
svens-unity, line 11: More of a general question; should all the converter things get it's own namespace?

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+        public override Type conversion => typeof(BuiltInToURPConverterContainer);
+
+        // Used to store settings specific to Graphics Tiers
+        GraphicsTierSettings m_GraphicsTierSettings;
+
+        // Settings items, currently tracks Quality settings only
+        List<SettingsItem> m_SettingsItems = new List<SettingsItem>();
+
+        // List of the rendering modes required
+        List<RenderingMode> m_RenderingModes = new List<RenderingMode>();
```
svens-unity, line 30: Isn't it safer to create the lists in `OnInitialize`? What if `OnInitialize` is called twice - we might still have entries in our list from the previous initialization.

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+            var defaultIndex = -1;
+            var renderers = new List<ScriptableRendererData>();
+            if (m_RenderingModes.Contains(RenderingMode.Forward))
+            {
+                renderers.Add(CreateRendererDataAsset(path, RenderingPath.Forward, "ForwardRenderer"));
+                if (GetEquivalentRenderMode(m_GraphicsTierSettings.RenderingPath) == RenderingMode.Forward)
+                    defaultIndex = m_RenderingModes.IndexOf(RenderingMode.Forward);
+            }
+
+            if (m_RenderingModes.Contains(RenderingMode.Deferred))
```
svens-unity, line 142: Should this be an `else if`? Assuming OnInitialize is called once, we are only adding one item to the list ever; 
`m_RenderingModes.Add(GetEquivalentRenderMode(tier.renderingPath));`
If this is the case do we even need to use lists?

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+        {
+            var item = context.item;
+            // is quality item
+            if (m_SettingsItems[item.index].GetType() == typeof(RenderSettingItem))
+            {
+                GeneratePipelineAsset(item.index, m_SettingsItems[item.index] as RenderSettingItem);
+            }
+        }
+
+        private void GeneratePipelineAsset(int index, RenderSettingItem settings)
```
svens-unity, line 117:
```suggestion
        private void GeneratePipelineAsset(RenderSettingItem settings)
```
This is not being used.

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// <param name="asset">Pipeline asset to set</param>
+        /// <param name="settings">The ProjectSettingItem with stored settings</param>
+        private void SetPipelineSettings(UniversalRenderPipelineAsset asset, RenderSettingItem settings)
+        {
+            // General
+            asset.supportsCameraDepthTexture = settings.SoftParticles;
+
+            // Quality
+            asset.supportsHDR = m_GraphicsTierSettings.HDR;
+            asset.msaaSampleCount = settings.MSAA == 0 ? 1 : settings.MSAA;
```
svens-unity, line 187: How does URP handle this setting? If it is set to 0 in the settings.MSAA we set it to 1. If it is set to 1 in the settings.MSAA we set it to 1 too? Are these settings equivalent?

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+using UnityEngine.Rendering;
+using UnityEngine.Rendering.Universal;
+using ShadowQuality = UnityEngine.ShadowQuality;
+using ShadowResolution = UnityEngine.ShadowResolution;
+
+namespace UnityEditor.Rendering.Universal
+{
+    internal class RenderSettingsConverter : RenderPipelineConverter
+    {
+        public override string name => "Rendering Settings Upgrade";
```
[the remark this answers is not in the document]
Verasl: I was following the material upgrade title, but now that you mention it I think it actually is nicer if we commit the word all together and just have **Render Settings** as this is already inside the converter so it should be contrived from the window it is in.

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+            switch (value)
+            {
+                case 0: // low
+                    return 1024;
+                case 1: // med
+                    return 2048;
+                case 2: // high
+                    return 4096;
+                case 3: // very high
+                    return 4096;
```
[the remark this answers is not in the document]
Verasl: yea for now, this lines up a close as I could with builtin based on a 1080p screen on desktop, builtin scales shadow resolution based of many criteria so it is hard to match 1:1 but these values should be similar on average.

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
Verasl: Added it to the new Converters namespace

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
Verasl: So this is a little future proofing(for the camera work to be adding in later PR) that the rendering modes will actually contain more than a single value as it will check each camera and add any that differ from using graphics settings.
This is why it's a list atm and also is not a 'else if'.

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
Verasl: so quality settings use 0, 2, 4, 8 for MSAA while RTs use 1, 2, 4, 8 and URP follows this convention, so all 0 values becomes 1 and the rest should line up as expected.

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+                    return RenderingMode.Forward;
+                case RenderingPath.DeferredLighting:
+                case RenderingPath.DeferredShading:
+                    return RenderingMode.Deferred;
+                default:
+                    return RenderingMode.Forward;
+            }
+        }
+
+        #region Structs
```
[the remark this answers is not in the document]
Verasl: changed to 'Data'

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
Verasl: Again this is part fo future proofing, I actually had the camera class here too, can remove now if you think it's worth it otherwise it will be used in the following camera PR.

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
Verasl: removed

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
svens-unity: I think its okay to keep 👍

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
svens-unity: My main worry here is the default index, We need it to get set otherwise we end up with an invalid index. Setting it is behind a few checks which make it brittle.  Also because we check forward first deferred will be preferred as default when both are active?

svens-unity (approved):

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
Verasl: Changed the index to be 0 default, which will make sure at least the first Renderer added to the pipeline asset will be the default one.
Later when the camera stuff is added and there is the chance of having both forward and deferred, forward will alway be added to the pipeline asset first, but depending on what is set in m_GraphicsTierSettings will be used as the default index.

CertifiedRice (approved):

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+                GetEquivalentMainlightShadowResolution((int)settings.ShadowResolution);
+
+            // Additional Lights
+            asset.additionalLightsRenderingMode = settings.PixelLightCount <= 1
+                ? LightRenderingMode.Disabled
+                : LightRenderingMode.PerPixel;
+            asset.maxAdditionalLightsCount = Mathf.Max(0, settings.PixelLightCount - 1);
+            asset.supportsAdditionalLightShadows = settings.Shadows != ShadowQuality.Disable;
+            asset.additionalLightsShadowmapResolution =
+                GetEquivalentAdditionalLightAtlasShadowResolution((int)settings.ShadowResolution);
```
andrew-theisen, line 207: This completely disables lights that should probably stay enabled. On Low settings in VikingVillage with Pixel Light Count = 0, lights are still enabled in built-in but set to per-vert. Completely disabling them might be confusing to users, especially with the main light or when deferred rendering is used.

andrew-theisen: Another question on this same block is when shadows are set to Hard Only in built-in, should we completely disable shadows in URP or set them to really low resolution? Maybe do this for only the Main Light at least?

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+                    defaultIndex = m_RenderingModes.IndexOf(RenderingMode.Forward);
+            }
+
+            if (m_RenderingModes.Contains(RenderingMode.Deferred))
+            {
+                renderers.Add(CreateRendererDataAsset(path, RenderingPath.DeferredShading, "DeferredRenderer"));
+                if (GetEquivalentRenderMode(m_GraphicsTierSettings.RenderingPath) == RenderingMode.Deferred)
+                    defaultIndex = m_RenderingModes.IndexOf(RenderingMode.Deferred);
+            }
+
```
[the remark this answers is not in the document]
Verasl: solved with starting default index as 0

com.unity.render-pipelines.universal/Editor/Converter/Converters/RenderSettingsConverter.cs
Verasl: Have tweaked the pixel light count logic to be more in line with builtin, if 0 lights, main light is still pixel, and I set 4 vertex lights for additional. If any number of pixel lights, these are only counted as Additional lights, so ignoring the main light which builtin seems to also do.

As for soft shadows, this is handled further down, pretty much if builtin supported 'All' I enable soft shadows, builtin it was more per light, but I think in URP we will be moving towards having a unified shadowing across all lights.

erikabar (approved):
Hakeem's testing doc: https://docs.google.com/document/d/1K3bfyz43_grrDUWDC_V3mo6ALkG0As1vh6Bnws3-PSI/edit#

there is one PR issue that will be addressed in a follow up PR

phi-lira (approved):

phi-lira (approved):

erikabar (approved):

phi-lira:
I reverted the MSAA commit. It will land in separate PR as it seems it needs some reference image updates due to MSAA and I don't want to block this PR.

PR #4479, opened 2021-05-07
SwapBuffer for Universal Renderer

robin-demoor:

### Checklist for PR maker

---
### Purpose of this PR
This PR adds a swap buffer system for the color target in the Universal Renderer. This avoids some management of temporary RT's and allows us to give the users a function that allows for easy post-processing.

This new API: `public void Blit(CommandBuffer cmd, ref RenderingData data, Material material, int passIndex = 0)`
Internally this will use the current color target as a source and the other buffer in the swapsystem as the destination to apply the material. This blit then swaps the 2nd target to be the color target of the renderer. This way users can easily apply effects to the color buffer without needing knowledge about the swapbuffer system.

---
### Testing status

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
@@ -931,7 +931,6 @@ namespace UnityEngine.Rendering.Universal
                         ConfigureCameraColorTarget(m_ColorBufferSystem.GetBackBuffer(cmd).id);
 
 
-                    ConfigureCameraColorTarget(m_ColorBufferSystem.GetBackBuffer(cmd).id);
                     m_ActiveCameraColorAttachment = m_ColorBufferSystem.GetBackBuffer(cmd);
                     cmd.SetGlobalTexture("_CameraColorTexture", m_ActiveCameraColorAttachment.id);
                 }
@@ -1034,7 +1033,12 @@ namespace UnityEngine.Rendering.Universal
         internal override void SwapColorBuffer(CommandBuffer cmd)
         {
             m_ColorBufferSystem.Swap();
-            ConfigureCameraTarget(m_ColorBufferSystem.GetBackBuffer(cmd).id, m_ColorBufferSystem.GetBufferA().id);
+
+            //Check if we are using the depth that is attached to color buffer
+            if (m_ActiveCameraDepthAttachment == RenderTargetHandle.CameraTarget)
+                ConfigureCameraTarget(m_ColorBufferSystem.GetBackBuffer(cmd).id, m_ColorBufferSystem.GetBufferA().id);
+            else ConfigureCameraColorTarget(m_ColorBufferSystem.GetBackBuffer(cmd).id);
+
             m_ActiveCameraColorAttachment = m_ColorBufferSystem.GetBackBuffer();
             cmd.SetGlobalTexture("_CameraColorTexture", m_ActiveCameraColorAttachment.id);
         }


com.unity.render-pipelines.universal/Runtime/RenderTargetBufferSystem.cs
```diff
@@ -0,0 +1,74 @@
+using System;
```
robin-demoor, line 1: The working of this class is not how we want it to be. I made it like this to quickly tests if we can achieve a swapbuffer system without RTHandles. I'm working on a better solution.

If you have any suggestions on how we should handle this class. Please feel free to suggest.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -509,7 +511,7 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
                 ConfigureCameraTarget(activeColorRenderTargetId, activeDepthRenderTargetId);
             }
 
-            bool hasPassesAfterPostProcessing = activeRenderPassQueue.Find(x => x.renderPassEvent == RenderPassEvent.AfterRendering) != null;
+            bool hasPassesAfterPostProcessing = activeRenderPassQueue.Find(x => x.renderPassEvent == RenderPassEvent.AfterRenderingPostProcessing) != null;
```
robin-demoor, line 514: This change is unrelated to this PR but I think it's a bug. Thought I would adjust this as well since I encountered the issue a few times.

com.unity.render-pipelines.universal/Runtime/RenderTargetBufferSystem.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public void Swap()
+        {
+            m_FirstIsBackBuffer = !m_FirstIsBackBuffer;
+        }
+
+        void Initialize(CommandBuffer cmd)
+        {
+            cmd.GetTemporaryRT(m_NameA, m_Desc, m_FilterMode);
+            cmd.GetTemporaryRT(m_NameB, m_Desc, m_FilterMode);
```
lukaschod, line 55: Where do we release second bufffer, I am guessing the first one is released with old logic in universal renderer?

robin-demoor: It doesn't release yet. 
So by default I think it should always release at the same point as the active color target.
But do we also want an option in the Swap() function (`Swap(bool release frontbuffer)`) so we can specify that we will no longer need a second buffer so we can get rid of the memory sooner?

Although we are never sure if the user will need the second buffer in their own passes so maybe it's not worth trying to get rid of it sooner.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -219,7 +221,7 @@ public UniversalRenderer(UniversalRendererData data) : base(data)
 
             // RenderTexture format depends on camera and pipeline (HDR, non HDR, etc)
             // Samples (MSAA) depend on camera and pipeline
-            m_CameraColorAttachment.Init("_CameraColorTexture");
+            m_ColorBufferSystem = new RenderTargetBufferSystem("_CameraColorAttachment");
```
lukaschod, line 224: Do we set somehwere _CameraColorTexture as property? As users can depend on this property and now it will be gone.

robin-demoor: No, I will add it!

com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
```diff
@@ -294,5 +294,15 @@ public override void FinishRendering(CommandBuffer cmd)
             if (m_CreateDepthTexture)
                 cmd.ReleaseTemporaryRT(k_DepthTextureHandle.id);
         }
+
+        internal override void SwapColorBuffer(CommandBuffer cmd)
```
lukaschod, line 298: Is there a reason why we do not implement it in 2D? IIRC they also have post processing

robin-demoor: Initially this branch was built upon the RTHandles work and they were only going to implement it for UniversalRenderer due to time constraints. Now that I am no longer built on top of their branch I suppose I can try and implement it as well for the 2D renderer? @phi-lira

phi-lira: If they API is not used in 2D today to avoid risk I'd say to add in follow up PR. This is internal API so users can't access it anyway.

com.unity.render-pipelines.universal/Runtime/Deprecated.cs
```diff
@@ -46,6 +46,16 @@ public override void FinishRendering(CommandBuffer cmd)
         {
             throw new NotSupportedException(k_ErrorMessage);
         }
+
```
phi-lira, line 49: Why is it being added in `Deprecated.cs`?

com.unity.render-pipelines.universal/Runtime/Passes/CopyColorPass.cs
```diff
@@ -75,6 +75,13 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
             }
 
             CommandBuffer cmd = CommandBufferPool.Get();
+
+            //It is possible that the given color target is now the frontbuffer
+            if (source == renderingData.cameraData.renderer.GetCameraColorFrontBuffer(cmd))
+            {
+                source = renderingData.cameraData.renderer.cameraColorTarget;
```
phi-lira, line 82: This highlights to me that we should make clear to users to not cache camera target and only access it before execute.
 To make it more clear to user we should update this pass to avoid caching the `source`. Not issue for this PR though, we can do that in follow up PR.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -439,6 +439,21 @@ public void Blit(CommandBuffer cmd, RenderTargetIdentifier source, RenderTargetI
...
         }
 
+        /// <summary>
+        /// Add a blit command to the context for execution. This applies the material to the color target.
+        /// </summary>
+        /// <param name="cmd">Command buffer to record command for execution.</param>
+        /// <param name="data">RenderingData to access the active renderer.</param>
+        /// <param name="material">Material to use.</param>
+        /// <param name="passIndex">Shader pass to use. Default is 0.</param>
+        public void Blit(CommandBuffer cmd, ref RenderingData data, Material material, int passIndex = 0)
```
phi-lira, line 449: Maybe a specific name for this `Blit`?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -439,6 +439,21 @@ public void Blit(CommandBuffer cmd, RenderTargetIdentifier source, RenderTargetI
             cmd.Blit(source, destination, material, passIndex);
         }
 
+        /// <summary>
+        /// Add a blit command to the context for execution. This applies the material to the color target.
```
phi-lira, line 443: Needs more description to tell this Blit happens with camera color target as source and dest?

com.unity.render-pipelines.universal/Runtime/RenderTargetBufferSystem.cs
phi-lira: If we release after swapping we have to make sure to get if from pool upon access again. Is there any downside of always releasing instead of adding option?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
phi-lira: Thank you for fixing it.

com.unity.render-pipelines.universal/Runtime/RenderTargetBufferSystem.cs
robin-demoor: I am no expert on the workings of GetTemporaryRT but as I understand it. Releasing it every time should actually be fine since everything is pooled anyway. That way we get rid of it as soon as possible and it really doesn't cost anything extra if we need it 2 lines further in the code again. But like I said I am no expert on it so maybe someone can confirm this?

lukaschod: Yes that is correct, we should release at the end of the frame or even sooner if possible. Because once it is released can be immedetiatelly re-used by other requests, which would create another one if not present

com.unity.render-pipelines.universal/Runtime/Deprecated.cs
robin-demoor: Because it is marked as an abstract function and the ForwardRenderer is deprecated but still inherits from ScriptableRenderer so it needs to have the function.

sandy-carter-unity (approved):

Jonasmortensen (approved):
Tested API from custom renderer features and everything looks good 👍

phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
Verasl: Agreed, needs to mention what it is doing behind the scenes with the swap a little bit

Verasl (approved):
Going to approve, feature works great, would love us to make sure we have decent docs and maybe an example usage in the API docs, but this can come after cutoff.

erikabar (approved):
Tested UniversalRenderingExamples project, also created a custom scene that covers all Post Processing effects, multiple cameras + camera stacking, SSAO, new Scripting API. Most of the issues were fixed, the remaining one is on OpenGLCore which will be fixed in a follow-up PR. 
Testing doc: https://docs.google.com/document/d/127tH2YudVeZppmS0-QBp6LKMRL1LUeRIroxF3Pkwt8E/edit#heading=h.6f6jaidlu5z

vaidasma:
Testing results: feature has issues


Projects used: SwapBuffer 2
Tested on devices:

:green_circle: Android (Vulkan / Gles3 / Gles2):
VLNQA00010, Samsung Galaxy S8 (SM-G950U), Android 7.0, CPU: Snapdragon 835 MSM8998, GPU: Adreno (TM) 540
VLNQA00267, Samsung Galaxy S10+ (SM-G975F), Android 11, CPU: Exynos 9 (9820), GPU: Mali-G76
----------, Google Pixel 5 (Pixel 5), Android 11, CPU: Snapdragon 765/765G, GPU: Adreno (TM) 620
VLNQA00267, Samsung Galaxy S10+ (SM-G975F), Android 11, CPU: NOT FOUND, GPU: Mali-G76
android



:green_circle: iOS (Metal):
 --------------, iPhone 12, iOS:14.4.2,  A14
 VLNQA00237,  iPhone 6s, iOS:14.3, A9
ios


Issues found:
:x: Gles3 appears darker on Mali-G76
Mali76




:question: Gles2 does not support Post Processing. Is that intended?
Gles2

MarioUnity (commented):
XR QA Review:

Editor: 2021.2.0a17.2411
Project Used: SwapBuffer2.zip, provided by Erika Baraisyte

In general, SwapBuffers has issues with XR. Using SwapBuffer with Multiview or Single Pass Instancing results in a black or gray screen on OpenGLES3 or Vulkan. In multipass, the post processing effect is not applied. 

See this screenshot on Oculus Quest using Vulkan and multipass, compared to general Android Mobile:
Quest2_Vulkan_multipass_postprocess

On Rift VR desktop, the issues are even worse:
RIft_DX11_multipass

Please coordinate with Thomas Zeng to see what work needs to be done to support this feature in XR.

MarioUnity (changes requested):
Please see [https://github.com/Unity-Technologies/Graphics/pull/4479#pullrequestreview-658464804](url)

vaidasma:
Attaching RenderDoc file for Gles3 Mali-G76 black material:

https://drive.google.com/file/d/1OFyyuF_q0MrC93uAGvhNF8ldnMSYcwQC/view?usp=sharing

vaidasma:
:wrench: 🟢 - Mali-G76 on Gles3 black materials issue fixed

fixedSwapOriginal png_R58M22TD79M

vaidasma (approved):
Final pushes seem to not have broken anything, approving.

MarioUnity (approved):
Spoke to Robin. He has tested his updates with Oculus desktop and has it working, so we know the feature works with XR. We will continue to check additional platforms and file bugs if any work is needed.

phi-lira:

PR #4483, opened 2021-05-07
[HDRP][Path Tracing] Stacklit material support

eturquin:
Added Stacklit material support for path tracing. Before, objects with stacklit shaders assigned would render black, now they produce the expected look.
Most of the parameterization is supported (haze settings, coating settings, different normals), with the exception of the alpha-blend-like transparent mode (could be added, but unless the Lit where simulation of alpha blend is done by modifying refraction parameters, here there is no refraction to begin with, and it would have to be added just to support this mode...).

Has been tested on ad hoc stacklit materials, and this scene:
https://github.com/Unity-Technologies/MeasuredMaterialLibraryHDRP

Would be good to test on additional stacklit variants.

### DIFF
--- a/TestProjects/HDRP_DXR_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D12/None/107_ReflectionsHybridFullRes.png
+++ b/TestProjects/HDRP_DXR_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D12/None/107_ReflectionsHybridFullRes.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:b7f673f8970ada9ae01b005ed21b5d421499e888cee4cf807be85dfeed3b041e
+size 164303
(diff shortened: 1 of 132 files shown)


eturquin:
stacklit

remi-chapelain:
Haven't fully tested yet but we need to update documentation that state that we don't support stacklit with Pathtracer. 
@Vic-Cooper
Thanks !

remi-chapelain (changes requested):
Made an in-depth depth of the stacklit in more complex scenes. 
- Coat / Coat Normal / Coat IOR / Coat Extinction Color / Thickness / Smoothness
- Dual Spec Lobes / Anisotropy
- Iridescence
- Anisotropy on base layer
- SSS / Transmission

For now there's a few issues and one that will block the PR. 
- Coat normal creates some NaNs that ends in the pathtracer rendering full screen black (Blocker)
- SubSurfaceScattering seems to differ in complex materials(Not used a lot with stacklit)
- Transparent seems to differ a lot from their raster counterpart (not used a lot with stacklit) 

**(Nans with coat normal)**
Uploading 0f7cbe31687e848167e2e64e643eb47b.mp4…

remi-chapelain:
Go for merge. 
- The blocking point has been fixed with the coat normal. 

We should just note on documentation the limitation of this support, mostly that it won't match rasterization for these things. @Vic-Cooper
- Subsurface scattering won't look the same  (for stacklit and fabric)
- Transparency won't look the same for stacklit 
- Refraction clear coat on underlying layers won't look the same on stack lit (subtle so not sure it's needed). 

@eturquin if you have more things in minds. 

I'm also building a test scene for stacklit feature for pathtracer.

remi-chapelain (approved):

sebastienlagarde (approved):

PR #4488, opened 2021-05-07
Remove depth prepass when MSAA is enabled

manuele-bonanno:
### Purpose of this PR
Currently, when the "Depth texture" option is selected or when post processing effects require depth, an extra depth prepass happens.
The reason for this was the lack of support for highp Texture2DMS. This is not the case anymore so it is now possible to copy the MSAA'd depth of the Opaque pass.

Removing the depth prepass can give significant gains on mobile platforms, especially for complex scenes. Enabling copying of MSAA'd depth also allows us to resolve MSAA depth on platforms with no hardware depth resolve support, enabling techniques which would require the resolved depth (i.e. depth priming + Vulkan + MSAA).

---
### Testing status
local (iOS/Metal) + Yamato

---
### Comments to reviewers

how to reproduce the depth prepass scenarion:
enable MSAA, enable "Depth Texture"

**Before:**
image

**After:**
image

image


manuele-bonanno:
after running all the tests against custom revision 276a9a6152d8, as suggested in #srp-nightly-status, all the tests pass except some ShaderGraph related failures and a single URP_Terrain failure (terrain holes) which should be unrelated to this PR

phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -995,9 +995,7 @@ bool CanCopyDepth(ref CameraData cameraData)
             bool supportsDepthTarget = RenderingUtils.SupportsRenderTextureFormat(RenderTextureFormat.Depth);
             bool supportsDepthCopy = !msaaEnabledForCamera && (supportsDepthTarget || supportsTextureCopy);
 
-            // TODO: replace hasHiddenSurfaceRemovalOnGPU to hasTiledGPU if it ever gets exposed. Then we can uncomment this and guarantee there's no regression on tiled GPUs.
-            //bool msaaDepthResolve = msaaEnabledForCamera && SystemInfo.supportsMultisampledTextures != 0 && !SystemInfo.hasHiddenSurfaceRemovalOnGPU;
-            bool msaaDepthResolve = false;
+            bool msaaDepthResolve = msaaEnabledForCamera && SystemInfo.supportsMultisampledTextures != 0;
```
phi-lira, line 998: Shouldn't we check here for `multisampleAutoResolve`?

phi-lira (approved):
Missing changelog.

phi-lira:
It would be nice to have a doc page to add information to users on how to handle depth resolve in Unity. 
I also remember there was something specific to auto resolve in some platform, so maybe we could document these as well.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
manuele-bonanno: multisampleAutoResolve doesn't affect this

manuele-bonanno:
> It would be nice to have a doc page to add information to users on how to handle depth resolve in Unity.
> I also remember there was something specific to auto resolve in some platform, so maybe we could document these as well.

Currently there is no engine depth resolve support yet. I have a trunk PR in review which adds it for Metal (as a start), but hasn't landed yet

erikabar (approved):

manuele-bonanno:
the test failures seem unrelated to this PR and the same as master on the 7th of May

nigeljw-unity (approved):

PR #4491, opened 2021-05-07
Fixed edge connections being replaced to not run the auto block remov…

joshua-davis:
…al logic. Fixes fogbugz 1334341

### Checklist for PR maker

---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1334341/
Replacing the connection to block node that was considered inactive from the target's perspective would hide the node if "auto add or remove blocks" is enabled.

---
### Testing status
Make an empty graph and manually add the "Base Color" block.
Connect a node to the "Base Color" Block
- [x] Replace the connection to the block with itself. The block doesn't disappear.
- [x] Drag the edge off and back on the block node (in one click). The block doesn't disappear.
- [x] Replace the connection with the block with the connection from another node. The block doesn't disappear.
- [x] Convert the block to a property (e.g. with a color node). The block doesn't disappear.
- [x] Convert the block to a sub-graph (might require more than a single block like color -> add). The block doesn't disappear.
- [x] Delete the edge. The block disappears
- [x] Delete the node connected to the block. The block disappears.
- [x] Drag the edge connection to something else (not connected to the block). The block disappears.
- [x] Drag the edge off the block and release. The edge and the block disappear.

---
### Comments to reviewers
Added a catch all in `HandleGraphChanges` to remove any unused blocks after all operations so that things like replacing an edge wouldn't trigger the check in-between the remove and add. This was slightly tricky as it requires checking after edges are removed which is after block views are removed. This requires visited the block views a second time so that the views are properly deleted. In addition, a flag was added to GraphData so this path is only checked when an edge is removed.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -84,6 +84,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue where an integer property would be exposed in the material inspector as a float [1330302](https://issuetracker.unity3d.com/product/unity/issues/guid/1330302/)
 - Fixed an issue where upgrading from an older version of ShaderGraph would cause Enum keywords to be not exposed [1332510]
 - Fixed an issue where a missing subgraph with a "Use Custom Binding" property would cause the parent graph to fail to load [1334621] (https://issuetracker.unity3d.com/issues/shadergraph-shadergraph-cannot-be-opened-if-containing-subgraph-with-custom-binding-that-has-been-deleted)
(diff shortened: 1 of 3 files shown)


jessebarker (approved):
LGTM

ghost (approved):
Tested that the bug is fixed for both vertex and fragment blocks
Tested that one or multiple targets still worked
Tested the bug is fixed for inactive blocks when targets are active
Tested that connecting something else to the inactive block also doesn't repro the bug
Tested deleting blocks, undoing, redoing, reordering blocks, deactivating and reactivating blocks with something plugged in by adding and removing targets

PR #4496, opened 2021-05-10
Strip DEBUG_DISPLAY shaders from URP test projects.

arttu-peltonen:
### Purpose of this PR

This PR is intended to 
- help decrease Yamato build times cause by tons of shader variants in URP
- hopefully allow some URP iPhone tests [like this one](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Furp_terrain-iphone-metal.yml%2523URP_Terrain_iPhone_Metal_Standalone_cache_il2cpp_Linear_CUSTOM-REVISION/6702252/job/artifacts) to start passing - they currently fail due to out-of-memory error, possibly because of excessive shader variants.
 
All URP test projects now include a scripting define `URP_TEST_AGGRESSIVE_SHADER_STRIPPING`, which is used to strip debug shader variants from the build.

---
### Testing status
Made a build of UniversalGraphicsTest_2D and made sure the shaders are stripped.


phi-lira (approved):

arttu-peltonen:
@phi-lira looks like the test still fails, but I guess that was expected - [the job was already failing on master](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Furp_terrain-iphone-metal.yml%2523URP_Terrain_iPhone_Metal_Standalone_cache_il2cpp_Linear_CUSTOM-REVISION/6702252/job/artifacts) before my [debug views PR](https://github.com/Unity-Technologies/Graphics/pull/4089) went in.

I would still expect this to speed up running test project Yamato jobs, but it's hard to quantify given how much variance there is on the build times, and AFAIK the number of variants is not reported.

phi-lira (approved):

hdb-unity (approved):
Built URP template project with vs. without the new symbol.

No regression seen locally in either case.

Do we need multiple entries of define symbols in each project settings asset?

PR #4499, opened 2021-05-10
[URP]Fix possible recursion loop from invalid camera stack data

alex-vazquez:
### Purpose of this PR
Fixing [1334531](https://fogbugz.unity3d.com/f/cases/1334531/)

---
### Testing status
Check that the bug is fixed

---
### Comments to reviewers
Cyclic dependecy on the camera stack, this was from the refresh of  the camera on the stack from the ctor, as that is always called from the inspector, no need to double call the method.


martint-unity (approved):
LGTM

erikabar (approved):
Verified case 1334531

PR #4502, opened 2021-05-10
[HDRP] Fix issue with velocity rejection when using physically-based DoF.

pmavridis:
### Purpose of this PR
There was flickering when a high velocity rejection value was used with TAA and physically-based DoF (This was reported by the demo team)

The reason is that we have enabled velocity rejection with post-DoF TAA in a previous PR, but it was never working properly, because the motion vectors were not bound. This PR fixes this issue.

---
### Testing status
- I have tested using a repro project that was provided my the demo team.
- DoF tests are green.

Note for QA: we can pair-test this PR if you believe more testing is necessary and/or I can provide test projects.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -1472,17 +1472,11 @@ class TemporalAntiAliasingData
...
-                passData.nextMVLen = builder.WriteTexture(renderGraph.ImportTexture(nextMVLen));
-            }
-            else
-            {
-                passData.prevMVLen = TextureHandle.nullHandle;
-                passData.nextMVLen = TextureHandle.nullHandle;
-            }
+
+            // Note: In case we run TAA for a second time (post-dof), we can use the same velocity history (saves the length of the mv)
+            GrabVelocityMagnitudeHistoryTextures(camera, out var prevMVLen, out var nextMVLen);
```
FrancescoC-unity, line 1477: We only need to read though so the write of nextMV is unnecessary (and should be in fact skipped in shader)

pmavridis: Yes, you are right, I was also thinking skipping this.

FrancescoC-unity (approved):

sebastienlagarde (approved):

pmavridis:
Could not find any other issue, aside from a small fix in reset, so I guess we can merge this for now. 
If there is something else I will create another PR.

PR #4504, opened 2021-05-11
Fix an nda-platform debug shader compilation

arttu-peltonen:
### Purpose of this PR
Fix this failure: https://yamato.cds.internal.unity3d.com/job/6720369


arttu-peltonen:
I found out this is already passing, no need to fix anything.

PR #4211, opened 2021-04-16
[OnHold] Exposable SubTarget property for ShaderGraph

adrien-de-tocqueville:
### Purpose of this PR

C++: https://github.cds.internal.unity3d.com/unity/unity/pull/3285

1.
This PR is adding a checkbox on the left of SubTarget properties on SG to expose/hide them on materials using this Shader

<p float="left">
  
   
</p>
<p align="center">Example with non exposed surface property<br></p>

The underlying generated shaders can be optimized to not generate unecessary #multi_compile for properties that are not exposed.
Some more optimizations can be done, but this will be done in another PR.

Not all properties can be exposed, they are essentially the SurfaceOptions for decal and lit shader types.

The feature is supported by shadergraph but must be explicitely enabled by each SRP. Only HDRP support has been implemented yet

2.
This PR is also implementing dependency tracking between shadergarph and materials so that all materials depending on a shadergraph can be reimported on demand, for example when we change a specific property whose value should be copied in the material.
This goes with a new c++ callback called ValidateMaterial, that will be called automatically to reset keywords and properties when the material is changed

---
### Comments to reviewers
URP: updated the shader gui to use the new c++ callback
Shadergraph: added an option to enable exposable properties
HDRP: same as URP + handle exposable properties

---
### Testing status
Manual tests:
- test upgrading existing shadergraphs (master nodes as well as shader stack)
- check generated code to make sure keyword removal is correct
- check non exposed properties are actually not exposed on the material UI

QA doc: https://confluence.unity3d.com/display/HDRP/Exposable+SubTarget+property+for+ShaderGraph

PR #4508, opened 2021-05-11
Material dependency

adrien-de-tocqueville:
### Purpose of this PR

Stripped down version of https://github.com/Unity-Technologies/Graphics/pull/4211
C++: https://github.cds.internal.unity3d.com/unity/unity/pull/3285

JIRA card: https://unity3d.atlassian.net/browse/UR-1450

This PR is optimizing dependency tracking between shadergraphs and materials: When we change some specific property on a SG, for example when we set a shadergaph to use subsurface sacttering in HDRP, we need to reset the keywords for all materials using this shader. Previously when saving, all materials from the project were loaded and resetted.
Now only the materials loaded in memory are reset.
This is possible thanks to the added callback `ValidateMaterial`, that will be called automatically to reset keywords and properties when a material is changed or imported, that will handle all the materials not in memory at saving time.
This callback will also permit to fix various issues with materials not initialized as well as reset keyword when the active render pipeline is changed.

---
### Testing status

This feature was implicitely tested during testing of the PR mentionned above as well as during testing for Material Variants
HDRP QA doc: https://confluence.unity3d.com/display/HDRP/Exposable+SubTarget+property+for+ShaderGraph

Manual tests:
- tested saving SG correctly updates materials
- tested modyfing .mat file outside of unity and make sure material is validated before being used

---
### Comments to reviewers

Changes are similar on HDRP and URP:
1. Update ShaderGUI to use the new ValidateMaterial callback
2. On shadergraph save, only reset loaded materials

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/HDShaderGUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/HDShaderGUI.cs
@@ -23,8 +23,15 @@ namespace UnityEditor.Rendering.HighDefinition
     /// </summary>
     public abstract class HDShaderGUI : ShaderGUI
     {
-        // The following set of functions are call by the ShaderGraph
-        // It will allow to display our common parameters + setup keyword correctly for them
+        /// <summary>
+        /// Sets up the keywords and passes for the material you pass in as a parameter.
+        /// </summary>
+        /// <param name="material">Target material.</param>
+        [Obsolete("SetupMaterialKeywordsAndPass has been renamed ValidateMaterial (UnityUpgradable) -> ValidateMaterial()", false)]
+        protected virtual void SetupMaterialKeywordsAndPass(Material material)
+        {
+            ValidateMaterial(material);
+        }
 
         /// <summary>
         /// Unity calls this function when it displays the GUI. This method is sealed so you cannot override it. To implement your custom GUI, use OnMaterialGUI instead.


alex-vazquez (approved):
About the ValidateMaterial, are we sure that never will be called with a null material?

If so, everything looks good to me

cinight (approved):
No regression on URP ([test doc](https://docs.google.com/document/d/1NfPDJXaKOE-ripeO6qWflZU4LRkTkWngUOGfmQybhZY/edit#)).

erikabar (approved):
Ming Wai already tested it, approving from URP QA side

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/ShaderGraphMaterialsUpdater.cs
```diff
@@ -50,25 +41,14 @@ static void OnShaderGraphSaved(Shader shader, object saveContext)
...
-                    if ((material != null) && material.shader.name == shader.name)
-                        ShaderUtils.UpdateMaterial(material, ShaderUtils.MaterialUpdateType.ModifiedShader);
-
-                    material = null;
-
-                    // Free the materials every 200 iterations, on big project loading all materials in memory can lead to a crash
-                    if ((i % 200 == 0) && i != 0)
-                        EditorUtility.UnloadUnusedAssetsImmediate(true);
+                    if (materials[i].shader.name == shader.name)
+                        ShaderUtils.UpdateMaterial(materials[i], ShaderUtils.MaterialUpdateType.ModifiedShader);
```
lukaschod, line 46: Why do we need this logic? I just tested and even on shader graph save, depending materials gets their ValidateMaterial message called.

com.unity.render-pipelines.universal/Editor/ShaderGUI/ShaderGraphUnlitGUI.cs
```diff
@@ -23,7 +23,7 @@ public static void UpdateMaterial(Material material, MaterialUpdateType updateTy
             BaseShaderGUI.UpdateMaterialSurfaceOptions(material, automaticRenderQueue: false);
         }
 
-        public override void MaterialChanged(Material material)
+        public override void ValidateMaterial(Material material)
         {
             if (material == null)
                 throw new ArgumentNullException("material");
```
lukaschod, line 29: I am guessing this needs removed too

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalUI.cs
```diff
@@ -30,18 +30,7 @@ enum Expandable : uint
 
         protected override void OnMaterialGUI(MaterialEditor materialEditor, MaterialProperty[] props)
         {
-            // always instanced
-            SerializedProperty instancing = materialEditor.serializedObject.FindProperty("m_EnableInstancingVariants");
```
sebastienlagarde, line 34: This part have disappeard and I don't think it is expected  (to setup instancing to true)

sebastienlagarde: ah sorry , see it is setup below

com.unity.render-pipelines.high-definition/Editor/Material/Unlit/BaseUnlitGUI.cs
```diff
@@ -254,18 +254,12 @@ public static void SetupBaseUnlitKeywords(this Material material)
...
-            bool doubleSidedGI = false;
             if (doubleSidedGIMode == DoubleSidedGIMode.Auto)
-                doubleSidedGI = doubleSidedEnable;
+                material.doubleSidedGI = doubleSidedEnable;
             else if (doubleSidedGIMode == DoubleSidedGIMode.On)
-                doubleSidedGI = true;
+                material.doubleSidedGI = true;
             else if (doubleSidedGIMode == DoubleSidedGIMode.Off)
-                doubleSidedGI = false;
-            // material always call setdirty, so set only if new value is different
```
sebastienlagarde, line 265: Sanity check you have fixed that on C++ side?

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Editor/Material/Unlit/BaseUnlitGUI.cs
adrien-de-tocqueville: yep

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalUI.cs
adrien-de-tocqueville: I moved it below in SetupCommonDecalMaterialKeywordsAndPass, which is called by the resetkeyword code. That way, we don't need the user to open the inspector to open the flag, and we don't make use of SerializedProperty, which works kind of unreliably for materials

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/HDShaderGUI.cs
```diff
@@ -23,52 +23,9 @@ namespace UnityEditor.Rendering.HighDefinition
...
-        internal protected bool m_FirstFrame = true;
-
         // The following set of functions are call by the ShaderGraph
         // It will allow to display our common parameters + setup keyword correctly for them
 
-        /// <summary>
-        /// Sets up the keywords and passes for the material you pass in as a parameter.
-        /// </summary>
-        /// <param name="material">Target material.</param>
-        protected abstract void SetupMaterialKeywordsAndPass(Material material);
```
adrien-de-tocqueville, line 35: does that count as an API change ? because the HDShaderGUI class is public

com.unity.render-pipelines.universal/Editor/ShaderGUI/ShaderGraphUnlitGUI.cs
adrien-de-tocqueville: yes

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/ShaderGraphMaterialsUpdater.cs
adrien-de-tocqueville: we do need it to reload materials in memory. There are a few things that can trigger a reset so maybe it was one of them, like if you have the material inspector openned.
I tested to disable the code and the material don't get reset

PR #4518, opened 2021-05-12
[2021.1] Misc SDET backports

theopnv:
### Purpose of this PR

- Backported #4517 - Bump gfx test framework to 7.8.11-preview for all test projects
- Backported #4497 - Use cache version of linux-standalone (HDRP) in the ABV/nightly

---
### Testing status
PR jobs


sophiaaar (approved):

theo-pnv:
The remaining pipelines are Android only. I'm canceling them and merging to save resources for those who need them before cut-off, as the changes of this PR would have broken other platforms as well if they were to break something.

PR #4520, opened 2021-05-12
Make the probe reference volume transform irrelevant

FrancescoC-unity:
As discussed, as a partially temporary (rotation and scale will likely be remain fixed) solution, we are fixing the transform so the reference volume is not rotated and always at origin for now. 

After the whole subdivision change goes in we might need to revisit the translation and likely put the origin at the same point as the center of the global bounds. For now let's fix at origin as any other position is broken anyway. 

By doing this, the transform of the reference volume transform is not going to impact anything and the relative warnings have been removed. 
The change can lead to further simplifications, but before doing this I want to make sure we can do the translation part correctly.


JulienIgnace-Unity (approved):

alelievr (approved):

alelievr:
Nice! less things to test :)

sebastienlagarde (approved):

PR #4522, opened 2021-05-12
Edited and restructured the Decal docs.

oleks-k:
Edited and restructured the Decal docs.


com.unity.render-pipelines.universal/Documentation~/renderer-feature-decal.md
```diff
@@ -1,51 +1,132 @@
...
 
-|__Propery__           | __Description__ |
-| -------------------- |---------------- |
-| __Surface Data__     | Allows specifying which decals surface data should be blended with surfaces. |
-| _Albedo_             | Decals will affect only base color and emission.|
-| _Albedo Normal_      | Decals will affect only base color, normal and emission|
-| _Albedo Normal MAOS_ | Decals will affect only base color, normal, metallic, ambient occlusion, smoothness and emission.|
+**Limitations:**
 
+* This technique requires the DepthNormal prepass, which does not work with tile-based rendering.
```
lukaschod, line 59: I feel like this sentance refers that technique does no work at all with tile based render. It does work, it just not effective as DepthNormal prepass in tile-based rendering only adds cost (Where in dekstop it can actually increase performance by a bit)

lukaschod (approved):
Really nice changes!!!

erikabar (approved):

com.unity.render-pipelines.universal/Documentation~/renderer-feature-decal.md
oleks-k: Thanks!
How about this:
***
* This technique requires the DepthNormal prepass, which makes the technique less efficient on GPUs that implement tile-based rendering.
***
And if this version is correct, can we add a short sentence about why it's less efficient (or just inefficient)?

lukaschod: Yeap looks good, I think this info should be in depth prepass docs if we have them

PR #4525, opened 2021-05-13
[HDRP][Path Tracing] AxF material support

eturquin:
Added *almost* exhaustive support for AxF material, both the SVBRDF and CarPaint variants. Match should be great with raster version (see screenshots below). This PR also brings few minor refactors/cosmetic changes in the other path traced materials.

Before: AxF materials would not be supported in path tracing, and would render black.
Now: AxF materials produce the expected results in path tracing.

This has been tested on AxF sample materials provided by @remi-chapelain (who himself got them from @slunity), on Unity balls. Would be worth testing it on other scenes/materials.

Known limitations:
- no texture filtering, which is visible for instance on the flakes in the car paint material
- no support for the useRefraction option, in SVBRDF and car paint

**Note:** testing this PR will require the AxF importer package


eturquin:
rasterized

eturquin:
path_traced
Path traced version

sebastienlagarde (approved):

PR #4532, opened 2021-05-13
Fixed bug where CopyDepthPass was always enqueued when deferred rende…

kaychang-unity:
---
### Purpose of this PR
An user reported that CopyDepthPass was always enabled when deferred rendering mode was enabled, while the depth texture copy would not always be used later in the pipeline. This PR fixes this bug and only enqueue CopyDepthPass when the UniversalRenderPipelineAsset requires depthTexture output.

---
### Testing status
Ran local project and confirmed CopyDepthPass was only enqueued when necessary.

---
### Comments to reviewers
I did not modify the changelog because deferred rendering mode is only officially enabled in the coming SRP package.
I did not add backport labels for same reason.


jonuuukas (approved):

kaychang-unity:
ABV is passing (same failures as master branch).

hdb-unity (approved):
As discussed, dev test coverage - manual and automated - seems sufficient for this change.

phi-lira (approved):

phi-lira (commented):
Missing changelog.

PR #3051, opened 2021-01-11
Removed LWRP

phi-lira:
### Purpose of this PR
 Removed LWRP dependency from URP package. This means that users can't upgrade from LWRP to URP 2021.2.
 They must first upgrade from LWRP to 11.x.x, then from 11.x.x to 2021.2
 
 This information was added to changelogs and upgrade guide.

---
### Testing status
 Tested locally

---
### Comments to reviewers
 The impact of this PR is that users cannot upgrade from URP to 2021.1 anymore. They will have to go to LWRP -> 11.x.x or earlier -> 2021.2 URP.

PR #4533, opened 2021-05-13
Fix sceneview draw modes for baked GI

nerites:

### Checklist for PR maker
- [N/A] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Fix for underlying issue behind https://fogbugz.unity3d.com/f/cases/1211436/
Although both SRPs have provided meta passes for their shaders, they weren't being picked up or used to render the sceneview draw modes. Instead, rendering was falling back to a built-in shader that did not work for instanced terrain. 

This PR adds a common MetaPass.hlsl file used for both Universal and HD sceneview draw mode rendering. Since Builtin's PBR albedo validation and PBR specular validation modes are unsupported in SRP, this file omitted the code used only for those modes. 

Realtime GI behavior is unreliable, but matches builtin behavior (at least with tip of trunk), so I'm going ahead with just the baked GI fixes for now.

I will be making additional changes to shadergraph as well, but I'll have that as a separate PR unless anyone objects.

---
### Testing status
Manually checked behavior against built-in (tip of trunk) for baked GI draw modes using progressive CPU.

Known inconsistencies:

- Terrain doesn't render in the baked lightmap culling mode for URP and built-in.  -> Fixed in URP by e7928d7
- Emissives don't render properly for HDRP. However this may be an issue with my scene setup, since emission doesn't appear to have an effect during normal rendering, either.
- The shadowmask debug mode is disabled for HDRP in the menu. Can someone from HDRP let me know if this is intentional?
- Shadergraphs may not behave as expected with this functionality, since I'm not sure if they add the meta pass with the appropriate. Will seek feedback about how to make the proper changes to enable this for shadergraph

Current status image:
image

---
### Comments to reviewers
@sebastienlagarde and @phi-lira can you please assign this PR to someone on your team with bandwidth?


iM0ve (approved):
I have checked on HDRP the baked view is fixed.
Note that other views like Texel Validity, Albedo, Emissive will require a separate fix


Baked view before the PR in the template
Unity_STcME5fRVq


Baked view after the PR in the template
Unity_BA2MFYJp94

sebastienlagarde:
>Emissives don't render properly for HDRP. However this may be an issue with my scene setup, since emission doesn't appear to have an effect during normal rendering, either.

Maybe your emissive intensity is 0? We have graphics test for all kind of emissive both bake and no bake, so we know it works. I expect you have an intensity of zero, this can be checked in our render pipelin debug windows with lighting mode -> Emissive lighting only.

>The shadowmask debug mode is disabled for HDRP in the menu. Can someone from HDRP let me know if this is intentional?

Shadowmask is disabled by default so I suppose you just forget to enable it?

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/ShaderPass/AxFSharePass.hlsl
```diff
@@ -17,7 +17,7 @@
 #if defined(_REQUIRE_UV2) || defined(_REQUIRE_UV3) || defined(DYNAMICLIGHTMAP_ON) || defined(DEBUG_DISPLAY) || (SHADERPASS == SHADERPASS_LIGHT_TRANSPORT)
 #define ATTRIBUTES_NEED_TEXCOORD2
 #endif
-#if defined(_REQUIRE_UV3) || defined(DEBUG_DISPLAY)
+#if defined(_REQUIRE_UV3) || defined(DEBUG_DISPLAY) || ((SHADERPASS == SHADERPASS_LIGHT_TRANSPORT) && defined(EDITOR_VISUALIZATION))
```
sebastienlagarde, line 20: I don't see any UV3 in the meta pass, are you sure you need this?  note that the meta pass used uv0, uv1 and uv2 but no uv3

com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.hlsl
```diff
@@ -55,7 +55,11 @@ LightTransportData GetLightTransportData(SurfaceData surfaceData, BuiltinData bu
 {
     LightTransportData lightTransportData;
 
+#if (SHADERPASS == SHADERPASS_LIGHT_TRANSPORT)
+    lightTransportData.diffuseColor = surfaceData.color;
```
sebastienlagarde, line 59: I suppose it is to match builtin behavior?   We chose to not handle color in HDRP as color don't have any light value. HDRP rely on physical light value and thus, only Emissive color have it, we should avoid to have regular color in it. (Also I expect it may break various project)

sebastienlagarde (commented):
Thanks to look at this issue and for the various change. few questoin on my side.

com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl
```diff
@@ -1,70 +1,85 @@
...
+}
+
+Varyings LightweightVertexMeta(Attributes input)
+{
+    Varyings ret;
+    MetaVaryings mRet = UniversalVertexMeta(input);
+    ret.positionCS = mRet.positionCS;
+    ret.uv = mRet.uv;
+    return ret;
+}
```
phi-lira, line 84: This is probably merge conflict? It should be removed as we removed LWRP upgrade tooling here: https://github.com/Unity-Technologies/Graphics/pull/3051/files#diff-42c6b6ac3d960a42860bdc7f7b585ebe5c4f1f2580fa2fde5e805d33e05f92e3L49-L59

com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl
```diff
@@ -1,70 +1,85 @@
...
+struct Varyings
 {
-    half3 Albedo;
-    half3 Emission;
-    half3 SpecularColor;
+    float4 positionCS   : SV_POSITION;
+    float2 uv           : TEXCOORD0;
 };
 
-float4 MetaVertexPosition(float4 positionOS, float2 uv1, float2 uv2, float4 uv1ST, float4 uv2ST)
```
phi-lira, line 28: There might be users using this, you could still define this and redirect to UnityMetaVertexPosition for backward compatibility.

com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl
```diff
@@ -1,70 +1,85 @@
...
+{
+    float4 positionCS   : SV_POSITION;
+    float2 uv           : TEXCOORD0;
+#ifdef EDITOR_VISUALIZATION
+    float2 VizUV        : TEXCOORD1;
+    float4 LightCoord   : TEXCOORD2;
+#endif
+};
 
-struct MetaInput
```
phi-lira, line 21: There could be users declaring this in their custom shaders, you could forward declare this to UnityMetaInput for backward compatibility / avoid breaking changes.

com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl
```diff
@@ -1,70 +1,85 @@
...
-        positionOS.z = positionOS.z > 0 ? REAL_MIN : 0.0f;
+        ret.VizUV = uv1 * unity_LightmapST.xy + unity_LightmapST.zw;
+        ret.LightCoord = mul(unity_EditorViz_WorldToLight, float4(TransformObjectToWorld(positionOS), 1));
     }
-    return TransformWorldToHClip(positionOS.xyz);
+#endif
+    return ret;
 }
 
-half4 MetaFragment(MetaInput input)
```
phi-lira, line 47: Same, for backward compatibility forward declare this to `UnityMetaFragment`

com.unity.render-pipelines.universal/Shaders/BakedLitMetaPass.hlsl
```diff
@@ -1,41 +1,12 @@
 #ifndef UNIVERSAL_BAKEDLIT_META_PASS_INCLUDED
 #define UNIVERSAL_BAKEDLIT_META_PASS_INCLUDED
 
-#include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl"
+#include "Packages/com.unity.render-pipelines.universal/Shaders/UnlitMetaPass.hlsl"
 
-struct Attributes
+//LWRP -> Universal Backwards Compatibility
+half4 LightweightFragmentMetaBakedLit(Varyings input) : SV_Target
```
phi-lira, line 7: Also probably merge conflict? was removed here: https://github.com/Unity-Technologies/Graphics/pull/3051/files#diff-9faa832cd3ae7ccf303e2c5580cba1ccaad39c4dbaac548dfe802d33a985b4a5L41-L50

com.unity.render-pipelines.universal/Shaders/BakedLit.shader
```diff
@@ -223,7 +223,7 @@ Shader "Universal Render Pipeline/Baked Lit"
             #pragma target 4.5
 
             #pragma vertex UniversalVertexMeta
-            #pragma fragment UniversalFragmentMetaBakedLit
+            #pragma fragment UniversalFragmentMetaUnlit
```
phi-lira, line 226: Not sure why these are changing?

com.unity.render-pipelines.universal/Shaders/Lit.shader
```diff
@@ -332,8 +332,9 @@ Shader "Universal Render Pipeline/Lit"
             #pragma target 4.5
 
             #pragma vertex UniversalVertexMeta
-            #pragma fragment UniversalFragmentMeta
+            #pragma fragment UniversalFragmentMetaLit
 
+            #pragma shader_feature EDITOR_VISUALIZATION
```
phi-lira, line 337: This is only needed in editor, how is this keyword stripped?

com.unity.render-pipelines.universal/Shaders/LitMetaPass.hlsl
```diff
@@ -3,46 +3,24 @@
...
     metaInput.Albedo = brdfData.diffuse + brdfData.specular * brdfData.roughness * 0.5;
     metaInput.SpecularColor = surfaceData.specular;
     metaInput.Emission = surfaceData.emission;
+    return UniversalFragmentMeta(input, metaInput);
+}
 
-    return MetaFragment(metaInput);
+half4 LightweightFragmentMeta(Varyings input) : SV_Target
+{
+    return UniversalFragmentMetaLit(VaryingsToMetaVaryings(input));
```
phi-lira, line 23: Should be removed

com.unity.render-pipelines.universal/Shaders/LitMetaPass.hlsl
```diff
@@ -3,46 +3,24 @@
...
-Varyings UniversalVertexMeta(Attributes input)
-{
-    Varyings output;
-    output.positionCS = MetaVertexPosition(input.positionOS, input.uv1, input.uv2, unity_LightmapST, unity_DynamicLightmapST);
-    output.uv = TRANSFORM_TEX(input.uv0, _BaseMap);
-    return output;
-}
-
-half4 UniversalFragmentMeta(Varyings input) : SV_Target
+half4 UniversalFragmentMetaLit(MetaVaryings input) : SV_Target
```
phi-lira, line 6: Not sure I follow why renaming this and why needed to rename `Varyings` to `MetaVaryings`.

com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl
```diff
@@ -1,70 +1,85 @@
...
+    float4 positionCS   : SV_POSITION;
+    float2 uv           : TEXCOORD0;
+#ifdef EDITOR_VISUALIZATION
+    float2 VizUV        : TEXCOORD1;
+    float4 LightCoord   : TEXCOORD2;
+#endif
+};
 
-struct MetaInput
+struct Varyings
```
phi-lira, line 28: I'm not sure why the rename to MetaVaryings is needed and I worry that now user could be using this as thsi is part of `ShaderLibrary` and will have compiler errors. At the bare least you could forward declare the old to the new MetaVaryings if needed.

com.unity.render-pipelines.universal/Shaders/SimpleLitMetaPass.hlsl
```diff
@@ -3,42 +3,22 @@
...
 
-    return MetaFragment(metaInput);
+    return UniversalFragmentMeta(input, metaInput);
+}
+
+//LWRP -> Universal Backwards Compatibility
+
+half4 LightweightFragmentMetaSimple(Varyings input) : SV_Target
+{
+    return UniversalFragmentMetaSimple(VaryingsToMetaVaryings(input));
```
phi-lira, line 21: Should be removed.

com.unity.render-pipelines.universal/Shaders/SimpleLitMetaPass.hlsl
```diff
@@ -3,42 +3,22 @@
...
-{
-    Varyings output;
-    output.positionCS = MetaVertexPosition(input.positionOS, input.uv1, input.uv2,
-        unity_LightmapST, unity_DynamicLightmapST);
-    output.uv = TRANSFORM_TEX(input.uv0, _BaseMap);
-    return output;
-}
-
-half4 UniversalFragmentMetaSimple(Varyings input) : SV_Target
+half4 UniversalFragmentMetaSimple(MetaVaryings input) : SV_Target
```
phi-lira, line 6: Same here, not sure why renamed.

com.unity.render-pipelines.universal/Shaders/UnlitMetaPass.hlsl
```diff
@@ -3,39 +3,17 @@
...
-    Varyings output;
-    output.positionCS = MetaVertexPosition(input.positionOS, input.uv1, input.uv2,
-        unity_LightmapST, unity_DynamicLightmapST);
-    output.uv = TRANSFORM_TEX(input.uv0, _BaseMap);
-    return output;
+    return UniversalFragmentMeta(input, metaInput);
 }
 
-half4 UniversalFragmentMetaUnlit(Varyings input) : SV_Target
+half4 LightweightFragmentMetaUnlit(Varyings input) : SV_Target
```
phi-lira, line 14: Should be removed.

phi-lira (changes requested):

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/ShaderPass/AxFSharePass.hlsl
nerites: The meta pass needs VARYINGS_NEED_TEXCOORD3 defined. So I guess I should move this down below as follows:


#ifdef ATTRIBUTES_NEED_TEXCOORD3  || ((SHADERPASS == SHADERPASS_LIGHT_TRANSPORT) && defined(EDITOR_VISUALIZATION))
#define VARYINGS_NEED_TEXCOORD3
#endif

com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.hlsl
nerites: Right, in builtin, unlit shaders still render out the diffuse color. If HDRP intends to have different behavior for this view, then I'm ok to revert this.

com.unity.render-pipelines.universal/Shaders/BakedLit.shader
nerites: Since BakedLit and MetaUnlit did the same thing in their fragment shader, I had BakedLit reuse the Unlit fragment shader. Would you prefer I keep BakedLit separate?

com.unity.render-pipelines.universal/Shaders/Lit.shader
nerites: How can I make sure it gets stripped from builds?

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/ShaderPass/AxFSharePass.hlsl
nerites: Ah, VertMesh requires that uv3 exists if texCoord3 is used. So I will leave it as is if that's okay.

nerites:
> 
> 
> > Emissives don't render properly for HDRP. However this may be an issue with my scene setup, since emission doesn't appear to have an effect during normal rendering, either.
> 
> Maybe your emissive intensity is 0? We have graphics test for all kind of emissive both bake and no bake, so we know it works. I expect you have an intensity of zero, this can be checked in our render pipelin debug windows with lighting mode -> Emissive lighting only.
> 
> > The shadowmask debug mode is disabled for HDRP in the menu. Can someone from HDRP let me know if this is intentional?
> 
> Shadowmask is disabled by default so I suppose you just forget to enable it?

Ah, thank you! When I enabled shadowmask, the debug view works correctly in HDRP:
image

I'm having a bit more trouble with emissives though. It looks like different shaders handle emission differently, but I enabled emission maps wherever I could, and chose a default checkerboard texture to use as the emission map. All emission colors are non-black, and all emission weights are 1. 

When I render in HDRP's emissive debug mode, only the unlits are rendered in color (even the Unlit that has emission set to black). Everything else has no emission. 
Do HDRP's debug modes also rely on the meta pass? Is it possible I've unintentionally modified HDRP's debug views with my changes?

com.unity.render-pipelines.universal/Shaders/UnlitMetaPass.hlsl
nerites: Fixed in 159a6ef

com.unity.render-pipelines.universal/Shaders/SimpleLitMetaPass.hlsl
nerites: Fixed in 159a6ef

com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl
nerites: Fixed in 5b3637f

com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl
nerites: Fixed in 5b3637f

com.unity.render-pipelines.universal/ShaderLibrary/MetaInput.hlsl
nerites: Fixed in 5b3637f

nerites:
@phi-lira I ended up splitting the attributes/varyings declaration into a separate UniversalMetaPass.hlsl file, so that MetaInput.hlsl can be included in Shadergraph with conflicting with Shadergraph's Attributes/Varyings declarations. 

It's not great for backwards compatibility since it moves the Attributes and Varyings declarations into a new file, but the split seems in line with other *Input.hlsl/*Pass.hlsl content splits. What do you think?

6688d1f

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/ShaderPass/AxFSharePass.hlsl
sebastienlagarde: can you point me where UV3 is used in meta pass, I haven't noticed it? thanks

com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.hlsl
sebastienlagarde: yes we should revert this one

sebastienlagarde:
> Do HDRP's debug modes also rely on the meta pass? Is it possible I've unintentionally modified HDRP's debug views with my changes?

nope, our debug mode are independent from Meta pass.

nerites:
> 
> 
> > Do HDRP's debug modes also rely on the meta pass? Is it possible I've unintentionally modified HDRP's debug views with my changes?
> 
> nope, our debug mode are independent from Meta pass.
> 
> > It looks like different shaders handle emission differently, but I enabled emission maps wherever I could, and chose a default >checkerboard texture to use as the emission map. All emission colors are non-black, and all emission weights are 1.
> 
> Maybe it is due to exposure? In HDRP emissive is lighting and behave like a light, maybe you have different exposure setup? Also we ahve an option to have emissive as pre-expose or not.

Oh yeah, it looks like it was due to exposure- when I zoomed in close, I was able to see the emissive color I expected. I could see it even better when I disabled the exposure settings on my sky and fog volume.

> I don't get it, you mean that a lit material with emissive don't appear and that an unlit with a black emissive is visible?

I meant that both looked the same. Though with exposure disabled I see the emissive cube is brighter. The unlit cube with black emissive is still rendered in the debug mode though, is this intended?
image

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/ShaderPass/AxFSharePass.hlsl
nerites: Line 65 of ShaderPassLightTransport.hlsl

https://github.com/Unity-Technologies/Graphics/pull/4533/files#diff-8a05acf5191b37d00df7c8de3aa78907221288291e6062afc6c3bd71b7905876

nerites: Hmm... I think I'm misunderstanding something about how the v2f struct is constructed. I assumed that each texcoord is a float2, but now I'm getting errors in HDRP shadergraphs because the texcoords seem to be float4s. If the texcoords can be float4s, then I would only need up to texcoord2.

nerites:
Fixed metapass hookup in HDTarget for shadergraph!
Verified that emissive matches expectations from HDRP Debug view: 
image

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/ShaderPass/AxFSharePass.hlsl
nerites: ah, just assigning to .xy only fixes it. :) Can you check whether the usage of texcoords in ShaderPassLightTransport.hlsl is correct in latest?

com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.hlsl
nerites: Done in fe87b58

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/ShaderPass/AxFSharePass.hlsl
sebastienlagarde: ah right, so indeed we will keep the support of uv3 because it can be used in shader graph (i.E an artists could create a graph with fetching albedo with uv3 ). So indeed my bad, I have forge't about this :)

sebastienlagarde: However you don't need to add the  ((SHADERPASS == SHADERPASS_LIGHT_TRANSPORT) as it is optional. it is IF the material use it then we will need to define it, so you should still remove this code

nerites: Since the meta pass needs texcoord3, these lines in VertMesh.hlsl cause an error if uv3 does not also exist:

#if defined(VARYINGS_NEED_TEXCOORD3) || defined(VARYINGS_DS_NEED_TEXCOORD3)
    output.texCoord3 = input.uv3;
#endif

nerites:
Holes don't get rendered in the sceneview debug modes in SRP, though they do get rendered for builtin. I'm still trying to get to the bottom of why this is happening, but renderdoc shows no holes texture is bound when rendering the meta passes in URP. I think this is to do with how it's not executing the SRP, but also not exactly executing the builtin render pipeline either... I'll do some more investigation tomorrow, but maybe this isn't worth fixing?

phocodes (approved):

erikabar (approved):
Baked Lightmap view is fixed on URP

com.unity.render-pipelines.universal/Shaders/Lit.shader
phi-lira: I'm not sure what's the behaviour regarding shader_feature that's not references in any material, I guess it's stripped automatically. @erikabar maybe you can confirm this.

phi-lira (approved):

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDTarget.cs
```diff
@@ -595,12 +595,17 @@ static class CoreRequiredFields
...
             HDStructFields.AttributesMesh.uv0,
             HDStructFields.AttributesMesh.uv1,
-            HDStructFields.AttributesMesh.color,
             HDStructFields.AttributesMesh.uv2,
+            HDStructFields.AttributesMesh.uv3,
+            HDStructFields.FragInputs.positionRWS,
+            HDStructFields.FragInputs.texCoord0,
+            HDStructFields.FragInputs.texCoord1,
+            HDStructFields.FragInputs.texCoord2,
+            HDStructFields.FragInputs.texCoord3,
```
nerites, line 608: can remove texcoords since they will be added when corresponding uvs are used

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/ShaderPass/LitSharePass.hlsl
```diff
@@ -14,7 +14,7 @@
 #if defined(_REQUIRE_UV2) || defined(_REQUIRE_UV3) || defined(DYNAMICLIGHTMAP_ON) || defined(DEBUG_DISPLAY) || (SHADERPASS == SHADERPASS_LIGHT_TRANSPORT)
 #define ATTRIBUTES_NEED_TEXCOORD2
 #endif
-#if defined(_REQUIRE_UV3) || defined(DEBUG_DISPLAY)
+#if defined(_REQUIRE_UV3) || defined(DEBUG_DISPLAY) || ((SHADERPASS == SHADERPASS_LIGHT_TRANSPORT) && defined(EDITOR_VISUALIZATION))
```
nerites, line 17: convert all ((SHADERPASS == SHADERPASS_LIGHT_TRANSPORT) && defined(EDITOR_VISUALIZATION)) -> defined(EDITOR_VISUALIZATION))

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDTarget.cs
nerites: If I remove these, the incorrect UVs are used to sample the debug texture, so I am leaving them in:
image

nerites:
In testing shadergraphs, I've found that the hair, cotton, silk, and eye shadergraphs do not render in meta passes. However, Arnold, 3DS, and Autodesk do. Do you know of a difference between these shadergraphs that could be causing this @sebastienlagarde?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/ShaderPass/LitSharePass.hlsl
nerites: Done in 8611bcaa08

sebastienlagarde (approved):

nerites:
Turns out I just forgot to enable GI on the MeshRenderers for those cubes, my bad! I did have to make a small change to get the hair SG working though. 
image

nerites:
It looks like in URP the meta passes for ComplexLit, SimpleLit, BakedLit, and Unlit aren't getting picked up for some reason (they're rendered with the fallback GIDebug shader). I need some more information on shader keywords in order to get to the bottom of this. For now, I don't think it makes sense to block this PR on that, since other URP shaders are working (including shadergraphs), and the fallback shader should do just fine for non-instanced objects. 

If I get QA approval before I figure this out, I'll file a bug for the issue.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassLightTransport.hlsl
```diff
@@ -99,18 +101,19 @@ float4 Frag(PackedVaryingsToPS packedInput) : SV_Target
...
-
-    if (unity_MetaFragmentControl.y)
-    {
-        // emissive use HDR format
-        res.rgb = lightTransportData.emissiveColor;
-    }
+    UnityMetaInput metaInput;
+    metaInput.Albedo = lightTransportData.diffuseColor.rgb;
+    metaInput.Emission = lightTransportData.emissiveColor;
+#if defined(UNLIT_CS_HLSL) || defined(HAIR_CS_HLSL)
```
sebastienlagarde, line 107: oh sorry I missed this change in my first review. It is forbidden to access surfaceData in any of ShaderPassXXX.hlsl  it will not compile. Each Material can implement what they want.

What you need is to move the code sampling the Material here to each of the function GetLightTransportData() in Material it has been implemented for this.

What you need to do is to add SpecularColor in LightTransportData struct and move the code 
metaInput.SpecularColor = surfaceData.specularColor.rgb in each of the Material (each material have a different way to define spec color so you need to adapt for each material (you can check of the debug mode to output spec color convert the property for each material).

Then in this current files you will have only this:
UnityMetaInput metaInput;
    metaInput.Albedo = lightTransportData.diffuseColor.rgb;
    metaInput.Emission = lightTransportData.emissiveColor;
    metaInput.SpecularColor = lightTransportData.specularColor.rgb;
#endif
#ifdef EDITOR_VISUALIZATION
    metaInput.VizUV = input.texCoord1.xy;
    metaInput.LightCoord = float4(input.texCoord2.xy, input.texCoord3.xy);
#endif

nerites:
Since dynamic GI has been fixed for HDRP in trunk, I was able to confirm that dynamic GI debug views also work, except for the emissive view. I still had to regenerate lighting a few times to get the other views working, and still saw the "failed to generate texture" message, which I will report to them.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassLightTransport.hlsl
nerites: Ah actuaully, this SpecularColor was only used in validation debug views that are unsupported by SRPs (likely because their lighting models do not match builtin lighting models). So I just removed SpecularColor from MetaInput in fac9bab

nerites:
Building URP_Terrain is consistently crashing. Attempting to build locally to repro.

com.unity.render-pipelines.universal/Shaders/Lit.shader
nerites: It looks like you have to explicitly strip it. Kuba added it to his PR for stripping lighting-related keywords here: https://ono.unity3d.com/unity/unity/pull-request/127683/_/lighting/bugfix/1307896-dirlightmap-stripping

sebastienlagarde:
Moving forward. Yamato is ok despite all the stuff broken, it come from our infrastructure and master

PR #4552, opened 2021-05-17
[2021.2] Removed _SHADOWS_SOFT variant

nigeljw-unity:
The soft shadows variant is only impacting a single predicate check. The variant was removed in favor of branching on a constant.

### Checklist for PR maker

---
### Purpose of this PR
To diffuse the combinatorial explosion of variants which will decrease build times.

---
### Testing status
Tested against default URP scene.

---

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
+++ b/com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
@@ -1751,7 +1751,6 @@ namespace UnityEngine.Rendering.Universal.Internal
 
                 int shadowLightIndex = m_AdditionalLightsShadowCasterPass != null ? m_AdditionalLightsShadowCasterPass.GetShadowLightIndexFromLightIndex(visLightIndex) : -1;
                 bool hasDeferredLightShadows = vl.light && vl.light.shadows != LightShadows.None && shadowLightIndex >= 0;
-                bool hasSoftShadow = hasDeferredLightShadows && renderingData.shadowData.supportsSoftShadows && vl.light.shadows == LightShadows.Soft;
 
                 CoreUtils.SetKeyword(cmd, ShaderKeywordStrings._DEFERRED_LIGHT_SHADOWS, hasDeferredLightShadows);
 


com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -1636,7 +1636,6 @@ void RenderStencilDirectionalLights(CommandBuffer cmd, ref RenderingData renderi
                 }
 
                 bool hasSoftShadow = hasDeferredShadows && renderingData.shadowData.supportsSoftShadows && vl.light.shadows == LightShadows.Soft;
```
ellioman, line 1638: Can't we delete this bool as well?

nigeljw-unity: Done!

ellioman (approved):

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -68,9 +68,7 @@ internal class ShaderPreprocessor : IPreprocessShaders
         ShaderKeyword m_ReflectionProbeBlending = new ShaderKeyword(ShaderKeywordStrings.ReflectionProbeBlending);
         ShaderKeyword m_ReflectionProbeBoxProjection = new ShaderKeyword(ShaderKeywordStrings.ReflectionProbeBoxProjection);
         ShaderKeyword m_DeferredLightShadows = new ShaderKeyword(ShaderKeywordStrings._DEFERRED_LIGHT_SHADOWS);
-        ShaderKeyword m_DeferredSoftShadows = new ShaderKeyword(ShaderKeywordStrings._DEFERRED_SHADOWS_SOFT);
```
phi-lira, line 71: The keywords are being unified between deferred and forward.
https://github.com/Unity-Technologies/Graphics/pull/4535

phi-lira (approved):
LGTM. Should we add a rule to stripper to always strip _SHADOW_SOFT keyword in case users have custom shaders using it?

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
nigeljw-unity: Hey @kaychang-unity, we should really just delete all as done in this PR rather than unified together.

kaychang-unity: @nigeljw-unity are you suggesting #4535 waits for yours to land, so I can delete _DEFERRED_SHADOWS_SOFT. Or should I land #4535 before yours, then you can remove the extra code on your side?

kaychang-unity (approved):

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
nigeljw-unity: I'm ok with either, but I think I would prefer this to land first to minimise code churn, and also to have this as a low complexity test for removing variants in favour of branches, since we can potentially regress on low tier platforms. This has already gone through performance testing and showed a huge build time improvement

erikabar (approved):
Tested and compared build times and CPU time on this PR, master, 12.0, 10.4, 7.5.2:
**Build times:** 
9e35952-a18 (PR)  ---------- 1102s 
5a87688-a18 (master) ------ 2830s
12.0-a18 --------------------- 763s
10.4 ------------------------- 1311s
7.5.2 ------------------------ 1540s

**CPU Time:**
9e35952-a18 (PR) ---------- 8.32ms
5a87688-a18 (master) ------ 8.34ms
12.0-a18 -------------------- 8.39ms
10.4 ----------------------- 30.99ms

Looks like there's no impact on CPU time after changes, build time is improved compared to master, 10.4 and 7.5.2 but is still worse than 12.0

erikabar:
added @ernestasKupciunas for mobile results

CertifiedRice (approved):

nigeljw-unity:
Remaining ABV failures are unrelated to this change: Backbuffer perspective difference and development build text plus the "some tests were inconclusive".

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
kaychang-unity: Ok, I'll wait your PR then.

ernestasKupciunas (approved):
PR looks good. I didn't notice any significant performance drop looking at the various counters between the two builds (before and after the changes).

Devices under test:
Google Pixel 5, Adreno 620
Samsung Galaxy S10+, 10.0.0, Exynos 9 9820, Mali-G76

Unity:
Version: 2021.2.0a19.2746
Revision: trunk c7b813f0d5d5
Built: Thu, 27 May 2021 19:32:01 GMT

Mali_GPU_Cycles

BvA

kaychang-unity:
@ernestasKupciunas Just curious, does MasterURP test scene has soft-shadows enabled or disabled?
GPU active -8.8% +- 4.3%.

ernestasKupciunas:
@kaychang-unity, Yes, Soft Shadows checkbox was enabled during the testing

nigeljw-unity:
@ernestasKupciunas can we also get a run with soft shadows disabled? The main intent is to verify that removing the variant does not reduce performance when soft shadows are disabled.

ernestasKupciunas:
New Data with the Soft Shadows disabled (sorry for not doing this in the first place 🤦 ):
Mali_GPU_Cycles
Pixel5MasterVSURP
Pixel4MasterVSPR


Devices under testing:
Google Pixel 5, Adreno 620
Google Pixel 4, Adreno 640
Samsung Galaxy S10+, 10.0.0, Exynos 9 9820, Mali-G76

Unity used for testing:
Version: 2021.2.0a19.2746
Revision: trunk c7b813f0d5d5
Built: Thu, 27 May 2021 19:32:01 GMT

nigeljw-unity:
@ernestasKupciunas @erikabar can we also get the delta between shader binary compilation size?

Can we also get a perf delta test for iOs devices?

ernestasKupciunas:
The general performance on iOS looks very similar or slightly decreased, but not critical(unfortunately I have only 2 iOS devices):
https://docs.google.com/spreadsheets/d/1jFt93xNLIa7cZjVJ0LERE3j3SMrvnVsKwhx9CArLLgk/edit?usp=sharing

Shaders binary compilation sizes comparison:
BuildReportComparison

PR #4553, opened 2021-05-17
[URP] Add GetUniversalAdditionalLightData

svens-unity:
### Purpose of this PR
Universal Render Pipeline exposes additional light data in a separate component. This PR adds a method (`GetUniversalAdditionalLightData`) that returns the additional data component for the given light or create one if it doesn't exist yet. The behavior and usage of the extension method should be identical to `GetAdditionalCameraData`. This also allows us to get rid off all usages of `ComponentSingleton` in URP. 

---
### Testing status
[URP split pass. Issues observed on IOS.](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252FGetUniversalAdditionalLightData/.yamato%252Fall-universal_split.yml%2523PR_Universal_Split_trunk/6842090/job/full-pipeline)

---

### DIFF
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added URP Global Settings Asset to the Graphics Settings - a common place for project-wide URP settings.
 - Added possibility to rename light layer values.
 - Added Light cookies support to directional, point and spot light. Directional light cookie is main light only feature.
+- Added GetUniversalAdditionalLightData, a method that returns the additional data component for a given light or create one if it doesn't exist yet.
 - VFX: Basic support of Lit output.
 
 ### Changed


com.unity.render-pipelines.universal/Runtime/UniversalAdditionalLightData.cs
```diff
@@ -27,6 +27,29 @@ public enum LightLayerEnum
...
     }
 
+    /// <summary>
+    /// Contains extension methods for Light class.
+    /// </summary>
+    public static class LightExtensions
+    {
+        /// <summary>
+        /// Universal Render Pipeline exposes additional light data in a separate component.
+        /// This method returns the additional data component for the given light or create one if it doesn't exists yet.
```
svens-unity, line 37:
```suggestion
        /// This method returns the additional data component for the given light or create one if it doesn't exist yet.
```

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -1174,7 +1174,7 @@ void SetupMainLightConstants(CommandBuffer cmd, ref LightData lightData)
             Vector4 lightPos, lightColor, lightAttenuation, lightSpotDir, lightOcclusionChannel;
             UniversalRenderPipeline.InitializeLightConstants_Common(lightData.visibleLights, lightData.mainLightIndex, out lightPos, out lightColor, out lightAttenuation, out lightSpotDir, out lightOcclusionChannel);
 
-            UniversalAdditionalLightData additionalLightData = ForwardLights.GetAdditionalLightData(lightData.visibleLights[lightData.mainLightIndex].light);
+            var additionalLightData = lightData.visibleLights[lightData.mainLightIndex].light.GetUniversalAdditionalLightData();
```
svens-unity, line 1177: I believe this is the only "unsafe" change out of the suggested changes. Before this code would pass if `light == null`. Now it won't.

ellioman: Should we then split this in to 2 or more lines to be sure we're not breaking anything?

`Light light = lightData.visibleLights[lightData.mainLightIndex].light
if (light == null)
    return;
`

phi-lira: Shouldn't the null check be performed in the extension method instead and make sure you return an instance to additional light data in case of null?

eh-unity: Maybe just get lights and these into an array (matching visible lights) in lightData once and reuse elsewhere. Cache them by hash of light/lightData or similar and only get them when necessary.

svens-unity: To clarify, I do not think `Light` would ever be null here. How can we have an index for our main light when the component itself is null.

I am pointing out that all other code would break before if `Light` _was_ null. This is the only place where before it would work if _Light is null_, after it will break with null reference exception. Taking Elvar's suggestion could work.
I did not add a null check to the extension method to keep behavior and error messages identical to the extension method for `GetUniversalAdditionalCameraData()`, we could add a check but we would need to do this to both places.

phi-lira: It would be nice to test this with particles lights for instance. Those will yield a null Light iirc.

svens-unity: I created a scene with a particle system that uses the lights module using deferred rendering in URP. I used a break point to make sure the code is ran, the component is not null in this case. 
image

We could change the code to this - just to be sure:
`            uint lightLayerMask = (uint)LightLayerEnum.LightLayerDefault;
            var light = lightData.visibleLights[lightData.mainLightIndex].light;
            if(light != null)
            {
                var additionalLightData = light.GetUniversalAdditionalLightData();
                lightLayerMask = (uint)additionalLightData.lightLayerMask;
            }`

phi-lira (approved):
LGTM. Seems no big issues with the light being null and I'm ok with the proposed approach.

kaychang-unity:
Test 029_Particles_Deferred in the UniversalGraphicsTest_Foundation project would also crash if null light reference was accessed (it's working fine).

kaychang-unity (approved):

eh-unity (approved):
Seems ok.

PR #4558, opened 2021-05-17
[10.x.x] Backport fix for shader graph sticky note copy pasting

joshua-davis:
Bugfix https://fogbugz.unity3d.com/f/cases/1336810
backport of https://github.com/Unity-Technologies/Graphics/pull/4395

### Checklist for PR maker

---
### Purpose of this PR
Backport of https://github.com/Unity-Technologies/Graphics/pull/4395
Sticky notes cannot be copied/cut/pasted right now.
Also fixed sticky notes not being able to be dragged into a group which would cause some odd behaviors since you could right-click to group a sticky note with other nodes, but you could never drag it in.

---
### Testing status

- Within a graph:
  - [x] Simply copy paste of a sticky note. Note's contents/title are the same and it's selected
  - [x] Copy two sticky notes. Both get pasted and selected
  - [x] Copy a sticky note and a node. Both are pasted and selected
  - [x] Cut/Paste works.
  - [x] Right click copy/paste/cut/duplicate works (menu option is in an odd location, see notes below)
  - [x] Copy of groups works (stickynote in a group, group with a sticky note, etc...)
  - [x] Copy a sticky note and a node and pan the graph view way to the side so the original location is very far away. Paste and verify that both nodes are within the current viewport (see comments for some details).
  - [x] Validate the size and theme is correct when copying sticky nodes (change to dark theme or change text size then copy/cut + paste)
- Do all of the above for:
  - [x] Between two different graphs
  - [x] From a graph to a sub-graph
  - [x] From a sub-graph to a graph
  - [x] Between two sub-graphs

- [x] Additionally verify you can drag sticky notes into a group.
 
---
### Comments to reviewers

There are a lot of cases with sticky notes, I hope I tried them all, but any additional testing is appreciated. 

Known issues:
- The copy/cut/duplicate commands on right-click are not in the same location as for normal nodes. This is controlled in GraphView so there's no way to fix this without changing the package.
- When notes/nodes are pasted they aren't always perfectly within the viewport, but they should be mostly inside (you can at least somewhat see them) as opposed to way off where they were copied. The core "framing" logic wasn't changed to be better, only to include sticky notes.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -12,6 +12,7 @@ The version number for this package has increased due to a version update of a r
 ### Fixed
 - Fixed an issue where an integer property would be exposed in the material inspector as a float [1332564]
 - Fixed an issue where the normal vector in object space would scale with the objects scale, causing non-normalized normal vectors
+- Fixed a bug in ShaderGraph where sticky notes couldn't be copied and pasted [1221042].
 
 ## [10.5.0] - 2021-04-19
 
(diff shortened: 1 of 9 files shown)


jessebarker (approved):
LGTM

ghost (approved):
developer testing is sufficient

PR #4573, opened 2021-05-18
[URP] Render Pass path store actions optimization

manuele-bonanno:
Fogbugz link:
https://fogbugz.unity3d.com/f/cases/1341452/

---
### Purpose of this PR
This PR introduces the same store actions optimization options introduced in [this PR](https://github.com/Unity-Technologies/Graphics/pull/4161) to the RenderPass path, to make sure we have performance parity between the RP and non-RP paths.

The main difference introduced is the concept of overriding Color/Depth store actions.
Since the RenderPass path merges render passes, it is important that the final store actions chosen for all the merged passes, matches the action requested **explicitly** by the last pass in the mergeable list.

All store actions are "Store" by default, the overridden flag helps to figure out which store actions are effectively requested by the user code and which ones were default store actions (which then are assumed safe to override in favour of the most optimal one)

---
### Testing status
Local, Yamato

---


manuele-bonanno:
the few test failures are not related to this PR and the same as in master nightly run

jonuuukas (approved):

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -135,6 +142,47 @@ internal void SetupNativeRenderPassFrameData(CameraData cameraData, bool isRende
             }
         }
 
+        internal void UpdateFinalStoreActions(int[] currentMergeablePasses, CameraData cameraData)
+        {
+            for (int i = 0; i < 8; ++i)
```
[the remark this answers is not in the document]
manuele-bonanno: I added this in a following PR, I think I could move that change here too and then merge later. WIll do

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -163,6 +163,16 @@ public RenderBufferStoreAction depthStoreAction
             get => m_DepthStoreAction;
         }
 
+        public bool[] overriddenColorStoreActions
+        {
+            get => m_OverriddenColorStoreActions;
+        }
+
+        public bool overriddenDepthStoreAction
```
phi-lira, line 171: I'm failing to understand why this public API is needed.

manuele-bonanno: the fact that it is public is an error, will change that to internal. 
Regarding why the overridden flag is used I moved the conversation to slack

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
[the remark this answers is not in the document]
manuele-bonanno: fixed in CS ad0a262

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
manuele-bonanno: changed to internal in CS 5254f56

phi-lira (approved):

ernestasKupciunas (approved):
Looks fine. I was using [Winter](https://github.com/ernestasKupciunas/Winter) and [PolyFarm](https://github.com/Unity-Technologies/PolygonFarmBenchmark) projects to test the visual correctness and Polyfarm to test the performance. 

Performance doc:
On the performance side, I was comparing RP version build vs non-RP build with different store action configurations..
https://docs.google.com/spreadsheets/d/1zggujz-shSxvaw7rSfLT1hYVz0ZGIGMDHMSyvGX_HhM/edit?usp=sharing

Devices under testing:

Unity used for testing:
Version: 2021.2.0a20.2866
Revision: trunk 0bd21e314fca
Built: Thu, 03 Jun 2021 18:52:14 GMT

simon-engelbrecht-soerensen (approved):
Tested on PC with no difference in performance or visuals.

manuele-bonanno:
after merging master into the branch seems like this is introducing a RenderPass test failure on Windows and Linux. All other test failures are not related. Investigation ongoing

manuele-bonanno:
the remaining test errors are unrelated to this PR (out of memory and provisioning file issues on test devices). The 111_CameraStackMSAA_RenderPass test failure was fixed

PR #4574, opened 2021-05-18
[12.x.x] Stripped unnecessary shaders from _Foundation project.

phi-lira:
### Purpose of this PR
 - Strips `LIGHTMAP_ON`, `DYNAMICLIGHTMAP_ON`, `DIRLIGHTMAP_COMBINED` keywords from _Foundation project. 
 - Strips post-processing shaders from _Foundation project. 

---
### Testing status

Validated using the ShaderVariantTool:
 
Before

After

---
### Comments to reviewers
 - The above lightmapping keywords are handled by Unity and configure in Graphics Settings strip settings. The project had set to auto strip unused lightmapping keywords, however some scenes in the project are setup in a way that it reference mixed lighting.  I changed the stripping setting to force stripping rather than fixing the multiple scenes. I decided to do like this as if I would just fix the scenes a regression could happen by someone adding a new test scene and forgetting to disable baked/mixed lighting. 
 
 - I also looked at all renderers and disabled post-processing as that's not the goal to be tested for this project. The post-processing stripping should be moved to Global Settings in a later PR as the workflow to disable post is not great and troublesome. 

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fshader-stripping-foundation/.yamato%252Fall-urp_foundation.yml%2523PR_URP_Foundation_trunk/6862558/job

### DIFF
--- a/TestProjects/UniversalGraphicsTest_Foundation/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/050_Shader_Graphs.png
+++ b/TestProjects/UniversalGraphicsTest_Foundation/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/050_Shader_Graphs.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5e2709bfaa485313837fbb5a47c439b9dad524af6ad0aa23b65d420a27a6177e
-size 144013
+oid sha256:9bb82cfc140e205d4ab6f6c782e6745d74f42cc82d6a90dd06b405da4bac1398
+size 122125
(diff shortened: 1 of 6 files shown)


ellioman (approved):

phi-lira:
Linux_Vulkan_Standalone failing due to Motion Vector. Being fixed in this PR https://github.com/Unity-Technologies/Graphics/pull/4574

phi-lira:
Updated Reference image for Motion Vector, as discussed with @HedvigAxelsson the reference image was wrong and that was necessary to fix iPhone.
 
 Updated Reference image for Android Vulkan and GLES3 to account for the difference in baked shadows / AO due to change in lighting to not user baked lighting. Same as other reference images. 
 
 It seems there's an issue in deferred + sky lighting in both iOS and Android (Vulkan) need to be investigated as it's not related to this PR.

PR #4534, opened 2021-05-14
[2020.3] [ShaderGraph] [10.x] Fix SamplerStates being used by SubGraphs (waiting for any 2021 fallout)

cdxntchou:
### Purpose of this PR
Fixing a number of issues with SamplerStates and SubGraphs.

https://fogbugz.unity3d.com/f/cases/1336089/
   Regression, where nested subgraphs that pass SamplerStates with the same default settings will cause compile failures

https://fogbugz.unity3d.com/f/cases/1336119/
   Subgraph SamplerState defaults are not respected

https://fogbugz.unity3d.com/f/cases/1336126/
   User can only edit SamplerState property display names when they are created.

2021.2: https://github.com/Unity-Technologies/Graphics/pull/4597
2021.1: https://github.com/Unity-Technologies/Graphics/pull/4580
2020.3: https://github.com/Unity-Technologies/Graphics/pull/4534

### Details:

The existing SamplerState properties would enforce a reference name override, forcing it to something like "SamplerState_Linear_Repeat".

This caused issues because SubGraph properties are treated as function input parameters, whereas ShaderGraph properties are global resources, and they could very easily have the same name, causing errors when we try to access the global resource within a subgraph with a conflicting input parameter.

This change differentiates the naming scheme of the function inputs from the global resources.  The function inputs are now named using the parameter naming scheme used by all other inputs:  basically type + GUID   i.e. "SamplerState_1340f13f".

The properties resources are now always named using the setting-based naming scheme, i.e.  "SamplerState_Linear_Repeat", and they are de-duplicated, so multiple instances of the same one are combined.

We change the UI so that the reference names show up as the property resource name ("SamplerState_Linear_Repeat") and is not editable.  We also now allow the SamplerState display names to be edited at any point, not just on creation.
image

Versioning is applied to the SamplerStateShaderProperties to remove the pre-existing reference name override ("SamplerState_Linear_Repeat"), causing them to fall back to their more unique default naming ("SamplerState_1340f13f").

Previously, subgraph samplerstates could specify defaults within the subgraph, however they were completely ignored -- the samplerstate slots created on a subgraph node would _always_ default to "SamplerState_Linear_Repeat".

Now it will correctly copy the default from the subgraph property.

Note that it does this in a "true inheritance" fashion -- changing the setting on the subgraph property will change the defaults on the slots in all of the corresponding subgraph nodes.

---
### Testing status

Added tests to check that the nested subgraph case continues to work, and also a new scene that checks expected SamplerState behavior.

[...]

PR #4580, opened 2021-05-18
[2021.1] [ShaderGraph] Fix SamplerStates being used by SubGraphs

cdxntchou:
### Purpose of this PR
Fixing a number of issues with SamplerStates and SubGraphs.

https://fogbugz.unity3d.com/f/cases/1336089/
   Regression, where nested subgraphs that pass SamplerStates with the same default settings will cause compile failures

https://fogbugz.unity3d.com/f/cases/1336119/
   Subgraph SamplerState defaults are not respected

https://fogbugz.unity3d.com/f/cases/1336126/
   User can only edit SamplerState property display names when they are created.

2021.2: https://github.com/Unity-Technologies/Graphics/pull/4597
2021.1: https://github.com/Unity-Technologies/Graphics/pull/4580
2020.3: https://github.com/Unity-Technologies/Graphics/pull/4534

### Details:

The existing SamplerState properties would enforce a reference name override, forcing it to something like "SamplerState_Linear_Repeat".

This caused issues because SubGraph properties are treated as function input parameters, whereas ShaderGraph properties are global resources, and they could very easily have the same name, causing errors when we try to access the global resource within a subgraph with a conflicting input parameter.

This change differentiates the naming scheme of the function inputs from the global resources.  The function inputs are now named using the parameter naming scheme used by all other inputs:  basically type + GUID   i.e. "SamplerState_1340f13f".

The properties resources are now always named using the setting-based naming scheme, i.e.  "SamplerState_Linear_Repeat", and they are de-duplicated, so multiple instances of the same one are combined.

We change the UI so that the reference names show up as the property resource name ("SamplerState_Linear_Repeat") and is not editable.  We also now allow the SamplerState display names to be edited at any point, not just on creation.
image

Versioning is applied to the SamplerStateShaderProperties to remove the pre-existing reference name override ("SamplerState_Linear_Repeat"), causing them to fall back to their more unique default naming ("SamplerState_1340f13f").

Previously, subgraph samplerstates could specify defaults within the subgraph, however they were completely ignored -- the samplerstate slots created on a subgraph node would _always_ default to "SamplerState_Linear_Repeat".

Now it will correctly copy the default from the subgraph property.

Note that it does this in a "true inheritance" fashion -- changing the setting on the subgraph property will change the defaults on the slots in all of the corresponding subgraph nodes.

---
### Testing status

Added tests to check that the nested subgraph case continues to work, and also a new scene that checks expected SamplerState behavior.

- [x] Tested against nested subgraph repro case for 1336089.
- [x] Added the repro case as a Test in the ShaderGraph tests.
- [x] Verified the added Test failed before this PR, and succeeds with this PR.
- [x] Added samplerstate graphics tests.
- [x] Verified the samplerstate graphics tests failed before this PR, and succeed now.
- [x] Tested that samplerstate display names can be renamed at any point, in either subgraphs or shadergraphs.
- [x] Tested that modifying display names is undo/redoable.
- [x] Verified that the samplerstate reference names only show up in ShaderGraphs, are not editable, and display the actual HLSL resource name.
- [x] Verified that existing shaders upgrade correctly, with samplerstates removing their override reference name, for both subgraphs and shadergraphs.
- [x] Verified that changing the samplerstate input settings on a subgraph result in the subgraph node defaults changing automatically, and the correct samplerstate value being used, both in loaded graph previews, and without any open graphs in scene view.
- [x] Verified that the samplerstate subgraph default inheritance works through multiple layers of subgraph
- [x] Verified that multiple SamplerStates with the same setting in a ShaderGraph will work, and are deduplicated.
- [x] Verified that multiple SamplerStates with the same setting across a number of subgraphs work, and are deduplicated.
- [x] Tested VFX shaders using SamplerStates continue to work (this even fixes the same bugs in VFX usage)
image

Yamato:

Nightly ShaderGraph: 🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/2021.1%252Fsg%252Ffix%252F1336089/.yamato%252Fall-shadergraph.yml%2523Nightly_ShaderGraph_2021.1/6867461/job/pipeline

URP PR Job: 🟢
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/2021.1%252Fsg%252Ffix%252F1336089/.yamato%252Fall-universal_split.yml%2523PR_Universal_Split_2021.1/6869698/job/pipeline

HDRP PR Job: 🟢
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/2021.1%252Fsg%252Ffix%252F1336089/.yamato%252Fall-hdrp.yml%2523PR_HDRP_2021.1/6869578/job/pipeline

### DIFF
--- a/.yamato/_green_job_revisions_2021.1.metafile
+++ b/.yamato/_green_job_revisions_2021.1.metafile
@@ -137,142 +137,142 @@ nightly_vfx_hdrp_-_2021_1:
       updated_at: '2020-12-16 00:05:36.049181'
       version: ''
 nightly_vfx_hdrp_-_2021.1:
-  updated_at: '2021-05-18 07:38:40.910165'
+  updated_at: '2021-05-25 07:11:29.627648'
   last_green_revisions:
     Android:
-      revision: 8605afcc71ad15eda94742b4de82ca8a4ce97908
-      updated_at: '2021-05-15 23:22:44.417497'
+      revision: 7a790e367ab34f384321451c38a63b40b6a0e167
+      updated_at: '2021-05-24 23:16:40.887093'
       version: ''
     Linux:
-      revision: 8605afcc71ad15eda94742b4de82ca8a4ce97908
-      updated_at: '2021-05-16 23:25:05.318622'
+      revision: 7a790e367ab34f384321451c38a63b40b6a0e167
+      updated_at: '2021-05-24 23:16:40.887093'
       version: ''
     OSX:
-      revision: 8605afcc71ad15eda94742b4de82ca8a4ce97908
-      updated_at: '2021-05-15 23:22:44.417497'
+      revision: 7a790e367ab34f384321451c38a63b40b6a0e167
+      updated_at: '2021-05-24 23:16:40.887093'
       version: ''
     Win:
-      revision: 8605afcc71ad15eda94742b4de82ca8a4ce97908
-      updated_at: '2021-05-15 23:22:44.417497'
+      revision: 7a790e367ab34f384321451c38a63b40b6a0e167
+      updated_at: '2021-05-24 23:16:40.887093'
       version: ''
     iPhone:
-      revision: 8605afcc71ad15eda94742b4de82ca8a4ce97908
-      updated_at: '2021-05-15 23:22:44.417497'
+      revision: 7a790e367ab34f384321451c38a63b40b6a0e167
+      updated_at: '2021-05-24 23:16:40.887093'
       version: ''
 vfx_hdrp_pr_job_-_2021.1:
-  updated_at: '2021-05-18 07:38:40.910165'
+  updated_at: '2021-05-25 07:11:29.627648'
   last_green_revisions:
     Android:
-      revision: 8605afcc71ad15eda94742b4de82ca8a4ce97908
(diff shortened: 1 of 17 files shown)


jessebarker (approved):
LGTM

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -307,6 +307,14 @@ public virtual void UpdateSlots()
                 // Copy defaults
                 switch (prop.concreteShaderValueType)
                 {
+                    case ConcreteSlotValueType.SamplerState:
+                        {
+                            var tSlot = slot as SamplerStateMaterialSlot;
```
jessebarker, line 312: I'm guessing this indentation is the source of the formatting issue (might need to pull this block left one stop).

LandonTownsendUnity (approved):
Developer testing is sufficient

LandonTownsendUnity:
Rerequesting review, would like to take a look at the original PR before approving the backports

LandonTownsendUnity (approved):
Reapproving since parent PR is approved

PR #4583, opened 2021-05-19
Enabled DXC for Metal API and URP stock shaders.

phi-lira:
### Purpose of this PR
 Enabled DXC compilation for metal on URP stock shaders. DXC compiles much faster than FXC.
 For now we can only enable on metal due to current limitations.
 
 I didn't enabled for the following shaders:
  - Particles, can't be enabled due to DXC and CB issues.
  - PostProcessing shaders except Uber. There's not much advantage to enable for those. We can enable later.
  - 2D shaders, want to test DXC stability first on 3D shaders.
  - ShaderGraph, there might be DXC limitations that we don't know of and cause regression/compilation issues in shader graph. 

### Testing status
  Tested locally.

---
### Comments to reviewers
   I had to fix Unlit shader vertex and fragment entry point as there was ambiguity with another function with same name. This caused an error in DXC.


com.unity.render-pipelines.universal/Shaders/Unlit.shader
```diff
@@ -55,8 +59,8 @@ Shader "Universal Render Pipeline/Unlit"
             #pragma multi_compile_fragment _ _DBUFFER_MRT1 _DBUFFER_MRT2 _DBUFFER_MRT3
             #pragma multi_compile _ DEBUG_DISPLAY
 
-            #pragma vertex UniversalVertexUnlit
-            #pragma fragment UniversalFragmentUnlit
+            #pragma vertex UnlitPassVertex
+            #pragma fragment UnlitPassFragment
```
phi-lira, line 63: This had to be renamed as there's a function in shader library with same name and that caused compiler error in DXC.

ellioman (approved):

CertifiedRice (approved):

erikabar (approved):
tested with a simple scene that covers all URP shaders. build time after the change is improved 2x: 
avg build time on master      612s
avg build time with PR          297s

PR #4584, opened 2021-05-19
Fix Regression: Gracefully handle when a filtered list of properties is passed into ShaderGraphPropertyDrawers.DrawShaderGraphGUI instead of throwing exception

hybridherbst:
### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`. _Not needed, fixes a regression_
- [ ] Have you updated the changelog? Each package has a `CHANGELOG.md` file. _Not needed, fixes a regression_

---
### Purpose of this PR
The new category workflow has introduced a regression: when passing a filtered list of properties (e.g. from a custom shader GUI) into one of the base property GUIs, an exception is thrown because some of the properties inside categories aren't found.
This PR fixes that by gracefully handling those missing properties (they're skipped) instead of throwing an exception and breaking custom shader GUIs this way.

---
### Testing status
Tested that custom shader GUIs that call into these with a filtered list work again.

### DIFF
--- a/com.unity.shadergraph/Editor/Drawing/MaterialEditor/ShaderGraphPropertyDrawers.cs
+++ b/com.unity.shadergraph/Editor/Drawing/MaterialEditor/ShaderGraphPropertyDrawers.cs
@@ -50,7 +50,7 @@ namespace UnityEditor.ShaderGraph.Drawing
                     return prop;
                 }
             }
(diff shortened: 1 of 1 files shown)


Nightmask3:
Closing this PR as the requested changes were merged by cherry-picking the commits from this branch.

PR #4585, opened 2021-05-19
[Fix] Ensuring SRP asset in Quality overrides Graphics' one

jenniferd-unity:
*Purpose of this PR*:

Our codebase needs to use the proper SRP reference. I have also cleaned up code in some small places.

---
### Testing status
- launched HDRP_Tests and URP_Lighting tests

This PR requires the following [one] (https://github.cds.internal.unity3d.com/unity/unity/pull/4258)
Fogbugz: [1335986](https://fogbugz.unity3d.com/f/cases/1335986/) The warning was removed because URP / HDRP now no longer need to have an asset assigned to both Graphics and Quality to work.

---
### Comments to reviewers
This PR requires the following [one] (https://github.cds.internal.unity3d.com/unity/unity/pull/4258)
Explanations:
Unity offers two locations to set an Scriptable Render Pipeline asset: one from the Graphics Settings window, and another for each Quality Level. The desired behavior is to consider the one from Quality Settings (`QualitySettings.renderPipeline`) as an override to the one in Graphics (`GraphicsSettings.renderPipelineAsset `or `GraphicsSettings.defaultRenderPipeline`). The result of this is accessible from `GraphicsSettings.currentRenderPipeline`.

For example:
| GraphicsSettings.renderPipelineAsset | QualitySettings.renderPipeline | GraphicsSettings.currentRenderPipeline  |
| :-------------: | :-------------: | :-------------: |
| SRP 1 | None | SRP 1 |
| None | None | None (Built-in) |
| SRP 1 | SRP 1| SRP 1 |
| SRP 1 | SRP 2| SRP 2 |
| None | SRP 2| SRP 2 |

The previous limitation to force users to have an SRP in Graphics window to be able to override it in Quality is no longer applicable and has been removed. This was possible with the introduction of fixes in URP and HDRP to remove the dependency in 2021.2.

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
+++ b/com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
@@ -644,7 +644,7 @@ namespace UnityEditor.Rendering.Universal
             Profiler.BeginSample(k_ProcessShaderTag);
 #endif
 
-            UniversalRenderPipelineAsset urpAsset = GraphicsSettings.renderPipelineAsset as UniversalRenderPipelineAsset;
+            UniversalRenderPipelineAsset urpAsset = UniversalRenderPipeline.asset;
             if (urpAsset == null || compilerDataList == null || compilerDataList.Count == 0)
                 return;
 


com.unity.render-pipelines.universal/Editor/UniversalAnalytics.cs
```diff
@@ -37,7 +37,7 @@ static void SendUniversalEvent()
             if (!EnableAnalytics())
                 return;
 
-            // Needd to check if this isn't null
+            // Need to check if this isn't null
             UniversalRenderPipelineAsset rendererAsset = GraphicsSettings.currentRenderPipeline as UniversalRenderPipelineAsset;
```
phi-lira, line 41: Ah we have something similar to `HDRenderPipeline.currentAsset` 

UniversalRenderPipeline.asset (I guess this was added before the quality settings one so hence the name). maybe we could align and rename this to `currentAsset`. It would make more sense.
https://sourcegraph.com/github.com/Unity-Technologies/Graphics@ba52521a7a0d7015b9d34e0f9d8fa47664ff23f1/-/blob/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs#L465-468

phi-lira (commented):
Not related to your PR but since we are talking about the API here. 
I'm confused as some API has `asset` some others don't. Are all of these concerning assets?
I'm also confused between what is `GraphicsSettings.currentRenderPipeline` and `GraphicsSettings.renderPipelineAsset`

com.unity.render-pipelines.universal/Editor/UniversalAnalytics.cs
jenniferd-unity: Do you want me to redo this line and change `GraphicsSettings.currentRenderPipeline as UniversalRenderPipelineAsset;` to `UniversalRenderPipeline.asset` ? It could be a good idea

jenniferd-unity:
> Not related to your PR but since we are talking about the API here.
> I'm confused as some API has `asset` some others don't. Are all of these concerning assets?
> I'm also confused between what is `GraphicsSettings.currentRenderPipeline` and `GraphicsSettings.renderPipelineAsset`

@phi-lira I entirely agree thats why I am trying to reduce the headache for people by going over the code. One of the issue is that we have two APIs doing the exact same thing.
So `GraphicsSettings.renderPipelineAsset == GraphicsSettings.defaultRenderPipeline` == asset in Graphics Settings
and `GraphicsSettings.currentRenderPipeline == QualitySettings.renderPipeline ?? GraphicsSettings.renderPipelineAsset;`

com.unity.render-pipelines.universal/Editor/UniversalAnalytics.cs
phi-lira: I'd appreciate if you could do it, but will not block your PR as this is not introduced in your PR.

phi-lira (approved):

sebastienlagarde (approved):

JMargevics (approved):

JMargevics (approved):
No issues from my side ✅

Tried the following: 
- Deleting assigned to a quality level HDRP asset
- Selecting various quality levels
- Undo/Redo everywhere

cinight:
There is a crash when opening projects using the C++ build so I'll resume testing on both the C++ PR and this Graphics PR when the crash issue is resolved.

jenniferd-unity:
@cinight i found what looks like a fix - tried locally and the crash no longer happens but let me know if that is not enough.
Some recent changes caused some code to fail.

cinight (approved):
Tested BoatAttack on Win DX11. I checked assigning and un-assigning RP assets in Quality / Graphics Settings and it's working expectedly. Switching Quality Levels in player is not affected and no other regressions.

erikabar (approved):

phi-lira (approved):

PR #4587, opened 2021-05-19
Removed Draft banners

Vic-Cooper:
Removed all "Draft" banners. 

Given all such pages an initial improvement pass.


julienf-unity (approved):
Does this require backports? I would say probably

PR #4591, opened 2021-05-20
Fix Foundation Linux Editor failure

phi-lira:
### Purpose of this PR
 PR https://github.com/Unity-Technologies/Graphics/pull/4574 updated reference images, however Linux Editor missed Linux Editor playmode because that one only run in nightly.
---
### Testing status
 Only need to test on yamato.
 
---


phi-lira:
007 Test is fixed. Failures in the above project due to decal.

PR #4592, opened 2021-05-20
Decal small fixes

lukaschod:
### Purpose of this PR
- Updating vulkan decal reference images as it regressed.
- Changing HDRP Decal Projector component path name with HDRP
- Changing URP Decal Projector component path name with URP

---
### Testing status

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalMenuItems.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalMenuItems.cs
@@ -6,7 +6,7 @@ namespace UnityEditor.Rendering.HighDefinition
 {
     class DecalMenuItems
     {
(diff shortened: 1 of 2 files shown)


phi-lira (approved):

jenniferd-unity (changes requested):
Could we confirm the path?
I understood we wanted "URP Decal Projector" without a submenu.

Also could you explain what you tested?

TheoWong-pixel:
Yes, if possible we would like to avoid a sub menu and just have the prefix. If the plan is to merge the two decal projectors together it keeps this menu location relatively the same and also menus with only one items are not encouraged as it makes the users work harder to get to the content.

TheoWong-pixel (approved):

jenniferd-unity (approved):
Thanks

lukaschod:
As this is UI and reference image changes. Will skip tests. Merging

PR #4595, opened 2021-05-20
[VFX] Staging fixes

julienf-unity:
### Purpose of this PR
- Per vertex cache motion vector optimization
- Remove some useless reimport of VFX with subgraph

---
### Testing status
Tested locally and with QA Pass

---
### Comments to reviewers
URP not needed, we only added a function in VFXCommon.hlsl in urp path

### DIFF
--- a/com.unity.visualeffectgraph/Shaders/RenderPipeline.meta
+++ b/com.unity.visualeffectgraph/Shaders/RenderPipeline.meta
@@ -1,8 +1,0 @@
-fileFormatVersion: 2
-guid: 642d455b4fec0114b8df4e7448e37b24
-folderAsset: yes
-DefaultImporter:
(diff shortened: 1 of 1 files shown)


PaulDemeulenaere (approved):
LGTM, I'm already using this revision locally.
It requires a recent trunk due to `VFXCameraXRSettings`
Otherwise, expected changes from :
- https://github.cds.internal.unity3d.com/unity/vfx-graphics/pull/246
- https://github.cds.internal.unity3d.com/unity/vfx-graphics/pull/253
🟢 
Potentially, we could use this opportunity to remove this .meta : https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.visualeffectgraph/Shaders/RenderPipeline.meta

PR #4597, opened 2021-05-20
[2021.2] [ShaderGraph] Fix SamplerStates being used by SubGraphs

cdxntchou:
### Purpose of this PR
Fixing a number of issues with SamplerStates and SubGraphs.

https://fogbugz.unity3d.com/f/cases/1336089/
   Regression, where nested subgraphs that pass SamplerStates with the same default settings will cause compile failures

https://fogbugz.unity3d.com/f/cases/1336119/
   Subgraph SamplerState defaults are not respected

https://fogbugz.unity3d.com/f/cases/1336126/
   User can only edit SamplerState property display names when they are created.

2021.2: https://github.com/Unity-Technologies/Graphics/pull/4597
2021.1: https://github.com/Unity-Technologies/Graphics/pull/4580
2020.3: https://github.com/Unity-Technologies/Graphics/pull/4534

### Details:

The existing SamplerState properties would enforce a reference name override, forcing it to something like "SamplerState_Linear_Repeat".

This caused issues because SubGraph properties are treated as function input parameters, whereas ShaderGraph properties are global resources, and they could very easily have the same name, causing errors when we try to access the global resource within a subgraph with a conflicting input parameter.

This change differentiates the naming scheme of the function inputs from the global resources.  The function inputs are now named using the parameter naming scheme used by all other inputs:  basically type + GUID   i.e. "SamplerState_1340f13f".

The properties resources are now always named using the setting-based naming scheme, i.e.  "SamplerState_Linear_Repeat", and they are de-duplicated, so multiple instances of the same one are combined.

We change the UI so that the reference names show up as the property resource name ("SamplerState_Linear_Repeat") and is not editable.  We also now allow the SamplerState display names to be edited at any point, not just on creation.
image

Versioning is applied to the SamplerStateShaderProperties to remove the pre-existing reference name override ("SamplerState_Linear_Repeat"), causing them to fall back to their more unique default naming ("SamplerState_1340f13f").

Previously, subgraph samplerstates could specify defaults within the subgraph, however they were completely ignored -- the samplerstate slots created on a subgraph node would _always_ default to "SamplerState_Linear_Repeat".

Now it will correctly copy the default from the subgraph property.

Note that it does this in a "true inheritance" fashion -- changing the setting on the subgraph property will change the defaults on the slots in all of the corresponding subgraph nodes.

---
### Testing status

Added tests to check that the nested subgraph case continues to work, and also a new scene that checks expected SamplerState behavior.

- [x] Tested against nested subgraph repro case for 1336089.
- [x] Added the repro case as a Test in the ShaderGraph tests.
- [x] Verified the added Test failed before this PR, and succeeds with this PR.
- [x] Added samplerstate graphics tests.
- [x] Verified the samplerstate graphics tests failed before this PR, and succeed now.
- [x] Tested that samplerstate display names can be renamed at any point, in either subgraphs or shadergraphs.
- [x] Tested that modifying display names is undo/redoable.
- [x] Verified that the samplerstate reference names only show up in ShaderGraphs, are not editable, and display the actual HLSL resource name.
- [x] Verified that existing shaders upgrade correctly, with samplerstates removing their override reference name, for both subgraphs and shadergraphs.
- [x] Verified that changing the samplerstate input settings on a subgraph result in the subgraph node defaults changing automatically, and the correct samplerstate value being used, both in loaded graph previews, and without any open graphs in scene view.
- [x] Verified that the samplerstate subgraph default inheritance works through multiple layers of subgraph
- [x] Verified that multiple SamplerStates with the same setting in a ShaderGraph will work, and are deduplicated.
- [x] Verified that multiple SamplerStates with the same setting across a number of subgraphs work, and are deduplicated.
- [x] Tested VFX shaders using SamplerStates continue to work (this even fixes the same bugs in VFX usage)
image

Yamato:

Nightly ShaderGraph: 🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252F1336089/.yamato%252Fall-shadergraph.yml%2523Nightly_ShaderGraph_trunk/6922505/job/pipeline

URP PR Job: 🟡
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252F1336089/.yamato%252Fall-universal_split.yml%2523PR_Universal_Split_trunk/6922942/job

  URP Lighting failed (1 test):
failure matches master build: 808217798554e113f5a8173c327cdcc2dc22ce12
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Fall-universal_split.yml%2523Nightly_Universal_Split_trunk/6894823/job/pipeline

HDRP PR Job: 🟢 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252F1336089/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/6922962/job/pipeline


TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ImportUpdateTests.cs
```diff
@@ -72,14 +72,15 @@ public void ImportMulti(string directory)
...
+                // pause asset database, until everything is copied
                 AssetDatabase.StartAssetEditing();
                 DirectoryCopy(sourceDir, targetDir, true, true);
             }
             finally
             {
                 AssetDatabase.StopAssetEditing();
             }
-
+//            AssetDatabase.Refresh();
```
jessebarker, line 83: Does this need to be here commented out?

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -352,6 +352,14 @@ public virtual void UpdateSlots()
                 // Copy defaults
                 switch (prop.concreteShaderValueType)
                 {
+                    case ConcreteSlotValueType.SamplerState:
```
jessebarker, line 355: As with the other port of these changes, I suspect your formatting error is this indentation (needs to come left one stop).

jessebarker (approved):
A couple of nits, but otherwise LGTM

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/ImportUpdateTests.cs
cdxntchou: I can strip it -- this was a case of fixing a bug in the test code, that was already fixed (though in a slightly different way) in master -- I commented out the colliding code here, but it can be removed (this test passes)

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
cdxntchou: Yup, that was it!  Fixed.

LandonTownsendUnity (approved):
Developer manual testing is sufficient.

PR #4599, opened 2021-05-21
[HDRP] Add support for the recorder / camera bridge in the graphics compositor

pmavridis:
### Purpose of this PR
Previously the output camera of the graphics compositor could not be directly captured by the Recorder, because the camera bridge interface was not implemented.
 
With this PR the recorder and other external scripts that use the camera bridge can directly capture the compositor’s output (without capturing the game view) 

[This was requested by an internal team]

---
### Testing status
I have tested this using the Recorder 3.0.0-pre.1 and by capturing the compositor's output camera:
image
both with and without a RenderTarget output. In both cases the image was captured correctly.

---
### Comments to reviewers
This looks easy and safe to backport, so I have added the tags.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -62,6 +62,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added support for tessellation for all master node in shader graph.
 - Added ValidateMaterial callbacks to ShaderGUI.
 - Added support for internal plugin materials and HDSubTarget with their versioning system.
+- Added support for the camera bridge in the graphics compositor
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


sebastienlagarde (approved):

iM0ve (approved):
I also verified using our HDRP FTP project scene.
Before the PR:
* Was possible to record compositor if the source was Game view
* Impossible to record if the source was set Targeted Camera. Errors in the console while trying to record.

After the PR
* Still possible to record Game view
* Now possible to record when using Targeted Camera. Tested with video and image capture
* No visual difference in the output, before vs after PR


https://user-images.githubusercontent.com/36502659/119350866-7f3a8000-bca8-11eb-9e5f-28a3e6680b43.mp4

adrien-de-tocqueville (approved):

CertifiedRice (approved):

PR #4603, opened 2021-05-21
[URP] RFC stripping disabled variants

svens-unity:
### Checklist for PR maker

---
### Purpose of this PR
This PR adds functionality to the shader preprocessor that allows us to strip the disabled variants of multi-compiles. These disabled variants are never used if all URP assets have the specific feature enabled. When we observe assets that have the feature both enabled and disabled we still need to include both variants as we cannot guarantee which is being used. When a feature is always disabled on all URP assets we already strip variants.

---
### Testing status
Tested using the shader variant tool building scene 154 and 155 on Windows having box projection enabled on all assets.

Before:
5037 compiled  variants in build
After:
4505 compiled variants in build

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
+++ b/com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
@@ -17,6 +17,7 @@ namespace UnityEditor.Rendering.Universal
     [Flags]
     enum ShaderFeatures
     {
+        All = ~0,
         None = 0,
         MainLight = (1 << 0),
         MainLightShadows = (1 << 1),
@@ -67,6 +68,7 @@ namespace UnityEditor.Rendering.Universal
         ShaderKeyword m_AdditionalLightShadows = new ShaderKeyword(ShaderKeywordStrings.AdditionalLightShadows);
         ShaderKeyword m_ReflectionProbeBlending = new ShaderKeyword(ShaderKeywordStrings.ReflectionProbeBlending);
         ShaderKeyword m_ReflectionProbeBoxProjection = new ShaderKeyword(ShaderKeywordStrings.ReflectionProbeBoxProjection);
+        ShaderKeyword m_ReflectionProbeBoxProjectionOff = new ShaderKeyword(ShaderKeywordStrings.ReflectionProbeBoxProjectionOff);
         ShaderKeyword m_DeferredLightShadows = new ShaderKeyword(ShaderKeywordStrings._DEFERRED_LIGHT_SHADOWS);
         ShaderKeyword m_DeferredSoftShadows = new ShaderKeyword(ShaderKeywordStrings._DEFERRED_SHADOWS_SOFT);
         ShaderKeyword m_CastingPunctualLightShadow = new ShaderKeyword(ShaderKeywordStrings.CastingPunctualLightShadow);
@@ -230,6 +232,10 @@ namespace UnityEditor.Rendering.Universal
 
             bool isReflectionProbeBoxProjection = compilerData.shaderKeywordSet.IsEnabled(m_ReflectionProbeBoxProjection);
             if (!IsFeatureEnabled(features, ShaderFeatures.ReflectionProbeBoxProjection) && isReflectionProbeBoxProjection)
+                return true;
+
+            bool isReflectionProbeBoxProjectionOff = compilerData.shaderKeywordSet.IsEnabled(m_ReflectionProbeBoxProjectionOff);
+            if (isReflectionProbeBoxProjectionOff && IsFeatureEnabled(ShaderBuildPreprocessor.alwaysEnabledFeatures, ShaderFeatures.ReflectionProbeBoxProjection))
                 return true;
 
             bool isPunctualLightShadowCasterPass = (snippetData.passType == PassType.ShadowCaster) && compilerData.shaderKeywordSet.IsEnabled(m_CastingPunctualLightShadow);
@@ -466,7 +472,20 @@ namespace UnityEditor.Rendering.Universal
             }
         }
 
-        private static ShaderFeatures _supportedFeatures = 0;
+        public static ShaderFeatures alwaysEnabledFeatures
+        {
+            get
+            {
+                if(_alwaysEnabledFeatures == ShaderFeatures.All)
+                {
+                    FetchAllSupportedFeatures();
+                }
+                return _alwaysEnabledFeatures;
+            }
+        }
+
+        private static ShaderFeatures _supportedFeatures = ShaderFeatures.None;
+        private static ShaderFeatures _alwaysEnabledFeatures = ShaderFeatures.All;
         public int callbackOrder { get { return 0; } }
 #if PROFILE_BUILD
         public void OnPostprocessBuild(BuildReport report)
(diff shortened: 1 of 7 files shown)


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -333,6 +333,7 @@ public static class ShaderKeywordStrings
         internal static readonly string ClusteredRendering = "_CLUSTERED_RENDERING";
         public static readonly string AdditionalLightShadows = "_ADDITIONAL_LIGHT_SHADOWS";
         public static readonly string ReflectionProbeBoxProjection = "_REFLECTION_PROBE_BOX_PROJECTION";
+        public static readonly string ReflectionProbeBoxProjectionOff = "_REFLECTION_PROBE_BOX_PROJECTION_OFF";
```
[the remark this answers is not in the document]
pbbastian: I like `_OFF`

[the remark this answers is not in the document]
ellioman: Same. It clearly shows what the keyword is about.

com.unity.render-pipelines.universal/Editor/VFXGraph/Shaders/VFXPasses.template
```diff
@@ -23,7 +23,7 @@ ${VFXBegin:VFXPassForwardLitAdditionalPragma}
 #pragma multi_compile _ _ADDITIONAL_LIGHTS
 #pragma multi_compile_fragment _ _ADDITIONAL_LIGHT_SHADOWS
 #pragma multi_compile_fragment _ _REFLECTION_PROBE_BLENDING
-#pragma multi_compile_fragment _ _REFLECTION_PROBE_BOX_PROJECTION
+#pragma multi_compile_fragment _REFLECTION_PROBE_BOX_PROJECTION_OFF _REFLECTION_PROBE_BOX_PROJECTION
```
phi-lira, line 26: What happens when we a shader don't declare this line?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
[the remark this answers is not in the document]
phi-lira: Seems reasonable, we have both _OFF and _ON. This is also what Unity generates when using attribute [ToggleOff] and [ToggleOn] respectively in shader properties.

https://sourcegraph.com/github.com/Unity-Technologies/Graphics/-/blob/com.unity.render-pipelines.universal/Shaders/ComplexLit.shader#L111

[ToggleOff] generates shader_feature FEATURE_OFF

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -466,7 +472,20 @@ public static ShaderFeatures supportedFeatures
             }
         }
 
-        private static ShaderFeatures _supportedFeatures = 0;
+        public static ShaderFeatures alwaysEnabledFeatures
```
phi-lira, line 475: The name is misleading but well naming is hard. You can at least add a comment to explain this property.

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -67,6 +68,7 @@ internal class ShaderPreprocessor : IPreprocessShaders
         ShaderKeyword m_AdditionalLightShadows = new ShaderKeyword(ShaderKeywordStrings.AdditionalLightShadows);
         ShaderKeyword m_ReflectionProbeBlending = new ShaderKeyword(ShaderKeywordStrings.ReflectionProbeBlending);
         ShaderKeyword m_ReflectionProbeBoxProjection = new ShaderKeyword(ShaderKeywordStrings.ReflectionProbeBoxProjection);
+        ShaderKeyword m_ReflectionProbeBoxProjectionOff = new ShaderKeyword(ShaderKeywordStrings.ReflectionProbeBoxProjectionOff);
```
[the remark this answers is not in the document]
phi-lira: +1 for this and also helper function to strip the enabled / disabled variants.

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -232,6 +234,10 @@ bool StripUnusedFeatures(ShaderFeatures features, Shader shader, ShaderSnippetDa
             if (!IsFeatureEnabled(features, ShaderFeatures.ReflectionProbeBoxProjection) && isReflectionProbeBoxProjection)
                 return true;
 
+            bool isReflectionProbeBoxProjectionOff = compilerData.shaderKeywordSet.IsEnabled(m_ReflectionProbeBoxProjectionOff);
+            if (isReflectionProbeBoxProjectionOff && IsFeatureEnabled(ShaderBuildPreprocessor.alwaysEnabledFeatures, ShaderFeatures.ReflectionProbeBoxProjection))
+                return true;
```
phi-lira, line 239: In line with the above comment, it would be less error prone to have a helper function to deal with the enabled/disabled stripping.

phi-lira (commented):
LGTM. Minor comments, my biggest worry is what happens when a shader doesn't declare the multi_compile now. Does Unity handle it automatically by let's say picking the first one from the set or do we have to make sure that _FEATURE_OFF is defined?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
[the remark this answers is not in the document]
pbbastian: I do like having both `_OFF` and `_ON`, but I think it'll be less breaking to not change the existing keyword. That being said, we should consider doing a pass over the keywords in one big go some time, so that we can make the naming entirely consistent.

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
lukaschod: I might be missing something, but why we need this new `alwaysEnabledFeatures`? 

Why not just do this:
`if (isReflectionProbeBoxProjectionOff && IsFeatureDisabled(features, ShaderFeatures.ReflectionProbeBoxProjection))`

svens-unity: We can only strip the disabled variants if all URP assets have the feature enabled - this guarantees no one will be using the off variant (as long we do not allow for runtime modifications). This is how the features are collected:
`                    _supportedFeatures |= supportedFeatures;
                    _alwaysEnabledFeatures &= supportedFeatures;
`
When a flag is enabled in `supportedFeatures` it means that at least one asset has that feature enabled, but there can also be assets that have it disabled. 
`_alwaysEnabledFeatures` on the other hand guarantees that the flag is enabled on all assets.

lukaschod: k, we should also decide to landing patter as my PR will conflict https://github.com/Unity-Technologies/Graphics/pull/4619

PR #4605, opened 2021-05-21
[HDRP][Path Tracing] Fixed transparent unlit

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1335500/
Small fix: the position used for shooting a continuation ray on transparent unlit materials was wrong. It is now fixed, so transparent unlit surfaces will produce the expected look.

The fix has been tested on the test provided in the corresp. bug report.


sebastienlagarde:
@eturquin  does it need to be backported?

sebastienlagarde (approved):

iM0ve (approved):
I have checked the repro case too. Please add screenshots if you can next time for auto-approval :)

Before the PR
Unity_kHyGKLXTCS

After PR
Unity_Bbqhait49G


For any unexpected regressions I also checked the template:

Before the PR
before

After the PR
Unity_JydVR4pAId

PR #4609, opened 2021-05-21
Probe dilation on GPU and across cells (90x times faster)

FrancescoC-unity:
I moved the dilation code to be working on the GPU instead of the CPU. 

This has a couple of advantages:
 
- It is roughly 90x faster :-) On simpler scenes this was the bottleneck for the whole baking so now it is much faster to iterate. 
- It now can span across various cells to find dilation targets.  [--> NOTE: this also mean that it might pick bad dilation across cells which can look worse in case of leaking, but occlusion is a known issue] 
- Can be improved without paying a huge cost. 

The dilation being fast is now always run (was before to, but with broken UX) and I simplified a fair bit the settings. 

I also fixed an issue with Global cbuffers in core (they did not compile). 

I am still not convinced dilation is great in itself and we might need to consider alternative options in the longer run (virtual offset mostly)


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -417,156 +407,93 @@ static void OnAdditionalProbesBakeCompleted()
...
+            Clear();
+        }
 
-            // Sort probes by distance to prioritize closer ones
-            for (int culledProbeIdx = 0; culledProbeIdx < culledProbes.Count; culledProbeIdx++)
-            {
-                float dist = Vector3.Distance(probePositions[culledProbes[culledProbeIdx].idx], probePositions[probeIdx]);
-                culledProbes[culledProbeIdx] = new DilationProbe(culledProbes[culledProbeIdx].idx, dist);
-            }
+        static float CalculateSurfaceArea(Matrix4x4 transform, Mesh mesh)
```
FrancescoC-unity, line 476: Haven't changed this part, I guess github got confused?

alelievr (approved):

sebastienlagarde (approved):

PR #4612, opened 2021-05-24
Adding RenderPass tests to URP_Foundation suite

jonuuukas:
### Purpose of this PR
RenderPass functionality is already added and we agreed to add tests after that. This PR introduces some copies of the existing tests to the URP_Foundation with Native RenderPass option enabled. All these tests help to cover pretty much all of the needed functionality but it might change in the future with new tests getting introduced. 

PR also adds one liner to truly disable RenderPass for XR players.

Few tests have been added to TestFilter due to issues that need fixing from engine side:
GLES3 - MSAA produces an RenderPass-specific error
DX12 - SplitScreen is broken

Tests added: 
001_SimpleCube_RenderPass
001_SimpleCube_deferred_RenderPass
013_CameraMulti_Splitscreen_RenderPass
050_Shader_Graphs_deferred_RenderPass
111_CameraStackMSAA_RenderPass
126_RendererFeaturesActive_RenderPass
204_SSAO_Deferred_RenderPass
---
### Testing status
Almost all of the URP_Foundation players are green (only iPhone_metal is red due to some other issue). Currently rerunning after master merge with hopes iPhone runner was fixed.

UPDATE: Seems like merging master actually broke some tests, so currently investigating

UPDATE 2: All of the added tests are green in the URP_Foundation suite, only Win Vulkan is red (same failures as in master branch)
---
### Comments to reviewers


erikabar (approved):

manuele-bonanno (approved):

CertifiedRice (approved):

phi-lira:
LGTM we should discuss as a follow up a better way to test render pass. Maybe we could run tests once with render pass disable and once with render pass enabled?

phi-lira (approved):
LGTM to get render pass coverage. We should discuss how to more efficiently test this as it doesn't scale to duplicate scenes to test render pass on/off

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -647,6 +646,8 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
                     SortStable(m_ActiveRenderPassQueue);
                 }
 
+                if (cameraData.xr.enabled)
+                    useRenderPassEnabled = false;
```
nigeljw-unity, line 650: Does it make sense to set this here? cameraData.xr.enabled should be relatively static. Any reason why we do not do this earlier?

jonuuukas: So I've actually tried multiple approaches on disabling RenderPass for XR and this one was the only way it worked (was a little bit difficult because locally i didn't have a machine i could reproduce the failing tests that required this change). I think I could find a better place as this was done just to be 100% sure it's always disabled. I can address moving this to a better place in the PR i have in the works (which is waiting for the test coverage this PR introduces), is that alright with you?

nigeljw-unity (approved):
LGTM

PR #4613, opened 2021-05-24
Make light cookies shader/public api more consistent

eh-unity:
### Purpose of this PR
Make light cookies shader api more consistent with rest of URP.
Make core blit shaders internal.
Pack cookies enable status into bits, 1 bit per light cookie. (More instructions, but gave consistent minor 1-1.5% perf improvement on MacBook).

---
### Testing status
Ran tests locally.

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
+++ b/com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
@@ -199,7 +199,8 @@ namespace UnityEngine.Rendering.Universal
             {
                 unsafe
                 {
-                    int len = bitCapacity;
+                    Debug.Assert(bitCapacity < 4096, "Bit string too long! It was truncated!");
+                    int len = Math.Min(bitCapacity, 4096);
                     byte* buf = stackalloc byte[len];
                     for (int i = 0; i < len; i++)
                     {


com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
```diff
@@ -115,6 +120,97 @@ public void Resize(int size)
             }
         }
 
+        private struct ShaderBitArray
+        {
+            const int kBitsPerElement = 32;
+            const int kElementShift = 5;
+            const int kElementMask = (1 << kElementShift) - 1;
+
```
svens-unity, line 128: I believe the convention is to do `k_`

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
```diff
@@ -115,6 +120,97 @@ public void Resize(int size)
...
+            private float[] m_data;
```
svens-unity, line 129:
```suggestion
            private float[] m_Data;
```

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
```diff
@@ -115,6 +120,97 @@ public void Resize(int size)
...
+
+                int newElemCount = ((bitCount + (kBitsPerElement - 1)) / kBitsPerElement);
+                if (newElemCount == m_data?.Length)
+                    return;
+
+                var newData = new float[newElemCount];
+                if (m_data != null)
+                {
+                    for (int i = 0; i < m_data.Length; i++)
+                        newData[i] = m_data[i];
```
svens-unity, line 148: Could also use [UnsafeUtility.MemCpy](https://docs.unity3d.com/ScriptReference/Unity.Collections.LowLevel.Unsafe.UnsafeUtility.MemCpy.html) iirc

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
```diff
@@ -115,6 +120,97 @@ public void Resize(int size)
...
+                }
+                m_data = newData;
+            }
+
+            public void Clear()
+            {
+                for (int i = 0; i < m_data.Length; i++)
+                    m_data[i] = 0;
```
svens-unity, line 156: If that is changed to memcpy this could be changed to memset

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
```diff
@@ -115,6 +120,97 @@ public void Resize(int size)
...
+                    }
+                }
+            }
+
+            public override string ToString()
+            {
+                unsafe
+                {
+                    int len = bitCapacity;
+                    byte* buf = stackalloc byte[len];
```
svens-unity, line 203: We do not have a cap on bitCapacity. As I recall it is strongly recommended to have a hard limit when using stackalloc and have a fallback path using non stackalloc array when this limit is exceeded.

svens-unity (approved):

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
eh-unity: Technically we do, maxLights / 32, but yes it would be better to not trust that as that can change.

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
eh-unity: Thanks. I'll see if I can use those instead.

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
eh-unity: fixed.

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
eh-unity: fixed.

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
eh-unity: It becomes a bit of hassle to do unsafe block + 2 fixed block pointers in order to use this. It's not that many elements usually (N/32), so I'll leave it as a simple for loop.

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
eh-unity: I decided to leave this as is, it avoids doing unsafe and fixed blocks.

com.unity.render-pipelines.universal/ShaderLibrary/LightCookie/LightCookieInput.hlsl
```diff
@@ -75,43 +70,58 @@ int URP_LightCookie_GetLightType(int lightIndex)
...
+
+    #if USE_STRUCTURED_BUFFER_FOR_LIGHT_DATA
+        uint elem = asuint(_AdditionalLightsCookieEnableBitsBuffer[elemIndex]);
+    #else
+        uint elem = asuint(_AdditionalLightsCookieEnableBits[elemIndex]);
+    #endif
+
+    return (elem & (1u << bitOffset)) != 0u;
+#endif
 }
```
ellioman, line 125: Will we enable the `#if 0 `section at some point?

ellioman (approved):
One question but overall a very nice API update.

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
```diff
@@ -11,18 +11,22 @@ static class ShaderProperty
...
-            public static readonly int additionalLightsWorldToLightBuffer = Shader.PropertyToID("_AdditionalLightsWorldToLightBuffer"); // TODO: really a light property
-            public static readonly int additionalLightsLightTypeBuffer = Shader.PropertyToID("_AdditionalLightsLightTypeBuffer"); // TODO: really a light property
+            public static readonly int additionalLightsCookieEnableBits = Shader.PropertyToID("_AdditionalLightsCookieEnableBits");
 
+            public static readonly int additionalLightsCookieAtlasUVRectBuffer = Shader.PropertyToID("_AdditionalLightsCookieAtlasUVRectBuffer");
             public static readonly int additionalLightsCookieAtlasUVRects = Shader.PropertyToID("_AdditionalLightsCookieAtlasUVRects");
-            public static readonly int additionalLightsWorldToLights = Shader.PropertyToID("_AdditionalLightsWorldToLights"); // TODO: really a light property
-            public static readonly int additionalLightsLightTypes = Shader.PropertyToID("_AdditionalLightsLightTypes"); // TODO: really a light property
+
+            // TODO: these should be generic light property
```
ellioman, line 24: You know me :)
Can we kill the TODO and make a Jira card about this task?

com.unity.render-pipelines.universal/ShaderLibrary/LightCookie/LightCookieInput.hlsl
eh-unity: No, that's the old version. I left that in so I can do perf testing on mobile later (new on is on my macbook around 1% faster). Also it's an emergency switch to restore the old behavior, just in case. I'll remove it with my next PR.

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
eh-unity: I decided to just clamp it just in case. It's a private method and mostly for debug.

com.unity.render-pipelines.universal/Runtime/LightCookieManager.cs
eh-unity: I'll do a task for it.

CertifiedRice (approved):

erikabar (approved):
no regressions found with URP template scene

phi-lira (approved):

eh-unity:
Merged, all green, all approved. Only failing test "Top Asset Store" test was a known failure not related to this PR.

PR #4498, opened 2021-05-10
[URP][HDRP] Global Settings polish

jenniferd-unity:
HDRP Global Settings:
image

URP Global Settings:
image

---
### Purpose of this PR
Iterating on feedback from previous PR: https://github.com/Unity-Technologies/Graphics/pull/4199
This includes:
- [Fix] wording similar between URP and HDRP on tooltips and labels
- [Maintenance] moved UI text to their own file for enhanced readability
- [Maintenance] renamed DefaultSettingsPanel to HDGlobalSettingsWindow for improved readability
- [Fix] Right click in Project window to create URP Global Settings asset is done in the current folder
create-urp-globalsettings-projectWindow
create-urp-globalsettings-graphicsWindow

- [Added] Added documentation links on some sections with available doc pages
- [Added] Added ability to reset light and decal layer names 
layernames-reset-all
layernames-reset-individual
layernames-reset-urp

- [Fix] Layout issue on minimal size
urp-global-settings-minSize-fix

- [Doc] Added documentation link to the Global Settings in the top right corners
- [Doc] Updated doc
- [Fixed] UI for probe volume is now properly aligned
- [Optim] Probe Volume checkbox does not reload the RP if not running HDRP

---
### Testing status
I have tested behaviors Editor only on windows. Maybe we could do a sweep on the doc links too to make sure it makes sense.
HDRP > Could you make sure I didnt break the Probe Volume checkbox behavior? I did an optim where it doesnt reload the render pipeline if not HDRP, I would like to confirm it still works as planned. 

---
### Comments to reviewers
Hakeem's doc is useful to test behavior: [link](https://docs.google.com/document/d/1SzGX9DcRmNsx_nWGuCb80qqtpRra1_ra9mcKp8qixzs/edit?usp=sharing)

PR #4616, opened 2021-05-25
Move supportRuntimeDebugDisplay flag to global settings for HDRP & URP

arttu-peltonen:
### Purpose of this PR
Improve URP build times to solve https://fogbugz.unity3d.com/f/cases/1339115/.

This PR removes `supportRuntimeDebugDisplay` flag from HDRP Settings Asset, and introduces it in HDRP Global Settings Asset and URP Global Settings Asset, for following reasons:

- This type of setting makes more sense on a "per project basis", rather than "per quality level basis".
- URP was missing this option altogether. Having this option will decrease shader compilation times for many users, since you now need to actively opt-in to compiling debug shader variants for player builds.

I've also removed the temporary URP_TEST_AGGRESSIVE_SHADER_STRIPPING define used to remove debug variants from test projects; since `supportRuntimeDebugDisplay` is false by default, debug shaders will still be stripped from test project player builds.

---
### Testing status
I've tested that when opening an existing project where the active HDRP asset has "Runtime Debug Display" option is set to true/false, the same value persists in the HDRP Global Settings Asset using this PR. I think it would be good to test that this migration happens correctly with older versions of the HDRP asset too.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineAsset.Migration.cs
```diff
@@ -204,10 +205,19 @@ enum Version
...
             {
                 data.m_RenderPipelineSettings.supportSurfaceGradient = false;
+            }),
+            MigrationStep.New(Version.MovedSupportRuntimeDebugDisplayToGlobalSettings, (HDRenderPipelineAsset data) =>
+            {
+#pragma warning disable 618 // Type or member is obsolete
+                HDRenderPipelineGlobalSettings.instance.supportRuntimeDebugDisplay = data.m_RenderPipelineSettings.m_ObsoleteSupportRuntimeDebugDisplay;
+#if UNITY_EDITOR
+                UnityEditor.EditorUtility.SetDirty(HDRenderPipelineGlobalSettings.instance);
+#endif
```
arttu-peltonen, line 219: I'm not sure if this is a good thing to do. I must set the global asset dirty to ensure to ensure it gets saved. Or is this code in the wrong place and I should instead increment the version on the GlobalSettins asset?

arttu-peltonen: I'm now realizing there could be multiple RP assets, and whatever gets migrated last would remain in effect on the global asset. So I guess I'm doing this backward now - I should instead migrate the global asset to a new version, and just look up the currently active RP asset for the value.

arttu-peltonen:
@RSlysz & @JulienIgnace-Unity please focus on the HDRP parts.
@jenniferd-unity & @phi-lira please focus on the URP parts.
Thanks!

TestProjects/HDRP_RuntimeTests/Assets/HDRPDefaultResources/HDRenderPipelineGlobalSettings.asset
```diff
@@ -62,8 +62,7 @@ MonoBehaviour:
     materialQuality: 0
   m_RenderPipelineResources: {fileID: 11400000, guid: 3ce144cff5783da45aa5d4fdc2da14b7,
     type: 2}
-  m_RenderPipelineRayTracingResources: {fileID: 11400000, guid: ba6e7f30e5cffc249a8bf7ee5756c196,
-    type: 2}
+  m_RenderPipelineRayTracingResources: {fileID: 0}
```
arttu-peltonen, line 65: This happened automatically because `ClearRayTracingResources()` was called - @jenniferd-unity is this expected?

TestProjects/VisualEffectGraph_URP/ProjectSettings/ProjectSettings.asset
```diff
@@ -576,19 +576,19 @@ PlayerSettings:
   webGLThreadsSupport: 0
   webGLDecompressionFallback: 0
   scriptingDefineSymbols:
-    1: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING;URP_ENABLE_DEBUG_DISPLAY
```
arttu-peltonen, line 579: FYI - `URP_ENABLE_DEBUG_DISPLAY` is not used anymore so removed it here as well.

RSlysz (approved):
Though, I wonder if we can totally remove it from HDRPAsset with API upgrader for better clarity (see below)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
```diff
@@ -320,8 +319,13 @@ internal bool SupportsAlpha()
...
         /// <summary>Support motion vectors.</summary>
         public bool supportMotionVectors;
+
         /// <summary>Support runtime debug display.</summary>
-        public bool supportRuntimeDebugDisplay;
+        public bool supportRuntimeDebugDisplay
+        {
+            get => HDRenderPipelineGlobalSettings.instance.supportRuntimeDebugDisplay;
+            set => HDRenderPipelineGlobalSettings.instance.supportRuntimeDebugDisplay = value;
+        }
```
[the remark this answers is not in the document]
RSlysz: Good point. I believe we should provide a public API at some point (perhaps editor only ?)

I guess it is out of scope for this PR anyway.

TestProjects/HDRP_RuntimeTests/Assets/HDRPDefaultResources/HDRenderPipelineGlobalSettings.asset
jenniferd-unity: We could revert this portion yes

com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
```diff
@@ -65,7 +65,7 @@ protected override bool DoShadersStripper(HDRenderPipelineAsset hdrpAsset, Shade
             // If requested by the render pipeline settings, or if we are in a release build,
             // don't compile fullscreen debug display variant
             bool isFullScreenDebugPass = snippet.passName == "FullScreenDebug";
-            if (isFullScreenDebugPass && (!Debug.isDebugBuild || !hdrpAsset.currentPlatformRenderPipelineSettings.supportRuntimeDebugDisplay))
+            if (isFullScreenDebugPass && (!Debug.isDebugBuild || !HDRenderPipelineGlobalSettings.instance.supportRuntimeDebugDisplay))
```
jenniferd-unity, line 68: you should probably use `globalSettings` instead of `HDRenderPipelineGlobalSettings.instance`

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/DefaultSettingsPanel.cs
```diff
@@ -84,6 +84,7 @@ public class Styles
 
             internal static readonly GUIContent lensAttenuationModeContentLabel = EditorGUIUtility.TrTextContent("Lens Attenuation Mode", "Set the attenuation mode of the lens that is used to compute exposure. With imperfect lens some energy is lost when converting from EV100 to the exposure multiplier.");
             internal static readonly GUIContent probeVolumeSupportContentLabel = EditorGUIUtility.TrTextContent("Probe Volumes (Experimental)", "Set whether Probe volumes are supported by the project. The feature is highly experimental and subject to changes.");
+            internal static readonly GUIContent supportRuntimeDebugDisplayContentLabel = EditorGUIUtility.TrTextContent("Runtime Debug Shaders", "When disabled, all debug display shader variants are removed when you build for the Unity Player. This decreases build time, but prevents the use of Rendering Debugger in Player builds.");
```
jenniferd-unity, line 87: FYI - with the polish PR https://github.com/Unity-Technologies/Graphics/pull/4498 this is moved to a new file. we need to coordinate

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.cs
```diff
@@ -699,6 +699,12 @@ internal bool AddDiffusionProfile(DiffusionProfileSettings profile)
         [SerializeField]
         internal bool supportProbeVolumes = false;
 
+        /// <summary>
+        /// Controls whether debug display shaders for Rendering Debugger are available in Player builds.
+        /// </summary>
+        [SerializeField]
+        public bool supportRuntimeDebugDisplay = false;
```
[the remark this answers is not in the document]
jenniferd-unity: actually - why is this public and not internal?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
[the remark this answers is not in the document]
jenniferd-unity: yeah right now the Global Settings class is internal as main point of entry remains the HDRP Asset. I guess the HDRP team can revisit this behavior in 2022.x if they like

com.unity.render-pipelines.universal/Editor/GlobalSettings/UniversalGlobalSettingsWindow.cs
```diff
@@ -45,6 +45,9 @@ public class Styles
             internal static readonly GUIContent lightLayerName6 = EditorGUIUtility.TrTextContent("Light Layer 6", "The display name for Light Layer 6.");
             internal static readonly GUIContent lightLayerName7 = EditorGUIUtility.TrTextContent("Light Layer 7", "The display name for Light Layer 7.");
 
+            internal static readonly GUIContent miscSettingsLabel = EditorGUIUtility.TrTextContent("Miscellaneous");
+            internal static readonly GUIContent supportRuntimeDebugDisplayContentLabel = EditorGUIUtility.TrTextContent("Runtime Debug Shaders", "When disabled, all debug display shader variants are removed when you build for the Unity Player. This decreases build time, but prevents the use of Rendering Debugger in Player builds.");
```
jenniferd-unity, line 49: same comment as above - this was moved to its own file - we need to sync

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -153,7 +153,7 @@ bool StripUnusedPass(ShaderFeatures features, ShaderSnippetData snippetData)
 
         bool StripUnusedFeatures(ShaderFeatures features, Shader shader, ShaderSnippetData snippetData, ShaderCompilerData compilerData)
         {
-            bool stripDebugDisplayShaders = !Debug.isDebugBuild;
+            bool stripDebugDisplayShaders = !Debug.isDebugBuild || !UniversalRenderPipelineGlobalSettings.instance.supportRuntimeDebugDisplay;
```
jenniferd-unity, line 156: can URP Global Settings be null?

jenniferd-unity (approved):

jenniferd-unity (changes requested):
Actually i just realized you would also need to update the documentation/manual for URP and HDRP Global Settings.
Have Vic and Kerry as reviewers for those changes.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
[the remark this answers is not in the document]
RSlysz: @sebastienlagarde (for awareness)

TestProjects/HDRP_RuntimeTests/Assets/HDRPDefaultResources/HDRenderPipelineGlobalSettings.asset
sebastienlagarde: you shouldn't commit this file anyway

TestProjects/VisualEffectGraph_URP/ProjectSettings/ProjectSettings.asset
```diff
@@ -576,19 +576,19 @@ PlayerSettings:
...
-    18: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING
-    19: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING
-    21: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING
-    23: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING
-    25: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING
-    26: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING
-    27: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING
-    28: UNITY_POST_PROCESSING_STACK_V2;URP_TEST_AGGRESSIVE_SHADER_STRIPPING
+    1: UNITY_POST_PROCESSING_STACK_V2
+    4: UNITY_POST_PROCESSING_STACK_V2
```
sebastienlagarde, line 580: I don't understand wher this define come from (UNITY_POST_PROCESSING_STACK_V2), we don't have PPV2 in our test project AFAIK

com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
```diff
@@ -74,7 +74,7 @@ protected override bool DoShadersStripper(HDRenderPipelineAsset hdrpAsset, Shade
             // we allow user to make explicit request for it and it bypass other request
             if (!hdrpAsset.currentPlatformRenderPipelineSettings.supportRuntimeAOVAPI)
```
sebastienlagarde, line 75: just for awareness. When we use our AOV API at runtime, we force this settings. This mean that we still have a behavior per configuration for HDRP. It is ok like this

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.Migration.cs
```diff
@@ -40,6 +41,16 @@ enum Version
             MigrationStep.New(Version.UpdateLensFlare, (HDRenderPipelineGlobalSettings data) =>
             {
                 FrameSettings.MigrateToLensFlare(ref data.m_RenderingPathDefaultCameraFrameSettings);
+            }),
+            MigrationStep.New(Version.MovedSupportRuntimeDebugDisplayToGlobalSettings, (HDRenderPipelineGlobalSettings data) =>
+            {
+#pragma warning disable 618 // Type or member is obsolete
+                var activePipeline = GraphicsSettings.currentRenderPipeline as HDRenderPipelineAsset;
```
sebastienlagarde, line 48: Please, update the upgrade guide of HDRP to mention this migration step as it will be destructive (i.e if some users have configure runtime debug display on other configuration but not on the one in graphics settings we will loose this information. It is ok as this settings is not really mean to be shipped or control like this but it is better to inform users in the upgrade guide (upgrading from 21.1 to 21.2 md file)

sebastienlagarde (commented):
so change is good for me, just need to add upgade guide section

CertifiedRice (approved):

TestProjects/VisualEffectGraph_URP/ProjectSettings/ProjectSettings.asset
arttu-peltonen: I didn't know what it is, and that's why I preferred not to touch it...

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
arttu-peltonen: Good point, the instance can indeed be null - I assumed the code would ensure it exists but looks like a lot of that code is editor-only, so there might be cases in player builds where it's hasn't been set. I'll add a check to be sure.

TestProjects/HDRP_RuntimeTests/Assets/HDRPDefaultResources/HDRenderPipelineGlobalSettings.asset
arttu-peltonen: This file must be committed. HDRP & URP runtime tests assert unless global settings asset is up-to-date. However, I can revert the ray tracing resources part if desired.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.cs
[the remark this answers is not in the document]
arttu-peltonen: It's because the original field was public, and I incorrectly thought that the global settings class itself was public (instead of internal). Still, if the class is made public at some point, I see no reason why this shouldn't be public.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/DefaultSettingsPanel.cs
arttu-peltonen: Yup, I'll wait for that to land and will fix my PR afterwards.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineGlobalSettings.Migration.cs
arttu-peltonen: Good, I'll add that. @RSlysz I wonder if the same thing should have been done for HDRP Asset -> HDRP Global Settings itself? Or is it not required because that migrates from `defaultRenderPipeline`, rather than from `currentRenderPipeline`? Should I also use `defaultRenderPipeline` here?

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
```diff
@@ -9,50 +9,50 @@ The High Definition Render Pipeline (HDRP) adds the HDRP Settings tab to Unity's
...
+The Default Volume Profile Asset references a Volume Profile in the HDRP Package folder called DefaultSettingsVolumeProfile by default. Below it, you can add [Volume overrides](Volume-Components.md), and edit their properties. You can also assign your own Volume Profile to this property field. Be aware that this property must always reference a Volume Profile. If you assign your own Volume Profile and then delete it, HDRP automatically re-assigns the DefaultSettingsVolumeProfile from the HDRP Package folder.
 
-[Frame Settings](Frame-Settings.md) control the rendering passes that Cameras make at runtime. This section allows you to set default values for the Frame Settings that all Cameras use if you do not enable their Custom Frame Settings checkbox. For information about what each property does, see [Frame Settings](Frame-Settings.md).
+The LookDev Volume Profile Asset references the Volume Profile that will be used in the [LookDev window](Look-Dev.md). It works the same way than the Default Volume profile except that in this asset you can't put a [Visual Environment Component](Override-Visual-Environment.md) or skies component because they are overwritten by the LookDev.
```
Vic-Cooper, line 18: Remove passive voice: 

The LookDev Volume Profile Asset references the Volume Profile HDRP uses in the [LookDev window](Look-Dev.md). This Asset works in almost the same way as the Default Volume Profile Asset, except that it overrides [Visual Environment Components](Override-Visual-Environment.md) and sky components.

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
```diff
@@ -9,50 +9,50 @@ The High Definition Render Pipeline (HDRP) adds the HDRP Settings tab to Unity's
...
+## Frame Settings
 
-The LookDev Volume Profile Asset references the Volume Profile that will be used in the [LookDev window](Look-Dev.md). It works the same way than the Default Volume profile except that in this asset you can't put a [Visual Environment Component](Override-Visual-Environment.md) or skies component because they are overwritten by the LookDev.
+[Frame Settings](Frame-Settings.md) control the rendering passes that Cameras make at runtime. This section allows you to set default values for the Frame Settings that all Cameras use if you do not enable their Custom Frame Settings checkbox. For information about what each property does, see [Frame Settings](Frame-Settings.md).
```
Vic-Cooper, line 27: Small tweaks: 

The [Frame Settings](Frame-Settings.md) control the rendering passes that Cameras make at runtime. Use this section to set default values for the Frame Settings that all Cameras use if you do not enable their Custom Frame Settings checkbox. For information about what each property does, see [Frame Settings](Frame-Settings.md).

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
```diff
@@ -9,50 +9,50 @@ The High Definition Render Pipeline (HDRP) adds the HDRP Settings tab to Unity's
...
+| ------------------------- | ------------------------------------------------------------ |
+| **Player Resources**      | Stores references to Shaders and Materials that HDRP uses. When you build your Unity Project, HDRP embeds all of the resources that this Asset references. It allows you to set up multiple render pipelines in a Unity Project and, when you build the Project, Unity only embeds Shaders and Materials relevant for that pipeline. This is the Scriptable Render Pipeline equivalent of Unity’s Resources folder mechanism. When you create a new HDRP Global Settings Asset, the HDRenderPipelineRuntimeResources from HDRP package is automatically referenced in it. |
+| **Ray Tracing Resources** | Stores references to Shaders and Materials that HDRP uses for ray tracing. HDRP stores these resources in a separate Asset file then the main pipeline resources so it can use less memory for applications that don't support ray tracing. When you create a new HDRP Global Settings Asset, the HDRenderPipelineRayTracingResources from HDRP package is automatically referenced in it if your project use ray tracing. |
```
Vic-Cooper, line 57: Replace "then" with "from": 

"...HDRP stores these resources in a separate Asset file from the main pipeline resources so that it can use less memory for applications that don't support ray tracing..."

Vic-Cooper (changes requested):
Requested a few changes.

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
arttu-peltonen: @Vic-Cooper thanks for looking over these! Just to clarify what I've changed here - I've only added documentation that relates to HDRP/URP Global Settings Assets. All other changes are simply reorganizing the documentation so that they appear in the same order as the settings UI itself, to make it easier for people to read; sadly the diffs don't make this easy to see. I'm happy to make the all of the suggested changes if you prefer, but just wanted to clarify that this PR didn't add them :)

Vic-Cooper: That's alright! As I take over docs I haven't touched before I might see the occasional thing I'd like to tweak (a few typos and things). 
The content here is otherwise great :)

Vic-Cooper (approved):

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
arttu-peltonen: Calling Ensure() here now (which asserts on failure to create) like HDRP does just to make sure.

jenniferd-unity (approved):

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -296,6 +296,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Splited HDProjectSettings with new HDUserSettings in UserProject. Now Wizard working variable should not bother versioning tool anymore (case 1330640)
 - Removed redundant Show Inactive Objects and Isolate Selection checkboxes from the Emissive Materials tab of the Light Explorer (case 1331750).
 - Renaming Decal Projector to HDRP Decal Projector.
+- The HDRP Render Graph now uses the new RendererList API for rendering and (optional) pass culling.
```
sebastienlagarde: looks like there is a merge issue with the changelog

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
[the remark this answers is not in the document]
sebastienlagarde: ok

com.unity.render-pipelines.high-definition/CHANGELOG.md
arttu-peltonen: @sebastienlagarde do you mean these lines shouldn't appear in my PR? I think if you look at the changes in the merge commit itself, it displays "unrelated changes" that happened in that merge. But if you view [this file in "Show all commits" view](https://github.com/Unity-Technologies/Graphics/pull/4616/files?file-filters%5B%5D=.md#diff-621f39c820c4b8b22426a93fb942adb48b98d3955ec07eec68d40a1726bea33f), it correctly displays only the one line that I've added.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/HDGlobalSettingsWindow.cs
```diff
@@ -455,14 +455,20 @@ static void DrawMiscSettings(SerializedHDRenderPipelineGlobalSettings serialized
...
 #endif
+
+                Type renderPipeManagerType = typeof(RenderPipelineManager);
+                var cleanupRenderPipeline = renderPipeManagerType.GetMethod("CleanupRenderPipeline", BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.Public);
                 EditorGUI.BeginChangeCheck();
                 EditorGUILayout.PropertyField(serialized.supportProbeVolumes, Styles.probeVolumeSupportContentLabel);
                 if (EditorGUI.EndChangeCheck())
                 {
                     // If we are running HDRP, we need to make sure the RP is reinitialized
                     if (HDRenderPipeline.currentPipeline != null)
```
JulienIgnace-Unity, line 466: Was that supposed to be removed since we seem to be calling it in a more generic way now?

jenniferd-unity: what generic way are you talking about?

JulienIgnace-Unity: The part where we get the method like this GetMethod("CleanupRenderPipeline"). Or am I misunderstanding things?
If not then I don't understand why this was added.

jenniferd-unity: actually it is a bad merge - i moved that code out of this function. @arttu-peltonen could you double check

arttu-peltonen: Whoops yes merge artifact indeed. Fixed now - thanks @JulienIgnace-Unity and @jenniferd-unity!

iM0ve (approved):
What I tested:
* I have created a new template project on 2019.4 and upgraded it to this branch, checked that the Runtime Debug Display option is still checked.
* Built the project with Runtime Debug Display On + Developer mode On - I could enable and see debug modes, as expected (adding a screenshot)
* Built the project with Runtime Debug Display Off + Developer mode Off - could not enable debug mode, as expected.
* Built the project with Runtime Debug Display Off + Developer mode On - cold enable debug mode but all shaders were just white. So likely expected, was just checking this to see if build doesn't fail.

Found some general UX issues, but they are not introduced with this PR so theyre not blocking anything. Will contact UX team about them

* Discoverability. The setting is still hard to find. If possible we could move it Build settings
* The most used scenario is very confusing. Most of the time you simply want to enable the debug mode to have it fully functional. However to do this you need to know to check both checkboxes Runtime Debug Display and Developer mode. They are also in completely  different locations
* There are some setting combinations that maybe dont make sense. For example Runtime Debug Display On + Developer mode Of. In this case you will compile the shaders but wont be able to use them in the build.
* Feels that these two options should be linked or at least in the same location. Or combined into one
* Errors - checked that it also happens when upgrading to master, will file a separate bug case


**Testing results**

Before and after upgrade template
Screenshot 2021-06-01 at 16 26 14
Screenshot 2021-06-01 at 16 40 16

Working debug view on build
Screenshot 2021-06-01 at 17 26 02

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
```diff
@@ -9,50 +9,50 @@ The High Definition Render Pipeline (HDRP) adds the HDRP Settings tab to Unity's
...
 | Use DLSS Custom Project Id            | Controls whether to use a custom project ID for the NVIDIA Deep Learning Super Sampling module. If you enable this property, you can use **DLSS Custom Project Id** to specify a custom project ID.<br/>This property only appears if you enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project. |
 | DLSS Custom Project Id                | Controls whether to use a custom project ID for the NVIDIA Deep Learning Super Sampling (DLSS) module. If you enable this property, you can use **DLSS Custom Project Id** to specify a custom project ID. If you disable this property, Unity generates a unique project ID. <br/>This property only appears if you enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project. |
+| Runtime Debug Shaders                 | Disable this property to remove all debug display shader variants when you build for the Unity Player. This decreases build time but as a result, you cannot use Rendering Debugger in Player builds. |
```
reallyfancy, line 48: Checking my understanding: 
* You can only use the Render Pipeline Debugger window with development builds if this setting is enabled
* It works by generating "debug display" shader variants

Questions:
* Which shaders generate debug display variants: the standard HDRP prebuilt ones/all shaders created using Shader Graph/something else? 
* What do the debug display variants do - e.g. enable special debug views like wireframe?
* Does a user have to do anything else to generate these variants?
* What happens in a development build if you don't generate these variants - you can't use it at all? Some modes don't work properly?

reallyfancy (commented):
Under review - need some extra info!

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
arttu-peltonen: Your understanding is correct. 
- The DEBUG_DISPLAY variants are generated by HDRP prebuilt and shader graph shaders. @JulienIgnace-Unity knows more about details related to HDRP.
- Things like wireframe, viewing material channels (albedo, specular etc.), lighting debug etc. Almost all debug options in Rendering Debugger Window work by activating these variants.
- If DEBUG_DISPLAY variants are not there, the debug modes do not work (renders as normal / white / black / pink - not sure actually). This PR doesn't affect this behavior.

reallyfancy (approved):
I've updated the text for the Runtime Debug Shaders description to be clearer and more consistent. Please tweak it if I've inadvertently changed the meaning!

com.unity.render-pipelines.high-definition/Documentation~/Default-Settings-Window.md
```diff
@@ -45,7 +45,7 @@ You have one list per post processing injection point: `After Opaque And Sky`, `
...
 | DLSS Custom Project Id                | Controls whether to use a custom project ID for the NVIDIA Deep Learning Super Sampling (DLSS) module. If you enable this property, you can use **DLSS Custom Project Id** to specify a custom project ID. If you disable this property, Unity generates a unique project ID. <br/>This property only appears if you enable the NVIDIA package (com.unity.modules.nvidia) in your Unity project. |
-| Runtime Debug Shaders                 | Disable this property to remove all debug display shader variants when you build for the Unity Player. This decreases build time but as a result, you cannot use Rendering Debugger in Player builds. |
+| Runtime Debug Shaders                 | When enabled, Unity includes shader variants that let you use the Render Pipeline Debugger window to debug your build. When disabled, Unity excludes ("strips") these variants. Enable this when you want to debug your shaders in the Render Pipeline Debugger window, and disable it otherwise. |
```
arttu-peltonen, line 48: @reallyfancy thanks! Only thing I'd restore is that the window is now called "Rendering Debugger" in the UI, so I believe the docs should use the same term.

```suggestion
| Runtime Debug Shaders                 | When enabled, Unity includes shader variants that let you use the Rendering Debugger window to debug your build. When disabled, Unity excludes ("strips") these variants. Enable this when you want to debug your shaders in the Rendering Debugger window, and disable it otherwise. |
```

cinight (approved):
Did similar testing as Motiejus's comment but with URP template project. Checked the shader variant count in build is correct. Only when Debug Shader ON + Build with Development the shader variant count is higher.
I agree on Motiejus's point of view regarding UX.

lukaschod (approved):

ellioman (approved):

simon-engelbrecht-soerensen (approved):
cnight has approved, so auto approving.

RSlysz:
Testing slack bot 2

PR #4617, opened 2021-05-25
Remove use of obsolete itemHeight.

arttu-peltonen:
### Purpose of this PR
Fixes warning in LookDev window.

---
### Testing status
Compiles without warnings, LookDev window is not empty. (I don't know what the window is supposed to look like so can't verify - code change should be a no-op though)


arttu-peltonen:
Definition of `itemHeight`: 
```
        [Obsolete("itemHeight is deprecated, use fixedItemHeight instead.", false)]
        public int itemHeight
        {
            get => (int)fixedItemHeight;
            set => fixedItemHeight = value;
        }
```

stramit (approved):

PR #4618, opened 2021-05-25
Fix scene hierarchy filtering

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1333684

The issue was twofold: 

- the sky creates some problems (i.e. it writes alpha 1, checked with @pmavridis  and that's expected). We now hide the sky if scene filtering is enabled. 
- Moving of PP to a different file messed a bit the changes that were made originally, specifically alpha was not enabled unless  the pipeline was rebuilt with scene filtering ON. This PR addresses the problem and fixes it. 

Before: 
image

After:
image

**What did I test**: the repro case and you can see something slightly different in the pictures.


pmavridis (approved):
Looks good to me.

CertifiedRice (approved):

sebastienlagarde (approved):

sebastienlagarde:
note to self: retarget to master once HD/bugfix is merge

PR #4620, opened 2021-05-25
Fix issues with initialization of cached area shadows

FrancescoC-unity:
This is a fix for https://fogbugz.unity3d.com/f/cases/1338347/ 

So the bug mentioned was fixed by another PR, but there was still an issue in that situation. The new issue just shows black area light shadows, but without any error in console. This PR fixes said issue. 

Note that while the bug in case only applies to 10.x.x without RG, there is an issue that slipped through more recent packages. 

**What did I test**: the repro case.  

**Note for backport**: this should backport trivially, but this line https://github.com/Unity-Technologies/Graphics/compare/HDRP%2Ffix-cached-area-shadows-init-params?expand=1#diff-df4fd395887dcbe1aa3489bddf28c6c3b8e6aadc100d58d1d051515d5e958184R395 won't exist in older packages. Please let me know if you want me to do an explicit backport.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -215,6 +215,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed overdraw in custom pass utils blur and Copy functions (case 1333648);
 - Fixed invalid pass index 1 in DrawProcedural error.
 - Fixed a compilation issue for AxF carpaints on Vulkan (case 1314040).
+- Fixed issue with hierarchy object filtering.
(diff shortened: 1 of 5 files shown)


sebastienlagarde:
Yes, we will need backport of it

sebastienlagarde:
it miss a changelog

FrancescoC-unity:
I pushed with wrong git account, the changelog is indeed me :D

TomasKiniulis (approved):
Looks good. Tried tweaking various shadow settings, changing update modes, adding multiple area lights on 2021.2 as well as 2020.2 with Render Graph disabled. Didn't encounter more shadow artifacts or compute shader errors.

sebastienlagarde (approved):

sebastienlagarde:
will redirtect this PR to master after hd/bugfix is merge as it may require different backport

PR #4621, opened 2021-05-25
Various Improvements to the volumetric clouds system

anisunity:
Various Improvements to the volumetric clouds system
- Added a slider to control how much the erosion affects the ambient occlusion term.
- Increased the minimal density for the volumetric clouds.
- Changed the format for storing the volumetric clouds presets for an easier edition.
- Reduced the maximal distance per ray step for the volumetric clouds.
- Improved the fly through ghosting artifacts in the volumetric clouds.
- Fixed some depth comparison instabilities with volumetric clouds.
- Added Animation curves for the simple/custom mode.

Testing status
Tests are still green locally.
Played with the new build extensively.


sebastienlagarde (approved):

sebastienlagarde:
@anisunity there is formatting issue

sebastienlagarde:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fclouds-improvements/.yamato%252Fall-hdrp.yml%2523Nightly_HDRP_CUSTOM-REVISION/6958321/job

anisunity:
Pushed the formatting @sebastienlagarde

Vic-Cooper (changes requested):
Docs look good. I've made some small suggestions to the changelog.

anisunity:
@Vic-Cooper review changes pushed

remi-chapelain:
Haven't fully tested everything but for now : 
- Validated that the flythrough was improved even though there's still room for improvement
- Tested Erosion occlusion factor in all the modes
Here's my suggestion for the tooltip : 
Before : "_Controls the erosion occlusion strength_" 
After : "_Controls how much Erosion Factor is taken into account when computing ambient occlusion. 
The Erosion Factor parameter is hard-coded in presets and only editable in Advanced or Manual Mode._"

remi-chapelain (changes requested):
Tested what I wanted to tested. 
Trusting @anisunity for the rest of the testing-improvements. 

Here's what I liked changed to approve : 
- Updating tooltip for Erosion Occlusion parameter to be more clear. 
Before : "_Controls the erosion occlusion strength_"
After : "_Controls how much Erosion Factor is taken into account when computing ambient occlusion. The Erosion Factor parameter is editable in the custom preset, Advanced and Manual Modes._" 
 
- The custom curve are really useful and help users achieve more granularity on how the clouds are rendered, however, for "**Custom Erosion Curve**" and "**Custom Ambient Occlusion Curve**", the default is a curve that starts at -1 in X axis and ends at 1 leading you to believe that the mapping on the cloud height goes from -1 to 1, whereas only from 0 to 1 is taken into account (which makes sense height wise) so I'd suggest having a default curve there, that goes from 0 to 1 in both X and Y

useful part between brackets
image

- Custom AO curve seems to be inverted : 0 in Y axis yields darker clouds, should be inverted
- Mission Erosion Occlusion documentation

remi-chapelain (approved):
All the issues in the previous comment have been fixed in the last commit. 
Double checked that the 3 automated tests were still passing ✔️ 
Approving 🟢

PR #4624, opened 2021-05-25
Changed light reset to preserve type

adrien-de-tocqueville:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1333875/

Note: when resetting a spot or area light, the type is preserved but not the shape (eg rectangle/tube/disc for area is always reset to rectangle).

Behavior now:
light-reset

---
### Testing status
Tested resetting point, area, directional and spot light


iM0ve (approved):
Looks good. Small change. Testing sufficient.

sebastienlagarde (approved):

PR #4628, opened 2021-05-25
[URP] Avoid rendering skybox in a different render pass when copy depth is required

manuele-bonanno:
---
### Purpose of this PR
The skybox being rendered in separate render encoder happens if Depth texture is required to be bounded as shader resource read when rendering transparent meshes.

When this happen URP triggers a copy depth pass that happens after opaque objects has been rendered. I submitted a patch that triggers the copy depth pass after skybox has been rendered instead to fix the issue.

---
### Testing status
local, iOS device, Yamato

---
### Comments to reviewers
Separate skybox renderpass:
SeparateSkyboxRenderPass (1)

### DIFF
--- a/TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/200_DepthDeferred.unity
+++ b/TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/200_DepthDeferred.unity
@@ -184,6 +184,10 @@ PrefabInstance:
       propertyPath: m_RendererIndex
       value: 2
       objectReference: {fileID: 0}
+    - target: {fileID: 1102861101, guid: 144d97346f1c24d4ebabd0eec382b95d, type: 3}
+      propertyPath: XRCompatible
+      value: 0
+      objectReference: {fileID: 0}
     - target: {fileID: 2105869044706732872, guid: 144d97346f1c24d4ebabd0eec382b95d,
         type: 3}
       propertyPath: m_RootOrder
--- a/TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/200_DepthForward.unity
+++ b/TestProjects/UniversalGraphicsTest_Lighting/Assets/Scenes/200_DepthForward.unity
@@ -139,6 +139,10 @@ PrefabInstance:
       propertyPath: m_RendererIndex
       value: 2
       objectReference: {fileID: 0}
+    - target: {fileID: 1102861101, guid: 144d97346f1c24d4ebabd0eec382b95d, type: 3}
+      propertyPath: XRCompatible
+      value: 0
+      objectReference: {fileID: 0}
     - target: {fileID: 2105869044706732872, guid: 144d97346f1c24d4ebabd0eec382b95d,
         type: 3}
       propertyPath: m_RootOrder


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -109,6 +109,7 @@ public Matrix4x4 GetGPUProjectionMatrix(int viewIndex = 0)
         public CameraType cameraType;
         public bool isDefaultViewport;
         public bool isHdrEnabled;
+        public bool postProcessRequiresDepthTexture;
```
phi-lira, line 112:
```suggestion
        public bool postProcessingRequiresDepthTexture;
```

phi-lira: Also needs public API docs. We have task to fix the ones missing atm.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -454,7 +457,29 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
-            m_CopyDepthPass.renderPassEvent = (!requiresDepthTexture && (applyPostProcessing || isSceneViewCamera || isGizmosEnabled)) ? RenderPassEvent.AfterRenderingTransparents : RenderPassEvent.AfterRenderingOpaques;
+            // Aim to have the most optimized render pass event for Depth Copy (The aim is to minimize the number of render passes)
+            if (requiresDepthTexture)
+            {
+                RenderPassEvent copyDepthPassEvent = RenderPassEvent.AfterRenderingTransparents;
+                // CameraData's requiresDepthTexture is based on the URP Pipeline Asset UI checkbox value.
+                if (cameraData.requiresDepthTexture)
+                {
+                    // If user explicitly enable the setting, we force the depth copy to be AfterRenderingSkybox
+                    copyDepthPassEvent = (RenderPassEvent)Mathf.Min((int)copyDepthPassEvent, (int)RenderPassEvent.AfterRenderingSkybox);
```
phi-lira, line 468: Isn't this the same as always setting copyDepthPassEvent to `RenderPassEvent.AfterRenderingSkybox`

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -454,7 +457,29 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+                    // Do depth copy before the render pass that requires depth texture as shader read resource
+                    copyDepthPassEvent = (RenderPassEvent)Mathf.Min((int)copyDepthPassEvent, ((int)renderPassInputs.requiresDepthTextureEarliestEvent) - 1);
+                }
+                m_CopyDepthPass.renderPassEvent = copyDepthPassEvent;
+            }
+            else if (cameraHasPostProcessingWithDepth || isSceneViewCamera || isGizmosEnabled)
+            {
+                // If only post process requires depth texture, we can re-use depth buffer from main geometry pass instead of enqueuing a depth copy pass, but no proper API to do that for now, so resort to depth copy pass for now
+                m_CopyDepthPass.renderPassEvent = RenderPassEvent.AfterRenderingTransparents;
+            }
```
phi-lira, line 482: Nitpick: It feels more readable to me that this logic setups the `inputSummary.requiresDepthTextureEarliestEvent`
 and `copydepthPass.rendererPassEvent = inputSummary.requiresDepthTextureEarlistEvent`

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
phi-lira: This might introduce the regression we talked before that moves the copy depth pass from afterRenderingOpaques -> afterRenderingSkybox.

I discussed with Andre a possible solution for this that would not only fix the copy  depth in between opaque and sky but also other causes of breaking render pass. The problem is the that default event is `AfterRenderingOpaques`, so users can inject other types of passes and break opaque/sky by mistake. I tagged you into a proposal fix.

manuele-bonanno: this can cause users code regression, will remove the first branch

manuele-bonanno: fixed in CS 79b78f0 as discussed

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
manuele-bonanno: fixed in CS 79b78f0

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
manuele-bonanno: after having a second look, I think it makes more sense the way requiresDepthTextureEarliestEvent is initialized now in GetRenderPassInputs.
Since GetRenderPassInputs is the place where all the other inputSummary fields are initialized I think could be confusing to move that logic somewhere else? (also might affect initialization orders?)

phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -111,6 +111,12 @@ public Matrix4x4 GetGPUProjectionMatrix(int viewIndex = 0)
         public bool isHdrEnabled;
         public bool requiresDepthTexture;
         public bool requiresOpaqueTexture;
+
+        /// <summary>
+        /// it is true if the post processing passes require the scene depth texture.
```
ellioman, line 116:
```suggestion
        /// Returns true if post processing passes require depth texture.
```

ellioman (approved):

ValentinHume (approved):
Regression tests in Boat Attacks and UniversalRenderingSamples in editor and player build, looks good !

manuele-bonanno:
all tests have passed except URP Top asset store, unrelated to this PR

ernestasKupciunas (approved):
Looks good. I was using [Winter](https://github.com/ernestasKupciunas/Winter) and [PolyFarm](https://github.com/Unity-Technologies/PolygonFarmBenchmark) projects for regression testing.

Devices under testing:
AppleTV 4k 64GB, tvOS: 14.5

Unity used for testing:
Version: 2021.2.0a19.2746
Revision: trunk c7b813f0d5d5
Built: Thu, 27 May 2021 19:32:01 GMT

simon-engelbrecht-soerensen (approved):
Tested in Boat attack & other depth dependent project - seems to work as expected.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -452,9 +455,32 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
 
             requiresDepthPrepass |= m_DepthPrimingMode == DepthPrimingMode.Forced;
 
+            // If possible try to merge the opaque and skybox passes instead of splitting them when "Depth Texture" is required.
+            // This optimization is temporarily disabled on VR/XR because of a depth copy y-flip issue on DX11.
+#if ENABLE_VR && ENABLE_XR_MODULE
+            m_CopyDepthPass.renderPassEvent = (!requiresDepthTexture && (applyPostProcessing || isSceneViewCamera || isGizmosEnabled)) ? RenderPassEvent.AfterRenderingTransparents : RenderPassEvent.AfterRenderingOpaques;
+#else
```
sandy-carter-unity, line 462: Do you intend to use `if (cameraData.xr.enabled)` here ?

manuele-bonanno: good point, added that check and changed the order of the define checks in https://github.com/Unity-Technologies/Graphics/pull/4628/commits/5f9715edc04a11e98fb76ca2149cf269e112f230

manuele-bonanno:
retested with latest VR fixes, all tests pass except few unrelated ones already failing on master

thomas-zeng:
Hi @manuele-bonanno!
Sorry for dropping the ball, we have looked into the XR issue more and propose the fix in the following PR:
https://github.com/Unity-Technologies/Graphics/pull/5121
I have added you as reviewer.
The issue is not introduced by your PR change, instead it is a issue in the custom render feature code in the test. I have put some explanation of the issue in the PR description. Feel free to reach out to me to discuss further :)

This PR is looking good to me!
I think I am gonna to greenlight the PR if we could do either 
A) merge the proposed PR that fixes the test 
or B) disable the copyDepth_200 test for XR only(this could be done by ticking off XR compatibility checkbox under camera insector-> UniversalGraphicsTestSettings).
Please let me know which plan you would like to go :D

manuele-bonanno:
> This PR is looking good to me!
> I think I am gonna to greenlight the PR if we could do either
> A) merge the proposed PR that fixes the test

Hi @thomas-zeng, thanks a lot for providing a fix!

It looks good to me so I just merged the branch, will merge to latest master and rerun all the tests just in case but looks like we should be good to go

Thanks!

sandy-carter-unity (approved):

thomas-zeng (approved):

PR #3988, opened 2021-03-24
[HDRP] Plugin and versionable subtarget material system

slunity:
# **Please read the [Contributing guide](CONTRIBUTING.md) before making a PR.**

* Read the [Graphics repository & Yamato FAQ](http://go/graphics-yamato-faq).

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Why is this PR needed, what hard problem is it solving/fixing?

Implements a simple versionable plugin material / subtarget system for HDRP subtargets. 
See google doc and JIRA UR-1895:
https://unity3d.atlassian.net/browse/UR-1895

---
### Testing status
Describe what manual/automated tests were performed for this PR

- All automated SG editor tests
- Importing old project, and various old shadergraphs (masternode based, early stack)
- Running our own private plugin material in the system 
-   Artificially changing subtarget or plugin material versions for the plugin

---
### Comments to reviewers

https://github.com/Unity-Technologies/Graphics/pull/3936 has been reverted for now.

See document here:
https://docs.google.com/document/d/1g_zbbhfLAw6SVBA8zmfK6ZdG2_m__ChOK5f1XQfdkNw/

PR #4630, opened 2021-05-25
[HDRP] Clarifying ProjectSettings/UserSettings Editor/Runtime assembly and proxy usage

RSlysz:
### Purpose of this PR
Recent PR have moved HDProjectSettings to Runtime API to use the folder path here.
We splited it to have a part in HDUserSettings.
Finally, we needed it back in Editor and introduce a proxy as we wanted to add enum type in it that are defined only in Editor (#3988).
The state is now mixed, one part in Editor, the other in Runtime. And the introduced proxy broke the Project/User data separation.

This PR is trying to solve the above.
Now both HDProjectSettings and HDUserSettings are in Editor assembly while the proxy is the only one in Runtime assembly.
When trying to access the folder in Runtime assembly, we now directly use the proxy.
Project/User data separation is restored as the proxy is removed from HDUserProject.

Also, instance usage in HDProjectSettings have been cleaned, ensuring we will create/load what needed when calling recently added methods.

Serialization have not changed.

---
### Testing status
Manually tested:
- opened the wizard and check that we still can toggle without error the Show at Start.

---


slunity (approved):

TomasKiniulis (approved):
Looks good. Checked upgrading template, tweaking global settings creating new project through wizard. Did not encounter any pr specific issues.

CertifiedRice (approved):

sebastienlagarde (approved):

PR #4632, opened 2021-05-25
Fix draw procedural invalid pass idx 1 on first template load

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1332630/ 

Boy howdy was this a wild ride :D 

Given the generic nature of the error I had to try comment out all DrawProcedural in HDRP and then  uncomment them to find the right one. Eventually I found it to be TAA. 

This was half of the problem as pass index 1 is defined there.  

However turns out the problem is that with async shader compilation, the pass was not compiled yet but the pass 0 is, so the draw call goes on. This is why the error only triggered once and only with fresh library.  The other half of the reasons why this happens on layout change is because it makes the game view visible for the first time, whereas default layout starts with scene view that does **not** use TAA.

This fixes the issue as it runs the second pass only if needed. 

I reported the behaviour to see if it is expected at all to start with. 

**What did I test:** the repro case.


TomasKiniulis (approved):
Looks good! Just in case checked that:
- Creating new Template Project and changing layout to 2 by 3 split does not cause the error
- Clearing library and reopening SampleScene trying all splits is not affected
- Creating empty project through wizard and changing layout works

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -1533,7 +1536,12 @@ TextureHandle DoTemporalAntialiasing(RenderGraph renderGraph, HDCamera hdCamera,
                         }
 
                         ctx.cmd.DrawProcedural(Matrix4x4.identity, data.temporalAAMaterial, 0, MeshTopology.Triangles, 3, 1, mpb);
-                        ctx.cmd.DrawProcedural(Matrix4x4.identity, data.temporalAAMaterial, 1, MeshTopology.Triangles, 3, 1, mpb);
+#if UNITY_EDITOR
+                        if (ShaderUtil.IsPassCompiled(data.temporalAAMaterial, 1))
```
sebastienlagarde, line 1540: note: instead of this we could add #pragma editor_sync_compilation in the TAA shader, guess it would be cleaner

CertifiedRice (approved):

FrancescoC-unity:
Oops pushed my other git account :-) But last commit is indeed from me.

sebastienlagarde (approved):

PR #4637, opened 2021-05-25
Adding "Allow Material Override" option to BuiltIn Targets

joshua-davis:

### Checklist for PR maker

---
### Purpose of this PR
The BuiltIn target was implemented with material override, but this was always true and couldn't be controlled via the shader graph. This PR is making the built-in target match URP by adding an option to shader graph to specify if the surface options should be controlled by the material or by the target on the shader graph.

---
### Testing status
- [x] For Material Override true (for both lit and unlit):
    - [x] Tested that assigning a new Shader to a Material will reset keywords properly
    - [x] Tested that newly created Materials have their keywords set up properly even if the inspector is not visible or run.
    - [x] Tested combinations of different Target settings, to see that they modify the default properties in the generated shader, and will populate newly created Materials with those defaults.
    - [x] Tested all cull mode settings (front/back/both) on Material
    - [x] Tested setting Opaque/Transparent, and all blend modes on Material
    - [x] Tested enabling/disabling alpha clipping on Material
    - [x] Tested Depth Write and Depth Test settings on Material
- [x] For Material Override false (for both lit and unlit):
    - [x] Tested that assigning a new Shader to a Material will reset keywords properly
    - [x] Tested that newly created Materials have their keywords set up properly even if the inspector is not visible or run.
    - [x] Tested combinations of different Target settings, to see that they modify the default properties in the generated shader, and will populate newly created Materials with those defaults.
    - [x] Tested all cull mode settings (front/back/both) on Target
    - [x] Tested setting Opaque/Transparent, and all blend modes on Target
    - [x] Tested enabling/disabling alpha clipping on Target
    - [x] Tested Depth Write and Depth Test settings on Target
- [x] Test swapping material's shader graph to another shader graph:
    - [x] Swap from material override false to material override true
    - [x] Swap from material override true  to material override false
- [x] Tested pre-existing BuiltIn graphs get upgraded correctly (as material override true)
- [x] Verified that shadergraph master previews show up correctly with BuiltIn Lit/Unlit subtarget, and selecting various combinations of default surface settings on them

Additional tests for extra bugs found:
- [x] Set an unlit shader to transparent. Set the alpha mode to "premultiply". Set the target back to opaque. This should properly compile and draw (used to be a shader error).
- [x] The "Disabled" mode for ZTest is now gone
  - [x] Setting all of the ZTest modes in the target still produce the correct value in the shader
  - [x] Setting all of the ZTest modes on the material when in "Allow Material Override" produce the correct values (verified both from rendering and inspecting the material with the inspector's debug mode).

---
### Comments to reviewers
In the added test project, a duplicate shader graph has to be added due to an existing bug not related to the built-in target: https://fogbugz.unity3d.com/f/cases/1328077/
A few addition bugs found:
 - Additive blend mode for material override got broken but the graphics tests weren't sensitive enough to catch it before. Updated the sensitivity on the tests
 - A shader compiler could happen in unlit if the material was opaque but the alpha mode was pre-multiply (depending on the last state it was set to).
 - The ZTest mode exposed "Disabled" which isn't valid inside a shader.
 - BuiltInUnlitSubtarget didn't inherit from the proper base class. This fixes a case where changing shaders on a material when "Allow Material Override" was true wasn't working. This is because no metadata was saved so the inspector didn't know what the inspector to use to update material keywords.

### DIFF
--- a/com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/Editor/ShaderGUI/BaseShaderGUI.cs
+++ b/com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/Editor/ShaderGUI/BaseShaderGUI.cs
@@ -35,6 +35,8 @@ namespace UnityEditor.Rendering.BuiltIn.ShaderGraph
             public static readonly string[] blendModeNames = Enum.GetNames(typeof(BlendMode));
             public static readonly string[] renderFaceNames = Enum.GetNames(typeof(RenderFace));
             public static readonly string[] zwriteNames = Enum.GetNames(typeof(UnityEditor.Rendering.BuiltIn.ShaderGraph.ZWriteControl));
+            // need to skip the first entry for ztest (ZTestMode.Disabled is not a valid value)
+            public static readonly int[] ztestValues = (int[])Enum.GetValues(typeof(UnityEditor.Rendering.BuiltIn.ShaderGraph.ZTestMode));
             public static readonly string[] ztestNames = Enum.GetNames(typeof(UnityEditor.Rendering.BuiltIn.ShaderGraph.ZTestMode));
 
             public static readonly GUIContent surfaceType = EditorGUIUtility.TrTextContent("Surface Type",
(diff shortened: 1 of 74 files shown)


jessebarker (approved):
LGTM

com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -179,7 +181,7 @@ public static void SetupSurface(Material material)
                         else if (blendMode == BlendMode.Premultiply)
                             SetBlendMode(material, UnityEngine.Rendering.BlendMode.One, UnityEngine.Rendering.BlendMode.OneMinusSrcAlpha);
                         else if (blendMode == BlendMode.Additive)
-                            SetBlendMode(material, UnityEngine.Rendering.BlendMode.One, UnityEngine.Rendering.BlendMode.One);
+                            SetBlendMode(material, UnityEngine.Rendering.BlendMode.SrcAlpha, UnityEngine.Rendering.BlendMode.One);
```
cdxntchou, line 184: Was this found to be incorrect before?

cdxntchou (approved):

GrantLamb-Unity (approved):
Excellent work on testing, finding bugs, and following them up with fixes.

PR #4638, opened 2021-05-25
Fix Jittered Project Matrix Infinite Far Clip Plane 

johnpars:
### Purpose of this PR

Partially fix: https://fogbugz.unity3d.com/f/cases/1336317/

The fogbugz noted two issues, both related to high far clip plane values:
**A)** No objects are rendered for the main camera at high far clip plane values _and with TAA enabled_ (jittered projection). 
**B)** No objects are rendered into shadow maps at high far clip plane values.

This PR fixes Problem A and does not fix Problem B. After some investigation into Problem B, it's my understanding that this behavior is currently "expected" due to the way Unity internally manages culling for shadows, due to its relationship with the main camera view projection matrix.

For Problem A, the issue is due to jittered projection matrix construction. The jittered matrix is constructed from the decomposed projection matrix into the frustum planes. For some reason, for very high far clip plane values, the decomposition evaluates to infinity. This PR addresses this scenario and reconstructs the far plane in that case, but it's still probably worth investigating why this user facing public API (`Matrix4x4.decomposeProjection`) failed.

---
### Testing status
I manually confirmed that when jittered projection is enabled (TAA, DLSS), objects continue to render at extremely high far clip plane values. 

---
### Comments to reviewers
~


sebastienlagarde (approved):

FrancescoC-unity (approved):
Interesting, fix is ok HDRP side, but worth pinging someone for the exposed API.

TomasKiniulis (approved):
Looks good. Objects are rendered now correctly except the shadows with large far clipping planes.

What's tested:
- All Anti-Aliasing modes work correctly with large far clipping plane values
- Multiple cameras in game view and scene view
- Edit mode and play mode

https://user-images.githubusercontent.com/50582134/119789171-81385500-bedb-11eb-8719-8fdad399ae3a.mp4

PR #4640, opened 2021-05-26
Fixed reflection probes being injected into the ray tracing light cluster even if not baked (case 1329083).

anisunity:
https://fogbugz.unity3d.com/f/cases/1329083/
Note that i didn't try to repro the bug, simply did the fix that was suggested by @IonutNedelcuUnity 

Testing status
DXR tests are green locally.
Note that I am still getting occasional crashes, but i don't think the ones i get are related to reflection probes


TomasKiniulis (approved):
Tested repro project switching through listed scenes, no crashes occur anymore.

https://user-images.githubusercontent.com/50582134/120492012-63c72780-c3c2-11eb-9989-b8ab9beedda2.mp4

sebastienlagarde (approved):

PR #4641, opened 2021-05-26
[HDRP] Fix AxF debug output in certain configurations

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1333780/
We run out of samplers in the AxF material when debug dislplay was enabled (this includes matcap and AOV output). There are two reasosn for this:
- In debug an extra sampler is used for mat cap
- The auto-generated debug code can visualize all texture properties so nothing is stripped out (unlike the non-debug case)

This problem is exagerated when high (or very high) shadow quality is used in the HDRP Asset, because the shadow filtering can also use a few samplers.

A reasonable solution for this problem is to re-use the sampler state of the color map for some other maps too (only for debug), in order to stay below the limit.

---
### Testing status
The debug output works as expected now in the repro project:
image
I have also tried a few material settings that are expected to use more samplers, like enabling decals and transparency.

---
### Comments to reviewers
Interestingly, we hit this limitaion even in Vulkan and DX12, I guess due to the shader compilation workflow in Unity.


slunity:
Yeah good choice of maps too. Unfortunately we can't do better for now until it's made into a target since the user is actually free to configure each texture (and thus sampling state) as he sees fit. The only side effects I can see is the AOV API.

slunity (approved):

sebastienlagarde (approved):

PR #4643, opened 2021-05-26
Fix IES Importer related to new Core API

skhiat:
Fix:
https://fogbugz.unity3d.com/f/cases/1336946/

Test:
Import any IES, no more error


sebastienlagarde (approved):

sebastienlagarde:
@soufianekhiat can you re-enable the 3 IES test that have been disabled? It is in the filter asset of the HDRP test project, thanks

PR #4645, opened 2021-05-26
[Yamato] Run only standalone jobs on PRs

sophiaaar:
### Purpose of this PR
Enable only Win DX11 Standalone jobs to run on PRs

---
### Testing status
n/a

---
### Comments to reviewers
slowly re-enabling automatic triggers , hence only doing one platform

### DIFF
--- a/.yamato/all-universal_split.yml
+++ b/.yamato/all-universal_split.yml
@@ -29,6 +29,8 @@ PR_Universal_Split_trunk:
          rerun: on_new_revision
       -  path: .yamato/all-urpupdate.yml#PR_URPUpdate_trunk
          rerun: on_new_revision
+    triggers:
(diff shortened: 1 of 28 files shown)


.yamato/all-universal_split.yml
```diff
@@ -29,6 +29,8 @@ PR_Universal_Split_trunk:
          rerun: on_new_revision
       -  path: .yamato/all-urpupdate.yml#PR_URPUpdate_trunk
          rerun: on_new_revision
+    triggers:
```
theo-pnv, line 32: Do we want to reenable the urp jobs as well? Doesn't the team need to fix the build times?

theo-pnv (approved):

.yamato/all-universal_split.yml
sophiaaar: i thought itd be okay since it was just one from each project, but i will double-check with tim

PR #4648, opened 2021-05-26
Fix SSR accumulation white flash

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1327036/ 

The issue is that for the way SSR accumulation works the color pyramid needs to be valid for more than a frame due to some feedback issues. 

This fixes the issue by clearing the history until we have proper color pyramid valid. 

**What did I test**: the repro project.


skhiat (approved):

TomasKiniulis (approved):
Looks good. Checked SSR with:
- PBR Accumulation, issue is fixed
- Approximation and tweaking other settings remains unaffected
- Multiple Game and Scene views playmode and edit mode

sebastienlagarde (approved):

PR #4649, opened 2021-05-26
Fixes for light anchor

adrien-de-tocqueville:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1336996/

Fixes undo for distance and up direction fields as described in the fogbugz

Also fixes minor issues i found while using the component:
* When adding the component on an existing gameobject, the transform was modified to some dirty values. Now it doesn't
a

* When moving the light using the handles, sometimes the roll was happily spinning. It is no more
b

---
### Testing status

Tested undo/redo with the different properties, tested that changing up direction gives the same results as before


skhiat (approved):

iM0ve (approved):
Looks good, I quickly verified the original issue. Undo now works properly

https://user-images.githubusercontent.com/36502659/119654411-1c282500-be31-11eb-878e-3a49cbe94d63.mp4

CertifiedRice (approved):

sebastienlagarde (approved):

PR #4653, opened 2021-05-26
Fix for discrepancies in saturation and intensity between screen space refraction and probe refraction

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1325298/ 

The probe version had an additional multiply by the transmittance, it was in https://github.com/Unity-Technologies/Graphics/pull/4653/files#diff-c0279e9ca5f2c9c431c0e83cefb9445734a2f0cbf08e2e3ad5208edb96ff70d9L2004 and  https://github.com/Unity-Technologies/Graphics/pull/4653/files#diff-c0279e9ca5f2c9c431c0e83cefb9445734a2f0cbf08e2e3ad5208edb96ff70d9L2017 (removed) 

This fixes the most glaring discrepancies. There are still some but are due to different source content and convolution methods to get the output for a given roughness. 

Before: 

https://user-images.githubusercontent.com/43168857/119658858-e46bae00-be2d-11eb-9a94-a67fcb0b6879.mp4

After:

https://user-images.githubusercontent.com/43168857/119658837-dfa6fa00-be2d-11eb-976a-cf34487a42db.mp4

I think this will change a lot of screenshots, but given I understood yamato is not in a good state now I didn't re-run. We'd need to to not have red tests.  

**What did I test**: repro and can see what I did in video.


sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -2014,7 +2014,7 @@ IndirectLighting EvaluateBSDF_Env(  LightLoopContext lightLoopContext,
         lighting.specularReflected = envLighting;
 #if HAS_REFRACTION
     else
-        lighting.specularTransmitted = envLighting * preLightData.transparentTransmittance;
+        lighting.specularTransmitted = envLighting;
```
sebastienlagarde, line 2017: can you check we don't have the same issue in stacklit (iirc stacklit support refraction?)

CertifiedRice (approved):

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
FrancescoC-unity: Nope, looks like it is not supported

PR #4654, opened 2021-05-26
[Fix] 1338844 - SRP GlobalSettings asset name mismatch

jenniferd-unity:
---
### Purpose of this PR
When a new SRP Global Settings asset is created for the user, a warning may occur "The main object name should match the asset filename.". This warning actually prevents changes to the Global Settings to be saved on file.
image

Fogbugz: (https://fogbugz.unity3d.com/f/cases/1338844)

The solution is to set the appropriate name before asset creation - instead of after.
Callstack:
image

---
### Testing status
On windows editor, i tried on both URP and HDRP to delete the active global settings asset to force the asset recreation.
I tried with and without my changes to ensure the asset no longer appears.


phi-lira (approved):

JMargevics (approved):
I've imported both HDRP and URP to the same project and tried the following steps:

- Check the settings asset
- Delete it, wait till it regenerates
- Select again and check for warnings

Both URP and HDRP work as expected, without any issues ✅

RSlysz (approved):

CertifiedRice (approved):

erikabar (approved):
HDRP QA already covered this, so I'm approving from URP side

PR #4657, opened 2021-05-26
Remove forced clear on mobile

rolandk-unity:
### Checklist for PR maker

---
### Purpose of this PR
Previously, we would always clear each camera target on mobile, even when the camera's target's background was set to "Don't care". This was based on the assumption that a clear comes for free on mobile. However, this assumption is not always true.

- Not all mobile GPUs are tile-based (even though the vast majority are, Tegra being an exception), so the assumption that clearing is always done through a load action on mobile is incorrect
- An empty render pass with a DontCare load action will be optimized away low level code. However, an empty render pass with a Clear load action cannot be optimized away, because we would expect the render target to be cleared afterwards. 

This means that, by forcing a clear on mobile, we sometimes take away an opportunity for low level code to optimize away a renderpass. This currently happens with the 2DRP.

Based on these downsides, I removed the forced clear for mobile when setting a camera's background to "Don't care". Instead, we use a DontCare load action in this case.

This is PR fixes https://fogbugz.unity3d.com/f/cases/1324880/

---
### Testing status
Manual testing done on
- [x] iOS
- [x] Android Vulkan
- [x] Android GLES 3
- [x] Android GLES 2
Automated testing ongoing

---
### Comments to reviewers
See https://unity.slack.com/archives/C89KFUUCT/p1620805082390500


phi-lira (approved):

CertifiedRice (approved):

erikabar (approved):
added mobile QA to reviewers to cover this. approving from URP QA side

aleksandrasdzikia (approved):
Looks good on mobile side. Tested Android and iOS, both works as expected. 
Tested devices: Huawei P40 Lite (9.0.0), Galaxy S21 5G (11.0.0), Huawei P20 (8.1.0), Meizu Pro 6 Plus (7.0.0), OnePlus Nord (10.0.0), iPhone SE (14.4.2), iPad Pro 12.9 (12.2), iPhone 12 pro (14.5)

PR #4660, opened 2021-05-26
Unlock Standalone_XR test config on Yamato

fabien-unity:
---
### Purpose of this PR
- make sure the 4 test projects for URP are available for Standalone_XR on Android (gles3, vk) and Windows (d3d11, d3d12, vk)
- cleanup platforms

---
### Testing status
Confirmed the jobs are now visible on Yamato

---
### Comments to reviewers
A next PR will fix a few failing tests and enable Standalone_XR for nightly runs


fabien-unity:
Note: the failed test is expected. The goal was to make it runnable from Yamato for now.

theo-pnv (approved):

DennisDeRykeUnity (approved):
These metafile changes do not add XR Android tests to SRP ABV, so the test failures should not make nightly status red.  

I see that Fabien is removing the extraneous Mac XR pipelines.  This will remove clutter from the Yamato UI.  I had planned to do this myself in an upcoming PR, so this is one less action item for me. 

Thanks Fabien!

PR #4661, opened 2021-05-26
Align LitTessellation to Lit and LayereLitTessellation to LayeredLit. + make the former fallback on the latter in R

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1335725

- Made Lit and LitTessellation more similar (fix some discrepancies) 
- Make LitTessellation/LayeredLitTessellation  fallback to Lit/LayeredLit with DXR 

**Re: testing** checked with @remi-chapelain and we also discussed so he'll add/update a test when yamato is back in a good shape. 

[Stealing image from @remi-chapelain to show testng] 
image


remi-chapelain (commented):
image

image

remi-chapelain (approved):
Tested while adding the test that the fallback is correct in raytracing ! ✔️

CertifiedRice (approved):

sebastienlagarde (approved):

PR #4662, opened 2021-05-26
[HDRP] Added an error message when we assign the wrong type of texture to the cookie field + reset it to null

alelievr:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1333947/

---
### Testing status
Tested this:

https://user-images.githubusercontent.com/6877923/119690264-cf9d1380-be49-11eb-9694-884359ab66dd.mp4

Note that there is also an error message when a cookie texture is assigned you switch the light type to a type that expects another kind of cookie (different texture dimension).
Personally, I prefer this approach as the user knows why this field has been reset to null when switching the light type. But it's also possible to do it silently, let me know what you think :)


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -767,19 +767,32 @@ static void DrawEmissionContent(SerializedHDLight serialized, Editor owner)
             {
                 serialized.settings.DrawCookie();
 
-                // When directional light use a cookie, it can control the size
-                if (serialized.settings.cookie != null && lightType == HDLightType.Directional)
+                if (serialized.settings.cookie is Texture cookie && cookie != null)
```
RSlysz, line 770: So if no texture is set, the coockie size field disappear?

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -767,19 +767,32 @@ static void DrawEmissionContent(SerializedHDLight serialized, Editor owner)
...
+                            serialized.shapeHeight.floatValue = size.y;
+                        }
+                        EditorGUI.indentLevel--;
+                    }
+                    else if (lightType == HDLightType.Point && cookie.dimension != TextureDimension.Cube)
                     {
-                        serialized.shapeWidth.floatValue = size.x;
-                        serialized.shapeHeight.floatValue = size.y;
+                        Debug.LogError($"The cookie texture '{cookie.name}' isn't compatible with the Point Light type. Only Cube textures are supported.");
+                        serialized.settings.cookieProp.objectReferenceValue = null;
```
RSlysz, line 789: When we clean, it means that user that return to a compatible light will don't have its value anymore. Is it what we want?

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -767,19 +767,32 @@ static void DrawEmissionContent(SerializedHDLight serialized, Editor owner)
...
+                    }
+                    else if (lightType == HDLightType.Spot && cookie.dimension != TextureDimension.Tex2D)
+                    {
+                        Debug.LogError($"The cookie texture '{cookie.name}' isn't compatible with the Spot Light type. Only 2D textures are supported.");
+                        serialized.settings.cookieProp.objectReferenceValue = null;
```
RSlysz, line 794: same

RSlysz (commented):
You should validate the new behaviour with UX

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
alelievr: Yes, see this new UI: https://github.cds.internal.unity3d.com/unity/unity/pull/4021

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
RSlysz: Can't we just hide the field when it is not relevant ?

alelievr: No, it's the same field for Point and Spot lights and because it's a Texture Field, it can accept any kind of texture dimension.
Btw, the behavior on built-in right now is to set to null the cookie when not compatible (like what I did here).
I asked UX for more information :)

RSlysz: In that case, I guess it is ok. Alignment is always better

CertifiedRice (approved):

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
alelievr: UX approved :) (so with this we're aligned with built-in and URP when their PR land)

sebastienlagarde (approved):

PR #4663, opened 2021-05-26
Ignore hybrid duplicated reflection probes during light baking

adrien-de-tocqueville:
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1313264/

When using DOTS live link, it looks like the gameobjects are getting duplicated, probably to compute the generation of entities
But HDRP doesn't differentiate between original objects and duplicated objects, so it actually tries to bake lighting twice.

This is not a complete solution for the issue, because for example when going into playmode, HDRP complains that there are multiple directinoal lights casting shadows whereas there is only one in the scene. But i don't know what the proper fix should be for that case

---
### Testing status

Tested to bake lighting on the repro project

---
### Comments to reviewers

@sebastienlagarde i don't know who should be reviewing the DOTS/Hybrid related PRs ?


CertifiedRice (approved):

TomasKiniulis (approved):
Thanks for the fix! Checked probes with multiple Reflection Probes baking lighting or baking individual probes. Works correctly now

sebastienlagarde (approved):

PR #4665, opened 2021-05-26
Added two toggles to control occluder rejection and receiver rejection for the ray traced ambient occlusion (case 1330168).

anisunity:
https://fogbugz.unity3d.com/f/cases/1330168/

Testing status
Pair tested with QA
DXR tests are green locally


CertifiedRice (approved):

remi-chapelain:
Added QA for awareness :)

anisunity:
Added the equivalent for RTGI

remi-chapelain (changes requested):
The fix allows users to be able to choose between rejecting history when the occluder or receiver object is moving based on what you prefer depending on the context (more noise or more ghosting). This works great and allows more flexibility.

image
 
One concern is that this does not work in the scene view. 
This is not something introduced by the PR, before this, the latest change to reject history based on the object moving was also not reflected in scene view (I checked a few commit before). 
This is probably something we would want to be reflected in scene view to avoid confusion.

sebastienlagarde:
Need update on this PR

remi-chapelain (approved):
The fact that it doesn't work in the scene view is due to the fact that we only have camera motion vectors (and not from the objects), so it's by design. 

Approving then. ✔️

sebastienlagarde (approved):

PR #4670, opened 2021-05-27
[URP][Converter Frame Work] Added priority to our converters

martint-unity:
---
### Purpose of this PR
Adding a priority system to the converters. 
This is needed so that we can have certain converters run before other converters.
There is also a few UI tweaks. Like paddings and such. Nothing that breaks. :)

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
+++ b/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
@@ -11,7 +11,7 @@ namespace UnityEditor.Rendering.Universal
     {
         public override string name => "Material Upgrade";
         public override string info => "This will upgrade your materials.";
-        public override int priority => -1000;
+        public override int priority => - 1000;
         public override Type container => typeof(BuiltInToURPConverterContainer);
 
         List<string> m_AssetsToConvert = new List<string>();


com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
```diff
@@ -120,18 +121,25 @@ void OnEnable()
...
             {
                 // Iterate over the converters
-                RenderPipelineConverter conv = (RenderPipelineConverter)Activator.CreateInstance(converters[i]);
+                RenderPipelineConverter conv = (RenderPipelineConverter)Activator.CreateInstance(converterList[i]);
                 m_CoreConvertersList.Add(conv);
+            }
+
+            // this need to be sorted by Priority property
+            m_CoreConvertersList = m_CoreConvertersList
+                .OrderBy(o => o.GetType().GetProperty("priority").GetValue(o, null)).ToList();
```
svens-unity, line 135: Perhaps we can get rid of the reflection here and on line 129 by using an interface.

com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
```diff
@@ -120,18 +121,25 @@ void OnEnable()
 
             // This is the drop down choices.
             m_ConverterContainers = TypeCache.GetTypesDerivedFrom<RenderPipelineConverterContainer>();
```
svens-unity, line 123: I think this can just be a local variable in the gui method now?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
```diff
@@ -11,6 +11,7 @@ internal sealed class UniversalRenderPipelineMaterialUpgrader : RenderPipelineCo
     {
         public override string name => "Material Upgrade";
         public override string info => "This will upgrade your materials.";
+        public override int priority => - 1000;
```
svens-unity, line 14:
```suggestion
        public override int priority => -1000;
```

ellioman (approved):

svens-unity (approved):
How confident are we that we need this? Just wondering if we should wait until we have a proven need for this?

com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
martint-unity: done

phi-lira (approved):

erikabar (approved):

PR #4676, opened 2021-05-27
Fixed gizmos no longer allocate memory in game view

lukaschod:
### Purpose of this PR
Fixed gizmos no longer allocate memory in game view. [case 1328852]

---
### Testing status
Tested if it fixes the case 1328852. As this is UI-likeish fix I think we can skip full ABV.
Also we do not expose gizmo enable/disable to script so I can not think automation for this case.


svens-unity (approved):

phi-lira (approved):

phi-lira:
I agree we can skip ABV. Graphics tests will not catch any issue here.

erikabar (approved):
I think no additional testing is required because Lukas already tested it

PR #4686, opened 2021-05-27
Fixed loading all materials from project when saving a ShaderGraph

adrien-de-tocqueville:
### Purpose of this PR
When checking that changing active pipeline was correclty calling material setup code, i noticed a new target for builtin was added and didn't benefit from the changes of PR #4508

Changes contain:
* Adding the `ValidateMaterial` callback to the builtin GUI, that gets called when the pipeline is changed, the material is created, and a modification is done in the inspector
* Disabling the code that loads all materials from the project because 1. it's not needed for builtin SG and 2. In case it is needed one day, only the material in memory need reset now (see PR mentionned above for more info)

---
### Testing status

Tested switching from HDRP/URP/builtin and that materials gets setup appropriately
Tested setup is called when modifying value from inspector

---
### Comments to reviewers


jessebarker:
@joshua-davis Does this intersect with your material override PR?

jessebarker (approved):
This looks fine to me, but I'd like to hear from Josh to make sure the PR he has in flight doesn't already have a version of these changes.

joshua-davis (approved):
Looks good and won't conflict with the upcoming "Allow Material Override" PR for built-in

sebastienlagarde (approved):

PR #4688, opened 2021-05-27
Fixed a memory leak related to not disposing of the RTAS at the end HDRP's lifecycle.

anisunity:
@IonutNedelcuUnity noticed that the Ray Tracing acceleration structures were not properly freed after couple HDRP creation/destructions. This fixes it.

Testing status
HDRP tests are still green.


INedelcu (approved):

sebastienlagarde (approved):

sebastienlagarde:
@QA: merging as there is nothing to check but added for awareness

anisunity:
@sebastienlagarde yes it needs a backport

PR #4693, opened 2021-05-27
Misc APV fixes

FrancescoC-unity:
Few APV fixes: 

- Serialize the APV asset as binary -> decrease disk space usage + faster loading. 
- Make sure we end up in a clean state when the lightmapper fails in baking probes, early out and avoid piling up of requests. 
- Make sure that if baking a scene not in the build settings it still works (this will need to be reverted when we have baking sets).


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
```diff
@@ -5,6 +5,7 @@
 
 namespace UnityEngine.Experimental.Rendering
 {
+    [PreferBinarySerialization]
```
alelievr, line 8: Disclaimer about that: Don't expect too much because by default all unity projects are using the serialization mode "Force text" which bypasses this attribute. Son unless someone changes the setting, there is no effect.

alelievr (approved):

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
FrancescoC-unity: At least our template doesn't, I noticed on my project cloned from it that it still serialized as binary, so there is hope :D

sebastienlagarde (approved):

PR #4697, opened 2021-05-27
[HOLD] LightLoop GPU Optimization, VGPR reduction through 16bit accumulation struct

kecho:
---
### Purpose of this PR
Saving 0.3 ms on PS4 / Consoles by reducing 8 vgprs (and going from 3 to 4 occupancy) on the deferred light loop.

* Accumulation buffer pre exposes, and packs to fp16 in software
* Accumulation buffer de-exposes back
* This Increases the total ALU cost, but makes everything go more parallel. So its a win. NOTE: before this PR can be merged I must test with other light rigs.
* ALU cost is negligible, largest threadgroup was around 0.07ms consistently before and after change. 

image

image

---
### Testing status
Have only tested on ps4, on a simple light rig made by @victorsclui 
QA should verify the perf wins on xbox and ps5 if possible.

---
### Comments to reviewers
Leave comments, but more testing is pending if we agree this is a good optimization.


com.unity.render-pipelines.high-definition/Runtime/Material/MaterialEvaluation.hlsl
```diff
@@ -36,6 +36,105 @@ void AccumulateIndirectLighting(IndirectLighting src, inout AggregateLighting ds
...
+    }
+};
+
+struct PackedAggregateLighting
+{
+    float2 m_multipliers;
+    PackedDirectLighting direct;
+    PackedIndirectLighting indirect;
+
+    void init(float exposureMultiplier, float exposureMultiplierInv)
```
sebastienlagarde, line 100: Note: I am not sure that our various platform support function declaration of function in the struct. We need to validate that

FrancescoC-unity: Also nitpicking here, should be capitalized (Init)

FrancescoC-unity:
Sounds good in principle to me, though I'd urge to test with many different amount/types of lights before merging this, though I feel optimistic :D

victorsclui:
looks good to me +1

about turning this feature off adaptively in other deferred lighting variant / forward shader, maybe we can do it later, 
since there may be other effort to reduce register usage that combined with this will lead to bumping up of occupancy.
of course we'll need to remember to do that then : p

FrancescoC-unity:
> looks good to me +1
> 
> about turning this feature off adaptively in other deferred lighting variant / forward shader, maybe we can do it later,
> since there may be other effort to reduce register usage that combined with this will lead to bumping up of occupancy?
> of course we'll need to remember to do that then : )


I would first profile to check how bad it is :-) I am not a fan of pushing regressions if it is trivial not to (just a matter of profiling and an #ifdef ); VGPR optimization is flaky usually, so better to be thorough than not for me.

victorsclui:
yea the perspective about avoiding pushing out regression sounds right
let's check the other deferred lighting variant, and also forward opaque / transparent +1

JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/Material/MaterialEvaluation.hlsl
kecho: I am confident they do since it should be part of the hlsl spec. However it will take a significant effort to verify unless this is deployed in the CI.
I will just take the conservative approach and do it in a glsl friendly way :). I might also put all this in its own utility header to keep it clean. This could also make it easier in the future to move it to hardware 16 bit.

kecho:
So this feature is a negative for variants 7 and 4. THat is, area lights / a lot of ALU we pay a small cost.
This indicates we are mostly ALU bound on these passes (due to the low occupancy). So unless we increase occupancy on thsoe passes, this would be a net loss. This also does not help the register count for those variants.
I am gonna turn it on only for variant 3 deferred mode.

kecho:
This PR is no longer valid. Taking another approach.

PR #4699, opened 2021-05-28
[URP] Enable Depth Priming on Vulkan w/ MSAA

NicoLeyman:
### Purpose of this PR
Depth Priming recently landed in master. One of the limitations was that we couldn't use when MSAA was enabled on Vulkan due to no depth resolve support. Since this support has now landed this limitation is no longer valid.
This limitation wasn't mentioned in the documentation.

---
### Testing status
Manually verified that depth priming now works on Vulkan when MSAA is enabled.
Results in a 3% CPU and 7.5% GPU speedup w/ 8x MSAA.

---


erikabar (approved):
PR was already tested by dev, approving

phi-lira (approved):

manuele-bonanno (approved):

PR #4705, opened 2021-05-28
[URP] Re-enable depth prepass instead of depth copy on GLES3

manuele-bonanno:
---
### Purpose of this PR

This [PR](https://github.com/Unity-Technologies/Graphics/pull/4488#pullrequestreview-661180116) removed the depth prepass which usually happened when MSAA > 1.
This caused motion vectors and decals regressions and a test failures (206 and 230) 

This PR re-enables depth prepass on GLES3 devices as a workaround as an investigation is needed on the device side, since the depth copied seems to be invalid (could be related to a similar VR issue).
GLES3 copy depth fogbugz issue: 1339401

---
### Testing status
Tested locally on Android GLES3

---


HedvigAxelsson (approved):
LGTM

ellioman (approved):

PR #4242, opened 2021-04-20
[2021.2] Control Volume Framework updates

ellioman:
# Purpose of this PR
This PR is adding the option to disable volume framework updates in the Camera Inspector and to the URP Asset. It also adds public API on the Camera for users to update the volume or to change the setting.

### Editor changes
* **Camera Inspector**: Added a Update Mode dropdown
* **URP Asset**: Added a Update Mode dropdown under `Advanced`

#### Public
* **VolumeManager**: Made the stack setter public
* **VolumeManager**: Added a m_DefaultStack variable and cache the original stack in that.
* **VolumeManager**: Added a public `ResetMainStack()` function.
* **Camera**: Added a public function `GetVolumeFrameworkUpdateMode()`
* **Camera**: Added a public function `SetVolumeFrameworkUpdateMode(VolumeFrameworkUpdateMode mode)`
* **Camera**: Added a public function `UpdateVolumeStack()`
* **Camera**: Added a public function `UpdateVolumeStack(UniversalAdditionalCameraData cameraData)`
* **AdditionalCameraData**: Added an public property `requiresVolumeFrameworkUpdate `
* **AdditionalCameraData**: Added an public property `volumeStack`
* **UniversalRenderPipelineAsset**: Added a public enum `VolumeFrameworkUpdateMode`
* **UniversalRenderPipelineAsset**: Added a public property `volumeFrameworkUpdateMode`

#### Internal
* **Camera**: Added an internal function `GetVolumeLayerMaskAndTrigger(UniversalAdditionalCameraData cameraData, out LayerMask layerMask, out Transform trigger)`
* **AdditionalCameraData**: Added an internal property `volumeFrameworkUpdateMode`

### Other changes
* **Testing**: Created a new test scene using a global and local volume with three cameras using different masks. This test scene will be enabled in a follow-up PR.
* **Other**: Added some Documentation that were missing for some public properties

#### Before
image
#### After
image

#### Before
image
#### After
image

# Testing
## What has been tested
* Tested locally in the editor (OS X)

## What needs testing
* **Editor**: Test if Volumes are always updated when not in playmode
* **Editor**: Test if Volumes are updated in playmode when EveryFrame is selected on the camera and/or URP Asset
* **Editor**: Test if Volumes are not updated in playmode when ViaScripting is selected on the camera and/or URP Asset
* **Editor/Builds**: Scripting: Camera: Test Getting & Setting VolumeFrameworkUpdateMode on the camera. 
[...]

PR #4708, opened 2021-05-28
[2022.1] Adding test scenes for the Volume Update Feature

ellioman:
# Purpose of this PR
The PR for controlling Volume Updates (#4242) added a test scene but was not enabled for our CI as we wanted to do that in a separate PR. This PR splits that test scene into three separate tests:
- Test EveryFrame
- Test UsePipeline
- Test ViaScripting

The test works by waiting a single frame and then moving the camera a little bit so it is out of the bounds for the local volume before capturing the screen. The EveryFrame and UsePipeline test results should look the same and only be affected by the global volume. The ViaScripting should keep the state done initially which is the settings done in the local volume. 

# Testing status
Tested locally in the editor.

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fenable-160-test/.yamato%252Fall-urp_postpro.yml%2523Nightly_URP_PostPro_trunk/7074845/job/pipeline


erikabar (approved):

phi-lira (approved):

phi-lira:
Fixed conflicts and merged manually to staging branch.

PR #4711, opened 2021-05-28
Fixed support of distortion with MSAA

sebastienlagarde:
### Purpose of this PR
This PR re-enable the support of distortion with MSAA which were disabled. It looks like RenderGraph change allowed the support.

---
### Testing status

Tested in a scene that distortion is working the same way with or without MSAA

---
### Comments to reviewers


sebastienlagarde:
note: I will merge this PR for the SRP2Core cutoff it is pretty safe and yamato is green.

JMargevics (commented):
image
Verified from my side on bugfix branch. Distortion looks good with MSAA enabled. Tested on DX11, DX12 and Vulkan.

PR #4715, opened 2021-05-31
Fix a bug introduced in https://github.com/Unity-Technologies/Graphics/pull/4479 which broke one of post process tests in UniversalGraphicsTest_2D

kennytann:
---
### Purpose of this PR
Fixes a bug in the 2D renderer where post process results are being overwritten
---
### Testing status
Verified post process is working correctly in UniversalGraphics_2D Test

---
### Comments to reviewers
@robin-demoor  Please verify the targets should not have been swapped and it does not affect any of your changes.


robin-demoor (approved):
Yes, the targets should not have been swapped. I am sorry that this got through and thank you for fixing it!

erikabar (approved):

PR #4719, opened 2021-05-31
Volumetric clouds shape offset and aliasing fix

anisunity:
- Fixed some aliasing issues with the volumetric clouds.
- Added slides to control the shape noise offset.

Testing status
The aliasing issues reported by @pierre-unity do not occur anymore.
Tested the new sliders and the position in the UI.
Local cloud tests are still green.


anisunity:
image

sebastienlagarde (approved):

remi-chapelain (approved):
quickly tested that shape offset were behaving properly with different sky parameters values.

✔️

pierre-unity:
https://user-images.githubusercontent.com/39901544/120833158-60b26f80-c559-11eb-9b91-73a3808a26dd.mp4

When overriding the offset, the clouds are jumping slightly.

Can we make sure that they remain in place when the user adds an override for 0,0?

anisunity:
@pierre-unity  you must have an other volume somwhere that overrides it to an othyer value for that to happen

PR #4720, opened 2021-05-31
Fix contact shadow debug views

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1339327/

The load coordinates were wrong and not resilient to change of view. 

**What did I test**: the repro scene, can see in video an exaggerated case that now works (the contact shadows are badly setup just to make sure I can see them easily) 

https://user-images.githubusercontent.com/43168857/120171325-3af83400-c202-11eb-9b6c-e1cb0c726143.mp4


remi-chapelain (approved):
Fix looks straightforward and testing sufficient ! ✔️ 
Kudos for including a video of the test, very much appreciated :) 

The issue only arise when resizing the view so there's no easy way to include this in automation  to prevent it from regressing, which is unlikely anyway !

sebastienlagarde (approved):

PR #4721, opened 2021-05-31
[XPipeline]When the Add Override of the Volume Profile is pressed, only show Volume Components from the current pipeline

alex-vazquez:
### Purpose of this PR
Fix this: https://fogbugz.unity3d.com/f/cases/1336295/

Adding a new attribute `SupportedOn` to specify that a `VolumeComponent` only is supported for a render pipeline, after this is filtered on the `VolumeComponentProvider`.

Also users can specify a `SupportedOn(typeof(HDRenderPipeline), typeof(UniversalRenderPipeline))` to specify on which pipelines are supported and avoid that volume components on a when a custom SRP is in use.

Added a tooltip on the Volume Component header that shows from which pipeline is the volume component, as we allow that volume profiles have different pipelines volume components.

---
### Testing status
- Check that volume profile can exist with multiple volume components of different pipelines
- Check that when no SRP asset is on the Graphics Settings the user can not add a Volume Component

no_srp_in_use

- Check that when URP is set, only URP volume components are visible, the same with HDRP
urp_only
hdrp_only

- Check that the Unit Test are working fine

---

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
+++ b/com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
@@ -31,6 +31,10 @@ namespace UnityEngine.Rendering
         }
     }
 
+    /// <summary>
+    /// This attribute allows you to add commands to the <strong>Add Override</strong> popup menu
+    /// on Volumes and specify for which render pipelines will be supported
+    /// </summary>
     public class VolumeComponentMenuForRenderPipeline : VolumeComponentMenu
     {
         /// <summary>


com.unity.render-pipelines.core/Editor/Volume/VolumeComponentProvider.cs
```diff
@@ -47,35 +48,35 @@ public VolumeComponentProvider(VolumeProfile target, VolumeComponentListEditor t
...
+                        {
+                            path = attrMenu.menu;
+                            break;
+                        }
+                        case HideInInspector attrHide:
+                        case ObsoleteAttribute attrDeprecated:
+                            skipComponent = true;
+                            break;
+                        case SupportedOnAttribute supportedOn:
+                            skipComponent = !supportedOn.pipelineTypes.Contains(currentPipelineType);
```
arttu-peltonen, line 77: So if you don't give an attribute, it will not be visible. That's probably a good choice, since it forces programmer to notice this attribute and use it, meaning we probably avoid more bugs where something only meant for SRP X accidentally ends up visible for SRP Y.

com.unity.render-pipelines.core/Runtime/Common/CoreAttributes.cs
```diff
@@ -2,6 +2,36 @@
...
+        /// <summary>
+        /// The pipeline types that define the support
+        /// </summary>
+        public Type[] pipelineTypes { get; }
+
+        /// <summary>
+        /// Creates a new <seealso cref="SupportedOn"/> instance.
+        /// </summary>
+        /// <param name="pipelineTypes">The list of pipeline types
+        /// create sub-menus.</param>
```
arttu-peltonen, line 20: "create sub-menus" - is this a copypaste bug?

com.unity.render-pipelines.core/Runtime/Common/CoreAttributes.cs
```diff
@@ -2,6 +2,36 @@
...
 namespace UnityEngine.Rendering
 {
+    /// <summary>
+    /// This attribute allows you to add information for a class to be supported on a render pipeline
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class SupportedOnAttribute : Attribute
+    {
+        /// <summary>
+        /// The pipeline types that define the support
```
arttu-peltonen, line 12: Maybe "The list of pipeline types that the target class supports"

arttu-peltonen (approved):
Looks great!

com.unity.render-pipelines.core/Runtime/Common/CoreAttributes.cs
```diff
@@ -2,6 +2,36 @@
 
 namespace UnityEngine.Rendering
 {
+    /// <summary>
+    /// This attribute allows you to add information for a class to be supported on a render pipeline
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class SupportedOnAttribute : Attribute
```
RSlysz, line 9: This remind me that there is this attribute used sparcely that rely on the asset. Perhaps we should rework it to use your introduced attribute instead ?
> [ScriptableRenderPipelineExtension(typeof(HDRenderPipelineAsset))]

It is in trunk. It is based on the type of asset in Graphics Settings. I had totally forgotten I done this some years ago :(
@jenniferd-unity for awareness

com.unity.render-pipelines.core/Tests/Editor/ReflectionUtils.cs
```diff
@@ -8,6 +8,38 @@ namespace UnityEngine.Rendering.Tests
...
+        /// <summary>
+        /// Finds a type by full name
+        /// </summary>
+        /// <param name="name">The full type name with namespace</param>
+        /// <returns>The found type</returns>
+        public static Type FindTypeByName(string name)
+        {
+            var type = AppDomain.CurrentDomain
+                .GetAssemblies()
+                .Reverse()
```
RSlysz, line 20: Why we need to reverse here?

RSlysz (approved):

com.unity.render-pipelines.core/Runtime/Common/CoreAttributes.cs
alex-vazquez: yes, thanks,

com.unity.render-pipelines.core/Runtime/Common/CoreAttributes.cs
alex-vazquez: changed :)

jenniferd-unity (approved):

jenniferd-unity:
That is a great PR! Very nice improvement that would be useful for the future! Thanks @alex-vazquez

com.unity.render-pipelines.core/Tests/Editor/ReflectionUtils.cs
alex-vazquez: no need

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
```diff
@@ -159,5 +160,23 @@ public void TestHandleParameterDecorators()
         }
 
         #endregion
+
+        [Test]
+        public void TestVolumeComponentProvider()
```
svens-unity, line 165: What is the aim of this test? In my understanding it checks if `FilterVolumeComponentTypes` returns an empty dictionary when it receives some VolumeComponets . Would it not also be useful to test if we get the correct result, e.g. get a component when it is supported by the render pipeline and a result is discarded when it is not?

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
```diff
@@ -4,6 +4,7 @@
 using System.Linq;
 using System.Reflection;
 using UnityEditor.Rendering;
+using Type = System.Type;
```
svens-unity, line 7: What other Type is there?

svens-unity (commented):
I have some questions on the approach taken. Creating a `SupportedOn` attribute sounds like it should handle a lot of general broad cases - basically you should have the ability to slap it on anything and magically it will handle SRP compatibility. The solution here on the other hand is extremely focused. It only solves SRP targets for volume components in the UI menu.

If this is a general purpose solution I would expect it to solve `CustomEditor` and `CustomEditorForRenderPipeline` etc. are we ready yet to take this step? 
If this attribute is meant for this specific problem consider renaming or taking a similar approach as `CustomEditor`, e.g. add a `VolumeComponentMenuForRenderPipelineAttribute`. 

On this note, CustomEditorForRenderPipeline supports inheritance. If an user inherits from URP they will still receive the URP editors (if they do not define any stricter editors). Should we not do the same here. e.g. if you inherit from URP you still receive the URP volumes.

com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed issue displaying a warning of different probe reference volume profiles even when they are equivalent.
 - Fixed missing increment/decrement controls from DebugUIIntField & DebugUIUIntField widget prefabs.
 - Fixed IES Importer related to new API on core.
+- Fixed errors on Adding Overrides to the Volume Profile. Only show Volume Components from the current Pipeline
```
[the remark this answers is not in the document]
alex-vazquez: thanks

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentProvider.cs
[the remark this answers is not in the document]
alex-vazquez: fixed, thanks

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentProvider.cs
```diff
@@ -47,35 +48,35 @@ public VolumeComponentProvider(VolumeProfile target, VolumeComponentListEditor t
...
 
-        public void CreateComponentTree(List<Element> tree)
+        private static Dictionary<string, Type> FilterVolumeComponentTypes(Type[] types, Type currentPipelineType)
         {
-            tree.Add(new GroupElement(0, "Volume Overrides"));
-
-            var types = VolumeManager.instance.baseComponentTypeArray;
-            var rootNode = new PathNode();
-
+            var volumes = new Dictionary<string, Type>();
```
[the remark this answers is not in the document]
alex-vazquez: Caching the valid types in a static dictionary, and avoid parsing the attributes everytime too.

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
alex-vazquez: Changed the test, checking that the `SupportedOn `does not avoid the skipping if there is `HideInInspector `defined.

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
alex-vazquez: No need, the autoinclude from visual added that to me while I was doing tests

com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -33,10 +33,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a large, visible stretch ratio in a LensFlare Image thumbnail.
 - Fixed Undo from script refreshing thumbnail.
 - Fixed cropped thumbnail for Image with non-uniform scale and rotation
+- Fixed errors on Adding Overrides to the Volume Profile. Only show Volume Components from the current Pipeline
 
 ### Added
 - Support for the PlayStation 5 platform has been added.
 - Support for additional properties for Volume Components without custom editor
+- Exposing new attribute SupportedOn
```
svens-unity, line 41: Something along the lines of
```suggestion
- Added VolumeComponentMenuForRenderPipelineAttribute to specify a volume component only for certain RenderPipelines.
```

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentProvider.cs
```diff
@@ -47,35 +48,49 @@ public VolumeComponentProvider(VolumeProfile target, VolumeComponentListEditor t
...
-
-                    if (attr is ObsoleteAttribute attrDeprecated)
-                        skipComponent = true;
+                    switch (attr)
+                    {
+                        case VolumeComponentMenu attrMenu:
+                        {
+                            path = attrMenu.menu;
+                            if (attrMenu is VolumeComponentMenuForRenderPipeline supportedOn)
+                                skipComponent |= !supportedOn.pipelineTypes.Contains(currentPipelineType);
```
svens-unity, line 86: Perhaps this should take into account inheritance similar to CustomEditorForRenderPipeline (CustomEditorAttributes.cs around line 70 in trunk). To handle cases where users inherit form URP but only make minor changes.

com.unity.render-pipelines.high-definition/Runtime/Debug/VolumeDebug.cs
```diff
@@ -140,14 +140,14 @@ static public List<Type> componentTypes
...
         }
 
-        /// <summary>Returns the name of a component from its VolumeComponentMenu.</summary>
+        /// <summary>Returns the name of a component from its VolumeComponentMenuForRenderPipeline.</summary>
         /// <param name="component">A volume component.</param>
         /// <returns>The component display name.</returns>
         static public string ComponentDisplayName(Type component)
         {
-            Attribute attrib = component.GetCustomAttribute(typeof(VolumeComponentMenu), false);
+            Attribute attrib = component.GetCustomAttribute(typeof(VolumeComponentMenuForRenderPipeline), false);
```
svens-unity, line 148: Should this not handle the base attribute?

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentProvider.cs
svens-unity: I dont know if it should. The CustomEditorForRenderPipeline is not taking into account the full inheritance tree.

svens-unity (approved):
Looks good :)

com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
```diff
@@ -30,6 +31,38 @@ public VolumeComponentMenu(string menu)
         }
     }
 
+    public class VolumeComponentMenuForRenderPipeline : VolumeComponentMenu
```
phi-lira, line 34: public API docs missing

phi-lira (approved):
What happens if when you have both pipelines installed and `VolumeComponentForRenderPipeline` with same path?

erikabar (approved):
Verified case 1336295, everything works as expected. Volumes are shown depending on what Render Pipeline I have set in Graphics settings, and if no pipeline asset is set, I can't add any overrides to the Volume.

alex-vazquez:
@phi-lira, now we have this scenario, and URP bloom can not be added. With my changes, only appear the ones for the current pipeline.

PR #4723, opened 2021-05-31
[HDRP][VT] Bugfix multiple resolvers

debaetsd:
### Checklist for PR maker
- [  ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.

---
### Purpose of this PR
This PR fixes an issue that caused VT to stop requesting pages when multiple cameras where (offscreen) rendering.

---
### Testing status
Added additional tests to HDRP_RuntimeTests to verify this exact senario
These also check if VT plays nicely along with the global mip bias (required for DLSS et al)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -244,7 +244,7 @@ public partial class HDRenderPipeline
 #if ENABLE_VIRTUALTEXTURES
                     // Note: This pass rely on availability of vtFeedbackBuffer buffer (i.e it need to be write before we read it here)
                     // We don't write it when doing debug mode, FullScreenDebug mode or path tracer. Thus why this pass is call here.
-                    m_VtBufferManager.Resolve(m_RenderGraph, hdCamera, vtFeedbackBuffer);
+                    hdCamera.ResolveVirtualTextureFeedback(m_RenderGraph, vtFeedbackBuffer);
```
sebastienlagarde, line 247: Question what is happening here with multiple camera? isn't the VTFeedback PR overwrite by each camera result? or there is a process on CPU side in PushFullScreenVTFeedbackDebugTexture which do it, mean that there is some sort of sync point between each camera?

remi-chapelain:
Hey, do you have in mind what could need further testing in the context of the fix ?
Thanks ! :)

FrancescoC-unity (approved):
This looks good to me, though I am curious to what the answer is to seb's question.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
debaetsd: This is exactly the change :)
Before this PR that VTBufferManager was part of the HDRenderPipeline but now it is owned by HDCamera. So now each camera has a (CPU, async readback) buffer rather then 1 shared across all cameras. 

This used to be working so I guess this broke recently (maybe with the introduction of RG?)
I added an explicit test for this in HDRP_RuntimeTests (where the other VT tests also live)

debaetsd:
> Hey, do you have in mind what could need further testing in the context of the fix ?
> Thanks ! :)

I modifies HDRP_RuntimeTests (003-VirtualTexturing) to explicitly test multiple cameras so not a whole lot needs to be tested imho. If the existing VT tests keep working & those new ones also pass, I'm fairly confident about it.

debaetsd:
@kecho This PR will fix the issues you were seeing wrt the global mipbias. The test I added is almost a copy of your previous setup (multiple off-screen VT quads rendered with various global mip bias applied).

sebastienlagarde (approved):

sebastienlagarde:
note: launched yamato with a custom revision that is suppose to be green for check: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fvt-bugfix-resolver/.yamato%252Fall-hdrp.yml%2523Nightly_HDRP_CUSTOM-REVISION/7060530/job/pipeline

JulienIgnace-Unity (approved):

debaetsd:
@sebastienlagarde Well that yamato builder cannot be green as I haven't updated the reference images yet ( not sure what is happening but the builder that should have invalidated them reports everything green.. investigating atm)

TomasKiniulis (approved):
Seems good. Tried creating new VT, checked with multiple Scene/Game views with multiple cameras in Editor and Playmode didn't encounter any PR related issues.

GraphineDieter:
@sebastienlagarde Updated reference images. PR is good to go

sebastienlagarde:
@GraphineDieter  should this PR be backported? can you create the backport pr for 21.1 and 20.3? (It is ok to not updated test for those)

PR #4725, opened 2021-05-31
Fix double sided option moving when toggling it in the material UI

alelievr:
### Purpose of this PR
Fix double sided option moving when toggling it in the material UI: https://fogbugz.unity3d.com/f/cases/1328877/

UX + Artists Validated!

---
### Testing status

Behavior before:
DoubleSided

After:
DoubleSidedFix


sebastienlagarde (approved):

remi-chapelain (approved):
Thanks for including gifs ! 

Simple fix but huge quality of life improvement !!! :)

RSlysz (approved):

RSlysz:
@TheoWong-pixel for awareness. I'm pretty sure this is the reason why we were not "hidding" but instead "disable" things before. Only happens when condition are interleaved.

RSlysz:
Woops sorry wrong manip

sebastienlagarde (approved):

PR #4727, opened 2021-05-31
Fix refraction tile artifacts near reflection probe edges

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1295701/ and https://fogbugz.unity3d.com/f/cases/1326126/

The fix is by changing how the refraction gets the reflection probe used for influence volume. 
Before it would pick the first env light in the tile, but that could change within the same object at borders between reflection probes. 

Now, instead we find the tile at the center of the object and pick the first env light at that tile. 

**What did I test**: repro of second bug mentioned + in same scene moving the spheres a bit and see how it pops changes to different influence volumes as the center enters new influence volumes. 

Before

image

After
image


FrancescoC-unity:
Oh missing aa commit one sec

remi-chapelain (approved):
Glad this is fixed :)

Testing seems sufficient thanks to the detailled description 👍

sebastienlagarde (approved):

PR #4729, opened 2021-05-31
[HDRP] Fix incorrect RTHandle scale in DoF CoC history texture when TAA is enabled

pmavridis:
### Purpose of this PR
DoF breaks when switching resolutions in game view if TAA was also enabled.
To repro:
- Download the test project from: https://fogbugz.unity3d.com/f/cases/1335687/ 
- Switch to gameview
- Change the gameview resoltion from 1080p, to 4k and back to 1080p
Observe that DoF is broken.

The problem is that, if we adjust the max RTHandle size, the CoC history texture is not taking that into account, because it was allocated with an fixed/absolute size. Changing the allocation to use the RTHandle.alloc(scale, ...) method fixes the problem.

---
### Testing status
- I have tested with TAA + DoF enabled 
- I have tested with DLSS + DoF enabled 
In both cases I got the desired results.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -66,6 +66,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added three animation curves to control the density, erosion, and ambient occlusion in the custom submode of the simple controls.
 - Added support for the camera bridge in the graphics compositor
 - Added slides to control the shape noise offset.
+- Added two toggles to control occluder rejection and receiver rejection for the ray traced ambient occlusion (case 1330168).
+- Added the receiver motion rejection toggle to RTGI (case 1330168).
 
 ### Fixed
 - Fixed Intensity Multiplier not affecting realtime global illumination.
@@ -231,6 +233,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed contact shadow debug views not displaying correctly upon resizing of view.
 - Fixed an error when deleting the 3D Texture mask of a local volumetric fog volume (case 1339330).
 - Fixed some aliasing ussues with the volumetric clouds.
+- Fixed reflection probes being injected into the ray tracing light cluster even if not baked (case 1329083).
+- Fixed the double sided option moving when toggling it in the material UI (case 1328877).
 - Fixed incorrect RTHandle scale in DoF when TAA is enabled.
 
 ### Changed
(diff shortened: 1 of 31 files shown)


remi-chapelain (approved):
Trusting what pavlos already tested ! Seems sufficient given the issue ! ✔️

remi-chapelain:
Just to clarify, this does not fix the issue linked. (i.e discrepencies between half and full rez in DoF)

kecho:
Good thing you tested DLSS! Means it should work for regular DRS. Approved!

kecho:
Could you test with Hardware and Software DRS (non DLSS) on dx12 as well?

kecho (commented):
Could you test with dx12 software and hardware DRS  (with CAS and also DLSS) ?

pmavridis:
> Could you test with dx12 software and hardware DRS (with CAS and also DLSS) ?

After doing some additional tests, it looks like using DRS + DLSS + DOF + TAA at the same time is broken, both on master and with this PR. I understand why this happens, I will poke you on Slack to discuss if/how we can fix it.

Every other combination was looking fine in my tests.

pmavridis:
@kecho After my last update, the post-process screen size is available during RG setup and all combinations that I could try seem to work.

@remi-chapelain when you have time, can you also take for a spin and try various combinations of Dynamic Res / DLSS / upscale filters by changing these settings (in case I missed something):
image
(also try toggling TAA  + DoF when various of these settings are on)

kecho (approved):

remi-chapelain (changes requested):
Think I managed to find a repro for the issue while fiddling with parameters

https://user-images.githubusercontent.com/57442369/120488870-725c1180-c3b7-11eb-9ce7-8c16434e5625.mp4

DRS is configured to be force 50% resolution software. 

1. Start with DoF at full on volume
2. Switch DoF to half  on volume
3. Enable DRS on camera
4. Enable DLSS on camera
5. Disable TAA on camera
6. Enable TAA on camera
7. Disable DLSS on camera
8. Disable DRS on camera

I know it's convoluted but still there may be something wrong here that could leak somewhere else ? 

_Tested this on master to be sure and this does not repro, so it might have been introduced in this PR._

pmavridis:
Good catch! Yes, this certainly looks like a bug. Even if the user plays all day with the settings, it should not break...

pmavridis:
@kecho Small change since you last reviewed: now I reset the history buffers when we switch the upscale schedule from after-post to pre-post. I think this is not a problem. This avoid flashes and other issues related to what Remi found.
 
@remi-chapelain I have fixed the issue, can you take it for another spin?

remi-chapelain:
Issue with TAA is fixed but I think I found another unexpected thing with Physically based DoF and DRS. 
But I'm not totally sure because AFAIK, DRS is applied after DoF so maybe the noise "should" be blown out like it after upscaling ? 

https://user-images.githubusercontent.com/57442369/120681618-d8ba6000-c49b-11eb-8640-d07150243338.mp4

remi-chapelain (approved):
Apprently my last issue is [known](https://fogbugz.unity3d.com/f/cases/1340551/) and will be taken care of in another PR. 

Approving then

sebastienlagarde (approved):

PR #4730, opened 2021-05-31
[URP][XR] Add support to generate Skybox shaders form ShaderGraph

maloyer-unity:
This solves an issue where the skybox would be depth clipped in stereo single-pass because the constant buffer for the matrices would not contain the right matrices... 
The Skybox pass currently use the built-in renderer, so we need to make sure that the generated sky shaders will use the built-in stereo matrices constant buffer instead of the one from URP.

To do that, ShaderGraph will add a "#define USING_BUILTIN_STEREO_MATRICES 1" for shaders built with the "Skybox" material type. This will handle all the necessary changes to make built-in stereo matrices work with URP and use them for single-pass stereo shader variation.

This will allow users to create ShaderGraph skyboxes with full support for XR SinglePass Instancing, until we fully port the skybox pass to URP.

---
### Purpose of this PR
Add a way to make skybox-compatible shaders with ShaderGraph so that it will use the right XR single-pass instanced matrix. We need to use the built-in matrices instead of the URP-provided stereo instancing matrices since the built-in skybox renderer will make some important modifications to these (skybox use an inflated world matrix and because of that require an infinite projection matrix)

Fix for https://issuetracker.unity3d.com/product/unity/issues/guid/1328062
And will provide the needed code to also fix https://issuetracker.unity3d.com/product/unity/issues/guid/1299691

---
### Testing status
Created a new shader with ShaderGraph in a new URP project with [a shader like described in the issue](https://forum.unity.com/threads/custom-shadergraph-shader-in-skybox-breaks-single-pass-instanced-rendering.1089931/) which would repro the issue (the skybox is using the wrong stereo matrices and is depth-clipped). With the fix and setting the ShaderGraph to use the new Skybox material, the skybox will appear on Oculus (Tested on Quest2 via Oculus link)

Also tested with : 
 - Editor Vulkan MockHMD
 - Editor DX11 MockHMD/Oculus
 - AndroidPlayer on Quest 2 with Vulkan
 - AndroidPlayer on Quest 2 with OpenGLES

---
### Comments to reviewers
This change will fix the issue until we port the skybox pass to URP and stop using the built-in skybox render pass. (Added XRTODO where we have code that will need to be removed once we have a fully URP skybox)


com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
```diff
@@ -153,7 +153,11 @@ float4 unity_MotionVectorsParams;
 CBUFFER_END
 
 #if defined(USING_STEREO_MATRICES)
+#if defined(USING_BUILTIN_STEREO_MATRICES) && (defined(SHADER_API_D3D11) || defined(SHADER_API_D3D12))
```
thomas-zeng, line 156: I am curious about the DX11 and DX12 logic here. Should we apply the logic for all gfx APIs?

thomas-zeng (commented):
Looks to me the PR will automatically handle the ShaderGraph case. We should consider including the info to our documentation so that users could use this new define for hand-written skybox shaders.

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
maloyer-unity: When I ran my tests, I was only able to repro the problem on DirectX APIs (with Oculus link or MockHMD), Vulkan PC worked normally (with MockHMD). On the Quest it was working, so I targeted that change for DX only. 

I'll re-run some tests later today to confirm everything still works on Quest.

maloyer-unity:
> Looks to me the PR will automatically handle the ShaderGraph case. We should consider including the info to our documentation so that users could use this new define for hand-written skybox shaders.

How do we update the documentation ? Is there a page for URP shaders doc ?

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
thomas-zeng: I am curious about how does it work on Vulkan but no DX11/DX12. Should we look into that? I think in SRP land, we should keep logic the same for all gfx APIs if possible.

thomas-zeng:
> > Looks to me the PR will automatically handle the ShaderGraph case. We should consider including the info to our documentation so that users could use this new define for hand-written skybox shaders.
> 
> How do we update the documentation ? Is there a page for URP shaders doc ?

Hey Marc, yes we have a doc page for URP XR here which we could consider adding the info to: 
https://docs.unity3d.com/Manual/xr-render-pipeline-compatibility.html

thomas-zeng (approved):

maloyer-unity:
There is an issue with some autotests where the shader type is auto-changed to skybox (should not happen).
I'll return the pull request to a draft until I figure out what is happening

erikabar (approved):

maloyer-unity:
Hi, can someone from @Unity-Technologies/universal-rp-team review this PR ?

Also, as far as I know, the last few failing tests are unrelated to this PR.

phi-lira:
Closing this PR as we decided we will fix this issue with removal of Built-in skybox dependency. This is being tracked in our JIRA. Please drop me a message if you need to discuss about it.

PR #4733, opened 2021-06-01
[VFX] Fix soft particles

PaulDemeulenaere:
### Purpose of this PR
TLDR, it fixes two fogbugz :
- https://fogbugz.unity3d.com/f/cases/1309961/
- https://fogbugz.unity3d.com/f/cases/1330697/

More details :
- Fix soft particles orthographic projection : it was initially fixed by @ludovic-theobald in this [PR](https://github.cds.internal.unity3d.com/unity/vfx-graphics/pull/211).
- Fix missing Flip-Y : implicitly integrated in `GetNormalizedScreenSpaceUV`. The issue occurs when there isn't an HDR or Opaque texture as intermediate texture (⚠️This intermediate texture can be implicitly enabled by renderer feature like decal) 
- Fix wrong behavior on OGL : We were loading instead of sampling, when the depth texture isn't available, the default texture can be either black or white depending on `reverse_z`
- Fix missing scale integration : See https://github.com/Unity-Technologies/Graphics/pull/4733/commits/e73718cda2f9976acdd2f664c5a11e8d9d8c65ec I'm not sure if the issue was already there, I didn't test it before, maybe relative to the change from `Load` to `Sample`

---
### Testing status

**1) Orthographic Depth**
*Before*
_ortho_vs_perspective

*After* : Fix Orthographic Depth Sampling
_ortho_fix_after

Also an equivalent test from Ludovic :
eaf1df00-9b79-11eb-88a1-b0ee4bf399f0

**2) Missing Y-Flip**
*Before*
_before_fix_y_flip

⚠️ Decal has the same effect than enabling hdr/opaque texture
_flip_with_decal

*After*
_fix_soft_particle_depth

**3) Fallback when there isn't a depth texture**
*Before* DX11 (no issue actually)
_soft_particle_dx11

*Before* OGL (on the left Shuriken, on the right VFX)
_shuriken_vs_vfx

*After*
_fix_no_depth_ogl

**4) Screen scale**
*Before*
_screen_scale_fail

*After*
_screen_scale_fix

---
### Comments to reviewers
This change will require a manual backport in 10.x.x & 7.x.x
See also this PR : https://github.cds.internal.unity3d.com/unity/vfx-graphics/pull/211

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraUI.Rendering.Drawers.cs
+++ b/com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraUI.Rendering.Drawers.cs
@@ -99,7 +99,8 @@ namespace UnityEditor.Rendering.Universal
                     ),
                 PostProcessingWarningDrawer,
                 CED.Group(
-                    DrawerRenderingClearDepth
+                    DrawerRenderingClearDepth,
+                    DrawerRenderingRenderShadows
                 )
             );
 
@@ -115,7 +116,6 @@ namespace UnityEditor.Rendering.Universal
                 BaseCameraRenderTypeDrawer,
                 OverlayCameraRenderTypeDrawer,
                 CED.Group(
-                    DrawerRenderingRenderShadows,
                     CameraUI.Rendering.Drawer_Rendering_CullingMask,
                     CameraUI.Rendering.Drawer_Rendering_OcclusionCulling
                 )
--- a/com.unity.visualeffectgraph/CHANGELOG.md
+++ b/com.unity.visualeffectgraph/CHANGELOG.md
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix unexpected Spawn context execution ordering
 - Fix incorrect buffer type for strips
 - Enabled an optimization for motion vectors, storing projected positions for vertices instead of the transform matrix
+- VFX Graph operators keep the same width when expanded or collpased so that the button does not change position
 - Fix Soft Particle depth computation when using an orthographic camera [Case 1309961](https://issuetracker.unity3d.com/product/unity/issues/guid/1309961)
 
 ## [11.0.0] - 2020-10-21
--- a/com.unity.visualeffectgraph/Editor/GraphView/Elements/VFXOperatorUI.cs
+++ b/com.unity.visualeffectgraph/Editor/GraphView/Elements/VFXOperatorUI.cs
@@ -208,16 +208,16 @@ namespace UnityEditor.VFX.UI
                 float controlWidth = 50;
                 GetPreferedWidths(ref labelWidth, ref controlWidth);
 
-                float newMinWidth = Mathf.Max(settingsLabelWidth + settingsControlWidth, labelWidth + controlWidth) + 20;
-
-                if (resolvedStyle.minWidth != newMinWidth)
-                {
-                    style.minWidth = newMinWidth;
-                }
-
                 ApplySettingsWidths(settingsLabelWidth, settingsControlWidth);
 
                 ApplyWidths(labelWidth, controlWidth);
+
+                // Do not let the UI reduce in width so that collapse button does not move
+                var newMinWidth = resolvedStyle.width;
+                if (resolvedStyle.minWidth.value < newMinWidth)
+                {
+                    style.minWidth = newMinWidth;
+                }
             }
             else
             {


com.unity.render-pipelines.universal/Runtime/VFXGraph/Shaders/VFXCommon.hlsl
```diff
@@ -116,14 +116,25 @@ float3 GetWorldStereoOffset()
 
 float VFXSampleDepth(float4 posSS)
 {
-    return LoadSceneDepth(uint2(posSS.xy));
+    float2 screenUV = GetNormalizedScreenSpaceUV(posSS.xy);
+
+    //In URP, the depth texture is optionnal and could be 4x4 white texture, Load isn't approriate in that case.
```
ellioman, line 121:
```suggestion
    // In URP, the depth texture is optional and could be 4x4 white texture, Load isn't appropriate in that case.
```

ellioman (approved):

com.unity.render-pipelines.universal/Runtime/VFXGraph/Shaders/VFXCommon.hlsl
PaulDemeulenaere: Thanks, I should enable automatic correction in hlsl file as well.

com.unity.render-pipelines.universal/Runtime/VFXGraph/Shaders/VFXCommon.hlsl
```diff
@@ -116,14 +116,25 @@ float3 GetWorldStereoOffset()
...
+
+    return depth;
 }
 
 float VFXLinearEyeDepth(float depth)
 {
     return LinearEyeDepth(depth, _ZBufferParams);
 }
 
+float VFXLinearEyeDepthOrthographic(float depth)
```
julienf-unity, line 133: Is there any reason why those methods are actually pipeline specific? I guess LinearDepthToEyeDepth is a URP specific implem of what we did in HDRP VFXCommon?

PaulDemeulenaere: Afaik, this helper doesn't exist in HDRP, but indeed, LinearDepthToEyeDepth could be in core and shared between pipelines, we can even use a generic code. I used the centralized function because it was available, I realized afterwards @ludovic-theobald was doing the same kind of fix.
See the implementation in URP : https://github.com/Unity-Technologies/Graphics/blob/b468fa4627d191f06200a0f33df3d9cddee2c3f9/com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesFunctions.hlsl#L328 in URP
...
Actually, it isn't exactly the same equation for `UNITY_REVERSED_Z` :
https://github.com/Unity-Technologies/Graphics/blob/6d552a7a5993e8296cea815dea63ef6ce0bf00e6/com.unity.render-pipelines.high-definition/Runtime/VFXGraph/Shaders/VFXCommon.hlsl#L154

PaulDemeulenaere: Good point, I didn't realize it was equivalent code. Actually, these two function can be store in cross SRP implementation : `VFXLinearEyeDepth` & `VFXLinearEyeDepthOrthographic` (VFXSampleDepth has to be implemented differently in URP and HDRP)

It was specific only for built-in renderpipeline : see [here](https://github.com/Unity-Technologies/Graphics/blob/7.x.x/release/com.unity.visualeffectgraph/Shaders/RenderPipeline/Legacy/VFXCommon.hlsl#L96) (_ZBufferParams is common to URP/HDRP)

I will move this code out from render pipeline specific implementation, it isn't needed anymore ⏳

PaulDemeulenaere: Fix at https://github.com/Unity-Technologies/Graphics/pull/4733/commits/91e182aecbfadb0ce7a813a08435759b2d950e45
I will propose another PR to move https://github.com/Unity-Technologies/Graphics/blob/b468fa4627d191f06200a0f33df3d9cddee2c3f9/com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesFunctions.hlsl#L328 into core.

ludovic-theobald (approved):
Looks good to me!

com.unity.render-pipelines.universal/Runtime/VFXGraph/Shaders/VFXCommon.hlsl
PaulDemeulenaere: Tested in URP :
image

And HDRP :
image

VitaSkruibyte:
Thank you for shiny tests and detailed description, @PaulDemeulenaere! 

Tested:
- All camera settings (ortho + perspective)
- URP asset settings
- Forward decal
- Output settings/inspector
- Standalone
- Graphics APIs (DX11, Vulkan, Metal)

Everything you fixed works like a charm, but observed several issues:
1. Soft particles do not work with camera stacking
Screenshot_1
2. Decal particles have 'sliding' problem when tweaking 'Render Scale' property in URP asset
A1wqlfeilZ

Repro can be found [here](https://drive.google.com/file/d/1vzvM2-91sEEODAuYfJtR0XNzE8gj_F8f/view?usp=sharing)

Is this something that could be fixed in this PR?

ludovic-theobald:
@VitaSkruibyte For the second issue, it's possible that it is fixed in the VFX HDRP Decals PR (at the same time as the sliding artifact was fixed).
I'll test and come back to you.

PaulDemeulenaere:
> Edit : It's not fixed in the other PR.

I have an idea about this issue, forward decal use a simple sampling which require to consider the renderscale while sampling the depth. I will take a look, if it's a simple fix, we could append it to this PR ⏳

PaulDemeulenaere:
Update for @VitaSkruibyte 
1. About `Soft particles do not work with camera stacking`
image
Can you confirm it doesn't work with shuriken as well ? If it's the case, it's probably a common issue within URP and should be reported in URP grab bag.

2. About `Soft particles do not work with camera stacking`, the issue is actually [here ](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.visualeffectgraph/Shaders/ParticleDecals/Pass.template#L112)
It should be
```clipPos.xy = (i.pos.xy * rcp(GetScaledScreenParams().xy)) * 2.0f - 1.0f;```
Instead of
```clipPos.xy = (i.pos.xy / _ScreenParams.xy) * 2.0f - 1.0f;``` (cc @ludovic-theobald)
Can you open a fogbugz (and assign it to me) for this one ? I will create an independent PR for this issue, it will probably require an new interface in VFX (GetScaledScreenParams is specific to URP).
I'm also wondering if there is a specific issue on SubpixelAA & ScreenSpaceSize with RenderScale & URP.

PaulDemeulenaere:
I isolated the fix about forward decal here : https://github.com/Unity-Technologies/Graphics/commit/5a4cb98702822a4fd86f9fbd3dc7e50e0ef1326c (not included ⚠️)

https://user-images.githubusercontent.com/22494706/122190132-d68bd480-ce91-11eb-8cba-ac760c8b4345.mp4

In the end, should I integrate this change in this PR @julienf-unity ?

VitaSkruibyte:
Heya @PaulDemeulenaere! Thank you for the update!

Here's the [bug](https://fogbugz.unity3d.com/f/cases/1343675/) for URP team regarding soft particles
Here's the [bug](https://fogbugz.unity3d.com/f/cases/1343674/) report regarding decals. Also, bumped into another issue where forward decal is not properly projected onto the surface with Overlay Camera, so regarding your comment could this be looked into in the independent PR you've mentioned? (of course, if you have bandwidth for this, otherwise I'll just open a new FB)

NSIixje5Dq

VitaSkruibyte (approved):
All PR related issues were fixed, newly found bugs are not PR related and gonna be fixed in following PR, hence approving.

ellioman (approved):

julienf-unity (approved):

PaulDemeulenaere:
VFX_HDRP on Win_DX11_editmode_mono_Linear is expected to fail 🔴

PaulDemeulenaere:
TODO : backport on 21.1, 10.x.x & 7.x.x

PR #4736, opened 2021-06-01
Fix volumetric fog in planar reflections

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1322845/ 

The issue was due to the Depth Buffer optimization relying on good linear depth being avaialble, that was not the case with Planar reflections due to the oblique projection. 

This PR fixes it as it calculates the linear depth the hard way. 

**What did I test**: the repo scene, can see here 

**Before**
image

**After**
image


remi-chapelain (approved):
Seems like the repro in enough ✔️

sebastienlagarde (approved):

PR #4739, opened 2021-06-01
Fixed SimpleLit smoothness input

Nzollner:
### Purpose of this PR
Link to FogBugz case: https://fogbugz.unity3d.com/f/cases/1318659/
This PR fixes simple lit not using smoothness and changing alpha when "Smoothness Source" is set to "Specular Alpha".
The Smoothness value isn't handled the same way as for the lit shader is SimpleLitInput.hlsl. 
Removed changes to Base Color from smoothness slider in SimpleLitGui.hlsl

The SimpleLit material now uses the same albedo color as the Lit material.

---
### Testing status
Describe what manual/automated tests were performed for this PR.

Manually tested on windows then using Fogbugs given project.

Tested for Each universal test project on windows.

Should have automated testing from nightly.
---
### Comments to reviewers
I change how objects using simple lit is rendered for their specular highlight

### DIFF
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -138,6 +138,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue where shadow artefacts appeared between cascades on Terrain Detail objects.
 - Fixed ShaderGraph materials to select render queue in the same way as handwritten shader materials by default, but allows for a user override for custom behavior. [case 1335795]
 - Fixed sceneview debug mode rendering (case 1211436)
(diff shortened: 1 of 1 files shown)


Nzollner:
Nightly is currently failing on master but this PR needs to pass on nightly because of changes to the SimpleLit shader

erikabar (approved):
it doesn't need additional testing from QA because developer already covered it.

Verasl (changes requested):
Late to look at this sorry, but it looks like by doing this we would need a Material upgrader as before the value was being read from color.a or specular.a the value get's lost as it `_Smoothness` does not currently exist.

Nzollner:
This solution comes with its own set of challenges. I will instead close the Fogbugz case by design since we want the simple lit to be as barebone as possible.

PR #4743, opened 2021-06-01
Add import dialog to add a new diffusion profile to the global settings

FrancescoC-unity:
Addressing https://unity.slack.com/archives/C6Y79CZM0/p1621941147179000 

This PR introduces a dialog on import of a material with a diffusion profile. The  user can then add a diffusion profile automatically to the global settings if there is space for it. 

https://user-images.githubusercontent.com/43168857/120357907-39179900-c306-11eb-83b3-77e6dbe8d7d2.mp4

Added Vic as I am not sure the wording on the dialog box is good enough :D 

Adding @alelievr  as I remember you worked on diffusion profiles? Ignore this review if I am misremembering.

**What did I test**: can see the video above, I also tested that an already added profile doesn't prompt a dialog box and if there is no space in the diffusion profile list similarly no dialog is displayed.


remi-chapelain:
Dumb question, did you try with a SG material ? Does it works the same ? 
meaning material.HasProperty("_DiffusionProfileAsset") will return true aswell ?

FrancescoC-unity:
> Dumb question, did you try with a SG material ? Does it works the same ?
> meaning material.HasProperty("_DiffusionProfileAsset") will return true aswell ?

Not a dumb question at all! 
Tried and works just fine

remi-chapelain (changes requested):
Quickly tested with importing material samples since there's 7 different diff profile there. 

And the only thing that fail is for Eye shader. There's this node that takes 2 diff profile in input that are not included in the batch in material importer. Is there something we can do about this since this is node accesible by everyone in utility SG nodes ? 

image

Also, since there's an upper limit of diffusion profile that can be added (weirdly 15), when your list is nearly full, when importing new materials with new diff profile, the new one gets added and then when the list is full, the rest are just skipped without info, is there a way to have either some sort of pop up telling you that the list is full, or a warning for each diffusion profile skipped and the reason why, for the users to know that something just failed silently ? :)

FrancescoC-unity:
> Quickly tested with importing material samples since there's 7 differents diff profile there.
> 
> And the only thing that fails is for Eye shader. There's this node that takes 2 diff profile in input that are not included in the batch in material importer. Is there something we can do about this since this is node accessible by everyone in utility SG nodes ?
> 
> image
> 
> Also, since there's an upper limit of diffusion profile that can be added (weirdly 15), when your list is nearly full, when importing new materials with new diff profile, the new one gets added and then when the list is full, the rest are just skipped without info, is there a way to have either some sort of pop up telling you that the list is full, or a warning for each diffusion profile skipped and the reason why, for the users to know that something just failed silently ? :)
> 
> Thanks !

Oh didn't know about the Eye diffusion profiles, I can look into that later on today. 

Regarding the silently fail, I thought about that but then if you load a lot of assets, you'll end up with potentially a lot of dialog stopping the process saying the same thing which sounds annoying.

I guess what I can do is have a single dialog saying "Diffusion profile don't fit anymore, anything imported from now on will not be added to settings" ?

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added a slider that controls how much the volumetric clouds erosion value affects the ambient occlusion term.
 - Added three animation curves to control the density, erosion, and ambient occlusion in the custom submode of the simple controls.
 - Added support for the camera bridge in the graphics compositor
+- Added a dialog box on import to add a diffusion profile to the global settings.
```
Vic-Cooper, line 68: Tweaked for clarity: 

Added a dialog box when you import a Material that has a diffusion profile to add the diffusion profile to global settings.

Vic-Cooper (changes requested):
One small adjustment to the changelog.

remi-chapelain:
> I guess what I can do the warning at the end of process in the console.

I am fine with this solution, I agree, it can be hassle when importing huge asset pack !

remi-chapelain (approved):
Issues have been addressed :), approving ✔️

alelievr (approved):

sebastienlagarde (approved):

sebastienlagarde:
note to self: retarget to master once hd/bugfix is merge

sebastienlagarde:
for unknow reasins it break test package: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fdiffusion-profile-import-add/.yamato%252F_projectcontext.yml%2523test_all_project_OSX_trunk/7157649/job/artifacts

FrancescoC-unity:
> for unknow reasins it break test package: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fdiffusion-profile-import-add/.yamato%252F_projectcontext.yml%2523test_all_project_OSX_trunk/7157649/job/artifacts

I'll check!

Vic-Cooper (approved):
Looks good :)

sebastienlagarde:
it works, thanks :)

PR #4745, opened 2021-06-01
[HDRP][Path Tracing] Support for shadow mattes

eturquin:
Added support for Unlit shadow mattes in path tracing, addressing: https://fogbugz.unity3d.com/f/cases/1335487

Before, shader graph Unlit, with shadow mattes enabled, would be rendered as if shadow matte was disabled, in path tracing.
Now, shadow mattes are supported, with or without RGBA shadow tint.

This has been tested on the scene provided in the case mentioned above (with added shadow tint), and produces results matching the rasterized ones.


iM0ve (commented):
I also verified to better understand the issue. Shadow Matte option is now working with Path tracer(shadow is cast upon the red plane), tested fully opaque and transparent shadows. No issues found. But would it be a good idea to add an automated test for this or it's too niche of a case?


https://user-images.githubusercontent.com/36502659/120614820-4f396c80-c460-11eb-80d3-cef6d4f15fb0.mp4

sebastienlagarde:
This PR indeed require an automated test

eturquin:
> This PR indeed require an automated test

I've just added one.

remi-chapelain (approved):
Test is 🟢 on yamato

iM0ve (approved):

PR #4746, opened 2021-06-01
Fix thin resize handles on Blackboard and Graph Inspector

dhsavell:
### Purpose of this PR

This PR fixes https://fogbugz.unity3d.com/f/cases/1329247/. The resize bounds on the Graph Inspector and Blackboard windows were only 1 pixel large, making them difficult to use.

---
### Testing status

Vertical/horizontal resize handles on the Blackboard and Graph Inspector were previously 1 pixel wide. Manual testing should show that resize handles are now 10 pixels on all sides. 

---
### Comments to reviewers


esmelusina (approved):
Doesn't impact product integrity-- the major impact will be on QA, whether it works well with the sliders/handles on those UI elements.

LandonTownsendUnity (approved):
This fix works great. Scrollbars are still accessible and blackboard/inspector is able to be resized even when stretched to the edges of shadergraph, while the scrollbars still look fine. Approving

PR #4748, opened 2021-06-01
[FogBugz 1339083] Fixing exposure issues with clouds and planar probes

kecho:
### Purpose of this PR
Fogbugz: https://fogbugz.unity3d.com/f/cases/1339083/
This PR has a few fixes for issues around volumetric clouds and exposures. The issues are the following:

1) Planar reflection volumetric cloud exposure history was wrong, it was using a pointer to the parents view history. This history was invalid since the parent view would write a new exposure right before rendering. This has been changed now so the probe has its own history, and instead we copy by value the parent's exposure. A new pass has been added for this, which is called  "UpdateParentExposure". This pass clears the current exposure texture to whatever was established by the parent. This effectively preserves the history of the current planar reflection probe.

Before vs After:
image

2) When exposure control was disabled on the reflection probe, the clouds would look very weird in the reflection. This was caused by the internal usage of a "Fast tonemap" function, which squeezed the range of the volumetric clouds. This range helps when we are pre exposued, but if we are not pre exposed we get luminance values that are bigger. This causes some weird floating point behaviour when luminance is very high (we lose tons of precision due to 16 bit storage). For now this PR proposes removing the tonemapping function if exposure control is disabled on the reflection.

Before vs After
image

3) In the editor, if we switch quickly between exposure control on and off in the planar reflection, we can get some NaNs and corruption because the history is not invalidated. Invalidated the history correctly and fixed the problem

4) Fixed feedback loop caused by mismatched exposure coming from the CPU & GPU. The exposure used to sample planar reflection was coming from the CPU which could be a few frame delayed, and the exposure used to render the planar reflection was coming from the GPU which could be ahead. This could cause some crazy flashing /feedback loop in exposure when adaptation time was very small.

---
### Testing status
* Tested on sample scene from test dx11 PC .
* Tested with planar probe with exposure control on and off.
* Tested with exposure modes: Fixed, Auto and using histogram & average metering.


anisunity:
@sebastienlagarde don't merge, this requires some reading/changes

JMargevics (approved):
image

Hello! Tested from my side: 

- With and without exposure control
- DX11, Vulkan (win), Metal. (DX12 has broken planar reflections for me atm)
- Entering playmode
- Switching to different exposure modes

All good now ✅

sebastienlagarde (approved):
seems fine on my side , waiting Anis reply

kecho:
Running Yamato, and waiting for Anis' review.

anisunity (changes requested):
Pretty much good, discussed with @kecho over the phone couple changes required, but adding them here just for tracking purposes.
 1 - Change the multicompile for the fast-tonemap into a scalar with a lerp, the shader is quite large and I would like to avoid increasing the compilation time.
 2 - Add the exposure control history flag to the pre-existing structure, it may be useful for ray tracing effects eventually and it would avoid having yet an other structure into the HDCamera.cs
3- Duplicate the shadow test to have a version without exposure control for coverage

JulienIgnace-Unity (approved):

kecho:
@anisunity all tests and the changes you requested have been added. Gonna now run the CI and get it all updated.

anisunity (approved):
lgtm

PR #4754, opened 2021-06-02
Fixed an issue with normal management for recursive rendering (case 1324082).

anisunity:
Making the normal always point towards the ray origin to fix some inconsitent behaviors.
https://fogbugz.unity3d.com/f/cases/1324082/
Also removed the incidentDirection from the payload was not necessary).

Testing status
All tests are green except 902_hair for some reason (not even related to ray tracing)
The issue reported in the fogbugz doesn't repro anymore.


remi-chapelain (approved):
Fogbogz repro case is fixed and tests are all green on yamato ✔️

sebastienlagarde:
We will not backport this one, afraid backport could be conflciting.

sebastienlagarde (approved):

PR #4759, opened 2021-06-02
- Updated the recursive rendering documentation (case 1338639).

anisunity:
https://fogbugz.unity3d.com/f/cases/1338639/
After a deeper investigation, it is not really possible to have a fallback for this the way I thought. I will mention it in the documentation, the only option right now is to increase the ray length.


com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Recursive-Rendering.md
```diff
@@ -4,6 +4,8 @@ This feature is a replacement pipeline for rendering Meshes in the High Definiti
 
 The smoothness of a Material does not affect the way a ray reflects or refracts, which makes this rendering mode useful for rendering multi-layered transparent GameObjects.
 
+If the ray-traced object is further than the maximal ray length defined in the volume component, the color of the sky is displayed. To solve this, increase the value of the maximal ray length.
```
Vic-Cooper, line 7: HDRP might display the sky color instead of a GameObject that has ray tracing applied. This happens when the GameObject is further away from the Camera than the **Max Ray Length** value set in the volume component. To make the GameObject appear correctly, increase the value of the **Max Ray Length** property.

Vic-Cooper (changes requested):
Suggested a change to the new content :)

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Recursive-Rendering.md
Vic-Cooper: If it's unambiguous that HDRP returns the color of the sky in this case, remove "might".

sebastienlagarde:
waiting for update

anisunity:
@sebastienlagarde review fixes done

sebastienlagarde:
review done, merging

sebastienlagarde (approved):

PR #4761, opened 2021-06-02
[XPipeline] Lazy init + clean up of CoreEditorStyles

alex-vazquez:
---
### Purpose of this PR
Lazy initialization of all the styles in order to be able to obtain the icons from a UIToolkit context.

All the calls to Editor.Styles are restricted to the method `OnGUI`, from UIToolkit window this is not further possible so we defer the initialization of those styles to the first call to obtaining the style, That call must be on OnGUI context.

Use case the development of Converters from @martint-unity

---
### Testing status
Check that the styles are correct, nothing is broken

---
### Comments to reviewers
Needed by @martin-unity3d

### DIFF
--- a/com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
+++ b/com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
@@ -8,20 +8,81 @@ namespace UnityEditor.Rendering
     {
         /// <summary>Speed of additional properties highlight.</summary>
         public static readonly float additionalPropertiesHightLightSpeed = 0.3f;
+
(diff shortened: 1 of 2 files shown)


jenniferd-unity (approved):
I am approving the change because code functionality seems correct but I do not understand use case or why Martin needs it. Could we clarify the PR description maybe?

martint-unity (approved):
LGTM
Need to have a quick QA pass on it.

alex-vazquez:
@Unity-Technologies/gfx-qa-urp could you do a pass on this PR, open some inspectors and windows, and see that everything is working fine

erikabar (approved):
LGTM

PR #4762, opened 2021-06-02
Hdrp/dlss fixes for particles half res

kecho:
### Purpose of this PR

The problem & Reason:
During Dynamic Resolution Scaling, half res transparency passes fall too much behind detail. This causes particles to look very aliased and other transparent geo look wrong. This is because we are downscaling half resolution, which gets rid of more detail.

The solution is to clamp how low we go on resolution (everything relative to the native resolution of course). 

This PR has a few improvements for DRS and low resolution transparency (UX and functionality as well):

1) Showing an info box in the material inspector if low resolution transparency is disabled in the current HDRP asset. This should make it easier for our users to find it.
image

2) Added new checkboxes  / low res transparency min threshold when DRS is enabled. This will let users threshold low res transparency when the current frame scale falls below a certain threshold. Included here also a similar info box to warn the user.
image

3) Documentation update, and CHANGELOG updates.

---
### Testing status
In Windows dx11, just created a new material with low res transparent. Proceeded to test:
* Live enabling and disabling low res transparent from the HDRP asset. Checked that both info boxes update appropiately.
* Set DRS as force resolution percentage, and set many resolutions.
* Set the new threshold at multiple settings and verified that low res transparent was always disabled when the forced percentage fell between a threshold.
* Verified low transparency was running during these tests, and also verified it was disabled by simply looking at the low res result.
* Tested this on CAS and Bilinear, software DRS.

Recommended additional tests for QA (have not tested):
* Test on DLSS dx12 and vulkan.
* Test on hardware DRS
* Test on spaceship demo, with DLSS and verify particles can pop in / out of low res.


com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -544,5 +544,16 @@ public Vector2Int GetLastScaledSize()
         {
             return m_LastScaledSize;
         }
+
+        /// <summary>
+        /// Returns true if the current configuration should allow for low res transparency, returns false otherwise.
+        /// </summary>
+        /// <returns>True if the current configuration should allow for low res transparency, returns false otherwise.</returns>
+        public bool AllowLowResTransparency()
```
FrancescoC-unity, line 552: This is actually problematic as it will pop when going below a certain threshold (with actually dynamic res and not hardcoded dimension).  Say we over around the minimum threshold, it will constantly go between full res and low res, which will be a big jump. 

We should instead probably just clamp the min resolution the low res transparent can reach. So we don't pop, but stop scaling down. 

Say the last accepted low res resolution is 360p (i.e. main res 720p), if the resolution with dynamic res drops below 720 we still use whatever is lower between 360p and the main res. This will still pop at very very low resolutions (less than half the percentage that is allowing the low res), but it should be way less often and jarring (since at that point we are likely an hodgepodge of pixels anyway and therefore the settings are likely to be set so that it won't happen in a shipped product)

kecho: This is doable, but obviously more work :) 
@sebastienlagarde  what are your thoughts? We were just wanting to do some minimal effort here.

FrancescoC-unity: I am just afraid this will be more jarring than the issue is actually trying to fix when using dynamic resolution. 

The pop will be very strong as we suddenly jump between very different resolutions and potentially in a flickering manner (as we hover the threshold).

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/SurfaceOptionUIBlock.cs
```diff
@@ -698,6 +699,14 @@ void SurfaceTypePopup()
                         renderQueueType = HDRenderQueue.ConvertFromTransparentRenderQueue(newRenderQueueTransparentType);
                         renderQueue = HDRenderQueue.ChangeType(renderQueueType, offset: (int)transparentSortPriority.floatValue);
                     }
+                    if (renderQueueTransparentType == HDRenderQueue.TransparentRenderQueue.LowResolution)
+                    {
+                        if (HDRenderPipeline.currentPipeline != null
+                            && !HDRenderPipeline.currentPipeline.currentPlatformRenderPipelineSettings.lowresTransparentSettings.enabled)
+                        {
+                            EditorGUILayout.HelpBox(Styles.lowResTransparencyNotSupportedText, MessageType.Info);
```
FrancescoC-unity, line 707: While I like this, one thing to keep in mind is that the asset used during authoring is not necessarily the asset that is used in the project on a given platform. 

I think this is the reason why we don't do this for similar things (e.g. subsurface scattering is a very similar situation)

FrancescoC-unity: I do like the idea though, worth pinging UX?

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -544,5 +544,26 @@ public Vector2Int GetLastScaledSize()
...
+
+        /// <summary>
+        /// Returns the resolved low res multiplier based on the low transparency threshold settings.
+        /// </summary>
+        /// <param name="targetLowRes"> the target low resolution.
+        ///     If by any chance thresholding is disabled or clamped, the exact same resolution is returned.
+        ///     This allows the caller to directly compare the float result safely with the floating point target resolution.
+        /// </param>
+        /// <returns>Returns the resolved low res multiplier based on the low transparency threshold settings.</returns>
+        public float GetLowResMultiplier(float targetLowRes)
```
FrancescoC-unity, line 556: I actually wonder if we need this on the handler? Not all pipelines will have low res stuff? 

I guess as stop gap I would make the documentation more generic (e.g. "The pipeline can use this to drive the scale for low res transparency if available etc.")

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -544,5 +544,26 @@ public Vector2Int GetLastScaledSize()
         {
             return m_LastScaledSize;
         }
+
+        /// <summary>
+        /// Returns the resolved low res multiplier based on the low transparency threshold settings.
```
FrancescoC-unity, line 549: low res transparency instead of just low

com.unity.render-pipelines.core/Runtime/Common/GlobalDynamicResolutionSettings.cs
```diff
@@ -102,5 +104,11 @@ public static GlobalDynamicResolutionSettings NewDefault() => new GlobalDynamicR
         public bool forceResolution;
         /// <summary>The resolution percentage forced in case forceResolution is set to true.</summary>
         public float forcedPercentage;
+
+        /// <summary>Enable to support lowResTransparencyMinimumThreshold</summary>
+        public bool enableLowResTransparencyThreshold;
```
FrancescoC-unity, line 109: I am actually now wondering if we need the kill switch, maybe just having the threshold is enough? This way 0.0f is equivalent to disable the threshold? Not sure, what do you think? Might make the UI less cumbersome.

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -544,5 +544,26 @@ public Vector2Int GetLastScaledSize()
...
+        ///     If by any chance thresholding is disabled or clamped, the exact same resolution is returned.
+        ///     This allows the caller to directly compare the float result safely with the floating point target resolution.
+        /// </param>
+        /// <returns>Returns the resolved low res multiplier based on the low transparency threshold settings.</returns>
+        public float GetLowResMultiplier(float targetLowRes)
+        {
+            if (!m_Enabled || !m_CachedSettings.enableLowResTransparencyThreshold)
+                return targetLowRes;
+
+            float thresholdPercentage = Math.Min(m_CachedSettings.lowResTransparencyMinimumThreshold / 100.0f, targetLowRes);
```
FrancescoC-unity, line 561: Probably here we want to safe guard for lowResTransparencyMinimumThreshold == 0? Probably just early out if it's the case, otherwise looks like we could return a 0 resolution.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
```diff
@@ -1052,10 +1055,14 @@ void DownsampleDepthForLowResTransparency(RenderGraph renderGraph, HDCamera hdCa
...
             using (var builder = renderGraph.AddRenderPass<DownsampleDepthForLowResPassData>("Downsample Depth Buffer for Low Res Transparency", out var passData, ProfilingSampler.Get(HDProfileId.DownsampleDepth)))
             {
-                // TODO: Add option to switch modes at runtime
-                if (m_Asset.currentPlatformRenderPipelineSettings.lowresTransparentSettings.checkerboardDepthBuffer)
+                if (hdCamera.isLowResScaleHalf)
                 {
                     m_DownsampleDepthMaterial.EnableKeyword("CHECKERBOARD_DOWNSAMPLE");
+                    m_DownsampleDepthMaterial.DisableKeyword("GREATER_THAN_HALF_DOWNSAMPLE");
+                }
+                else if (m_Asset.currentPlatformRenderPipelineSettings.lowresTransparentSettings.checkerboardDepthBuffer)
```
FrancescoC-unity, line 1063: Do we need the check on checkerboardDepthBuffer here? I think it's always true for now and if it wasn't the following is not doing checkerboard anyway.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassForward.hlsl
```diff
@@ -113,7 +113,7 @@ void Frag(PackedVaryingsToPS packedInput,
     FragInputs input = UnpackVaryingsToFragInputs(packedInput);
 
     // We need to readapt the SS position as our screen space positions are for a low res buffer, but we try to access a full res buffer.
-    input.positionSS.xy = _OffScreenRendering > 0 ? (input.positionSS.xy * _OffScreenDownsampleFactor) : input.positionSS.xy;
+    input.positionSS.xy = _OffScreenRendering > 0 ? (uint2)round(input.positionSS.xy * _OffScreenDownsampleFactor) : input.positionSS.xy;
```
FrancescoC-unity, line 116: I think the cast does the round implictly?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -235,6 +235,11 @@ float SampleCameraDepth(float2 uv)
     return LoadCameraDepth(uint2(uv * _ScreenSize.xy));
 }
 
+float4 GatherCameraDepth(float2 uv)
```
FrancescoC-unity, line 238: This doesn't seem to be used anywhere?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/DownsampleDepth.shader
```diff
@@ -60,9 +61,15 @@ Shader "Hidden/HDRP/DownsampleDepth"
         float4 _DstOffset;
 #endif
 
+        float4 _ScaleBias; // x: uv offset x, uv offset y, uv x scale, uv y scale,
+
         void Frag(Varyings input, out float outputDepth : SV_Depth)
         {
             UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
+#if GREATER_THAN_HALF_DOWNSAMPLE
+            float4 depths = GATHER_RED_TEXTURE2D_X(_CameraDepthTexture, s_trilinear_clamp_sampler, input.texcoord * _ScaleBias.xy + _ScaleBias.zw);
```
FrancescoC-unity, line 70: No need for trilinear, this doesn't have mips, linear is enough

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -235,6 +235,11 @@ float SampleCameraDepth(float2 uv)
...
+{
+    return GATHER_RED_TEXTURE2D_X(_CameraDepthTexture, s_trilinear_clamp_sampler, uv * _RTHandleScale.xy);
```
FrancescoC-unity, line 240: If we keep this, the above still applies, we shouldn't trilinear here.

FrancescoC-unity (approved):
Small comments here and there, but looks fine to me.

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: if lowResTransparencyMinimumThreshold the condition at line 563 will catch it and just return the target low res. 
I will add a guard for the input instead. So you cant request a low res below some epsilon

com.unity.render-pipelines.core/Runtime/Common/GlobalDynamicResolutionSettings.cs
kecho: yep killing this

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/SurfaceOptionUIBlock.cs
kecho: Yep, pinged maddalena, she liked it as well

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassForward.hlsl
kecho: im pretty sure that cast is a floor.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/DownsampleDepth.shader
kecho: Done

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
kecho: leftover. Removed.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
kecho: Removing this, unused

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Prepass.cs
kecho: Cleaned this up.

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: Updated doc comment

iM0ve (approved):
**Pair tested with Kleber:**
* As shown in the video below, there was an issue that with DRS that particles were being downscaled more than expected, resulting in them disappearing. In the first half of the video notice when Low res is toggled, some particles are disappearing. With the help of this PR, now we can set the min threshold (50% in the second part of the video) to prevent the half rez particles of being downscaled, resulting in them no longer disappearing
* Tested software and hardware drs modes, different drs scaling settings
* Multiple parts of Spaceship demo project
* Dx11 dx12 and Vulkan

**Demo:**

https://user-images.githubusercontent.com/36502659/121058023-9315d400-c7c8-11eb-89e3-103a0cbac7d8.mp4

**Issues:**
* **FIXED** - sometimes the info box was displayed incorrectly

image

**PR unrelated issues found:**
Crash on Vulkan when toggling Low res (same on master)
https://fogbugz.unity3d.com/f/cases/1341576/
 Visual lighting, blending difference with low res
https://fogbugz.unity3d.com/f/cases/1341567/

sebastienlagarde:
@Vic-Cooper I will merge this one as it change a lot of code, but please report any doc correction here it will be done on another PR.

PR #4766, opened 2021-06-03
Remove orphan metas

theopnv:
Feel free to merge after reviewing, thanks


phi-lira (approved):
It would be nice to have githooks for these.

theo-pnv:
it is on the sdet's roadmap to have both client side (git hooks) and server side (yamato jobs) to detect these. Should come by end of summer.
Thanks for the merge.

PR #4769, opened 2021-06-03
Urp/depth priming tests

NicoLeyman:
### Purpose of this PR
The recently landed PR adding Depth Priming didn't contain any automated test coverage. This PR aims to correct that by adding a new graphics test.
Since this test ran into an issue related to the recent swapbuffer changes this also contains a fix for that to allow the test to succeed.

---
### Testing status
New graphics test added. Ref images are being added.

image

The test shows a set of boxes rendered with a camera stack, splitscreen rendering. Since Depth Priming can have fallout I opted to use some of the existing materials on the scene objects. SSAO and transparent materials are included and definitely need to stay part of the test.

---
### Comments to reviewers
Code changes provided by Robin.

### DIFF
--- a/TestProjects/UniversalGraphicsTest_PostPro/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/240_DepthPriming.png
+++ b/TestProjects/UniversalGraphicsTest_PostPro/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/240_DepthPriming.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7a3bff3f6970ca1f7811827e58d09cf1dcfac4d0a8c91357b01abb714085805f
-size 310721
+oid sha256:79c264c51f2fd261d7114735767f92a17b5fef97cfbe2ca93255f8ae9a349004
+size 258375
--- a/TestProjects/UniversalGraphicsTest_PostPro/Assets/ReferenceImages/Linear/Android/Vulkan/None/240_DepthPriming.png
+++ b/TestProjects/UniversalGraphicsTest_PostPro/Assets/ReferenceImages/Linear/Android/Vulkan/None/240_DepthPriming.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7a3bff3f6970ca1f7811827e58d09cf1dcfac4d0a8c91357b01abb714085805f
-size 310721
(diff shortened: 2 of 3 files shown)


com.unity.render-pipelines.universal/Runtime/RenderTargetBufferSystem.cs
```diff
@@ -29,11 +29,6 @@ public RenderTargetBufferSystem(string name)
             RTB.Init(name + "B");
         }
 
-        public RenderTargetHandle GetBackBuffer()
-        {
-            return m_FirstIsBackBuffer ? RTA : RTB;
-        }
```
phi-lira, line 35: Note for RM: this class is internal, not breaking change to user.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -389,12 +389,14 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
                 m_DeferredLights.IsOverlay = cameraData.renderType == CameraRenderType.Overlay;
             }
 
+            CommandBuffer cmd = CommandBufferPool.Get();
```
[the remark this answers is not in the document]
phi-lira: Dont' we have already a `cmd` available in `UniversalRenderer`?

TestProjects/UniversalGraphicsTest_Foundation/ProjectSettings/TagManager.asset
```diff
@@ -17,8 +17,8 @@ TagManager:
   - VolumeLayer2
   - VolumeLayer3
   - VolumeLayer4
-  - 
-  - 
+  - Terrain
+  - Props
```
phi-lira, line 21: Why is it needed?

NicoLeyman: It's not, I've changed it to just use the existing UI tag so I don't need to any new ones.

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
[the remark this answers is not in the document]
robin-demoor: Turns out we won't need the cmd in the UniversalRenderer Setup so this is all removed

simon-engelbrecht-soerensen (approved):
Would be nice to have screenshots of ref images that were added in the PR.
Otherwise looks good.

phi-lira (approved):

PR #4772, opened 2021-06-03
APV Scene bounds inflation so they match brick sizes.  

FrancescoC-unity:
As the title says, mostly this is prep work to spare some space in the index. 

When we will have the probes not generated at cell distance, we can then stop rounding the index by cell and save a significant amount of space.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneBounds.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneBounds.cs
@@ -110,7 +110,7 @@ namespace UnityEngine.Experimental.Rendering
             float minBrickSize = refVol.MinBrickSize();
 
             float minSideInBricks = Mathf.CeilToInt(minSizedDim / minBrickSize);
-            int subdivLevel = Mathf.CeilToInt(Mathf.Log(minSideInBricks, 3));
+            int subdivLevel = Mathf.FloorToInt(Mathf.Log(minSideInBricks, 3));
 
             return subdivLevel;
         }


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneBounds.cs
```diff
@@ -102,6 +102,37 @@ public void OnBeforeSerialize()
...
 #if UNITY_EDITOR
+        private int FindInflatingBrickSize(Bounds b)
+        {
+            var refVol = ProbeReferenceVolume.instance;
+            float minSizedDim = Mathf.Min(b.size.x, Mathf.Min(b.size.y, b.size.z));
+
+            float minBrickSize = refVol.MinBrickSize();
+
+            float minSideInBricks = Mathf.CeilToInt(minSizedDim / minBrickSize);
+            int subdivLevel = Mathf.CeilToInt(Mathf.Log(minSideInBricks, 3));
```
alelievr, line 113: Isn't it a FloorToInt here? because we want to take the min brick size that fits inside the min side length of the probe volume? (at least this is what I understood from your drawings yesterday)

alelievr (approved):

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneBounds.cs
FrancescoC-unity: Yup I think you are right, in my sketches I started counting from 1 and not 0 :D Will fix up.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #4773, opened 2021-06-03
[URP][Converter] anim clip converter

martint-unity:
---
### Purpose of this PR
Added an animation Clip Converter.
In menu Window -> Rendering -> Render Pipeline Converter

image

We now have a new converter called animation clip converter.
This will convert animation clip properties if needed. 

This can be tested in UniversalUpgradeTest project
for example changing 
attribute: material.**_Color.a** to attribute: material.**_BaseColor.a**

If opening the test project and initialize the converters you will have the following
image

After running convert you will have something that looks similar to this
image

---

### DIFF
--- a/com.unity.testing.urp-upgrade/Tests/Editor/AnimationClipUpgrader_UnitTests.cs
+++ b/com.unity.testing.urp-upgrade/Tests/Editor/AnimationClipUpgrader_UnitTests.cs
@@ -273,6 +273,28 @@ namespace UnityEditor.Rendering.Tests
                 .SetName("Single target material, upgraded, float property"),
             new TestCaseData(
                 new[] { ("ID1", "NewShader") },
+                "material._BaseColor.r",
+                new[] { (From: "_BaseColor", To: "_BaseColor") },
+                new[]
+                {
+                    ("OldShader", "NewShader", new[] { (From: "_Color", To: "_BaseColor", Type: (int)MaterialPropertyType.Color) })
+                }
+            )
+                .Returns(SerializedShaderPropertyUsage.UsedByUpgraded)
+                .SetName("Single target material, upgraded, color property already upgraded"),
+            new TestCaseData(
+                new[] { ("ID1", "NewShader") },
+                "material._BaseMap_ST_ST.x",
+                new[] { (From: "_BaseMap_ST_ST", To: "_BaseMap_ST_ST") },
+                new[]
+                {
+                    ("OldShader", "NewShader", new[] { (From: "_MainTex_ST", To: "_BaseMap_ST_ST", Type: (int)MaterialPropertyType.Float) })
+                }
+            )
+                .Returns(SerializedShaderPropertyUsage.UsedByUpgraded)
+                .SetName("Single target material, upgraded, float property already upgraded"),
+            new TestCaseData(
+                new[] { ("ID1", "NewShader") },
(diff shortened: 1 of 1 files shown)


simonbz:
I'm hitting an error after going through domain reload. I'm running the converter and then I `git reset --hard` to reset the assets. The converter window then opens up with the following error:

`TargetException: Non-static field requires a target
System.Reflection.RuntimeFieldInfo.SetValue (System.Object obj, System.Object val, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Globalization.CultureInfo culture) (at <56a09cea74224b4cafcbda9f814f7d23>:0)
System.Reflection.FieldInfo.SetValue (System.Object obj, System.Object value) (at <56a09cea74224b4cafcbda9f814f7d23>:0)
UnityEditor.Rendering.Universal.Converters.RenderPipelineConvertersEditor.DontSaveToLayout (UnityEditor.EditorWindow wnd) (at /Users/sbz/work/Graphics/com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs:115)
UnityEditor.Rendering.Universal.Converters.RenderPipelineConvertersEditor.ShowWindow () (at /Users/sbz/work/Graphics/com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs:94)`

simonbz:
I spent some time playing with the animation clip converter trying to make sense of the assets that were shown in the UI after having successfully updated them. When saving the assets to disk and reopening the converter, I'm retrieving the exact same list of animation clips to upgrade, but this time the operation mostly ends in failure.

Screen Shot 2021-06-04 at 1 53 23 PM

Some things to consider:
- The clips that are green are the "WithMaterialProperties_OnlyUsedByNotUpgradable" clips that are showing up in the upgraded list. Since our filter is set to "UsedByUpgraded | UsedByNonUpgraded", these show up as upgraded while we haven't done any work on them. Although not an error per se, these should probably be in the nonUpgraded list still.
- Clips that were already upgraded are now added to the nonUpgraded list with a usage set to "Unknown". Should these clips be in the list to begin with? Would it be possible to remove clips from the list that either have a usage set to "UsedByNonUpgraded" or "Unknown"?

martint-unity:
@simonbz Domain reload is an issue and there are other PR's that will fix that. On is in the default Material Converter and on is in A Lazy load GUI states. 
@amechtley Could you maybe sync with @simonbz and figure out what we would / should do for the 
"Some things to consider:

The clips that are green are the "WithMaterialProperties_OnlyUsedByNotUpgradable" clips that are showing up in the upgraded list. Since our filter is set to "UsedByUpgraded | UsedByNonUpgraded", these show up as upgraded while we haven't done any work on them. Although not an error per se, these should probably be in the nonUpgraded list still.
Clips that were already upgraded are now added to the nonUpgraded list with a usage set to "Unknown". Should these clips be in the list to begin with? Would it be possible to remove clips from the list that either have a usage set to "UsedByNonUpgraded" or "Unknown"?"
And maybe we can address it?

amechtley (approved):
lgtm

simonbz (approved):

erikabar (approved):
Tested if Animation clips appear in the converter window and are successfully converted to URP.

phi-lira (approved):

PR #4775, opened 2021-06-03
Fix UX issues in URP Rendering Debugger

arttu-peltonen:
### Purpose of this PR
Fixes issues listed in https://fogbugz.unity3d.com/f/cases/1340619/

- Rewrite logic for URP debug widget conditional hiding.
- Hide Hue Tolerance & Saturation Tolerance fields for Default Luminance under Material Albedo validation.
- Change Map Size to percentage of screen instead of pixels.

**NOTE**: Setting wireframe color is **not fixed** - there is no API to set this in SRP right now. Scene View wireframe modes seem to use a special wireframe shader on most backends, with Vulkan being the only one to support colored wireframe at API level.

Conditional UI for albedo validation:
hide-tolerance-ui

Map Size as percentage (default = 50%):
overlay-map-size-percentage

---
### Testing status
Tested UI conditional behaves correctly both in Editor and Runtime. Map Size now respects viewport resolution as expected.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
@@ -178,35 +178,6 @@ namespace UnityEngine.Rendering.Universal
         internal DebugMaterialMode DebugMaterialModeData { get; private set; }
         internal DebugVertexAttributeMode DebugVertexAttributeIndexData { get; private set; }
 
-        static void DebugMaterialValidationModeChanged(DebugUI.Field<int> field, int value)
-        {
-            // Hacky way to hide non-relevant UI options based on displayNames.
-            var mode = (DebugMaterialValidationMode)value;
-            var validationWidgets = field.parent.children;
-            foreach (var widget in validationWidgets)
-            {
-                if (mode == DebugMaterialValidationMode.None && (
-                    widget.displayName == k_AlbedoSettingsContainerName ||
-                    widget.displayName == k_MetallicSettingsContainerName))
-                {
-                    widget.isHidden = true;
-                }
-                else if (mode == DebugMaterialValidationMode.Albedo && widget.displayName == k_MetallicSettingsContainerName)
-                {
-                    widget.isHidden = true;
-                }
-                else if (mode == DebugMaterialValidationMode.Metallic && widget.displayName == k_AlbedoSettingsContainerName)
-                {
-                    widget.isHidden = true;
-                }
-                else
-                {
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
```diff
@@ -251,7 +222,7 @@ internal static class WidgetFactory
                 setter = (value) => {},
                 getIndex = () => (int)data.albedoDebugValidationPreset,
                 setIndex = (value) => data.albedoDebugValidationPreset = (AlbedoDebugValidationPreset)value,
-                onValueChanged = (field, value) => DebugManager.instance.ReDrawOnScreenDebug()
+                onValueChanged = (_, _) => DebugManager.instance.ReDrawOnScreenDebug()
```
alex-vazquez, line 225: normally for unused parameters I think is better to put `(_, __)`

alex-vazquez (approved):

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
arttu-peltonen: I think in order to make it a [C#7 discard](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards), the single underscore is what I need to use. My guess is that before C#7, this was more of a common naming convention and hence you couldn't use duplicate variable names.

alex-vazquez: maybe I am becoming and old guy :)

TheoWong-pixel (approved):

TheoWong-pixel:
@oleks-k Might need to note in the docs later that Map size: is a percentage of the screen. 
@arttu-peltonen I think we need to apply this change the HDRP mapsize implementations as well? unless that is already the case.

arttu-peltonen:
> @arttu-peltonen I think we need to apply this change the HDRP mapsize implementations as well? unless that is already the case.

On HDRP, there is already "Lighting > Debug Overlay Screen Ratio" which goes from 0 to 1 (essentially a percentage), and scales the shadow debug overlay accordingly.

simon-engelbrecht-soerensen (approved):
Looking good to me.

Verasl (approved):
LGTM

Would have liked the wireframe colour but this seems to be a Unity trunk change

lukaschod (approved):

ellioman (approved):

PR #4777, opened 2021-06-03
In radius subgraph

MarieGuff:
**Distance comparison subgraph** in VFX graph additions package
[TriggerSphereTestElements.zip](https://github.com/Unity-Technologies/Graphics/files/6599334/TriggerSphereTestElements.zip)

image

MODIFICATIONS:

Changed to sphere behavior + naming "Trigger (Sphere)"

### Checklist for PR maker

---
### Purpose of this PR
Providing a simple way for user to create behavior when a particle reaches a target

PR-InRadiusDemo

Demo graph:
image

See complete informations in the following document:
https://docs.google.com/document/d/1lsJOTzRpPgf__4ehpXiu2V46GRozY08xciSdrd_6uFk/edit?usp=sharing

---
### Testing status
Local and world space compatibility, and two use cases. See attached zip folder. 

---
### Comments to reviewers
Testing material includes two VFX graph effects used in a prefab. Drop the prefab in you scene for testing.
The first prototype subgraph is also included as additionnal information


VladNeykov:
Hi! It's looking good, here's the initial [test doc](https://docs.google.com/document/d/1r62zxZ_XTs6A0qcqtAkxPgYv3WNCwV6sg4vlYyM5XVg/edit#) with some notes.

MarieGuff:
MODIFICATIONS: 

Changed to sphere behavior + naming

VladNeykov (approved):
Thanks for the PR! The subgraph is looking good and feels intuitive to use:
kSe0mZcf3f

There's an edge case due to [this](https://fogbugz.unity3d.com/f/cases/1163500/) bug about space not propagating into the subgraph, leading to cases like this below. This is a separate issue which will need addressing, unrelated to this PR.
39j0WrCdit

[Test doc](https://docs.google.com/document/d/1r62zxZ_XTs6A0qcqtAkxPgYv3WNCwV6sg4vlYyM5XVg/edit#) for reference.

peeweek (changes requested):
Node seems pretty functional at the exception of the space of the system which is not transmitted (not related to this PR). 


However I see some issues that should be adressed:

1) Node name "Trigger (Sphere)" does not reflect the actual outcome of the operator : In VFX Graph triggers are always blocks that trigger the creation of other particles through GPU events. Which is not the case here. However. what this node does is return whether the particle Is contained within a sphere or is Inside a sphere. Renaming the node to `Is Inside (Sphere)` or `Is Contained (Sphere)` would be relevant.

2) Absolute value is not necessary for the distance check, as it will be always positive 
image

3) For sphere distance checks, there's already a node that performs the distance computation in an optimized manner:

image

4) As the Space of the system is not transmitted, and we don't know when (and if) it will be fixed, it doesn't mean that we shouldn't make it functional out of the box in the meantime. Here's a suggestion on how to prevent issues. (Note: this suggestion also takes the sphere distance operator into account)

image

image


We will remove the System Space input property when the "Get Attribute" node will output a correctly spaced property.

MarieGuff:
Requested changes: 
1. Naming pending (waiting Vic cooper's help)
2. Absolute Value deleted, thanks!
3. Only the position is used in the distance check, the radius is used later. The sphere is only one of the possible use case, but not the only one (the demo graph uses a position and a float for the range, for example). Not sure the distance(sphere) attribute would be a good choice in that case. I put it on hold and wil lsee again with the naming choice. 
4. Done, thanks!

julienf-unity (commented):
I'm not sold on the name "Trigger" I think "IsInside" better convey what the subgraph is actually doing.

Also nodes we have generally will have the position passed as input, here it's directly fetching the position attribute within the subgraph. It's ok (even if it causes potential space issue then but we cannot really handle that atm). I'm wondering if this should be reflected in the name or not

VladNeykov:
> I'm not sold on the name "Trigger" I think "IsInside" better convey what the subgraph is actually doing.

I misled Marie with the trigger suggestion based on the Unity collision system, both Thomas and you are right about that meaning something else in the VFX Graph context, especially in the context of GPU events.
image

MarieGuff:
Changes summary: 
- Naming to "Is Inside"
- Local/World space droplist
- deleted "Absolute" node
- Sphere version

peeweek (changes requested):
image

Tested and works gracefully! However namings should be adjusted to the following:
- Output: `Is Triggered` should be renamed `Is Inside` (*)
- Input : `Comparison Sphere` should be simplfied to `Sphere`
- Input : `Comparison Range` should be simplified to `Padding` or `Radius Padding` (**)

(*) Trigger is a concept in VFX Graph that creates particles for GPU events, which is not the case in that operator, it's just a boolean that tells whether it's inside or not.

(**) Comparison Range is a bit confusing as the range is actually the radius + the `Comparison Range`. Naming that as padding, and set the tooltip to something like "Defines optional extra padding to the sphere, for the check, for instance to take particle size into account".

MarieGuff:
namings updated

peeweek (approved):

PR #4782, opened 2021-06-03
[HDRP] Fix console errors when trying to upgrade template to master

RSlysz:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1340369/ (error appearing when migrating due to proxy not being init yet)

Way it is fixed:
- Proxy is removed
- Base class named `HDProjectSettingsReadOnlyBase` that implement the loading and only one field `projectSettingsFolderPath` in readonly access that is in runtime assembly where we need it when we create a **HDRenderPipelineGlobalSettings** asset by migrating the **HDRenderPipelineAsset** in GraphicSettings
- `HDProjectSettings` inherite from this base class. It is in editor assembly and we can thus add editor types here. It also add write access to the `projectSettingsFolderPath`.

Additional cleaning:
- The ScriptableSingleton is now usable (don't know from how long). But due to the above and the check it do in its constructor we cannot use it.
- But we can use it for HDUserSettings. Code is updated to use it.

---
### Testing status
Tested:
- From 2019.4.27f1 HDTemplate, close editor, update package.json to use this branch and open it with 2021.2.0a20. No error seen. HDGlobalSettings is created in projectSettingsFolderPath folder with Migration from HDRenderPipelineAsset in GraphicsSettings. HDUserSettings is still created with migration from HDProjectSettings. [I make several test with small code variation to ensure the repertoir used for the creation is the one that come from the HDProjectSettingsReadOnlyBase.projectSettingsFolderPath]
- From 2019.4.27f1 HDTemplate, close editor, remove the HDProjectSettings, update package.json to use this branch and open it with 2021.2.0a20. No error seen. HDGlobalSettings is created in projectSettingsFolderPath folder. HDProjectSettings and HDUserSettings are still created.

---
### Comments to reviewers
@slunity What is the intend of the [InitializeOnLoadMethod] I heard that we must avoid it in SRPs. Can we do it another way ?

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/HDProjectSettings.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/HDProjectSettings.cs
@@ -132,7 +132,6 @@ namespace UnityEditor.Rendering.HighDefinition
 
         HDProjectSettings()
         {
-            //s_Instance = this; //Todo: should be done in base static contstructor. To check
             // s_Instance.FillPresentPluginMaterialVersions(); <
             // We can't call this here as the scriptable object will not be deserialized in yet
         }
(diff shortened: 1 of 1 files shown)


remi-chapelain (approved):
Thanks to the detailed description, seems like no further testing needed for this PR ! ✔️

RSlysz:
@remi-chapelain Edit: I failed my test on the migration.
- Data are migratted correctly from HDRenderPipelineAsset to HDRenderPipelineGlobalSettings
- Migration is not done from HDProjectSettings to HDUserSettings (data is unchanged)

I have an idea to fix this and will try to do it tonight.

RSlysz:
@remi-chapelain The above fix the migration HDProjectSettings -> HDUserSettings call when loaded from HDProjectSettingsReadOnlyBase.
I checked again HDRenderPipelineGlobalSettings and it is still correctly migrated (Light layer names are kept)

slunity:
@RSlysz : We use InitializeOnLoadMethod in our MaterialPostProcessor for the version check polling function (RegisterUpgraderReimport) though? But to think of it yes it's possible to do it in another way: we just need a non serialized instance state variable that indicates the singleton hasn't been accessed yet and on first access, to call reset and set that state flag so that it doesn't get called repeatedly.

com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/HDProjectSettings.cs
```diff
@@ -302,16 +320,19 @@ internal enum Version
...
 
-        Version IVersionable<Version>.version { get => instance.m_Version; set => instance.m_Version = value; }
+        //Never use "instance" here as it can create infinite call loop. Use s_Instance instead.
+        Version IVersionable<Version>.version { get => (s_Instance as HDProjectSettings).m_Version; set => (s_Instance as HDProjectSettings).m_Version = value; }
 
+        //Never use "instance" here as it can create infinite call loop. Use s_Instance instead.
         static readonly MigrationDescription<Version, HDProjectSettings> k_Migration = MigrationDescription.New(
             MigrationStep.New(Version.SplittedWithHDUserSettings, (HDProjectSettings data) =>
             {
+                HDProjectSettings inst = s_Instance as HDProjectSettings;
```
slunity, line 330: @RSlysz I was just wondering why not use directly the "data" param of MigrationStep here (it will be called with s_Instance above)

RSlysz: Excellent point. I have no reason really. I may suffer bug :D (I'll change this)

RSlysz: No in fact, this is totally correct.
Normally we should use the data param but here everything is grabed through the static instance due to the singleton pattern. So `data.something == HDProjectsettings.instance.something`.
But here we also are in the migration and the instance accessor ensure the migration is done because when it is loaded from HDProjectSettingsReadonlyBase, it will not migrate automatically. Thus if we call instance here (or use data) we will trigger a recursive call look which will lead to stack overflow error.

jenniferd-unity: we should probably add this last comment as a comment in the code :)

jenniferd-unity (approved):

RSlysz:
It is right before at line 326 :)

com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/HDProjectSettings.cs
slunity: I must be missing something (sorry :) ) but isn't that loop prevented by the call at line 284 above  if (k_Migration.Migrate(inst)), inst is set to the static s_instance, not the accessor ?

jenniferd-unity: @RSlysz actually i saw that comment but it is missing the explanation of the singleton pattern. i feel like if i read the code again in a couple of months i would forget about that

RSlysz: @jenniferd-unity ok I'll add it then :)

@slunity what hapens is this (I'll add comment on this too):
**Case 1:** 
- no early migration occured, we need to access one field here and this pass by a static accessor such as HDProjectSettings.materialVersionForUpgrade
- This will request the **instance** which is not loaded/created yet. So the LoadOrCreate will be called that should trigger the migration.

**Case 2:**
- When opening unity Editor, the package.json have changed and request a newer version of the HDRP package. While evreything is reimported by the ADB, the render loop will start to be called (when modal say "Load Scene" in fact).
- This can trigger a migration from former system without GlobalSettings to the new one. In this case we will try to create a HDRenderPipelineGlobalSettings out of the HDRenderPipelineAsset in the GraphicSettings. For this asset to be created, we must access the m_ProjectSettingFolderPath. This is done from the RuntimeAssembly and we have no certitude that the editor assembly is loaded. This is the reason we have a base class HDProjectSettingsReadonlyBase that lives in Runtime assembly.
- Though HDProjectSettingsReadonlyBase don't see every part of the object and can just try to load it. When this is the case, the migration is **NOT triggered**.
- So next time we will need access to HDProjectSettings, we need to say "Hey migrate if you need".
- For this reason, in the **instance**, we need to check if we need to migrate or not. This is done by checking the version at line 123 and remove the static instance in this case to force its reloading in the HDProjectSettings way (which means with Migration)

When migration is triggered:
- When we call the Migrate(), the System will try first to check the version to see if migration is needed. So if IVersionable<Version>.version use instance, we will suppress the static instance to reload it if the version is not last (see Case 2). In the end, as it is loaded again, it will request a Migration. This is a recursive call that will lead to stack overflow. So we must not use **instance** in IVersionable<Version>.version
- Then if version is not the last, we will execute missing migration steps. Once again if inside a step we use **instance** we will cause a recursive call like in above and will produce a stack overflow. So no HDProjectSettings.materialVersionForUpgrade but we can safely use (s_Instance as HDProjectSettings).m_LastMaterialVersion or data.m_LastMaterialVersion (as pointed by you slunity, data is always the s_Instance as a HDProjectSettings)

RSlysz: Explaining it make me realize I was a little wrong on the data. you can use it directly. What is forbidden is trying to use "instance" and data is already the instance that we want.
I added this whole explanation in comments as it will be simpler to debug later with this.

slunity: Good! Thanks for adding

slunity (approved):

sebastienlagarde (approved):

PR #4784, opened 2021-06-03
[HDRP] APV - Remove useless bricks subdivison apv

alelievr:
### Purpose of this PR
As discussed with @FrancescoC-unity , this PR removes the bricks bigger than the min edge of a probe volume (see gif)

Subdiv2

This PR also prevents the probe volumes to generate bricks of the size of the cell when there is no geometry inside the probe volume.

Also adjusted the function that calculates the bounds of the scene to match the new subdivision algorithm:
image
(you can see the calculated bounds in purple matches the subdivision for cases where the probe volume overlaps only with a small part of a cell)
Another example with a smaller probe volume (this time the bounds are in blue):
image

---
### Testing status
Tested that the subdivision still works correctly with the real-time mode


JulienIgnace-Unity:
>This PR also prevents the probe volumes to generate bricks of the size of the cell when there is no geometry inside the probe volume.

Do we actually want that? IMO if the user put a volume somewhere then they want at least the minimum density of probes.
Otherwise you have no way to handle objects flying through empty space for example.

alelievr:
It's still possible to do that with the subdivision controller in the probe volume.

This PR actually improves a lot the memory usage of the index dimension when you have a high simplification level

JulienIgnace-Unity (approved):

FrancescoC-unity:
> It's still possible to do that with the subdivision controller in the probe volume.
> 
> This PR actually improves a lot the memory usage of the index dimension when you have a high simplification level

No it won't really because when we compute the index size we do not know if the probe volume contains objects or not (we cannot do that without loading scenes), so we pad as if it was full anyway

FrancescoC-unity:
Also since it landed in master, could you update the padding in scene bounds to match the split of volumes we discussed yesterday?

alelievr:
> No it won't really because when we compute the index size we do not know if the probe volume contains objects or not (we cannot do that without loading scenes), so we pad as if it was full anyway

but we don't need to know that, we only need to know the size of the probe volume which directly gives the max possible brick size and if we do that for all probe volumes in all scenes, we can reduce the index dimension size by a great deal. Or am I missing something?

FrancescoC-unity (approved):

PR #4790, opened 2021-06-04
Quality Audit - Layer mask pup up fix

alex-vazquez:
---
### Purpose of this PR
https://jira.unity3d.com/browse/XPIPELINE-193

image

---
### Testing status
Check the pop up is working

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.core/CHANGELOG.md
+++ b/com.unity.render-pipelines.core/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - LensFlare (SRP) can be now disabled per element
 - LensFlare (SRP) tooltips now refer to meters.
 - Serialize the Probe Volume asset as binary to improve footprint on disk and loading speed.
+- Layer mask pop up to the standard popup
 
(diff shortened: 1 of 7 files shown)


RSlysz (changes requested):
I bet we better update trunk to make the method public instead of making it public in CoreRP.

Making it public in CoreRP means that users will be able to use it. But we cannot be garant of it working fine. There is no test here and trunk can change its signature without realizing it is breaking this method.

For sustenability, I don't think this is the good way to do it. Or we need to have it used in an automated test at least. (Minima is: "does this signature still exist")

RSlysz:
@TheoWong-pixel Actually, the one used is the default provided to anyone to use. This is the basic way of drawing a layer field when you use the `EditorGUILayout.PropertyField`. If the result of this function is not correct, we must change it as this is the version available to user.
This should affect the WHOLE unity currently.

RSlysz:
@TheoWong-pixel Additionally, the "dark" version support multiple differente value bitwise in the layer.
image

Be sure the old white one support this prior changing it. (I'm pretty sure it does not undestand and will say either checked or nothing)

jenniferd-unity (approved):

PR #4802, opened 2021-06-04
[HDRP] Fix cropping issue with the graphics compositor when using the camera bridge

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1340549/
This is a small follow-up to a [previous PR ](https://github.com/Unity-Technologies/Graphics/pull/4599) in order to fix a strange cropping issue that we have encountered in certain configurations.

---
### Testing status
- Did the same test as this PR: https://github.com/Unity-Technologies/Graphics/pull/4599 to make sure we did not break the previous functionality. 
- I have also tested using the repro project provided by the internal team

---
### Comments to reviewers
It turns out the cmd.blit in Unity has different behavior depending on the type of the source argument.
I'm now using a temporary RT, and the behavior is the same as when recording an HDRP camera.

The difference is that under the hood, the blit was translated before to a copySubResourceRegion command, but now it doing a fullscreen draw with the BlitCopy material (which also scales the input as required instead of cropping).


YohannVaastUnity (approved):
It works perfectly with our HDRP Blackmagic project, the issue is gone. Thanks a lot !

sebastienlagarde (approved):

PR #4805, opened 2021-06-04
[ShaderGraph] Fixed auto generated reference names for properties to consistently h…

joshua-davis:
…ave an underscore pre-pended for all property types.

### Checklist for PR maker

---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1336937/
The auto-generated reference name was not consistent for all types. Some types would get pre-pended with an underscore and some wouldn't. This was because some names were reserved names (Float) but some weren't (Vector2) which caused the difference. Now they are all pre-pended if renaming of reference names is allowed.

---
### Testing status
- [x] Added new property tests
- [x] Verified all new primtive types have a pre-pended underscore based upon the display name
- [x] Verified that changing the display name will produce a reference name with an underscore
- [x] Verified that manually overriding the reference name and then changing the display name doesn't affect the overriden reference name
- [x] Verified that resetting an overriden reference name back to default does the correct underscore logic
- [x] Verified that a display name with an underscore doesn't produce a reference name with a double under score
- [x] Verified that Material Quality produces the same name as before
- [x] Verified Sampler States in a main graph aren't changed (as they have a reference name based upon the proeprties)

---
### Comments to reviewers


jessebarker (approved):
LGTM.

GrantLamb-Unity (approved):
Developer testing is sufficient, and good work on the DefaultNamePropertyTest!

PR #4806, opened 2021-06-04
Fixed some Shader Graph shaders fail to compile for Android XR

thomas-zeng:
---
### Purpose of this PR
* Fixed fogbugz case https://fogbugz.unity3d.com/f/cases/1340356/

---
### Testing status
- QA
- ABV

---
### Comments to reviewers
- Added special handling for fields within `UNITY_STEREO_INSTANCING_ENABLED` define so that they could be placed at the end of PackedVaryings struct. This is required for SV_RenderTargetArrayIndex to avoid error: `Non system-generated input signature parameter () cannot appear after a system generated value`
- To QA: I have tested the repro project on Quest device and was able to see objects with shader graph materials render as expected.

### DIFF
--- a/com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
+++ b/com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
@@ -151,7 +151,7 @@ namespace UnityEditor.ShaderGraph
                 {
                     // special case, "UNITY_STEREO_INSTANCING_ENABLED" fields must be packed at the end of the struct because they are system generated semantics
                     //
-                    if (subscript.HasPreprocessor() && (subscript.preprocessor.Contains("UNITY_STEREO_INSTANCING_ENABLED")))
+                    if (subscript.HasPreprocessor() && (subscript.preprocessor.Contains("INSTANCING")))
                         postUnpackedSubscripts.Add(subscript);
                     // special case, "SHADER_STAGE_FRAGMENT" fields must be packed at the end of the struct,
                     // otherwise the vertex output struct will have different semantic ordering than the fragment input struct.


com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
```diff
@@ -154,6 +154,10 @@ internal static void GeneratePackedStruct(StructDescriptor shaderStruct, ActiveF
                     //
                     if (subscript.HasPreprocessor() && (subscript.preprocessor.Contains("SHADER_STAGE_FRAGMENT")))
                         postUnpackedSubscripts.Add(subscript);
+                    // special case, "UNITY_STEREO_INSTANCING_ENABLED" fields must be packed at the end of the struct because they are system generated semantics
```
fabien-unity, line 157: This will not conflict with the previous one (SHADER_STAGE_FRAGMENT) which also needs to be at the end ?

thomas-zeng: SHADER_STAGE_FRAGMENT field is also a system generated semantic so it won't conflict with our XR use case. However the comment mentioned semantic ordering in fragment shader I am not sure about that use case. Will dig in more.
https://github.com/Unity-Technologies/Graphics/blob/c6ae7599b33ca48852eddd592b3e40f33f2dd982/com.unity.shadergraph/Editor/Generation/TargetResources/StructFields.cs#L60

esmelusina: The `SHADER_STAGE_FRAGMENT` case is there for situations where frag only SVs coming in wouldn't've been part of the vertex output but still needs to be in the interpolator structure. Since the custom interpolator feature allows arbitrarily adding more interpolators, we had to make sure frag only fields were pushed to the bottom.

A similar issue shows up with instancing, but the instance ids are optionally established in the vertex stage. So-- I think what we want here is for the `UNITY_STEREO_INSTANCING_ENABLED` case to come before the `SHADER_STAGE_FRAGMENT` case. They both need to be pushed to the end, but fragment should come after instancing.

I think it's safe (and preferred) to generalize this for any type of instancing, not just stereo.

esmelusina (changes requested):
Changes requested in the comment thread there (and mentioned on slack).

Let me know if that's unclear.

DennisDeRykeUnity:
Dev branch (universal/xr/fix-sg-packedvaryings bb08c6a) is green on Yamato for Android XR OpenGL URP Foundation, Lighting, PostProc, and Terrain per links below.  For Vulkan these builders fail, but per discussion with XR Graphics devs this is a separate bug(s).  I see that PR 4806 is still under dev review, so there more be further commits that need to be tested before it can be reviewed by QA.  

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fxr%252Ffix-sg-packedvaryings/.yamato%252Furp_foundation-android-opengles3.yml%2523Build_URP_Foundation_Android_OpenGLES3_Standalone_XR_il2cpp_Linear_trunk/7176785/job 

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fxr%252Ffix-sg-packedvaryings/.yamato%252Furp_lighting-android-opengles3.yml%2523URP_Lighting_Android_OpenGLES3_Standalone_XR_il2cpp_Linear_trunk/7176794/job 

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fxr%252Ffix-sg-packedvaryings/.yamato%252Furp_postpro-android-opengles3.yml%2523URP_PostPro_Android_OpenGLES3_Standalone_XR_il2cpp_Linear_trunk/7176801/job

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fxr%252Ffix-sg-packedvaryings/.yamato%252Furp_terrain-android-opengles3.yml%2523URP_Terrain_Android_OpenGLES3_Standalone_XR_il2cpp_Linear_trunk/7176803/job

com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
thomas-zeng: Hi esmelusina, thanks for the feedback!
I have moved the `UNITY_STEREO_INSTANCING_ENABLED` case before the `SHADER_STAGE_FRAGMENT` case. I was checking the usage of SV_InstanceID in shader graph. It is not used in Varyings. So I think we don't need to handle it in the PackedVaryings here.

thomas-zeng: Hi @esmelusina , thanks for the feedback!
I have moved the UNITY_STEREO_INSTANCING_ENABLED case before the SHADER_STAGE_FRAGMENT case in 3774019. I also  checked the usage of SV_InstanceID in shader graph. It is only used as vertex shader input and not being used in Varyings. So I think we don't need to handle it in the PackedVaryings logic here.

fabien-unity (approved):

com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
esmelusina: It can be optionally pulled in if the "Instance ID" node is used in the fragment stage. I'd prefer if you went ahead and generalized it.

Thanks!

thomas-zeng: Thanks @esmelusina ,
I have made another commit that generalize the instance case here: 2a2afbe

esmelusina (approved):
Thanks for the changes. I'm approving for shadergraph, I'm going to @tian-unity to see if they want to check that this interacts correctly with speed tree.

DennisDeRykeUnity (approved):
This PR’s dev branch last inherited from master around 7/4/2021 so I am comparing Nightly from back then on master against today’s Nightly on the dev branch 

* Graphics (master 45a5ae2)
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Fall-vfx.yml%2523Nightly_VFX_trunk/7130646/job/pipeline

* Graphics (universal/xr/fix-sg-packedvaryings 2a2afbe) 
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fxr%252Ffix-sg-packedvaryings/.yamato%252F_abv.yml%2523all_project_ci_nightly_trunk/7365566/job

The dev branch is green wherever master was green except for several jobs.  These became green upon re-run except for these two: 

* URP_Performance_BoatAttack on Win_DX12_performance_build_editmode_mono_Linear

* URP_Performance_BoatAttack on Win_Vulkan_performance_build_editmode_mono_Linear

Both of these jobs had the following error and crashed.  I do not believe this is due to this PR because I see the same error and crash in recent jobs on Graphics master for these two pipelines. 

Error retrieving packages from source 'https://artifactory-slo.bf.unity3d.com/artifactory/api/nuget/slough-nuget':
 Could not connect to the feed specified at 'https://artifactory-slo.bf.unity3d.com/artifactory/api/nuget/slough-nuget'. Please verify that 
[12:17:30.831 Information] the package source (located in the Package Manager Settings) is valid and ensure your network connectivity.

esmelusina (approved):
Re-Approving to clear lock.

PR #4809, opened 2021-06-07
Added support for reflection probes as a fallback for ray traced reflections (case 1338644)

anisunity:
Added support for reflection probes as a fallback for ray traced reflections (case 1338644)
Fixed an issue where disbaled reflection probes were still sent into the the ray tracing light cluster.
When a ray hits the sky in the ray marching part of mixed ray tracing, it is considered a miss.
image
https://fogbugz.unity3d.com/f/cases/1338644/
Testing status
Added new tests to cover this new feature.
The previous tests are still green.


remi-chapelain (commented):
Thanks for adding the tests. ✊ 

I updated the  "both" and the "reflection" because they were using the same diffusion profile so "both" was not tested in it. I modified the probe box size for the test "both" to be able to see the probe contribution and the sky on the same capture. 

109_ReflectionFallbackBoth

I also renamed the volumes to avoid confusion.

What I tested : 
- Perf Mode / Quality Mode
- Checked if Layer masks were still functioning properly
- Mixed mode
- Denoiser

Tests are still green locally, also launched [yamato](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-1338644/.yamato%252Fall-hdrp_dxr.yml%2523PR_HDRP_DXR_trunk/7150260/job). 

One quick comment, even though it seems logical what do you think of changing "Reflection probe and Sky" to "Reflection Probe then Sky" to emphasize the fact than there is an order ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRenderPipeline.RaytracingDeferredLightLoop.cs
```diff
@@ -176,7 +177,7 @@ static void RayMarchGBuffer(CommandBuffer cmd, in DeferredLightingRTRPassData da
             cmd.SetComputeFloatParam(data.parameters.rayMarchingCS, HDShaderIDs._RayMarchingThicknessScale, thicknessScale);
             cmd.SetComputeFloatParam(data.parameters.rayMarchingCS, HDShaderIDs._RayMarchingThicknessBias, thicknessBias);
             cmd.SetComputeIntParam(data.parameters.rayMarchingCS, HDShaderIDs._RayMarchingSteps, data.parameters.raySteps);
-            cmd.SetComputeIntParam(data.parameters.rayMarchingCS, HDShaderIDs._RayMarchingReflectSky, data.parameters.raytracingCB._RaytracingIncludeSky);
+            cmd.SetComputeIntParam(data.parameters.rayMarchingCS, HDShaderIDs._RayMarchingReflectSky, 0);
```
sebastienlagarde, line 180: is it expected?

sebastienlagarde (commented):
I am not convince about the enum. Why to propose all the fallback mode? 
If I understand correctly we just want to avoid cost of reflection fallback if we don't want it. But I think it introduce way to many settings, we should keep it simple.

I suggest just to be able to enable the fallback or reflection probe or not if performance is a concern but we always have the sky as a fallback. Thought? (And if adding the code of reflection probe don't imply cost (due to the dynamic branching and maybe because it doesn't add extra vgpr pressure), let's have it all the time.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRenderPipeline.RaytracingDeferredLightLoop.cs
anisunity: Yes it is

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -229,6 +230,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed support of Distortion with MSAA
 - Fixed contact shadow debug views not displaying correctly upon resizing of view.
 - Fixed an error when deleting the 3D Texture mask of a local volumetric fog volume (case 1339330).
+- Fixed an issue where disbaled reflection probes were still sent into the the ray tracing light cluster.
```
remi-chapelain, line 233: disbaled -> disabled

anisunity:
Found a bug in the wieght mixing code of the probes, need to fix it.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflection.cs
```diff
@@ -141,6 +141,11 @@ public int rayMaxIterations
         #endregion
 
         #region Ray Tracing
+        /// <summary>
+        /// When enabled, SSR handles sky reflection.
```
anisunity, line 145: Out of date comment

anisunity:
review fixes done

sebastienlagarde (approved):

PR #4812, opened 2021-06-07
Fix issue with OnDemand directional shadow map being corrupted when reflection probe are updated same frame

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1340544/ 

The issue is that because cached directional lights are view dependent, if a reflection probe is rendered, we get updates only for the first camera / face of the reflection probe, caching that result. The result is broken looking shadows in the main view. 

This is somewhat expected, but because sounds like a common enough scenario, the shadow is not marked as rendered if the rendering comes from a reflection camera. 

Added some doc to mention this as well. 

**What did I test**: The repro scene + I made my own repro script with an update explicitly called upon button pressing. Tested with punctual and directional lights. 

NOTE FOR QA: The original bug has an error printed, but that comes from the user's script, I verified that the system now works as expected so that error can be ignored when verifying.


RemyUnity (approved):
I double checked the repro out of curiosity, everything seems fine 👍

sebastienlagarde (approved):

PR #4813, opened 2021-06-07
Quality audit - Adding doc browsing button on camera inspectors

alex-vazquez:
### Purpose of this PR
https://jira.unity3d.com/browse/XPIPELINE-195

image

This applies to HDRP and URP

---
### Testing status
Open the links and see that they are working fine

---
### Comments to reviewers
Improved the documentation sections, sorting them as well.
Merge the Base and Overlay cameras, added warnings on the docs


arttu-peltonen:
I have a couple of questions/suggestions :) 

From the screenshot, it seems like the question-mark icons are very low-res compared to the ones above and below from it - could this be fixed at the same time?

I also wonder if it was intended that all of these icons point to the same page? Should the behavior instead do one or both of following things?
1) Instead of having a docs link on each subsection, make the main Camera component doc link point to URP Camera page.
2) If each subsection is to have a link, should it directly to the section on the page? I.e. instead of "...camera-component-reference.html", point to "...camera-component-reference.html#PhysicalCamera"?

alex-vazquez:
@arttu-peltonen 
 1- Has been solved
 2- The url is being generated with the correct sections at the end, but as the documentation that is only doesn't have those sections ( as they are comming from my changes ) the section is discarded. Once everything is deployed the sections will work as you mentioned.

Thanks for the review

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Drawers.cs
```diff
@@ -1,17 +1,38 @@
...
 namespace UnityEditor.Rendering.HighDefinition
 {
     using CED = CoreEditorDrawer<SerializedHDCamera>;
 
     static partial class HDCameraUI
     {
-        static readonly ExpandedState<CameraUI.Expandable, Camera> k_ExpandedState = new ExpandedState<CameraUI.Expandable, Camera>(CameraUI.Expandable.Projection, "HDRP");
+        /// <summary>Enum to store know the expanded state of a expandable section on the camera inspector</summary>
+        [HDRPHelpURL("HDRP-Camera")]
+        public enum Expandable
```
RSlysz, line 13: This continue to move \o/ ( Hope it will finally stay in place in the future :D )

RSlysz (approved):
looks good.
Though, I feel we have this section wise help only for Camera at the moment and we have foldout header in many other component and material. Should we wait to have a batch landing all at once for uniformity ?

arttu-peltonen (approved):

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Drawers.cs
alex-vazquez: Lights have this change, materials are coming in another PR too(only URP for the moment)

Vic-Cooper (changes requested):
Docs changes suggested.

Vic-Cooper (approved):
Docs feedback applied. Thanks :)

simon-engelbrecht-soerensen (approved):
Links are working and haven't found any issues besides the fact that they all lead to the Camera page.

PR #4819, opened 2021-06-07
UX - Moving camera Allow Dynamic Resolution checkbox to rendering section, so its consistent with HDRP.

kecho:
### Purpose of this PR
* Dynamic Resolution is more related to Antialiasing and post processes than it is with camera's output.
* Moving camera Allow Dynamic Resolution to the Rendering section of the camera inspector.
* Updating documentation
* Minimal changes, shouldnt affect any functionality. 

---
### Testing status
Opened a scene, verified checkbox works and is moved.

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Documentation~/camera-component-reference.md
+++ b/com.unity.render-pipelines.universal/Documentation~/camera-component-reference.md
@@ -52,10 +52,15 @@ The Camera Inspector has the following sections when the Camera has its **Render
 <a name="base-rendering"></a>
 ## Rendering
 
+Note that when a Camera's *Render Mode* is set to *Base* and its *Render Target* is set to *Texture*, Unity does not expose the following properties in the Inspector for the Camera:
+
+* Allow Dynamic Resolution
+
 | __Property__               | __Description__                                              |
 | -------------------------- | ------------------------------------------------------------ |
 |__Renderer__ |Controls which renderer this Camera uses. |
 |__Post Processing__ |Enables post-processing effects. |
+|__Allow Dynamic Resolution__|Enables Dynamic Resolution rendering for this Camera.<br/>This property only appears when you select **Screen** from the **Output Target** drop-down.|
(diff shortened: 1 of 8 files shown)


alex-vazquez (approved):
Thank you :)

sebastienlagarde (approved):

sebastienlagarde:
can you trigger yamato? thanks

svens-unity (changes requested):
This is changing public API no? We should either offer an upgrade path or deprecate the old API

com.unity.render-pipelines.universal/Documentation~/camera-component-reference.md
```diff
@@ -52,10 +52,15 @@ The Camera Inspector has the following sections when the Camera has its **Render
...
 
+Note that when a Camera's *Render Mode* is set to *Base* and its *Render Target* is set to *Texture*, Unity does not expose the following properties in the Inspector for the Camera:
+
+* Allow Dynamic Resolution
+
 | __Property__               | __Description__                                              |
 | -------------------------- | ------------------------------------------------------------ |
 |__Renderer__ |Controls which renderer this Camera uses. |
 |__Post Processing__ |Enables post-processing effects. |
+|__Allow Dynamic Resolution__|Enables Dynamic Resolution rendering for this Camera.<br/>This property only appears when you select **Screen** from the **Output Target** drop-down.|
```
phi-lira, line 63: Note: there are some UX implications when moving this setting to Rendering. As stated above the dynamic resolution setting is hidden once Render Target is `Texture`. Today these setting lie next to each other, if you move you might cause confusion. 
 
 URP also groups all settings that affect camera target properties in Output category. I'm not convinced moving to Rendering is a better design.
 
 @TheoWong-pixel thoughts?

erikabar (approved):

com.unity.render-pipelines.universal/Documentation~/camera-component-reference.md
kecho: The interesting part is in HDRP this setting being in Output is far from Rendering, which will also cause the same problem since it affects more rendering properties, such as TAA (in HDRP). This whole PR's purpose was to remove that confusion.
If this is gonna cause confusion in URP then we should completely rethink this.

What I recommend is to rethink this: Why is Core enforcing a grouping at all? if this is an issue, then we should not have Core pipeline enforce a group. Thus maybe the core team should just change the core setting instead, and keep the groupings in the pipelines since they behave differently?

kecho:
Spoke with @alex-vazquez and @TheoWong-pixel.
@phi-lira brings up a good point, we are basically just moving the ball now to URP (in terms of confusion).
Thus we need to rethink the strategy, likely remove this from core and have each pipeline handle it properly, since DRS means something slightly different for each SRP.

Closing this pull request.

PR #4820, opened 2021-06-07
Write 0 instead of micro sized motion vectors + fix extremely fast velocities

FrancescoC-unity:
"Fix" for https://fogbugz.unity3d.com/f/cases/1337082/ 

The issue is that given we do various operations some precision is lost along the way, this is usually not a problem, but if we had significant motion vectors that cancel out (e.g. an object moving alongside the camera so with total motion vector of 0). 

Some floating imprecision are to be expected, but as a  workaround for the above case that might be annoying this PR adds a way to kill movements that are extremely small (1% of a pixel worth, such small movements are not going to be relevant with reprojection anyway). 

**Before**

https://user-images.githubusercontent.com/43168857/121043592-10374e00-c7b5-11eb-8272-496755e973b5.mp4

**After**

https://user-images.githubusercontent.com/43168857/121043617-14fc0200-c7b5-11eb-917c-52ce43a962b9.mp4

To look at this kind of issue I also added a debug mode to not display any motion vector smaller than a certain threshold (in pixels)

https://user-images.githubusercontent.com/43168857/121043699-27763b80-c7b5-11eb-8e1d-409efc5c7c57.mp4

ALSO https://github.com/Unity-Technologies/Graphics/pull/4820#issuecomment-856540285 another issue is fixed. 

The problem is that without clamp if we have a motion vector bigger than a screen-worth, it is detected as a "no motion" and so goes to 0 wrongly. Compeltely different than the issue above, for QA: to repro just put a plane and move the camera really fast.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/MotionVectors/CameraMotionVectors.shader
```diff
@@ -57,6 +57,13 @@ Shader "Hidden/HDRP/CameraMotionVectors"
 
             // Convert from Clip space (-1..1) to NDC 0..1 space
             float2 motionVector = (positionCS - previousPositionCS);
+
+#ifdef KILL_MICRO_MOVEMENT
+            // We multiply by 2 as the motion vectors have been already converted to NDC when we reach this function.
```
JulienIgnace-Unity: I don't see the *2 described in the comment. Mistake or outdated comment? (same above)

com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinUtilities.hlsl
```diff
@@ -14,6 +14,13 @@ float2 CalculateMotionVector(float4 positionCS, float4 previousPositionCS)
     previousPositionCS.xy = previousPositionCS.xy / previousPositionCS.w;
 
     float2 motionVec = (positionCS.xy - previousPositionCS.xy);
+
+#ifdef KILL_MICRO_MOVEMENT
+    // We multiply by 2 as the motion vectors have been already converted to NDC when we reach this function.
```
sebastienlagarde, line 19: where is the multiply by 2 mention in the comment?

FrancescoC-unity: Oh because I moved the code and forgot to remove the comment, will do remove that

JulienIgnace-Unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/MotionVectors/CameraMotionVectors.shader
FrancescoC-unity: Leftover from older code, removed!

Aydindeveloper:
There are problems not only with micro movements but with very fast motion too, here is a video

https://user-images.githubusercontent.com/49202311/121059866-a164ef80-c7ca-11eb-88c9-a27b7f3940c6.mp4

FrancescoC-unity:
> There are problems not only with micro movements but with very fast motion too, here is a video
> 
>  movie_006.mp4

Hey I am not sure what I am looking at there (especially given the compression artifacts :P ) , by micro movement I mean relative micro movements btw. If a camera moves at extreme speed but an object is linked to it, it will end up being a micro motion vector rendered to it.

FrancescoC-unity:
Regardint the bug in the video above, indeed looks like a different issue and I fixed it in a commit I just pushed. 

Before


https://user-images.githubusercontent.com/43168857/121144470-508ede80-c83e-11eb-9891-d0aff394ae03.mp4


After


https://user-images.githubusercontent.com/43168857/121144487-54bafc00-c83e-11eb-8108-56adc68a2b5d.mp4



I also had a chat with the user, further problems found will be flagged as separate bugs.

Aydindeveloper:
this solution doesn't work for angular camera movements as you move away from origin.

Camera position 0,10,0
https://user-images.githubusercontent.com/49202311/121386289-8a510980-c952-11eb-93c4-5656410c785d.mp4

Camera position 1000,10,0
https://user-images.githubusercontent.com/49202311/121386306-8f15bd80-c952-11eb-9f3f-0a0d72151699.mp4

sebastienlagarde (approved):

sebastienlagarde:
This PR kill the micro movement for the initial case and is an accepted fix for some imprecision. Move forward with this PR.
QA: The current coverage by Francesco should be enough.

PR #4822, opened 2021-06-07
Update SSAO ref images for WinEditor + Vulkan now that ST8 has subsurface enabled

nerites:
Fix in response to bisection on my earlier ST8 PR. 
Confirmed in the test scene that ST8 appearance is intended to change (due to subsurface scattering only being enabled on the rightmost tree) while ST7 appearance stays the same (since that shader was not modified).

Slack thread here: https://unity.slack.com/archives/CFZERNTS7/p1623056781144000


theo-pnv:
Rerunning the failing/canceling tests as some timed out.

theo-pnv (approved):

PR #4826, opened 2021-06-07
Fixed an epsilon in the Hue node that was larger than it should be, g…

joshua-davis:
…iven that it was only being used to prevent a zero division. Fixes fogbugs 1340849.

* Read the [Graphics repository & Yamato FAQ](http://go/graphicsQuad_e4yamato-faq).

### Checklist for PR maker
---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1340849/

There is an edge case where the hue node would produce very glaring seams when the transformation should've just produced the input data. This is easy to see on any sphere by checking around the negative octant:
OriginalSphere

There's multiple reasons for this.

The first issue is easy to fix. In order to avoid a zero division, the conversion from RGB to HSV uses an epsilon which is currently set to 1e-4. It's easy to see that the size of this epsilon contributes directly to the seam. Pictured below is the uvs of a quad converted to spherical coordinates and the passed through the Hue Node:

| Epsilon: | 1e-2 | 1e-4 | 1e-10 |
| - | - | - | - |
|  | Quad_e2 | Quad_e4 | QuadBaseColorFixed |

You may have to click the 1e-4 case to see the seam.
This epsilon is used purely to prevent a zero division and avoid a branch. To prevent a zero division you want to add a small number that's unlikely to affect the total result but will avoid a zero division. FLT_EPS is around 1e-7 which is the smallest number you can add to 1 and not get 1. FLT_MIN is the smallest number you can add to 0 and not get 0. The number we should add is in-between these two. 1e-10 was chosen for how it looks and because HALF_MIN is 1e-14, so the value works for both precisions.

Even after this, there was an edge case which is very precise. This is easiest to see on a cube when using the orientation gizmo to align down a negative axis:
CubeEdgeCase
The issue here is due to value becoming zero when the input vector was not zero, causing the hsv to rgb conversion to fail due to hsv.z being zero (value):
```
Out = hsv.z * lerp(K2.xxx, saturate(P2 - K2.xxx), hsv.y);
```
To get around this, epsilon is added to value unless the input vector was zero (which can be inferred from D) as pointed out by Landon.

---
### Testing status
Manually tested various shapes and values:
- Testing all of the object positions of a sphere
- Testing all of the object positions of a box
- Tested all uv coordinates on a plane converted to spherical coordinates (for easy visualization)

---


joshua-davis:
Linux failures are known failures currently on master

jessebarker (approved):
LGTM

hkr:
@Unity-Technologies/mobile-qa Please test on iOS and also on Android/Mali and Android/Adreno with OpenGL ES 3.x and Vulkan.
Please also check BoatAttack demo. There was an issue in the past where the Hua node would cause NaN on some object that use it in this demo. IIRC it's used to generate colors for the wind surfers and other assets.

LandonTownsendUnity:
The grey seams can be fixed by adding epsilon to the value when converting to HSV
image
image
image

LandonTownsendUnity:
If we want to make sure that plugging the zero vector (0, 0, 0) into the node always returns the zero vector, to prevent possible regressions, this code will work
image

LandonTownsendUnity:
Can you add the bug the new code fixes to the changelog? Something like "Fixed a bug with the Hue node where colors with negative values whose max component is 0 would change to the zero vector"
It's not exactly the same bug as the negative values one so I want to make sure both bugs are in the changelog

ghost (approved):

evelinastaniulyte:
Mobile testing progress can be observed in this document: https://docs.google.com/document/d/1EPqBskNSAaCQoBxLaKYP4VmwL7xR3WcsZskdF_ulaFE/edit#

evelinastaniulyte (approved):
No PR related issues were found during testing.

Tested projects: Boat Attack and project from [Case 1340849](https://fogbugz.unity3d.com/f/cases/1340849/)

Tested devices:
- Samsung Galaxy Note8 (SM-N950F), Android 7.1.1, CPU: Exynos 9 Octa 8895, GPU: Mali-G71
- Samsung Galaxy Note20 Ultra 5G (SM-N986B), Android 10, CPU: Exynos 990, GPU: Mali-G77
- Samsung Galaxy J5(2016) (SM-J510FN), Android 7.1.1, CPU: Snapdragon 410 MSM8916, GPU: Adreno (TM) 306
- Samsung Galaxy S8 (SM-G950U), Android 7.0, CPU: Snapdragon 835 MSM8998, GPU: Adreno (TM) 540
- Motorola moto g(7) power (moto g(7) power), Android 9, CPU: Snapdragon 625 MSM8953, GPU: Adreno (TM) 506
- Huawei Y6p, CPU: MediaTek MT6762R, GPU: PowerVR Rogue GE8320, Android: 10
- Google Pixel 4, Android 10, CPU: Snapdragon 855 (SM8150), GPU: Adreno (TM) 640
- Samsung Galaxy S10e (SM-G970F), Android 10, CPU: Exynos 9 9820, GPU: Mali-G76
- iPhone X, A11, 14.0.1

Testing doc: https://docs.google.com/document/d/1EPqBskNSAaCQoBxLaKYP4VmwL7xR3WcsZskdF_ulaFE/edit#

joshua-davis:
Passing green run before changelog conflicts: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Fhue-seam/.yamato%252Fall-shadergraph.yml%2523Nightly_ShaderGraph_trunk/7454203/job

PR #4833, opened 2021-06-08
Fixed aliasing artifacts that are related to numerical imprecisions of the light rays in the volumetric clouds (case 1340731).

anisunity:
https://fogbugz.unity3d.com/f/cases/1340731/
Due to the large scale of the earth, the ray-sphere intersection can lead to precision issues that cause some aliasing. This is a way to avoid those imprecisions.

Testing status
The issue does not repro anymore on the dataset shared by the user.
Volumetric cloud tests are still green locally.


RemyUnity (approved):
Next time a before/after screenshots in the description would be nice, but it seems ok for approval.

PR #4835, opened 2021-06-08
[URP] Fix usage of 'this' during OnAfterDeserialize()

unity-chris:
---
### Purpose of this PR

This PR fixes an issue uncovered while making this trunk PR : https://ono.unity3d.com/unity/unity/pull-request/126367/_/assetdatabase/mainobject/loaded-asset-management

That PR is an overhaul of how AssetDatabase.Refresh() handles objects that are loaded during its lifetime, and makes sure that objects that were previously unloaded are still unloaded, and those that were previously loaded are still loaded.

Previously, we were not correctly 'tidying up' the object used as a this-pointer in OnAfterDeserialize(), and it would still be available after the refresh. This meant that the object would still be available in the EditorApplication.delayCall, and all would be well.

After this trunk change, the object is no longer available, and a later call to EditorUtility.SetDirty() will throw a null argument exception.

The simplest/safest fix for now is to adapt the existing code to use InstanceIDs instead. Longer term, we might want to consider a different approach.

The relationship between paths and InstanceIDs is preserved by the PersistentManager, even when objects are unloaded, and InstanceIDs are valid for the current session, which should support this use-case.

---
### Testing status

I have used these changes locally on the 'TestCleanConsole' test that was failing on my branch, and it now passes
perl utr.pl --suite=integration --testtarget=TestVeryVerySlowEditorOnly --fail-on-assert --testfilter=Unity\.IntegrationTests\.EditorApplicationTests\.CleanStartup\.com\.unity\.template\.universal 

---
### Comments to reviewers

The corresponding trunk PR is considered critical for release, so the plan is to land the changes as part of the embedded trunk version of the code, after the current SRP update PR lands, and land this PR in parallel, ready for the next SRP update to trunk

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
@@ -1117,14 +1117,16 @@ namespace UnityEngine.Rendering.Universal
 #if UNITY_EDITOR
             if (k_AssetPreviousVersion != k_AssetVersion)
             {
-                EditorApplication.delayCall += () => UpgradeAsset(this);
+                EditorApplication.delayCall += () => UpgradeAsset(this.GetInstanceID());
             }
 #endif
         }
 
 #if UNITY_EDITOR
-        static void UpgradeAsset(UniversalRenderPipelineAsset asset)
+        static void UpgradeAsset(int assetInstanceID)
         {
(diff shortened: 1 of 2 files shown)


simon-engelbrecht-soerensen (approved):

unity-chris:
The corresponding trunk PR has now landed: https://ono.unity3d.com/unity/unity/pull-request/126367/_/assetdatabase/mainobject/loaded-asset-management, with changes to the embedded version of the URP code as mentioned above, in this changeset https://ono.unity3d.com/unity/draft/changeset/9ac88c3cac72

Marking this PR as ready for review now..

PR #4836, opened 2021-06-08
Use non jittered projection in outline pass

adrien-de-tocqueville:
### Purpose of this PR

Partial fix for https://fogbugz.unity3d.com/f/cases/1304189/

Fix outline rendering when TAA is enabled on the scene view (needs always refresh enabled)

The fix from aras for outline was apparently not working anymore because outline shader user the jittered matrix, and it caused flickering
before:
Flicker
after:
NoFlicker

This PR does not fix gizmo flickering though
This doesn't seem possible by using the same method as the one for the outline, because changing the depth offset for intersecting geometry just moves the intersection elsewhere, but does not remove it. For outline it does, because the issue is z-fighting

---
### Testing status
Tested outline rendering for AxF, Unlit, Lit, LayeredLit, LitTessellation, LayeredLitTessellation, SG with/without tessellation

For tessellated shaders, we don't have access to the non jittered matrix frustum planes, so the outline tessellation is a bit different from the object tessellation in some case like this one, but it's minor i think
FlickerT


RemyUnity (approved):
LGTM

sebastienlagarde (approved):

PR #4839, opened 2021-06-08
Fix data mismatch in APV debug view

FrancescoC-unity:
Previous debug view had a few issue: 
 
- Imprecise sample position, this can lead to fetching the wrong data from the atlas. 
- For same imprecision the outer borders of the cells can end up sampling ambient probe for the debug probes, which is of course not what we want. 

This is instead grabbing the information using directly the fixed position in atlas for each probe to get the SH info, leading to very precise data being sampled. 

To do so I needed to add some things and rewrite a chunk of the work done in prepping debug data, but all seems fine now (Tested with cases discussed with Jesper and Torbjorn last week).


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -293,6 +298,8 @@ public override bool Equals(object obj)
         Dictionary<RegId, List<Chunk>>  m_Registry = new Dictionary<RegId, List<Chunk>>();
 
         internal Dictionary<int, Cell> cells = new Dictionary<int, Cell>();
+        private Dictionary<int, CellChunkInfo> chunkInfo = new Dictionary<int, CellChunkInfo>();
```
JulienIgnace-Unity, line 301: Private member should be named "m_Stuff" and private keyword is not necessary here.

FrancescoC-unity: yeah but was internal, I added private as a test to see if I could've made it private but then forgot to change the naming and remove the keyword. Will change this.

alelievr (approved):

JulienIgnace-Unity (approved):

PR #4845, opened 2021-06-08
[Fogbugz 1341576] Memory leaks when the player is paused, and the user changes pipeline…

kecho:
### Purpose of this PR
Toggling any pipeline settings while having the player paused causes memory leaks on the GPU resources.
Fogbugz: https://fogbugz.unity3d.com/f/cases/1341576/
Fixed by detecting if we are paused and using the correct API to delete render target objects.

---
### Testing status
Followed the use case in the fogbogz:
* Tested also on dx11, vulkan and dx12 on PC
* Checked that I didnt crash
* Checked task manager, GPU memory usage and saw that GPU memory is now flat.

### DIFF
--- a/com.unity.render-pipelines.core/CHANGELOG.md
+++ b/com.unity.render-pipelines.core/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed issue displaying a warning of different probe reference volume profiles even when they are equivalent.
 - Fixed missing increment/decrement controls from DebugUIIntField & DebugUIUIntField widget prefabs.
 - Fixed IES Importer related to new API on core.
+- Fixed memory leak when changing SRP pipeline settings, and having the player in pause mode.
 
 ### Added
 - Support for the PlayStation 5 platform has been added.


FrancescoC-unity (approved):

iM0ve (approved):
Looks safe

hknielsen (approved):
From my point of view this LGTM, but I think personally I would still see if it was possible to cover with some test

kecho:
> From my point of view this LGTM, but I think personally I would still see if it was possible to cover with some test

Thanks for your review. Will instruct QA to do further testing.

kecho:
@iM0ve the bug is ready for some further testing:
* Do some stress testing on live editing while editor is played (on windows migth suffice).
* Add / remove objects, play unplay, try to see if we hit a crash.
* Try the usecase of playing and pausing, play pause multiple times while the editor is updating
* Test with multiple viewports
* Test with vulkan and dx11. Let me know if you find out anything else. If it looks solid @sebastienlagarde and you agree then I think we can ship this?

sebastienlagarde (approved):

TomasKiniulis (commented):
Tested Dx11 and Vulkan editing objects in playmode paused, unpaused, or going back to editmode. I get no leaks or crashes anymore. 

Adding assets in edit mode, playmode paused/unpaused:

https://user-images.githubusercontent.com/50582134/122027701-01afee80-cdd4-11eb-9ecc-2c8ed5b0a63a.mp4

Removing objects in editor playmode paused/unpaused:

https://user-images.githubusercontent.com/50582134/122027842-2015ea00-cdd4-11eb-8c45-646f2fe0fb55.mp4

Edit: Below occurs in master:
~However, toggling asset options like Low res Transparency in paused playmode sometimes ends up with a warning in console:
`Releasing render texture that is set to be RenderTexture.active!
UnityEngine.GUIUtility:ProcessEvent (int,intptr,bool&)
`~
~@kecho any idea if this can be resolved? Can't repro this in master, since it involves a lot of spam clicking and master almost immediately crashes.~

~Releasing Render Texture Warnings:~

https://user-images.githubusercontent.com/50582134/122028070-55bad300-cdd4-11eb-9527-b076db43a437.mp4

TomasKiniulis (approved):

PR #4848, opened 2021-06-08
Fix Resolution Issues for Physically Based Depth of Field

johnpars:
### Purpose of this PR
Fix: https://fogbugz.unity3d.com/f/cases/1340551/

This PR addressed two issues regarding buffer resolution and physically based depth of field:
1) Sampling artifacts at the edge of the viewport, when changing screen resolutions (from bigger -> smaller).
2) Varying results / DoF shallowness when changing resolutions / dynamic resolution. 

For 1, I fixed this by using a similar UV clamping strategy as was done for the normal depth of field path, when sampling the color pyramid. 

Before:
Before

After:
After

For 2, I addressed this by adding a resolution scaling term to the circle of confusion limit. I did this by fitting the term to a 1920x1080 reference resolution. 

(I thought .MP4 could embed into a PR, but I guess not)

Before:
https://user-images.githubusercontent.com/28882975/121268063-33343180-c88b-11eb-8709-d2c7b8c36f50.mp4

After:
https://user-images.githubusercontent.com/28882975/121268118-4cd57900-c88b-11eb-9261-b9bd4ae0384f.mp4

---
### Testing status
I manually confirmed that the artifacts no longer produce after changing between various resolutions.

I have not yet locally run the DoF tests to confirm how the resolution independence term may impact the result, but I will update the PR when I do this.

---
### Comments to reviewers
I may have misunderstood the scope of the fogbugz ticket by adding the resolution independence part. It is true that without it, you will get wildly differing results by changing resolution (especially dynamic resolution with, say, 25% scale), but I also am concerned that it is "too late" to add a change like this, since fitting the scale to a reference resolution (1920x1080) would change results for users in existing projects.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -245,6 +245,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed cropping issue with the compositor camera bridge (case 1340549).
 - Fixed the transparent cutoff not working properly in semi-transparent and color shadows (case 1340234).
 - Fixed object outline flickering with TAA.
+- Fixed issue with sky settings being ignored when using the recorder and path tracing (case 1340507).
 - Fixed some resolution aliasing for physically based depth of field (case 1340551).
 - Fixed an issue with resolution dependence for physically based depth of field.
 
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
@@ -160,17 +160,8 @@ namespace UnityEngine.Rendering.HighDefinition
 
 #endif // UNITY_EDITOR
 
-        private void CheckDirtiness(HDCamera hdCamera)
+        private void CheckDirtiness(HDCamera hdCamera, int camID, CameraData camData)
         {
-            if (m_SubFrameManager.isRecording)
-            {
-                return;
-            }
-
-            // Grab the cached data for the current camera
-            int camID = hdCamera.camera.GetInstanceID();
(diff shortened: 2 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFGather.compute
```diff
@@ -34,7 +34,6 @@ RW_TEXTURE2D_X(CTYPE, _OutputTexture);
 // A set of Defines to fine-tune the algorithm
 #define ADAPTIVE_SAMPLING
 #define STRATIFY
-#define FORCE_POINT_SAMPLING
```
pmavridis, line 37: If I remember correctly, trilinear sampling can introduce some "leaking" between in-focus and out-of-focus regions, that's why I was forcing point sampling. Ideally this choice should be per tile (use trilinear only in tiles without in-focus pixels), but since we don't do that now, I think it's more safe to use point sampling.

pmavridis (approved):
Looks good to me, other than the comment on the point sampling. You might also have to generate new reference images too for the DoF test.

sebastienlagarde (approved):

TomasKiniulis (approved):
Looks good. Checked with multiple game views, forced dynamic resolution and resizing Windows player build. No artifacts are visible after resizing.

master:

https://user-images.githubusercontent.com/50582134/121888245-6dd11a80-cd20-11eb-89cf-0a37318f0897.mp4

PR editor:
https://user-images.githubusercontent.com/50582134/121888284-7a557300-cd20-11eb-934b-fb2a540dfd4a.mp4

PR player:
https://user-images.githubusercontent.com/50582134/121888312-83dedb00-cd20-11eb-9bc3-4753e1d5cdf5.mp4

johnpars:
Graphics test reference images updated to match the new resolution fit for 1080.

@pmavridis, I will re-request your review just to confirm that switch back from trilinear to point sampling checks out.

pmavridis (approved):

PR #4849, opened 2021-06-09
Set depth storeAction to Store when XR requires the depth to share with XR providers

thomas-zeng:
---
### Purpose of this PR
- Set depth storeAction to Store when XR requires the depth to share with XR providers

---
### Testing status
- QA
- ABV

---
### Comments to reviewers
- Handles the case where depth is required for XRDepthCopyPass
- Add `cameraData.xr.enabled` check to guard the XRCopyDepth logic


manuele-bonanno (approved):
LGTM

fabien-unity (approved):

simon-engelbrecht-soerensen (approved):

ellioman (approved):

PR #4850, opened 2021-06-09
[URP] MSAA Backbuffer Optimization when Intermediate render targets are created

manuele-bonanno:
### Purpose of this PR
if MSAA is enabled and if intermediate render targets are created, the final postprocessing/UI rendering is processed on top of a Metal back buffer with MSAA enabled. Please see the attached GPU trace screenshot for reference.

This PR disables/enables MSAA back buffer depending on whether intermediate render targets are created or not. 
This gives a good memory footprint gain (Around 30 MB for iPhone SE 1334x750 res with 4x sample count) and potentially little bit of performance increases for projects with a lot of UI quads rendered on top of the back buffer.

---
### Testing status
local testing (iOS + editor)
Yamato ABV

---
### Comments to reviewers

Before:
image

After:
image


com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -566,6 +566,39 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+            // The condition depends on whether the intermediate MSAA render targets has been created or not
+            if (cameraTargetDescriptor.msaaSamples > 1 && renderingToBackBufferTarget)
+            {
+                int qualitySettingsMsaaSampleCount = QualitySettings.antiAliasing > 0 ? QualitySettings.antiAliasing : 1;
+
+                if (createColorTexture)
+                {
+                    if (qualitySettingsMsaaSampleCount > 1)
+                    {
+                        QualitySettings.antiAliasing = 1;
```
phi-lira, line 581: Code looks fine to me, however changing QualitySettings.antiAliasing causes changes in user project and that's not ok to change serialised data without user interaction. This will cause a regression related to this bug https://fogbugz.unity3d.com/f/cases/1195272/
 
Is it possible to add API to control backbuffer format instead of overriding quality settings?

phi-lira (changes requested):

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -566,6 +566,39 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+#if ENABLE_VR && ENABLE_XR_MODULE
+                        XRSystem.UpdateMSAALevel(1);
```
thomas-zeng, line 583: I think we should not update XR MSAA level here as that could potentially trigger eye texture reallocation. Also in some cases MSAA eye texture is desired for final XR composition pass.

thomas-zeng (changes requested):

simon-engelbrecht-soerensen (approved):
Approved and added Mobile QA for specific mobile testing.

manuele-bonanno:
this cannot be merged because of Felipe comment and potential regressions. Investigations and work need to be done in order to remove the need to change the quality settings MSAA value

manuele-bonanno:
closing temporarily because of the comment above, until we decide about the trunk work that needs to be done

PR #4853, opened 2021-06-09
Cleanup some dead code in APV 

FrancescoC-unity:
There was a bunch of visual noise of functions not used or being in the wrong place.

This is a bit of a cleanup.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
+++ b/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
@@ -500,29 +500,6 @@ namespace UnityEngine.Experimental.Rendering
             Clear();
         }
 
-        static float CalculateSurfaceArea(Matrix4x4 transform, Mesh mesh)
-        {
-            var triangles = mesh.triangles;
-            var vertices = mesh.vertices;
-
-            for (int i = 0; i < vertices.Length; ++i)
-            {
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -664,6 +641,47 @@ public static ProbeSubdivisionResult BakeBricks(ProbeSubdivisionContext ctx)
             return result;
         }
 
+        // Converts brick information into positional data at kBrickProbeCountPerDim * kBrickProbeCountPerDim * kBrickProbeCountPerDim resolution
+        internal static void ConvertBricksToPositions(List<Brick> bricks, Vector3[] outProbePositions)
+        {
+            Matrix4x4 m = ProbeReferenceVolume.instance.GetRefSpaceToWS();
```
alelievr, line 647: Do we still need that?

alelievr (approved):

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
FrancescoC-unity: Not really for now, but because we need sooner rather than later moving the origin too I'd rather keep that there for now so the transition is easier (same reason why the transform is a bit everywhere)

PR #4856, opened 2021-06-09
[HDRP][Path Tracing] Sky settings now properly taken into account when using recorder

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1340507

Before: When recording (using the recorder package) in path tracing mode, sky settings would be ignored (sky was always set to clear color).
Now: Settings are properly taken into account.

This has been tested on the HDRP template (which was the scene used as an example in the accompanying ticket).


remi-chapelain (approved):
I'm trusting what has already been tested. ✔️ 
Will be tested in exploratory testing phase later this week using the bugfix branch.

sebastienlagarde (approved):

sebastienlagarde:
@eturquin can you run Yamato DXR for this change? If there is no change in test, it mean we should add one test :)

eturquin:
> @eturquin can you run Yamato DXR for this change? If there is no change in test, it mean we should add one test :)

The issue was only when using recorder, I'm not sure how we could test that?

sebastienlagarde:
> > @eturquin can you run Yamato DXR for this change? If there is no change in test, it mean we should add one test :)
> 
> The issue was only when using recorder, I'm not sure how we could test that?

ah ok, fine thne

PR #4860, opened 2021-06-09
[HDRP] Fix custom pass volume not executed in scene view

alelievr:
### Purpose of this PR
Fixed an issue where we can't see the custom pass effect if the custom pass volume layer is anything else than `Default`.

This happens because we can't easily change the default volume layer mask of the scene view camera, so it was impossible to see a custom pass volume effect in the scene view if no camera in the scene had the same volume layer mask than the custom pass.

Issue reported here:
https://unity.slack.com/archives/CC68SH5SB/p1623235154019000

---
### Testing status
Tested with a simple Blur custom pass with a custom pass volume on a layer not included in the camera volume layer mask: as expected it renders in the scene view (that was not the case before).

https://user-images.githubusercontent.com/6877923/121363765-85329100-c937-11eb-87bf-8e5aee1bcb9f.mov


pmavridis (approved):
Looks good to me!

I'm a bit concerned that projects that use a full screen blur pass for off-screen cameras will get an unusable/blurry scene view, but since the workaround is trivial (keep the volume disabled while editing), I don't think it's a big issue.

alelievr:
Yes, it only affects custom passes in scene view that were on a different GameObject Layer than Default. Also there is an API in the custom pass code to disable rendering in scene view:
https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@12.0/api/UnityEngine.Rendering.HighDefinition.CustomPass.html#UnityEngine_Rendering_HighDefinition_CustomPass_executeInSceneView

sebastienlagarde (approved):

PR #4864, opened 2021-06-09
[2021.2] Updating decal reference images for android vulkan

lukaschod:
### Purpose of this PR
Updating reference images for failing CI.


phi-lira:
The reference image on decals looks completely broken. Why is that?

PR #4865, opened 2021-06-09
[HDRP][VT] Allow VT to use the new inline anisotropic sampler

debaetsd:
### Purpose of this PR

This PR changes how VT declares the samplers used for sampling the GPU cache. We now use a single inline sampler for all caches (rather then 1 per layer). This reduces sampler pressure when having multiple VT stacks in a material.

**This requires a trunk of at least 2021.2.0a21.2890_1c5963d60e8d**

### Testing status
inline anisotropic sampler is tested in (trunk) gfx tests.
If HDRP VT tests do not break, it is working


debaetsd:
cc @robin-demoor

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -199,13 +177,13 @@ float4 SampleVT_##layerSamplerName(StackInfo info, int lodCalculation, int quali
         vtProperty.layerIndex[3] = layer3Index; \
         \
         vtProperty.cacheLayer[0].TextureArray = stackName##_c##layer0Index; \
-        vtProperty.cacheLayer[0].Sampler = sampler##stackName##_c##layer0Index;\
+        vtProperty.cacheLayer[0].Sampler = cacheSampler;\
```
cdxntchou, line 180: Is it possible to declare VT_SAMPLER in the file above, and use it directly here, instead of taking it as a parameter?

cdxntchou (changes requested):

robin-demoor (approved):

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
debaetsd: good idea, this can indeed be a lot simpler.
Pushed an updated version

cdxntchou (approved):

PR #4452, opened 2021-05-05
[2021.2][SG] Blackboard Categories 

Nightmask3:
---
### Purpose of this PR

This PR adds the ability to create categories in the blackboard, that can then be used to organize your blackboard items, both within the shader graph, and in the Material Inspector, for materials created based on a shader graph.

This branch is a staging branch, where each individual feature that composes the categories as a whole was developed on a separate feature branch that was reviewed before being merged into the staging branch. 

There has also been QA testing of the features as they were progressively added, so some of this code has had QA coverage prior to this PR being created. 

This feature can be broken down into 2 parts, at a high level:
- The code particular to the categories UI stuff and data management in the shader graph
- The code that handles the material inspector representation of the shader graph categories and properties/keywords

This PR also adds code that changes how shader graph properties are rendered across URP and HDRP for shader graph based materials, inserting a custom property drawer (see `ShaderGraphPropertyDrawers`) that overrides the default material inspector handling of shader graph properties, allowing for a custom UI field representation of:
- Vector2/Vector3/Vector4s, where before they were all represented by a single vec4 field 
- Virtual Textures, which now indent the layers that compose the virtual texture property into a foldout that can be collapsed and expanded
-  The categories and the properties/keywords contained within them.

The categories working in a URP project:

https://user-images.githubusercontent.com/13033542/117181039-ad076580-ad89-11eb-8133-c2bd27cfb651.mp4

The categories working in a HDRP project

https://user-images.githubusercontent.com/13033542/117186031-4e44ea80-ad8f-11eb-92e7-c7493a3f519d.mp4

Link to the QA test doc by Landon: https://docs.google.com/document/d/1iwLU5TM3WbQCKRzm1iSMQAlO1SjVkZv74nK5JlLLvJQ/edit

Link to the spec doc **(which has details of the expected user interactions, might be useful for testing)**: https://docs.google.com/document/d/1l87jXXVkVTgDFA-OnKK5Je4BPw0_C5QcCL8uaTwYyaM/edit#heading=h.9a8vlbymtc9e

Link to the Jira card: https://jira.unity3d.com/browse/GSG-258

---
### Testing status

- QA testing performed by Landon over the course of development, as well as on this PR specifically 
- Tech-Artist testing being performed by some volunteers organized by Cici and Tracy, details of which can be found in this Slack channel: https://unity.slack.com/archives/C01T63JMB2T
[...]

PR #4870, opened 2021-06-10
[2021.2][SG] Cleanup Blackboard Categories

Nightmask3:
---
### Purpose of this PR

This PR implements a number of fixes to the shader graph blackboard categories feature that was recently checked into master by this PR : #4452 

These are issues that were reported by an external partner as well as some issues that were raised by user testing after the previous PR landed.

The issues fixed by this PR:

image4

- Renaming a category would also select all children within that category, interfering with future operations such as adding another category

Fixed behavior:

https://user-images.githubusercontent.com/13033542/121451017-a011ff00-c951-11eb-81bb-e9a6f98fffa7.mp4

image1

- Was pointed out that its a bit hard to notice the drag indicator when its at the bottom of a category, this is made more prominent now

Fixed behavior: 

image

https://user-images.githubusercontent.com/13033542/121454678-19aceb80-c958-11eb-82d6-be8e1e8fd554.mp4

- An error introduced by the previous PR was an inability to use the Graph Settings menu in the Inspector when a category was selected due to it auto-switching to the node settings tab, this is fixed now

TQv4NhMbGm

- An issue when the ShaderGUI is called with a subset of properties – a thing that's common for custom shader GUIs, handle some properties that need specific handling and then call the ShaderGUI that would have been used originally with a filtered set of properties.
    - Integrated a PR submitted by an external partner: https://github.com/Unity-Technologies/Graphics/pull/4584 that fixes this by not throwing an exception when a property is not found, instead just continues iterating through the properties. 

- A ShaderGraph with no `ShaderGraphMetadata` asset breaks the Material inspector
    - This is fixed now, we simply just fall back on the previous material inspector behavior handling for shader properties if no `ShaderGraphMetadata` asset can be found

- Duplicating categories leaves selection in stuck state

Fixed behavior: 

https://user-images.githubusercontent.com/13033542/121452241-f1bb8900-c953-11eb-9c78-8ce8b08e9302.mp4

^^ Also fixed some other related issues with copy-pasting of properties/categories/etc where I realized the editor conventions of clearing current selection and then selecting the newly copied items post the duplication/paste operation were not being followed

- Moving properties around now has a slight insertion offset, doesn't match previous (expected) behaviour for where the insertion happens based on position above/below center of the hovered property

Fixed behavior: 

https://user-images.githubusercontent.com/13033542/121452334-20396400-c954-11eb-9f2b-4952ea161d11.mp4

^^ This issue also led me to discovering a host of other issues that existed with the handling of moving items within categories, where it was also not in line with the editor conventions set by the Scene Hierarchy list. We now follow those conventions exactly.

image

- Shader property order is messed up, not the same as in Blackboard

This was an issue introduced by the categories PR, and has been fixed now so that properties and keywords will remain in consistent order from the blackboard when getting inserted as property blocks into the shader. 

Bugs fixed by this PR:

- https://fogbugz.unity3d.com/f/cases/1337100/

---
### Testing status

- Testing all of the cases outlined above to ensure they no longer repro
- Extensive user testing across a variety of scenarios for inserting/moving items into and between categories:
    - Moving single items within a category, and save/load graph to ensure output is correct
    - Moving multiple consecutive items within a category, and save/load graph to ensure output is correct
    - Moving multiple non-consecutive items within a category, and save/load graph to ensure output is correct
    - Inserting single items from another category into a category, and save/load graph to ensure output is correct
    - Inserting multiple consecutive items from another category into a category, and save/load graph to ensure output is correct
    - Inserting multiple non-consecutive from another category into a category, and save/load graph to ensure output is correct
    - Combining the various above permutations for selections of items within the same category and items outside the category, to ensure that the order of items on the blackboard is persisted even after the move/insert operation
        - What this means is that, we are now sorting the items that are being moved/inserted such that 
             - the items are inserted into their destination category in such a way that the items that are within the category will always be inserted first
             - then items from other categories will be inserted after
                 - For items from other categories, their relative order across the blackboard will be preserved as they get inserted into the destination category
                 - This gif below should demonstrate the idea, notice that the items being inserted from the other categories, are retaining the order of the category that they come from, relative to the destination category, prior to this the items were inserted in the order that they were selected, which was not robust and inconsistent feeling as far as the UX goes. 

https://user-images.githubusercontent.com/13033542/121453981-e6b62800-c956-11eb-8d7c-0ce7c7de9fc1.mp4

---
### Comments to reviewers

Doesn't include a changelog because none of these are changes to anything that was user facing since the last release, just fixing stuff intended for the next release (i.e. 21.2).

### DIFF
--- a/com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
+++ b/com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
@@ -24,9 +24,9 @@ namespace UnityEditor.ShaderGraph
     // sure that all shader graphs get re-imported. Re-importing is required,
     // because the shader graph codegen is different for V2.
     // This ifdef can be removed once V2 is the only option.
-    [ScriptedImporter(121, Extension, -902)]
+    [ScriptedImporter(122, Extension, -902)]
 #else
-    [ScriptedImporter(50, Extension, -903)]
+    [ScriptedImporter(51, Extension, -902)]
 #endif
 
     class ShaderGraphImporter : ScriptedImporter


GrantLamb-Unity (approved):
Thank you for the before and after videos for these bugfixes!

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
```diff
@@ -26,7 +26,7 @@ namespace UnityEditor.ShaderGraph
     // This ifdef can be removed once V2 is the only option.
     [ScriptedImporter(115, Extension, -902)]
 #else
-    [ScriptedImporter(48, Extension, -902)]
+    [ScriptedImporter(49, Extension, -902)]
```
marctem, line 29: Why only update one of these numbers? You need to update both if the serialization actually requires a reimport on upgrade.

marctem (changes requested):
One potential issue with the importer and then I think it's good to land!

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
elizabeth-legros: correct, both should be updated. My mistake :)

Nightmask3: Thanks Liz! Will amend this 👍

Nightmask3:
Test run with most jobs passing, one test failure and formatting failure: 
image

Ran the formatter (pass of the formatter test after my last commit can be seen here on Github)

And also fixed the sole test failure and reran that job, which succeeded:
image

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
marctem: Ah, no, you changed the entirely wrong thing. you need:

```
 [ScriptedImporter(116, Extension, -902)]
#else
    [ScriptedImporter(49, Extension, -902)]
```

The first number is the call is the version, unique to SG, which must be bumped by 1 independently for each branch of the #if. The second number is the global importer order which should be the same for both branches of the #if, and should not change.

Nightmask3: Ahh that makes sense 👍 
Currently in master the state of the Importer parameters is as such:

```
    [ScriptedImporter(121, Extension, -902)]
#else
    [ScriptedImporter(50, Extension, -902)]
#endif
```

post the merge into this branch and bumping the appropriate version numbers: 

```
    [ScriptedImporter(122, Extension, -902)]
#else
    [ScriptedImporter(51, Extension, -902)]
```

marctem (approved):

PR #4875, opened 2021-06-10
Fixed issue that caused MSAA to always be disabled 

phi-lira:
### Purpose of this PR
 Fixed issue that caused MSAA to be disabled on cameras that didn't had `AdditionalCameraData`

---
### Testing status
  Issue was discovered in the QualitySettings upgrade PR. I tested that when URP doesn't have AdditionalCameraData MSAA behaves as expected now.

---
### Comments to reviewers
  We should actually guarantee that all cameras and lights have additional camera data. I wanted to push a fix that is easy/low risk to backport. Plan to work on follow up PR to always add AdditionalCameraData.


hdb-unity (approved):
This fixes the discussed issue (found in testing Quality Settings converter PR).
Nice changelog entry to accompany the change.

Note: Did not do regression checks for this change.

I think this change should be mentioned in the upgrade guide too.

Verasl (approved):

PR #4878, opened 2021-06-10
fix to read from _DepthTexure as _CameraDepthTexture not bound to dep…

peterjohnlong:
This PR changes changes the depth read within RayMarching.hlsl to use _DepthTexure rather than _CameraDepthTexture, as this is consistent with the raymarch compute shader that includes this file, and the binding in the corresponding .cs source setup. On some consoles where canReadBoundDepthBuffer is true the CameraDepthTexture is not bound to the depth pyramid/atlas. 
Tested locally that it fixes issues for example HDRP_DXR test 309_GlobalIlluminationHybridFullRes running on such a console platform.


anisunity (approved):
Good catch, LGTM, we just need to make sure the test pass on yamato

sebastienlagarde (approved):

PR #4879, opened 2021-06-10
[HDRP] Realtime subdivision improvements

alelievr:
### Purpose of this PR
Replaced the real-time subdivision budget with a progressive update system. There are now two settings to control the speed of real-time subdivision:
- Number of cell updated per frame
- Delay between two updates in seconds

This ensures that a fixed number of cells are updated each frame with a delay between two full updates of 1 second by default.
The culling is still here, as you can see in the video, each cell culling step is performed just before starting to update the cells, so the update frequency here is 1 second + time for all the cells to be updated.

image

---
### Testing status

https://user-images.githubusercontent.com/6877923/121536194-23d7f400-ca03-11eb-818d-fdd10f94409f.mov

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.Debug.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.Debug.cs
@@ -242,64 +242,53 @@ namespace UnityEngine.Experimental.Rendering
                     continue;
 
                 float largestBrickSize = cell.bricks.Count == 0 ? 0 : cell.bricks[0].subdivisionLevel;
-
                 List<Matrix4x4[]> probeBuffers = new List<Matrix4x4[]>();
                 List<MaterialPropertyBlock> props = new List<MaterialPropertyBlock>();
-                List<int[]> probeMaps = new List<int[]>();
+                CellChunkInfo chunks;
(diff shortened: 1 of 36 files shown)


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.Debug.cs
```diff
@@ -18,7 +18,9 @@ class ProbeVolumeDebug
         public bool drawBricks;
         public bool drawCells;
         public bool realtimeSubdivision;
-        public float realtimeSubdivisionBudget = 100.0f;
+        public int subdivisionCellUpdatePerFrame = 4;
+        public float subdivisionDelayInSeconds = 1;
+        // public float realtimeSubdivisionBudget = 100.0f;
```
FrancescoC-unity, line 23: Can remove?

FrancescoC-unity (approved):

JulienIgnace-Unity (approved):

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.Debug.cs
alelievr: done

PR #4610, opened 2021-05-21
[2021.2][URP][ShaderGraph] Enable automatic render queue control on SG-based material inspector

jessebarker:
### Checklist for PR maker
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1335795/

The material inspector behaves differently for handwritten shader-based materials than it does for ShaderGraph based materials.  The handwritten shaders automatically force the setting of the render queue based on the surface type property, whereas the ShaderGraph materials allow for independent control of the render queue.  This PR adds a new "Queue Control" on the material inspector with "Auto" and "UserOverride" options.  The "Auto" option is selected by default and behaves the same way as a material created from the handwritten URP shaders (Lit.shader, Unlit.shader).  The "UserOverride" option allows the current behavior of independent control over surface type and render queue.  When "Auto" is selected, there is no Render Queue field (as with handwritten shaders).

The UI had a bit of a review after the original screen shot below in the testing status section.  Currently, a "Queue Control" of "UserOverride" enables the "Render Queue" field, whereas "Auto" does not.  "Sorting Priority" is always present as it is relevant regardless of where the render queue setting comes from.  See these material inspector shots for detail:

Screen Shot 2021-06-09 at 6 19 59 AM
Screen Shot 2021-06-09 at 6 19 15 AM

It is perhaps worth mentioning that those two materials and shadergraphs were created with 2021.1 and dropped into a project that is using this branch (so they've been upgraded and not just created by these changes).

---
### Testing status
Created a scene with 3 objects:
- A material created using Lit.shader
- A material created using a Lit Shader Graph
- A material created using an Unlit Shader Graph

See this image:
Screen Shot 2021-05-21 at 9 18 22 AM

Observed the following behaviors/test cases:
[...]

PR #4884, opened 2021-06-10
Fixed the built-in target to have collapsible menus to match other pipeline

joshua-davis:

### Checklist for PR maker

---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1339256/

The material inspector for shader graph built-in target shader didn't have collapsible menus like URP did. This required a new assembly reference on the srp.core/editor assembly which was deemed fine.
Additionally a small bug with sorting priority not working and not always being exposed was fixed.

---
### Testing status
- [x] View a built-in SG material without material override:
  - [x] The Surface Options section displays no properties
  - [x] All of the user's SG properties are in Surface Inputs section
  - [x] The advanced section contains the render queue (which is currently locked and can't be changed as intended)
  - [x] The advanced section contains the Sorting Priority (queue offset) which can be changed and updates the render queue accordingly
- [x] Change a built-in SG material to material override
  - [x] All of the same surface properties as before are in the Surface Options section
  - [x] All of the user's SG properties are in Surface Inputs section
  - [x] The advanced section contains the render queue (which is currently locked and can't be changed as intended)
  - [x] The advanced section contains the Sorting Priority (queue offset) which can be changed and updates the render queue accordingly
- [x] Verify that toggling On/Off material override updates the Surface Inputs section dynamically, even while the inspector has focus
- [x] Check all of the above for an unlit and lit target

---
### Comments to reviewers
Currently some options are duplicated in the Surface Options setting. This is being handled in another [PR](https://github.com/Unity-Technologies/Graphics/pull/4610)
"Enable GPU Instancing" wasn't added in the Advanced Options section as it doesn't currently work (I believe).
You can currently manually set the RenderQueue to "From Shader" which is fine.
An assembly reference to srp.core/editor was added to shader graph so `MaterialHeaderScopeList` can be used.


jessebarker (approved):
LGTM.  I'll follow this up with the "Queue Control" stuff I added to the URP material inspector.

GrantLamb-Unity:
Sufficient testing for the original bugfix. 

I have a couple questions for the secondary issue you found during this though.
> Additionally a small bug with sorting priority not working and not always being exposed was fixed.
Was this specifically just an issue for Built-In like the original bugfix? It kind of looks like @jessebarker is referencing he's fixing it in URP as well.
Is this something we could create Unit test(s) for?

joshua-davis:
> Sufficient testing for the original bugfix.
> 
> I have a couple questions for the secondary issue you found during this though.
> 
> > Additionally a small bug with sorting priority not working and not always being exposed was fixed.
> 
> Was this specifically just an issue for Built-In like the original bugfix? It kind of looks like @jessebarker is referencing he's fixing it in URP as well.
> Is this something we could create Unit test(s) for?

Yes this was just a bug in built-in. Jesse is fixing an independent issue in URP (which will also need to be fixed, but after my change). High level I had made the sorting priority field but it wasn't being used. Jesse is fixing how the render queue gets applied, either always from shader (which then uses the priority) or manually controlled.

I don't think we can create a test for this as it'd involve verifying that changing the sort priority on the inspector also changes the render queue value, but maybe I'm wrong. I haven't seen any tests for this kind of control yet.

jessebarker:
> > Was this specifically just an issue for Built-In like the original bugfix? It kind of looks like @jessebarker is referencing he's fixing it in URP as well.
> > Is this something we could create Unit test(s) for?
> 
> Yes this was just a bug in built-in. Jesse is fixing an independent issue in URP (which will also need to be fixed, but after my change). High level I had made the sorting priority field but it wasn't being used. Jesse is fixing how the render queue gets applied, either always from shader (which then uses the priority) or manually controlled.
> 

Right.  What I did for the URP shader GUI in !4610 + !4920 was fixed a bug where the material settings were sort of stepping on each other, so I enabled a new "queue control" setting to allow the user to choose between the hand-written (Lit.shader, Unlit.shader) shader behavior, and the flexibility of the shader graph behavior unambiguously.  Once this PR goes in, I can apply that same behavior to the built-in target and shader GUI so it will "match" URP (not in all actual rendering behavior due to pipeline differences, but in UX/UI at least).  In theory, Josh could do that on this PR, but it would make it much larger, and I think we'd prefer to keep them separate.

GrantLamb-Unity (approved):
> > > Was this specifically just an issue for Built-In like the original bugfix? It kind of looks like @jessebarker is referencing he's fixing it in URP as well.
> > > Is this something we could create Unit test(s) for?
> > 
> > 
> > Yes this was just a bug in built-in. Jesse is fixing an independent issue in URP (which will also need to be fixed, but after my change). High level I had made the sorting priority field but it wasn't being used. Jesse is fixing how the render queue gets applied, either always from shader (which then uses the priority) or manually controlled.
> 
> Right. What I did for the URP shader GUI in !4610 + !4920 was fixed a bug where the material settings were sort of stepping on each other, so I enabled a new "queue control" setting to allow the user to choose between the hand-written (Lit.shader, Unlit.shader) shader behavior, and the flexibility of the shader graph behavior unambiguously. Once this PR goes in, I can apply that same behavior to the built-in target and shader GUI so it will "match" URP (not in all actual rendering behavior due to pipeline differences, but in UX/UI at least). In theory, Josh could do that on this PR, but it would make it much larger, and I think we'd prefer to keep them separate.

Ah, I see! Thanks for the additional context. Also thank you for linking the other PR's as well, I took a look into them and read up a bit more and that cleared up what your discussions were about!

joshua-davis:
Passing green nightly run before merging to fix changelog: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Fbuiltin-material-inspector/.yamato%252Fall-shadergraph.yml%2523Nightly_ShaderGraph_trunk/7286748/job

PR #4890, opened 2021-06-14
Fix reflection probe tootltip

adrien-de-tocqueville:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1341990/

Changed the capture position tootltip.
Before:
image

After:
image

The behavior is not unexpected, so I think it's only a UX issue


TomasKiniulis (approved):
That seems to clarify the issue. Thanks!

sebastienlagarde (approved):

PR #4893, opened 2021-06-14
Fixed remapping of depth pyramid debug

JulienIgnace-Unity:
### Purpose of this PR
This PR improves depth pyramid visualization.
- Controls for remapping are now hidden if remapping is not enabled.
- Remapping as been reworked to have the same behavior as in render doc.
  Previously it would remap [near, far] to [minRange, maxRange] (the debug params) which meant that we were compressing all 
  world space values into a tiny range making effectively all values near the camera appear black. Changing max value would just 
  make far values appear grey instead of white.
  Now we remap [near*minValue, far*maxValue] to [0, 1] which means that users can choose what range they want to display in 
  shades of grey. To make the UI more consistent we use the [0, 1] range instead of [near, far] and use them as multipliers 
  (because the near/far values are camera dependent).

Fix for https://fogbugz.unity3d.com/f/cases/1342238/

---
### Testing status
- Verified that remap controls were hidden if not enabled.
- Tested the new remap scheme
- Tested that disabling remap would still display the full range of depth (as required by AOV)


sebastienlagarde (approved):

TomasKiniulis:
Thanks for the fix! It works well, one question though, shouldn't this fix also keep the visualization results same between "Enable Depth Remap" enabled and disabled? I guess that means changing default depth remapping values to the ones that are used when checkbox is disabled? But in my opinion that would still make more sense than having remap set to min 0 and max 1 as this way barely any depth gradient is actually visible even with near infinite scale meshes.

https://user-images.githubusercontent.com/50582134/121877099-ef21b080-cd12-11eb-844c-4df7d0db3cb2.mp4

TomasKiniulis (approved):

simon-engelbrecht-soerensen (approved):

PR #4895, opened 2021-06-14
Fix wobble-like (or tearing-like) SSAO issues when temporal reprojection is enabled.

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1341704/ 

The issue was due to the fact that history was loaded, so we might have got some discontinuities at the threshold between a texel and another being sampled (e.g. 10.49 and 10.51), because this was delicate it would end up looking as moving moving lines. 

Depending the motion characteristic, that ends up looking a bit like a wobble effect (as the motion settles down). 

To solve the issue I had to change the format used for the history texture (from UINT 32 to RGBA8_UNorm, so same size and bw cost) and change how history data is packed. Once this is done, it is possible to use a linear sampler to get the data and get rid of straight lines and the wobble. 

There was also a bug with AOBufferSize being half a texel wrong if the source resolution was odd, leading to various visual imprecisions. This PR fixes that too. 

**Before**

https://user-images.githubusercontent.com/43168857/121886057-3f9a0d80-cd15-11eb-8909-5440306773fe.mp4

**After**

https://user-images.githubusercontent.com/43168857/121886187-6a846180-cd15-11eb-8951-e0263e5e1572.mp4


TomasKiniulis (approved):
Thanks for the fix and nice description! Testing looks sufficient. Approving

sebastienlagarde (approved):

PR #4898, opened 2021-06-14
[Cross-pipeline] HDRP Wizard window appears on opening a new URP project if both pipelines are present.

alex-vazquez:
---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1336296/
---
### Testing status

- Check that the wizard is not shown when opening the editor with URP as current pipeline
- Check that the first time that the user changes to HDRP the wizard is shown
- Check that the HDRP Wizard can not be opended if HDRP is not the current pipeline

image

---
### Comments to reviewers
I closed the editor with visual studio and I was not able to show the Wizard because the setting `HDUserSettings.wizardPopupAlreadyShownOnce` was never set to false. Fixed that issue in this PR, adding a comment in code to explain the issue and why is that setting set there.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -78,7 +78,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed Intensity Multiplier not affecting realtime global illumination.
 - Fixed an exception when opening the color picker in the material UI (case 1307143).
 - Fixed lights shadow frustum near and far planes.
-- Fixed that the HDRP Wizard is only open when a SRP in use is of type HDRenderPipeline.
+- The HDRP Wizard is only opened when a SRP in use is of type HDRenderPipeline.
 - Fixed various issues with non-temporal SSAO and rendergraph.
 - Fixed white flashes on camera cuts on volumetric fog.
 - Fixed light layer issue when performing editing on multiple lights.
--- a/com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
@@ -228,7 +228,7 @@ namespace UnityEditor.Rendering.HighDefinition
         static bool CanShowWizard()
         {
             // If the user has more than one SRP installed, only show the Wizard if the pipeline is HDRP
-            return RenderPipelineManager.currentPipeline.GetType() == typeof(HDRenderPipeline);
+            return HDRenderPipeline.isReady;
         }
 
         void OnGUI()


com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
```diff
@@ -277,6 +303,17 @@ static void CheckPersistencyPopupAlreadyOpened()
...
+            // If the wizard does not need to be shown at start up, do nothing.
+            if (!HDProjectSettings.wizardIsStartPopup)
+                return;
+
+            // Specify that the wizard has not been shown
+            // Set it here as the value stored on the settings might be wrong if Unity has stopped in scenarios like:
+            // - The user might stop the engine process.
+            // - The developer has ended the process from the engine side while attached
+            // - The editor crashed and that setting was not saved
+            HDUserSettings.wizardPopupAlreadyShownOnce = false;
```
RSlysz, line 315: The main intent of this boolean was to not reload script on domain reload.

Please check that domain reload will not trigger the pop up (when wizard is setup to pop at start)

RSlysz: If this is the case, we prefer having this small issue that only occures as a result of a crash.
Having the popup popping at each domain reload is a way higher issue.

alex-vazquez: Ok, I didnt know that, removed that line of code :)

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
```diff
@@ -246,22 +253,41 @@ static HDWizard()
...
+            {
+                // Delay the show of the wizard for the first time that the user is using HDRP
+                RenderPipelineManager.activeRenderPipelineTypeChanged += ShowWizardFirstTime;
+                return;
             }
+
+            // If we reach this point can be because
+            // - That the user started Unity with HDRP in use
+            // - That the SRP has changed to HDRP for the first time in the session
+            if (!HDUserSettings.wizardPopupAlreadyShownOnce)
```
RSlysz, line 289: HDProjectSettings.wizardIsStartPopup is not checked anymore.
Have you tested if this bevahe correctly when you check/uncheck the checkbox at the end of the window?

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
```diff
@@ -277,6 +303,10 @@ static void CheckPersistencyPopupAlreadyOpened()
         [Callbacks.DidReloadScripts]
         static void WizardBehaviour()
         {
+            // If the wizard does not need to be shown at start up, do nothing.
+            if (!HDProjectSettings.wizardIsStartPopup)
```
alex-vazquez, line 307: @RSlysz, we do not need to do all the frame waiting if the wizard doesn't need to show on startup

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
alex-vazquez: Check my other comment

RSlysz (approved):
Right. Its starting to be a bit hard to follow as it is spread over 3 methods. Can you add a small comment on that? (HDProjectSettings.wizardIsStartPopup could only be true at the point I spoted above). Thanks

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -78,6 +78,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed Intensity Multiplier not affecting realtime global illumination.
 - Fixed an exception when opening the color picker in the material UI (case 1307143).
 - Fixed lights shadow frustum near and far planes.
+- Fixed that the HDRP Wizard is only open when a SRP in use is of type HDRenderPipeline.
```
jenniferd-unity, line 81: I would argue that we can only say "The HDRP Wizard is only opened when a SRP in use is of type HDRenderPipeline."

jenniferd-unity: (we are in the category Fixed)

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
```diff
@@ -224,6 +224,13 @@ static void OpenWindow()
             HDUserSettings.wizardPopupAlreadyShownOnce = true;
         }
 
+        [MenuItem("Window/Rendering/HDRP Wizard", priority = 10000, validate = true)]
+        static bool CanShowWizard()
+        {
+            // If the user has more than one SRP installed, only show the Wizard if the pipeline is HDRP
+            return RenderPipelineManager.currentPipeline.GetType() == typeof(HDRenderPipeline);
```
jenniferd-unity, line 231: Please use HDRenderPipeline.currentAsset != null or HDRenderPipeline.isReady

jenniferd-unity (changes requested):
Just a quick note about our log, but otherwise I would prefer that we use HDRenderPipeline APIs instead of GetType to improve readability

jenniferd-unity (approved):
Thanks for acting on suggestions too!

alex-vazquez:
@Unity-Technologies/gfx-qa-hdrp , could you please test
- If URP is the active pipeline is not HDRP the wizard is not opened ( when starting the editor )
- When chaning from URP to HDRP for the first time the wizard is opened
- If HDRP is the active pipeline, the wizard is shown ( when starting the editor )

TomasKiniulis (approved):
Checked listed requests on Graphics and Quality sections. Also checked that changing pipelines grays out or shows HDRP wizard section when assets are changed in Graphics or overriden in quality settings. Works as expected. Nice changes!

TomasKiniulis:
Looks like I've missed regression introduced by this PR. When completely empty project is created HDRP Wizard is not showing up after adding hdrp package, and the option is grayed out. Workaround is to create HDRP asset and assign it manually, then the wizard can be opened up manually.
image

PR #4903, opened 2021-06-14
For the time being use GGX LTC table for hair. 

FrancescoC-unity:
Because the generation of a proper Kajiya Kay spec LTC table is non trivial, we fall back on GGX for now. 

I decided to change only shader side as I think is cleaner to keep C# as is to make the modification easier when we'll have a proper table.


anisunity (approved):
So much code change, it is scary, but looks good to me.

sebastienlagarde (approved):

sebastienlagarde:
Note to self: merge master once hd/bugfix is merge

PR #4911, opened 2021-06-15
Fix LitTessellation heightmap displacement with planar and triplanar mapping

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1342302/ 

There was an issue with keywords that removed the planar and triplanar mapping calculations from domain shaders, which would of course lead to problems. 

This is fixed and is also adding a test with displacement with the various mapping. 

I am going to be running yamato to get the ref images. 

**Before**
beforeTess

**After**
afterTess

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -248,7 +248,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed volumetric fog in planar reflections.
 - Fixed error with motion blur and small render targets.
 - Fixed issue with on-demand directional shadow maps looking broken when a reflection probe is updated at the same time.
- Fixed cropping issue with the compositor camera bridge (case 1340549).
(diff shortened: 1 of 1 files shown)


FrancescoC-unity:
Note: I have to push an additional change for the vertex part.

FrancescoC-unity:
I need to make a different scene as the repro scene was not covering all issues and I misunderstood the degree of the issue. Code fix is correct, but more can be done in testing. Adding it now.

TomasKiniulis:
Testing is sufficient, but looks like yamato is completely failing, DX12 that actually manages to run tests skips this test scene.

sebastienlagarde (approved):

sebastienlagarde:
Currently DX12 yamato is good. All the issue are instability (like all the other builder), go to merge

sebastienlagarde:
ah sorry I see we need to regenerate pcitures, laaunching Yamato

FrancescoC-unity:
Argh yeah sorry forgot to re-kick the jobs myself as the one I triggered failed for unrelated reasons.

sebastienlagarde (approved):

PR #4912, opened 2021-06-15
Properly validate displacement mode on lit shader

adrien-de-tocqueville:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1342333/
https://fogbugz.unity3d.com/f/cases/1343374/

Filter displacement mode when validating material properties, so that lit tessellation cannot have pixel displacement when setting it in the .mat file, or when switching from lit shader

Also when switching back to the lit shader, the displacement mode is now preserved (if it wasn't change when in lit tessellation, like any other setting).

---
### Testing status
Tested that pixel displacement is disabled when switching from lit to lit tessellation
Tested that pixel displacement is disabled when enabling it in the .mat file for a LitTessellation material

Tested to switch back and forth between lit and lit tessellation and checked option is preserved in both cases
Search in all places in code using displacement mode an verified they still work as before (a warning in UV detail, and the heightmap settings)
Tested multi editing
Quickly tested with layeredlit (should be the same code so only did basic testing)


TomasKiniulis (approved):
Looks good. Testing is sufficient for the scope of fix. 

Thanks for bringing up the issue with Displacement Mode switching. I have opened up a ticket: https://fogbugz.unity3d.com/f/cases/1343374/ if anything can be done to keep the values

TomasKiniulis (approved):
Thanks for the Displacement Mode switching improvement, did a quick check switching through various HDRP and Shadergraph shaders and it works nicely.

sebastienlagarde (approved):

PR #4915, opened 2021-06-15
Fixed Volume Gizmo size when rescaling parent GameObject

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1342960/

volume gizmo scale now takes into account it's parent
VolumeGizmo

---
### Testing status
Tested on the repro project

When rotating the child gameobject, the scale is skewed like on the picture below
Tested that the displayed gizmo is actually the same as the box used by the volume system to detect if we are inside the volume
image


TomasKiniulis (approved):
Looks good. Thanks for the change!

PR #4921, opened 2021-06-15
Optimizing and Burstifying CPU of Decal Projection updates

kecho:
### Purpose of this PR
The purpose of this PR is to optimize and ultimately remove the expensive Update() (Editor only) and LateUpdate() (Editor + Runtime) functions in HDRP DecalProjectors.

HDRP Decal Projectors do their entire transform and update in the "LateUpdate()" function. For example if all decals move at the same time, this will incur into really heavy / non parallel and slow wall time.

This optimization entirely removes the LateUpdate() function, and ultimately creates a job that performs the exact same transformation. This job is also burstified.

A few extras in this PR:
* Cleanup of Array resizing, and a nicer common file to hold array utilities. Taken from URP
* Removal of the Update() function (editor only). Its not necessary since live editing of the layer mask can be done from the DecalProjectorEditor
* Adding burst and mathematics dependencies (similar to URP's)

Bug report: https://fogbugz.unity3d.com/f/cases/1253158/
---
### Testing status
* Did a basic performance test on PC, an intel i9 CPU. The test involved the HDRP template, selecting all the decals (their root object) and then moving them (through live edit) while the player & the profiler is running. Then I picked a frame and look at its contents.
image

Before we would get a total of 0.378ms on the LateUpdate marker of DecalProjector :
image

After my changes, this then becomes a total cost of 0ms (hidden in a job). And the job cost is 0.025ms
image

Recommended tests for qa:

* Stress test that live editing workflow is ok: as in adding / removing / and live editing the decals has the exact same behaviour as before this PR.
* Create a new scene, add a ton of decal projectors, say 600. Write a script that moves them very slow.
* Stress test changing the Layers of a decal projector
** Check frame rate on a ps4.
** Get this PR changes in and repeat test. Compare frame rate.
I dont recommend grabbing the A2 project, its too much work. Its faster to setup your own project and test the performance of this marker.
* Test transparency decals

---
### Comments to reviewers
We probably need to discuss if its ok to roll burst as a hdrp dependency (I mimiced what URP has as dependencies).
^We had a short discussion and this seams to be ok.


RSlysz (approved):
I think this is good.

I recommand also some test regarding not only scripting but also editor edition (and especially multi-edition) in case of.

com.unity.render-pipelines.high-definition/Runtime/Utilities/ArrayUtils.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using Unity.Collections;
+using UnityEngine.Jobs;
+
+namespace UnityEngine.Rendering.HighDefinition
+{
+    internal static class ArrayUtils
+    {
+        public static void ResizeNativeArray<T>(ref NativeArray<T> array, int capacity) where T : struct
```
sebastienlagarde, line 9: is it a copy from URP? in this case I suggest we move this one in Core and remove it from URP

kecho: Done, made this a public API on core.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
```diff
@@ -57,7 +57,7 @@ static void UpdateColorsInHandlesIfRequired()
         SerializedProperty m_FadeFactor;
         SerializedProperty m_DecalLayerMask;
 
-        int layerMask => (target as Component).gameObject.layer;
+        int layerMask => (target as DecalProjector).cachedEditorLayer;
```
RSlysz, line 60: Should you also fix the below layerMaskHasMultipleValue for coherency then?

RSlysz (changes requested):
This is good. But I like the small fix below to be done for coherency.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/DecalProjectorEditor.cs
kecho: Added the fix

RSlysz (approved):

kecho:
@sebastienlagarde the failures seam to steam from one of the previously seen instabilities. See your thread on slack (crash on editor side).
I am merging master again and running tests again. Otherwise this PR is ready to go (need QA pass)

com.unity.render-pipelines.core/Runtime/Utilities/ArrayUtils.cs
```diff
@@ -0,0 +1,62 @@
+using System;
+using Unity.Collections;
+using UnityEngine.Jobs;
+
+namespace UnityEngine.Rendering
+{
+    /// <summary>
+    /// Array utilities functions
+    /// </summary>
+    public static class ArrayUtils
```
arttu-peltonen, line 10: This seems useful. You could consider making them extension methods, so you could just say 
`decalToWorlds.Resize(newCapacity);` instead of  `ArrayUtils.ResizeNativeArray(ref decalToWorlds, newCapacity);`. Though on second thought I'm not certain if it's good practice to have such things in this namespace or if they would belong in a separate extension namespace instead.

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
```diff
@@ -38,6 +40,16 @@
...
+        },
+        {
+            "name" : "com.unity.mathematics",
+            "expression" : "1.2.1",
+            "define" : "ENABLE_MATHEMATICS_1_2_1"
+        },
+        {
+            "name": "com.unity.burst",
+            "expression": "1.5.0",
+            "define": "ENABLE_BURST_1_0_0_OR_NEWER"
```
arttu-peltonen, line 52: Is it 1.5.0 or 1.0.0? The expression should probably agree with the define name.

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalSystem.Jobs.cs
```diff
@@ -0,0 +1,291 @@
...
+            {
+                m_UpdateJobHandle.Complete();
+                var updateJob = new UpdateJob()
+                {
+                    positions = m_Positions,
+                    rawRotations = m_Rotations,
+                    rawScales = m_Scales,
+                    resolvedScales = m_ResolvedScales,
+                    resolvedRotations = m_ResolvedRotations,
+                    resolvedSizedOffsets = m_ResolvedSizeOffsets,
```
arttu-peltonen, line 142: Slight typo here? "sized" vs "size"

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
```diff
@@ -38,6 +40,16 @@
             "name": "com.unity.modules.nvidia",
             "expression": "1.0.0",
             "define": "ENABLE_NVIDIA_MODULE"
+        },
+        {
+            "name" : "com.unity.mathematics",
+            "expression" : "1.2.1",
+            "define" : "ENABLE_MATHEMATICS_1_2_1"
```
arttu-peltonen, line 47: Here you've added a define but are not using it in the code. Does the define need to exist in the first place? Maybe this is just fine, I'm just wondering what the logic in using these is.

arttu-peltonen (approved):
Looks great and seems like a great use case for jobs & burst, exciting to see this! Just a couple of small typos & suggestions.

Interesting to hear if this completely solves the issue in the original bug where LateUpdate() took >10ms.

com.unity.render-pipelines.core/Runtime/Utilities/ArrayUtils.cs
kecho: Hey so in c# you cant add methods / functions directly in a namespace. The way to do it is to have a static class, which is what I've done.
Let me know if this makes sense.

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
kecho: This is matching URP's define, also the define is 1_0_0_OR_NEWER, so this is newer, but its on par with what URP has.

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
kecho: I had to add this so the package could be referenced. Not being used in the code.  You know another way to have mathematics referenced without adding this?

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalSystem.Jobs.cs
kecho: Fixed!

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
arttu-peltonen: Ok, could be there's a mistake on URP as well. Still, if the define says "1_0_0_OR_NEWER", then the expression should be "1.0.0".  If the expression is "1.5.0", the define should read "1_5_0".

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
arttu-peltonen: I'm not sure tbh. If other defines are used in the same way, it's probably fine.

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
kecho: the way I undesrtood it was "_OR_NEWER" covered essentially anything above or equal to 1.0.0. Ill make the change but if I have to touch URP that might have to be done in a separate PR since it could break things / impact things. Ill do this change just on HDRP.

com.unity.render-pipelines.core/Runtime/Utilities/ArrayUtils.cs
kecho: Ah ok I had no idea this existed in c# :) 
The question is can it work with refs on structs (notice these functions replace the refs of the arrays)? If not then we cant use it. Ill try it now and see if it works.

kecho: @arttu-peltonen I was successful with refs of structs, but not refs of managed arrays. Thus, this is still gonna look like a static method syntax for managed arrays (since I have to instantiate inside this method lazilly). Otherwise native arrays and tansform array access are now with the syntax you suggested. Tested URP to ensure it didnt break.

kecho:
For URP QA: there should be no functional changes in URP, all I changed was renamed some functions that do string resizing. Feel free to regress decal behaviour in URP.

com.unity.render-pipelines.core/Runtime/Utilities/ArrayUtils.cs
arttu-peltonen: Ok good to know there's a limitation like that. Thanks for considering it!

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
arttu-peltonen: Ah I see. Yeah I'm fine with not touching URP right now. My understanding is that when you write `"expression": "1.5.0"`, it's equivalent to saying "enable this define if package version >= 1.5.0", and the define name should have the equivalent name - for this expression that's `..._1_5_0_OR_NEWER".

com.unity.render-pipelines.high-definition/Runtime/Unity.RenderPipelines.HighDefinition.Runtime.asmdef
sebastienlagarde: should be fine like this

sebastienlagarde (approved):

kecho:
I will undo the changes from URP, and instead do them in a separate PR. This is so we get this code in the wild quicker and maximize testing.

lukaschod (approved):
Noice!

LaserYGD:
Will this land in 2021.2?

kecho:
Seams like yamato found something new?
https://yamato-artifactviewer.prd.cds.internal.unity3d.com/19fd35fd-f100-46b9-bf31-dcc731667606%2Flogs%2FTestProjects%2FHDRP_RuntimeTests%2Ftest-results/TestReportV1.html
No idea if its caused by this, but its a test with decals. Will investigate

kecho:
Seams unrelated, I found the same issue in another branch with completely unrelated changes:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fdlss-taa-particle-ghost-fix/.yamato%252Fhdrp-win-dx11.yml%2523HDRP_Win_DX11_Standalone_XR_mono_Linear_trunk/7663620/job/artifacts

RemyUnity (approved):
Here are my testing notes : https://docs.google.com/document/d/1mN8qPuBJp33uCFFkGRyqyZUX1_DdN1zMfqjO6yiYqhg/edit?usp=sharing

Looking at the CPU profiler, updating 10000 decals went from 38.55ms late update to 2.65ms with burst jobs, all moving, and 10% changing the decal layer each frame.
@JMargevics Ran the same scene on PS4 and went from 3fps on master to 7fps with this branch, but will try to profile it.
Decals editing is also pretty responsive (if you ignore the editor stalling when trying to select hundreds of objects).

RemyUnity:
Also note that the jobs are not spread across the workers. Still faster than before, but could probably be better ?
image

JMargevics:
Hello! 

I've done a quick test on PS4 and here are the results:

Test decal scene on this PR: 

https://user-images.githubusercontent.com/50581102/125733219-550ab17c-9487-49e9-870e-c77d032b0234.mp4
7 fps.

Same scene on master:

https://user-images.githubusercontent.com/50581102/125733281-7022d905-d700-41c6-a555-961cbbb07842.mp4
4 fps

3 frames difference on paper however as you can see in reality burstified scene feels a lot smoother.

sebastienlagarde:
>4 fps

>3 frames difference on paper however as you can see in reality burstified scene feels a lot smoother

for your information, it is 140ms, so twice as fast which is very good. FPS is not a good metric to use :)

PR #4922, opened 2021-06-15
Quality Audit - Improvements on Volume Editor

alex-vazquez:
---
### Purpose of this PR
Related JIRA's

https://jira.unity3d.com/browse/XPIPELINE-219
https://jira.unity3d.com/browse/XPIPELINE-220

- Changing the help box and the button for a `FixMeHelpBox`
image
- Adding two new warnings:
image
image
- Fixing the wrapping of the `FixMeHelpBox` ( See the before and after in the picture bellow )
image
- When global volume is selected the Transform tools are hidden, states are recovered when the edition is end, or if the type of volume is set to local
- Adding shortcuts to the context menu of the Volume component
image

---
### Testing status
Check that the warnings and the Fix button work as expected in the following scenarios
- Local Volume without Collider
- Local Volume with Collider, but disabled
- Global Volume with Collider enabled
- Check that the short cuts work as expected

- Check that the `FixMeBox` is correct in other places that is being used
image

---
### Comments to reviewers
- I've cleaned some code and fixed some memory garbage by the GUIStyles

### DIFF
--- a/com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
+++ b/com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
@@ -1,5 +1,7 @@
+using System;
 using UnityEngine;
 using UnityEngine.Experimental.Rendering;
+using Object = UnityEngine.Object;
 
 namespace UnityEditor.Rendering
 {
@@ -25,6 +27,10 @@ namespace UnityEditor.Rendering
         static System.Lazy<GUIStyle> m_SmallMixedTickbox = new(() => new GUIStyle("ShurikenToggleMixed"));
         /// <summary>Style for a small checkbox in mixed state</summary>
         public static GUIStyle smallMixedTickbox => m_SmallMixedTickbox.Value;
+
+        static System.Lazy<GUIStyle> m_HelpBoxLabelStyle = new(() => new GUIStyle(EditorStyles.miniLabel) { wordWrap = true });
+        /// <summary>Style for a small checkbox in mixed state</summary>
+        internal static GUIStyle helpBoxLabelStyle => m_HelpBoxLabelStyle.Value;
 
         static GUIStyle m_MiniLabelButton;
         /// <summary>Style for a minilabel button</summary>
@@ -112,6 +118,28 @@ namespace UnityEditor.Rendering
 
         /// <summary>RenderPipeline Global Settings icon</summary>
         public static readonly Texture2D globalSettingsIcon;
+
+        /// <summary>
+        /// Gets the icon that describes the <see cref="MessageType"/>
+        /// </summary>
+        /// <param name="messageType">The <see cref="MessageType"/> to obtain the icon from</param>
+        /// <returns>a <see cref="Texture2D"/> with the icon for the <see cref="MessageType"/></returns>
+        internal static Texture2D GetMessageTypeIcon(MessageType messageType)
+        {
+            switch (messageType)
+            {
+                case MessageType.None:
+                    return null;
+                case MessageType.Info:
+                    return iconHelp;
+                case MessageType.Warning:
+                    return iconWarn;
+                case MessageType.Error:
+                    return iconFail;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(messageType), messageType, null);
+            }
+        }
 
         #endregion
 
--- a/com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
+++ b/com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
@@ -99,24 +99,35 @@ namespace UnityEditor.Rendering
 
         // UI Helpers
         /// <summary>Draw a help box with the Fix button.</summary>
+        /// <param name="message">The message with icon if need.</param>
+        /// <param name="action">When the user clicks the button, Unity performs this action.</param>
+        public static void DrawFixMeBox(GUIContent message, Action action)
+        {
+            GUILayout.Space(2);
+            using (new EditorGUILayout.HorizontalScope(EditorStyles.helpBox))
+            {
+                EditorGUILayout.LabelField(message, CoreEditorStyles.helpBoxLabelStyle);
+                GUILayout.FlexibleSpace();
+                using (new EditorGUILayout.VerticalScope())
+                {
+                    GUILayout.FlexibleSpace();
+
+                    if (GUILayout.Button("Fix", GUILayout.Width(60)))
+                        action();
+
+                    GUILayout.FlexibleSpace();
+                }
+            }
+            GUILayout.Space(5);
+        }
+
+        /// <summary>Draw a help box with the Fix button.</summary>
         /// <param name="text">The message text.</param>
         /// <param name="messageType">The type of the message.</param>
         /// <param name="action">When the user clicks the button, Unity performs this action.</param>
         public static void DrawFixMeBox(string text, MessageType messageType, Action action)
         {
-            EditorGUILayout.HelpBox(text, messageType);
-
-            GUILayout.Space(-32);
-            using (new EditorGUILayout.HorizontalScope())
-            {
-                GUILayout.FlexibleSpace();
-
-                if (GUILayout.Button("Fix", GUILayout.Width(60)))
-                    action();
-
-                GUILayout.Space(8);
-            }
-            GUILayout.Space(11);
+            DrawFixMeBox(EditorGUIUtility.TrTextContentWithIcon(text, CoreEditorStyles.GetMessageTypeIcon(messageType)), action);
         }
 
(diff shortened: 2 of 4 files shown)


com.unity.render-pipelines.core/Editor/Volume/VolumeEditor.cs
```diff
@@ -49,42 +78,100 @@ void RefreshEffectListEditor(VolumeProfile asset)
...
+                    CoreEditorUtils.DrawFixMeBox(Styles.addColliderFixMessage, AddOverride);
                 }
 
                 EditorGUILayout.PropertyField(m_BlendRadius);
                 m_BlendRadius.floatValue = Mathf.Max(m_BlendRadius.floatValue, 0f);
             }
+            else
+            {
+                // When the volume is global hide all the transform gizmos
+                Tools.current = Tool.None;
```
arttu-peltonen, line 167: Doing this not only hides transform gizmos, it also prevents their selection in the toolbar - so none of these are selected and clicking them does nothing:

image

For example, now you cannot use the View tool when a Global volume is selected. Is this intended? I'm not sure if "no active tool mode" is a well-defined state. The new overlays UI seems to allow it though (not sure if it was always allowed by the old toolbar UI as well).

arttu-peltonen (approved):
Looks good! I tested the UI and the transform tool disabling logic seems to work as intended. Just want to make sure @TheoWong-pixel is fine with View tool not being available for global volumes.

com.unity.render-pipelines.core/Editor/Volume/VolumeEditor.cs
TheoWong-pixel: No we just want the associated gizmo to to be unavailable. It should not be affecting anything else in the world.

alex-vazquez: We do not have a way of disabling the gizmos for transforms, this is the only way, I can hide the first time that they are selected as global, and then if the user clicks this will force to show them. Do you want this @TheoWong-pixel

arttu-peltonen: I feel like positioning the volume object is still a valid operation, even if the global volume logic then ignores the position. It can still be useful to position the global volumes in a certain place, e.g. for the purpose of organizing your level how you want it. Otherwise users would have to switch to Local mode, reposition it and then switch back to global. Maybe what Alex is suggesting is a good compromise - hiding the transform handles to indicate the position is irrelevant, but user is not prevented from selecting transform tool and moving the global volume to where they like.

jenniferd-unity:
@ellioman do you think you can take over from Martin while he is on PTO or is there someone better suited?

com.unity.render-pipelines.core/Editor/Volume/VolumeEditor.cs
ellioman: Agree with @arttu-peltonen here.
During the time I was working in Game Development we often placed things like global volumes, directional lights, etc in proper places to structure our scenes as we wanted. Not allowing that will introduce a pain for users in my opinion.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -99,25 +99,34 @@ public static void DrawFixMeBox(string text, Action action)
 
         // UI Helpers
         /// <summary>Draw a help box with the Fix button.</summary>
-        /// <param name="text">The message text.</param>
-        /// <param name="messageType">The type of the message.</param>
+        /// <param name="message">The message with icon if need.</param>
         /// <param name="action">When the user clicks the button, Unity performs this action.</param>
-        public static void DrawFixMeBox(string text, MessageType messageType, Action action)
+        public static void DrawFixMeBox(GUIContent message, Action action)
```
ellioman, line 104: Is this public to users?
If so, then this is a breaking change to public API and we need to deprecate it accordingly.

com.unity.render-pipelines.core/Editor/Volume/VolumeEditor.cs
```diff
@@ -9,6 +10,33 @@ namespace UnityEditor.Rendering
...
+            public static readonly GUIContent capsuleBoxCollider = EditorGUIUtility.TrTextContent("Add a Capsule Collider");
+            public static readonly GUIContent meshBoxCollider = EditorGUIUtility.TrTextContent("Add a Mesh Collider");
+
+            public static readonly GUIContent addColliderFixMessage = EditorGUIUtility.TrTextContentWithIcon("Add a Collider to this GameObject to set boundaries for the local Volume.", CoreEditorStyles.iconWarn);
+            public static readonly GUIContent disableColliderFixMessage = EditorGUIUtility.TrTextContentWithIcon("Global Volumes do not need a collider. Disable or remove the collider.", CoreEditorStyles.iconWarn);
+            public static readonly GUIContent enableColliderFixMessage = EditorGUIUtility.TrTextContentWithIcon("Local Volumes need a collider enabled. Enable the collider.", CoreEditorStyles.iconWarn);
+
+            public static readonly GUIContent profileInstance = EditorGUIUtility.TrTextContent("Profile (Instance)", "A copy of a profile asset.");
+            public static readonly GUIContent profile = EditorGUIUtility.TrTextContent("Profile", "A reference to a profile asset.");
```
ellioman, line 32: Today I created a PR (#5000) updating those tooltips. These need to be updated to use the same.

ellioman (changes requested):

com.unity.render-pipelines.core/Editor/CoreEditorStyles.cs
```diff
@@ -113,6 +119,23 @@ public static GUIStyle miniLabelButton
...
+        /// <summary>
+        /// Gets the icon that describes the <see cref="MessageType"/>
+        /// </summary>
+        /// <param name="messageType">The <see cref="MessageType"/> to obtain the icon from</param>
+        /// <returns>a <see cref="Texture2D"/> with the icon for the <see cref="MessageType"/></returns>
+        internal static Texture2D GetMessageTypeIcon(MessageType messageType)
+        {
+            return messageType switch
+            {
+                MessageType.None => m_TransparentTexture,
```
RSlysz, line 131: Shouldn't this be null instead? (transparent texture seams a unneeded waste of memory)

Looking into code, 
> EditorGUIUtility.TrTextContentWithIcon(text, icon)

 is just sugar to write
> EditorGUIUtility.TrTextContent(text, icon: icon)

which fully support null icon as the default value is in fact `null`.

RSlysz (commented):
I don't think we have the tools right now in unity to hide one gizmo contextually. Not sure you will be able to do that without clunky workflow. I would enquire with the team handling the scene management.

com.unity.render-pipelines.core/Editor/Volume/VolumeEditor.cs
TheoWong-pixel: Wait so the position of a global volume DOES matter? I was under the assumption that it would be the same wherever it is.

TheoWong-pixel (approved):

com.unity.render-pipelines.core/Editor/Volume/VolumeEditor.cs
arttu-peltonen: @TheoWong-pixel I don't think it matters for in terms of its functionality. BUT like @ellioman said, as a gamedev, you want to organize your levels (_spatially_) so you find your stuff and it can be annoying not to be able to move a gameobject, just because it's a global volume that doesn't strictly care about its position.

ellioman: Yes exactly. For the end result Global Volumes, Directional Lights, etc do not matter. But when working in scenes, especially big ones, you want to be able to organize these things just like users want with the hierarchy, project, etc.

alex-vazquez: I will remove this then :)

alex-vazquez:
@ellioman , I am waiting to your #5000 is submitted, to rebase master. I will let you know.

com.unity.render-pipelines.core/Editor/Volume/VolumeEditor.cs
phi-lira: Approving the PR, but I guess this needs to be removed before merging right?

phi-lira (approved):

alex-vazquez:
@phi-lira , yeah, I need to merge @ellioman PR #5000, before merging, and remove the stuff from the tool

alex-vazquez:
@ellioman everything merged :)

ellioman (approved):

PR #4923, opened 2021-06-15
[HDRP] Improve DoF consistency between full, half and quarter resolution modes

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1335687/

The amount of defocus blur was not consistent previously between the Full, Half and Quarter resolution modes in Depth of Field.
To and extend this was "by design", but it could confuse the users.

This PR improves the consistency between these modes. 

Full resolution:
image

Half resolution:
image

Half resolution (before this PR):
image

The clarity (readability of the numbers) at the upper part of the blue cube has also improved a lot. This bluriness was introduced by the rotated grid downsampling in the prefilter mode (that can hide aliasing some times at the expense of sharpness). Now this is only enabled only when the "High Quality Filtering" option is enabled. So disabling this option will iprove clarity at the expense of some (potential) aliasing.

Since this is a breaking change in terms of visual output, I have introduced a "Compatibility mode" option to revert all processing to the way it was before (for people who update old projects):
image

Note: For large blur radius with low sample counts we still overblur the image by using low-res texture mip maps. This hides the noise and other undersampling artifacts but in such cases the modes might not look consistent anymore. In this case the user will have to use more samples or reduce the max radius.

---
### Testing status
I have tested with the project that was attached in the bug ticket. 
- Different DoF resolutions look consistent.
- Different game view resolutions look consistent.
- Free aspect also  works

---
### Comments to reviewers
I will generate new reference images for DoF after the PR is approved.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/PostProcessing/DepthOfFieldEditor.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/PostProcessing/DepthOfFieldEditor.cs
@@ -50,7 +50,6 @@ namespace UnityEditor.Rendering.HighDefinition
         SerializedDataParameter m_HighQualityFiltering;
         SerializedDataParameter m_Resolution;
         SerializedDataParameter m_PhysicallyBased;
-        SerializedDataParameter m_CompatibilityMode;
 
         public override void OnEnable()
         {
@@ -74,8 +73,6 @@ namespace UnityEditor.Rendering.HighDefinition
             m_HighQualityFiltering = Unpack(o.Find("m_HighQualityFiltering"));
             m_Resolution = Unpack(o.Find("m_Resolution"));
             m_PhysicallyBased = Unpack(o.Find("m_PhysicallyBased"));
-
-            m_CompatibilityMode = Unpack(o.Find("m_CompatibilityMode"));
 
             base.OnEnable();
         }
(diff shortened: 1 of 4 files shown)


TomasKiniulis (approved):
Looks good! The top part of blue box is less blurry now aligning more with full res.

- Tested with multiple camera setup 
- Checked with Compatiblity Mode on/off with switching between Quarter, Half, Full resolutions
- Swtiching between Physical Camera and Manual modes

remi-chapelain:
I'm having doubt about compatibility mode option. 
I'm all for letting users decide when a change is introducing a benefit AND a drawback (that's what we did [here](https://github.com/Unity-Technologies/Graphics/pull/4665#pullrequestreview-672190500)), for some it could be very problematic to have a breaking rencering change. 

However, this one doesn't look like the very light drawback is enough to introduce another confusing checkbox that only very few people will understand what it does.
I'm also conflicted about how the compatibilty mode will evolve, since it mentions "old HDRP", but in a few month, where can we draw the line between current and old HDRP DoF algorithm in a few HDRP versions ? 

What I would suggest is to remove the compatibility mode and use the new version by default which already fixes issues. 
And if that causes real problems we could always re-introduce this in the future, but from my point of view it feels "overkill" in that situation :)

pmavridis:
Note: I will re-target this to master, since I think it's better to not backport it.

I agree with Remi, I will remove the compatibility option (will make the code cleaner too) and put a note in the upgrade guide about this change.

sebastienlagarde:
>Note: I will re-target this to master, since I think it's better to not backport it.
yes please, thanks

sebastienlagarde (approved):

PR #4710, opened 2021-05-28
[Yamato] Pipeline reorganization

kivistikliis:
### Purpose of this PR
Reorganizes Yamato test architecture according to https://jira.unity3d.com/browse/SDET-189 
Job hierarchy per area from highest to lowest goes: `Nightly `> `<area or project>` > (`PR` or `Extra`)> individual jobs. 
`Nighty` jobs exist only for ShaderGraph (`shadergraph.metafile`), URP (new `urp.metafile`), HDRP (`hdrp.metafile`), and VFX (new `vfx.metafile`) areas, and each of these nightlies encompasses the rest of their dependencies.

`Nightly` contains the nightlies of the 4 above-mentioned areas
`ABV (all PR jobs)` contains the PR jobs of the 4 above-mentioned areas

---
### Testing status
Checked that all dependencies are correct, and the 4 area entrypoints for CI are according to the proposed architecture
- [Nightly HDRP](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/yamato%252Fpipeline-reorg/.yamato%252Fall-hdrp.yml%2523Nightly_HDRP_trunk/7045762/job/pipeline)
- [Nightly ShaderGraph](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/yamato%252Fpipeline-reorg/.yamato%252Fall-shadergraph.yml%2523Nightly_ShaderGraph_trunk/7045788/job/pipeline)
- [Nightly URP](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/yamato%252Fpipeline-reorg/.yamato%252Fall-urp.yml%2523Nightly_URP_trunk/7046027/job/pipeline)
- [Nightly VFX](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/yamato%252Fpipeline-reorg/.yamato%252Fall-vfx.yml%2523Nightly_VFX_trunk/7046108/job/pipeline)
- [General checks](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/yamato%252Fpipeline-reorg/.yamato%252F_abv.yml%2523General_Checks_trunk/7078755/job) (formatting was added in a later commit than the above jobs)

---
### Comments to reviewers
- Triggers exist for:
  - **Nightly**
  -  **ABV (all PR jobs)** 
  - **General checks**
  - **PR HDRP**
  - **PR URP**
  - **PR VFX**
  - **PR SG**
  - PR PostProcessing 
[...]

PR #4650, opened 2021-05-26
[Yamato] Add retries to commands

kivistikliis:
---
### Purpose of this PR
Adds retries to commands to reduce instability failures.

---
### Testing status
[BoatAttack PR job](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/yamato%252Frich-commands/.yamato%252Fall-urp_performance_boatattack.yml%2523PR_URP_Performance_BoatAttack_trunk): retries can be seen in execution logs

PR #4925, opened 2021-06-15
[Yamato][7.x.x] Several SDET backports

kivistikliis:
---
### Purpose of this PR
Backports #4710 and #4650

---
### Testing status
[Weekly](https://unity-ci.cds.internal.unity3d.com/project/902/branch/7.x.x%2Fyamato%2Fsdet-backports/jobDefinition/.yamato%2F_abv.yml%23all_project_ci_weekly_2019.4)

---
### Comments to reviewers
- Cache is not backported due to low traffic on this branch
- `Extra` jobs are empty right now, as pipelines for PR/Extra separation were taken as was before (only PR jobs were present). We can decide if we want to keep this layer to keep consistency with the rest of the branches and possibly increase coverage, or remove these redundant layers.
- Compare to the last [weekly ](https://unity-ci.cds.internal.unity3d.com/job/7237576) (it is a few days older and therefore may have some differences)
- This PR enabled HDRP_DXR which was not in the weekly before. Do we want to include it? (It has a few tests failing)


sophiaaar (approved):

.yamato/_abv.yml
```diff
@@ -7,51 +7,52 @@
...
+    variables:
+        UTR_VERSION: "current"
+        TEST_FILTER: .*
+    dependencies:
+      -  path: .yamato/_projectcontext.yml#all_package_ci_project_2019.4
          rerun: on_new_revision
-      -  path: .yamato/universal-iphone-metal.yml#Universal_iPhone_Metal_Standalone_il2cpp_Linear_2019.4
+      -  path: .yamato/all-smoke_test.yml#Smoke_Test_2019.4
          rerun: on_new_revision
-    triggers:
```
theo-pnv, line 49: I think the weekly trigger was accidentally removed?

.yamato/_projectcontext.yml
```diff
@@ -420,8 +420,6 @@ all_package_ci_project_2019.4:
       -  npm install upm-ci-utils@stable -g --registry https://artifactory.prd.cds.internal.unity3d.com/artifactory/api/npm/upm-npm
       -  upm-ci package izon -t
       -  upm-ci package izon -d
-    triggers:
```
theo-pnv, line 423: Is this one supposed to go?

theo-pnv: https://unity.slack.com/archives/G0155CW8683/p1623758613025500

theo-pnv (changes requested):

.yamato/_projectcontext.yml
kivistikliis: Good catch, I don't think so, will add it back now

.yamato/_abv.yml
kivistikliis: Good catch this one too, will push a fix!

kivistikliis:
Last commit
- adds back weekly trigger (using pr_triggers workaround, since the script doesn't yet support per-editor triggers for custom job definitions, see [JIRA](https://jira.unity3d.com/browse/SDET-278) card) 
- changes the general checks trigger to always run on open PRs (to make sure package and smoke tests are run), because on `7.x.x/release` we don't have per-area triggers enabled
- per-area triggers are manually excluded from the commit due to the above-mentioned jira card, and therefore running the script will show changes not included in this PR, which may cause confusion. This will be fixed as soon as possible in a follow-up PR.

.yamato/_projectcontext.yml
kivistikliis: Replaced it with general checks triggers, since this runs both package and smoke tests.

theo-pnv (approved):

kivistikliis:
Triggers mentioned in previous comment are now fixed in the script (and included in this PR)

PR #4926, opened 2021-06-15
Fix Asymmetric Projection Matrices and Fog / Pathtracing

johnpars:
### Purpose of this PR

Fix: https://fogbugz.unity3d.com/f/cases/1330290

This PR addresses some issues that asymmetric projection matrices pose for volumetric fog, as well as the path tracer. 

Before:
Asymmetry

After:
AsymmetryAfter

HDRP already somewhat handles asymmetry (lens shift for physical cameras), however this does not handle the case of a user providing a custom projection matrix for a camera. This PR now determines if a camera's projection is asymmetric and defers to a generic matrix for the pixel coord to world direction transformation. 

This PR also fixes a slight issue with the ray generation for the path tracer; it was not accounting for the translation component / asymmetry within the pixel coord to world direction matrix.

---
### Testing status
- I manually confirmed that the usage of custom asymmetric projection matrices provides expected results for fog and the path tracer. 
- I ran the DXR tests locally to determine that my change to the ray generation did not produce any unexpected issues, however I got some failures. I am 80% sure this is because I am currently stuck with a laptop that uses the DXR fallback API. I will run Yamato to know for certain.

---
### Comments to reviewers
This PR is made in follow up to #4428, which was reverted on master due to an issue with the path tracer (this PR attempts to fix that part).


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -1737,13 +1737,20 @@ Matrix4x4 GetJitteredProjectionMatrix(Matrix4x4 origProj)
...
+            // asymmetry, and this is accounted for in the optimized path below. Additionally, Unity C++ will
+            // automatically disable physical camera when the projection is overridden by user.
+            useGenericMatrix |= HDUtils.IsProjectionMatrixAsymmetric(viewConstants.projMatrix) && !camera.usePhysicalProperties;
+
+            if (useGenericMatrix)
             {
                 var transform = Matrix4x4.Scale(new Vector3(-1.0f, -1.0f, -1.0f)) * viewConstants.invViewProjMatrix;
                 transform = transform * Matrix4x4.Scale(new Vector3(1.0f, -1.0f, 1.0f));
                 transform = transform * Matrix4x4.Translate(new Vector3(-1.0f, -1.0f, 0.0f));
                 transform = transform * Matrix4x4.Scale(new Vector3(2.0f * resolution.z, 2.0f * resolution.w, 1.0f));
```
adrien-de-tocqueville, line 1753: In my PR, i simplified this to reduce the amount of matrix multiplication
I think it gives the same result but it might be worth rechecking
```
            var viewSpaceRasterTransform = new Matrix4x4(
                new Vector4(2.0f * resolution.z, 0.0f, 0.0f, -1.0f),
                new Vector4(0.0f, -2.0f * resolution.w, 0.0f, 1.0f),
                new Vector4(0.0f, 0.0f, 1.0f, 0.0f),
                new Vector4(0.0f, 0.0f, 0.0f, 1.0f));

            var transformT = viewConstants.invViewProjMatrix.transpose * Matrix4x4.Scale(new Vector3(-1.0f, -1.0f, -1.0f));
            return viewSpaceRasterTransform * transformT;
```

johnpars: Just added it back, I can confirm (at least locally) that it produces the exact same result

eturquin (approved):

adrien-de-tocqueville (approved):

sebastienlagarde (approved):

PR #4929, opened 2021-06-16
[HDRP] Disable TAA jitter when frame debugger is active

aras-p:
### Purpose of this PR

In 2021.2 a19 FrameDebugger.enabled API has landed (https://github.cds.internal.unity3d.com/unity/unity/pull/4073), which we can use to disable TAA jitter, since it's very distracting while using the frame debugger. See https://docs.google.com/document/d/1rkSqVcF5gpFa8T4JoolNc9IfYBppHjZbYv5tm-GI9nE/edit

---
### Testing status
Manually tested that jitter while using frame debugger is gone.

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -1678,6 +1678,13 @@ Matrix4x4 GetJitteredProjectionMatrix(Matrix4x4 origProj)
                 taaJitter = Vector4.zero;
                 return origProj;
             }
+            #if UNITY_2021_2_OR_NEWER
```
ellioman, line 1681: Don't think we need that #if here - @sebastienlagarde please correct me if I'm wrong. 

Did the new API land in a19 or later? I ask because we're part of trunk right now and we have the minimum version (the package.json file) here set to 2021.2.0a19.

aras-p: Ah I wasn't sure how it goes these days; found 12 other cases of `UNITY_2021_2_OR_NEWER` in the Graphics code and assumed it's still needed :)

aras-p: The API landed in a19

sebastienlagarde: Hey, yes the define is not needed, safe to remove it but also doesn't harm if someone see the fix and think to backport it, so let's let it here.

sebastienlagarde (approved):

ellioman (approved):

sebastienlagarde:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fframedebugger-no-jitter/.yamato%252Fall-hdrp.yml%2523HDRP_trunk/7323982/job

PR #4931, opened 2021-06-16
[HDRP] Add velocity delta input to master node

sebastienlagarde:
### Purpose of this PR

This PR add support for "custom velocity" (a.k.a custom motoin vector) , in the sense that it allow to add an extra delta velocity to the existing motion vector.

image

---
### Testing status

Manual test have been done with the HairVertex node of the demo team and an additional vertex animation to validate that motion vector react correctly in this context (so custom position load from compute + custom vertex animation on top + custom delta velocity).

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/SurfaceOptionUIBlock.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/SurfaceOptionUIBlock.cs
@@ -79,7 +79,7 @@ namespace UnityEditor.Rendering.HighDefinition
 
             public static GUIContent doubleSidedEnableText = new GUIContent("Double-Sided", "When enabled, HDRP renders both faces of the polygons that make up meshes using this Material. Disables backface culling.");
 
-            public static GUIContent customVelocityText = new GUIContent("Add Custom Velocity", "When enabled, the provided custom object space velocity is added to the motion vector calculation.");            
+            public static GUIContent customVelocityText = new GUIContent("Add Custom Velocity", "When enabled, the provided custom object space velocity is added to the motion vector calculation.");
 
             public static GUIContent useShadowThresholdText = new GUIContent("Use Shadow Threshold", "Enable separate threshold for shadow pass");
             public static GUIContent alphaCutoffEnableText = new GUIContent("Alpha Clipping", "When enabled, HDRP processes Alpha Clipping for this Material.");
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/MotionVectorVertexShaderCommon.hlsl
@@ -145,7 +145,7 @@ PackedVaryingsType MotionVectorVS(VaryingsType varyingsType, AttributesMesh inpu
         effectivePositionOS -= inputPass.precomputedVelocity;
 #endif
 
-        
+
 
     // Need to apply any vertex animation to the previous worldspace position, if we want it to show up in the motion vector buffer
 #if defined(HAVE_MESH_MODIFICATION)


com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/SurfaceOptionUIBlock.cs
```diff
@@ -79,6 +79,8 @@ internal static class Styles
 
             public static GUIContent doubleSidedEnableText = new GUIContent("Double-Sided", "When enabled, HDRP renders both faces of the polygons that make up meshes using this Material. Disables backface culling.");
 
+            public static GUIContent customVelocityText = new GUIContent("Add Custom Velocity", "When enabled, HDRP will happen the provided per vertex custom velocity to the motion vector calculation.");
```
TomasKiniulis, line 82: I don't fully understand what the tooltip is trying to say. Maybe it should sound something like this: "When enabled, the provided custom velocity is added to the motion vector calculation."?

TomasKiniulis (changes requested):
Should we have automation coverage for this?

Tried to validate with alembic animation, the custom velocity seems to add up correctly.
https://user-images.githubusercontent.com/50582134/124761800-d768d100-df3a-11eb-8b48-e6cbba3ad4a1.mp4

Only minor nitpick is the tooltip. It sounds incorrect

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDBlockFields.cs
```diff
@@ -11,6 +11,9 @@ public struct VertexDescription
         {
             public static string name = "VertexDescription";
 
+            public static BlockFieldDescriptor CustomVelocity = new BlockFieldDescriptor(VertexDescription.name, "CustomVelocity", "Add Velocity (OS)", "VERTEXDESCRIPTION_CUSTOMVELOCITY",
```
johnpars, line 14: Should we move the (OS) note to the tooltip/documentation? Since it goes against the existing way it is already done, for Normal or Tangent input. Perhaps we would just name the block "Velocity" and be clear in the tooltip that this input is integrated **Additively** in **Object Space**.

johnpars (approved):
LGTM, left one note regarding tooltip and block naming.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDBlockFields.cs
sebastienlagarde: Sadly there is no tooltip for block. But I suppose you mean on the option to enable this input?

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/SurfaceOptionUIBlock.cs
sebastienlagarde: good suggestion

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDBlockFields.cs
sebastienlagarde: ok will do the change. Sadly as I was thinking there is only a tooltip on the option in graph settings, hope users will not be confuse. What I like with normal for example is that you can have the space you want and it is show but the concept will not exist for velocity

sebastienlagarde:
@Vic-Cooper I will need to merge this PR. But please still review the doc and I will do a next PR where I will correct the doc and add the mandatory test coverage as suggested by Tomas

For QA: the qa was done and I adressed the  tooltip feedback, but Tomas is in vacation. PR was approve.

Vic-Cooper:
@Seb Thanks for the tag! Reviewed the docs, changes are in this PR: https://github.com/Unity-Technologies/Graphics/pull/5310

sebastienlagarde:
@Vic-Cooper thanks a lot!

PR #4934, opened 2021-06-16
[HDRP][SRP] Restore old version of the RendererList structs/api for compatibility

pmavridis:
### Purpose of this PR
To avoid breaking existing projects, this PR restores the older version of the RendererList API (and marks it as obsolete).

---
### Testing status
I have tested using the HDRP template and RenderGraph passes are still culled properly. 

---
### Comments to reviewers
In files that include both the experimental rendering and the new RendererList namespace, I had to resolve the ambiguity in the name manually.


alelievr (approved):

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraph.cs
```diff
@@ -4,6 +4,9 @@
 using UnityEngine.Rendering;
 using UnityEngine.Rendering.RendererUtils;
 
+// Typedef for the in-engine RendererList API (to avoid conflicts with the experimental version)
+using CoreRendererListDesc = UnityEngine.Rendering.RendererUtils.RendererListDesc;
```
pmavridis, line 8: `using RendererListDesc = UnityEngine.Rendering.RendererUtils.RendererListDesc;`
did not work in this case, because we are using it inside the Experimental.Rendering namespace, which also includes the other RendererListDesc definition, so I had to "rename" the class in the alias/typedef.

In other places that we are outside of the experimental namespace (see HDRenderPipeline.cs) the rename in the alias is not necessary.

If you think there is a better or more elegant way to solve the ambiguity let me know!

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2021.1-to-2021.2.md
```diff
@@ -74,3 +74,7 @@ From 2021.2, the range for **Sorting Priority** values has decreased from betwee
 If you used transparent materials (**Surface Type** set to **Transparent**) with a sorting priority lower than -50 or greater than 50, you must remap them to within the new range.
 
  HDRP does not clamp the Sorting Priority to the new range until you edit the Sorting Priority property.
+
+## RendererList API
+
+From 2021.2, an updated RendererList API is available in the `UnityEngine.Rendering.RendererUtils` namespace. This API has lower overhead when submitting the RendererList for drawing and also offers the posibility to query if the list of visible objects is empty. The older version of the API (in the `UnityEngine.Experimental.Rendering` namespace) is still available for compatibility purposes but is now deprecated.
```
Vic-Cooper, line 80: Suggested change: 

From 2021.2, HDRP includes an updated `RendererList` API in the `UnityEngine.Rendering.RendererUtils` namespace. This API uses fewer resources than the previous version of the `RendererList` API when it submits the RendererList for drawing. You can use this new version to query if the list of visible objects is empty. 

The previous version of the API in the `UnityEngine.Experimental.Rendering` namespace is still available for compatibility purposes but it is now deprecated.

Vic-Cooper (commented):
Tweaked the Upgrade guide entry slightly.

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2021.1-to-2021.2.md
pmavridis: I revised the text as you suggested, but I kept the expression "lower overhead" instead of "fewer resources". The reason is that the savings are because we now avoid some copy/flush operations in the command buffer, which I think can be better described as "overhead".

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2021.1-to-2021.2.md
```diff
@@ -80,3 +80,5 @@ If you used transparent materials (**Surface Type** set to **Transparent**) with
 From 2021.2, HDRP includes an updated `RendererList` API in the `UnityEngine.Rendering.RendererUtils` namespace. This API has lower overhead than the previous version of the `RendererList` API when it submits the RendererList for drawing. You can use this new version to query if the list of visible objects is empty.
 
 The previous version of the API in the `UnityEngine.Experimental.Rendering` namespace is still available for compatibility purposes but is now deprecated.
+
+When the **Dynamic Render Pass Culling** option is enabled in the HDRP Global Settings, HDRP will use the new API to dynamically skip certain drawing passes based on the type of currently visible objects. For example if no objects with distortion are drawn, the Render Graph passes that draw the distortion effect (and their dependencies - like the color pyramid generation) will be skipped.
```
Vic-Cooper, line 84: Suggested change. I removed "dynamically" because, as a rule, we avoid adjectives in technical writing and I think we describe its behavior clearly: 

When the **Dynamic Render Pass Culling** property is enabled in the HDRP Global Settings, HDRP uses the new `RendererList` API to skip some drawing passes based on the type of GameObjects that are currently visible. For example, if HDRP doesn't draw any GameObjects with distortion, it skips the Render Graph passes that draw the distortion effect and their dependencies like color pyramid generation.

Vic-Cooper (changes requested):

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2021.1-to-2021.2.md
```diff
@@ -80,3 +80,5 @@ If you used transparent materials (**Surface Type** set to **Transparent**) with
 From 2021.2, HDRP includes an updated `RendererList` API in the `UnityEngine.Rendering.RendererUtils` namespace. This API has lower overhead than the previous version of the `RendererList` API when it submits the RendererList for drawing. You can use this new version to query if the list of visible objects is empty.
```
Vic-Cooper, line 80: I think we can clarify "lower overhead" here, based on your description: 

This API performs fewer operations than the previous version of the `RendererList` API when it submits the RendererList for drawing.

Vic-Cooper (changes requested):
Suggested a small tweak to the "lower overhead" sentence.

com.unity.render-pipelines.high-definition/Documentation~/Upgrading-from-2021.1-to-2021.2.md
pmavridis: Sounds good, thanks.

Vic-Cooper (approved):
Happy with the docs at this stage :)

JulienIgnace-Unity (approved):

pmavridis:
@Unity-Technologies/gfx-qa-hdrp : There is no new functionality in this PR, just restored some old classes that were deleted. If something went wrong, it would show as compile errors when opening a project, I did not notice anything on my side, not sure if you want to do any additional tests.

sebastienlagarde:
> @Unity-Technologies/gfx-qa-hdrp : There is no new functionality in this PR, just restored some old classes that were deleted. If something went wrong, it would show as compile errors when opening a project, I did not notice anything on my side, not sure if you want to do any additional tests.

For this PR there is no QA requirement but it was bring for your awareness, merging.

TomasKiniulis:
Had the branch merged while testing Antoine PRs. Haven't ran into anything either.

PR #4475, opened 2021-05-07
Preparation for New Hair Material Type (Marschner)

johnpars:
### Purpose of this PR

Provides the infrastructure to support the new Marschner-based shading model for HDRP's hair target.
It is a preparatory PR that will be followed up by a subsequent PR for the core BSDF evaluation. 
More information on the BSDF can be found in [this](https://docs.google.com/document/d/1tGHT-sxm3FRSAZ3n4l0b2atQ2vl1ivre8lNeqBIZwms/edit?usp=sharing) brief document, and progress can be tracked on [this branch](https://github.com/Unity-Technologies/Graphics/tree/HDRP/hair-material-type-2).

What this PR does include:
- (Skeleton) precomputation and binding of necessary LUTs (LTC, azimuthal scattering for TT lobe).
- Parameterization of the model for Shader Graph inputs.
- Enablement of all necessary code paths for BSDF preparation and evaluation (analytic, area, and IBL light); predicated by the compile-time flag.
- Switching to/from Kajiya-Kay model.

What this PR does **not** include: 
- The HLSL for the BSDF or BSDF Reference evaluation itself. 

### Testing Status
Note that from a user perspective, this PR has no impact. The ability to choose the new material type for Hair will remain hidden until the following PR lands. This means that for this PR, there is nothing to manually test aside from confirming that hair targets will assume the existing Kajiya-Kay model, & passes CI.

### Comments to Reviewers
Apologies for the lack of commit history- I needed to squash a bunch of commits from the [other](https://github.com/Unity-Technologies/Graphics/tree/HDRP/hair-material-type-2) branch.

PR #4935, opened 2021-06-16
Add BCSDF Approximation for Second Hair Material Type

johnpars:
This PR introduces a Marschner-based BSDF (technically, BCSDF, as this is a far-field model which assumes that the fibers are viewed from a distance at which the near-field variations can be ignored) model to the HDRP hair shader (as a second model option, not a replacement). It is a follow up to [this PR](https://github.com/Unity-Technologies/Graphics/pull/4475). 

Marschner3

Currently, HDRP’s hair shader already supports a dual-specular lobed Kajiya-Kay model, with an additional specular transmission lobe. It is a phenomenological model, rather than physically based. This creates a challenge for artists, as there are many parameters to tune the model, and it is their responsibility to balance the energy between specular lobes.

This Marschner-based model takes a physically based approach, by analyzing the scattering events that occur within a hair fiber. It offers a much simpler parameterization for artists, with the added benefit of conforming well to HDRP’s physically based lighting by default. 

### Usage
Users can access the new model through the following steps:
- Create a new Hair Shader Graph asset and open the graph. 
- Navigate to the Graph Inspector window in the graph. 
- Change the Material Type from Kajiya-Kay to Marschner.

Usage is otherwise very similar to the existing Kajiya-Kay model, and for the most part a user should be able to safely defer to the existing documentation regarding this (namely the usage of hair strand direction, proper use of which is crucial for getting correct results with the model)

### Parameterization
The table below specifies the properties of note for controlling the model. This model shares one mutual parameter with the Kajiya model, _Hair Strand Direction_. It behaves in the exact same way.

| Parameter  | Description |
| :-------------: | ------------- |
| _Base Color_  | Defines the fiber’s cortex color, governing the absorption that occurs before the transmission (TT), second specular reflection (TRT), and scattering.  |
| _Smoothness_   | Surface smoothness of the fiber, defining the width of the specular highlight.   |
| _Radial Smoothness_  | Controls the internal scattering of the light paths and absorption that occurs within the fiber. “Allow Radial Smoothness” must be enabled on the hair target to use this property.   |
| _Cuticle Angle_ | The angle in degrees at which the scales on the hair fiber are tilted from the strand direction. (for human hair, typically between 2 to 3 degrees). Use this property to “shift” the highlight.  |
| _Primary Reflection Smoothness_ | A multiplier of the primary reflection’s (R) smoothness. Use this parameter to approximate the layer of shiny coating that can be observed on rougher-looking animal fur.  |

Note, @SebastienLa have organized this [hair analysis document](https://docs.google.com/document/d/1zSxEvHnru2osUiatAvvN9irBWxei55KRMyMU5PDQRNw/edit#) that goes in depth on parameterization. This PR does not currently address everything proposed there, but we are discussing on how to improve it. 

### Performance
Overall, performance is comparable to HDRP’s other forward-based shading models (Kajiya, Fabric, etc.). Performance metrics were captured from a standalone Windows built with an NVIDIA GTX 1070 (Max-Q, Laptop Card), at a 1920x1080 resolution, no MSAA. 

**Scenario #1**: Two screen-aligned quads (Marschner, Kajiya), each taking up 50% of the viewport. Illuminated by one point, one spot, one directional, one environment light.

| Model  | Forward Opaque Time Per-Quad (ms) |
| Kajiya-Kay  | 1.54  |
| Marschner  | 1.36 |
| Marschner (Reference)  | 22.32  |

image

**Scenario #2**: Two hair groom files (7000 strands represented with screen-aligned strips). Illuminated by three spot lights.

**NOTE: These metrics were also taken 2X MSAA.** 

| Model  | Forward Opaque Time Per-Groom (ms) |
| Kajiya-Kay  | 4.87  |
| Marschner  | 4.32 |
| Marschner (Reference)  | 57.38  |

image

### Light Loop Integration
| Light Type |   |
|:---:|---|
| _Analytic_  |  All analytic lights in HDRP (point, spot, directional) are supported. |
| _Area_  | For now, like Kajiya, we must fall back to GGX for area lights. This is due to an issue with LTC which cannot capture anisotropic lobes well in the table. More research is required to support this. In the nearer term, it probably makes sense to explore a “Most Representative Light” approach to add support for anisotropic BSDFs to area lights in HDRP. |
| _Environment_  | Environment lighting is calculated by re-purposing the probe lighting sample. The sample is treated as a directional light, and the BSDF is invoked with respect to it (with the transmission lobe disabled).   |

### Reference
This PR also includes a reference BSDF, which the approximation is based on. It can also be used by the path tracer, but right now is uniformly sampled.

[TODO IMAGE]

### Additional Notes
- Due to the transmission lobe, this model is extremely dependent on shadows.
- During the process of adapting the model from reference to real-time, I realized that I would need to remove the Refraction Index (IOR) artist parameter. This is because most of the approximations need to be made with respect to human hair IOR (1.55) for optimal performance. 
- Additionally, the approximated model only evaluates the first three scattering events (R, TT, TRT) for performance reasons. The residual TR*T lobe is not accounted for (resulting in a ~15% energy loss). 
- For now, the model re-uses the Kajiya-Kay diffuse term as a matter of scattering approximation. 
- For roughened azimuthal scattering (enabled with the Allow Radial Smoothness toggle), a LUT is sampled per-BSDF invocation. I created this LUT in python and stored it as an EXR. I will add a Yamato test to confirm that this is okay for the various plaforms.

---
### Testing status

Updated Graphics Test 1401 to also test for the new model. Launching a Yamato to confirm everything checks out on the various platforms. 

1401_HairGraph

The original reference implementation was verified by plotting the various distribution curves and referencing them to the original papers I derived this model from. During the approximation process, I used the reference to make informed optimization decisions at the cost of quality / accuracy.


com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPropertyBlocks.cs
```diff
@@ -28,7 +28,7 @@ public HairSurfaceOptionPropertyBlock(SurfaceOptionPropertyBlock.Features featur
         protected override void CreatePropertyGUI()
         {
             // TODO: Un-hide me when Marschner BSDF is available.
```
eturquin, line 30: Is this comment still relevant?

eturquin (approved):

com.unity.render-pipelines.high-definition/Documentation~/snippets/shader-graph-blocks/cuticle-angle.md
```diff
@@ -0,0 +1,6 @@
+<tr>
+<td>**Smoothness**</td>
+<td>The angle in degrees at which the scales on a hair fiber are tilted from the strand direction. (for human hair, typically between 2 to 3 degrees). Use this property to “shift” the highlight..</td>
```
Vic-Cooper, line 3: <td>The angle (in degrees) that the scales on a hair fiber tilt from the strand direction. For human hair, this value is usually between 2 to 3 degrees. Use this property to “shift” the highlight.</td>

com.unity.render-pipelines.high-definition/Documentation~/snippets/shader-graph-blocks/smoothness-radial.md
```diff
@@ -0,0 +1,6 @@
+<tr>
+<td>**Smoothness**</td>
+<td>Controls the internal scattering of the light paths and absorption that occurs within the hair fiber. “Allow Radial Smoothness” must be enabled on the hair target to use this property.</td>
```
Vic-Cooper, line 3: <td>Controls the internal scattering of light paths and the amount of light the hair fiber absorbs. To use this property, enable **Allow Radial Smoothness**.</td>

com.unity.render-pipelines.high-definition/Documentation~/snippets/shader-properties/advanced-options/allow-radial-smoothness.md
```diff
@@ -0,0 +1,4 @@
+<tr>
+<td>**Allow Radial Smoothness**</td>
+<td>Adds a secondary **Radial Smoothness** block to the target, controlling the internal scattering of the light paths and absorption that occurs within the fiber. Enabling this option comes at a slight performance cost to the shader; when disabled, it will fall back to an approximation of the internal scattering.</td>
```
Vic-Cooper, line 3: <td>Adds a secondary **Radial Smoothness** block to the target that controls the internal scattering of light paths and the amount of light the hair fiber absorbs. Enabling this option makes the shader use slightly more resources on the GPU. When this property is disabled, the shader approximates internal scattering.</td>

com.unity.render-pipelines.high-definition/Documentation~/snippets/shader-properties/advanced-options/geometry-type.md
```diff
@@ -0,0 +1,4 @@
+<tr>
+<td>**Geometry Type**</td>
+<td>Indicates the type of geometry being used to represent the hair, allowing the shading model to make informed approximations. Note that switching between these two types will produce slightly differing results, due to how the underling shading model approximates the hair fiber. The options are:<br/>&#8226; **Cards**: Representation by hair "cards"; a projection of hair fibers onto a drastically simplified proxy geometry.<br/>&#8226; **Strands**: Representation by screen-aligned ribbons, or tubes; an explicit representation of a hair fiber.</td>
```
Vic-Cooper, line 3: <td>Indicates what geometry is used to represent the hair, which allows the shading model to make informed approximations. The shading model approximates the hair fiber differently in each of the following representations: <br/>&#8226; **Cards**:  Hair fibers are projected onto simplified proxy geometry called "cards".<br/>&#8226; **Strands**: Hair fibers are represented by screen-aligned ribbons or tubes.</td>

com.unity.render-pipelines.high-definition/Documentation~/master-stack-hair.md
```diff
@@ -154,5 +156,6 @@ Depending on the [Graph Settings](#graph-settings) you use, Shader Graph can add
 [!include[](snippets/shader-properties/advanced-options/override-baked-gi.md)]
 [!include[](snippets/shader-properties/advanced-options/support-lod-crossfade.md)]
 [!include[](snippets/shader-properties/advanced-options/add-precomputed-velocity.md)]
-[!include[](snippets/shader-properties/advanced-options/use-light-facing-normal.md)]
+[!include[](snippets/shader-properties/advanced-options/geometry-type.md)]
+[!include[](snippets/shader-properties/advanced-options/allow-radial-smoothness.md)]
```
Vic-Cooper, line 160: These [!includes[] look good, but I'd encourage you to test them using the Docs Tools package to make sure they work.

Vic-Cooper (changes requested):
As noted, it's worth texting to make sure the `include`s work properly. Added some changes to the docs too.

slunity (approved):

sebastienlagarde (approved):

sebastienlagarde:
@johnpars All good. Can you update the documentation as mention by Vic and we are ready to go with this PR (before further improvement)

sebastienlagarde:
Also be sure to merge master and update screenshots

PR #4945, opened 2021-06-17
Updated URP decal graph to have color and sample normal correctly

Jonasmortensen:
---
### Purpose of this PR
Included decal shader graph didn't use the correct `Type` for the normal map `Sample Texture2D` node. The PR also adds a color property to the shader which makes it more versatile.


Jonasmortensen:
# Old Graph
image
image

# New Graph
image
image

simon-engelbrecht-soerensen (approved):
Looks good to me.

lukaschod (approved):
Nice

PR #4950, opened 2021-06-17
Color field keeps updating when using eye dropper and Esc key is pressed

julienamsellem:
### Purpose of this PR

Stop updating the color preview in the color field when the Eye dropper is used and the Escape key is pressed.
Additionally when Escape key is pressed the initial color is recovered.

https://fogbugz.unity3d.com/f/cases/1291991/
---
### Testing status
Tested locally

---
### Comments to reviewers
This fix depends on the following ono PR:
https://ono.unity3d.com/unity/unity/pull-request/128105/_/graphics/vfx/fix/1291991-colorpicker-preview-keeps-updating


VitaSkruibyte:
Looking neat! Tested:
- If color pickers work fine with inline nodes, exposed properties, properties in blackboard, blocks, Subgraphs, VFX Subgraph Operators (blackboard, inline nodes, blocks)
- Color picker of the color window, color picker of the property (that appears in the node/block)

VitaSkruibyte (approved):

VitaSkruibyte:
And just a kind reminder, please don't forget to add/update the changelog. 😇

iTris666 (approved):

julienamsellem:
Yamato is failing because of a dependency on an ono PR which needs to be merge for this PR to compile.

PR #4954, opened 2021-06-17
APV Virtual offset proof of concept

FrancescoC-unity:
This PR adds a way to push the position of probes very close to a backface of geometry (i.e. inside) to outside the geometry.  
This fixes a bunch of leaking issues that come due to bad dilation; dilation is still very much needed, especially with this proof of concept version. 

**NOTE** This is not very precise and it is purely a proof of concept while waiting for the proper system that will need to come from the lightmapper. 

The physics test look also not super precise, so if a point is inside a thin mesh it fails; I am chatting with someone from the physics team to check more about it. 

**Before**
NoVO
2_NoVO

**After**
VO
2_VO


kristian15959:
I currently work on my own Probe Volume system and I fixed most of the leaking by discarding probes inside geometry and if there are probes around occluded probe than occluded probe samples value from them, but moving probes is nice solution I would like to look into aswell, combining these two would be the best option I would guess.

JulienIgnace-Unity (approved):

alelievr (approved):

sebastienlagarde (approved):

PR #4974, opened 2021-06-25
Added a better support for LODs in the ray tracing acceleration structure.

anisunity:
I found a way to make LODs without introducing new artifacts (and removing some previous ones).
image

Testing status
Added a new test to cover the new behavior
HDRP DXR tests are green locally

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
@@ -339,6 +339,12 @@ namespace UnityEngine.Rendering.HighDefinition
             return worldSpaceSize / (2.0f * relativeScreenHeight * halfAngle);
         }
 
+        internal float GetWorldSpaceSize(LODGroup lodGroup)
+        {
+            Vector3 lossyScale = lodGroup.transform.lossyScale;
+            return Mathf.Max(Mathf.Max(Mathf.Abs(lossyScale.x), Mathf.Abs(lossyScale.y)), Mathf.Abs(lossyScale.z)) * lodGroup.size;
+        }
+
         internal void BuildRayTracingAccelerationStructure(HDCamera hdCamera)
         {
             // Clear all the per frame-data
@@ -479,7 +485,7 @@ namespace UnityEngine.Rendering.HighDefinition
                 LODGroup lodGroup = lodGroupArray[i];
 
                 // Compute the distance between the LOD group and the camera
-                float lodGroupDistance = (hdCamera.camera.transform.position - lodGroup.transform.position).magnitude;
+                float lodGroupDistance = (hdCamera.camera.transform.position - lodGroup.transform.TransformPoint(lodGroup.localReferencePoint)).magnitude;
 
                 // Flag to track if the LODGroup has been represented
                 bool lodGroupProcessed = false;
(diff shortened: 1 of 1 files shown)


remi-chapelain:
We might want to update documentation [here](unsupported-features-of-ray-tracing) that says that we only add the top level LOD in the RTAS since it won't be true after this PR

remi-chapelain (changes requested):
The basic idea works (LOD gets changed depending on the distance) but there's a mismatch between the distance we got in raster and the distance we calculate in the ray tracing manager. 

And after looking quickly at the code, I believe it's due to the fact that LODgroup.size gives the **local** size but doesn't take into account the scale of the gameobject. 


https://user-images.githubusercontent.com/57442369/123670364-6f7d0100-d83d-11eb-9617-d5607a032e77.mp4

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
```diff
@@ -331,6 +331,14 @@ static bool IsAlphaTestedMaterial(Material currentMaterial)
             return (!materialIsOnlyTransparent && hasTransparentSubMaterial) ? AccelerationStructureStatus.TransparencyIssue : AccelerationStructureStatus.Added;
         }
 
+        // This function mimics the LOD calculation that is used for the rasterization
+        // THe idea is to have rasterization and ray tracing perfectly match.
+        internal float CalculateCameraDistance(float fieldOfView, float worldSpaceSize, float relativeScreenHeight)
+        {
+            float halfAngle = Mathf.Tan(Mathf.Deg2Rad * fieldOfView * 0.5F);
+            return worldSpaceSize / (2.0f * relativeScreenHeight * halfAngle);
```
remi-chapelain, line 339: I think here we need to multiply worldSpaceSize by the localScale of the LODgroup to have a match with the LOD transition.

remi-chapelain (changes requested):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.cs
anisunity: Gonna fix that and update the test scene

anisunity:
Fixed Pushed

remi-chapelain (changes requested):
So, with the new fix, LOD transition match perfectly with simple geometry but with more complex (and real) geometry, there's a very small lag between RT and Raster transition. Meaning the transition in raytracing occurs a tiny bit before the raster transition resulting in discrepancies. 

This looks like a rounding error or something compared to what rasterization is doing but I don't know much more. 

Here's a repro scene to help debug. 

[repro-lod-rtx-transition.zip](https://github.com/Unity-Technologies/Graphics/files/6742845/repro-lod-rtx-transition.zip)


https://user-images.githubusercontent.com/57442369/124004890-c0752c80-d9d8-11eb-8a79-ee41794bc7f8.mp4

https://user-images.githubusercontent.com/57442369/124004900-c2d78680-d9d8-11eb-919f-f869e44be7f1.mp4

anisunity:
Last issue found fixed and pair tested with @remi-chapelain

remi-chapelain (approved):
All good, one more feature supported in raytracing. 🥳 
Let's merge this 🟢

sebastienlagarde (approved):

PR #4976, opened 2021-06-28
Fix Vertex Color Mode documentation

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1345568

When using the add mode, the vertex color is remapped to [-1, 1], fixed the doc that was giving a wrong range.
Modifying the code to give the same result as the fogbug wants would risk breaking existing projects.

Also remove some old comment in the code about wind which doesn't exist anymore I believe

---
### Testing status
I didn't try to edit the mesh vertex color to verify that putting 0.5 for neutral value works, but i expect it does  :)


Vic-Cooper (approved):
Only tiny changes, all good!

PR #4977, opened 2021-06-28
Fix diffusion profile info box in VFX graph (and potentially other tools)

alelievr:
### Purpose of this PR
Address a small UI bug where the diffusion profile infobox was not displayed in VFX Graph:

Before:
image

After:
image

Issue Reference: https://unity.slack.com/archives/C6Y79CZM0/p1623934772104100

Bonus: for scripts that have access to the HDRP internals: they now have the same infobox by default:
image

---
### Testing status

Tested for regression everywhere there is a diffusion profile visible:

HDRP material + Shadergraphs:
image
image

Volume profile override:
image

HDRP Global Settings:

image


RSlysz (commented):
Seams good. 

@TheoWong-pixel it seams that we have 3 different design for this fix button in helpbox:
- The one that we used in Material (if still the same)
- The one that we use here and in DiffusionProfileMaterial
- The one that is done here https://github.com/Unity-Technologies/Graphics/pull/4922#pullrequestreview-695784974
Tell us which one we should use :/

(This can be be fixed in a later quality pass if needed)

RSlysz (approved):

peeweek (approved):

iM0ve (approved):
No issues found

**Whats tested:**
* Upgraded a simple scene with assets + particles using Diffusion profiles
* Checked that the new info boxes appear and the existing ones are sitll present
* Tried the FIX button to see it actually works
* Upgraded HDRP template to check on the bamboo trees to see if any unexpected behaviour is present

Before PR (no fix on vfx)

Unity_FpwYbRXkHW


After PR (fix for materials and vfx is available)

Unity_Xb8lIg85Oa


After FIX is pressed

TwKtd2U2N6

sebastienlagarde (approved):

PR #4980, opened 2021-06-28
Standardization of SSR, SSGI and AO regarding naming (case 1335387)

fredericv-unity3d:
### Purpose of this PR
Standardize naming of several volume components (case case 1335387):
- SSGI
- SS Reflection
- SS Refraction
- AO

---
### Testing status
Manual Test:
1. Checkout master and create an empty project
2. Added SSGI, SS Reflection, SS Refraction, AO volumes to a Volume profile
3. Added some custom values to them
4. Checkout this branch
5. > Observe that 
   a. We have all previous volumes with their associated data
   b. The naming of the volume components are correct.

Note: The SSGI volume component was renamed only after Unity was restarted.

Running Yamato tests as well.
* So far a few failure caused by `kTextureUploadUninitialized` that are also appearing on master.
* However 4082_DepthOfField-PB: Rendered failed, this one need investigation

---
### Comments to reviewers

Doc: I also moved the documentation pages as I changed the links as well.


JulienIgnace-Unity (commented):
I see a lot of class renaming. Does this break old projects?

remi-chapelain (changes requested):
Quickly checked upgrade of those 4 overrides (Reflection, Refraction, Global Illumination, Ambient Occlusion) and surprisingly everything looks fine. I test on the template and Small Office RayTracing

However, for the documentation, we still have a "Override-Screen-Space-GI.md". 
Also, the naming have been changed for "Override-Reflection.md" and "Override-Refraction.md" but there's still a lot of backlinks to the old pages ("Override-Screen-Space-Reflection.md", "Override-Screen-Space-Refraction.md") that will fail now, so we need to update those as well. 

Also, there's a bunch of tooltips that still include "Screen Space". We do not need to get rid of every screen space since in context it can make total sense but for example the enable has screen space in it and doesn't make sense in every situation. 
image

fredericv-unity3d:
(Rebased on green master)

fredericv-unity3d:
@JulienIgnace-Unity Not it does not, I took care of keeping appropriate GUID in meta files and `VolumeComponent` are serialized as `Unity.Object` reference. So the migration does not impact technically previous projects.

fredericv-unity3d:
@remi-chapelain Regarding doc, I am talking with @Vic-Cooper about it.
I think I'll do a first version by just updating all references.

Although I forgot to check about tooltips, will do the update.

fredericv-unity3d:
@remi-chapelain Regarding the Debug Menu and the HDRP Asset, a straightforward renaming is probably not the appropriate answer. I need to double check what algorithm(s) is(are) actually behind those choices, and options can be to add/split/rename thoses.

fredericv-unity3d:
@remi-chapelain Regarding the Refraction: I check the UX and they like the change as well.

The motivation is that the volumes are named regarding the rendering goal to achieve, the technical details are inside. (So Refraction as goal, SSRefraction as algorithm/rendering technique)

fredericv-unity3d:
Note: We are still designing how we should rename the API, so this PR is to be changed.

remi-chapelain:
Will put our testing on hold. Please re-request review when you think the PR will be ready again :)

fredericv-unity3d:
The real issue is that the workflows are not clear, so I am closing this PR, another will come to refactor a bit the volumes parameters and clarify the workflows with additional documentation

RemyUnity:
Reopened case [1335387](https://fogbugz.unity3d.com/f/cases/1335387/) as this PR never landed.

PR #4147, opened 2021-04-11
Remove Opaque and Depth Textures from Overlay Camera properties

berniegp:
### Purpose of this PR
The documentation says that the Opaque Texture and Depth Texture properties are used for Overlay cameras. This is not true AFAICT and the opposite is said in `Graphics/com.unity.render-pipelines.universal/Documentation~/camera-component-reference.md`.

---
### Testing status
n/a

---
### Comments to reviewers
I do not know if this should be backported and I don't think this needs a changelog entry. Please note that I'm external to the Graphics team.

PR #4988, opened 2021-06-29
Remove Opaque and Depth Textures from Overlay Camera properties (backport #4147).

oleks-k:
Remove Opaque and Depth Textures from Overlay Camera properties (backport #4147).


eh-unity (approved):
lgtm

PR #4991, opened 2021-06-29
Global Settings can now be unassigned in the Graphics tab (case 1343570)

jenniferd-unity:
---
### Purpose of this PR
Adding the ability to un-assign a Global Settings via the UI so users can remove it from the build.
Also adding a "Fix" button so a Global Settings is automatically assigned (the first one found, or a new one if none available).
1343570-unassign-hd-globalsettings

Fogbugz case [1343570](https://fogbugz.unity3d.com/f/cases/1343570/)

---
### Testing status
Tested on Windows Editor the various behaviors:
- if the SRP is the active pipeline => can we (un-)assign the asset? is the fix button assigning a Global Settings for us?
- checked both with URP/HDRP/Built-in as the active pipeline

Note: if the SRP is active and the Game/Scene view is opened, unassigning will have no effect because we need one Global Settings to properly render the view. There used to be a warning but it is temporarily disabled until the template is updated.


arttu-peltonen:
> Note: if the SRP is active and the Game/Scene view is opened, unassigning will have no effect because we need one Global Settings to properly render the view. There used to be a warning but it is temporarily disabled until the template is updated.

How does this work exactly? From what I can reason, this PR causes `(...)PipelineGlobalSettings.instance` to become null when "None" is selected.

jenniferd-unity:
> > Note: if the SRP is active and the Game/Scene view is opened, unassigning will have no effect because we need one Global Settings to properly render the view. There used to be a warning but it is temporarily disabled until the template is updated.
> 
> How does this work exactly? From what I can reason, this PR causes `(...)PipelineGlobalSettings.instance` to become null when "None" is selected.

yes it is doing the same thing as if the active RPGlobalSettings was deleted. we use to have a warning saying "i m gonna create one for you" but it got removed, so a RPGlobalSettings is assigned automatically for you without any warning :)

arttu-peltonen:
> > > Note: if the SRP is active and the Game/Scene view is opened, unassigning will have no effect because we need one Global Settings to properly render the view. There used to be a warning but it is temporarily disabled until the template is updated.
> > 
> > 
> > How does this work exactly? From what I can reason, this PR causes `(...)PipelineGlobalSettings.instance` to become null when "None" is selected.
> 
> yes it is doing the same thing as if the active RPGlobalSettings was deleted. we use to have a warning saying "i m gonna create one for you" but it got removed, so a RPGlobalSettings is assigned automatically for you without any warning :)

I'm confused :D the gif above shows user _unassigning_ the RPGlobalSettings, and it stays as "None", yet you're saying one gets assigned automatically for you? Is the gif above depicting a situation where HDRP is not active, i.e. the unassigning of RPGlobalSettings asset is only possible for a SRP that is not active?

jenniferd-unity:
> > > > Note: if the SRP is active and the Game/Scene view is opened, unassigning will have no effect because we need one Global Settings to properly render the view. There used to be a warning but it is temporarily disabled until the template is updated.
> > > 
> > > 
> > > How does this work exactly? From what I can reason, this PR causes `(...)PipelineGlobalSettings.instance` to become null when "None" is selected.
> > 
> > 
> > yes it is doing the same thing as if the active RPGlobalSettings was deleted. we use to have a warning saying "i m gonna create one for you" but it got removed, so a RPGlobalSettings is assigned automatically for you without any warning :)
> 
> I'm confused :D the gif above shows user _unassigning_ the RPGlobalSettings, and it stays as "None", yet you're saying one gets assigned automatically for you? Is the gif above depicting a situation where HDRP is not active, i.e. the unassigning of RPGlobalSettings asset is only possible for a SRP that is not active?

Sorry my explanation is a bit unclear you are right ^^
So in the UI: now one can unnassign a RP Global Settings, the same way we could do via script.
However, if you are rendering a scene with a RP (game view open for ex.), the renderer will automatically create/pick one RP Global Settings for you to avoid any mistake.
But, lets imagine you have both pipeline and you want to create a build with only URP scenes, you want to make sure you dont include the HDRP Global Settings, well now you can :)

RSlysz (approved):
> But, lets imagine you have both pipeline and you want to create a build with only URP scenes, you want to make sure you dont include the HDRP Global Settings, well now you can :)

Ha. That is the reason. It wasn't clear for me as well.
Shouldn't it be part of the trimming process that we should have when building instead of manually have to remove a global settings?

Seams good otherwise.

arttu-peltonen:
> Sorry my explanation is a bit unclear you are right ^^
> So in the UI: now one can unnassign a RP Global Settings, the same way we could do via script.
> However, if you are rendering a scene with a RP (game view open for ex.), the renderer will automatically create/pick one RP Global Settings for you to avoid any mistake.
> But, lets imagine you have both pipeline and you want to create a build with only URP scenes, you want to make sure you dont include the HDRP Global Settings, well now you can :)

Ok, thanks for the clarification - makes sense :) So as long as long as we don't want ability to remove global settings in a single-pipeline workflow (which most likely doesn't make sense since, as you say, it would make the SRP unusable), this seems correct.

arttu-peltonen (approved):

JMargevics (approved):
Please disregard.

JMargevics:
https://user-images.githubusercontent.com/50581102/123973685-a6315380-d9c4-11eb-8541-782354e4cfb7.mp4

image

If I stop the video I can see the fix button etc

JMargevics (changes requested):
Oops wrongly approved. The issue is above

jenniferd-unity:
@JMargevics in your capture it looks like you are trying to unassign a Global Settings for an active SRP. this will not work indeed if there is something requiring for a frame to be rendered. however you can unassign it for a pipeline not in use or if no frame is rendered (no game view, no scene view refresh for ex).

JMargevics (commented):
Thank you @jenniferd-unity, I misunderstood a bit the aim of PR. Checking unassigning settings of inactive SRP and it works fine. Also tried Fix buttons, and they work correctly ✅

I have two minor issues left:

1)
URP
vs
hdrp

The layout of HDRP and URP warning messages and fix buttons are quite different. 

2) Can't undo setting asset assignation (and removal)

jenniferd-unity:
> Thank you @jenniferd-unity, I misunderstood a bit the aim of PR. Checking unassigning settings of inactive SRP and it works fine. Also tried Fix buttons, and they work correctly ✅
> 
> I have two minor issues left:
> 
> 1. The layout of HDRP and URP warning messages is quite different.
>    URP
>    vs
>    hdrp
> 2. Can't undo setting asset assignation (and removal)

Perfect! I definitely forgot to submit a file for the first one. And I will look into undo right now then!

jenniferd-unity:
@JMargevics I pushed the fix for the UX misalignment
But regarding Undo the solution is actually more complex and requires C++ change - would it be okay to add this as a Fogbugz instead? (we already have this one: https://fogbugz.unity3d.com/f/cases/1342987/)

JMargevics (approved):
Everything looks good now ✅ 

I will update the fogbugz case then 👍

jenniferd-unity:
@JMargevics to confirm my understanding, did you test both URP and HDRP? if so, maybe we could skip URP QA

JMargevics:
Yeah, I tried unassigning and clicking fix in both SRPs. All good in both 👍

ellioman (approved):

PR #4992, opened 2021-06-29
old: Hd/bugfix-1339945: Fix fabric IBL performance & quality

JarkkoUnity:
### Checklist for PR maker

---
### Purpose of this PR
Improve quality and performance of fabric IBL pre-convolution. The rendered result is much closer to the reference implementation and uses 1000x less samples and thus performs significantly better.

---
### Testing status
Tested the results against reference implementation for 1352 HDRP test and with various HDR probes showing closer match with the reference. The quality comparison in different scenes (reference implementation ("reference" layers), current implementation ("current" layers), new implementation using 100x less samples ("new" layers), new implementation using 1000x less samples ("new/10" layers): https://drive.google.com/open?id=1-ZDUlRbN1IiwoyCB-7RaLnfDZiQ0N94L&authuser=jarkko.lempiainen%40unity3d.com&usp=drive_fs

The offending project that renders the scene 7 times (bad) also went from ~1fps back to ~10fps. RenderDoc timing for the project shows ~50x perf improvement for single scene rendering (~166ms => ~3ms).

---
### Comments to reviewers
Need to update reference images for test 1352 to pass with the improved fabric IBL quality.

You can enable reference rendering by uncommenting the following line in Fabric.hlsl:
// #define FABRIC_DISPLAY_REFERENCE_IBL

At the end of CharlieConvolve.shader there's the old convolution (IntegrateLDCharlie call) commented out for testing.

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
@@ -538,97 +538,11 @@ real4 IntegrateLD(TEXTURECUBE_PARAM(tex, sampl),
 }
 
 real4 IntegrateLDCharlie(TEXTURECUBE_PARAM(tex, sampl),
-                   real3 V,
-                   real3 N,
-                   real roughness,
-                   uint sampleCount,
-                   real invOmegaP,
-                   bool prefilter)
-{
-    // Local frame for the local to world sample transformation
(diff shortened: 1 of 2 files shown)


JMargevics:
https://fogbugz.unity3d.com/f/cases/1339945/ OG case 👆

JMargevics:
Just checked our materials performance test scene: we don't have any reflection probes there. 

Thinking it's worth updating our scene to cover the performance hits like this

PR #5002, opened 2021-06-29
[HDRP][Samples] Remove EventSystem from ParticleShaderGraphsSampleScene.unity

TomasKiniulis:
Issue found while smoke testing 10.6: https://docs.google.com/document/d/1cveO-glh6TLTRPOWcSefeskndxyiOg3nYxjgZ1zO4Ho/edit?usp=sharing

We should remove Canvas EventSystem from the samples since it causes the following error spam in Playmode:
`InvalidOperationException: You are trying to read Input using the UnityEngine.Input class, but you have switched active Input handling to Input System package in Player Settings.`
image (43)

Should be safe to remove as it serves no purpose to the scene anyway(no clickable UI).

Testing status:
Made sure sample scene works correctly when imported to 2020.3.14f1 and 2021.2.0b1


JMargevics (approved):
Simple deletion seems safe enough ✅

alelievr (approved):

PR #5006, opened 2021-06-29
Global wind parameter

adrien-de-tocqueville:
### Purpose of this PR

Make orientation and speed of wind parameters for various overrides available from the visual environment. These can then be overriden locally for each volume override using them. This is an UX improvement requested in this slack discussion: https://unity.slack.com/archives/C6Y79CZM0/p1624351744140500

For HDRI Sky and Volumetric clouds:
image

There are 4 override modes (as requested by artist):
- Custom: use a custom value
- Global: use the value from the visual environment
- Multiply: multiply from the env
- Additive: add from the env

Also unified orientation so all overrides go along the x axis by default.
Since the hdri sky distortion cannot be expressed in km/h, it is remapped by a constant factor to roughly match volumetric clouds speed for default settings. Also the distortion orientation is now independant of the skybox rotation

This is how the visual env looks like:
image
(Note that i also changed the info message for static lighting which was "Current Static Lighting Sky use None of profile None.")

I updated the doc pages accordingly

---
### Testing status

Tested interpolation between different volumes
Tested migration of the HDRI sky:
- tested the speed remapping
- tested the wind orientatiod remapping (now along x vector)
- tested skybox rotation remapping (now distortion is independant of rotation)
Tested migration of the Volumetric Clouds

---
### Comment to reviewers

Important: This migration isn't entirely correct in because it doesn't take into account hierarchy of volumes (that depends on where the camera is located). So it may break things if projects modify distortion locally, but there's no way around it. I think it should still be mentionned


JMargevics (approved):
I have played on my side with Global settings on HDRI Sky distortion and Volumetric clouds and everything works as expected ✅

One minor issue:: you mentioned we can't express distortion speed in km/h and it's mapped to Volumetric Clouds default speed. On the video putting speed to 5k results in way too different scroll animations. Only with 0.01 altitude volumetric clouds have similar visual speed,. Could it be possible to match speeds closer, using volumetric clouds default settings? 

https://user-images.githubusercontent.com/50581102/124116189-5364a400-da77-11eb-90e6-2814d655bad5.mp4

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricClouds.cs
```diff
@@ -347,15 +347,15 @@ public sealed class CloudMapResolutionParameter : VolumeParameter<CloudMapResolu
         /// <summary>
         /// Sets the global wind speed in kilometers per hour.
         /// </summary>
-        [Tooltip("Sets the global wind speed in kilometers per hour.")]
-        public MinFloatParameter globalWindSpeed = new MinFloatParameter(100.0f, 0.0f);
+        [Tooltip("Sets the global wind speed in kilometers per hour.\nThis value can be relative to the Global Wind Speed defined in the Visual Environment.")]
+        public WindSpeedParameter globalWindSpeed = new WindSpeedParameter();
```
anisunity, line 351: relativeWindSpeed with previously serialized?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricClouds.cs
```diff
@@ -347,15 +347,15 @@ public sealed class CloudMapResolutionParameter : VolumeParameter<CloudMapResolu
         /// <summary>
         /// Sets the global wind speed in kilometers per hour.
```
anisunity, line 348: Out of date comment

com.unity.render-pipelines.high-definition/Runtime/Sky/VisualEnvironment.cs
```diff
@@ -12,12 +12,19 @@ namespace UnityEngine.Rendering.HighDefinition
...
         /// <summary>Type of clouds that should be used for rendering.</summary>
         public NoInterpIntParameter cloudType = new NoInterpIntParameter(0);
         /// <summary>Defines the way the ambient probe should be computed.</summary>
         public SkyAmbientModeParameter skyAmbientMode = new SkyAmbientModeParameter(SkyAmbientMode.Static);
 
+        /// <summary>Controls the global orientation of the wind relative to the X world vector.</summary>
+        [Header("Wind")]
+        public ClampedFloatParameter windOrientation = new ClampedFloatParameter(0.0f, 0.0f, 360.0f);
+        /// <summary>Controls the global wind speed in kilometers per hour.</summary>
+        public MinFloatParameter windSpeed = new MinFloatParameter(100.0f, 0.0f);
```
anisunity, line 26: I don't know if this is the default wind value we want for everything

com.unity.render-pipelines.high-definition/Runtime/Sky/VisualEnvironment.cs
```diff
@@ -74,4 +81,146 @@ public sealed class SkyAmbientModeParameter : VolumeParameter<SkyAmbientMode>
...
+                hash = hash * 23 + value.additiveValue.GetHashCode();
+                hash = hash * 23 + value.multiplyValue.GetHashCode();
+
+                return hash;
+            }
+        }
+
+        /// <summary>Returns interpolated value from the visual environment.</summary>
+        /// <returns>The value for this parameter.</returns>
+        public float GetValue(HDCamera camera)
```
anisunity, line 182: Not a big fan of doing volume evaluation inside this nested function, we should pass it the global windValue directly as a parameter instead of passing the camera and evaluating the volume component.

anisunity (commented):
Did a quick review, couple comments.

I didn't try it though, no idea if it works properly, we need a QA pass on this anyway!

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricClouds.cs
adrien-de-tocqueville: to simplify things i marked the old one as obsolete. It may break some code but i don't think a lot of people are using it in a script, do you think it's an issue ?

com.unity.render-pipelines.high-definition/Runtime/Sky/VisualEnvironment.cs
adrien-de-tocqueville: then the question is do we want a global wind parameter if different systems need different value. I think we will see when we have smoething other than the sky using this parameter

com.unity.render-pipelines.high-definition/Runtime/Sky/VisualEnvironment.cs
adrien-de-tocqueville: do we really want to evaluate the volume ourselves when getting the value ? plus that means we would evaluate it for nothing when in custom mode (not a big deal but still)

sebastienlagarde (approved):

pierre-unity:
I've tested it for HDRI, Cloud layers and Volumetric clouds, and it's looking great, and tooltip are fine as well. Thanks!

pierre-unity (approved):

PR #5016, opened 2021-06-30
Expose override camera rendering api

alelievr:
### Purpose of this PR
Expose an API to override the camera parameters inside a custom pass. It adds a lot more flexibility compared to the CustomPassUtils.RenderFrom functions.

Here's an example of the API:

```csharp
using (new CustomPassUtils.OverrideCameraRendering(ctx, override1))
{
    CustomPassUtils.DrawRenderers(ctx, overrideLayerMask);
    ...
}
```

Note that this API works well with `CustomPassUtils.DisableSinglePassRendering` because it allows rendering objects from a single point of view in VR instead of rendering them twice.

Furthermore, there are 4 more functions in the CustomPassUtils API that allows direct render of objects from another camera into a render texture:

- RenderFromCamera(in CustomPassContext ctx, Camera view, RenderTexture targetRenderTexture ...
- RenderDepthFromCamera(in CustomPassContext ctx, Camera view, RenderTexture targetRenderTexture ...
- RenderNormalFromCamera(in CustomPassContext ctx, Camera view, RenderTexture targetRenderTexture ...
- RenderTangentFromCamera(in CustomPassContext ctx, Camera view, RenderTexture targetRenderTexture ...

---
### Testing status
Updated the automated test in the custom pass API scene that tests nested camera override:
image

Updated the examples in https://github.com/alelievr/HDRP-Custom-Passes to use the new API and it works well :) 

https://user-images.githubusercontent.com/6877923/124012721-c02d5f00-d9e1-11eb-81d5-23208bc9ae11.mov

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Fexpose-override-camera-rendering-api/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7581851/job


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -112,6 +112,8 @@ internal static HDRenderPipeline currentPipeline
         ShaderVariablesXR m_ShaderVariablesXRCB = new ShaderVariablesXR();
         ShaderVariablesRaytracing m_ShaderVariablesRayTracingCB = new ShaderVariablesRaytracing();
 
+        internal ref ShaderVariablesGlobal GetShaderVariablesGlobalCB() => ref m_ShaderVariablesGlobalCB;
```
JulienIgnace-Unity, line 115: We should probably not return a ref here. It's usually fine to make your own copy when needed. This will avoid weird unintended side effects.

alelievr: 👍 I pushed the modifications

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Looks good! Tried going through the functions and adding FullScreenCustomPass only for the overridden camera rendering with Antoine help. Haven't encountered any PR specific issues.

sebastienlagarde (approved):

iGNi7e:
Sorry, where can I see updated the examples in https://github.com/alelievr/HDRP-Custom-Passes?

PR #5018, opened 2021-07-01
Keep VFX Graph operator the same width to ease collapse/expand

julienamsellem:
### Purpose of this PR
When collapsing a VFX Graph operator the box becomes smaller moving the collapse/expand button at the same time. This was preventing quick expand/collapse operation since we had to move the mouse each time.

With this fix the operator width is kept constant so that the expand/collapse button always stays at the same position.

### Testing status
Tested only a few operators

### Fogbugz
https://fogbugz.unity3d.com/f/cases/1342313/


VitaSkruibyte (approved):
Looks good! Tried to expand/collapsed inline nodes, blocks of graph, subgraph and VFX Subgraph Operator by using collapsing/expanding arrow, double clicking. For testing used graph that contains all VFX operators. 

Thanks for looking into this, Julien!

julienf-unity (approved):

PR #5019, opened 2021-07-01
Hd/bugfix-1339945: Fix fabric IBL performance & quality

JarkkoUnity:
### Purpose of this PR
Improve quality and performance of fabric IBL pre-convolution. The rendered result is much closer to the reference implementation and uses 1000x less samples and thus performs significantly better.

---
### Testing status
Tested the results against reference implementation for 1352 HDRP test and with various HDR probes showing closer match with the reference. The quality comparison in different scenes (reference implementation ("reference" layers), current implementation ("current" layers), new implementation using 100x less samples ("new" layers), new implementation using 1000x less samples ("new/10" layers): https://drive.google.com/open?id=1-ZDUlRbN1IiwoyCB-7RaLnfDZiQ0N94L&authuser=jarkko.lempiainen%40unity3d.com&usp=drive_fs

The offending project that renders the scene 7 times (bad) also went from ~1fps back to ~10fps. RenderDoc timing for the project shows ~50x perf improvement for single scene rendering (~166ms => ~3ms).

---
### Comments to reviewers
The original Fogbugz case: https://fogbugz.unity3d.com/f/cases/1339945/

You can enable reference rendering by uncommenting the following line in Fabric.hlsl:
// #define FABRIC_DISPLAY_REFERENCE_IBL


com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
```diff
@@ -588,39 +583,15 @@ real4 IntegrateLDCharlie(TEXTURECUBE_PARAM(tex, sampl),
...
         real3 val = SAMPLE_TEXTURECUBE_LOD(tex, sampl, L, mipLevel).rgb;
 
         // Use the approximation from "Real Shading in Unreal Engine 4": Weight ~ NdotL.
-        lightInt +=  val * F * D * Vis;
+        lightInt += val * F * D * Vis;
         cbsdfInt += F * D * Vis;
     }
 
-    return real4(lightInt / cbsdfInt, 1.0);
+    return 0.5f * real4(lightInt / cbsdfInt, 1.0);
```
sebastienlagarde, line 594: can you add a comment to say that the 0.5 improve the result compare to the reference but we don't know exactly why :)

sebastienlagarde:
For QA: Please goal is not to look only at performance here but also to do some comparison with different HDRI between after and previous this PR on multiple fabric object. to see if the result is still coherent and ok. Thanks

sebastienlagarde (approved):

com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
eturquin: I'm gonna try and see if I can understand why we'd need this.

JarkkoUnity: Looks like the reference model was missing pi factor, so now the error is pi/2 instead of 0.5 :) Needs to be investigated where this error is coming from

TomasKiniulis (approved):
Nice to see huge performance leap in the repro project! Seems to stay similar comparing with BSDF disabled now.
Comparing Fabric samples with various HDRIs and exposures with the reference from the drive HDRP/Fabric/CottonWool does look closer to the reference. The results appear to be better in my opinion, with a drawback that it pops more in underexposed areas. Might be good for tech art to look into this too.

A few comparison shots of mine:
**Default HDRI**
BSDF Convolution Disabled
fabricConvolutionDisabled
master
fabricConvolutionEnabledMaster
PR
fabricConvolutionEnabledPR

**HDRI 2** 
master
hdri2_master
PR
hdri2_pr

**Night HDRI**
master
Night_master
PR
Night_pr

**Overexposed HDRI**
master
overexposed_master
PR
overexposed_PR

sebastienlagarde (approved):

sebastienlagarde (approved):

sebastienlagarde:
@eturquin I moved forward with this PR but please have a look and give your through about it to be sure everything is correct. thanks

PR #5029, opened 2021-07-01
HDRP custom passes improvements

alelievr:
### Purpose of this PR
- Add the possibility to hide custom pass from the create menu with the HideInInspector attribute
- Add support of multi-editing on custom pass editor.

Note that the custom pass list is not supported for multi-editing since it's tricky to handle multi-selection in a reorderable list containing different types of data.

Hide the compositor CustomClear pass from the list:
image

---
### Testing status
Played with the custom pass UI

CustomPassMultiSelect

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fcustom-passes-fixes/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7587400/job


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/CustomPass/CustomPassVolumeEditor.cs
```diff
@@ -11,14 +11,15 @@
 
 namespace UnityEditor.Rendering.HighDefinition
 {
-    [CustomEditor(typeof(CustomPassVolume))]
+    [CustomEditor(typeof(CustomPassVolume)), CanEditMultipleObjects]
```
RSlysz, line 14: If this become multi-editable, we must convert all location where we use `target` 
e.g.:
```
        void OnEnable()
        {
            m_Volume = target as CustomPassVolume;
```
in case of multiple selection we must work on each targets

RSlysz (changes requested):
We must fully support each target if we go multi editing it.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/CustomPass/CustomPassVolumeEditor.cs
alelievr: I removed the custom pass list, so all references to m_Volume are unused in multi-selection now.

RSlysz (approved):

pmavridis:
> do you think the HideInInspector attribute should be added to the CustomClear pass?

Yes, definitely! Can you add it in this PR?

pmavridis (approved):

alelievr:
> Yes, definitely! Can you add it in this PR?

Yes :)

TomasKiniulis (approved):
Looks good. HideInInspector is properly hiding passes, Checking Multi-editing works when:
- Same multi selected Mode, Injection Point, Priority values are changed
- Varying multi selected Mode, Injection Point, Priority values are changed 

https://user-images.githubusercontent.com/50582134/124928226-1f095e80-e008-11eb-9e55-363a5dc7b743.mp4

sebastienlagarde (approved):

PR #5034, opened 2021-07-02
Urp/fix to add tooltip of ss-shadows

Seongdae:
### Checklist for PR maker

---
### Purpose of this PR
- Fix case 1327356, [https://fogbugz.unity3d.com/f/cases/1327356/](url)

- And While I was trying large-scale shadows using a RendererFeature of screen space shadows, those shadows were faded out from cascade distance. So I request my idea of shadow-fading on only realtime shadows in screen space shadows pass. Here is what I tried, [https://idea-registry.cloud.unity3d.com/groups/hackweek-2021/ideas/1316](url)

---
### Testing status

---
### Comments to reviewers

### DIFF
--- a/TestProjects/HDRP_RuntimeTests/Assets/Scenes/004-CloudsFlaresDecals/HDRP_Test_Def_Clouds.asset
+++ b/TestProjects/HDRP_RuntimeTests/Assets/Scenes/004-CloudsFlaresDecals/HDRP_Test_Def_Clouds.asset
@@ -42,7 +42,7 @@ MonoBehaviour:
     supportDecalLayers: 1
     supportSurfaceGradient: 0
     decalNormalBufferHP: 0
-    msaaSampleCount: 8
+    msaaSampleCount: 2
     supportMotionVectors: 1
     supportRuntimeAOVAPI: 0
     supportDitheringCrossFade: 0
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -37,7 +37,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added "Conservative Enclosing Sphere" setting to fix shadow frustum culling issue where shadows are erroneously culled in corners of cascades [case 1153151](https://issuetracker.unity3d.com/issues/lwrp-shadows-are-being-culled-incorrectly-in-the-corner-of-the-camera-viewport-when-the-far-clip-plane-is-small)
 - Fixed memory leak with XR combined occlusion meshes. [case 1366173]
 - Fixed a bug with Sprite Targets in ShaderGraph not rendering correctly in game view [1352225]
+- Fix for rendering thumbnails. [case 1348209](https://issuetracker.unity3d.com/issues/preview-of-assets-do-not-show-in-the-project-window)
 - Fixed a regression bug where XR camera postion can not be modified in beginCameraRendering [case 1365000]
+
 
 ## [12.0.0] - 2021-01-11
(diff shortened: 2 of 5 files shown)


com.unity.render-pipelines.universal/Documentation~/renderer-feature-screen-space-shadows.md
```diff
@@ -0,0 +1,41 @@
+# Screen Space Shadows Renderer Feature
```
Seongdae, line 1: @oleks-k I wrote a doc of screen space shadows, please check it :)

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
```diff
@@ -398,7 +398,11 @@ half MainLightShadow(float4 shadowCoord, float3 positionWS, half4 shadowMask, ha
     half shadowFade = half(1.0);
 #endif
 
+#if defined(LIGHTMAP_SHADOW_MIXING) || defined(SHADOWS_SHADOWMASK) || !defined(_MAIN_LIGHT_SHADOWS_SCREEN)
     return MixRealtimeAndBakedShadows(realtimeShadow, bakedShadow, shadowFade);
```
Seongdae, line 402: @phi-lira the main goal of this PR is to fix a fugbuz.

And,
While I was working on far-distance shadows using my another screen space shadows renderer feature, the cascade distance faded out those shadows although there are far distance shadows in the scene.
I thought it needs to exclude fade-out in getting shadows from screen space shadow texture when it didn't have baked shadows. This is my opinion and if this idea had problems in the current situation. let me revert it to the original codes. and I'll request another PR for it.

phi-lira: @lukaschod can you have a look at this?

phi-lira (approved):
Please add graphics tests. Can be in follow up PR or in this one.

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
lukaschod: I am not sure I undestand fully the intention behind the macro usage here. @Seongdae could you eleborate a bit more, as seems like u are removing shadow fade in case of no mixed lighting or am I wron?

Seongdae: Move shadow fade to screen space shadows pass when screen space shadows' feature was enabled in case of no mixed lighting.

In this hackweek, I tried to make a custom screen space shadows' feature for shadowing large-scale, and it could cast shadows on anywhere, but it was faded out beyond the distance urp set. At that moment I though any user can also have trouble with it by their own shadowing solution.

lukaschod: k I understand I would prefer to avoid adding screen space logic into this function. As we already have user pain where we have defines that change completetly functionality without users aware of it.
Maybe we can simply have ScreenSpaceShadows.shader instead of calling MainLightRealtimeShadow use MainLightRealtimeShadow in case of no mixed lighting?

Seongdae:
Hey @JordanL8 thanks :)

JordanL8 (approved):

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
Seongdae: @lukaschod let me think about it and request another PR. the code will be restored in this PR.

lukaschod (approved):

PR #5039, opened 2021-07-05
Fix depthDescriptor.bindMS condition

robin-demoor:
---
### Purpose of this PR
Some Scarlet tests have started failing with the exception: "A non-multisampled texture being bound to a multisampled sampler. Disabling in order to avoid undefined behavior. Please enable the bindMS flag on the texture."

This change fixes the issue.. The BindMS flag was incorrectly disabled on some platforms due to the depthPriming setting.


manuele-bonanno (changes requested):
this whole line of code was added by the depth priming PR. If you look at the VR only #ifdef above, this is exactly the same code, and used to be VR specific code.

I think we should revert the bindMS setting to what it was before the depth priming PR, which seems to have introduced the issue, so just removing this line completely

Are there any issues with that?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -988,7 +988,7 @@ void CreateCameraRenderTarget(ScriptableRenderContext context, ref RenderTexture
                     depthDescriptor.bindMS = depthDescriptor.msaaSamples > 1 && !SystemInfo.supportsMultisampleAutoResolve && (SystemInfo.supportsMultisampledTextures != 0);
 #endif
 
-                    depthDescriptor.bindMS |= depthDescriptor.msaaSamples > 1 && primedDepth && !SystemInfo.supportsMultisampleAutoResolve && (SystemInfo.supportsMultisampledTextures != 0);
+                    depthDescriptor.bindMS |= depthDescriptor.msaaSamples > 1 && !SystemInfo.supportsMultisampleAutoResolve && (SystemInfo.supportsMultisampledTextures != 0);
```
manuele-bonanno, line 992: this is the line of code to remove I mentioned in my comment (which I posted in the wrong place ops :) )

manuele-bonanno (approved):
LGTM, thanks for the changes!

NicoLeyman:
@thomas-zeng Any update on this ? This has been causing certain platform tests to fail for a while.
The fix shouldn't have any fallout for XR but should probably get an ok from the XR side.

thomas-zeng (approved):

thomas-zeng:
Sorry @NicoLeyman for dropping the ball. I have greenlighted the PR.

ellioman (approved):

ernestasKupciunas (approved):
Approving without testing. The issue was triggered in Yamato, and it is fixed now.

ellioman:
Issues on Yamato were not caused by this PR.

PR #5046, opened 2021-07-05
[case 1346694] Make sure Light Editor displays the current RP content on RP switch

jenniferd-unity:
---
### Purpose of this PR
When the active Render Pipeline is switched to a different type, the Light Editor needs to display the custom Editor for the new active SRP.

Fogbugz: https://fogbugz.unity3d.com/f/cases/1346694/ 

Demonstration from HDRP to URP:
1346694-light-inspector-changeSRP-updateEditor

---
### Testing status
Tested in Editor Windows. From URP to HDRP and vice versa. And tested also to Built-in but not as extensively.

---
### Comments to reviewers
I also decided to add an info box in case the Selection behavior breaks. I couldn't find a proper fix that would work 100% of the time. De-selecting and reselecting seems to work more reliably than other methods I have tried.
Note: if the selection workaround breaks, instead of having an error in the Console and empty custom Inspector we now have an infobox for better handling of those situations:
image


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightEditor.cs
```diff
@@ -71,6 +71,14 @@ void OnUndoRedo()
 
         public override void OnInspectorGUI()
         {
+            if (HDRenderPipeline.currentAsset == null)
+            {
+                EditorGUILayout.HelpBox("HDRP is not the active Render Pipeline.", MessageType.Info);
+                Selection.activeObject = null;
+                EditorApplication.delayCall += () => Selection.activeObject = (serializedObject.targetObject);
```
alex-vazquez, line 78: Can we extract this to `CoreEditorUtils`?

alex-vazquez (changes requested):
Looks good, minor comments to improve it a little :)

jenniferd-unity:
We have found a way to rebuild the custom inspectors on RP type changed. Closing this PR and reopening one with the trunk change instead.

PR #5049, opened 2021-07-05
Expose camera motion vector custom pass

alelievr:
### Purpose of this PR

Exposed motion vectors in custom pass. This change shouldn't have too many side effects because the motion vector texture was already accessible but only on the shader side.

In term of API, there is now a member called `cameraMotionVectorsBuffer` in the CustomPassContext class.

---
### Testing status
Added a sample in the HDRP Custom pass sample repository: motion vectors are accessible as expected:
image

Also checked that when motion vectors are disabled, there is no error.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fexpose-camera-motion-vector-custom-pass/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7602013/job


JulienIgnace-Unity (approved):

Vic-Cooper (approved):
Pushed some changes. Happy with it if you are.

TomasKiniulis (approved):
Trying out Motion Vectors for Custom Passes seems to work correctly. Should the Custom Pass API graphics test be updated with the motion vector support?

alelievr:
@TomasKiniulis I don't think it's necessary, our current motion vector tests should be enough.

sebastienlagarde (approved):

PR #5050, opened 2021-07-05
[HDRP] Improve DoF when MSAA is enabled

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1347291/

The resolved depth pyramid (when MSAA is enabled) is using the MAX depth. However, this creates some problems with DoF at the edges of in-focus objects, and is especially visible in thin objects like hair. 

The MAX depth is expanding the edges of in-focus objects, and pixels at the edges who are mostly in the out-of-focus background are treated as in-focus. This creates a 1-pixel ring around the objects, as you see in the screenshot:

image

With this PR, when MSAA is enabled, DoF will use the MIN depth. This mitigates the problem, by conservatively shrinking the edges of the in-focus objects. 
image

The improvement is more noticeable in hair (left before, right after) :
image
(The last screeshot is using the com.unity.unityhair package in 2021.1, so it was created when testing the backport PR.)

---
### Testing status
I have tested with the scenes that you see above.
Note for QA: I can provide the test scenes if you want to double check the PR.

---
### Comments to reviewers
The ideal solution is of course to run DoF per-sample. But we cannot afford to do that, so the solution in this PR is to be "conservative". 

This solution will increase the memory bandwidth when creating the CoC because we read an extra texture. However I don't see another option. We could make the MIN-depth optional, but MAX is clearly full of artifacts, so I don't think it is worth adding yet another option.


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DepthOfFieldCoC.compute
```diff
@@ -45,7 +47,7 @@ void KMainPhysical(uint3 dispatchThreadId : SV_DispatchThreadID)
 {
     UNITY_XR_ASSIGN_VIEW_INDEX(dispatchThreadId.z);
 
-    float depth = LoadCameraDepth(dispatchThreadId.xy);
+    float depth = CameraDepth(_DepthMinMaxAvg, dispatchThreadId.xy);
```
FrancescoC-unity, line 50: Note I guess for me if this gets merged before TAAU, these coordinates for posSS are not valid as they assume output resolution is same as depth buffer res.

FrancescoC-unity (approved):

sebastienlagarde (approved):

iM0ve (changes requested):
**What's tested:**
* Loaded both projects to see the issue and the change
* Set up DOF from scratch on the template and tried various settings
* Checked the Debug mode still works
* DOF with and without MSAA to find unexpected behavior
* All properties on DOF override to see they still work
* DOF with properties from the physical camera, such as Aperture, Focal length, Focus Distance
* Tested the backport PR on the hair project 21.1

**Issues:**
1) In certain setups new errors are appearing _A multisampled texture being bound to a non-multisampled sampler._

https://user-images.githubusercontent.com/36502659/125279342-82eb9a00-e31c-11eb-9ecf-1a60fdb93fd6.mp4

iM0ve (approved):
Updated my review, its good to go

PR #5053, opened 2021-07-05
Add custom post process injection point after DoF and moblur

robcupisz:
### Purpose of this PR
An injection point after depth of field and motion blur allows for rendering a transparent/composited overlay, which still takes advantage of the entire post process stack rendered afterwards.

Examples from our demo are unlit fire and smoke. Neither has a well defined depth, and rendering them at the regular transparency stage won't work well with depth of field or motion blur, both when these transparencies write depth or not. This can be solved by rendering them to a side buffer (using the custom pass hdrp feature, for example), applying a custom simplified DoF/moblur if needed, and compositing with the main buffer in the new AfterPostProcessBlurs injection point.

To achieve the same result without the new injection point, the user would have to disable DoF and moblur, and reimplement both and TAA in user code - all without sharing code with hdrp, since that's not really possible. Reimplementing TAA would be needed as well, since physically-based DoF makes use of an additional TAA pass. As an extra obstacle, custom post process calls aren't currently run as render graph, which means DoF, moblur and TAA implementations would have to be ported to the old style and maintained.

---
### Testing status
- Updated the custom post process automated test to include the new injection point. The output image is unchanged, but skipping any one of the post process effects will still change the final output.
- The code is being used in our demo.

---
### Comments to reviewers
Taking suggestions for an alternative name to AfterPostProcessBlurs.

Need to contact the docs team to prepare an updated version of the [control flow diagram in the docs](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@12.0/manual/Custom-Post-Process.html) to include the new injection point. But I already see it's outdated, since it doesn't include the BeforeTAA injection point.


TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/4x_PostProcessing/4060_CustomPostProcess/BluePP.shader
```diff
@@ -43,7 +43,7 @@
         UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
         uint2 positionSS = input.texcoord * _ScreenSize.xy;
-        float2 outColor = LOAD_TEXTURE2D_X(_InputTexture, positionSS).xy;
+        float2 outColor = LOAD_TEXTURE2D_X(_InputTexture, positionSS).xz;
```
sebastienlagarde, line 46: why this change?

com.unity.render-pipelines.high-definition/Documentation~/Custom-Post-Process.md
```diff
@@ -213,12 +215,14 @@ HDRP allows you to customize the order of your custom post-processing effect at
...
 
-1. Before Transparent.
+1. After Opaque And Sky.
 
-2. Before TAA
+2. Before TAA.
 
 3. Before Post Process.
 
+4. After Post Process Blurs.
```
sebastienlagarde, line 224: need to increment next Postprocess to 5.0

alelievr (approved):

sebastienlagarde:
(Sorry the label was missing and thus haven't gone through my various review - will be checked soon. poke @alelievr as well.

TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/4x_PostProcessing/4060_CustomPostProcess/BluePP.shader
sebastienlagarde: @alelievr / @robertcupisz  is it expected?

sebastienlagarde (approved):

PR #5054, opened 2021-07-05
Fix GBuffer depth debug mode

adrien-de-tocqueville:
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1346923/

In deferred mode, some materials are not rendered during the depth prepass if not needed. They wil later be renderer by the gbuffer pass.
Changed the gbuffer debug pass to use the full depth buffer instead of the one copied at the end of the depth prepass, which is incomplete

---
### Testing status
Tested on the repro project
All materials writing to depth are now visible with the debug mode
Note: Even forward materials are visible, but it was already the case before this PR


sebastienlagarde (approved):

remi-chapelain (approved):
Thanks for the explanation on the description. 
Trusting developers testing skills on this one :)

PR #5057, opened 2021-07-06
HDRP - HDRP Wizzard can be opened despite there is no Render Pipeline in use

alex-vazquez:
---
### Purpose of this PR
FIX : https://fogbugz.unity3d.com/f/cases/1347461/

---
### Testing status
- Check that if no RPAsset is in use, the HDRP wizard menu is enabled.


arttu-peltonen (approved):

TomasKiniulis (approved):
Thanks for the fix!

- Checked that wizard still pop ups and is not grayed out in the list when hdrp is added to manifest in built-in.
- Checked that assigning URP asset grays out wizard and assigning back to built-in leaves it available again.

PR #5059, opened 2021-07-06
Added an option to disable runtime debug window hotkey.

adrien-de-tocqueville:
---
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1345783/

it is now possible to disable the debug window hotkey by doing:
```cs
DebugManager.instance.enableWindowHotkey = false;
```

Ideally, the hotkey should be customizable but it is a bit more involved to do since we have to support the two input systems. (Created a JIRA task: https://jira.unity3d.com/browse/HDRP-1239)

---
### Testing status

tested in playmode and in player build


sebastienlagarde:
@adrien-de-tocqueville  can you update the documentation so users know they can do this. Put it in the debug menu

iM0ve (approved):
Testing should be sufficient for the scope of the change

sebastienlagarde (approved):

PR #5061, opened 2021-07-06
Added a property on the HDRP asset to allow users to avoid ray tracing effects running at too low percentages (case 1342588).

anisunity:
After some thinking, the only viable solution we found was to add a parameter to define until which sub-resolution half res ray tracing effects are allowed.
https://fogbugz.unity3d.com/f/cases/1342588/

Testing status
Due to a bug, tests are failing randomly even on master (locally). So we'll have to run on yamato and make sure it's fine


kecho (approved):
Make sure to run the formatter, otherwise looks good.

anisunity:
Formatting fixed

remi-chapelain (changes requested):
We need to update the documentation for this new checkbox in the dynamic resolution section of the HDRP asset. 

Apart from that, feature is working well, it's not ideal but that's the best we can do for now. 
If the screen percentage drops below the RT Half resolution threshold, all RT effects running at half resolution (RTR perf & RTGI perf) are force to full screen. 

I added two new tests to prevent regression for this. 
Basically, camera on the left uses DRS at 90% and the RT half threshold is 100%, so the half resolution get forced back to full resolution. Native camera on the right stays un-affected so keeps the half resolution. 

1001_HalfResolutionThreshold_RTR
1001_HalfResolutionThreshold_RTGI

remi-chapelain (approved):
All good 🟢

anisunity:
All green except formatting (that i need to get from master), test will probably be -reinstable after rebase 
image

anisunity:
hallelujah

sebastienlagarde (approved):

PR #5063, opened 2021-07-06
Fix object disappearing from lookdev

alelievr:
### Purpose of this PR
Fixed objects disappearing from Lookdev window when entering playmode: https://fogbugz.unity3d.com/f/cases/resolve/1309368/

---
### Testing status
Entered play mode after putting a cube from hierarchy in the lookdev

https://user-images.githubusercontent.com/6877923/124590734-c4cb9a80-de5b-11eb-8c25-42f5ef080093.mov

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Flookdev-object-disappear/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7655211/job


com.unity.render-pipelines.core/Editor/LookDev/LookDev.cs
```diff
@@ -43,6 +49,16 @@ internal static Context currentContext
...
         }
 
+        static Context TryFindCurrentContext()
+        {
+            Context context = s_CurrentContext;
+
+            if (context != null)
+                return context;
+
+            return Resources.FindObjectsOfTypeAll<Context>().FirstOrDefault();
```
RSlysz, line 59: Can we be sure that we not load a context that should be garbage collected ?

RSlysz (approved):
🎉

com.unity.render-pipelines.core/Editor/LookDev/LookDev.cs
alelievr: I'm not sure to see why it matters?

In any case, the static field is cleared after the domain reload, if the ScriptableObject survived the domain reload, then I expect it to never be garbage collected.

sebastienlagarde (approved):

TomasKiniulis (approved):
Tested with multiple environment libraries, environments entering and exiting out of playmode with prefabs and regular objects dragged from Hierarchy and Project view. Objects are properly showed in edit mode and playmode now.

PR #5068, opened 2021-07-06
Fixed exposure not being properly handled in ray tracing performance (RTGI and RTR, case 1346383).

anisunity:
Original problem described here:
https://forum.unity.com/threads/feature-request-better-denoiser-for-gi.1132984/

Major fixes of this PR
Exposure is now properly handeled as well as exposure weight for materials.
Sky is always clamped in HSV space
All materials are clamped in RGB space for RTR
All materials are clamped in HSV for RTGI

A regression was introduced in this PR https://github.com/Unity-Technologies/Graphics/pull/4340
image

Added new tests to cover properly all the cases.

Testing status
The new tests pass locally.
Pair tested with @remi-chapelain


remi-chapelain (approved):
Has already been pair tested extensively. 
Thanks for extending the coverage of the previous tests (added RTGI and high exposure that makes intensity values go over the buffer limit maximum value)

com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/UnlitRaytracing.hlsl
```diff
@@ -13,7 +13,7 @@ void FitToStandardLit( BSDFData bsdfData
     //Note: we have to multiply everything with the inverse exposure, since the result buffer expects everything to be 'pre exposed'.
```
anisunity, line 13: Update this

remi-chapelain (approved):

anisunity:
Fixed comment and launched tests

sebastienlagarde (approved):

PR #5071, opened 2021-07-06
 [HOLD][case 1347209] Only process materials of the same SRP that the MaterialPostProcessor

jenniferd-unity:
---
### Purpose of this PR
When having both URP and HDRP packages in the same project, we have an error in the Console with infinite import loops regarding Decals.
This PR aims to enforce the idea that each SRP's Material Post Processor is only triggered on the SRP it serves.

Fogbugz https://fogbugz.unity3d.com/f/cases/1347209/

---
### Testing status
On Windows Editor only: tried the repro step described in the fobgubz above both with 2022.1 and 2021.2.

---
### Comments to reviewers
The halo effect for this PR can be high, make sure to give any feedback about how this change could impact users or our tests.


com.unity.render-pipelines.universal/Editor/AssetPostProcessors/MaterialPostprocessor.cs
```diff
@@ -25,6 +26,24 @@ class MaterialReimporter : Editor
     {
         static bool s_NeedToCheckProjSettingExistence = true;
 
+
+        internal static bool IsURPShader(Shader shader)
```
RSlysz, line 30: Should be part of ShaderUtils

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/MaterialPostprocessor.cs
```diff
@@ -38,6 +57,11 @@ static void ReimportAllMaterials()
                 materialIdx++;
                 var path = AssetDatabase.GUIDToAssetPath(asset);
                 EditorUtility.DisplayProgressBar("Material Upgrader re-import", string.Format("({0} of {1}) {2}", materialIdx, totalMaterials, path), (float)materialIdx / (float)totalMaterials);
+                var material = (Material)AssetDatabase.LoadAssetAtPath(asset, typeof(Material));
```
RSlysz, line 60: Sadly this will slow up the import :/
Not sure how we can prevent it.

RSlysz (commented):
I'm not that fluent in URP but it seams to me that this is the fix to do.
Though there was no loading or the material in URP and this will slow everything a little :/

In HDRP there was no reason to not skip early. And this was potentially importing shadergraph from other RP. Seamed weird.

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/MaterialPostprocessor.cs
jenniferd-unity: @bastienunity do you have an idea on how i can properly enforce to process Materials from only the RenderPipeline I am interested in?

bastienunity: I don't think you can as you have to load them in memory to figure out the shader.
However, there are probably some significant improvements that could be made in this method:
- Don't use AssetDatabase.FindAssets("t:material", null); but AssetDatabase.FindAssets("glob:\"*.mat\""); instead. This will find guids only for .mat assets and will avoid loading in memory materials that you are not supposed to change.
- What is AssetDatabase.ImportAsset doing? You probably want to do an AssetDatabase.SaveAssetIfDirty instead?
- If Importers create Materials using a Shader that needs to be updated based on the SRP installed/active, then they need to declare a dependency upon it. We can discuss that separately on what is expected and how we can implement it. In any case, it's not viable to modify them after the import like that. They are never being saved back anywhere, and when unloaded they may be reverted to before your change (especially given the ModelImporter is always using a local caching now)

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/MaterialPostprocessor.cs
jenniferd-unity: I would agree but in URP ShaderUtils is part of the Runtime portion and most of these function are Editor only (IsShaderGraphAsset for ex)

com.unity.render-pipelines.universal/Editor/AssetPostProcessors/MaterialPostprocessor.cs
jenniferd-unity: Thanks! 

- I implemented the glob search suggestion,

- i think we are retriggering import to ensure the Material Upgrader is run. I dont think SaveAsset will trigger this behavior. no?
- i guess this point is out of scope for this bugfix but yes lets discuss it - do you agree?

jenniferd-unity: note: Bastien and I had a discussion there are some things I could implement for this PR.
cc @phi-lira

erikabar (approved):
verified the fix on mac

jenniferd-unity:
Will not submit for now -- it actually introduces other issues (for ex. adding URP to an empty builtin project) and it may be better to fix everything at the same time

PaulDemeulenaere (approved):
I was worried about this change :
> string[] guids = AssetDatabase.FindAssets("t:material", null);

to

> string[] guids = AssetDatabase.FindAssets($"glob:\"*.mat\"");

Because the VFX is storing material as subasset, these materials are generated during VisualEffectAsset import and shouldn't be modified by something else.
The search by extension seems safer from that aspect.
🟢 

> Will not submit for now -- it actually introduces other issues (for ex. adding URP to an empty builtin project) and it may be better to fix everything at the same time

👍

RSlysz (approved):

phi-lira:
@RSlysz this PR is staled for a while. Does your team plan to act on it?

RSlysz:
FB have been fixed partly by Unity update and other PR here: https://github.com/Unity-Technologies/Graphics/pull/6185 This is not required anymore.

PR #5073, opened 2021-07-06
[Fogbugz # 1346871] Fixing nvidia install button and preventing user from spamming it

kecho:
### Purpose of this PR
NVIDIA package which contains DLSS is not installed on HDRP by default. For this we have an install button when the user first explores Dynamic Resolution settings.
The button we had was not the standard 'fix' button so I changed it to this.
I also added a global that prevents the button from being spammed (see the fogbugz)
Fogbugz: https://fogbugz.unity3d.com/f/cases/1346871/

---
### Testing status
* Attempted to reproduce the fogbugz (see repro steps)
* Tested with new button on windows. 

Before: 
image

After:
image

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.cs
@@ -488,7 +488,10 @@ namespace UnityEditor.Rendering.HighDefinition
                 serialized.renderPipelineSettings.lightLoopSettings.maxLightsPerClusterCell.intValue = Mathf.Clamp(serialized.renderPipelineSettings.lightLoopSettings.maxLightsPerClusterCell.intValue, 1, HDRenderPipeline.k_MaxLightsPerClusterCell);
         }
 
+#if ENABLE_NVIDIA && !ENABLE_NVIDIA_MODULE
         static bool s_DisplayNvidiaModuleButtonInstall = true;
+#endif
+
         static void Drawer_SectionDynamicResolutionSettings(SerializedHDRenderPipelineAsset serialized, Editor owner)
         {
             EditorGUILayout.PropertyField(serialized.renderPipelineSettings.dynamicResolutionSettings.enabled, Styles.enabled);


remi-chapelain (approved):
Thanks for the fix. I trust developer's mashing button skill enough to approve 🟢 
The new behavior (button disappearing) and naming is the same than in the HDRP wizard so it makes sense in terms of consistency.

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -272,6 +272,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed remapping of depth pyramid debug view
 - Fixed an issue with asymmetric projection matrices and fog / pathtracing. (case 1330290).
 - Fixed gbuffer depth debug mode for materials not rendered during the prepass.
+- Fixed NVIDIA install button, changed it to a Fix button and also prevent user from spamming and corrupting installation of nvidia package.
```
alex-vazquez, line 275: If I were you I would split the changelog in two.
```
### Fixed
- Prevent user from spamming and corrupting installation of nvidia package.

### Changed
- Changed the NVIDIA install button to the standard FixMeButton.
```
As the change of the button is not a fix

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.cs
```diff
@@ -581,13 +582,13 @@ static void Drawer_SectionDynamicResolutionSettings(SerializedHDRenderPipelineAs
             --EditorGUI.indentLevel;
 
 #if ENABLE_NVIDIA && !ENABLE_NVIDIA_MODULE
-            EditorGUILayout.BeginHorizontal();
-            EditorGUILayout.HelpBox(Styles.DLSSPackageLabel, MessageType.Info);
-            if (GUILayout.Button(Styles.DLSSInstallButton, GUILayout.ExpandWidth(true), GUILayout.ExpandHeight(true)))
+            if (s_DisplayNvidiaModuleButtonInstall)
```
alex-vazquez, line 585: We shouldn't check if the `com.unity.modules.nvidia` is being already added into the `PackageManager`?

kecho: 1) When the nvidia package is already in the manager the preprocessor directive ENABLE_NVIDIA_MODULE is true, so this UI doesnt get painted at all. Is through pre processor directives that we can see if a moudle is included or not. See the asmdef of this package, thats where you map the preprocessor directives.

2) The global s_DisplayNVIDIAModuleButtonInstall is part of a bug, there is a 3 second delay when the button gets pressed and until HDRP gets recompiled. During this time if the user spams click then we get a failure.

com.unity.render-pipelines.high-definition/CHANGELOG.md
kecho: Done

alex-vazquez (approved):
Looks good, Thanks

sebastienlagarde (approved):

PR #5076, opened 2021-07-06
Fix camera rotation uncontrollable with low framerate

alelievr:
### Purpose of this PR
Fix camera moving too much with low framerate in FreeCamera, SimpleCameraController and LookWithMouse.

https://fogbugz.unity3d.com/f/cases/resolve/1340344/

---
### Testing status

Tested FreeCamera, SimpleCameraController and LookWithMouse with both new and old input system, movement in consistent with framerate:

60fps:

https://user-images.githubusercontent.com/6877923/124631106-5a2e5500-de83-11eb-9b0b-fc84d412c1ac.mov

5fps:

https://user-images.githubusercontent.com/6877923/124631102-5995be80-de83-11eb-9b74-94accbd63ea0.mov

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fcamera-controller-template/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7721737/job


JulienIgnace-Unity (commented):
Do we really want to remove any deltaT component from the computation?
As it is here, I reckon the camera will feel very different depending on the framerate. Did you try to test with various frame rates to see how it behaves?

alelievr:
Yeah, I tried with different framerates, the issue is that the mouse movement we get is already a delta from last frame, so multiplying a delta with another screwed the result.

Note that this is only true if the update loop actually slows down (CPU bound), and not if we lower the framerate manually with Application.targetFrameRate

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

TomasKiniulis (approved):
Looks good to me. Comparing template between master and PR while stressing my CPU to force low framerate does feel like a subtle change for the better to camera responsiveness

PR #5078, opened 2021-07-06
Improving the denoiser for RTGI (case 1346383)

anisunity:
This PR improves the upscaling process as well as the denoising of RTGI (and RTAO), it is one of the two PRs to solve this bug
https://fogbugz.unity3d.com/f/cases/1346383/

Removes the upscaling parameter for RTGI (forced to 3 for half res, 4 for full res).
Reduces signficantly the cost of the upscaling pass.
Resolves bleeding (and ghosting artifacts) that come from the upscale.
Change the default parameters for quality settings for RTGI.
Fix edges issues with the half resolution denoiser.
Deduce automatically the world space denoiser radius for surfaces (instead of fixed for the whole scene)

https://youtu.be/J5UJRT5fF6E
https://youtu.be/GZwg9ditAzk

Testing status
DXR Tests are green after updating the screenshots.


anisunity:
https://youtu.be/JNtsRVlo6Iw

anisunity:
After running the test a thousand times, all green

remi-chapelain (changes requested):
I quickly tested performance on the ray tracing apartment scene in quality mode and in performance medium mode and surprisingly, I [found performance a bit worse than the current master](https://docs.google.com/spreadsheets/d/1uGIh7dviI6N36tmZFXAf2TnaA42KjiknrqevB2l7itI/edit#gid=0). Needs to be discussed later. 

I honestly don't know how we can solve this but since values are serialized in the HDRP asset, the only way to get the new values from the quality settings is to create a new HDRP Asset, which is a terrible pain for upgrading. It should at least be mentionned in the upgrade guide or something. 

Also, only managed to repro those issues on those specifics projects, but they didn't repro on master so they must be linked with the improvement in a way. 

Found some NaNs when we use the perf mode denoiser in this project. Doesn't repro in quality mode and on master. 
https://user-images.githubusercontent.com/57442369/125320240-cc97ad00-e33b-11eb-8241-d9c489ab7ff5.mp4


And in perf mode again, I can see some green appearing on the right side of the view. Doesn't repro in quality mode and on master. 
Uploading Appartement-Green.mp4…


Hit me up when you are getting back on this to either pair debug it or for me to send you repros.

anisunity:
@remi-chapelain  The counters for the perf are not working properly and cannot be trusted. I'll check the perf comparison using a GPU capture to make sure that there is no major regression.

Let's sync for the issues that you found, so that you give me a repro.

anisunity:
I profiled, it is impossible that it is more expensive 
image

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalLightingQualitySettings.cs
```diff
@@ -110,9 +106,9 @@ internal GlobalLightingQualitySettings()
             RTGIHalfResDenoise[(int)ScalableSettingLevelParameter.Level.Medium] = false;
             RTGIHalfResDenoise[(int)ScalableSettingLevelParameter.Level.High] = false;
 
-            RTGIDenoiserRadius[(int)ScalableSettingLevelParameter.Level.Low] = 0.75f;
-            RTGIDenoiserRadius[(int)ScalableSettingLevelParameter.Level.Medium] = 0.5f;
-            RTGIDenoiserRadius[(int)ScalableSettingLevelParameter.Level.High] = 0.25f;
+            RTGIDenoiserRadius[(int)ScalableSettingLevelParameter.Level.Low] = 1.0f;
```
FrancescoC-unity, line 109: Given it is all the same value, does it make sense to keep as a scalability setting? (is it beneficial if user change this defaults or should they always be the same value?)

anisunity:
image
 The perf is equal at worst (full res, but higher visual quality), much better at best (half res)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalLightingQualitySettings.cs
anisunity: Users can chose to lower it for lower perf modes it as it can affect performance a lot

FrancescoC-unity: If it has an high impact, shouldn't this have different values by default too though? Out of the box the scalability does nothing here.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/Denoising/DiffuseDenoiser.compute
```diff
@@ -165,30 +177,62 @@ void BILATERAL_FILTER(uint3 dispatchThreadId : SV_DispatchThreadID, uint2 groupT
...
+    #endif
+    float depthValue = LOAD_TEXTURE2D_X(_DepthTexture, sampleCoord).x;
+    gs_cacheDepth[groupIndex] = depthValue;
+}
+
+uint OffsetToLDSAdress(uint2 groupThreadId, int2 offset)
+{
+    // Compute the tap coordinate in the 8x8 grid
+    uint2 tapAddress = (uint2)((int2)(groupThreadId) + offset);
+    return clamp((uint)(tapAddress.x) + tapAddress.y * GATHER_REGION_SIZE, 0, GATHER_REGION_SIZE_2 - 1);
```
FrancescoC-unity, line 203: No need for cast here, tapAddress.x is already a uint

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/Denoising/DiffuseDenoiser.compute
```diff
@@ -165,30 +177,62 @@ void BILATERAL_FILTER(uint3 dispatchThreadId : SV_DispatchThreadID, uint2 groupT
...
+    // Read the high res depth
+    int outputIdx = OffsetToLDSAdress(groupThreadId, int2(0, 0));
+    float targetDepth = gs_cacheDepth[outputIdx];
+
+    // Compute the 2x2 pixelregioncorner
+    uint2 corner = centerCoord.xy - uint2(centerCoord.x & 1, centerCoord.y & 1);
+    uint2 cornerGroupThread = corner - groupId * DIFFUSE_DENOISER_TILE_SIZE;
+
+    // Grab the indices of the sub-region to use
+    int ldsIdx0 = OffsetToLDSAdress(cornerGroupThread, int2(0, 0));
```
FrancescoC-unity, line 226: So I see you don't load a safety border so it is clamp at _group_ border which I guess can give problems if we have signficiant depth discontinuities happening at the edge of the group? Might be ok, but leaving comment here to see if needs further discussion or not.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalLightingQualitySettings.cs
anisunity: To be disccused

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/IndirectDiffuse/RaytracingIndirectDiffuse.compute
```diff
@@ -128,109 +132,323 @@ Texture2DArray<float> _BlueNoiseTexture;
...
+    // Initially the weight of this pixel is evaluate using the depth
+    float weight = 1.0f;
+
+    // If the candidate pixel is a background pixel, we cannot use it
+    if(neighborData.linearDepth == 1.0)
+        weight = 0.0f;
+
+    // If the normals of both pixels are too different, we cannot use it
+    if(dot(neighborData.normalWS, normalWS) < NORMAL_REJECTION_THRESHOLD)
+        weight *= 0.1f;
```
FrancescoC-unity, line 156: Would it make sense to soft reject this? Lerping towards 0.1 (or close to) multiplier as we approach the threshold?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/Denoising/DiffuseDenoiser.compute
anisunity: Not sure to get what you are describing, let's chat

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/IndirectDiffuse/RaytracingIndirectDiffuse.compute
anisunity: I guess we can have a transition region. But the idea here is to not use these unless there are not better samples

FrancescoC-unity (approved):

sebastienlagarde (approved):

anisunity:
Everything is green

anisunity:
Conflits, need to merge

anisunity:
Everything is green @remi-chapelain depends on your review

remi-chapelain (approved):
Couldn't repro any of the issues I reported above. ✔️ 

For the rest, I could only see improvements, couldn't find any more regression on the scenes I tested. ✔️ 

Also tested animation because we don't have coverage for moving objects. ✔️ 

Pushed the changes regarding the upscale filter to make the documentation match.

PR #5079, opened 2021-07-06
[Shadergraph] Don't dirty all the nodes 25 times when selecting a blackboard keyword

cdxntchou:
### Purpose of this PR
Fix for perf regression https://fogbugz.unity3d.com/f/cases/1347666/

This was an existing piece of very slow code that was now being called as a result of selecting a keyword property (whereas previously it would only run when a value was changed by the user).

Several changes here:
1) We make the slow code a good bit faster (300 ms => ~45 ms) in the case where it actually needs to run because of a user change.  This is the change to the Enum keyword UpdatePorts in KeywordNode.cs -- basically it was doing a bunch of individual modifications to remove, add, and reconnect the slots, that called the expensive ValidateGraph() at each step.
2) When the enum keyword node slots are not re-ordered we don't call the expensive Dirty(Topological), taking UpdatePorts down to < 0.1ms.
3) ShaderInputViewController.ModelChanged was taking a lot of time calling the expensive DirtyNodes() function unnecessarily.  The major problem was that every view controller was responding to every action, even if the action didn't affect them.  So one action would result in N calls into DirtyNodes(), where N is the number of properties, instead of just one call for the property that was changed.
4) In the case of the ChangedFlag action -- we were getting this action invoked when we selected a keyword property.  It was always a no-op action, so I added code to detect the no-op case and skip calling DirtyNodes() unnecssarily.

These changes took the HandleGraphChanges() function from ~2500 ms to < 0.07 ms (yay 35000x faster!) when selecting a keyword property in the blackboard, and should vastly improve perf when modifying blackboard properties as well.

You get similar performance improvements when changing values on properties now too!  Changing the exposed flag on a keyword property used to take ~6700 ms, now it takes ~720 ms (90% of this is kicking off new preview shader compilation).

There is still a minor hitch when selecting _any_ property, but it is now the result of "Update Style" being applied to change the appearance of the selected nodes -- which is running ~220 ms.   I'll leave that for a separate investigation.

---
### Testing status

- [x] Tested enum keyword selection -- performance hitch is now inline with previous behavior, same as other properties, and less than 1/4 s.
- [x] Tested modifying enum keywords:
- [x] Exposed checkbox works, is undo/redoable.
- [x] Entry reordering reorders the node slot order, keeps connections and default values.
- [x] Changing display name changes name of nodes, is undo/redoable.
- [x] Changing any of the other states is undo/redoable.
- [x] Changing any bool keyword state works, is undo/redoable.

Issues not fixed in this PR:
- [ ] Reordering enum keyword entries is not undo/redoable.  Pre-existing bug.
- [ ] Renaming an enum keyword entry will lose slot connection and default values.  Pre-existing bug.
- [ ] Minor hitch on selecting any property because of "Update Style" taking ~220 ms.

Yamato:

ShaderGraph PR: 🟢
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252F1347666/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_2021.2/7570693/job/pipeline

---
### Comments to reviewers


Nightmask3 (approved):
LGTM!

com.unity.shadergraph/Editor/Data/Nodes/Utility/KeywordNode.cs
```diff
@@ -85,54 +85,41 @@ void UpdatePorts()
                     break;
                 }
                 case KeywordType.Enum:
+                using (var inputSlots = PooledList<MaterialSlot>.Get())
```
Nightmask3, line 88: Neat, I didn't know we had a PooledList implementation we could use 🙂

Nightmask3 (approved):
Some good optimizations in there for the enum keywords in general
Approving 👍

Nightmask3:
Needs a formatter run and changelog resolution but looks good to merge other than that

com.unity.shadergraph/Editor/Drawing/Controllers/ShaderInputViewController.cs
```diff
@@ -244,27 +251,46 @@ protected override void ModelChanged(GraphData graphData, IGraphDataAction chang
...
                 case ChangeDisplayNameAction changeDisplayNameAction:
-                    ViewModel.inputName = Model.displayName;
-                    DirtyNodes(ModificationScope.Topological);
-                    m_SgBlackboardField.UpdateFromViewModel();
+                    if (changeDisplayNameAction.shaderInputReference == Model)
+                    {
+                        ViewModel.inputName = Model.displayName;
+                        DirtyNodes(ModificationScope.Topological);
+                        m_SgBlackboardField.UpdateFromViewModel();
+                    }
```
jessebarker, line 293: Can this not be done as:
`if (changeDisplayNameAction.shaderInputReference != Model)
 {
      return;
 }`
outside of the switch?  You'd still want to preserve the additional condition in the changeExposedFlagAction case (only call DirtyNodes if the value changed).

jessebarker (approved):
LGTM

com.unity.shadergraph/Editor/Drawing/Controllers/ShaderInputViewController.cs
cdxntchou: Only if we modified the interface to require that all actions implement the shaderInputReference property... I don't think all of them operate on ShaderInputs..

cdxntchou: We could clean it up further to make it nice, but all of this code is going to die in GTF...

jessebarker: Gotcha.  Good to go.

ghost (approved):
Developer testing is sufficient

PR #5080, opened 2021-07-06
[Fobguz # 1348357] VFX Particle templates missing stencil flags

kecho:
### Purpose of this PR
VFX particles missing the stencil flags. This means that forward particles are not getting SSS / option to exclude from TAA isnt working. 
The fix involves introducing the flags for forward permutations.

Fogbugz case: https://fogbugz.unity3d.com/f/cases/1348357/

---
### Testing status
* Verified by folling bug repro steps
* Verified by looking at stencil value for "exclude from TAA" from renderdoc.

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #5081, opened 2021-07-06
[Fogbugz # 1345143] DLSS: Aleviating some of the ghosting artifacts for partciles by using the color bias mask

kecho:
### Purpose of this PR
Fogbugz: https://fogbugz.unity3d.com/f/cases/1345143/

DLSS has heavy ghost artifacts on pixels that lack velocity. This change uses the color bias mask feature of DLSS to aleviate some of the heavy ghosting.

This PR creates a new pass in graphics that reads stencil, and blits a mask of 1,1,1,1 if  the stencil passes. The stencil ref used is  transparent passes "Exclude from TAA". This results in a mask for particles that DLSS can use.

---
### Testing status
Tested locally by using the HDRP template on windows dx11, dropping some butterflies and falling leaves VFX, making them transparent, and clicking on the "Exclude from TAA" checkbox.

Further testing:
* Must test on dx12 & vulkan
* Should test on hardware/software 

Dont expect ghosting to be completely gone, there is still ghosting remaining. This change should alleviate some of the ghosting.

---
### Comments to reviewers
I discovered  that VFXStencilForward macro isnt used, thus the "Exclude from TAA" option never worked. I have another PR https://github.com/Unity-Technologies/Graphics/pull/5080 that fixes this problem (is an isolated cherry pick). 5080 must go in first.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DLSSBiasColorMask.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DLSSBiasColorMask.shader
@@ -1,0 +1,70 @@
+Shader "Hidden/HDRP/DLSSBiasColorMask"
+{
+    Properties
+    {
+        [HideInInspector] _StencilRef("_StencilRef", Int) = 2
+        [HideInInspector] _StencilMask("_StencilMask", Int) = 2
+    }
+
+    HLSLINCLUDE
+    #pragma target 4.5
+    #pragma only_renderers d3d11 playstation xboxone xboxseries vulkan metal switch
+    #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
+
+    struct Attributes
+    {
+        uint vertexID : SV_VertexID;
+        UNITY_VERTEX_INPUT_INSTANCE_ID
+    };
+
+    struct Varyings
+    {
+        float4 positionCS : SV_POSITION;
+        float2 texcoord   : TEXCOORD0;
+        UNITY_VERTEX_OUTPUT_STEREO
+    };
+
+    Varyings Vert(Attributes input)
+    {
+        Varyings output;
+        UNITY_SETUP_INSTANCE_ID(input);
+        UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(output);
+        output.positionCS = GetFullScreenTriangleVertexPosition(input.vertexID);
+        output.texcoord = GetFullScreenTriangleTexCoord(input.vertexID);
+        return output;
+    }
+
+    void DLSSBiasColorMaskPS(Varyings input, out float4 outColor : SV_Target0)
+    {
+        UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
+        outColor = float4(1,1,1,1);
+    }
+
+    ENDHLSL
+
+    SubShader
+    {
+        Tags{ "RenderPipeline" = "HDRenderPipeline" }
+
+        Pass
+        {
+            Stencil
+            {
+                ReadMask [_StencilMask]
+                Ref     [_StencilRef]
+                Comp Equal
+                Pass Keep
+            }
+
+            ZWrite Off ZTest Always Blend Off Cull Off
+
+            HLSLPROGRAM
+                #pragma vertex Vert
+                #pragma fragment DLSSBiasColorMaskPS
+            ENDHLSL
+        }
+    }
+
+    Fallback Off
+}
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
@@ -173,6 +173,7 @@ namespace UnityEngine.Rendering.HighDefinition
         System.Random m_Random;
 
         bool m_DLSSPassEnabled = false;
+        Material m_DLSSBiasColorMaskMaterial;
         DLSSPass m_DLSSPass = null;
         void InitializePostProcess()
         {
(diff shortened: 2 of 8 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -599,6 +601,44 @@ TextureHandle RenderAfterPostProcessObjects(RenderGraph renderGraph, HDCamera hd
...
+
+        TextureHandle DoDLSSColorMaskPass(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle inputDepth)
+        {
+            TextureHandle output = TextureHandle.nullHandle;
+            using (var builder = renderGraph.AddRenderPass<DLSSColorMaskPassData>("DLSS Color Mask", out var passData, ProfilingSampler.Get(HDProfileId.DeepLearningSuperSamplingColorMask)))
+            {
+                output = builder.UseColorBuffer(renderGraph.CreateTexture(
+                    new TextureDesc(Vector2.one, true, true)
+                    {
+                        colorFormat = GraphicsFormat.R8G8B8A8_UNorm,
```
FrancescoC-unity, line 619: I assume the format is required by DLSS? Sounds a bit of a shame to use 32 bit for a binary mask

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -599,6 +601,44 @@ TextureHandle RenderAfterPostProcessObjects(RenderGraph renderGraph, HDCamera hd
...
+                passData.colorMaskMaterial = m_DLSSBiasColorMaskMaterial;
+
+                passData.destWidth = hdCamera.actualWidth;
+                passData.destHeight = hdCamera.actualHeight;
+
+                builder.SetRenderFunc(
+                    (DLSSColorMaskPassData data, RenderGraphContext ctx) =>
+                    {
+                        Rect targetViewport = new Rect(0.0f, 0.0f, data.destWidth, data.destHeight);
+                        ctx.cmd.SetViewport(targetViewport);
```
FrancescoC-unity, line 634: We should be setting the stencil bits from here too instead of relying on the fact that the hardcoded  one will remain the same, i.e. 

`data.colorMaskMaterial.SetInt(HDShaderIDs._StencilMask, (int)StencilUsage.ExcludeFromTAA);
data.colorMaskMaterial.SetInt(HDShaderIDs._StencilRef, (int)StencilUsage.ExcludeFromTAA);`

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DLSSBiasColorMask.shader
```diff
@@ -0,0 +1,73 @@
...
+        [HideInInspector] _StencilRef("_StencilRef", Int) = 2
+        [HideInInspector] _StencilMask("_StencilMask", Int) = 2
+    }
+
+    HLSLINCLUDE
+    #pragma target 4.5
+    #pragma only_renderers d3d11 playstation xboxone xboxseries vulkan metal switch
+    #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
+    #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl"
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Builtin/BuiltinData.hlsl"
```
FrancescoC-unity, line 14: Do we need this and the following 2 includes? (My guess is not? In that case we can remove and make this compile a bit faster)

FrancescoC-unity (approved):
Small comments, but generally lgtm

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
kecho: The documentation says that DLSS can take an R8_Unorm, however DLSS crashes hard (on the CPU) when we pass anything that doesnt have an RGBA swizzle.
Im prepping something to ask nvidia why is this the case. For now this will have to do.

FrancescoC-unity: Ouch, yeah whatever works for now

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
kecho: Done

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DLSSBiasColorMask.shader
kecho: Done

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
sebastienlagarde: you should add a comment in the code about it

sebastienlagarde (approved):

sebastienlagarde:
be sure you run Yamato

iM0ve (approved):
This was difficult for me to reproduce and the effect/improvement is very subtle and likely setup dependant. I noticed a small improvement as mentioned in the description

**Whats tested:**
* Modified the template to get the visible artifact
* Compared the issue before vs after the PR
* For regression testing in editor mode switched to DX11, DX12 and Vulkan
* Tried both Software and Hardware modes

No issues found

**Demo**

Before

https://user-images.githubusercontent.com/36502659/127343856-15cb8ec5-412d-42d9-892d-c6a880f4c251.mp4

After


https://user-images.githubusercontent.com/36502659/127343910-425d4b4b-ac40-4f31-ba91-6a4466b22a8b.mp4

PR #5085, opened 2021-07-07
Fixed the RTAO debug view being broken.

anisunity:
No proper debug texture was returned in the case of RTAO and the debug view was displaying garbage. We had a test, but it was updated with a wrong screenshot.

Testing status
Tests pass locally and now the screenshot for debug RTAO makes sense


remi-chapelain (approved):
Nice catch. ✔️

FrancescoC-unity (approved):

anisunity:
All green after launching a thousand times

sebastienlagarde (approved):

PR #5086, opened 2021-07-07
Added an SG node to get the Main light in HDRP and URP [22.1]

adrien-de-tocqueville:
### Purpose of this PR

Added a ShaderGraph node to get the sun light direction for URP and HDRP.

---
### Testing status
Tested the node in a shadergraph with a HDRP and URP target.
Tested on the preview material.
Verified the node preview output.
Tested with no directional light (gives 0 vector)

Verified the node follows these guidelines: https://confluence.unity3d.com/pages/viewpage.action?spaceKey=SHAD&title=Node+Style+Guide

Example in HDRP:
image

### DIFF
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/IPhonePlayer/Metal/None/InputNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/IPhonePlayer/Metal/None/InputNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:cd68f56c4faccbee8526c4c4414e31b37a07a140d533eb8e84381a91fc261410
-size 552912
+oid sha256:376458b49f8380dacb2f3cca95f219a1b4f549572befb44de084810e88d9ff22
+size 570383
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/InputNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/InputNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:cd68f56c4faccbee8526c4c4414e31b37a07a140d533eb8e84381a91fc261410
-size 552912
+oid sha256:b706a9f3578fd8bb0acd73a425a3b643f800fedc6e77c047bff3b08a0b7b2eec
+size 562711


com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Nodes/HDSunLightDirectionNode.cs
```diff
@@ -0,0 +1,62 @@
...
+            registry.ProvideFunction(GetFunctionName(), s =>
+            {
+                s.AppendLine("$precision3 {0}()", GetFunctionName());
+                using (s.BlockScope())
+                {
+                    s.AppendLine("#if defined(SHADERPASS) && (SHADERPASS != SHADERPASS_LIGHT_TRANSPORT)");
+                    s.AppendLine("if (_DirectionalLightCount > 0)");
+                    s.AppendLine("  return _DirectionalLightDatas[0].forward;");
+                    s.AppendLine("#endif");
+                    s.AppendLine("return $precision3(0.0f, 0.0f, 0.0f);");
```
alelievr, line 49: I suppose this line is used when rendering the node previews in ShaderGraph, I think it would be better for authoring to return an arbitrary direction (like `normalize(float3(1, 1, 1))`) so we can visualize something directly in SG.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -640,7 +640,12 @@ enum ClusterDepthSource : int
         int m_CurrentShadowSortedSunLightIndex = -1;
         HDAdditionalLightData m_CurrentSunLightAdditionalLightData;
         DirectionalLightData m_CurrentSunLightDirectionalLightData;
-        internal Light GetCurrentSunLight() { return m_CurrentSunLight; }
+
+        /// <summary>
+        /// Main directional Light for the HD Render Pipeline.
+        /// </summary>
+        /// <returns>The main directional Light.</returns>
+        public Light GetMainDirectionalLight() => m_CurrentSunLight;
```
alelievr, line 648: Nice :) I think it would also make sense to have a similar function in shaders like URP have to get the main light: `GetMainLight()`.

Source: https://blog.unity.com/technology/custom-lighting-in-shader-graph-expanding-your-graphs-in-2019

adrien-de-tocqueville: if urp has this function, it would make more sense to put the node in core RP and have it work for both RP
I'll try to do that

alelievr (approved):

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderGraphFunctions.hlsl
```diff
@@ -46,4 +59,10 @@ float3 shadergraph_HDBakedGI(float3 positionWS, float3 normalWS, float2 uvStatic
 #define SHADERGRAPH_BAKED_GI(positionWS, normalWS, uvStaticLightmap, uvDynamicLightmap, applyScaling) shadergraph_HDBakedGI(positionWS, normalWS, uvStaticLightmap, uvDynamicLightmap, applyScaling)
 
 
+#ifdef SHADERGRAPH_GET_SUN_LIGHT
+#undef SHADERGRAPH_GET_SUN_LIGHT
+#endif
+#define SHADERGRAPH_GET_SUN_LIGHT shadergraph_HDGetSunLight
```
sebastienlagarde, line 65: Guess it should be name GetMainDirectionLight

sebastienlagarde, line 65: we should name it, GetMainLight

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderGraphFunctions.hlsl
```diff
@@ -27,6 +27,19 @@ float3 shadergraph_HDBakedGI(float3 positionWS, float3 normalWS, float2 uvStatic
...
 }
 
+void shadergraph_HDGetSunLight(out float3 direction, out float3 color)
+{
+    direction = 0.0f;
+    color = 0.0f;
+#if defined(SHADERPASS) && (SHADERPASS != SHADERPASS_LIGHT_TRANSPORT)
+    if (_DirectionalLightCount > 0)
+    {
+        direction = -_DirectionalLightDatas[0].forward;
```
sebastienlagarde, line 37: I am not sure this work currently. Is the main line always the slot 0 when we have multiple directional light? Not sure about that.

adrien-de-tocqueville: indeed, when the sun light has shadows, it may not always be the first one, so i used _DirectionalShadowIndex in this case

simon-engelbrecht-soerensen (approved):
Working well on URP lit/unlit. This will be a great addition for unlit!

phi-lira (approved):
LGTM. I think this solves some issues we have as well with users trying to do this with custom node.

I wonder about shadows. We see users do custom node for main light shadows as well. In URP this would be easy to support either as separate node or part of same node.

adrien-de-tocqueville:
> I wonder about shadows. We see users do custom node for main light shadows as well. In URP this would be easy to support either as separate node or part of same node.

Sadly this is not as easy for HDRP, so this should probably be URP only

alex-vazquez (approved):
Looks good to me, tested on a XPipeline Project

com.unity.shadergraph/Editor/Data/Nodes/Input/Lighting/GetMainLightNode.cs
```diff
@@ -0,0 +1,40 @@
...
+using UnityEngine;
+
+namespace UnityEditor.ShaderGraph
+{
+    [Title("Input", "Lighting", "Main Light")]
+    class GetMainLightNode : CodeFunctionNode
+    {
+        public GetMainLightNode()
+        {
+            name = "Main Light";
```
bencloward, line 11: I recommend adding a line like the following at line 12:
`synonyms = new string[] { "sun", "direction" };`
These synonyms are used by the Create Node menu's search filter and can help users find your node even if they don't know the exact name.  Use whatever synonyms you think people might type in the search field when looking for your node.

adrien-de-tocqueville: Oh that's nice ! i added the synonym

Verasl (commented):
Not sure that there was a thorough planning for this feature, in URP this is a pointless node as a user would still have to make a custom main light node to get the colour and shadows. This means a broken incomplete workflow which I'm not sure we want to go down.
If HDRP cannot provide this information then I ask why bother with this node if it only supports 1/3 for the data a user would want in URP and not a great experience on HDRP side?

Furthermore with the coming Forward+ in URP does it make sense to have this separation in shader graph, as the concept of a main light is a URP only thing and something made as a fast path for simple Forward games that only required a single directional light?

adrien-de-tocqueville:
@Verasl There are certainly uses cases that would need colour and shadow information for URP, but these wouldn't be cross pipeline workflows.
On the other hand there are use cases that only require the direction of the light, so this node allows them to be done in a cross pipeline way.

Regarding the main light concept, it's something we have in HDRP regardless of the shading mode. And for URP the node is a wrapper around GetMainLight already available in shaders so if there is a problem it should be handled there

sebastienlagarde:
> @Verasl There are certainly uses cases that would need colour and shadow information for URP, but these wouldn't be cross pipeline workflows.
> On the other hand there are use cases that only require the direction of the light, so this node allows them to be done in a cross pipeline way.
> 
> Regarding the main light concept, it's something we have in HDRP regardless of the shading mode. And for URP the node is a wrapper around GetMainLight already available in shaders so if there is a problem it should be handled there

As a note: we have indeed get request for a cross pipeline way of getting main light direction. 
Users use it to do effect for the sun light, like darkening grass based on sun orientation or doing self shadow for POM for sun. In automotive they also have other use case.

Indeed for URP itself users will prefer to access other light properties but for HDRP due to pre-exposure concept that can quicky create a wrong visual this is not recommended.

sebastienlagarde (approved):

PR #5088, opened 2021-07-07
[HDRP] Fix render object after taa jittering

alelievr:
### Purpose of this PR
Fixed TAA jittering when rendering objects just after the TAA pass (in the after post process injection point).

---
### Testing status
Check that my Render video without TAA example works again: https://github.com/alelievr/HDRP-Custom-Passes#render-video-without-taa

With TAA:

https://user-images.githubusercontent.com/6877923/124771557-84dbe480-df3b-11eb-92d2-7ca07f2ea47c.mp4

Without TAA:

https://user-images.githubusercontent.com/6877923/124771567-85747b00-df3b-11eb-957a-5b17c1fa9f7a.mp4

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Frender-object-after-taa-jittering/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7568328/job


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -482,6 +482,7 @@ TextureHandle GetPostprocessUpsampledOutputHandle(RenderGraph renderGraph, strin
                     if (hdCamera.antialiasing == HDAdditionalCameraData.AntialiasingMode.TemporalAntialiasing)
                     {
                         source = DoTemporalAntialiasing(renderGraph, hdCamera, depthBuffer, motionVectors, depthBufferMipChain, source, postDoF: false, "TAA Destination");
+                        RestoreNonjitteredMatrices(renderGraph, hdCamera);
```
FrancescoC-unity, line 485: This looks good, **however** if physically based dof is enabled we have an additional TAA pass which I wonder if it means we should move the "Post TAA" injection point

kecho (approved):
Good fix!

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
alelievr: The post TAA point is called before post-process, so I don't think it's a good idea to put it after dof. In any case, I think it will break a lot of stuff if we do that.

About the additional TAA pass, is it done on the whole screen or another buffer?

FrancescoC-unity: > The post TAA point is called before post-process, so I don't think it's a good idea to put it after dof. In any case, I think it will break a lot of stuff if we do that.

Before post process but after TAA I assume, no? What I mean is that by doing so DoF runs on the full final color buffer after, so it will smudge what the user didn't want to smudge.

FrancescoC-unity: I guess can be a matter of documentation, but is something we should address _somehow_

JulienIgnace-Unity (approved):

FrancescoC-unity (approved):

sebastienlagarde (approved):

TomasKiniulis (approved):
Took a quick look with multiple cameras, combined with Depth Of Field, volumetric fog. Looks good, don't see any jittering compared with master either.

https://user-images.githubusercontent.com/50582134/125442587-e07d538f-2da4-455c-b18f-9be73427e4e6.mp4

PR #5089, opened 2021-07-07
[HDRP] Fix white flash with SSR when resetting camera history

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1335263/

SSR was flashing because the values of GetPreviousExposureMultiplier() and GetInversePreviousExposureMultiplier() were not correct for a frame after a history reset. 

The reason is that the first frame was rendered with a neutral EV100 exposure multiplier, but on the first frame (and only on the first frame) when we apply the computed exposure with an extra pass that overwrites teh previous exposure texture.

---
### Testing status
Flashing issue does not repro anymore when testing the project attached in the bug ticket.
Notes for QA: 
- The test project can be updated to latest HDRP/Trunk, but you will get errors in one Editor package tthat you have to remove.
- If you hit play, after teh timeline ends, you can find a new video recording in the Cars_Cinemachine\Recordings. You can check there if there is a flashing.
- The flashing can generally reproduced if you have a surface with SSR, dynamic exposure, and toggle TAA off and on again.


FrancescoC-unity (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -377,8 +377,10 @@ void BeginPostProcessFrame(CommandBuffer cmd, HDCamera camera, HDRenderPipeline
                     }
                 }
 
+                // If the history was reset in the previous frame, then the history buffers were actually rendered with a neutral EV100 exposure multiplier
+                var prevExposureTexture = camera.didResetPostProcessingHistoryInLastFrame ? m_EmptyExposureTexture : GetPreviousExposureTexture(camera);
```
FrancescoC-unity, line 381: Actually I changed my mind, I think is cleaner and simpler if we do the branch logic  inside GetPreviousExposureTexture itself, in case other user code grab this previous texture

pmavridis: Indeed!

JulienIgnace-Unity (approved):

iM0ve (approved):
**Whats tested:**
* Also checked on the repro project to better understand
* Created my own scenario in Amalienborg with more reflective materials for better investigation and real world case

**DEMO**
**Before PR**
There was a prominent flash on camera change

https://user-images.githubusercontent.com/36502659/124940798-f470d300-e012-11eb-89ec-4f7dab1a292e.mp4



**After PR:**

https://user-images.githubusercontent.com/36502659/124940701-de631280-e012-11eb-92bb-d6a79cb36721.mp4

**Amalienborg after PR**

https://user-images.githubusercontent.com/36502659/124940968-18341900-e013-11eb-9ca0-6d38489221e2.mp4

**Other notes:**
* Found a PR non-related concern - after camera change 1st frame does not contain geometry in SSR. Same on master. Will investigate if its intended

sebastienlagarde (approved):

PR #5090, opened 2021-07-07
[HDRP] Add mode for custom pass volume to execute on one camera only

alelievr:
### Purpose of this PR
Improve the workflow of custom pass volumes when you need to apply the effect on only one camera.

There is a new Mode named "Camera" that allows you to select on which camera the custom pass volume will be executed:
image

In term of C# API, there is a new field to set the target camera:
```CSharp
CustomPassVolume.targetCamera
```

setting the target camera in code will automatically switch the volume mode to Camera if the camera is not null.

---
### Testing status
- Tested that the new mode executes only with the camera set in the inspector.
- Tested that the volume mask doesn't affect how the volume is executed in camera mode.
- Tested that If the target camera field is set to null, the volume is not executed.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffeature%252Fcustom-pass-camera-only/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7638261/job


com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass-Volume-Workflow.md
```diff
@@ -20,6 +20,8 @@ The image below visualises the outer bounds of the fade radius as a wireframe bo
 
 ![A Custom Pass Volume visualised using a Box Collider, with its fade radius visualised as a wireframe box.](images/CustomPassVolumeBox_Collider.png)
 
+If the **Mode** of your Custom Pass Volume is set to **Camera**, the volume will only be executed for the target camera in parameter. When the target camera is set, the volume ignores the volume layer mask of the camera so that it's always executed for this camera.
```
Vic-Cooper, line 23: Suggested change: 

When a Custom Pass Volume's **Mode** is set to **Camera**, HDRP only executes the custom pass volume in its **Target Camera**. To make sure HDRP always executes the custom pass volume, it ignores the Camera's volume layer mask when a **Target Camera** is set.

Vic-Cooper (changes requested):
Suggested one change.

pmavridis (approved):
That's an awesome feature! Thanks for introducing this. Using this, the compositpr setup code can be simplified a lot.

For consistency, I'm wondering if we can introduce the same feature for the rest of the volume system (not only the custom pass).

Vic-Cooper (approved):
Looks good!

alelievr:
> For consistency, I'm wondering if we can introduce the same feature for the rest of the volume system (not only the custom pass).


That's a good idea :) It's probably possible to do something similar, I'll add this to the list of feature improvements

TomasKiniulis (changes requested):
Love the idea of this! Testing it out there seems to be an issue switching from local/global modes to targetCamera through the API. Custom Pass Volume is correctly updated only when the object is selected. 

https://user-images.githubusercontent.com/50582134/125053580-f76cd180-e0ad-11eb-85ea-23d8bcadfcb6.mp4

TomasKiniulis (approved):
Checked switching through Global and Local volume to target camera, selected and deselected custom volumes are updated correctly now

sebastienlagarde (approved):

PR #5092, opened 2021-07-07
[2021.2][ShaderGraph] Fix wrapping and scrolling due to VFX graph target warning

jessebarker:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1343124/

#3467 introduced a warning about a change in the way VFXgraph would be configured by the user.  This warning tickled an interaction between the Shader Graph graph inspector's scroll view, and the default wrap mode of the actual message in the UI.  This PR does the following:

- Default the GraphInpsector to vertical scrolling only (for graph settings inspector view).
- Enable horizontal + vertical scrolling for node settings inspector view.
- Switch between those scroll view modes as appropriate.
- Enable wrapping on the VFXgraph target warning message.

As this the VFXgraph refactor was done during 2021.2, this should not need a backport.

Here is what the original situation looked like (N.b., there are no +/- controls on the target list and no dropdown indicators on the Precision and Material controls as they are well off the right side of the view and would need to be scrolled to.  Also, the warning message is not fully visible.):
graph-inspector-nowrap

And here is what it looks like with the fix (N.b., the previously missing controls are there, and the message is nicely wrapped and fully visible):
graph-inspector-wrap

And one of the node settings to show that it still has a horizontal scroll:
Screen Shot 2021-07-07 at 1 06 09 PM

---
### Testing status
Manual testing according to the repro case in the bug.
Additional manual testing to exercise the scroll view toggling (i.e., switching between graph settings and node settings), as well as resizing the graph inspector to engage/disengage the scroll bars.  Confirmed that these behave as expected.

---
### Comments to reviewers


Nightmask3 (approved):
LGTM 👍

Nightmask3 (changes requested):
Oh one thing, needs a changelog

jessebarker:
> Oh one thing, needs a changelog

Thanks!  I wanted to get the actual changes pushed up first.  The changelog is there now.

Nightmask3 (approved):
changelog added, approving!

ghost (approved):
Thank you for the images

PR #5099, opened 2021-07-08
VFX Multi-selection and automatic attach on selection

julienamsellem:
### Purpose of this PR
This PR addresses two different purposes that are connected.
1. When multiple VFX are selected the VFX overlay should apply to all selected
https://fogbugz.unity3d.com/f/cases/1345753
2. When a VFX asset is being edited, automatically attach it to the selection (if compatible).
https://jira.unity3d.com/browse/GVFXG-134    

_Design docs_    
https://docs.google.com/document/d/1VjEjkRHN-GHeTx5bDW9F66GzGzPV9QQoAyh9QGsB1f4/edit?usp=drivesdk
https://docs.google.com/document/d/1jNWIv1FwW1SwU-qgblXfb4Oxw9CFEuTemahm0vku48E/edit?usp=drivesdk

---    
### Preview    
multi-selection
auto-attach

---
### Testing status
Tech artists have tested as real users so probably not every corner cases.

---
### Comments to reviewers

### DIFF
--- a/TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXPickerTests.cs
+++ b/TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXPickerTests.cs
@@ -2,8 +2,6 @@ using System.Reflection;
 
 using NUnit.Framework;
 
-using UnityEditor.Search;
-
 namespace UnityEditor.VFX.Test
 {
     [TestFixture]
@@ -18,7 +16,8 @@ namespace UnityEditor.VFX.Test
             var viewStateInfo = quickSearchType?.GetProperty("viewState", BindingFlags.Instance | BindingFlags.NonPublic);
             Assert.NotNull(viewStateInfo);
 
-            var flagsInfo = typeof(SearchViewState).GetField("flags", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
+            var searchViewStateType = typeof(Search.SearchService).Assembly.GetType("UnityEditor.Search.SearchViewState");
+            var flagsInfo = searchViewStateType.GetField("flags", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
             Assert.NotNull(flagsInfo);
         }
     }
--- a/com.unity.visualeffectgraph/Editor/GraphView/Views/VFXPicker.cs
+++ b/com.unity.visualeffectgraph/Editor/GraphView/Views/VFXPicker.cs
@@ -28,7 +28,8 @@ static class VFXPicker
         // Until the "viewState" API is made public (should be in 2022.1) we use reflection to remove the inspector button
         var quickSearchType = typeof(SearchService).Assembly.GetType("UnityEditor.Search.QuickSearch");
         var viewStateInfo = quickSearchType?.GetProperty("viewState", BindingFlags.Instance | BindingFlags.NonPublic);
-        if (viewStateInfo?.GetValue(view) is SearchViewState state)
+        var state = viewStateInfo?.GetValue(view);
+        if (state != null)
         {
             var flagsInfo = state.GetType().GetField("flags", BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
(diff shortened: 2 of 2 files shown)


VitaSkruibyte:
Logged some issues that can be looked into already, here's the work in progress [test doc](https://docs.google.com/document/d/1Cw1BICfGJfNiDXQawv9VMZwYr6qXW1GD7sr99VtLTHQ/edit?usp=sharing).

com.unity.visualeffectgraph/Editor/Utilities/EventTester/VFXEventTesterWindow.cs
```diff
@@ -262,31 +274,34 @@ static void WindowGUI()
...
 
-            // then send event with attributes
-            if (name == "OnPlay")
-                m_Effect.Play(attrib);
-            else if (name == "OnStop")
-                m_Effect.Stop(attrib);
-            else
-                m_Effect.SendEvent(name, attrib);
+                // then send event with attributes
+                if (name == "OnPlay")
```
PaulDemeulenaere, line 298: You didn't change this code but these is a global static value for this string : https://docs.unity3d.com/ScriptReference/VFX.VisualEffectAsset.PlayEventName.html
You can replace with :
```
name == VisualEffectAsset.PlayEventName
```

PaulDemeulenaere (approved):
I don't have any important behavior change to recommend.
I have only coding rules issues (especially the `this.`)

The issue reported by @VitaSkruibyte in [test doc](https://docs.google.com/document/d/1Cw1BICfGJfNiDXQawv9VMZwYr6qXW1GD7sr99VtLTHQ/edit) are probably a simply missing null check somewhere, so I can preapprove 🟢

VitaSkruibyte (approved):
Looks neat 💯! All the raised issues were fixed, couldn't find any other problems after additional pass. Thanks for all the work done on this! 🟢

com.unity.visualeffectgraph/Editor/GraphView/VFXComponentBoard.cs
```diff
@@ -747,28 +708,25 @@ static bool IsDefaultEvent(string evt)
...
-            if (context is VFXBasicEvent)
+            switch (context)
             {
-                if (!IsDefaultEvent(name))
-                    yield return (context as VFXBasicEvent).eventName;
-            }
-            else if (context is VFXSubgraphContext)
-            {
-                foreach (var subContext in (context as VFXSubgraphContext).subChildren.OfType<VFXContext>())
+                case VFXBasicEvent basicEvent when !IsDefaultEvent(name):
```
PaulDemeulenaere, line 720: Nice, I wasn't aware of this new condition : https://docs.microsoft.com/fr-fr/dotnet/csharp/language-reference/statements/selection-statements

PaulDemeulenaere (approved):
LGTM! Thanks!

peeweek:
As stated in the feedback document, currently the object field displays all effects in the scene but can only attach to the instances that are compatible with the currently opened graph, which is confusing as this field's purpose is to find instances from the scene that can be attached to:
- There is no way to figure out which effect is compatible with current graph
- Selecting an incompatible instance in the scene has no feedback

attach-objectpicker-shows-all

The issue should be addressed or the field removed, if can't be fixed. As the occurrence of this usability issue is extremely high. (It is extremely unlikely that a scene will only use effects from a single graph).

MarieGuff (changes requested):
Red flag on TA side: in the attach search, the incompatible VFX are displayed the same way as compatible ones. The incompatible ones should be greyed or hidden (using the search filter as Thomas suggested is a good idea). 
Otherwise, at least an error message (feedback) when selecting a incompatible VFX is necessary if nothing else can be done for now!

```
[Thomas finding]
❌[UX] Objectpicker shows all VisualEffect game objects and not only the ones with compatible attachments. 
Selecting an incompatible game object does not have any feedback. This makes the object picker rather useless in 
scenes with hundreds of game objects using dozens of different asset graphs (commonplace in game production)

Note : Using ref:/Assets/Path/To/Asset.vfx in the object picker enables filtering the correct ones.

```

julienamsellem:
Yamato `General Checks on trunk` is green but `General Checks on 2021.2` is red for unrelated reason

kroatoa (approved):

PR #5103, opened 2021-07-08
Improve node search result sorting

julienamsellem:
### Purpose of this PR
https://jira.unity3d.com/browse/GVFXG-142

When searching for a node to add, the search results were not sorted correctly.
For example an exact match could be listed after a partial match.

Now the sorting is done according the following criteria:
1. node name is equal to the search string
2. node name contains exactly the search string
3. node name contains all search words but not in the same order

In addition to these criteria, in case of equality shorted node names are listed before.

_Design doc_    
https://docs.google.com/document/d/199Rw34rkn3TuNDiDxEuVcXg7Qg-EsAKI5GBDU4C5aVU/edit?ts=60e6b80c#

goRpTzFNMw

---
### Testing status
I tested specifically the cases mentioned in the linked google doc

---
### Comments to reviewers

### DIFF
--- a/com.unity.visualeffectgraph/Editor/Models/Blocks/Implementations/Orientation/Orient.cs
+++ b/com.unity.visualeffectgraph/Editor/Models/Blocks/Implementations/Orientation/Orient.cs
@@ -81,7 +81,7 @@ namespace UnityEditor.VFX.Block
         private bool canTestStrips => flattenedParent as VFXAbstractParticleOutput; // Cannot check strip in subblock context or not child of a context
         private bool hasStrips => ((VFXAbstractParticleOutput)flattenedParent).HasStrips(); // direct cast as canTestStrips is supposed to have been called priorly
 
-        public override string name { get { return "Orient : " + ObjectNames.NicifyVariableName(mode.ToString()); } }
+        public override string name { get { return "Orient: " + ObjectNames.NicifyVariableName(mode.ToString()); } }
 
         public override VFXContextType compatibleContexts { get { return VFXContextType.Output; } }
         public override VFXDataType compatibleData { get { return VFXDataType.Particle; } }
--- a/com.unity.visualeffectgraph/Editor/Models/Blocks/Implementations/Position/PositionSequential.cs
+++ b/com.unity.visualeffectgraph/Editor/Models/Blocks/Implementations/Position/PositionSequential.cs
@@ -68,7 +68,7 @@ namespace UnityEditor.VFX.Block
         [Tooltip("Specifies how the sequence should behave at the end. It can either wrap back to the beginning, clamp, or continue in a mirrored direction.")]
         private VFXOperatorUtility.SequentialAddressingMode mode = VFXOperatorUtility.SequentialAddressingMode.Clamp;
 
-        public override string name { get { return string.Format("{0} Position (Sequential : {1})", VFXBlockUtility.GetNameString(compositionPosition), shape); } }
+        public override string name { get { return string.Format("{0} Position (Sequential: {1})", VFXBlockUtility.GetNameString(compositionPosition), shape); } }
         public override VFXContextType compatibleContexts { get { return VFXContextType.InitAndUpdateAndOutput; } }
         public override VFXDataType compatibleData { get { return VFXDataType.Particle; } }
 
--- a/com.unity.visualeffectgraph/Editor/Models/Blocks/Implementations/Update/AngularEulerIntegration.cs
+++ b/com.unity.visualeffectgraph/Editor/Models/Blocks/Implementations/Update/AngularEulerIntegration.cs
@@ -7,7 +7,7 @@ namespace UnityEditor.VFX.Block
     [VFXInfo(category = "Implicit")]
     class AngularEulerIntegration : VFXBlock
     {
-        public override string name { get { return "Integration : Update Rotation"; } }
+        public override string name { get { return "Integration: Update Rotation"; } }
         public override VFXContextType compatibleContexts { get { return VFXContextType.Update; } }
         public override VFXDataType compatibleData { get { return VFXDataType.Particle; } }
         public override IEnumerable<VFXAttributeInfo> attributes
--- a/com.unity.visualeffectgraph/Editor/Models/Blocks/Implementations/Update/EulerIntegration.cs
+++ b/com.unity.visualeffectgraph/Editor/Models/Blocks/Implementations/Update/EulerIntegration.cs
@@ -7,7 +7,7 @@ namespace UnityEditor.VFX.Block
     [VFXInfo(category = "Implicit")]
     class EulerIntegration : VFXBlock
     {
-        public override string name { get { return "Integration : Update Position"; } }
+        public override string name { get { return "Integration: Update Position"; } }
         public override VFXContextType compatibleContexts { get { return VFXContextType.Update; } }
         public override VFXDataType compatibleData { get { return VFXDataType.Particle; } }
         public override IEnumerable<VFXAttributeInfo> attributes


com.unity.visualeffectgraph/Editor/FilterPopup/VFXFilterWindow.cs
```diff
@@ -262,8 +256,7 @@ private void CreateComponentTree()
                         break;
 
                     // Try to find a child of the same name as we had before
-                    List<Element> children = GetChildren(activeTree, match);
-                    Element childMatch = children.FirstOrDefault(c => c.name == m_Stack[level].name);
+                    Element childMatch = GetChildren(activeTree, match).FirstOrDefault(c => c.name == m_Stack[level].name);
```
julienamsellem, line 259: Avoid allocating a complete list to grab a single element

com.unity.visualeffectgraph/Editor/FilterPopup/VFXFilterWindow.cs
```diff
@@ -373,7 +366,7 @@ private void HandleKeyboard()
                     if (evt.keyCode == KeyCode.DownArrow)
                     {
                         activeParent.selectedIndex++;
-                        activeParent.selectedIndex = Mathf.Min(activeParent.selectedIndex, GetChildren(activeTree, activeParent).Count - 1);
+                        activeParent.selectedIndex = Mathf.Min(activeParent.selectedIndex, GetChildren(activeTree, activeParent).Count() - 1);
```
julienamsellem, line 369: Again avoid allocation to only get the count

com.unity.visualeffectgraph/Editor/FilterPopup/VFXFilterWindow.cs
```diff
@@ -589,13 +563,13 @@ private void ListGUI(Element[] tree, GroupElement parent)
 
             EditorGUIUtility.SetIconSize(new Vector2(16, 16));
 
-            List<Element> children = GetChildren(tree, parent);
+            Element[] children = GetChildren(tree, parent).ToArray();
```
julienamsellem, line 566: ToArray is known to be slightly faster than ToList. Here we don't need a list ...

VitaSkruibyte:
Hi @julienamsellem, PR looks very solid! Nevertheless logged some concerns in the [test doc](https://docs.google.com/document/d/1BaMCjOfSbw2g4599tTxfvwZMTKtew_kC04JBlFt5iN4/edit?usp=sharing).

com.unity.visualeffectgraph/Editor/FilterPopup/VFXBlockProvider.cs
```diff
@@ -151,26 +156,18 @@ protected override string title
...
+                .Select(t => (Descriptor) new NewBlockDescriptor(t))
+                .Concat(SubGraphCache.GetItems(typeof(VisualEffectSubgraphBlock))
+                    .Where(t =>
+                        (((SubGraphCache.AdditionalBlockInfo)t.additionalInfos).compatibleType &
+                            m_ContextController.model.contextType) != 0 &&
+                        (((SubGraphCache.AdditionalBlockInfo)t.additionalInfos).compatibleData &
+                            m_ContextController.model.ownedType) != 0)
+                    .Select(t => (Descriptor) new SubgraphBlockDescriptor(t)))
+                .OrderBy(x => x.category)
+                .ThenBy(x => x.name);
```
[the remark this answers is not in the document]
PaulDemeulenaere: I also prefer OrderBy syntax than Sort approach which require a list.

com.unity.visualeffectgraph/Editor/FilterPopup/VFXFilterWindow.cs
PaulDemeulenaere: By the way, we aren't respecting all our coding rules. Here, it should have been `var` : https://ono.unity3d.com/unity-extra/unity-meta/files/@/ReferenceSource/CSharp/Assets/CSharpReference.cs#L222

com.unity.visualeffectgraph/Editor/FilterPopup/VFXFilterWindow.cs
```diff
@@ -660,36 +634,20 @@ private void ListGUI(Element[] tree, GroupElement parent)
...
-                    continue;
-
-                children.Add(e);
-            }
+            var childrenLevel = parent.level + 1;
+            return tree.SkipWhile(x => x != parent)
+                .Skip(1)
+                .TakeWhile(x => x.level >= childrenLevel)
+                .Where(x => x.level == childrenLevel || (x.level > childrenLevel && hasSearch));
+        }
```
PaulDemeulenaere, line 644: I prefer the Linq approach than the previous difficult to read algorithm, but you have to be careful modifying this : Linq doesn't perform as fast as general .NET application on Unity. However, in your case, it seems correct.

com.unity.visualeffectgraph/Editor/FilterPopup/VFXBlockProvider.cs
```diff
@@ -44,37 +45,42 @@ public void CreateComponentTree(List<VFXFilterWindow.Element> tree)
             tree.Add(new VFXFilterWindow.GroupElement(0, title));
             var descriptors = GetDescriptors();
 
+            var depth = 1;
             string prevCategory = "";
-            int depth = 1;
+            string[] prevSplit = new string[0];
+
+            List<T> noCategory = new List<T>();
```
PaulDemeulenaere, line 52: You have to be consistent with convention of this current file, so, if it uses "explicit" declaration, it's fine but it should also be `var` [here](https://ono.unity3d.com/unity-extra/unity-meta/files/@/ReferenceSource/CSharp/Assets/CSharpReference.cs#L222)

PaulDemeulenaere (commented):
It looks good to me. I didn't check the outgoing new ordering, but if it's validated by QA & Artist, it will be sufficient.

If you are curious about the performance implications of your changes, you want to easily compare before/after and you want a way to (later) track potential performance regression, feel free to add an editor performance test :
You can take inspiration from [Load_VFXLibrary](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.testing.visualeffectgraph/PerformanceTests/Editor/VFXPerformanceEditorTests.cs#L103), maybe we could add a timing for the call `VFXFilterWindow.CreateComponentTree`, poke me if you think it's relevant or ignore if it doesn't deserve to be measure (the current automation doesn't provide any aggregated results anyway)

Also, I run [Yamato Extra_VFX](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/vfx%252Ffeature%252FGVFXG-142-improve-node-search) on trunk and 2021.2.
But actually, only the editor test are relevant for this PR (see [here](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/vfx%252Ffeature%252FGVFXG-142-improve-node-search/.yamato%252Fvfx_hdrp-win-dx11.yml%2523VFX_HDRP_Win_DX11_editmode_PerformanceTests_mono_Linear_2021.2/7632150/job))

PaulDemeulenaere (approved):
There is only the expected fail among editor test :
```
[07:30:27.139 Information] [453/1105] TestFinish
	name: UnityEditor.VFX.Test.VFXSpawnerTest.Sanitize_VFXSpawnerCustomCallback_Namespace
	result: FAILED
	message:   Expected: not null
  But was:  null
```
It's not related to your PR 🟢

com.unity.visualeffectgraph/Editor/FilterPopup/VFXFilterWindow.cs
julienamsellem: Thanks for the link I'll refer to it in the future when in doubt :)

VitaSkruibyte:
Looking good, all the raised concerns were fixed couldn't find any other outstanding issues. [Test doc](https://docs.google.com/document/d/1BaMCjOfSbw2g4599tTxfvwZMTKtew_kC04JBlFt5iN4/edit?usp=sharing) for visibility.

PR #5108, opened 2021-07-09
Better feedback for when cached shadows are not fitting in atlas

FrancescoC-unity:
Before this PR a cached shadow that didn't fit the atlas wouldn't render without much immediate feedback given, the user needed to go to debug settings, print log of lights that fitted and cross reference. 

This was of course a bit cumbersome. 

Now there is additional feedback on the light and an option to set the resolution to the maximum that would fit the atlas

https://user-images.githubusercontent.com/43168857/125065505-313fd700-e0b2-11eb-8b14-d8ea8ef0ed7d.mp4

If it cannot fit, the option to fit doesn't appear and the warning displays a different message 
image

**What did I test**: I tested with punctual lights that the error appears when atlas is full, that it goes away when resolution is changed (both atlas and on the light) and that the button sets to a resolution that will lead to fitting into the atlas. 
In all the above tests I verified that lights were visible (or not) depending matching the warning status. 

The test was conducted on a toy project spamming new lights (both spot and point)


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -1046,20 +1049,78 @@ static void DrawShadowMapContent(SerializedHDLight serialized, Editor owner)
...
+
+                            if (showFitButton)
+                            {
+                                if (GUI.Button(rect, "Set resolution to the maximum that fits"))
+                                {
+                                    var scalableSetting = ScalableSettings.ShadowResolution(lightType, hdrp);
+                                    int res = additionalLightData.GetResolutionFromSettings(lightType, hdrp.currentPlatformRenderPipelineSettings.hdShadowInitParams);
+                                    int foundResFit = -1;
+                                    // Round up to multiple of 64
+                                    res = (res / slotSize) * slotSize + ((res % slotSize) == 0 ? 0 : slotSize);
```
RSlysz, line 1087: I wonder if `(int)(Mathf.Ceil(res / (float)slotSize) * slotSize)` is more readable

Or, have a dedicated method for that with an explicit name.

RSlysz (approved):

RSlysz:
Also, you let some TODO. Perhaps you should address them now or fill a task in Jira or something...

FrancescoC-unity:
> Also, you let some TODO. Perhaps you should address them now or fill a task in Jira or something...

Leftover comments :D

sebastienlagarde (approved):

FrancescoC-unity:
> Could we use the Light Explorer to indicate if a light doesn't fit the atlas?

We could yes. Gonna add that. 

> Alternatively, we could maybe have a list of all the lights that don't fit in the Render Debug window, in the Shadow section.

This is already there (see https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@12.0/manual/Shadows-in-HDRP.html Log Cached Shadow Atlas Status).  This PR is just an addition to that to also streamline the "fix" without having to do math. 

> Forcing the user to check each light manually to find out if it fits in the atlas or not is not really viable.

That's not what the workflow should be though. Can expand more in slack.

FrancescoC-unity:
@pierre-unity 


https://user-images.githubusercontent.com/43168857/125500688-70f9c2a3-84ef-4a23-9c7d-59540fb50fa5.mp4

Also I fixed some naming issues in the explorer and some missing things from explorer that I found while looking at this (cached shadow state was not updated and shadow resolution was misnamed)

iM0ve (commented):
**Review: Done**

**What's tested:**
* New warning is present on different types of lights (point, spot, area)
* New Fix button works as expected, increasing the shadow resolution on the selected light, changing the preset to custom.
* On enable shadows are still working
* The Light Explorer window added on 2nd iteration, it really brings a major improvement in terms of tracking/solving missing cached shadows
* Changing cached resolution in HD asset
* Checked that the warnings disappear if switched to HD asset with big shadow atlas

**Issues:**
**1) FIXED** It's great to be able to inspect and select not fitting lights in the Light inspector, however, the sorting currently is not working as expected. 

https://user-images.githubusercontent.com/36502659/125600228-05276e3d-a5ca-4044-95b5-16a2223dacbf.mp4

**2) FIXED** The search no longer works as intended. Its not possible to see whats being typed, but the filter updates. Works fine on master.

https://user-images.githubusercontent.com/36502659/126981603-8e717481-8d7f-4cac-8e1b-7fa5a91288b9.mp4

**3) FIXED** It is extremely laggy when trying to use the Light explorer. Need a different way to perform checks, else the window might be unusable in real production

https://user-images.githubusercontent.com/36502659/126981846-47618e91-1574-47fd-864a-ebfb176f0195.mp4


**Concerns:**
**1) FIXED** Since the functionality was updated with Light Explorer, maybe its worth also changing the text of the Warning, mentioning Light explorer? 

example:
_The shadow for this light doesn't fit the cached shadow atlas and therefore won't be rendered. Please open the Light Explorer and review the Shadows Fit Atlas column for each light. For more information about cached shadow management, consult HDRP Shadow documentation._

image

**2) FIXED** Do we still need this? The new Light Explorer now has the same functionality ?

image

iM0ve (changes requested):
When the above issues are addressed I think it should be good to go

FrancescoC-unity:
> **1)** It's great to be able to inspect and select not fitting lights in the Light inspector, however, the sorting currently is not working as expected.

Fixed in new commit!



> **1)** Since the functionality was updated with Light Explorer, maybe its worth also changing the text of the Warning, mentioning Light explorer?

Done this too! 



> 1. Do we still need this? The new Light Explorer now has the same functionality ?

Indeed no, removed.

iM0ve (changes requested):
Updated the review with issues # 2 and # 3

iM0ve (approved):
I have updated the review and took another look. All issues are now fixed and no new ones found.

PR #5112, opened 2021-07-12
Moved the Spotlight range validation logic into HDAdditionalLightData.OnValidate

fredericv-unity3d:
### Purpose of this PR
Fixes [case 1345264](https://fogbugz.unity3d.com/f/cases/1345264/)

Light range values were validated inside the inspector instead of the `HDAdditionalLightData.OnValidate()` function.
Previous code did not support multi edition appropriately and is only executed when editing with the inspector or gizmo.

I moved as well other fields in similar case:
* Shape Width and Shape Height (Box)
* Shape Radius (Cone, Pyramid)

---
### Testing status
Manual tests:
#### Testing that multi edition works appropriately
1. Created a new scene
2. Delete everything in the hierarchy
3. Add 3 spot lights
4. Set their radius to 0, 1, 2 respectively
5. Select them all
6. Use the `Inner Angle` slider in the inspector ui
7. > The radius value is not updated for any spotlight

#### Testing that the validation logic properly execute
1. Select any spot light
2. Set to the cone shape
3. Try to set the range value below 0
4. > The radius value is 0

1. Select any spot light
2. Set the box shape
3. Try to set the width and height values below `0.1`
4. > The width and height values are exactly 0.1

---


JulienIgnace-Unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
note to self: to redirect to master once hd/bugfix is merge

sebastienlagarde (approved):

JMargevics (approved):
What's tested seems sufficient. From my side I've checked if Undo still works. No issues ✅

erikabar (approved):
Changes are in HDRP package, so approving without testing from URP side

PR #5114, opened 2021-07-12
Reformat remaining URP projects

theopnv:
### Purpose of this PR

In order to add the remaining URP test projects to the Katana QV, let's comply with the code convention and reformat the remaining files.

Note: Only the code itself was reformatted. Some file names don't comply with the convention but an exception will be added to unity/unity.

---


nigeljw-unity:
LGTM

nigeljw-unity (approved):

PR #5115, opened 2021-07-12
[2021.2+] Fix URP Smoke Test to do SRP assignation in both Quality and Graphics Settings

jenniferd-unity:
---
### Purpose of this PR
This PR is needed to fix https://github.cds.internal.unity3d.com/unity/unity/pull/4258 Katana tests that are failing. The goal is to ensure that when switching to an SRP or Built-in we take into account both Graphics and Quality Settings.

---
### Testing status
Tested in Windows Editor to run the test scene (with and without my changes on the other PR).

---
### Comments to reviewers
This is step 1 of 3 for the PR to use the right SRP in our codebase.

### DIFF
--- a/com.unity.render-pipelines.universal/Tests/Runtime/RuntimeTests.cs
+++ b/com.unity.render-pipelines.universal/Tests/Runtime/RuntimeTests.cs
@@ -73,7 +73,7 @@ class RuntimeTests
         if (RenderPipelineManager.currentPipeline == null)
             Assert.Pass("Render Pipeline Asset is Null, test pass by default");
 
-        Assert.AreEqual(RenderPipelineManager.currentPipeline.GetType(), typeof(UniversalRenderPipelineAsset),
+        Assert.AreEqual(RenderPipelineManager.currentPipeline.GetType(), typeof(UniversalRenderPipeline),
             "Pipeline Asset is not Universal RP");
     }
 }


com.unity.render-pipelines.universal/Tests/Runtime/RuntimeTests.cs
```diff
@@ -64,12 +68,12 @@ public IEnumerator PipelineSetsAndRestoreGlobalShaderTagCorrectly()
 
     void AssetCheck()
     {
-        //Assert.IsNotNull(currentAsset, "Render Pipeline Asset is Null");
+        //Assert.IsNotNull(currentAssetGraphics, "Render Pipeline Asset is Null");
         // Temp fix, test passes if project isnt setup for Universal RP
-        if (currentAsset == null)
+        if (RenderPipelineManager.currentPipeline == null)
```
Verasl, line 73: `currentPipeline` returns an instance of the pipeline rather than the asset, either grab the asset file, or down further check against `UniversalRenderPipeline` rather than `UniversalRenderPipelineAsset`

Verasl (changes requested):
new logic changed to checking a pipeline instance against a pipeline asset, this has led to test failing, will need to be changed for the tests to pass.

com.unity.render-pipelines.universal/Tests/Runtime/RuntimeTests.cs
jenniferd-unity: so i am unsure about what this should check then. is it that an URP asset will be selected to be the active pipeline? (if so it means a combination of graphics/quality SRP asset?

Verasl: Oh actually it's not so much an issue with this line, more that `currentAsset` was switched out for `RenderPipelineManager.currentPipeline`, the real issue is actually a few lines lower where it does the `AreEquals` check

jenniferd-unity: oh yep! okay will rerun locally on batch mode and editor mode and compare everything is green!

jenniferd-unity:
Tested locally on 2021.2 and 2022.1 and seemed to work now (issue before is that package cache was not cleaned and local tests were false positives)

erikabar (approved):
Looks like the change is simple enough, so QA is not needed.

com.unity.render-pipelines.universal/Tests/Runtime/RuntimeTests.cs
```diff
@@ -64,12 +68,12 @@ public IEnumerator PipelineSetsAndRestoreGlobalShaderTagCorrectly()
...
             Assert.Pass("Render Pipeline Asset is Null, test pass by default");
```
phi-lira, line 74: Minor, related to above. the message should be updated as this is not testing the pipeline asset anymore.

phi-lira (approved):

PR #5116, opened 2021-07-12
[Hold] EditorWindow for diffusion profiles registration

adrien-de-tocqueville:
### Purpose of this PR

Small tweak to the behavior introduced in PR #4743. Basically it waits until import is finished to show the popup for adding profiles.
That way we don't pause import process in the middle, and choice can be made over which profiles to add with the full list available.

It also implicitely fixes warning that would get printed without showing the popup in case there is no more room in the global settings.
Added an option in the HDRP Global Settings Misc section to disable the window. It is shown by default, but the HDRP_tests project have it disabled

Window looks like that: Clicking on an object pings it in the project window
image

---
### Testing status

see the gif below
pretty much only contains UX change compared to francesco's changes so i didn't test in depth the behavior

diffusion_profiles

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -315,7 +315,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue with reflection probe normalization via APV when no probes are in scene.
 - Fixed Volumetric Clouds not updated when using RenderTexture as input for cloud maps.
 - Fixed custom post process name not displayed correctly in GPU markers.
-- Fixed diffusion profile upgrade window to by delaying the windows to end of the import process
+- Fixed the diffusion profile upgrade window by delaying its appearance until HDRP has finished the import process.
 
 ### Changed
 - Changed Window/Render Pipeline/HD Render Pipeline Wizard to Window/Rendering/HDRP Wizard
--- a/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/HDGlobalSettingsWindow.Skin.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/HDGlobalSettingsWindow.Skin.cs
@@ -77,9 +77,9 @@ namespace UnityEditor.Rendering.HighDefinition
             public static readonly GUIContent fixAssetButtonLabel = EditorGUIUtility.TrTextContent("Fix", "Ensure a HD Global Settings Asset is assigned.");
 
             public static readonly GUIContent probeVolumeSupportContentLabel = EditorGUIUtility.TrTextContent("Probe Volumes (Experimental)", "Set whether Probe volumes are supported by the project. The feature is highly experimental and subject to changes.");
-            public static readonly GUIContent rendererListCulling = EditorGUIUtility.TrTextContent("Dynamic Render Pass Culling", "When enabled, rendering passes are automatically culled based on what is visible on the camera.");
+            public static readonly GUIContent rendererListCulling = EditorGUIUtility.TrTextContent("Dynamic Render Pass Culling", "When enabled, HDRP automatically culls render passes based on what is visible in the Camera.");
             public static readonly GUIContent supportRuntimeDebugDisplayContentLabel = EditorGUIUtility.TrTextContent("Runtime Debug Shaders", "When disabled, all debug display shader variants are removed when you build for the Unity Player. This decreases build time, but prevents the use of Rendering Debugger in Player builds.");
-            public static readonly GUIContent showMissingDiffusionProfilesContentLabel = EditorGUIUtility.TrTextContent("Missing Diffusion Profiles Window", "When enabled, if a material with missing diffusion profiles is imported, a window will be displayed to allow adding the profile to the HDRP Global Settings.");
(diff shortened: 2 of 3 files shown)


merpheus-dev:
If you are still seeking any UI suggestions, for the UX of this, rather than showing plain text description, three things can be done. ( I agree that UI is pretty un-descriptive)

- Show a grid layout of texture previews of the diffusion profiles with names attached to their bottom.
- Plain text UI can stay still, but a little button can be added to the same line to show a preview or inspect the diffusion profile upon clicking it.
- Combination of 1st and 2nd option where it shows a grid layout diffusion profile preview+button to inspect the whole thing.

adrien-de-tocqueville:
@merpheus-dev Thanks for your suggestions :slightly_smiling_face: unfortunately diffusion profile don't have a preview, only a static icon. But still i did your second point for the inspection (see the new screenshot in the description)

merpheus-dev:
Great! That's a clever solution. Thanks!

sebastienlagarde (approved):

FrancescoC-unity (approved):

RemyUnity (changes requested):
Here is what the window looks like when I try to import a .unitypackage with 15 profiles, none beeing in the global settings : 
image

Two things bug me : 
- The phrase saying that if a profile is not set in the global settings then HDRP can't use it is a false statement, as I can use the "Diffusion Profile Override" in volumes.
- If would be nice to point to the override in the error box at the bottom so users can learn how to use them to go over that 15 profiles limit.

adrien-de-tocqueville:
I updated the window to mention the diffusion profile override with a link to the doc
diffusion_profiles

I also noticed that when saving a material, it gets reimported and so the window pops up everytime. To workaround that, i made the checkbox accessible directly from the window to easily disable it, and i made the API public so that asset store assets can explicitely require the window to be openned for some assets, which bypasses the checkbox

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/HDGlobalSettingsWindow.Skin.cs
```diff
@@ -79,6 +79,7 @@ internal class Styles
             public static readonly GUIContent probeVolumeSupportContentLabel = EditorGUIUtility.TrTextContent("Probe Volumes (Experimental)", "Set whether Probe volumes are supported by the project. The feature is highly experimental and subject to changes.");
             public static readonly GUIContent rendererListCulling = EditorGUIUtility.TrTextContent("Dynamic Render Pass Culling", "When enabled, rendering passes are automatically culled based on what is visible on the camera.");
             public static readonly GUIContent supportRuntimeDebugDisplayContentLabel = EditorGUIUtility.TrTextContent("Runtime Debug Shaders", "When disabled, all debug display shader variants are removed when you build for the Unity Player. This decreases build time, but prevents the use of Rendering Debugger in Player builds.");
+            public static readonly GUIContent showMissingDiffusionProfilesContentLabel = EditorGUIUtility.TrTextContent("Missing Diffusion Profiles Window", "When enabled, if a material with missing diffusion profiles is imported, a window will be displayed to allow adding the profile to the HDRP Global Settings.");
```
Vic-Cooper, line 82: Capitalize "Material", also removed passive voice: 

"When enabled, if HDRP imports a Material with missing diffusion profiles, it displays a window where you can add diffusion profiles to the HDRP Global Settings."

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -315,6 +315,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue with reflection probe normalization via APV when no probes are in scene.
 - Fixed Volumetric Clouds not updated when using RenderTexture as input for cloud maps.
 - Fixed custom post process name not displayed correctly in GPU markers.
+- Fixed diffusion profile upgrade window to by delaying the windows to end of the import process
```
Vic-Cooper, line 318: - Fixed the diffusion profile upgrade window by delaying its appearance until HDRP has finished the import process.

Vic-Cooper (changes requested):
Adjusted some of the UI text and the changelog entry.

RemyUnity:
> I updated the window to mention the diffusion profile override with a link to the doc
> 
> I also noticed that when saving a material, it gets reimported and so the window pops up everytime. To workaround that, i made the checkbox accessible directly from the window to easily disable it, and i made the API public so that asset store assets can explicitely require the window to be openned for some assets, which bypasses the checkbox
> 
> image

I like the checkbox to avoid being harassed by the window, but if for some reason I would like to re-enable it, how would I do it ?
Should it be considered UX wise ?

adrien-de-tocqueville:
> I like the checkbox to avoid being harassed by the window, but if for some reason I would like to re-enable it, how would I do it ?
Should it be considered UX wise ?

There is an option in the HDRP global settings. It is explained in the tooltip of the checkbox

RemyUnity:
> > I like the checkbox to avoid being harassed by the window, but if for some reason I would like to re-enable it, how would I do it ?
> > Should it be considered UX wise ?
> 
> There is an option in the HDRP global settings. It is explained in the tooltip of the checkbox

Wouldn't it make sense to have it in the HDRP preferences ? 
image

adrien-de-tocqueville:
I am not sure, I think HDRP preferences not per project. For example the window is disabled in HDRP_Tests but we might want it in other projects.
If I am wrong and the HDRP preferences are per project then I am ok to change, it would make that window a bit more populated, but that would spread the settings even more

RemyUnity:
It's more that HDRP global settings mostly hold settings for the behavior of the renderer itself, while this window is an editor only thing. Eventually in the HDRP wizard ? UX should give advice here.

PR #5119, opened 2021-07-12
[ShaderGraph][21.2] Properly cast values of Float properties in Integer mode

dhsavell:
### Purpose of this PR

This PR fixes the FogBugz case https://fogbugz.unity3d.com/f/cases/1330302/ in which `Float` properties in `Integer` mode could still return fractional values. If a user created a Float property with default of 0.5, then changed its mode to integer (with default value now displaying 0), 0.5 would still be used in the graph.

The property block was fixed in https://github.com/Unity-Technologies/Graphics/pull/4371, and this PR fixes the in-graph value. The user's choice is still preserved, so going from Integer back to Float without changing the value will cause the fractional part to return (see videos below).

---
#### Before Fix

Creating a Float node, giving it a fractional value (i.e. 0.9), then changing its mode to Integer results in the fractional value still being returned. This is illustrated with the Fraction node as suggested in the report. The expected result is a black preview (0).

image

https://user-images.githubusercontent.com/10332426/125334438-5f732000-e319-11eb-82f8-83b5c9df785b.mp4

#### After Fix
image

https://user-images.githubusercontent.com/10332426/125334650-9f3a0780-e319-11eb-8c7e-d46f11cd3ef2.mp4

The user's original entry is preserved even after saving/reloading.

#### Undo/Redo

The user's data is preserved and the previews update accordingly when stepping back and forth through history.

https://user-images.githubusercontent.com/10332426/125335468-9564d400-e31a-11eb-9de8-4331f666d84d.mp4

#### Other Options

The other Float modes, Enum and Slider, still behave as they did before.

image
image
image
image

#### Parity with Material Properties

Both the material properties and Shader Graph previews result in the same defaults (0.9 and -0.9 test cases, both truncate to 0).

https://user-images.githubusercontent.com/10332426/125482819-08ec31d1-540b-44fa-803a-f4fc9bf9317f.mp4

https://user-images.githubusercontent.com/10332426/125483399-ccc270a1-d26e-48f6-81a0-9bd441bd81e1.mp4


jessebarker (approved):
LGTM.  Great illustration of the fix via videos and images.

LandonTownsendUnity:
Casting a float as an int will truncate the value (round towards 0). Can you validate that this is the same thing that happens with the property block? Test 0.9 and -0.9 and make sure that the values you get in the graph are the same values you get as the default when you make a material.

dhsavell:
Tested and the defaults match, updated the PR with a new section using the suggested test cases.

marctem (approved):

LandonTownsendUnity (approved):

PR #5121, opened 2021-07-12
[URP] Enabled copyDepth optimization for XR

thomas-zeng:
---
### Purpose of this PR
- Addressed some issues caught by the automation

---
### Testing status
- Describe what manual/automated tests were performed for this PR

---
### Comments to reviewers
The issue is that:
- The custom OutputTextureFeature draws a full screen mesh to do the blit pass. Due to the render target is a renderTexture, unity yflip the projection matrix.
    - In non-XR case, uv is adjusted by `_ScaleBiasRt` which is set in previous `CopyDepth` pass to flip.
    - In XR case, uv is adjusted by `_ScaleBiasRt` which is set in previous `CopyDepth` pass to *NOT* flip.
    - This causes the yflip issue to only show up in XR
- Propose a fix to explicitly set `_ScaleBiasRt` in the current RenderFeature Pass to avoid any inconsistency. Otherwise, the flip behavior is defined by the pass before this render feature.

### DIFF
--- a/TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/200_Assets/Scripts/OutputTextureFeature.cs
+++ b/TestProjects/UniversalGraphicsTest_Foundation/Assets/Scenes/200_Assets/Scripts/OutputTextureFeature.cs
@@ -78,9 +78,9 @@ public class OutputTextureFeature : ScriptableRendererFeature
             CommandBuffer cmd = CommandBufferPool.Get();
             using (new ProfilingScope(cmd, m_ProfilingSampler))
             {
-				// SetRenderTarget has logic to flip projection matrix when rendering to render texture. Flip the uv to account for that case.
-				CameraData cameraData = renderingData.cameraData;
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -456,14 +456,6 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
             // If possible try to merge the opaque and skybox passes instead of splitting them when "Depth Texture" is required.
-#if ENABLE_VR && ENABLE_XR_MODULE
-            // This optimization is temporarily disabled on VR/XR because of a depth copy y-flip issue on DX11.
-            if (cameraData.xr.enabled)
-            {
-                m_CopyDepthPass.renderPassEvent = (!requiresDepthTexture && (applyPostProcessing || isSceneViewCamera || isGizmosEnabled)) ? RenderPassEvent.AfterRenderingTransparents : RenderPassEvent.AfterRenderingOpaques;
-            }
-            else
-#endif
             {
```
sandy-carter-unity, line 459: This open and close brackets should be removed and the block unindented.

thomas-zeng: Done in 
f7926b2

PR #5126, opened 2021-07-13
Added public GetMainLight API for HDRP

adrien-de-tocqueville:
Added a public function to get the sun light from script in HDRP:
```cs
var sunLight = (RenderPipelineManager.currentPipeline as HDRenderPipeline).GetMainLight();
```
or simply
```cs
var sunLight = HDRenderPipeline.currentPipeline.GetMainLight();
```
Note: This will return null until hdrp renders a camera.

A SG node doing the same thing will be added by #5086


merpheus-dev:
This is not just adding an api callback for GetMainLight but replaces/removes GetCurrentSunLight function. Adding a notice about "this replaces GetCurrentSunLight" to change log would be useful, unless adding a stub for 
```

[System.Obsolete]
GetCurrentSunLight()=>GetMainLight() 

```
is an intention.

adrien-de-tocqueville:
@merpheus-dev That's right, however, the function GetCurrentSunLight was marked as internal, so it was not accessible to users. So I don't think renaming the function will break users project

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -636,7 +636,12 @@ enum ClusterDepthSource : int
         int m_CurrentShadowSortedSunLightIndex = -1;
         HDAdditionalLightData m_CurrentSunLightAdditionalLightData;
         DirectionalLightData m_CurrentSunLightDirectionalLightData;
-        internal Light GetCurrentSunLight() { return m_CurrentSunLight; }
+
+        /// <summary>
+        /// Main directional Light for the HD Render Pipeline.
+        /// </summary>
+        /// <returns>The main directional Light.</returns>
+        public Light GetMainLight() { return m_CurrentSunLight; }
```
sebastienlagarde, line 644: Guess is weird to expose this Method inside the LightLoop class. Guess it should be in another location, maybe HDRenderPipeline. What are your thought?

adrien-de-tocqueville: There isn't actually any lightloop class, it's just the name of the file. The function is inside the class HDRenderPipeline, which is `partial`, so defined in multiple files.

For user it will be accessible this way:
```cs
var sunLight = (RenderPipelineManager.currentPipeline as HDRenderPipeline).GetMainLight();
```

sebastienlagarde: ah right :)

sebastienlagarde (approved):

PR #5129, opened 2021-07-13
[VFX] Remove shader warnings using VFX graph with builtin in 10.x

gabrieldelacruz:
---
### Purpose of this PR
VFX graph package is a dependency for URP and HDRP so, when the package is added, no SRP is present and will try to compile with legacy code for builtin.
This PR undefines SAMPLE_DEPTH_TEXTURE and SAMPLE_DEPTH_TEXTURE_LOD macros that will be redefined later.
In later versions of VFX graph this is not an issue, since legacy support for builtin has been removed.

---
### Testing status
Just adding the VFX graph package version 10.x would generate some shader warnings before. 

---


julienf-unity (approved):

VladNeykov (approved):
Tested by developer, affects only the legacy code.
Auto-approving.

PR #5130, opened 2021-07-13
Handle bindMS in rendergraph fallback textures

adrien-de-tocqueville:
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1348232/

When disabling opaque objects, the normal buffer is not written to but is read, so there is a fallback mechanism in rendergraph to assign a black texture. This breaks when the texture is supposed to be MSAA because the fallback texture is not.

Fixed it by force allocating a texture when it has not been written to and no rendergraph default resource can be used as a fallback

### Test status

tested the repro case


JulienIgnace-Unity (commented):
I'm not sure the fix is good on all platforms.
Some platforms will not react the same to an unbound texture. I guess the actual proper fix is to provide MSAA default textures as well.

iM0ve (approved):
Additionally, I tested for regressions on a more complex scene (RTX template). No issues found

https://user-images.githubusercontent.com/36502659/128873659-9fd14748-c22a-4fed-8d1a-9e46e3c7ab6d.mp4

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #5132, opened 2021-07-13
Prevent material from having infinite intensity

adrien-de-tocqueville:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1349738

Clamped material intensity to prevent reaching infinity
Also fix a regression when changing from ldr to hdr which messed the emissive color (this came from a bad copy paste, doesn't need more testing)

---
### Testing status

The bug report mentions two issues:
- slider even goes to INF, should be clamped
- scene view breaks and doesn't recover after reaching high intensity values

This PR is for the first one
Anis has made a second PR to fix an issue with nans in ssgi


TomasKiniulis (changes requested):
For permanent black screen. When intensity is reached until the black screen occurs reducing intensity value leaves the screen still black. But moving camera away from particles and looking back fixes it. I think that's what the bug report meant.

This PR doesn't seem to fix the issue though. Dragging EV100 intensity slider in the repro case all the way to max still results in black screen.

https://user-images.githubusercontent.com/50582134/125607529-42d15441-9eaf-4232-a37d-bc02ae83b861.mp4

My suggestion would be to clamp the value until EV100 reaches 115 which would prevent any black screens having the highest exposure value(15) set on the Exposure slider before warning icon is shown. This would at least minimize the encounter rate of this issue. I think we should also have Light(Intensity) and Material(Emission Intensity) docs mentioning this. Something to raise discussion again I suppose.

sebastienlagarde (approved):

TomasKiniulis:
I'm off to vacation next week. Requesting @Unity-Technologies/gfx-qa-hdrp to help out rechecking this

iM0ve (approved):
I have verified that it is no longer possible to reach Infinity emission value. 

**1) EXPECTED** There is still some blinking ~~but I assume this is what is being fixed with another PR "_Anis has made a second PR to fix an issue with nans in ssgi_"~~
https://user-images.githubusercontent.com/36502659/127151548-efeb11ad-ba02-4c04-a339-3ab72ab29a4f.mp4
**2) FIXED** The value snaps to 131 when anythin above 127 is used

https://user-images.githubusercontent.com/36502659/127153539-a5792f66-7d3d-4db5-acce-4ce4abba16cd.mp4

iM0ve (changes requested):
Talked with Adrien, he mentioned its not the same issue and will try to look into if it can be fixed here

iM0ve (approved):
The issue with snapping to 131 is fixed and now the maximum is a rounded 130.
The issue with blinking near max values still exists, but is explained to be by design. There is no easy solution since it depends on other things such as multiplication by color end etc. Since this is a very edge case I think its okay to leave the behavior as is for now.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightUtils.cs
```diff
@@ -194,7 +196,7 @@ public static float ConvertEvToLux(float ev, float distance)
         public static float ConvertLuminanceToEv(float luminance)
         {
             float k = ColorUtils.s_LightMeterCalibrationConstant;
-            return (float)Math.Log((luminance * 100f) / k, 2);
+            return Mathf.Log(luminance, 2) + s_evToLuminanceFactor;
```
sebastienlagarde, line 199: name of the constant is a bit weird in this context

adrien-de-tocqueville: I made two variables to clarify things

sebastienlagarde (approved):

PR #5137, opened 2021-07-13
[ShaderGraph] Add searcher DB provider, base class for registry nodes

dhsavell:
### Purpose of this PR
This PR creates the necessary classes to add a node from the GTF Searcher. Most notably, it adds:
- `RegistryNodeModel`, which is a node model whose topology is delegated to the registry. Akin to `CustomizableNode` from the Playground, but stores a known key instead of arrays of ports.
- `RegistryNode`, which is a view for a `RegistryNodeModel`.
- `RegistryNodeSearcherItem`, which is a searcher item associated with a registry key. It creates a `RegistryNodeModel` when selected, and exposes its associated key for ease of filtering.
- `ShaderGraphSearcherDatabaseProvider`, which generates searcher databases from the registry. Focused for now on nodes, will expand to types eventually.
- `ShaderGraphSearcherFilterProvider`, which will eventually provide searcher filtering based on registry keys.

---
### Comments to reviewers

Still WIP - currently blocked on registry interface and types, numerous TODO/FIXME comments. `PlaceholderRegistryKey` should be removed before merging. At that point, the associated field names can probably made more descriptive too.

Review items:
- Seems like RegistryNode doesn't need to cache its ports since the function mapping its inputs and type to its topology is deterministic -- any oversight here?

### DIFF
--- a/com.unity.shadergraph/Editor/Generation.meta
+++ b/com.unity.shadergraph/Editor/Generation.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: e60bcf879d5f67245b7d7613a611f02e
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/com.unity.shadergraph/Editor/Generation/Processors.meta
+++ b/com.unity.shadergraph/Editor/Generation/Processors.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 08d3b851f6777b1479d19e1f33a4c404
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/com.unity.shadergraph/Editor/GraphDelta/AssemblyInfo.cs
+++ b/com.unity.shadergraph/Editor/GraphDelta/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Unity.ShaderGraph.GraphDelta.Tests")]
--- a/com.unity.shadergraph/Editor/GraphDelta/AssemblyInfo.cs.meta
+++ b/com.unity.shadergraph/Editor/GraphDelta/AssemblyInfo.cs.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: ed78f92558fb16344abc674de786f265
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/com.unity.shadergraph/Editor/GraphDelta/ContextLayeredDataStorage.cs
+++ b/com.unity.shadergraph/Editor/GraphDelta/ContextLayeredDataStorage.cs
@@ -1,0 +1,257 @@
+using System.Collections.Generic;
+
+namespace UnityEditor.ShaderGraph.GraphDelta
+{
+    internal class ContextLayeredDataStorage
+    {
+        public class Element
+        {
+            public string id;
+            public List<Element> children;
+            public Element parent;
+
+            internal Element()
+            {
+                id = "";
+                parent = null;
+                children = new List<Element>();
+            }
+
+            internal Element(string id)
+            {
+                this.id = id;
+                parent = null;
+                children = new List<Element>();
+            }
+
+            //It is a path id if there are any '.' characters
+            public static bool IsPath(string id)
+            {
+                return id.IndexOf('.') >= 0;
+            }
+
+            public void AddChild(Element child)
+            {
+                child.parent = this;
+                children.Add(child);
+            }
+
+            public Element<T> AddData<T>(string id, T data)
+            {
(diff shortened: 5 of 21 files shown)


com.unity.shadergraph/Editor/GraphUI/GraphElements/RegistryNode.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    public class RegistryNode : CollapsibleInOutNode
+    {
+        protected override void BuildPartList()
+        {
+            base.BuildPartList();
+
+            if (Model is not RegistryNodeModel registryNodeModel) return;
+
+            // TODO: Build part list from node definition
```
Nightmask3, line 18: Wondering what this might look like more formally...

if  `NodeModel` exposes the actual NodeDefinition from registry (as it contains the registry key for lookup)

```
var nodeDefinition = Model.GetNodeDefinition();
foreach(var nodeDefinitionPart : nodeDefinition.Parts)
{
      if(nodeDefinitionPart.IsInput)
            AddDataInputPort(nodeDefinitionPart.name, nodeDefinitionPart.TypeHandle, .....)
      else
            AddDataOutputPort(nodeDefinitionPart.name, nodeDefinitionPart.TypeHandle, .....)
            
}
```

Will clarify with Esme what to expect in terms of the Node Topology and Type Information we can retrieve from the Registry

Nightmask3: Also, that has me thinking about where we will define whether a node has a preview or not attached to it, probably makes sense for it to be part of the `NodeDefinition.Parts` ? 

Then we can just react to it and add a 'NodePreviewPart` to the node part list in response

Curious on your thoughts here

Nightmask3 (commented):
Looks good overall! Gave me some new points to clarify with Esme, spawned some further questions we need to answer, but all of it makes sense 👍 

Good work! 😄 

Not sure if you had the intention to merge this just yet due to the `Obsolete` attribute on `PlaceHolderRegistryKey`, as that breaks the editor compilation

com.unity.shadergraph/Editor/GraphUI/DataModel/Searcher/ShaderGraphSearcherDatabaseProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace UnityEditor.ShaderGraph.GraphUI.DataModel
+{
+    public class ShaderGraphSearcherDatabaseProvider : DefaultSearcherDatabaseProvider
+    {
+        public ShaderGraphSearcherDatabaseProvider(Stencil stencil)
+            : base(stencil)
+        {
+        }
+
+        SearcherDatabaseBase CreateNodeDatabaseFromRegistry(IGraphModel graphModel)
```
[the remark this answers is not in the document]
dhsavell: Yeah, if we can eventually query by arbitrary criteria then I'd think we can get a list of everything. Let me know what you find out.

It would be great if we could get some enumerator of nodes to map into `SearcherItems`. One thing I'm curious about is how categories (like Artistic, Channel, and especially nested ones like Input/2D etc.) will be presented to us.

Also not accounted for yet is the dynamic stuff -- blackboard variables, custom interpolators, etc. We can probably generate another `SearcherDatabaseBase` for that and add it to the list a few lines down in `GetGraphElementsSearcherDatabases` when the time comes.

com.unity.shadergraph/Editor/GraphUI/DataModel/Searcher/ShaderGraphSearcherDatabaseProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        public override List<SearcherDatabaseBase> GetGraphElementsSearcherDatabases(IGraphModel graphModel)
+        {
+            var databases = base.GetGraphElementsSearcherDatabases(graphModel);
+            databases.Add(CreateNodeDatabaseFromRegistry(graphModel));
+            return databases;
+        }
+
+        SearcherDatabaseBase CreateTypeDatabaseFromRegistry()
```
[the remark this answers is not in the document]
dhsavell: If all of our TypeHandle stuff could go through the same place that'd be great. Since we're mostly focused on appearance and connectivity on our end, that should spare us some of the painful details. Let's look over this once we get to types -- I'm hoping we'll find a convenient registry type <-> TypeHandle mechanism.

com.unity.shadergraph/Editor/GraphUI/GraphElements/RegistryNode.cs
dhsavell: Starting with the ports specifically: those will be handled in `RegistryNodeModel.OnDefineNode` with a loop similar to your sample. Based on what I understand, the registry will have a deterministic function that maps (node type, node connections) -> topology for us to use. GTF automatically makes port parts for us, so that's already accounted for.

Now, other node parts are a bit more interesting. The registry could give us a list of parts (i.e. swizzle node -> [mask text field, preview]), but because the registry isn't responsible for anything GTF, we'd have to figure out how to map its concepts into visual elements. This sounds like something we'd want to discuss further as a whole -- I'm not actually sure where the responsibility should be.

dhsavell:
Yeah, the Obsolete attribute is intentional -- I'm hoping to have PlaceholderRegistryKey removed and replaced with the actual registry node type class before merging.

dhsavell:
Merging into the staging branch so @esmelusina and I can work on registry integration.

PR #5141, opened 2021-07-14
FogBugz#1348462 Random Crash fix from HDRP Wizard

alex-vazquez:
### [fogbugz](https://fogbugz.unity3d.com/f/cases/1348462/)

I wasn't able to reproduce it, but from the call stack I guess that the problem was comming that the the check that the wizard needs to be shown at startup was changed to be on frame 0 instead of frame 10 as it was before.

This PR rollsback that line where it was before.

---
### Testing status
Check the wizard is opened at startup when HDRP is in use.

---
### Comments to reviewers
- I would like that QA retest what was tested on the #4898.
- No need of changelog, as this was recently introduced.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
@@ -252,10 +252,6 @@ namespace UnityEditor.Rendering.HighDefinition
 
         static void WizardBehaviourDelayed()
         {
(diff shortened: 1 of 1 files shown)


RSlysz (approved):
It is from long time now. But yes it is possible that one part of the delayed frame was to be able to access the file that register the "should it pop up?".

If this work, it worth a comment on this on the frame delaying.

sebastienlagarde (approved):

PR #5148, opened 2021-07-15
Only show HDRP camera preview if HDRP is the active pipeline [case 1350767]

jenniferd-unity:
---
### Purpose of this PR
When switching from HDRP to URP with 2 camera inspectors open, we end up with an exception being thrown by the HD Camera inspector. This PR enforces the fact that this custom inspector is not shown if HDRP is not the active pipeline.

Fogbugz: https://fogbugz.unity3d.com/f/cases/1350767/

---
### Testing status
Tested on 2021.2 and 2022.1 in a Windows Editor with steps described in the Fogbugz. If  testing on Mac could be done it would be reassuring.

---
### Comments to reviewers
This bug has been discovered while fixing: https://fogbugz.unity3d.com/f/cases/1348038/ (where solution is to reload all inspectors on SRP type changed)


TomasKiniulis (approved):
Thanks for the fix! Checked on a mac with two locked/unlocked inspectors while Main Camera is selected, switching render pipelines back and forth does not reproduce the issue anymore.

RSlysz (approved):
I think this is tricky and can be forgotten easily :/ Would be great to have an automated test that check changing renderpipeline while component X is active works fine, X from a list of critical component to check. (can be done later)

PR #5149, opened 2021-07-15
[VFX] Mobile fixes

julienf-unity:
### Purpose of this PR
Some mobile fixes for VFX Graph:
- https://fogbugz.unity3d.com/f/cases/1348666/
- https://fogbugz.unity3d.com/f/cases/1149057/

This is a first pass, and other fixes are required for mobile devices (in particular some C++ changes)

For 1348666, I've checked with @hkr, and decided to guard functions using cubemap arrays between SHADER_AVAILABLE_CUBEARRAY check. They may revert to old behavior possibly at some point.

For 1149057, The issue is caused by driver bug for certain integer instructions (in this case, it's with bitFieldExtract). I've worked around them by using float instruction instead.
I've tested all outputs: There's still a potential issue with cube output uv computation, but as it is an experimental feature that has other bugs, I don't have an immediate fix for that (and not sure if it's worth it anyway)

---
### Testing status
Tested locally on Oculus Quest 2 with GLES3.1 (Vulkan does not exhibit the issue)

---
### Comments to reviewers


com.unity.visualeffectgraph/Shaders/ParticleHexahedron/Pass.template
```diff
@@ -28,8 +28,8 @@ VFX_VARYING_PS_INPUTS vert(uint id : SV_VertexID, vs_input i)
 	
 	float3 offsets = (float3)0;
 	offsets.x = float(id & 1);
-	offsets.y = float((id & 2) >> 1);
-	offsets.z = float((id & 6) >> 2);
+	offsets.y = (id & 2) * 0.5f;
+	offsets.z = (id & 4) * 0.25f;
```
gabrieldelacruz, line 32: Why is it (id & 4) and before it was (id & 6)? I assume before it was a mistake, right?

gabrieldelacruz (approved):
Looks good!

com.unity.visualeffectgraph/Shaders/ParticleHexahedron/Pass.template
julienf-unity: Yep most probably a typo but as it was shifted by 2 in previous code, &6 was equivalent to &4. Of course it wasnt equivalent anymore with new code.

VitaSkruibyte:
Hey @julienf-unity! Here's the WIP [Test Doc](https://docs.google.com/document/d/1D3twDRhJv3siFxLikPuPe_xXUL6NW9ZJNy-_At_DQHA/edit?usp=sharing). Main issue of _extended_ particles on Android is fixed, it haven't occured on any of the devices during first QA pass. _Whew!_ 

Speaking further, some of the issues are already summarized with provided reproduction and can be worked on. Some of them still need some attention, thus dropped some questions in the doc. Also, if there's anything that could make investigation easier, let me know.

VitaSkruibyte (approved):
Approving as the main issue is fixed, other issues will be looked into in another fix pass.

julienf-unity:
https://github.com/Unity-Technologies/Graphics/pull/5582 must be backported at the same time as this one

PR #5152, opened 2021-07-15
AMD FidelityFX 1.0

kecho:
### Purpose of this PR
Adding new AMD Fidelity FX 1.0 super resolution.

---
### Testing status
Added new yamato tests for pass.


sebastienlagarde (approved):

sebastienlagarde:
> any chance to make FSR available for unity 2019.4 ?

Nope. FSR is 2021.2 only and above.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -4503,6 +4509,137 @@ TextureHandle ContrastAdaptiveSharpeningPass(RenderGraph renderGraph, HDCamera h
...
+                    passData.destination = builder.WriteTexture(GetPostprocessUpsampledOutputHandle(renderGraph, "Robust Contrast Adaptive Sharpen Destination"));
+                    passData.casParametersBuffer = builder.CreateTransientComputeBuffer(new ComputeBufferDesc(1, sizeof(uint) * 4) { name = "Robust Cas Parameters" });
+
+                    builder.SetRenderFunc(
+                        (RCASData data, RenderGraphContext ctx) =>
+                        {
+                            ctx.cmd.SetComputeFloatParam(data.rcasCS, HDShaderIDs._RCASScale, 1.0f);
+                            ctx.cmd.SetComputeTextureParam(data.rcasCS, data.mainKernel, HDShaderIDs._InputTexture, data.source);
+                            ctx.cmd.SetComputeTextureParam(data.rcasCS, data.mainKernel, HDShaderIDs._OutputTexture, data.destination);
+                            ctx.cmd.SetComputeBufferParam(data.rcasCS, data.initKernel, "RCasParameters", data.casParametersBuffer);
```
FrancescoC-unity, line 4554: We should use the HDShaderIDs._* pattern here (also need renaming in shader to have a prepended _

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RobustContrastAdaptiveSharpen.compute
```diff
@@ -0,0 +1,63 @@
...
+#define A_GPU 1
+#define A_HLSL 1
+#define FSR_RCAS_F 1
+
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/ffx_a.hlsl"
+
+// two elements:
+// [0] = const0
+// [1] = const1
+// ComputeBuffer is allocated with stride sizeof(int)*4, 2 elements
```
FrancescoC-unity, line 16: Looks like it is actually a single uint4 and not 2 elements

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/EdgeAdaptiveSpatialUpsampling.compute
```diff
@@ -0,0 +1,81 @@
...
+    c.rgb *= c.rgb;
+#endif
+
+    _OutputTexture[COORD_TEXTURE2D_X(gxy)] = float4(c.rgb,1.0);
+}
+
+
+/*
+Doing this on the GPU despite the fact that CAS provides a CPU version
+of FsrRcasCon(). This is done to prevent us from having to rewrite a C#
```
FrancescoC-unity, line 61: I am not sure this is an advantage? We need to have a new package version even if we change any shader file. I am inclined to think that using a cbuffer is better here, unless I miss something? (same for other shader)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/EdgeAdaptiveSpatialUpsampling.compute
```diff
@@ -0,0 +1,81 @@
...
+    AU4 con2 = EASUParameters[2];
+    AU4 con3 = EASUParameters[3];
+    AF4 c;
+    FsrEasuF(c.rgb, gxy, con0, con1, con2, con3);
+
+#ifdef GAMMA2_INPUT
+    c.rgb *= c.rgb;
+#endif
+
+    _OutputTexture[COORD_TEXTURE2D_X(gxy)] = float4(c.rgb,1.0);
```
FrancescoC-unity, line 55: This will break if we have Alpha support, we need to have a variant for that

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -4503,6 +4509,137 @@ TextureHandle ContrastAdaptiveSharpeningPass(RenderGraph renderGraph, HDCamera h
...
+
+        TextureHandle EdgeAdaptiveSpatialUpsampling(RenderGraph renderGraph, HDCamera hdCamera, TextureHandle source)
+        {
+            if (hdCamera.DynResRequest.enabled && hdCamera.DynResRequest.filter == DynamicResUpscaleFilter.EdgeAdaptiveScalingUpres)
+            {
+                using (var builder = renderGraph.AddRenderPass<EASUData>("Edge Adaptive Spatial Upsampling", out var passData, ProfilingSampler.Get(HDProfileId.EdgeAdaptiveSpatialUpsampling)))
+                {
+                    passData.easuCS = defaultResources.shaders.edgeAdaptiveSpatialUpsamplingCS;
+                    passData.easuCS.shaderKeywords = null;
+                    passData.easuCS.EnableKeyword("GAMMA2_INPUT");
```
FrancescoC-unity, line 4598: Seems like we enable this always, any reason to keep as a multi compile? (If not, we should make it a define to reduce compile time)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RobustContrastAdaptiveSharpen.compute
```diff
@@ -0,0 +1,63 @@
...
+
+[numthreads(64, 1, 1)]
+void KMain(uint3 LocalThreadId : SV_GroupThreadID, uint3 WorkGroupId : SV_GroupID, uint3 dispatchThreadId : SV_DispatchThreadID)
+{
+    UNITY_XR_ASSIGN_VIEW_INDEX(dispatchThreadId.z);
+
+    // Do remapping of local xy in workgroup for a more PS-like swizzle pattern.
+    AU2 gxy = ARmp8x8(LocalThreadId.x) + AU2(WorkGroupId.x << 3u, WorkGroupId.y << 3u);
+
+    AF4 c = (AF4)0;
```
FrancescoC-unity, line 43: Does this mean we always output alpha 0? I think we need an ALPHA safe version here too

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -4503,6 +4509,137 @@ TextureHandle ContrastAdaptiveSharpeningPass(RenderGraph renderGraph, HDCamera h
...
+                                inputTextureSize = new Vector4(maxScaledSz.x, maxScaledSz.y);
+                            }
+                            ctx.cmd.SetComputeTextureParam(data.easuCS, data.mainKernel, HDShaderIDs._InputTexture, data.source);
+                            ctx.cmd.SetComputeVectorParam(data.easuCS, HDShaderIDs._EASUViewportSize,   new Vector4(data.inputWidth, data.inputHeight));
+                            ctx.cmd.SetComputeVectorParam(data.easuCS, HDShaderIDs._EASUInputImageSize, inputTextureSize);
+                            ctx.cmd.SetComputeTextureParam(data.easuCS, data.mainKernel, HDShaderIDs._OutputTexture, data.destination);
+                            ctx.cmd.SetComputeVectorParam(data.easuCS, HDShaderIDs._EASUOutputSize, new Vector4(data.outputWidth, data.outputHeight));
+                            ctx.cmd.SetComputeBufferParam(data.easuCS, data.initKernel, "EASUParameters", data.easuParameterBuffer);
+                            ctx.cmd.SetComputeBufferParam(data.easuCS, data.mainKernel, "EASUParameters", data.easuParameterBuffer);
```
FrancescoC-unity, line 4626: Same as before re the HDShaderIDs._* stuff

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Asset.md
```diff
@@ -84,7 +84,7 @@ These settings control the draw distance and resolution of the decals atlas that
...
+| **- Upscale Filter**            | Use the drop-down to select the filter that HDRP uses for upscaling.<br />&#8226; **Bilinear**: A low quality upsample. The least resource intensive option.<br />&#8226; **Catmull-Rom**: A bicubic upsample with 4 taps.<br />&#8226; **Lanczos**: A sharp upsample. This method can potentially introduce artifacts so you should not use it for extreme upsampling cases for example, when the screen percentage is less than 50%.<br />&#8226; **Contrast Adaptive Sharpen**: An ultra sharp upsample. This option is not meant for screen percentages less than 50% and still sharpens when the screen percentage is set to 100%. This uses **FidelityFX (CAS) AMD™**: For information about FidelityFX and Contrast Adaptive Sharpening, see [AMD FidelityFX](https://www.amd.com/en/technologies/radeon-software-fidelityfx). <br />&#8226; **FidelityFX Super Resolution 1.0 AMD™**: a spatial super-resolution technology that leverages cutting-edge algorithms to produce impressive upscaling quality at very fast performance.|
```
FrancescoC-unity, line 87: Unrelated to this change, but is this even true for CAS

> An ultra sharp upsample. This option is not meant for screen percentages less than 50% and still sharpens when the screen percentage is set to 100%

Unless something got changed recently, CAS is part of Dynamic res and that is turned off for anything but DLSS when resolution is 100%

FrancescoC-unity: Also, if we want to make it work at 100%, we should have the option elsewhere. (To do in a different PR, as I said, unrelated to this)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
kecho: It was there for debugging. Ill remove it since we should always have it on.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RobustContrastAdaptiveSharpen.compute
kecho: ill copy the alpha out.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
kecho: easy

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Asset.md
kecho: What do you mean? take RCAS and FSR completely out of DRS panel?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RobustContrastAdaptiveSharpen.compute
FrancescoC-unity: Worth adding the alpha multi compile like it is used in a lot of PP (search for ENABLE_ALPHA)

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Asset.md
FrancescoC-unity: Nope, have the "run with 100% screen percentage" option outside (so CAS outside of DRS). Not sure if is better to force it running when 100% too as well, but that sounds somewhat limiting (thinking of like temporal upscale, with that you might still want to run CAS after, but being both filters of upscale it won't be possible).

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/EdgeAdaptiveSpatialUpsampling.compute
kecho: Will keep it in compute:
The fsr shader will be tricky to get it compiled in c#. The actual packing function is doing some expensive ALU (which could be done in scalar registers) but still, its simpler / more mantainable to keep this in the GPU and avoid paying this cost for every wave dispatched.
Will update the comment.

FrancescoC-unity: Gotcha, I assumed there was a simple callback, if it is easier in here wherever is easier I am happy with.

com.unity.render-pipelines.high-definition/Documentation~/HDRP-Asset.md
kecho: gotit. I think it shouldnt be an option, should be on %100 of the time dont you think?

kecho: Just for these passes*

FrancescoC-unity: Well yes, but still we want to have an option somewhere outside for when DRS  is _not_ used. 

If I run normal rendering, it is not an obvious UX experience to enable DRS if I never intend to do DRS, just to get a sharpening filter on my normal 100% rendering.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RobustContrastAdaptiveSharpen.compute
kecho: yep. Copy pasted from CAS. Fixing this and all comments.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/EdgeAdaptiveSpatialUpsampling.compute
kecho: For the transition barrier, we could move the init pass all the way to the top of the frame, and that would mitigate some of the barrier issues. But easier said than done, we'd have to check internally in the renderer how barriers are handled

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/EdgeAdaptiveSpatialUpsampling.compute
kecho: as discussed, on it

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/EdgeAdaptiveSpatialUpsampling.compute
FrancescoC-unity: Yeah no worry bothering about it now, we can always do later as optimization when we are so lucky this is one of our problems :D

FrancescoC-unity (approved):

kecho:
Everything is done from the feedback. Rerunning yamato and making sure this lands all green now.

mahdisml:
Can FidelityFX FSR port to URP too?

PR #5153, opened 2021-07-15
[Universal] [Draft] SetDirty all loaded materials when a URP shadergraph is saved

cdxntchou:
This is attempting to fix https://fogbugz.unity3d.com/f/cases/1335798/... but I don't like the fix.

The problem seems much deeper than suggested in the bug.

When initially loading a project from a clean state, many of the URP material previews will show up as flat gray.
I verified it is calling RenderStaticPreview on each material -- it is just not rendering correctly somehow?
image
image
The one that does show correctly appears to be whichever one is first rendered via RenderStaticPreview..

Newly created materials also show up flat gray:
image

Re-importing the material asset does not fix the preview -- (it appears to not even make the call into RenderStaticPreview).

Bringing up the Material inspector does not fix the preview either.

Changing a value on the Material inspector DOES fix the preview (calling RenderStaticPreview) -- I believe this is because SetDirty(material) will invalidate/rebuild the preview cache for that material.
image

Opening ShaderGraph, changing constant colors in the ShaderGraph, and saving results in the pre views all being re-rendered (RenderStaticPreview is called on each), however they all show up gray.
image
This behavior happens even if the materials are being inspected.
A material that was previously previewed correctly by changing a value on the material, will retain the old preview, and not get the new colors (it actually calls RenderStaticPreview on ALL of the other materials, but not that one).  Subsequently, if you reimport the material, it DOES fix the preview (Wha...??)

With the change in this PR, saving a shadergraph will result in all loaded materials that use that shadergraph being marked dirty, and rebuild those previews.  However it only works if the materials were loaded.  And not all materials that have a preview will be loaded.  At least, any materials that have been inspected appear to always update properly now.
image
image

This PR also doesn't fix any of the other previous behaviors.

Thoughts?  Is there a more fundamental fix we could make to the material previews to fix all of these issues?

I'm thinking possibly the materials being previewed don't show up properly if they are not fully loaded somehow, which might be why they are gray initially..

The background gray changes as well ...  perhaps it is rendering the flat gray previews before the graphics settings (linear/gamma) and/or render pipeline are fully initialized?

Changing AssetPreview logic a bit makes material previews re-render correctly when reimported...  hmm

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/AssetPostProcessors/ShaderGraphMaterialsUpdater.cs
+++ b/com.unity.render-pipelines.universal/Editor/AssetPostProcessors/ShaderGraphMaterialsUpdater.cs
@@ -23,9 +23,6 @@ namespace UnityEditor.Rendering.Universal
             if (!(saveContext is UniversalShaderGraphSaveContext universalSaveContext))
                 return;
 
-            if (!universalSaveContext.updateMaterials)
-                return;
-
             // Iterate over all loaded Materials
             Material[] materials = Resources.FindObjectsOfTypeAll<Material>();
             try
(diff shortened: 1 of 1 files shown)


cdxntchou:
Closing, as the bug is postponed waiting on the larger problems to be solved.

PR #5156, opened 2021-07-19
Fixed LUT initialization in Wireframe mode

adrien-de-tocqueville:
### Purpose of this PR

fix https://fogbugz.unity3d.com/f/cases/1350026/

When a scene view in wireframe mode is opened during HDRP initialization (after a domain reload for example), the LUT initialization is wrong because the DrawFullScreen is done in wireframe.

The fix is to delay LUT computation until a non wireframe scene is displayed.
Note: I tried setting GL.wireframe to false and back to true after computation but it didn't work, altough that would probably be a better fix.

---
### Testing status

Tested on the repro project and on master.
The repro project had a sceneview in wireframe and a gameview (not wireframe) openned at the same time, that is the important part to get a repro on other hdrp versions

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -357,6 +357,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Reduced the number shader variants for the volumetric clouds.
 - Fixed motion vector for custom meshes loaded from compute buffer in shader graph (like Hair)
 - Fixed incorrect light list indexing when TAA is enabled (case 1352444).
+- Fixed Additional Velocity for Alembic not taking correctly into account vertex animation
 - Fixed wrong LUT initialization in Wireframe mode.
 
 ### Changed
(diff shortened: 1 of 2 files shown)


FrancescoC-unity (approved):

sebastienlagarde (commented):
I do'nt really like the current approach. I will suggest (if it is possible) that we temporarily disable the wireframe request in order to correctly initialize the texture. i.e I suggest to do this.

bool wireframe = GL.wireframe;
if (GL.wireframe)
 GL.wireframe = false;
 
 DrawFullScreen
 
 GL.wireframe  = wireframe.
 
 (note: I don't know if code above work, but this would be prefered method)

adrien-de-tocqueville:
That's exactly what I tried to do but it didn't work for some reason, the draw calls are still made in wireframe.
I think it's because the draw calls are actually delayed compared to GL.wireframe, which is immediate

JMargevics (approved):
Good to know it was connected to wireframe mode. 
Testing done by @adrien-de-tocqueville is sufficient in this case, can safely approve 👍✅

sebastienlagarde (approved):

PR #5158, opened 2021-07-19
Fix cubemap static preview + remove custom inspector (#2041)

pema99:
---
### Purpose of this PR
This PR is a backport of #2041. The reason for backporting this is to fix case 1349126 which is a HUP regression only affecting 2019.4. The issue lies within the custom cubemap inspector this version of HDRP currently uses, which was later replaced with the builtin inspector in the PR being backported.

Backporting this PR without changes to core would cause a regression, as 2019.4 does not currently have an exposure slider on the builtin cubemap inspector, but the custom HDRP inspector does. This was fixed here https://ono.unity3d.com/unity/unity/pull-request/103240/_/lighting/bugfix/1226322 (case 1226322), and is in the process of being backported to 2019.4 here https://ono.unity3d.com/unity/unity/pull-request/129435/_/2019.4/lighting/bugfix/1226322 (case 1349126).

Thus, this fix cannot land before the fix to core 2019.4 lands.

---
### Testing status
The fixes (both that being backported, and the HUP regression) were verified manually. No automated tests added.

---
### Comments to reviewers
cherry-pick did not cause conflicts with the exception of changes to the changelog.

### DIFF
--- a/com.unity.render-pipelines.core/ValidationExceptions.json
+++ b/com.unity.render-pipelines.core/ValidationExceptions.json
@@ -5,7 +5,7 @@
         {
              "ValidationTest": "API Validation",
 			 "ExceptionError": "Breaking changes require a new major version.",
-             "PackageVersion": "7.6.0"
+             "PackageVersion": "7.7.0"
         }
     ]
 }


kubacupisz (approved):
Looks good! Thank you. :)

alelievr (approved):

sebastienlagarde:
Hi, please poke when the mentionned C++ PR land. 
Also we can have few month between when the next 7.x pacakge for 19.4 will be done and when the binary will be release. Hope this will not perturb too much users

pema99:
@sebastienlagarde The core PR has landed just this morning.

PR #5160, opened 2021-07-19
Fixed ghosting issues if the exposure changed too much (RTGI).

anisunity:
Due to the nature of clamping, it is sometime impossible to re-use the result of previous frames if the different between the two frames is too different (one frame to an other). This rejects the history if the exposure is too different. While creating a exposure adaptation impression, we don't want that at this point of the pipeline.

Testing status
Pair tested on zoom with @remi-chapelain
Shouldn't change anything on the test


anisunity:
https://user-images.githubusercontent.com/42026998/126186026-22e484c5-9e4b-4e7b-a17a-f12f3440bb6e.mp4


https://user-images.githubusercontent.com/42026998/126186029-81416672-bf84-4cd4-afc7-00bb0f859941.mp4

remi-chapelain (approved):

anisunity:
image
 Tests were green before the reformating

sebastienlagarde (approved):

PR #5163, opened 2021-07-19
[ShaderGraph][21.2] Fix wrong searcher appearing when edge of stack is hovered

dhsavell:
### Purpose of this PR

FogBugz case: https://fogbugz.unity3d.com/f/cases/1351733/. This PR fixes an issue where hovering over a context block but not its inner stack (i.e. mouse on the heading or edge of the block, see images below) would cause the global searcher instead of the block-specific one to appear.

This PR doesn't change the behavior of right-click menus.

---
### Testing status

Note that in this test graph, the fragment context searcher should start with "Alpha," the vertex one should start with "Normal," and the global one should start with "Artistic."

#### Before Fix
image
image
image

#### After Fix
image
image
image

#### Mousing Over Stack Still Works

Before:
image
After:
image

#### Undo/Redo Interaction

Moving a block, opening the searcher, undoing/redoing the move without moving the mouse, and re-opening the searcher correctly reflects the change beneath the cursor.

https://user-images.githubusercontent.com/10332426/126213516-1fda3249-1adb-4fee-99a6-75aee9c10a90.mp4

#### Overlaps

I confirmed that the correct searcher window appears when edges overlap with both nodes and other context blocks. This test was motivated by the fact that I worked with mouse over/out events.

https://user-images.githubusercontent.com/10332426/126213444-cfbe02a6-9c9b-436a-8041-e678135efb1b.mp4

https://user-images.githubusercontent.com/10332426/126213625-06cf8fb3-d215-4e9a-9059-af8a2e178bfe.mp4

#### Additional Cases

- Empty block works both inside and outside the "Spacebar to Add Node" frame
  image
  image

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -119,6 +119,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed ShaderGraph sub-graph stage limitations to be per slot instead of per sub-graph node [1337137].
 - Fixed ShaderGraph exception when trying to set a texture to "main texture" [1350573].
 - Fixed a ShaderGraph issue where Float properties in Integer mode would not be cast properly in graph previews [1330302](https://fogbugz.unity3d.com/f/cases/1330302/)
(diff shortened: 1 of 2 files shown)


jessebarker (approved):
LGTM

ghost (approved):
Good testing evidence, very nice job!

PR #5165, opened 2021-07-19
Improve volumetric cloud upsampling

kecho:
### Purpose of this PR
This PR addresses some of the issues found here https://fogbugz.unity3d.com/f/cases/1347250/

The upsampler of the volumetric clouds is not considering the destination pixel centroid within the 3x3 neighborhood of the sample space.
To improve the upsampler, we are also adjusting the weights depending on centroid distance of the 3x3 grid.

Before (1080p x5 zoom):

image
(red dots are the output pixels, the X are the input samples. Since they are weighted equally this will cause a stair stepping artifact)
image

After (1080p x5 zoom):

image
(red dots are the output pixes, even though they use the same neighborhood, they now consider the distance to the samples as part of the weighting, and this smooths the signal).
image

This issue was especially noticeable when DRS was on, but it was still present if zooming in enough in a native resolution, just less noticeable.

Pending: Will have to update a lot of yamato tests screenshots for this one.
 
---
### Performance
* Pending performance analysis impact on consoles.

### Testing status
* Tested with DLSS and DRS. Tested also on native resolution zooming in. 
* Tested on Dx11 and Dx12


anisunity:
Good catch, I didn't explore more given that the problem seemed to appear with DLSS. That said i want to investigate a bit more the weighting so i'll take it from here.

kecho:
@anisunity will take a closer look. Closing this PR for now.

PR #5166, opened 2021-07-20
Disable the roughness based radius optimization for RTR denoising (case 1343012).

anisunity:
https://fogbugz.unity3d.com/f/cases/1343012/
Unfortunately, most of the optimization/tricks that works very well for single bounce ray traced reflections cannot be used in multibounce. I've decided to just disable the roughness based denoise radius adaptation if we have more than one bounce in quality mode.

Testing status
DXR tests are green locally
The issue doesn't repro anymore on the user dataset


sebastienlagarde (approved):

remi-chapelain (approved):
Change makes sense. 
Took a look at the repro and at the updated test. ✔️

PR #5167, opened 2021-07-20
Fixed infinite propagation of nans for RTGI, SSGI and RTSSS(case 1349738).

anisunity:
https://fogbugz.unity3d.com/f/cases/1349738/
Users can specify values that lead to nans in the emissive section. This leads to frames with nan values, which end up stored in the history buffers and propagate even beyond the user resolving the source of nans which is an issue. This PR blocks the Nans from the history buffers for RTGI, SSGI and RTSSS (issue already solved for RTR). The other ray tracing effects store unlit data which do not require this change

Testing status
The issue doesn't repro anymore.
Shouldn't affect the tests


remi-chapelain (approved):

adrien-de-tocqueville (approved):

anisunity:
Getting some really weird erros related to ray tracing failures on linux and vulkan...

anisunity:
https://github.com/Unity-Technologies/Graphics/pull/5193 This PR should fix the issues mentionned above

sebastienlagarde (approved):

JMargevics (commented):
Testing done seem sufficient ✅👍

PR #5171, opened 2021-07-20
[Light Anchor] Fixes for Light Component behavior and Min Distance

pierre-unity:
### Purpose of this PR
Following [this discussion on Slack](https://unity.slack.com/archives/C6Y79CZM0/p1626784574040000), I created this PR.

It solves the following issues:

- It removes the automatic addition of a light component, because this light will be unwanted in several cases, notably with Prefab that do not contain a Light component at the root
image

- When dragging the slider near 0 and beyond, the transform (position) can still be seen changing! Clamping the distance with "Math.Clamp" avoid impacting the transform while dragging the Distance slider and reaching near 0 values.

- The Min distance value is also now set to 0.001m (1mm). Previously, there was a mismatch between the default value (0.0f) and the allowed value in the UI (0.01f). Now, it will be 0.001 by default and it won't let the user use a lower value, as 0 messes up the rotation transform.
image

- The max distance is also clamped at 1000m, as higher values will start to badly affect the precision (the position starts to shift). 1000m is more than enough for any realistic scenario.

---
### Testing status
None

---
### Comments to reviewers
Fairly safe change, but I'm no coder, so I've no idea if this is following good code practices, especially for the way I name variables and use the clamp function on a slider, so feel free to fix any errors. :)

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lights/LightAnchor.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lights/LightAnchor.cs
@@ -20,7 +20,7 @@ namespace UnityEngine
         const float k_AxisLength = 10;
 
         [SerializeField]
-        float m_Distance = minDistance;
+        float m_Distance = k_MinDistance;
         [SerializeField]
         UpDirection m_FrameSpace = UpDirection.World;
 


com.unity.render-pipelines.core/Runtime/Lights/LightAnchor.cs
```diff
@@ -10,17 +10,17 @@ namespace UnityEngine
     /// Represents camera-space light controls around a virtual pivot point.
     /// </summary>
     [AddComponentMenu("Rendering/Light Anchor")]
-    [RequireComponent(typeof(Light))]
     [ExecuteInEditMode]
     [DisallowMultipleComponent]
     [CoreRPHelpURLAttribute("View-Lighting-Tool")]
     public class LightAnchor : MonoBehaviour
     {
+        const float minDistance = 0.001f;
```
RSlysz, line 18: In convension, constant should be prefixed with k_
```suggestion
        const float k_MinDistance = 0.001f;
```

com.unity.render-pipelines.core/Runtime/Lights/LightAnchor.cs
```diff
@@ -10,17 +10,17 @@ namespace UnityEngine
...
         const float k_ArcRadius = 5;
         const float k_AxisLength = 10;
 
         [SerializeField]
-        float m_Distance = 0f;
+        float m_Distance = minDistance;
```
RSlysz, line 23:
```suggestion
        float m_Distance = k_MinDistance;
```

com.unity.render-pipelines.core/Runtime/Lights/LightAnchor.cs
```diff
@@ -70,7 +70,7 @@ public float roll
         public float distance
         {
             get { return m_Distance; }
-            set { m_Distance = Mathf.Max(value, .01f); }
+            set { m_Distance = Mathf.Clamp(value, minDistance, 1000); }
```
RSlysz, line 73:
```suggestion
            set { m_Distance = Mathf.Clamp(value, k_MinDistance, 1000); }
```

RSlysz (changes requested):
Ok. Just the code convention.

pierre-unity (commented):
All changes from Remi accepted.

RSlysz (approved):

PR #5174, opened 2021-07-20
Fix CustomPassUtils.Copy not working on depth buffers

alelievr:
### Purpose of this PR

- Fix CustomPassUtils.Copy not working on depth buffers 
- Add an override of DrawRenderers function with shader tags.
- Fix a nullref when binding an RTHandle allocated from a RenderTextureIdentifier with CoreUtils.SetRenderTarget.

---
### Testing status
Manually tested that the CustomPassUtils.Copy works with depth buffers with this custom pass:

```CSharp
using System.Collections;
using UnityEngine;
using UnityEngine.Rendering;
using UnityEngine.Rendering.HighDefinition;

public class CopyCustomColorBufferPass : CustomPass {
        public RenderTexture rt;            
 
        protected override void Execute(CustomPassContext ctx) {
            var temp = RTHandles.Alloc(rt);
            CustomPassUtils.Copy(ctx, ctx.cameraDepthBuffer, temp);
        }    
    }
```

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fcustom-pass-utils-copy-depth/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7800952/job


pmavridis (approved):
Looks good to me. 

Perhaps it is worth clarifying in the docs what to expect when running this in the AfterPostProcess injection point when DLSS/DRS is enabled.

sebastienlagarde (approved):

JMargevics (approved):
DrawRenderer

Things I checked: 

- Copying entire camera's depth buffer as in Antoine's example ✅
- Drawing objects from specific layer via DrawRenderers into a custom depth buffer and copying it to another one ✅
Both cases verified in RenderDoc

PR #5176, opened 2021-07-20
[FogBugz # 1343363] Fixing render target warning on pipeline settings modification.

kecho:
### Purpose of this PR
Fogbugz: https://fogbugz.unity3d.com/f/cases/1343363/
Upon render pipeline modification, sometimes we would get the warning
"Releasing render texture that is set to be RenderTexture.active!"

This was caused in the dispose call of HDRP, we had some render target still bound. This fix just cleans the render target and UAV states so that deletion can be clean. No more warning!

---
### Testing status
Tested on windows editor, spamming some HDRP settings in live editing, and in play mode / paused play mode.


sebastienlagarde (approved):

sebastienlagarde:
For QA, fix is trivial so I will merge this PR.

JMargevics (commented):
Testing seem sufficient ✅👍

PR #5189, opened 2021-07-21
[HDRP] Fix viewport size when TAA is executed after dynamic res upscale [Hold]

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1348541/

There is a "rare" configuration where we execute TAA after DLSS, to accumulate PB DoF samples. In this case, the viewport of TAA was not set correctly, creating the issue in the bug ticket. For post upscale passes, the scale of the history textures need to also be adjusted.  

---
### Testing status
- PB DOF + TAA seems to work fine in the repro project after the fix.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
@@ -1525,7 +1525,7 @@ namespace UnityEngine.Rendering.HighDefinition
             if (resGroup == ResolutionGroup.AfterDynamicResUpscale)
             {
                 // If TAA runs after dynamic res upscale, then previousScreenSize is always the camera.postProcessScreenSize (is not affected by rthandle scale)
-                passData.previousScreenSize = new Vector4(camera.postProcessScreenSize.x, camera.postProcessScreenSize.y, 1.0f / camera.postProcessScreenSize.x, 1.0f / camera.postProcessScreenSize.y);
+                passData.previousScreenSize = new Vector4(postProcessViewportSize.x, postProcessViewportSize.y, 1.0f / postProcessViewportSize.x, 1.0f / postProcessViewportSize.y);
                 passData.stencil = false;   // we don't upscale the stencil after DLSS (probably not worth the extra cost)
             }
             else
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -1398,10 +1399,11 @@ RTHandle Allocator(string id, int frameIndex, RTHandleSystem rtHandleSystem)
 
         void GrabVelocityMagnitudeHistoryTextures(HDCamera camera, out RTHandle previous, out RTHandle next)
         {
+            Vector2 scale = new Vector2(camera.postProcessScreenSize.x * camera.screenSize.z, camera.postProcessScreenSize.y * camera.screenSize.w);
```
FrancescoC-unity, line 1402: Does this create GC Alloc? It is captured in the lambda, I solved this in the TAAU PR with a different way to handle allocators.

pmavridis: This function is not executed inside a lambda (In this version of the code), so there is no GC alloc. I guess you have moved this in your PR? We will certainly have to merge a few things in the end.
(I'm also runnign a yamato job as we speak to check for any errors and gc allocs)

FrancescoC-unity: The parameter scale is captured by the allocator lambda https://github.com/Unity-Technologies/Graphics/pull/5189/files#diff-617165acfec9a5b4c1b26809291f30938d7a152b8e9fec0e5ae2c661078201dbR1403

pmavridis: Yes,  you are right. Interestingly our tests did not catch any GC allocs, but maybe the guards are not ON when this is executed.

Since you have solved this in the TAAU PR, I guess I should rebase this PR on top of TAAU and also fix any merge conflicts.

FrancescoC-unity: One thing we can do is that we can close this PR and I do the change on top of TAAU _or_ we wait for TAAU to land and you add this on top of the rest.

pmavridis: Ok, for now I have added [Hold] in the title to wait for TAAU, but if you have time to apply this change in your PR, poke me and I will close this one. Otherwise I will merge after your PR has landed.

kecho (approved):

pmavridis:
I'm closing this one, final PR after TAAU got merged is here: https://github.com/Unity-Technologies/Graphics/pull/5270

PR #5190, opened 2021-07-21
Rewrite of APV indirection buffer

FrancescoC-unity:
This PR changes fundamentally how the index buffer in APV works. 

We know have two data structure. The old index buffer is now 

- A flat array of memory with no spatial correspondence (solves the issue of having to allocate a huge index buffer with a big world and we can budget memory)
- Said buffer is divided in chunks of 243 bricks, each cell will grab a contiguous set of chunks out of the free ones. Doesn't really matter from where in the index.  One important thing here is that the unit is not anymore the absolute minimum brick size, but the minimum brick size of the cell (so if min subdivision is 3rd level, we save a lot of space)

There is now an extra data structure, with one entry per cell containing the following packed metadata (a uint3 per cell):
- The index of the first chunk in the above data structure.
- The minimum brick size for the cell.
- The bounds of the area of the cell that actually contains bricks (as we now don't have a cell-sized brick anymore)

This changes  the evaluation at runtime too (see changes ProbeVolume.hlsl) 

The advantages are  several, among the more important: 

 - Decouple index with spatial area covered by the world, this has many advantages, but chief among those is to avoid huge allocations and allow any kind of settings from user. 
 - Memory for the index buffer is manageable from user same way the pool is, it will just determine how much streaming will happen (streaming to come)
 - This new structure operates on cells on a completely independent way, it will make streaming much much easier. Only thing left is to decide what cells to be removed (single function call) and what added (again a single function call now). No need to update the full index buffer anymore. 
 - If only a cell has super high subdivision, the other cells will not pay the cost, the resolution of the index is not _per cell_ ; this will lead to significant memory improvements and unlock more complex setups without crashing OOM :D

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
@@ -209,7 +209,7 @@ namespace UnityEngine.Experimental.Rendering
             int firstValidChunk = -1;
             for (int i = 0; i < m_IndexInChunks; ++i)
             {
-                if (!m_IndexChunks[i])
+                if (!m_IndexChunks[i] && (i + numberOfChunks) < m_IndexInChunks)
                 {
                     int emptySlotsStartingHere = 0;
                     for (int k = i; k < (i + numberOfChunks); ++k)


sebastienlagarde (approved):
I am a tourist

alelievr (approved):

PR #5192, opened 2021-07-22
Added the support of volumetric clouds for baked and realtime reflection probes.

anisunity:
This PR adds the support of volumetric clouds for baked and real-time reflection probes.
For baked reflection probes, the clouds are evaluated at full resolution
For realtime reflection probes, the clouds are evaluated at low resolution, but without temporal reprojection.

By default volumetric clouds are disabled on planars and realtime reflection probes because of the performance cost.
By default volumetric clouds are enabled on the baked reflection probes if the asset allows it.

Testing status
Tests on couple reflection probes (baked and realtime).


com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricCloudsLighting.cs
```diff
@@ -0,0 +1,71 @@
...
+            // Linear (L1) + constant (L0) polynomial terms
+            x1.x = Vector4.Dot(shAr, vA);
+            x1.y = Vector4.Dot(shAg, vA);
+            x1.z = Vector4.Dot(shAb, vA);
+
+            return x1;
+        }
+
+        Vector3 SHEvalLinearL2(Vector3 N, Vector4 shBr, Vector4 shBg, Vector4 shBb, Vector4 shC)
+        {
```
sebastienlagarde, line 20: Do'nt we already have this code somewhere? 
We should not hide generic function like this in file like volumetrics, should be in spherical harmonics. cs file in Core or LightUtil.cs or similar

sebastienlagarde: Pretty sure we have this kind of function for PBS sky already

sebastienlagarde (approved):
Please add in the documentation of cloud the following thing

"By default volumetric clouds are disabled on planars and realtime reflection probes because of the performance cost.
By default volumetric clouds are enabled on the baked reflection probes if the asset allows it."

So users know they need to enable it for realtime reflection if they want it

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricCloudsLighting.cs
anisunity: This PR just moves it inside the clouds code, didn't add it in this PR

anisunity:
@sebastienlagarde I updated the documentation, i'll run the tests.

pierre-unity:
When the volumetric clouds are set to "Local", they will disappear from the reflection probe, regardless of the far clip plane distance in the reflection probe capture settings.

anisunity:
@pierre-unity you didn't set the frame setting

anisunity:
Ok indeed something wrong with the rt probes in local mode

anisunity:
Fixed realtime local clouds

anisunity:
This PR requires some complex conflict resolving :(

remi-chapelain (changes requested):
Couples questions and (possibly) issues
- What's the reason behind capturing the real-time reflection at a quarter resolution? 
I understand it could be performance if they are used in update "every frame" mode but what about the OnDemand or OnEnable mode, could they not benefit for full resolution too? 
- If we keep this (realtime at quarter, baked at full resolution), could we add this info to the documentation to avoid confusion from users ? 
- Temporal accumulation factor doesn't have any effect (regardless if the probe is baked or real time) but PR description suggest otherwise, (i.e only disabled for real-time reflection probe).
Is this expected ? 
- Lastly, I believe this is expected but just want to make sure. 
When there's no **static lighting sky** setup in lighting tab and you're using dynamic **ambient mode** in your visual environment override, the **ambient light probe dimmer** parameter works properly on the volumetric clouds but it looks like it's not taken into account when the reflection probe is in baked mode, is this expected too ? 
(note : this works fine if you set your lighting sky properly and your ambient mode to static)

https://user-images.githubusercontent.com/57442369/127019017-6fe8f0cd-19e2-4628-9439-6c885b19a6cf.mp4

anisunity:
Merged master, added new tests relaunching the tests to be able to update the screenshots

anisunity:
Everything is fine, except, i'll fix an related issue of clouds on vulkan in this PR first.

sebastienlagarde:
Discussed this Pr with Anis and Remi before Remi leave for vacation. Was agree to merge it after Yamato green result. Merging

PR #5193, opened 2021-07-22
Fixed failures on platforms that do not support ray tracing due to an engine behavior change.

anisunity:
Creating an RTAS on platforms that do not support ray tracing leads to an assert now, this avoids that initial creation (that was useless anyway).
https://unity.slack.com/archives/CBBPCEPTQ/p1626948897231700

Testing status
DXR tests still pass locally.


sebastienlagarde (approved):

sebastienlagarde:
Will fix issue https://fogbugz.unity3d.com/f/cases/1352512/

PR #5195, opened 2021-07-22
Fix white flash when changing upsampling to before post process.

FrancescoC-unity:
Currently in master this is only a bug for DLSS I suspect, but I need to fix this for TAAU. 

The wrong exposure texture was used with the reset history configuration that happens when switching to an upsampler that happens before post processing. I suspect other situations will trigger the same bug, but this is where I noticed it. 

@remi-chapelain  this fixes the issue you mentioned in the TAAU PR about white flashes, this PR is merged into that branch if you want to check. 

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-exposure-used-when-history-is-reset/.yamato%252Fall-hdrp.yml%2523PR_HDRP_2021.2


sebastienlagarde (approved):

anisunity (approved):

kecho (approved):

remi-chapelain (approved):
Already tested in [other branch](https://github.com/Unity-Technologies/Graphics/pull/5066) ! ✔️

PR #5200, opened 2021-07-22
Small fix for shader analysis

adrien-de-tocqueville:
### Purpose of this PR

`PackageManager.PackageInfo.GetAll` is not an internal API anymore and it has been renamed `PackageManager.PackageInfo.GetAllRegisteredPackages`

which breaks current package

Also added an error message if symbolic links creation fails. I didn't manage to fix the creation itself, but at least this error message will maybe help someone in the future

---
### Testing status

I tried to use the shader analysis package, now it doesn't crash because of the first issue and it throws in case symbolic link creation fails

I don't know if there is tests to launch for the analyser (I guess not otherwise they wouldn't work ?)

### DIFF
--- a/com.unity.shaderanalysis/Editor/Internal/PackagesUtilities.cs
+++ b/com.unity.shaderanalysis/Editor/Internal/PackagesUtilities.cs
@@ -49,7 +49,7 @@ namespace UnityEditor.ShaderAnalysis.Internal
             if (!links.AreSymbolicLinks())
             {
                 if (!SymbolicLinkUtilities.CreateSymbolicLinks(links, targets))
-                    throw new Exception("Failed to create symbolic links");
+                    throw new Exception("Failed to create packages symlinks in ProjectFolder/Library/ShaderAnalysis/PackageSymlinks/Packages/*. Please check if you have sufficient privileges to create symlinks or manually create the symlinks for each package.");
             }
         }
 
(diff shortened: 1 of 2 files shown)


fredericv-unity3d (changes requested):
Good catch on the API update, it is definitely a code quality improvement.

Also good thinking with the error message, but you can push it further: You should include a better explanation and some resolution hints. The users running into this error won't probably have the knowledge to solve that, so we need to provide them.

I am thinking about something like: 
"Failed to create packages symlinks at MyProjects/Packages/*, please check if you have the privileges to create symlinks or create the symlinks your self for each package in the MyProject/Packages/ directory."


Final comment (but this one is not mandatory for this PR): on windows, creating symlinks requires specific privileges, so you either need to grant them for the current user, or use an elevated process. This may be the cause of your symlink creation failure.

adrien-de-tocqueville:
I changed the message as requested.

For the symlink creation, yes it was indeed the issue, even though my windows account is admin. But it's probably that Unity is not launched with admin rights, so I changed the process to always execute in "runas" mode. That fixes the issue for me, do you think it's correct ?

fredericv-unity3d:
To me its ok, but it would be best to ask for someone more familiar with window privilege systems, perhaps with people in the SDET team?

Modifying privilege can have undesirable side effects.

fredericv-unity3d (approved):

sebastienlagarde:
@theo-pnv do you have any thought about this topics? thanks

theo-pnv (approved):
Sorry for the delay, I apparently missed some notification.
I discussed the elevated rights changes with Adrien on Slack. Everything looks good.

PR #5201, opened 2021-07-22
[2021.2+] Explicit Preset support in SRP for Lights, Cameras, and Reflection Probes (case 1335979)

jenniferd-unity:
---
### Purpose of this PR
Camera, Light, and Reflection Probes are implemented in our SRPs with Additional Data Components. However, the Preset system is not yet ready to support those additional components on Objects. This PR aims to clarify in the UI of Presets for these components which properties are supported (those who are not are now hidden).

### URP:
URP Light Preset:
image

URP Camera Preset:
image

URP Reflection Probe system reuses Built-in so there is no change.

### HDRP:  (updated after TA+UX approval)
HDRP Light Preset:
image

HDRP Camera Preset:
image

HDRP Reflection Probe Preset:
image

---
### Testing status
Tried on Windows Editor of trunk to switch between SRPs and check which fields were supported. tested mainly on 2021.2 branch. Please test on trunk too and Mac if possible.

---
### Comments to reviewers
I could not find a good place in URP to have the temporary utility function IsPresetEditor() so I decided to copy/paste it in two places.
A good way to test if a property is supported in the Preset is to Exclude All Properties from it, properties that don't have a Red line on the left side are not supported. For example, before the change we had:
image


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -240,7 +240,8 @@ static void DrawShapeContent(SerializedHDLight serialized, Editor owner)
 
                 case HDLightType.Point:
                     EditorGUI.BeginChangeCheck();
-                    EditorGUILayout.PropertyField(serialized.shapeRadius, s_Styles.lightRadius);
+                    using (new EditorGUI.DisabledScope(false))
```
RSlysz, line 243: A disableScope with (false) do nothing. The value is to know if this should be disable or not.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Rendering.Drawers.cs
```diff
@@ -6,9 +6,9 @@ namespace UnityEditor.Rendering.HighDefinition
 {
     using CED = CoreEditorDrawer<SerializedHDCamera>;
 
-    static partial class HDCameraUI
+    internal static partial class HDCameraUI
```
RSlysz, line 9: internal is default visibility for class outside struct / class. This is not needed

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDEditorUtils.cs
```diff
@@ -276,5 +276,11 @@ internal static void HandlePrefixLabelWithIndent(Rect totalPosition, Rect labelP
             labelPosition.x += EditorGUI.indentLevel * 15;
             EditorGUI.HandlePrefixLabel(totalPosition, labelPosition, label);
         }
+
+        // IsPreset is an internal API - lets reuse the usable part of this function
+        static internal bool IsPresetEditor(UnityEditor.Editor editor)
+        {
+            return (int)((editor.target as Component).gameObject.hideFlags) == 93;
```
RSlysz, line 283: Any case where editor or editor.target can be null?

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -259,6 +259,12 @@ public new void OnDisable()
             base.OnDisable();
         }
 
+        // IsPreset is an internal API - lets reuse the usable part of this function
+        static internal bool IsPresetEditor(UnityEditor.Editor editor)
+        {
+            return (int)((editor.target as Component).gameObject.hideFlags) == 93;
```
RSlysz, line 265: same missing null check?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineLightEditor.cs
```diff
@@ -15,11 +15,24 @@ protected override void OnEnable()
             serializedLight = new UniversalRenderPipelineSerializedLight(serializedObject, settings);
         }
 
+        // IsPreset is an internal API - lets reuse the usable part of this function
+        static internal bool IsPresetEditor(UnityEditor.Editor editor)
+        {
+            return (int)((editor.target as Component).gameObject.hideFlags) == 93;
```
RSlysz, line 21: same missing null check?

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.PresetInspector.cs
```diff
@@ -0,0 +1,30 @@
...
+    static class HDLightUIPreset
+    {
+        [HDRPHelpURL("Light-Component")]
+        enum Expandable
+        {
+            General = 1 << 0,
+            Emission = 1 << 2,
+        }
+
+        static readonly ExpandedState<Expandable, Light> k_ExpandedState = new ExpandedState<Expandable, Light>(~(-1), "HDRP");
```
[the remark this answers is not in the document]
jenniferd-unity: okay i will rename to HDRP-Preset

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.Skin.cs
```diff
@@ -4,9 +4,9 @@
 
 namespace UnityEditor.Rendering.HighDefinition
 {
-    partial class HDLightUI
+    internal partial class HDLightUI
```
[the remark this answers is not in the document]
jenniferd-unity: thanks! will remove for simplicity

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
jenniferd-unity: that code was not supposed to be submitted, my apologies, will remove those changes (locally i was testing disabled(true) versus hiding fields)

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Rendering.Drawers.cs
jenniferd-unity: thanks!

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDEditorUtils.cs
jenniferd-unity: not as far as i know, I am passing "this" every time

RSlysz (approved):

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.PresetInspector.cs
```diff
@@ -0,0 +1,30 @@
...
+    static class HDLightUIPreset
+    {
+        [HDRPHelpURL("Light-Component")]
+        enum Expandable
+        {
+            General = 1 << 0,
+            Emission = 1 << 2,
+        }
+
+        static readonly ExpandedState<Expandable, Light> k_ExpandedState = new ExpandedState<Expandable, Light>(~(-1), "HDRP-preset");
```
alex-vazquez, line 21:
```suggestion
        static readonly ExpandedState<Expandable, Light> k_ExpandedState = new ExpandedState<Expandable, Light>(~(-1), "HDRP:LightUIPreset");
```

Overlaping key with different sections on : `HDCameraUIPreset`

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -732,91 +743,93 @@ static void DrawEmissionContent(SerializedHDLight serialized, Editor owner)
 
             EditorGUI.BeginChangeCheck(); // For GI we need to detect any change on additional data and call SetLightDirty
 
-            if (lightType != HDLightType.Area)
+            if (!isInPreset)
```
alex-vazquez, line 746: If I were you I would put this code in a `CED.Conditional` instead

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.PhysicalCamera.Drawers.cs
```diff
@@ -9,7 +9,7 @@ namespace UnityEditor.Rendering.HighDefinition
 
     static partial class HDCameraUI
     {
-        partial class PhysicalCamera
+        public partial class PhysicalCamera
```
alex-vazquez, line 12: If you change this to be `public` I would change the `partial` declaratio too on the file `.Skin,cs`

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.PresetInspector.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            /// <summary> Projection</summary>
+            Projection = 1 << 0,
+            /// <summary> Physical</summary>
+            Physical = 1 << 1,
+            /// <summary> Rendering</summary>
+            Rendering = 1 << 2,
+        }
+
+        static readonly ExpandedState<Expandable, Camera> k_ExpandedState = new ExpandedState<Expandable, Camera>(Expandable.Projection, "HDRP-preset");
```
alex-vazquez, line 23:
```suggestion
        static readonly ExpandedState<Expandable, Camera> k_ExpandedState = new ExpandedState<Expandable, Camera>(Expandable.Projection, "HDRP:CameraUIPreset");
```
Overlaped with the one declared in `HDLightUIPreset`.

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraUI.PresetInspector.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            /// <summary> Projection</summary>
+            Projection = 1 << 0,
+            /// <summary> Physical</summary>
+            Physical = 1 << 1,
+            /// <summary> Rendering</summary>
+            Rendering = 1 << 2,
+        }
+
+        static readonly ExpandedState<Expandable, Camera> k_ExpandedState = new(Expandable.Projection, "URP-preset");
```
alex-vazquez, line 24: Is this the same as `UniversalRenderPipelineLightUIPreset` ?

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.PresetInspector.cs
```diff
@@ -0,0 +1,34 @@
...
+    static class UniversalRenderPipelineLightUIPreset
+    {
+        [URPHelpURL("light-component")]
+        enum Expandable
+        {
+            General = 1 << 0,
+            Emission = 1 << 1
+        }
+
+        static readonly ExpandedState<Expandable, Light> k_ExpandedState = new ExpandedState<Expandable, Light>(~(-1), "URP-preset");
```
alex-vazquez, line 20: Same as the one in the `UniversalRenderPipelineCameraUIPreset`

alex-vazquez (changes requested):
Good PR, some concerns:

- The expanded states that we are saving are correct? do the light mess the camera and viceversa? You could simply check that the sections are properly saved an restored?
- As the sections are the same on HDRP and URP, I would share the camera preset section states between URP and HDRP and the same for the Light Preset states.
- For the behaviour of the `!isPreset`, that you have two method passing one with true and the other with false, I would recommend you to simplify it with a `CED.Conditional`, If you have time to do it in this PR that will be nice, otherwise I would note it down and have a look later as it will simplify the code readability.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.PresetInspector.cs
jenniferd-unity: if that's the case then should I also update for unique strings HDProbeUI, LocalVolumetricFogUI, HDRenderPipelineUI etc?
For ex. i see we have :
```
static readonly ExpandedState<Expandable, HDRenderPipelineAsset> k_ExpandedState = new(Expandable.Rendering, "HDRP");
static readonly ExpandedState<ExpandableShadows, HDRenderPipelineAsset> k_LightsExpandedState = new(0, "HDRP:Shadows")
```
 but also
```
readonly static ExpandedState<Expandable, ProbeVolume> k_ExpandedStateVolume = new ExpandedState<Expandable, ProbeVolume>(Expandable.Volume, "HDRP");
readonly static ExpandedState<Expandable, ProbeVolume> k_ExpandedStateProbes = new ExpandedState<Expandable, ProbeVolume>(Expandable.Probes, "HDRP");
```
@RSlysz thoughts?

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
jenniferd-unity: what's the difference?

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.PhysicalCamera.Drawers.cs
jenniferd-unity: what would it bring? since i only need internal visibility access to the DrawerPreset

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.PresetInspector.cs
alex-vazquez: Ok, I saw the implementation and it is doing: `string key = $"{prefix}:{typeof(TTarget).Name}:{typeof(TState).Name}:UI_State";` so they won't be overlaped. Then you do not need the - preset as it will get the name of the class HDLightUIPreset:Expanded and HDCameraUIPreset:Expanded.

alex-vazquez: I do not know why but I got confused :) to much time from the game jam maybe xD

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.PresetInspector.cs
alex-vazquez: my bad, no overlap at the end

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.PresetInspector.cs
alex-vazquez: no overlap

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.PresetInspector.cs
RSlysz: Well, If I remember well this string is a prefix of the key which is computed this way:
`<Prefix>:<TypeName>:<StateName>:UI_State`
Prefix: to distinguish amongst SRP (usually HDRP / URP / CoreRP)
TypeName: the type name of the component, second template argument (here "Light")
StateName: the type name of the enum encoding the state, first template argument argument ("here "Expandable")

So yes, using ':' is a better separator than '-'. But at the same time it is still part of the prefix disambiguation. Just here we do not only disambiguish by SRP but also by inspector kind (preset or not).
So for me "HDRP-Preset" works fine.
We can chose to use the separator, adding a new segment in the key but in this case, "Light" and "UI" should not be in it as already in the kay name. So it would be "HDRP:Preset"

This is temporary anyway so both works for me (but try not to duplicate information please, the key is already long)

RSlysz: (the key used to store the EditorPref value)

RSlysz: "HDRP:Shadows" for instance is odd as Shadow is already induced but the enum used "ExpandableShadows". IMO it should have remained 
`static readonly ExpandedState<ExpandableShadows, HDRenderPipelineAsset> k_LightsExpandedState = new(0, "HDRP")
`

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.PhysicalCamera.Drawers.cs
alex-vazquez: is just to mantain coherence between partial declarations, that's it

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
alex-vazquez: you will have the full flow using the core editor drawers, now we have a mix of them.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.PresetInspector.cs
alex-vazquez: yeah, we should remove the :Shadows:

RSlysz: For the ProbeVolume it is ill constructed and conflict o_o It should not

jenniferd-unity: this should be fixed in the next commit then :)

RSlysz: Sorry Alex, just saw your message. Git don't refresh itself :D

reallyfancy (approved):
Added an extra sentence where needed to make it clear to users that unsupported properties are hidden.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
jenniferd-unity: hum i tried to have a conditional here but how do i access isInPreset since it is not part of the serialized value and it doesnt make sense to have it as a static property (unless it does?))?

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.PhysicalCamera.Drawers.cs
jenniferd-unity: done in next commit

SeanPuller:
These parameters aren't visible with the new 'hidden' layout but they are still stored
image

SeanPuller (approved):

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -259,6 +259,12 @@ public new void OnDisable()
             base.OnDisable();
         }
 
+        // IsPreset is an internal API - lets reuse the usable part of this function
+        static internal bool IsPresetEditor(UnityEditor.Editor editor)
```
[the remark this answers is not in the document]
svens-unity: At the moment nothing is calling this outside of the class correct? Why not keep it private for the time being? Also taking in the argument is unnecessary in it's current usage?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineLightEditor.cs
```diff
@@ -15,11 +15,24 @@ protected override void OnEnable()
             serializedLight = new UniversalRenderPipelineSerializedLight(serializedObject, settings);
         }
 
+        // IsPreset is an internal API - lets reuse the usable part of this function
+        static internal bool IsPresetEditor(UnityEditor.Editor editor)
```
svens-unity, line 19: Same comments as for `UniversalRenderPipelineCameraEditor.cs `

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.PresetInspector.cs
svens-unity:
```suggestion
        static readonly ExpandedState<Expandable, Light> k_ExpandedState = new ExpandedState<Expandable, Light>(0, "URP-preset");
```
~(-1) == 0

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.Drawers.cs
```diff
@@ -112,6 +112,16 @@ static Func<int> SetGizmosDirty()
...
+        {
+            DrawGeneralContentInternal(serializedLight, owner, isInPreset: false);
+        }
+
+        internal static void DrawGeneralContentPreset(UniversalRenderPipelineSerializedLight serializedLight, Editor owner)
+        {
+            DrawGeneralContentInternal(serializedLight, owner, isInPreset: true);
+        }
+
+        static void DrawGeneralContentInternal(UniversalRenderPipelineSerializedLight serializedLight, Editor owner, bool isInPreset)
```
svens-unity, line 124: Personally I would only keep one `DrawGeneralContent` and take in an argument whether Preset mode is active. `DrawGeneralContent` and `DrawGeneralContentInternal` have the same accessibility level, so it does not enforce an usage. Furthermore, if more options or modes are added in the future the number of methods would expand a lot. That being said this is internal code and will be easy to refactor in the future.

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraUI.PresetInspector.cs
```diff
@@ -0,0 +1,36 @@
+using System.Linq;
+using UnityEngine;
+using UnityEngine.Rendering;
+using UnityEngine.Rendering.Universal;
+
+namespace UnityEditor.Rendering.Universal
+{
+    using CED = CoreEditorDrawer<UniversalRenderPipelineSerializedCamera>;
+
+    static class UniversalRenderPipelineCameraUIPreset
```
svens-unity, line 10: A lot of the access modifiers are changed from private to internal or public. Could we avoid this by integrating the Preset inspector into `UniversalRenderPipelineCameraUI`? This could be a nested class or not. 
In `UniversalRenderPipelineCameraEditor.cs` it could look something along the lines of:
`UniversalRenderPipelineCameraUI.PresetInspector.Draw(m_SerializedCamera, this);`

As a result of this all drawers could remain private. Same for the Styles in the Skin

svens-unity (commented):
Did a quick review over the URP and Core files

com.unity.render-pipelines.core/Editor/Camera/CameraUI.Skin.cs
```diff
@@ -54,6 +54,11 @@ public static class Styles
             /// Far plane content
             /// </summary>
             public static readonly GUIContent farPlaneContent = EditorGUIUtility.TrTextContent("Far", "The furthest point relative to the camera that drawing occurs.");
+
+            /// <summary>
+            /// Info box for unsupported fields
+            /// </summary>
+            public static readonly string unsupportedFieldsPresetInfoBox = "When using Preset of Camera Component, only a subset of properties are supported.  Unsupported properties are hidden.";
```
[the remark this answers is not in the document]
jenniferd-unity: great suggestion, i am changing that right now.

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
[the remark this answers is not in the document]
jenniferd-unity: Comment is not adding new information

> I am talking about the API IsPreset that already exists in PresetEditor.cs - this is a temp fix while the Preset dev makes it public so we can reuse it (UI also duplicated the code in their package)

Yes will change to Private for now for URP - HDRP needed internal visibility- you are right. Changed in next commit.

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.Drawers.cs
jenniferd-unity: Yes i do agree with you. The problem is that we are building the UI with the CoreEditorDrawer APIs which expects a specific signature so i cannot just add the parameter "IsPreset".

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraUI.PresetInspector.cs
jenniferd-unity: we could. i dont really understand the benefit but if thats what we want for URP codebase I dont mind updating the code.
@alex-vazquez any opinion on this? you have seen more URP code than I

com.unity.render-pipelines.core/Editor/Camera/CameraUI.Skin.cs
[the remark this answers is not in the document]
alex-vazquez: maybe also we are missing a translation?

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraUI.PresetInspector.cs
alex-vazquez: I think @svens-unity is having a good point here. If we are avoiding making public API is always a good path to follow.

jenniferd-unity: okay then! thanks for the feedback, i ll work on this asap

jenniferd-unity:
Re-requesting dev reviews since the architecture was changed based on @svens-unity's great feedback!

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightEditor.cs
```diff
@@ -72,15 +73,22 @@ void OnUndoRedo()
...
+
+            if (HDEditorUtils.IsPresetEditor(this))
+            {
+                HDLightUIPreset.Inspector.Draw(m_SerializedHDLight, this);
+            }
+            else
+            {
+                using (new EditorGUILayout.VerticalScope())
+                    HDLightUI.Inspector.Draw(m_SerializedHDLight, this);
+            }
```
alex-vazquez, line 91: Now that I see this it will be nice to have something like `HDEditorUtils.DrawPresetEditor(HDLightUIPreset.Inspector, HDLightUI.Inspector, m_SerializedHDLight, this)` as this else if statement is in multiple places. Internally the method will decide what to draw

jenniferd-unity: I dont think having Preset editors are a great pattern and this is a temporary fix until we can land proper preset support.

alex-vazquez: perfect then :)

com.unity.render-pipelines.core/Editor/Camera/CameraUI.Skin.cs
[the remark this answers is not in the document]
RSlysz:
```suggestion
            /// Info box for unsupported fields
            /// </summary>
            public static readonly string unsupportedFieldsPresetInfoBox = L10n.Tr("When using Preset of Camera Component, only a subset of properties are supported.  Unsupported properties are hidden.");
```

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.Skin.cs
```diff
@@ -160,6 +160,7 @@ sealed class Styles
 
             // Warnings
             public readonly string unsupportedLightShapeWarning = "This light shape is not supported by Realtime Global Illumination.";
```
RSlysz, line 162:
```suggestion
            public readonly string unsupportedLightShapeWarning = L10n.Tr("This light shape is not supported by Realtime Global Illumination.");
```

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.Skin.cs
```diff
@@ -160,6 +160,7 @@ sealed class Styles
...
+            public readonly string unsupportedFieldsPresetInfoBox = "When using Preset of Light Component, only a subset of properties are supported.  Unsupported properties are hidden.";
```
RSlysz, line 163:
```suggestion
            public readonly string unsupportedFieldsPresetInfoBox = L10n.Tr("When using Preset of Light Component, only a subset of properties are supported.  Unsupported properties are hidden.");
```

com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/HDProbeEditor.cs
```diff
@@ -34,6 +34,11 @@ abstract class HDProbeEditor<TProvider, TSerialized> : Editor, IHDProbeEditor, I
 
         public override void OnInspectorGUI()
         {
+            if (HDEditorUtils.IsPresetEditor(this))
+            {
+                EditorGUILayout.HelpBox("Presets of HD Reflection Probe is not supported.", MessageType.Info);
```
RSlysz, line 39: Text should be outside in Styles class and with L10n.Tr for translation

RSlysz (approved):
Even if temporary we don't know exactly for how long. So we need to support localization.

The remains looks good so I pre approve it.

TheoWong-pixel:
Just for the message in the HDRP Reflection probes: "Presets for HDRP Reflection Probes are not supported" do we do not confuse Reflection probes to be a 'HD' feature which is different from a HDRP feature

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.PresetInspector.cs
jenniferd-unity: Thanks Sven! It was applied in latest commit.

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
jenniferd-unity: There is no HideFlags for this yet. so i need to cast to int because we go from Enum => int comparison.
When the API will be public this will not be necessary anymore

jenniferd-unity:
@svens-unity @TheoWong-pixel could you approve/deny changes?
I implemented the changes discussed on the PR and in this morning sync. Thanks

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.PresetInspector.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using UnityEngine;
+using UnityEngine.Rendering;
+using UnityEngine.Rendering.Universal;
+
+namespace UnityEditor.Rendering.Universal
+{
+    using CED = CoreEditorDrawer<UniversalRenderPipelineSerializedLight>;
+
+    internal partial class UniversalRenderPipelineLightUI
```
RSlysz, line 10: internal is already default here. Not needed. (Same for other partial definition of this one)

RSlysz (approved):

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.Drawers.cs
alex-vazquez: Could it be done with a ConditionalDrawer?

alex-vazquez (approved):

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
svens-unity: You could also cast 64 to the enum until the missing entry has been exposed. This will allow you to construct the combination of flags used here and compare using the enum, like [this](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUKgZgAIMiBhIgbzyNqIG0AxAGwEMBzAZwF0a7CiAUygQAtkQASASziCWHTn1rVcdNUQD0GogEEiUAPZhRrZjCIA3KZykAjZoKIHbAK0EBjYADoiAFQAW1kRBwP6OsgBmrBDM3krqAHIGUI4AvEQADDB48WpafmFOrh7ARADuUszM+galrAAO9YKsYMFQRKGOgYJgLe7+AJ5ErFBw5ZXVhqWc/gZlRBD1bR2F9WAGbp6WUoLzUhHBpUGcwEaCY1LtI8OcnIJxquq00rIAklDSPX2DROmY5locLhcnR8q8jpwaqV6gZbnYHB0DNtdodlp02pwmp4jCDnjJBO93piSkZfkR0NlgY8nvkAo5nJtShUqlCiLZHJxWBZzoiVhz3MJBF5cUQACLJYAAZS5BKgAFE4FJTq10gAWAEaACsOWp6lphQZJWCkKmbMc5yVrHsjkufIxWOVIqSwAVlutZIAHDB8pgCDqnrR9fTiltmZNamaiJzuWMymF2rYIJVFVB2MMiPU2AMeiLxVApTL3gAhJPMMZ/ABs3o06AAnP6A0GiozxizTeyFlBmAZWLIxrYhgBVLs9uDDiB3OA6W73RS6tR54DD7u98eT6d3UrpAgUhtPRfS7lk1VERBETAV09EHeUkX5PTuAyiWyXVjAKTJJwHU0zObtW3ot0vRgP0QwjGMP4ymMpy8pwAopJwwrznQLyCDoowHjKZKYFeJ5nh6V4XleO7xAAvnuRDxBg6CUbqKgBpYLQdC07D3AA8iGwCofIXBkgAFNxbBcAAlLWBAANwUWoFhMdYAAKYCCJugkKGSKlcF4qHvJ8wGgbRDEGeoyCSPiPGIZh3LvK6jrIYZTzGepiHOtZVoOCKdl0A5plCeZEqHrKJbJu5HlEMZAneQowkVqqklUiKqCYLWfEACQAESvJwClKfcIBUMALHsZxjm/Ok8mKcpEVcKRqXCbF8WJSl6WZeVOVUI5XgWWkpXNdlXGVZw1W1bid4aAlSX5WArHABxjLFQAZCZshmV4SQpENtmaKNDUTVNM0lPNi1yD5mn4tpOy6YM60MVoY18TthWzf1RALe1WlQESDpGFdjZbeNBXTUVT0vf1HV+YW8qKsq300r9d3/XtngHe1zmQ65gjQ3qsP3QDj1LT5z2HctnXFqWcAY3kWPw4DeOqcDNMaYuy6jmu5wbvc5OgrDZDJNyYDeL4BiSsAYCXOwfF8ZcwDCdjCN9fTnDmOgwkY+RuCkUAA=).

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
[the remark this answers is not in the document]
svens-unity: Did you forget to push? I do not see the change?

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.Drawers.cs
```diff
@@ -112,6 +112,16 @@ static Func<int> SetGizmosDirty()
         }
 
         static void DrawGeneralContent(UniversalRenderPipelineSerializedLight serializedLight, Editor owner)
+        {
+            DrawGeneralContentInternal(serializedLight, owner, isInPreset: false);
+        }
+
+        internal static void DrawGeneralContentPreset(UniversalRenderPipelineSerializedLight serializedLight, Editor owner)
```
svens-unity, line 119: This no longer needs to be internal with the PresetInspector being included in the class?

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.Drawers.cs
```diff
@@ -235,7 +245,7 @@ static void DrawAreaShapeContent(UniversalRenderPipelineSerializedLight serializ
                 serializedLight.settings.DrawArea();
         }
 
-        static void DrawerColor(UniversalRenderPipelineSerializedLight serializedLight, Editor owner)
+        internal static void DrawerColor(UniversalRenderPipelineSerializedLight serializedLight, Editor owner)
```
svens-unity, line 248: This can remain private too?

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.Drawers.cs
```diff
@@ -263,7 +273,7 @@ static void DrawerColor(UniversalRenderPipelineSerializedLight serializedLight,
             }
         }
 
-        static void DrawEmissionContent(UniversalRenderPipelineSerializedLight serializedLight, Editor owner)
+        internal static void DrawEmissionContent(UniversalRenderPipelineSerializedLight serializedLight, Editor owner)
```
svens-unity, line 276: This can remain private too?

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.PresetInspector.cs
```diff
@@ -0,0 +1,33 @@
...
+            CED.Group((serialized, owner) => EditorGUILayout.Space()),
+            CED.FoldoutGroup(Styles.generalHeader,
+                Expandable.General,
+                k_ExpandedStatePreset,
+                UniversalRenderPipelineLightUI.DrawGeneralContentPreset),
+            CED.FoldoutGroup(Styles.emissionHeader,
+                Expandable.Emission,
+                k_ExpandedStatePreset,
+                CED.Group(UniversalRenderPipelineLightUI.DrawerColor,
+                    UniversalRenderPipelineLightUI.DrawEmissionContent)),
```
svens-unity, line 26: `UniversalRenderPipelineLightUI` is unnecessary for the draw calls? Same comments might apply to the Camera changes

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.PresetInspector.cs
jenniferd-unity: yes I agree, this is for consistency with the other partial declarations as suggested in the above conversations.

svens-unity (approved):
It would be good to address these minor things before landing the PR. Especially the change from `internal` to `private`, that was the reason behind moving the `PresetInspector`.

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.PresetInspector.cs
RSlysz: No I mean, the other also don't need internal as it is the default one already applied

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
jenniferd-unity: i really don't understand the benefits of your proposal, the code I have could be searched for and replaced once the API is made public. maybe @RSlysz you could find the right words to explain to me what i fail to understand?
@svens-unity Regarding the rest of the PR, is there anything blocking?

RSlysz: Honestly I dunno if it will ever made public. Preset are such an internal things.

I think what @svens-unity propose is just to be extra explicit here:
```suggestion
            const HideFlags preset = 64 //usefull comment about why HideFlags 64 is not exposed here
                | HideFlags.DontSaveInEditor
                | HideFlags.NotEditable
                | HideFlags.DontSaveInBuild;
            return (editor.target as Component).gameObject.hideFlags == preset;
```
This way we don't lost information like "this is not saved".

svens-unity: It'd improve code readability. The check here in `IsPresetEditor` grabs the GameObject of the targeted component. Then it checks if exactly all of the following flags are enabled:
- HideFlags.HideInHierarchy 
- HideFlags.DontSaveInEditor
- HideFlags.NotEditable
- HideFlags.DontSaveInBuild
- (HideFlags)64;

The suggested change makes this much more explicit. Anyone reading the code will understand it from reading this code right here. Understanding the `93` on the other hand would require the dev to convert it to bits and check in source code what flags are enabled.

svens-unity: As an aside the access modifier and static keyword are in an incorrect order. Also this code could break if a new element is added to HideFlags that is always enabled. 

I did another pass 30 minutes ago.

jenniferd-unity: 93 is not a combination of the flags actually but a magic number (asked Bastien, Preset dev about it). I have added the explanation as a comment for the future. Let me know if this is better?

jenniferd-unity:
@TheoWong-pixel @Verasl Could you confirm the PR is okay with you still?

TheoWong-pixel:
For the HDRP Reflection Probe Preset can we have the info box come after the 'Type' section. its much easier to associate that way.

jenniferd-unity:
> For the HDRP Reflection Probe Preset can we have the info box come after the 'Type' section. its much easier to associate that way.

Sorry @TheoWong-pixel I pulled up the before image instead of after. The image should be updated in a couple of minutes :)

TheoWong-pixel (approved):

com.unity.render-pipelines.universal/Editor/Camera/UniversalRenderPipelineCameraEditor.cs
svens-unity: [HideFlags](https://github.cds.internal.unity3d.com/unity/unity/blob/22f5c73389b2d2d6a06e7d4a795d1757b0198ffa/Runtime/Export/Scripting/UnityEngineObject.bindings.cs#L11) use the [Flags attribute](https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-5.0), each individual flag can be enabled or disabled. In this code you're checking if the `int` representation of those flags is equal to 93. In bits `93` is equal to `1011101`, in other words the 1st, 3rd, 4th, 5th and 7th flag are enabled, all other flags are disabled. Therefore 93 is a combination of the flags I mentioned before.

Hence the other concerns I raised before. If someone introduces a new flag that is always enabled the value jumps from 93 to 221. Furthermore, this combination of flags might not be exclusive to presets? A user or any unity dev could create a GameObject with these HideFlags on a GameObject.

Verasl (approved):
LGTM

RemyUnity:
Well, it was merged anyway, but apart from the missing "Type" field for the lights (not related to this PR), it seems good to me.

PR #5202, opened 2021-07-22
[HDRP] Make custom post process template work and use cmd.Blit by defaukt

alelievr:
### Purpose of this PR

Updated the custom post-processes template to use cmd.Blit instead of HDUtils.DrawFullscreen which is more intuitive for users. 
Also, the custom PP shader template was not working with cmd.Blit because of the missing Property scope in the shader and mismatching source texture name.

I also updated the doc about the usage of cmd.Blit and HDUtils.DrawFullscreen
image

---
### Testing status

Tested that the greyscale example in our doc works as expected with cmd.Blit. Also replaced one of the custom post-process in our test project to use Blit() instead of HDUtils.DrawFullscreen.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fcustom-post-process-blit/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/7800994/job


FrancescoC-unity (approved):
LGTM (note I have not read the doc, leaving that to Vic!)

sebastienlagarde (approved):

Vic-Cooper (approved):
Docs changes done :)

JMargevics (approved):
Things tested: 

Adding a Fullscreen Custom Pass
image

Adding a DrawRenderer Custom Pass + SG with HDSceneColor node
image

No issues in functionality found 👍✅

PR #5204, opened 2021-07-22
[2021.2][ShaderGraph] Add BuiltIn Target variant stripping

jessebarker:
---
### Purpose of this PR
This PR addresses https://fogbugz.unity3d.com/f/cases/1345580/

Currently, the BuiltIn Target does no variant stripping (apart from what the ShaderLab compiler does for us).  For users testing assets against, say, both BuiltIn and URP, this can yield a huge set of variants for compilation.  This code is heavily based on the stripping done by URP, primarily because the BuiltIn Target and some of its shader library are also derived from the URP implementation (though most of the BuiltIn shader library are the same cginc files used by hand-written shaders for the built-in render pipeline).   Some things to note (this should all be documented in comments in the code, but feel free to comment if it isn't clear:
- BuiltIn cannot strip many variants that URP or HDRP can (e.g., there is no "global" rendering path setting for built-in the way there is for URP and HDRP via the asset, so we have to assume that deferred and forward can be used).
- BuiltIn will strip all SRP shader variants only if there is no SRP asset assigned in settings (i.e., we are sure that built-in is the pipeline in use)
- Only non-SRP BuiltIn Shader Graph shader snippet stripping activity is logged to the console, so the console log will not show the full stripping results (see caveat below).

Additional caveat on the logging of variant stripping in general: There are a large number of shader variants that are not logged by the existing URP and HDRP processors (the same will be true for the BuiltIn processor), even though they are seen/processed.  Many of these occur _after_ the ones logged by these processors, which means the final numbers (totals/percentages) in the console log are not the real final numbers.  For example (using the URP Lit Shader Graph mentioned below), filtering so that only the BuiltIn passes are logged makes it appear as though we have included 12.5K out of 295K variants, when the real number is the 13K out of 410K variants mentioned below in the testing status.  This is due to the large number of hidden/legacy/other shaders that are actually included in the project, but don't pass the logging filter.

---
### Testing status

Manual testing according to the repro shadergraph from the bug (result after stripping is a similar number of total variants passed to the compiler as for URP - roughly 7K out of 280K variants, or about 2%).
Manual testing similar to above, but with the proposed new URP Lit Shader Graph (roughly 13K out of 410K variants, or about 3%).
Working on automated tests using the URP Lit Shader Graph.

#### Yamato jobs:
- [BuiltIn Foundation PR on 2021.2](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg-builtin-variant-stripping/.yamato%252Fall-shadergraph_builtin_foundation.yml%2523PR_ShaderGraph_BuiltIn_Foundation_2021.2/7786929/job/pipeline)
- [BuiltIn Lighting PR on 2021.2](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg-builtin-variant-stripping/.yamato%252Fall-shadergraph_builtin_lighting.yml%2523PR_ShaderGraph_BuiltIn_Lighting_2021.2/7786934/job/pipeline)
- [ShaderGraph PR on 2021.2](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg-builtin-variant-stripping/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_2021.2/7786924/job/pipeline)

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -123,6 +123,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed ShaderGraph exception when trying to set a texture to "main texture" [1350573].
 - Fixed a ShaderGraph issue where Float properties in Integer mode would not be cast properly in graph previews [1330302](https://fogbugz.unity3d.com/f/cases/1330302/)
 - Fixed a ShaderGraph issue where hovering over a context block but not its node stack would not bring up the incorrect add menu [1351733](https://fogbugz.unity3d.com/f/cases/1351733/)
+- Fixed the BuiltIn Target to perform shader variant stripping [1345580] (https://issuetracker.unity3d.com/product/unity/issues/guid/1345580/)
 
 ## [11.0.0] - 2020-10-21
 


joshua-davis (approved):

cdxntchou (approved):
Looks good!

PR #5205, opened 2021-07-22
Fix shader warnings when switching platforms

FrancescoC-unity:
When switching platforms there is always _a lot a lot_ (hundreds if not collapsed) of warnings because of macro redefinition in API files. 

The solution to this is to undefine the macros that will be shortly redefined. This way no shader warnings are there anymore. 

Yamato: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Ffix-warning-on-platform-change/.yamato%252F_abv.yml%2523all_project_ci_2021.2

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
@@ -168,6 +168,8 @@
 #define DEPTH_OFFSET_SEMANTIC SV_Depth
 
 // Include language header
+// Before declaring macros we need to undef some of them to avoid warnings.
+// Said warnings are happening as these macros are included in HLSLSupport.cginc and API files re-define them.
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

sebastienlagarde:
@FrancescoC-unity do you know if it is a new thing ? Poking @aleks01010101 to be sure we don't do a mistake here :)

aleks01010101 (commented):
Why do you need to include HLSLSupport.cginc in the first place?
Also, it sounds a bit odd to me that this fixes shader warnings when switching platforms.

FrancescoC-unity:
> Why do you need to include HLSLSupport.cginc in the first place?
> Also, it sounds a bit odd to me that this fixes shader warnings when switching platforms.

This because something platform side includes said file, when switching platform to said platform, the shaders are compiled with that file and the problem ensues.

FrancescoC-unity:
We agreed with @aleks01010101 to move this to the platform specific packages. Closing this and moving this there

PR #5207, opened 2021-07-22
Fix artifacts in billboard LODs due to LOD smoothing/crossfade behavior

nerites:

### Checklist for PR maker
- [n/a] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [n/a] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Fixes this bug: https://fogbugz.unity3d.com/f/cases/1348407/
A bug was previously filed for this issue, but was mistakenly closed despite the initial PR for that bug not addressing the issue: https://fogbugz.unity3d.com/f/cases/1198135/

With this fix, the logic for the handwritten shader now matches the ST8 shadergraph.

---
### Testing status
Confirmed the following:

- [x] When LOD Fade mode is set to SpeedTree, branches interpolate as expected without popping between non-billboard LODs.
speedtreelods
speedtreenocrossfade

- [x] When Animate Crossfade is enabled and LOD Fade mode is set to SpeedTree, crossfade dithering does not scale up due to interpolation
speedtreecrossfade

- [x] When Animate Crossfade is disabled and LOD Fade mode is set to Crossfade, behavior matches LOD Fade mode = none. (see popping between LODs)
speedtreenoanim

---
### Comments to reviewers
Here are the two bugs related to this issue:
Popping between non-billboard LODs:
lodsmoothpopping

Incorrect scaling on crossfade to billboard:
speedtree_lod_change

This should be covered with a graphics test, but since the effect is transient, it would take some thinking to set up a test to capture it properly. 
Since this bug has been around for so long, I'd rather just merge the fix quickly and add test coverage to the test audit task we have in our backlog.


LandonTownsendUnity:
can you post gifs of the LOD blending for all combinations of "fade mode" options, and having "Animate Cross-Fading" checked or unchecked?

LandonTownsendUnity (approved):

russoh (approved):
Does this now match the Shader Graph versions?

nerites:
> 
> 
> Does this now match the Shader Graph versions?

No: the shader graph versions perform the SpeedTree vertex lerp even if "Crossfade" or "None" is selected as the LOD blend mode. So this PR has the correct behavior and the Shader Graphs do not. This is due to the default keyword limitation on shader graph. In the future, if we add a target postprocessor for speedtree, we could work around this by adding the keywords SpeedTree8 uses on the passes directly, circumventing this limit. For now I think this discrepancy in behavior is acceptable.

phi-lira (approved):

hdb-unity (approved):
Nice PR description here - thanks for the animated gifs!

Did not test locally. Approving for URP QA based on Landon's approval / dev testing. 

I will reach out to ask about the automation plan - this change should be protected by some tests.

Does this change require any documentation updates?

nerites:
This change should not require any documentation updates since it matches the intended behavior of the LOD blend modes. 

For test coverage, ideally we'd have a scene with 6 sets of 3 different speedtree instances, each configured with different LOD transitions such that as the camera is being moved for the test, one tree transitions from LOD0 to LOD1, another transitions from LOD1 to LOD2, and the third transitions from LOD2 to LOD3/Billboard. 
Each set would correspond to: 
1. Blend Mode = None, Animate Crossfade = Off
2. Blend Mode = None, Animate Crossfade = On
3. Blend Mode = Crossfade, Animate Crossfade = Off
4. Blend Mode = Crossfade, Animate Crossfade = On
5. Blend Mode = SpeedTree, Animate Crossfade = Off
6. Blend Mode = SpeedTree, Animate Crossfade = On

nerites:
Hmm. I guess another bug was the incorrect interpolation that was occuring between the transition points. So we'd want to add 2 more speedtrees per set, with LOD distances configured to test passing through LOD1 (interpolating LOD0->LOD2) and LOD2 (interpolating LOD1->LOD3).

We should probably run this test plan by Greg Croft to confirm that the behavior in Unity is what's intended.

PR #5209, opened 2021-07-23
[HDRP] Add a built-in custom pass to draw object IDs

pmavridis:
### Purpose of this PR
Add a built-in custom pass to draw object IDs. The users can select it from the menu with the other built-in custom passes:
image

The interface is similar to the DrawRenderers custom pass, but it does not allow you to averride the material (because the material is fixed to the objectID shader).
image

---
### Testing status
- The "AOVbuffers2" test has been updated to use the built-in custom pass to output the object IDs and the output looks correct.


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/CustomPass/DrawRenderersCustomPassDrawer.cs
```diff
@@ -226,40 +228,45 @@ void DoFilters(ref Rect rect)
         void DoMaterialOverride(ref Rect rect)
         {
             //Override material
-            EditorGUI.BeginChangeCheck();
-            EditorGUI.PropertyField(rect, m_OverrideMaterial, Styles.overrideMaterial);
-            if (EditorGUI.EndChangeCheck())
+            if (showMaterialOverride)
```
pmavridis, line 231: @alelievr I just added an if statement here, sorry for the mess in the diff ;)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/ObjectIDCustomPass.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            int index = 0;
+            foreach (Renderer renderer in rendererList)
+            {
+                MaterialPropertyBlock propertyBlock = new MaterialPropertyBlock();
+                float hue = (float)index / rendererList.Length;
+                propertyBlock.SetColor("ObjectColor", Color.HSVToRGB(hue, 0.7f, 1.0f));
+                renderer.SetPropertyBlock(propertyBlock);
+                index++;
+            }
```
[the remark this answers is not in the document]
pmavridis: Yeah, that's a good idea. Done!

alelievr (approved):

sebastienlagarde (approved):

sebastienlagarde:
@pmavridis  please update documentation as well to explain there is this pass. thanks

pmavridis:
Adding Vic for the docs.

iM0ve (approved):
I have tested that the new ID custom pass works on mac

Screenshot 2021-08-02 at 14 41 37


Reported an issue not caused by this pr - unable to remove items from custom pass volume
https://fogbugz.unity3d.com/f/cases/1354871/

com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass-Creating.md
```diff
@@ -161,6 +161,11 @@ When Unity renders a material that is not supported by the current injection poi
 
 <a name="Custom-Renderers-Pass-shader"></a>
```
Vic-Cooper, line 162: Move this title tag to just above the ## Custom Renderers Pass shader header.

com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass-Creating.md
```diff
@@ -161,6 +161,11 @@ When Unity renders a material that is not supported by the current injection poi
...
 
+## Object ID Custom Pass
+HDRP includes a specialization of a Draw Renderers Custom Pass that draws the objects in the scene with a unique color (Object ID). This pass has the same parameters as the Draw Renderers Custom Pass except the material override option, which is hidden because this pass uses a built-in material that draws the Object ID colors. The Object IDs are generated before the first execution of the pass occurs. If a user script generates new objects procedurally, then the method AssignObjectIDs should be called manually, to generate new IDs.
+
+If some customization is required to the way Object IDs are generated, then we recommend writing a new Draw Renderers Custom Pass with a Custom Renderers Pass shader that provides the required functionality. You can find more details regarding Custom Renderers Pass shaders in the next section.
+
```
Vic-Cooper, line 168: - Could we have a short example of the `AssignObjectIDs` implementation to generate new Object ID's here? 
- An example of a Custom Renderers Pass that changes how HDRP generates Object IDs would be helpful too!

Vic-Cooper: Current re-write: 

The **Object ID Custom Pass** draws GameObjects in the scene with a unique color (`Object ID`). 

This pass has the same parameters as the [Draw Renderers Custom Pass](#Draw-Renderers-Custom-Pass) except the **Material** property is hidden because this pass uses a built-in Material to draw the Object ID colors. 

HDRP generates Object IDs before executing this custom pass. If you use a script to generate new GameObjects procedurally, call `AssignObjectIDs`  to generate new Object IDs.

To customize the way Unity generates Object IDs, write a new Draw Renderers Custom Pass using a [Custom Renderers Pass](#Custom-Renderers-Pass-shader) shader.

Vic-Cooper (changes requested):
Reviewed the new Object ID custom pass section. I've made some suggestions. We can push it as is and make these changes in the future, or if there's no urgency with this PR, I'm happy to work it out here.

com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass-Creating.md
pmavridis: Thanks for the feedback Vic, I will shortly update the PR based on the suggestions and I will also add the missing examples.

Vic-Cooper:
Updated the docs. I feel like the Object ID behavior info should be captured in more detail somewhere else but that might not be one for this PR :D I'll add that to my expanding list of custom pass docs improvement to-do's.

Vic-Cooper (commented):
Pushed changes, please review for accuracy. If you're happy, I'll give the green dot :)

com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass-Creating.md
```diff
@@ -159,6 +159,83 @@ However, not all materials are supported by every injection point in a draw rend
...
+|Write Depth                     |Enable **Write Depth** to instruct Unity to write depth values for GameObjects that use this material. <br/>Disable it if you do not want Unity to write depth values for each GameObject.                                                              |
+|  Sorting                     | Select the way Unity sorts the GameObjects in your scene before it renders them.<br/><br/>For more information, see [Sorting criteria](https://docs.unity3d.com/ScriptReference/Rendering.SortingCriteria.html).                                          |
+
+### Generating Object IDs for a custom pass
+
+HDRP generates Object IDs before executing an Object ID custom pass.
+
+If you use a script to generate new GameObjects procedurally, call `AssignObjectIDs` to generate new Object IDs.
+
+The following script example uses the default `AssignObjectIDs`  implementation that assigns Object ID colors incrementally:
```
pmavridis, line 195: Small suggestion: 
"The following script example **uses** the default" ---> "The following script example **shows** the default"

The script shows the default implementation of the AssignObjectIDs method (so users know what to expect), it's not "using" this method.

Vic-Cooper: Thanks for this clarification, I'll make the change :)

PR #5211, opened 2021-07-23
Documentation pass on Custom Post Process supporting DRS / DLSS

kecho:
### Purpose of this PR
Documentation on custom post process so they can be compatible with DRS and DLSS.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Custom-Post-Process.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Custom-Post-Process.md
@@ -163,8 +163,6 @@ Shader "Hidden/Shader/GrayScale"
     {
         UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
-        // NOTE: As of 12.x beta, for new custom post-processes to be compatible with Dynamic resolution, please use _PostProcessScreenSize.xy as the reference size of our source textures.
-        // You must use this size for color / normal and velocity. Otherwise, your custom post-process won't work when dynamic resolution or DLSS is enabled.
         float3 sourceColor = SAMPLE_TEXTURE2D_X(_MainTex, s_linear_clamp_sampler, input.texcoord).xyz;
 
         // Apply greyscale effect
--- a/com.unity.render-pipelines.high-definition/Editor/PostProcessing/Templates/CustomPostProcessingShader.template
+++ b/com.unity.render-pipelines.high-definition/Editor/PostProcessing/Templates/CustomPostProcessingShader.template
@@ -48,7 +48,7 @@ Shader "Hidden/Shader/#SCRIPTNAME#"
     {
         UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
-        // Note that if HDUtils.DrawFullScreen is used to render the post process, input.texcoord needs to be multiplied by _RTHandleScale.xy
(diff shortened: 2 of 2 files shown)


com.unity.render-pipelines.high-definition/Documentation~/Custom-Post-Process.md
```diff
@@ -151,6 +151,11 @@ Shader "Hidden/Shader/GrayScale"
         UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
         uint2 positionSS = input.texcoord * _ScreenSize.xy;
+
+        //NOTE: As of 12.x beta, for new custom post processes to be compatible with Dynamic resolution, please use:
```
alelievr, line 155: I think we need to put that in the custom post process template too (file: `CustomPostProcessingShader.template`)

com.unity.render-pipelines.high-definition/Documentation~/Custom-Post-Process.md
```diff
@@ -274,6 +279,20 @@ sealed class GrayScaleEditor : VolumeComponentEditor
 ```
 This custom editor is not really useful as it produces the same result as the editor that Unity creates. Custom Volume component editors also support an [additonal properties toggle](More-Options.md). To add it, you have to set hasAdvancedMode override to true. Then, inside the OnInspectorGUI, you can use the isInAdvancedMode boolean to display more properties.
 
+## Dynamic resolution and DLSS support
+
+If you are planning to fully take advantage of DLSS and / or dynamic resolution on your pass, and you require to interpolate / uv sample either color / normals or velocity, you must use the following functions to calculate the correct uvs:
```
Vic-Cooper, line 284: Phrasing change: 

If you want to use DLSS and/or dynamic resolution on your pass, and you need to interpolate or sample UVs from color / normal or velocity, you must use the following functions to calculate the correct UVs:

com.unity.render-pipelines.high-definition/Documentation~/Custom-Post-Process.md
```diff
@@ -151,6 +151,11 @@ Shader "Hidden/Shader/GrayScale"
         UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
         uint2 positionSS = input.texcoord * _ScreenSize.xy;
+
+        //NOTE: As of 12.x beta, for new custom post processes to be compatible with Dynamic resolution, please use:
+        //uint2 positionSS = input.texcoord * _PostProcessScreenSize.xy;
+        // This position must be used ultimately for color / normal and velocity. If not, your custom post process won't work when dynamic resolution / DLSS is enabled.
```
Vic-Cooper, line 157: Small adjustment: 

You must use this position for color / normal and velocity. Otherwise, your custom post-process won't work when dynamic resolution or DLSS is enabled.

com.unity.render-pipelines.high-definition/Documentation~/Custom-Post-Process.md
Vic-Cooper: Also, please hyphenate post-processes :D:

//NOTE: As of 12.x beta, for new custom post-processes to be compatible with Dynamic resolution, please use:

Vic-Cooper (changes requested):
Made some docs suggestions.

com.unity.render-pipelines.high-definition/Documentation~/Custom-Post-Process.md
kecho: > I think we need to put that in the custom post process template too (file: `CustomPostProcessingShader.template`)

Updated the template file. Good catch!

alelievr (approved):

PR #5218, opened 2021-07-26
catch thrown errors in PPv2 Render (case 1221972)

fredericv-unity3d:
### Purpose of this PR
Fixes [1221972](https://fogbugz.unity3d.com/f/cases/1221972/)

**Investigation:**
In PPv2 when an error occurs, it is thrown as an exception, although, there are no try/catch to handle those errors.

During the `PostProcessLayer.OnPreCull` call, GPU resources are modified or allocated. But if an exception occurs during it, those resources are not cleaned up. resulting in a GPU crash.

**Workaround:**
`PostProcessEffectRenderer.Render` is the offending function here, it is the place where user can provide code and have to call the `PropertySheet.Get` function (which does the throw on invalid shader).

Other places seems to not trigger the issue, so I only updated all call sites of `PostProcessEffectRenderer.Render`, so places where the user can potentially throws exception in the framework.
(Basically, I added a try/catch statement and logged any thrown Exception)

**Notes and suggestion:**
This is a technical debt as the error management paradigm is not suited for realtime application.
Using try/catch statements can be costly, and for some platforms the exception flow can be disabled (like for iOS).

I recommend to change the error management to a more C-style error management for critical section (where and exception is returned, not thrown) and error are handled whenever they occurs.
(Basically, don't use the exception flow, instead return Exception and caller must check for errors)

---
### Testing status
Manual test:
1. Without using the workaround
2. Load the user project
3. Start play mode
4. > Observe a crash around 3000 error logs

1. Use the workaround
2. Load the user project
3. Start play mode
4. > No crash observed by 50 000 error logs

---

### DIFF
--- a/com.unity.postprocessing/CHANGELOG.md
+++ b/com.unity.postprocessing/CHANGELOG.md
@@ -4,15 +4,10 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [3.1.3] -
+## [3.1.2] - 2021-02-06
 
 ### Fixed
 - Catch thrown errors in `PostProcessEffectRenderer`, preventing resources allocation leaks and crash (case 1221972)
-
-## [3.1.2] - 2021-02-06
-
-Version Updated
-The version number for this package has increased due to a version update of a related graphics package.
 
 ## [3.1.1] - 2021-03-08
 


sebastienlagarde (approved):

sebastienlagarde (approved):

sebastienlagarde:
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Fbugfix_case_1221972_ppv2_crash/.yamato%252Fall-postprocessing.yml%2523PR_PostProcessing_2021.2/7871053/job

RemyUnity (approved):
LGTM ! For sanity check I've also reopened my testing project with the fix and the crash doesn't occur anymore.

PR #5219, opened 2021-07-26
[HDRP] Fix APV UI loosing focus when changing a param

alelievr:
### Purpose of this PR
Fixed the APV UI loosing focus when the helpbox about baking appears in the probe volume.
LooseFocus

---
### Testing status
Verified that the UI don't loose focus anymore + checked that the helpbox size is calculated correctly when resizing the editor.


sebastienlagarde (approved):

FrancescoC-unity (approved):

JMargevics (approved):
Testing coverage seems sufficient, the style fix is pretty simple 👍 Approving ✅

PR #5224, opened 2021-07-26
Fixed for enabling a lensflare in playmode

adrien-de-tocqueville:
### Purpose of this PR
fix https://fogbugz.unity3d.com/f/cases/1352771/

The bug was only happening in the editor because OnValidate would get executed when entering playmode.
Now it will only registers the flare if the component is active

---
### Testing status

Tested on the repro project in the editor and in a player build


JMargevics (approved):
Testing on the repro project is enough. Approving ✅👍

sebastienlagarde (approved):

PR #5225, opened 2021-07-26
Added disabled Projected Space Geometric Normal Filtering

adrien-de-tocqueville:
### Purpose of this PR

Implemented geometric normal filtering method described here https://yusuketokuyoshi.com/papers/2021/Tokuyoshi2021SAA.pdf
The code is currently under a defined so it is not accessible directly for users

---
### Testing status

You can open the images in different tabs for easier comparison, but there isn't any improvement that would justify changing the method

- Current HDRP
previous
- This PR
new


sebastienlagarde (approved):

sebastienlagarde:
Note for QA: there is no need for testing, we add the code for completeness purpose but it is disable by default.

PR #5229, opened 2021-07-26
[ShaderGraph][21.2] Improve node creation context menu option in stacks

dhsavell:
### Purpose of this PR
FogBugz case: https://fogbugz.unity3d.com/f/cases/1307853/. Previously, the "Create Node" context menu item was difficult to access because it only showed up when right-clicking on a separator. I discussed the issue with @alindmanUnity, and they suggested the following improvements which I've implemented here.

- Rename "Create Node" to "Add Block Node"
- Make the "Add Block Node" option also accessible when right-clicking on the stack node body, not just the separators (right-clicking on a node should still bring up its original context menu)

Before (click for full resolution):
image

After:
image

---
### Testing status

- [x] Add Block Node
  - [x] using space bar
    - Space bar add behavior was unaffected by this PR and still functions as before
  - [x] from separator context menu
    - Item has the same behavior as before: adds the selected block node at the bottom of the stack or does nothing if it is already present
  - [x] from stack context menu
    - Adding a node from the new stack context menu adds to the bottom of the stack or does nothing if present, matching separator behavior
  - [x] to empty stack
    - The new "Add Block Node" option is both present and functional when right clicking anywhere on an empty stack. Space still works as well
    image
- [x] Disconnect All option still works
  - The "Disconnect All" context menu item is still present only when right-clicking on a block node, and is enabled and functional when that node has incoming connections
- [x] Undo/redo, other pipelines, subgraphs - not affected by this PR, tested and behavior matches master (tried undoing/redoing added blocks, tried adding nodes in other pipelines, no stacks in subgraphs).
 
---
### Comments to reviewers

Are there any block nodes that have their own context menus?


jessebarker (approved):
LGTM.  Looks friendlier based on the screenshots.

ghost (approved):
Developer manual testing looks good!

PR #5231, opened 2021-07-26
Improving DLSS cpu load, and not recreating DLSS state every change of res

kecho:
### Purpose of this PR

DLSS was recreating state every change of resolution for DRS. This was a bit of an oversight, caused by not checking the min and max resolution that DLSS suggest
Turns out that we should always query for the min and max resolution regardless if auto parameters are enabeld or not. This range is the only range were we are allowed to utilize hot DRS scaling. Otherwise, if we request outside this range, we must in turn recrete the DRS feature state.

The non technical summary: we optimized CPU usage and straining of system. DRS should run much smoother.

Fixed also a problem were if the viewport is odd size of pixels, the DRS scaler would return a 1 pixel extra. So to fix this, clamped the actualWidth/actualHeight post resolution.

---
### Testing status
* Tested on vulkan and dx11 using the test scene from the bug.
* Tried to repro crash, could not. So ended up finding this issue and fixing it.
* QA Must do a full test of DLSS: dx12, retest vulkan and do this with software and hardware drs. Check on CPU usage.

---
### Comments to reviewers
I renamed some functions but the jist of the change is that we now call GetOptimalSettings every frame, use the safe ranges for DRS, and optionally use the sharpness / autoscaler on the setting.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
+++ b/com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
@@ -502,9 +502,6 @@ namespace UnityEngine.Rendering
 
             Vector2Int scaledSize = ApplyScalesOnSize(size);
 
-            scaledSize.x = Math.Min(scaledSize.x, size.x);
-            scaledSize.y = Math.Min(scaledSize.y, size.y);
-
             m_LastScaledSize = scaledSize;
             return scaledSize;
         }
@@ -529,6 +526,9 @@ namespace UnityEngine.Rendering
                 scaledSize.x += (1 & scaledSize.x);
                 scaledSize.y += (1 & scaledSize.y);
             }
+
+            scaledSize.x = Math.Min(scaledSize.x, size.x);
+            scaledSize.y = Math.Min(scaledSize.y, size.y);
 
             return scaledSize;
         }
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/DLSSPass.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/DLSSPass.cs
@@ -351,7 +351,6 @@ namespace UnityEngine.Rendering.HighDefinition
 
             public void UpdateViewState(
                 in DlssViewData viewData,
-                /*uint backbufferWidth, uint backbufferHeight,*/
                 CommandBuffer cmdBuffer)
             {
                 bool shouldUseOptimalSettings = ShouldUseAutomaticSettings();


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -902,8 +902,9 @@ internal void Update(FrameSettings currentFrameSettings, HDRenderPipeline hdrp,
             if (canDoDynamicResolution)
             {
                 Vector2Int scaledSize = DynamicResolutionHandler.instance.GetScaledSize(new Vector2Int(actualWidth, actualHeight));
-                actualWidth = scaledSize.x;
-                actualHeight = scaledSize.y;
+                //ensure that the scale is within and we dont round up an extra pixel
+                actualWidth = Math.Min(scaledSize.x, actualWidth);
```
FrancescoC-unity, line 906: Feels to me this should be done inside DynamicResolutionHandler.instance.GetScaledSize itself, I cannot think of a scenario in which we want to exceed. 

Moreover if at some point we modify the dynamic resolution to scale _up_ as well, it'd be good to have it self contained in the handler.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/DLSSPass.cs
```diff
@@ -333,18 +339,20 @@ private bool ShouldUseOptimalSettings()
 
             public void UpdateViewState(
                 in DlssViewData viewData,
+                /*uint backbufferWidth, uint backbufferHeight,*/
```
FrancescoC-unity, line 342: Leftover comment?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/DLSSPass.cs
```diff
@@ -333,18 +339,20 @@ private bool ShouldUseOptimalSettings()
...
                 CommandBuffer cmdBuffer)
             {
-                bool shouldUseOptimalSettings = ShouldUseOptimalSettings();
+                bool shouldUseOptimalSettings = ShouldUseAutomaticSettings();
                 bool isNew = false;
                 if (viewData.outputRes != m_Data.outputRes ||
-                    (viewData.inputRes != m_Data.inputRes && !m_UsingOptimalSettings) ||
-                    (viewData.CanFitInput(m_Data.inputRes) && m_UsingOptimalSettings) ||
+                    (viewData.inputRes.width > m_BackbufferRes.width || viewData.inputRes.height > m_BackbufferRes.height) ||
+                    (viewData.inputRes != m_BackbufferRes && !m_OptimalSettingsRequest.CanFit(viewData.inputRes)) ||
```
FrancescoC-unity, line 349: Is it fine to do this checks (m_OptimalSettingsRequest.CanFit) when optimal settings are not used?

FrancescoC-unity:
> This range is the only range were we are allowed to utilize hot DRS scaling. Otherwise, if we request outside this range, we 
> must in turn recrete the DRS feature state.

For my curiosity: How inclusive are those "optimal settings"? Is there a threshold in which we thrash back and forth the recreation?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/DLSSPass.cs
kecho: yep!

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
kecho: yeah its possible since the drs handler is aware of the viewport size.
Just need to be more careful with this though since now we are making this function dependant on a final viewport size and test a bit more.

FrancescoC-unity: Oh no I mean clamping to the thing that is passed to GetScaledSize (in this case is actualWidth/Height)

kecho: ah! right. Yes this should be easy

kecho:
> > This range is the only range were we are allowed to utilize hot DRS scaling. Otherwise, if we request outside this range, we
> > must in turn recrete the DRS feature state.
> 
> For my curiosity: How inclusive are those "optimal settings"? Is there a threshold in which we thrash back and forth the recreation?
Depending on the quality, it can go down to 20% (for the high performance settings) up to 50% (for the high quality settings).
But is usually between 20% or 50%. Anything below they force you to recreate the state. If you dont, then you get internal exceptions in the library.

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -501,6 +501,10 @@ public Vector2Int GetScaledSize(Vector2Int size)
             }
 
             Vector2Int scaledSize = ApplyScalesOnSize(size);
+
+            scaledSize.x = Math.Min(scaledSize.x, size.x);
```
FrancescoC-unity, line 505: I think we should do this in ApplyScalesOnSize though as that is public API too, specifically this variant :

> Vector2Int ApplyScalesOnSize(Vector2Int size, Vector2 scales)

remi-chapelain (commented):
So, for now, I quickly tested and can still repro the crash. 
I've narrowed the fact that it's DX12 related (doesn't happen in vulkan or dx11). 
@kecho managed to repro it so hopefully we'll be able to find the cause. 

in the meantime, there's a crash folder zipped, if needed. 
[Crash_2021-07-27_151018460.zip](https://github.com/Unity-Technologies/Graphics/files/6886639/Crash_2021-07-27_151018460.zip)

kecho:
> So, for now, I quickly tested and can still repro the crash.
> I've narrowed the fact that it's DX12 related (doesn't happen in vulkan or dx11).
> @kecho managed to repro it so hopefully we'll be able to find the cause.
> 
> in the meantime, there's a crash folder zipped, if needed.
> [Crash_2021-07-27_151018460.zip](https://github.com/Unity-Technologies/Graphics/files/6886639/Crash_2021-07-27_151018460.zip)

@remi-chapelain for now I have removed the fogbugz on this PR, and instead keep this PR as an optimization. So as long as DLSS is running well on this PR we should still pursue it. 
I have another PR in ono with second attempt fix for the crash (fix in c++)

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -421,8 +421,10 @@ public void Update(GlobalDynamicResolutionSettings settings, Action OnResolution
             if (!m_ForceSoftwareFallback && type == DynamicResolutionType.Hardware)
             {
                 hardwareResolutionChanged = FlushScalableBufferManagerState();
-                if (ScalableBufferManager.widthScaleFactor != m_PrevHWScaleWidth ||
-                    ScalableBufferManager.heightScaleFactor != m_PrevHWScaleHeight)
+                float sfw = ScalableBufferManager.widthScaleFactor;
```
FrancescoC-unity, line 424: This change is a no-op right? I don't see the advantage as I tend to like more verbose naming vs acronyms, but am not opposed. However, if it is for clarity let me be an anal nitpicker and either we have _sfw_ and _sfh_ **or**  _wsf_ and _hsf_, now is a bit of a mixup of names.,

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
```diff
@@ -391,7 +391,7 @@ public void Update(GlobalDynamicResolutionSettings settings, Action OnResolution
         {
             ProcessSettings(settings);
 
-            if (!m_Enabled && !s_ActiveInstanceDirty)
+            if (!m_Enabled || !s_ActiveInstanceDirty)
```
FrancescoC-unity, line 394: I am not sure what this change is doing? 

So s_ActiveInstanceDirty is determining whether we changed anything (is set to false at the end of this), so essentially this is going to call the update every 2 frames instead of every frame which of course is not what we want to do. 
Am I missing something?

kecho: This should have been an or conditional.
Basically dont run this function if we already called it in the frame. 
The reason is that we have 2 loops for cameras within the same frame. 
Each loop calls UpdateAndUseCamera. So a camera calls Update() twice. But we only need to flush all the settings once per frame.

ClearSelectedCamera gets called at the end of the frame and sets this dirty flag to true.

THe problem is that this was false, it should be true. This fixes some crazy artifacting when resizing Hardware resolution scaler.

FrancescoC-unity: Gotcha! I missed the 

> ClearSelectedCamera gets called at the end of the frame and sets this dirty flag to true.

part. 

Makes all sense.

com.unity.render-pipelines.core/Runtime/Common/DynamicResolutionHandler.cs
kecho: This is left over on my side! i was debugging stuff. Good catch!

FrancescoC-unity:
@kecho  still missing https://github.com/Unity-Technologies/Graphics/pull/5231/files#r677527579 and https://github.com/Unity-Technologies/Graphics/pull/5231/files#diff-e967f213a4a311774292867c47260bc3ca8c133c098e42f023b0e0809e55d318R342

remi-chapelain (approved):
This doesn't fix the crash (reported [here](https://fogbugz.unity3d.com/f/cases/1352848/)) but after some quick tests on the template, it seems to improve slightly performance around 10% give or take. 

I also tested on Vulkan, DX11 and DX12 to be sure that it didn't cause any issues there.

kecho:
Indeed. The crash will be fixed at a later point. I will do a debug session with @remi-chapelain since we can only repro it on his mahcine. This however is a good change for performance improvements.

kecho:
> @kecho still missing https://github.com/Unity-Technologies/Graphics/pull/5231/files#r677527579 and https://github.com/Unity-Technologies/Graphics/pull/5231/files#diff-e967f213a4a311774292867c47260bc3ca8c133c098e42f023b0e0809e55d318R342
> 
> rest LGTM
@FrancescoC-unity done!

FrancescoC-unity (approved):

remi-chapelain (approved):
In addition to the slight improvement, this PR fixes this issue ([#1354184](https://fogbugz.unity3d.com/f/cases/1354184/))

sebastienlagarde (approved):

PR #5232, opened 2021-07-26
[Fogbugz # 1352093] Fixing bad copy texture when DRS is on with ray tracing.

kecho:
### Purpose of this PR
Fogbugz: https://fogbugz.unity3d.com/f/cases/1352093/
Copy was failing because the mip texture size computed was off by 1. Made it conservative by using a ceiling instead of casting down a float to an int.

---
### Testing status
* Tested repro case from fogbugz. Windows with hardware DRS on dx11.
* QA should test this with many DRS changes on hardware mode.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
```diff
@@ -642,7 +642,8 @@ public void ComputePackedMipChainInfo(Vector2Int viewportSize)
                 }
                 while ((mipSize.x > 1) || (mipSize.y > 1));
 
-                textureSize = new Vector2Int((int)((float)hardwareTextureSize.x * textureScale.x), (int)((float)hardwareTextureSize.y * textureScale.y));
+                textureSize = new Vector2Int(
```
FrancescoC-unity, line 645: IIRC there is a disparity of rounding behaviour between HW and SW, have you checked both with a collection of resolution? (odd resolutions will be the problem I think). 

If you didnt I guess that's a ping for QA

sebastienlagarde (commented):
please target master

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
kecho: It should work since that computation is what writes on acualWidth and actualHeight.
Will indeed ask QA to test for this.

kecho:
> please target master

Retargetting and rerunning yamato.
Adding QA as well

FrancescoC-unity (approved):
LGTM, pending the QA check I mentioned in the other comment, but code looks good

sebastienlagarde (approved):

remi-chapelain (commented):
Found a bunch of DRS issues ([1354180](https://fogbugz.unity3d.com/f/cases/1354180/), [1354181](https://fogbugz.unity3d.com/f/cases/1354181/), [1354184](https://fogbugz.unity3d.com/f/cases/1354184/)) while testing, but all of theses seems to repro on current master so the PR didn't created that. 

However, since hardware mode is kind of broken right now cf [1354184](https://fogbugz.unity3d.com/f/cases/1354184/), I'm not sure I should approve even though I wasn't able to repro the console "bad copy texture" testing on all the filters (including DLSS) and a lot of different resolutions (classic and odd).

FrancescoC-unity:
@remi-chapelain  I closed 1354180 as Lanczos is about to get removed altogether :-) 

I wonder if 1354184 is the one that @kecho  fixed in another open PR (https://github.com/Unity-Technologies/Graphics/pull/5231)

remi-chapelain:
I'll have a quick look to 1354184 with the other PR, to be sure !

remi-chapelain (approved):
Okay, so indeed, I tried this small change on top of [this PR](https://github.com/Unity-Technologies/Graphics/pull/5231) to be able to test hardware mode. ✔️ 

And I confirm that [this PR](https://github.com/Unity-Technologies/Graphics/pull/5231) indeed fixes the [1354184](https://fogbugz.unity3d.com/f/cases/1354184/). I resolved it.

PR #5014, opened 2021-06-30
[2021.2] Quality Audit: URP Asset facelift

Jonasmortensen:
This PR adds a collection of UI improvements to the Universal Render Pipeline Asset as part of the Quality Audit

- `General` section has been renamed to `Rendering`
- Settings in `Advanced` section have been distributed among existing sections, and hidden behind `Show Additional Properties` setting:
  - SRP Batcher -> Rendering
  - Dynamic Batching -> Rendering
  - Mixed Lighting -> Lighting
  - Light Layers -> Lighting
  - Debug Level -> Rendering
  - Shader Variant Log Level -> Rendering
  - Store Actions -> Rendering
  - Volume Update Mode -< Post-processing
- Replaced `EditorGUILayout.BeginFoldoutHeaderGroup` with `CoreEditorUtils.DrawHeaderFoldout` for consistency
- `Shader Variant Log Level` option `Only Universal RP Shaders` renamed to `Only URP Shaders`
- The default names for newly created pipeline assets and renderers now include spaces

Current Pipeline Asset Inspector:
image
New Pipeline Asset Inspector (All additional properties visible)
image
image

PR #5234, opened 2021-07-27
Quality audit - UX Improvements on URP Asset Inspector

alex-vazquez:
---
### Purpose of this PR

[JIRA](https://jira.unity3d.com/browse/XPIPELINE-244)

- `General` section has been renamed to `Rendering`
- Replaced `EditorGUILayout.BeginFoldoutHeaderGroup` with `CoreEditorDrawer` classes.
- `Shader Variant Log Level` option `OnlyUniversalRPShaders` renamed to `OnlyURPShaders`
- The default names for newly created pipeline assets and renderers now include spaces.
- Forcing a `Repaint` when the highlight is being updated., as the animation was not correctly rendered.
- Settings in `Advanced` section have been distributed among existing sections, and hidden behind `Show Additional Properties settings`:

1. SRP Batcher -> Rendering
2. Dynamic Batching -> Rendering
3. Debug Level -> Rendering
4. Shader Variant Log Level -> Rendering
5. Store Actions -> Rendering
6. Mixed Lighting -> Lighting
7. Light Layers -> Lighting
8. Volume Update Mode -> Post-processing

Some screenshoots, with/out additional properties:

image
image
image
image
image
image
image
image
image

---
### Testing status
- Check the additional data works as expected
- Check old inspector and new inspector and that the values are correctly changed

---
### Comments to reviewers
- Changed to CoreEditorDrawers as it was need by the additional properties.
- @Unity-Technologies/gfx-qa-urp, as this has been changed to the new system, needs a full QA pass on the inspector.
- @Jonasmortensen I've already included your changes on: https://github.com/Unity-Technologies/Graphics/pull/5014

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
@@ -83,10 +83,10 @@ namespace UnityEngine.Rendering.Universal
     public enum ShaderVariantLogLevel
     {
         Disabled,
-        OnlyURPShaders,
-        AllShaders,
-        [Obsolete("OnlyUniversalRPShaders is deprecated, use OnlyURPShaders instead (UnityUpgradable) -> OnlyURPShaders", true)]
+        [InspectorName("Only URP Shaders")]
         OnlyUniversalRPShaders,
+        [InspectorName("All Shaders")]
+        AllShaders
     }
 
     [Obsolete("PipelineDebugLevel is unused and has no effect.", false)]


com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
```diff
@@ -158,7 +158,13 @@ void IDrawer.Draw(TData data, Editor owner)
                     m_ActionDrawers[i](data, owner);
 
                 if (m_Anim != null)
+                {
                     CoreEditorUtils.EndAdditionalPropertiesHighlight();
+
+                    // While the highlight is being changed, force the Repaint of the editor
+                    if (m_Anim.value > -Mathf.Epsilon)
```
jenniferd-unity: we dont have something like "isZero-ish" in C#?

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
```diff
@@ -158,7 +158,13 @@ void IDrawer.Draw(TData data, Editor owner)
                     m_ActionDrawers[i](data, owner);
 
                 if (m_Anim != null)
+                {
                     CoreEditorUtils.EndAdditionalPropertiesHighlight();
+
+                    // While the highlight is being changed, force the Repaint of the editor
+                    if (m_Anim.value > Mathf.Epsilon)
```
jenniferd-unity, line 165: is there any "is zero-ish" functionality in Maths or C#?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -1,789 +1,123 @@
...
-                m_ShadowNormalBiasProp.floatValue = EditorGUILayout.Slider(Styles.shadowNormalBias, m_ShadowNormalBiasProp.floatValue, 0.0f, UniversalRenderPipeline.maxShadowBias);
-                EditorGUILayout.PropertyField(m_SoftShadowsSupportedProp, Styles.supportsSoftShadows);
-                EditorGUI.indentLevel--;
-                EditorGUILayout.Space();
-                EditorGUILayout.Space();
-            }
-            EditorGUILayout.EndFoldoutHeaderGroup();
+                drawElementCallback = OnDrawElement,
+                drawHeaderCallback = (Rect rect) => EditorGUI.LabelField(rect, Styles.rendererHeaderText),
+                onCanRemoveCallback = li => li.count > 1,
```
jenniferd-unity, line 40: what is "li"? listIndex?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -1,789 +1,123 @@
...
+                onRemoveCallback = OnRemoveElement,
+                onReorderCallbackWithDetails = (reorderableList, index, newIndex) => UpdateDefaultRendererValue(index, newIndex) // Need to update the default renderer index
+            };
         }
 
-        private void DrawCascadeSliders(int splitCount, bool useMetric, float baseMetric)
+        void OnRemoveElement(ReorderableList li)
```
jenniferd-unity, line 46: could we rename "li" to something more explicit?

jenniferd-unity (approved):
Approving - minor comments that can be tackled later on

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
alex-vazquez: is the `ReorderableList`. I did not change what was before.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
alex-vazquez: I will

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
alex-vazquez: I do not found any, in this case the editor was always being repainted, my mistake I hade the minus added.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
jenniferd-unity: oh, could we change it? it aint clear to me at all

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
jenniferd-unity: thanks <3

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
alex-vazquez: yeah, I will tackle it with the other one :)

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
alex-vazquez: I do not know, I've been looking at it until I found this solution.

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
jenniferd-unity: could we use https://docs.unity3d.com/ScriptReference/Mathf.Approximately.html?

alex-vazquez: ok, so that is checking if a value is between a tolerance of epsilon, might work. Thank you!!

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
RSlysz:
```
if (!Mathf.Approximately(m_Anim.value, 0f))
```
But basically it do the same with more cost. So I think this is ok (I dunno any other one)

Can the animbool can go other way too ?

RSlysz (approved):

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -83,7 +83,7 @@ public enum LightRenderingMode
     public enum ShaderVariantLogLevel
     {
         Disabled,
-        OnlyUniversalRPShaders,
+        OnlyURPShaders,
```
phi-lira, line 86: This is breaking change. Use API Upgrader.

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -70,6 +70,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - The behavior of setting a camera's Background Type to "Dont Care" has changed on mobile. Previously, "Dont Care" would behave identically to "Solid Color" on mobile. Now, "Dont Care" corresponds to the render target being filled with arbitrary data at the beginning of the frame, which may be faster in some situations. Note that there are no guarantees for the exact content of the render target, so projects should use "Dont care" only if they are guaranteed to render to, or otherwise write every pixel every frame.
 - Stripping shader variants per renderer features instead of combined renderer features.
 - When MSAA is enabled and a depth texture is required, the opaque pass depth will be copied instead of scheduling a depth prepass.
+- URP Asset Inspector - Advanced settings have been reordered under `Show Additional Properties` on each section.
+- Change the default name when a new urp asset is created.
```
phi-lira, line 74:
```suggestion
- Changed the default name when a new urp asset is created.
```

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -1,789 +1,123 @@
...
-        SerializedProperty m_SupportsLightLayers;
-        SerializedProperty m_DebugLevelProp;
-
-        SerializedProperty m_ShaderVariantLogLevel;
-        SerializedProperty m_VolumeFrameworkUpdateModeProp;
-
-        SerializedProperty m_ColorGradingMode;
-        SerializedProperty m_ColorGradingLutSize;
-        SerializedProperty m_UseFastSRGBLinearConversion;
+        public ReorderableList rendererList => m_RendererDataList;
```
phi-lira, line 14: Is this new public API?

phi-lira (changes requested):

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
alex-vazquez: when the show starts it is put to 1 and it is going towards 0

simon-engelbrecht-soerensen (approved):
Tested all values, they seems to work and no out of bounds issues etc. 
Though I feel that having "Show additional properties" in the 3 dots menu(and in the editor preferences instead of project settings) seem kind of strange? Feels like awful UX to me 🤔 
Anyway, it all seems to work besides.

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
RSlysz: I think it is ok.

Just to be sure: Why not just `if (m_Anim.value > 0f)`? It seams like we update until we reach 0 right?

alex-vazquez: Floating point imprecision makes comparing floats using the equals operator inaccurate.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
alex-vazquez: Thanks

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
RSlysz: Yes you don't use equal here. So I don't get the issue. Beside sign is encoded separatly when you inspect float encoding so it cannot miss normally when checking >0f. Am I wrong ?

alex-vazquez:
@simon-engelbrecht-soerensen  you can discuss your thoughts with @TheoWong-pixel about the UX

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
alex-vazquez: yes your are right, I will change it

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
alex-vazquez: changed to internal

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -85,6 +85,8 @@ public enum ShaderVariantLogLevel
         Disabled,
         OnlyURPShaders,
         AllShaders,
+        [Obsolete("OnlyUniversalRPShaders is deprecated, use OnlyURPShaders instead (UnityUpgradable) -> OnlyURPShaders", true)]
+        OnlyUniversalRPShaders,
```
phi-lira, line 89: OnlyURPShaders and OnlyUniversalRPShaders have different int values. Will that cause any issue with upgrading?
 Actually, do we need to change this? Is this change driven by how it appears on UI? Maybe if that's the case we could change the UI name only?

alex-vazquez: I've checked, and the enums are stored with their underlaying type, so an int,as the value of the renamed enum is the same as before the serialization is working as expected, and the UnityUpgradable is checking code string so it was working fine, but maybe an user can be doing something that we are not expecting, so I decided to go on your approach and only change the way that we display the enum

Added an inspector name attribute to have:

image

simon-engelbrecht-soerensen (approved):
Approved

TheoWong-pixel (approved):

phi-lira (approved):

PR #5235, opened 2021-07-27
[HDRP] Fix update order in Graphics Compositor causing jumpy camera updates

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1345566/

If a user script changed the state of a camera, then the compositor could see the change with one-frame delay because the compositor's Update() could be called after the user's script Update().

The solution is rather simple, I have moved the compositor's update functionality in LateUpdate.

---
### Testing status
- Issue does not repro anymore with the project attached in the ticket.
- Compositor test is green.


JMargevics (approved):
Looks good to me. Approving ✅👍

adrien-de-tocqueville (approved):

sebastienlagarde (approved):

PR #5240, opened 2021-07-27
Allow to switch Dynamic Resolution Upscale filter via script

FrancescoC-unity:
Added a small API to change upscale filter via API as I got asked via dm and it was trivial to do so. I also remember Thomas I wanting something like that at some point. 

I wrote a small script to cycle through them via script (and print what filter is active too, but you can see how visually it matches, the log is just in my script, not part of this PR). 

https://user-images.githubusercontent.com/43168857/127177241-e1e94990-3f36-4522-a7a6-e34acc43ed56.mp4

I am about to start to write documentation for TAAU, so I am going to add some words about this API with that PR (starting to work on that tomorrow morning at the latest.  Hence no documentation here, but coming real soon. 

Note that this does _not_ include DLSS API because this controls only stuff on the core system and DLSS is HDRP specific.


sebastienlagarde (approved):

kecho (approved):

kecho:
very nice ship it!

peeweek:
This seems very noice! I desperately need this API in order to fix spaceship settings in 2021.2 (and thefefore be able to add switch between DLSS/CAS, etc). 
Do you think this small helper can be backported to 2021.2? @sebastienlagarde

FrancescoC-unity:
@peeweek  keep in mind DLSS is not activated through this API as it is HDRP only, but you can se the fallback with this and then enable disable DLSS on the camera I think? 

Tagging @kecho for more DLSS details.

iM0ve (approved):
Thanks for the video showcase. No further testing needed

sebastienlagarde:
>Do you think this small helper can be backported to 2021.2?
yes

PR #5242, opened 2021-07-27
X pipeline - Fix static reset with instance provided error

alex-vazquez:
### Purpose of this PR

It can no exist a method static named `Reset` in a class that inherits from `ScriptableObject` because it might be called during `ScriptableObject.CreateInstance` 
The reference for this method is [here](https://docs.unity3d.com/2020.1/Documentation/ScriptReference/ScriptableObject.Reset.html)

image

---
### Testing status
- URP project -> Install HDRP -> See the error

---
### Comments to reviewers
This log happened on C++ side as there is a flag that handles if the ScriptableObject that needs to be reset, so in that moment that is creating an instance is calling the method `Reset` of the object.

I found this debugging the C++ code of the Scriptable Object, but I added a new method (non-static) to the class and added some logs, to show the use behaviour here:

image

Also removed the default ctor as the code was commented.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -274,6 +274,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed cases in which object and camera motion vectors would cancel out, but didn't.
 - Fixed HDRP material upgrade failing when there is a texture inside the builtin resources assigned in the material (case 1339865).
 - Fixed custom pass volume not executed in scene view because of the volume culling mask.
+- When the HDProjectSettings was being loaded on some cases the load of the ScriptableObject was calling the method `Reset` from the HDProjectSettings, simply rename the method to avoid an error log from the loading.
 - Fixed remapping of depth pyramid debug view
 - Fixed an issue with asymmetric projection matrices and fog / pathtracing. (case 1330290).
 - Fixed rounding issue when accessing the color buffer in the DoF shader.


com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/HDProjectSettings.cs
```diff
@@ -130,15 +130,9 @@ static HDProjectSettings instance
...
-        HDProjectSettings()
-        {
-            // s_Instance.FillPresentPluginMaterialVersions(); <
-            // We can't call this here as the scriptable object will not be deserialized in yet
-        }
-
         //// We force the instance to be loaded/created and ready with valid values on assembly reload.
         [InitializeOnLoadMethod]
-        static void Reset()
+        static void InitializeFillPresentPluginMaterialVersionsOnLoad()
```
RSlysz, line 135: @slunity When you named it Reset in https://github.com/Unity-Technologies/Graphics/commit/99384cf466fe3074e6e712f692a9191fb8f06084 , Was it to be also called when reseting the ScriptableObject from inspector contextual menu?

slunity: @RSlysz Good question :) I'm trying to remember if the naming was coincidental or not, although I wouldn't have put a static in front if it wasn't in the API. From what @alex-vazquez is saying, this is called by name match only, but more importantly, on load and/or create, and not only when the contextual menu is called? ie basically everytime the SO is created or loaded? In any case this change is ok, but I'm just wondering if there's a scenario where we should also hook reset to do this, I guess not, as the HDProjectSettings is not touched by the user it's not even in the project's Assets hierarchy, so no context menu interactions there.

RSlysz: It was allowed to have static Reset before. Unity have changed and there is a new fence as this was causing silent issues it seams

RSlysz (approved):

jenniferd-unity (approved):

com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/HDProjectSettings.cs
slunity: Ah I see. Thanks for the info!

slunity (approved):

PR #5221, opened 2021-07-26
added store actions API docs

manuele-bonanno:
### Purpose of this PR
Adding missing public API docs for [this PR](https://github.com/Unity-Technologies/Graphics/pull/4161).

PR #5243, opened 2021-07-27
[10.x.x] Backport load/store optimization for forward renderer Drawobjects pass to 10.x

thomas-zeng:
---
### Purpose of this PR
- Backport part of the changes from https://github.com/Unity-Technologies/Graphics/pull/4161 to 10.x.x to remove the redundant depth store in XR simple scene test case. 
- This optimized the case where URP renders direct to eye without copy color and copy depth, which is pretty common in XR.
- fogbugz case for depth store optimization: https://fogbugz.unity3d.com/f/cases/1354459/
- ES fogbugz case for color resolve and store optimization: https://fogbugz.unity3d.com/f/cases/1353964/
- PR requires core side changes: 69a75e448302
---
### Testing status
- ABV
- QA

---
### Comments to reviewers
- Removed public functions introduced in the original PR.
- Only backported the depth store optimization part because the color resolve and store optimization requires trunk side changes.
To QA:
- Manually tested cube scene case and verified the extra depth store is properly removed(by checking renderdoc).


manuele-bonanno:
I am about to backport (all the way down to 19.4) the trunk PR that fixes the store actions assert. Once I do that you should be able to set the store actions in URP

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -177,6 +192,25 @@ public void ConfigureInput(ScriptableRenderPassInput passInput)
...
+            {
+                m_ColorStoreActions[i] = storeActions[i];
+            }
+        }
+
+        internal void ConfigureDepthStoreAction(RenderBufferStoreAction storeAction)
+        {
+            m_DepthStoreAction = storeAction;
+        }
+
```
manuele-bonanno, line 213: these 3 methods are public in master, was it intentional to change them to internal?

manuele-bonanno: also, when they are changed back. to public, we will need to add API docs (which I forgot to add in the original PR, ooops).
I've added the docs in this [PR](https://github.com/Unity-Technologies/Graphics/pull/5221), currently waiting for review, so once my docs PR is approved we should adds the docs too to the backport

manuele-bonanno:
this trunk PR fixes the store actions related asserts. If you can wait for it to land you could raise the minimum version requirement and be able to backport all the store action changes
https://ono.unity3d.com/unity/unity/pull-request/129952/_/2020.3/graphics/store_actions_fixes

manuele-bonanno:
In general unfortunately I don't think the changes can be backported without also backporting the public "Store Actions" UI option, because potentially we can break existing user projects. (which is the main reason why those options were added in the original PR).
It's not safe to always discard depth since we don't know if users inject extra custom passes that try to access it, in that case the original PR would fall back to storing the depth as soon as an injected pass is detected (if Store Actions option set to "Auto" which is the default).
Probably it's better to move the discussion to slack so I can give more details

manuele-bonanno (approved):

sebastienlagarde:
Is this PR still alive? is it mean to be merge ?

thomas-zeng:
Yes, this PR is still alive. I am just back from vacation and will keep pushing this PR.

thomas-zeng:
QA found issue with this PR, further investigation reveals trunk side issue. Submitted trunk PR here https://ono.unity3d.com/unity/unity/pull-request/133068/_/xr/graphics/fix-gles-quest-depthstencil-discard
and will try to land this PR after QA green lighted the trunk PR and tested the fixed trunk with this PR revision.

amaestas13 (approved):
Tested with

Version: 2020.3.20f1.1086.2020.3/xr/graphics/fix-gles-quest-depthstencil-discard.1
Revision: 2020.3/xr/graphics/fix-gles-quest-depthstencil-discard b6908a38da1d
Built: Tue, 28 Sep 2021 21:37:05 GMT

Verified that with core change this feature works as intended n Quest and Quest 2. 

We see an improvement of 2ms in app gpu time when set to auto (when not using a render feature using depth) or discard, compared to store.

PR #5244, opened 2021-07-27
Fixed banding in the volumetric clouds (case 1353672).

anisunity:
https://fogbugz.unity3d.com/f/cases/1353672/
This PR fixes 3 type of issues.
- Artifacts when the density at height 0 or 1 is not null (when defined by the curve)
- Banding artifacts
- Cloud discontinuity

Testing status
Tried to test extensively, cannot reproduce any of the issues anymore.


pierre-unity (approved):
All good on my side.
- all 4 test scenes are working fine
- shadows have been fixed

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricClouds.compute
```diff
@@ -366,25 +366,25 @@ struct CloudCoverageData
     float maxCloudHeight;
 };
 
-float3 AnimateCloudMapPosition(float3 positionWS)
+float3 AnimateCloudMapPosition(float3 positionPS)
```
sebastienlagarde, line 369: what mean PS here?

anisunity: PlanetSpace

anisunity:
Fixed formatting, adding tests

anisunity:
need to update the screenshots

anisunity:
Updated the screenshots and relaunched the tests

remi-chapelain (commented):
It looks like the fix works properly, 
Here's two example from the tests scene that shows the issues. 

https://user-images.githubusercontent.com/57442369/127474090-afceade0-c229-4b52-859a-ee5043883b9c.mp4

https://user-images.githubusercontent.com/57442369/127474093-2a831b4c-5edd-48ef-8471-759e1885a995.mp4

As for the test, I've updated it, lessen the thresholds, put it more contrast and more light to be able to make it fail when the fix is not there (was still passing without the fix before). 
I've also removed wind in the tests for better robustness. 

Yamato has been relaunch, and I'll update the screenhots when it's done.

sebastienlagarde (approved):

remi-chapelain (approved):
Tests are not fully green (linux and osx still fail) but volumetric clouds pass, so we should be fine.

anisunity:
i'll_rerun_test_untill_they_pass_exe

anisunity:
All green

PR #5246, opened 2021-07-28
[HDRP] Fix issue with path tracing when switching between non-persistent cameras

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1337843/

When disabling and re-enabling a camera that has non-persistent history, the converged path tracing image is lost/reset, so we also have to reset the accumulation.

---
### Testing status
A more reliable way to repro the issue in the ticket is:
- Hit play and let the first camera to fully converge
- Hit S to switch to the second camera and then hit A to switch back to the first camera
- The first camera will be black (because the history was reset but the accumulation manager thinks it's converged)

With this PR the issue does not repro anymore.

I have also tested by making the history persistent in the CamSwitch script:
```
 cam1.GetComponent<HDAdditionalCameraData>().hasPersistentHistory = true; 
 cam2.GetComponent<HDAdditionalCameraData>().hasPersistentHistory = true; 
```
and I got the expected behavior.


iM0ve (approved):
I think the PR is well tested for the change it introduces

sebastienlagarde (approved):

PR #5250, opened 2021-07-28
[HDRP] Fixed mask value stored by the LayerMaskParameter class

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1345515/

Before:
the mask value stored by the class would be "concatenated", meaning it would not take non-defined layers into account, which would results in values being incorrectly shifted after a non-defined layer

Now:
the mask value is not "concatenated" anymore, it takes non-defined layers properly into account

Tested on the scene provided in the corresponding ticket.


sebastienlagarde (approved):

JMargevics (approved):
Testing done looks sufficient 👍✅

PR #4190, opened 2021-04-15
Hd/new-draw-api

JarkkoUnity:
### Purpose of this PR
Enables  motion vector rendering when using Graphics.RenderMeshInstanced()

---
### Testing status
Tested motion vectors are properly rendered for RenderMesh() and RenderMeshInstanced() with a local HDRP project

---
### Comments to reviewers
Requires landed  C++ PR #126439 ("New draw API" for 2021.2)

PR #5252, opened 2021-07-28
[2021.2][ShaderGraph] Fix for undefined matrix recently added with the new motion vectors work

jessebarker:
---
### Purpose of this PR
#4190 added some new matrix property macros to support proper motion vector behavior for SRPs.  Unfortunately, the macros reference matrix properties that are not defined for the built-in render pipeline.  This PR defines them to return a 0 matrix (as the feature work is not intended to support builtin, this should be fine).

---
### Testing status
Local testing to ensure there are no errors in SpaceTransforms.hlsl.

Yamato as below.

---
### Comments to reviewers
This change went in earlier today, so there isn't even an available build with it yet.

### DIFF
--- a/com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/ShaderLibrary/Input.hlsl
+++ b/com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/ShaderLibrary/Input.hlsl
@@ -90,7 +90,8 @@ CBUFFER_END
 #define UNITY_MATRIX_I_M   unity_WorldToObject
 #define UNITY_MATRIX_I_P   (float4x4)0
 #define UNITY_MATRIX_I_VP  (float4x4)0
-#define UNITY_PREV_MATRIX_I_M unity_MatrixPreviousMI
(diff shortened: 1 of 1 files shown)


com.unity.shadergraph/Editor/Generation/Targets/BuiltIn/ShaderLibrary/Input.hlsl
```diff
@@ -90,7 +90,8 @@ CBUFFER_END
 #define UNITY_MATRIX_I_M   unity_WorldToObject
 #define UNITY_MATRIX_I_P   (float4x4)0
 #define UNITY_MATRIX_I_VP  (float4x4)0
-#define UNITY_PREV_MATRIX_I_M unity_MatrixPreviousMI
```
joshua-davis, line 93: Should these PREV matrices just be defined to the current ones instead? This way things should kind of work (I think) instead of not at all?

joshua-davis (approved):
Should at least be fine to get us green

sebastienlagarde (approved):

PR #5253, opened 2021-07-28
[2021.2][ShaderGraph] Pass emission along during lighting meta pass for BuiltIn Target

jessebarker:
---
### Purpose of this PR
This PR addresses https://fogbugz.unity3d.com/f/cases/1352861/

It turns out we were not passing emission along during the lighting meta pass for the BuiltIn target.  This fixes that omission.  It is expected that the material inspector GUI will have differences not only between the render pipelines, but even between the shader source types (i.e., that materials based on a Shader Graph will have different GUIs from materials based on the handwritten "standard" shaders from the various pipelines).

---
### Testing status
Manual testing with the repro project from the bug.

Yamato details below.
---
### Comments to reviewers


joshua-davis (approved):
Double checked that the Emission field always exists even if no input node is connected. Seems Good

PR #5256, opened 2021-07-29
Remove redundant upscale filters 

FrancescoC-unity:
As new filters got recently added, some old filters (Bilinear and Lanczos) have always better alternatives and there is not much reason to keep them in as they just add noise and confusion, with essentially no value added anymore. 

This PR also adds migration to make sure old projects using these options go to the equivalent filter that is kept. Also added a note on the upgrade guide. 

More documentation to come.


FrancescoC-unity:
Oh of course I need to also update the visual test, will do straight after lunch.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/UpsampleScene.compute
```diff
@@ -25,9 +25,9 @@ void MainUpsample (uint3 dti : SV_DispatchThreadID)
 
     float2 uvs = ((float2)dti.xy + 0.5f) * _ViewPortSize.zw;
 #if defined(UPSAMPLE_COLOR)
-    _OutputTexture[COORD_TEXTURE2D_X(dti.xy)] = Bilinear(_InputTexture, uvs);
+    _OutputTexture[COORD_TEXTURE2D_X(dti.xy)] = Sample(TEXTURE2D_X_ARGS(_InputTexture, s_linear_clamp_sampler), uvs);
 #endif
```
kecho, line 29: I think you removed this pass altogether with TAAU, since now we are just doing a min/max upsampler on depth and a trilinear on motion vectors.
Thus, I would say we remove this shader altogether as well.

kecho (approved):

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/UpsampleScene.compute
FrancescoC-unity: Indeed, I even removed it C# side, just forgot the shader side. Will do now.

FrancescoC-unity: Done

sebastienlagarde (approved):

sebastienlagarde:
For QA: there is nothing to check here, it is for awreness

LaserYGD:
upgrade guide is missing some information. currently "From 2021.2, Bilinear and Lanczos upscale filters.." should be something like *From 2021.2, Bilinear and Lanczos are now removed upscale filters*

PR #5257, opened 2021-07-29
Update reference image for 049 after OIDN upgrade

KEngelstoft:
### Purpose of this PR
Fix case 1353374: Gfx test UniversalGraphicsTests.049_Lighting_Mixed_Subtractive is failing.

### Testing status
Yamato UniversalGraphicsTests for DX11 is green now https://unity-ci.cds.internal.unity3d.com/job/7889629

### Comments to reviewers
The scene is set to 'auto lightmap filtering' which means that Optix denoiser is used when available and we fall back to OIDN denoiser if not.
On Bokken, the Nvidia driver is quite old so the Optix denoiser isn't used. When manually setting filtering to OIDN I can reproduce the artifacts locally. Therefore I am accepting the new output as reference image.

Let me know if this is targeting the correct branch.


phi-lira (approved):
Approving the PR, there are some sky light changes that need to be investigated.

KEngelstoft:
Posting Slack message here for posterity: The oidn upgrade PR does not affect sky lighting. Comparing it to the previous revision shows that it is only introducing denoiser artifacts. Perhaps some other PR before it changed it slightly but the diff was too small to trigger a failure?
f92578c3ab52-old
b808e73af0e4-new

PR #5259, opened 2021-07-29
[APV] Avoid allocate textures only used with SH2 if using SH1 + Display approx amount of vmem used by APV

FrancescoC-unity:
So up until now we were always allocating textures used by APV as if SH2 was always used; now we only allocate them only when actually needed. 

I also added a small info box to show how much is allocated for the APV 
Unity_aI4JdsFvem

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
@@ -646,7 +646,7 @@ namespace UnityEngine.Experimental.Rendering
         /// Get approximate video memory impact, in bytes, of the system.
         /// </summary>
         /// <returns>An approximation of the video memory impact, in bytes, of the system<returns>
-        public int GetVMemCost()
+        public int GetVideoMemoryCost()
         {
             if (!m_ProbeReferenceVolumeInit)
                 return 0;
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -638,6 +642,18 @@ public void Cleanup()
             m_IsInitialized = false;
         }
 
+        /// <summary>
+        /// Get approximate video memory impact, in bytes, of the system.
+        /// </summary>
+        /// <returns>An approximation of the video memory impact, in bytes, of the system<returns>
+        public int GetVMemCost()
```
alelievr, line 649: When I read this function name I instinctively think about virtual memory and not video memory. Should we rename it to be clearer? (Maybe it's just me)

alelievr (approved):

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
FrancescoC-unity: Oh I can see how it can be indeed confusing, renaming.

alelievr: 👍

sebastienlagarde (approved):

PR #5262, opened 2021-07-29
Ps5/raytracing

peterjohnlong:
## Purpose of this PR
The purpose of this PR is to activate HDRP ray tracing functionality for PS5, by enabling compilation of associated shaders and code paths.

## Testing status
Tested locally on HDRP_DXR_Tests on the console against reference images https://github.cds.internal.unity3d.com/unity/ScriptableRenderPipelinePrivate/tree/ps5-dxr/ReferenceImages/HDRP_DXR_Tests/PS5/PlayStation5/None 
Tested locally on HDRP_DXR_Tests are remain green on DX12.
Currently yamato SRP PS5-DXR test is broken due to unrelated issue.

### DIFF
--- a/com.unity.render-pipelines.high-definition/ValidationConfig.json
+++ b/com.unity.render-pipelines.high-definition/ValidationConfig.json
@@ -26,6 +26,17 @@
 				]
 			},
 			{
+				"Targets": "+PS5",
+				"Files":
+				[
+					"Editor/Material/ShaderGraph/HDTarget.cs"
+				],
+				"Patterns":
+				[
+					"Pragma\\.OnlyRenderers.*Platform\\.PS5"
+				]
+			},
+			{
 				"Targets": "+XboxOne",
 				"Files":
 				[


com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
```diff
@@ -693,9 +699,12 @@ void CheckPersistantNeedReboot()
...
 
         bool IsDXRResourcesCorrect()
-            => IsHdrpGlobalSettingsUsedCorrect()
-            && HDRenderPipelineGlobalSettings.instance.AreRayTracingResourcesCreated()
-            && SystemInfo.supportsRayTracing;
+        {
+            var selectedBuildTarget = CalculateSelectedBuildTarget();
+            return IsHdrpGlobalSettingsUsedCorrect()
+                && HDRenderPipelineGlobalSettings.instance.AreRayTracingResourcesCreated()
+                && (SystemInfo.supportsRayTracing || selectedBuildTarget == BuildTarget.PS5);
```
sebastienlagarde, line 706: can't we have SystemInfo.supportsRayTracing returning true on PS5 instead of having this separate test (selectedBuildTarget == BuildTarget.PS5) ?

sebastienlagarde (approved):
outside of the quetion about SystemInfo.supportsRayTracing (which could be done in a next PR), all looks good.

Other question: With this PR, can we trigger test on scriptableRenderPipelinePrivate on PS5 for raytrace?

Also I expect it need to be backport up to 20.3 ?

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
peterjohnlong: The issue was when building a player for PS5, on a non RTX PC the check would fail.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingRenderer.raytrace
```diff
@@ -2,7 +2,11 @@
 #pragma max_recursion_depth 10
 
 // WARNING: This define must be kept in sync with the c# code
+#if defined(SHADER_API_PSSL)
+#define RAYTRACING_MAX_RECURSION 1
```
anisunity, line 6: why this one only? RayTracingReflection.raytracing and RaytracingIndirectDiffuse need to have their max recursion force to 1 aswell

anisunity (commented):
I think we need to rework a bit the wizard ps5 code as the wizard was centered around the dx12 problem. I'll take it from here @peterjohnlong if you're fine with that. The ps5 related changed are fine (except my question on the max recursion).

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingRenderer.raytrace
peterjohnlong: @anisunity I am fine with you reworking the wizard code. Good point about RaytracingRenderer.raytrace - mainly it was done there as a reaction to some HDRP_DXR tests e.g. 902_Materials_SG_Variants_xxxxx where a third of the screen is testing recursive raytrace, in the case of PS5 I wanted to achieve a consistent result of just the background. I will rethink best way to do this.

peterjohnlong:
@sebastienlagarde 
I am looking into running test on scriptableRenderPipelinePrivate on PS5 for raytrace again, which is broken ATM. 
I will need to backport up to 20.3, but only once the C++ PR for the backports 2020.3 and 2021.1 are ready otherwise it will break PS5.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingRenderer.raytrace
peterjohnlong: @anisunity I have reverted this change to RayTracingRenderer.raytrace, and added a scripting define in ProjectSettings NO_RAY_RECURSION, and tests to set the max recursion in the shader setup .cs code for this and the other cases

anisunity:
@peterjohnlong i'll do a quick pass on the pr later this week to change the wizard, etc.

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
```diff
@@ -144,57 +144,106 @@ public Entry(QualityScope scope, InclusiveMode mode, Style.ConfigStyle configSty
             }
         }
 
+
+        [InitializeOnLoadMethod]
+        static void Bobby()
```
RSlysz, line 149: Please find a meaningful name :)

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
```diff
@@ -144,57 +144,106 @@ public Entry(QualityScope scope, InclusiveMode mode, Style.ConfigStyle configSty
...
+                new Entry(QualityScope.CurrentQuality, InclusiveMode.DXROptional, Style.dxrGI, IsDXRGICorrect, null, forceDisplayCheck: true, skipErrorIcon: true, displayAssetName: true),
+                new Entry(QualityScope.Global, InclusiveMode.DXROptional, Style.dxrGIFS, IsDXRGIFSCorrect, null, forceDisplayCheck: true, skipErrorIcon: true, displayAssetName: false),
+            });
+
+            return entryList.ToArray();
+        }
+
+        internal void ReBuildEntryList()
+        {
+            m_Entries = null;
```
RSlysz, line 231: Not needed. The line below already replace it. Former will be cleaned later by garbage collection.

Note that alternately if you only nullify it, the list will be rebuild by the lazy constructor anyway in `entries`. But in this case the name of this method is not accurate anymore.

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
```diff
@@ -144,57 +144,106 @@ public Entry(QualityScope scope, InclusiveMode mode, Style.ConfigStyle configSty
...
+        static void Bobby()
+        {
+            if (EditorWindow.HasOpenInstances<HDWizard>())
+            {
+                HDWizard window = (HDWizard)EditorWindow.GetWindow(typeof(HDWizard));
+                window.ReBuildEntryList();
+            }
+        }
+
+        Entry[] BuildEntryList()
```
RSlysz, line 158: [Optional] I think that you can simplify things a little by removing BuildEntryList() and only keep ReBuildEntryList().

You should be able to make it work with only 
```
internal void ReBuildEntryList()
{
    //compute List<Entry> entryList as below this comment
    m_Entries = BuildEntryList();
}
```

and then entries would become
```
Entry[] entries
{
    get
    {
        // due to functor, cannot static link directly in an array and need lazy init
        if (m_Entries == null)
            ReBuildEntryList();
        return m_Entries;
    }
 }
```
Especially because when cleaning current ReBuildEntryList, it will become a one liner method.

RSlysz (changes requested):
Looks good. But please rename **Bobby**

RSlysz:
Above is a fix for error logged in the console when user switch platform and Wizard window is open.

(@alex-vazquez for awareness. If this PR don't land, we would still need to do this fix. It seams unrelated to current Wizard change and could have been introduced 2 month ago in https://github.com/Unity-Technologies/Graphics/commit/31b7b5d146e55850efc2de4a8414a5cce91e2160)

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
peterjohnlong: Renamed Bobby to InitializeEntryList

com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Configuration.cs
peterjohnlong: Removed the unneeded code

sebastienlagarde (approved):

PR #5268, opened 2021-07-30
[HDRP][DXR] Make vertex color default to 1.0 instead of 0.0

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1348821

Before:
When not defined, in ray/path tracing, vertex color would default to 0.0

Now:
When not defined (or null), in ray/path tracing, vertex color defaults to 1.0, matching the raster behaviour

Tested on the scene provided in the corresponding ticket.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingIntersection.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingIntersection.hlsl
@@ -106,6 +106,7 @@ void FetchIntersectionVertex(uint vertexIndex, out IntersectionVertex outVertex)
     outVertex.color      = UnityRayTracingFetchVertexAttribute4(vertexIndex, kVertexAttributeColor);
 
     // We want to default to white in case there is no specified color, to match the raster behaviour
+    // FIXME: This could be addressed in UnityRayTracingFetchVertexAttribute4(), but until then we use this workaround
     if (!any(outVertex.color))
         outVertex.color = 1.0;
 


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingIntersection.hlsl
```diff
@@ -104,6 +104,11 @@ void FetchIntersectionVertex(uint vertexIndex, out IntersectionVertex outVertex)
 
     #ifdef ATTRIBUTES_NEED_COLOR
     outVertex.color      = UnityRayTracingFetchVertexAttribute4(vertexIndex, kVertexAttributeColor);
+
+    // We want to default to white in case there is no specified color, to match the raster behaviour
+    if (!any(outVertex.color))
```
sebastienlagarde, line 109: hmm... I don't think we should do this here. Can you poke Ionut about this one, it should be done on C++ side I guess or inside UnityRayTracingFetchVertexAttribute4

eturquin: I kinda of agree, but since the UnityRayTracingFetchVertexAttribute4(...) is generic, for all kinds of attributes, it makes sense to return 0 as a default value, rather than 1. What could be done maybe, is passing the result floatN as an attribute, and then return a boolean to indicate if the attribute was found or not?

eturquin: Ok there's a bunch of tests on the attribute data format already in UnityRayTracingFetchVertexAttribute4(...), so I guess we could add one on the attribute type, when returning the value if not found. I'll check with Ionut.

eturquin: Discussion ongoing with Ionut, in the meantime we could still merge this (and I added a FIXME comment) until we address this on the engine side.

sebastienlagarde: I am good with the fixme for now. And see what Ionut come with.

sebastienlagarde (approved):

PR #5270, opened 2021-07-30
[HDRP] Fix viewport size when TAA is executed after dynamic res upscale

pmavridis:
### Purpose of this PR
[rebase of [this PR](https://github.com/Unity-Technologies/Graphics/pull/5189)  after TAAU got merged)

There is a "rare" configuration where we execute TAA after DLSS, to accumulate PB DoF samples. In this case, the viewport of TAA was not set correctly, creating the issue in the bug ticket. For post upscale passes, the scale of the history textures need to also be adjusted.

---
### Testing status
- PB DOF + TAA seems to work fine in the repro project after the fix.
- TAA test is green


FrancescoC-unity (approved):
I assume this means the issue was only with DLSS and not TAAU by looking at the code? 

Either way LGTM

pmavridis:
> I assume this means the issue was only with DLSS and not TAAU by looking at the code?
Yes, TAAU was working great!

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -1662,22 +1662,23 @@ class TemporalAntiAliasingData
...
 
             passData.prevFinalViewport = (camera.prevFinalViewport.width < 0 || needToUseCurrFrameSizeForHistory) ? camera.finalViewport : camera.prevFinalViewport;
             var mainRTScales = RTHandles.CalculateRatioAgainstMaxSize(camera.actualWidth, camera.actualHeight);
 
-            var historyRenderingViewport = TAAU ? new Vector2(passData.prevFinalViewport.width, passData.prevFinalViewport.height) :
+            var historyRenderingViewport = (TAAU || runsAfterUpscale) ? new Vector2(passData.prevFinalViewport.width, passData.prevFinalViewport.height) :
                 (needToUseCurrFrameSizeForHistory ? RTHandles.rtHandleProperties.currentViewportSize : camera.historyRTHandleProperties.previousViewportSize);
 
-            if (TAAU && postDoF)
+            if ((TAAU && postDoF) || runsAfterUpscale)
```
FrancescoC-unity, line 1673: When TAAU && postDoF is true, runsAfterUpscale will always be true too, so no need for both checks

iM0ve (approved):
Verifying the fogbuz case should be sufficient

kecho (approved):

sebastienlagarde (approved):

PR #5272, opened 2021-08-02
Fix issue with 0-sized dispatch with extremely low resolutions

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1354181/ 

The issue was due to an integer division that would round down to 0 when resolution is _extremely_ small (not that I expect that would happen in any real project). 

What did I test: downloaded repro and set the resolution to 30x30 to repro. Before it would spam an error, now it doesn't.


sebastienlagarde (approved):

sebastienlagarde:
Added QA for awareness but the change is small enough , test is sufficient (helping with low number of QA this week)

PR #5273, opened 2021-08-02
Update DRS docs and TAAU

FrancescoC-unity:
Adding some more info on Dynamic Resolution doc page and adding some info on TAAU there. 

@Vic-Cooper  is about to push something on this branch, so other reviewers maybe wait for them :-) 

Addresses a couple of fogbugz issues too: https://fogbugz.unity3d.com/f/cases/1354326/ and https://fogbugz.unity3d.com/f/cases/1354324/

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Dynamic-Resolution.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Dynamic-Resolution.md
@@ -25,8 +25,7 @@ The policy type can be one of the following:
 - `DynamicResScalePolicyType.ReturnsPercentage`:  The DynamicResolutionHandler expects the `scaler` to return a screen percentage value. The value set will be clamped between the minimum and maximum percentage set in the HDRP Asset.
 - `DynamicResScalePolicyType.ReturnsMinMaxLerpFactor`:  The DynamicResolutionHandler expects the `scaler` to return a factor `t` that is the [0 ... 1] range and that will be used as a lerp factor between the minimum and maximum screen percentages set in the HDRP asset.
 
-It is best practice to keep the screen percentage above 50%. However, HDRP supports values up to 5, 100.
-Dynamic resolution cannot go above 100% of the rendering resolution.
+It is best practice to keep the screen percentage above 50%. However, HDRP supports values from 5% to 100%, anything below or above this range gets clamped accordingly.
 
 The example below shows how to call this function. In a real production environment, you would call this function depending on the performance of your application.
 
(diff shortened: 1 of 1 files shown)


kecho (approved):

sebastienlagarde (approved):

remi-chapelain:
Added HD QA but waiting for @Vic-Cooper commit to have a look :)

remi-chapelain (changes requested):
In addition to this, could it be mentioned that only FSR / DLSS and TAAU will still be applied when the resolution is forced at 100% ? And the reason why CAS and Catmulll-Rom won't is because they don't help with aliasing so it doesn't make sense to run those at 100% ! 

Thanks :)

FrancescoC-unity:
> In addition to this, could it be mentioned that only FSR / DLSS and TAAU will still be applied when the resolution is forced at 100% ? And the reason why CAS and Catmulll-Rom won't is because they don't help with aliasing so it doesn't make sense to run those at 100% !
> 
> Thanks :)

Tbf, FSR and CAS could both run at 100%. FSR is not doing any anti-aliasing, I think we allow it at 100% to have the extra sharpening it provides. 

DLSS and TAAU otoh do need to run 100% to avoid popping. 

Will add stuff later today (setting a reminder for when I finished catching up, feel free to give me a ping if I forget :) )

com.unity.render-pipelines.high-definition/Documentation~/Dynamic-Resolution.md
```diff
@@ -20,84 +25,91 @@ The policy type can be one of the following:
 - `DynamicResScalePolicyType.ReturnsPercentage`:  The DynamicResolutionHandler expects the `scaler` to return a screen percentage value. The value set will be clamped between the minimum and maximum percentage set in the HDRP Asset.
 - `DynamicResScalePolicyType.ReturnsMinMaxLerpFactor`:  The DynamicResolutionHandler expects the `scaler` to return a factor `t` that is the [0 ... 1] range and that will be used as a lerp factor between the minimum and maximum screen percentages set in the HDRP asset.
 
-The example below shows how to call this function. In a real production environment, you would call this function depending on the performance of your application.
-
+It is best practice to keep the screen percentage above 50%. However, HDRP supports values up to 5, 100.
```
remi-chapelain, line 28: Here's, I believe it's oddly phrased. 
I would say something like, "HDRP supports values from 5% up to 100%"

remi-chapelain (changes requested):
Had a thorough read of the new sections. 
Very useful information, especially parts of the different filters and TAA notes. 👍 

I allowed myself to fix two broken links in the previous commits. 

Before approving, I have 3 little remarks: 
- First one has already been raised in [my previous comment](https://github.com/Unity-Technologies/Graphics/pull/5273#pullrequestreview-731826096). Since it has been [asked on the forums](https://forum.unity.com/threads/using-cas-alongside-taa-just-for-the-sharpening-how-to-use-upscale-filters-at-100.1150967/#post-7425311), I think It would be great to add this on this page too. 
- Second, the doc says that "HDRP supports values up to 5, 100", in HDRP if you put a value above 100 (like 105%), it gets clamped to 100, which makes sense. However, if you put a value lower than 5 (like 2%), it does not get clamped to 5%. I think we should be on par here either on the doc or HDRP. 
- Lastly, having an image to compare filters is a great idea. Even though it's fairly obvious which is which when you know what to look for, I think it's still lacking a legend below or above like A. TAAU / B. Catmull-Rom, especially that they are announced in the wrong order compared to the image and that they unused have letter labels on the image :)
image

remi-chapelain (approved):
all good for me ! ✔️ 
Thanks for addressing the small concerns 🥳

PR #5274, opened 2021-08-02
Fixed an issue with volumetric clouds on vulkan (case 1354802)

anisunity:
https://fogbugz.unity3d.com/f/cases/1354802/
On vulkan only due to a permanent invalidaiton of the shaders at a missing reference, the clouds disappeared if you switched from a scene with clouds and pbr sky to a scene with clouds and an HDRI sky (see repro steps). 

Testing status
Tests are still green and the issues doesn't repro anymore.


iM0ve (approved):
Since the change only touches clouds and our developer checked the repro I think it's safe. As a sanity check, I did a basic test on DX11 to see that clouds still function properly. Built player

PR #5275, opened 2021-08-02
Fix error when changing layer of Post Process Volume

adrien-de-tocqueville:
---
### Purpose of this PR

fix https://fogbugz.unity3d.com/f/cases/1345869/

some logic is done manually at prost process manager update time to handle gameobjects changing layer.
But if a script changes the layer just after creating a volume, or just before destroying one, the update would not have time to run and the volume will not be on the correct layer

---
### Testing status
Tested to enter playmode with both scenes present in the repro case.
No more errors are thrown.


iM0ve (approved):
Small fix, testing the repro case should be sufficient

sebastienlagarde (approved):

sebastienlagarde:
@adrien-de-tocqueville  for this one it is PPV2, please target master. I will do it once I emrge hd/staging in master

PR #5276, opened 2021-08-02
Fix for lens model being taken into account if changed at runtime

FrancescoC-unity:
When we added the lens model (perfect or imperfect lens), we didn't have global settings yet. At the time a change of value would've rebuilt the pipeline so the matching code logic was done at HDRP build. 

Now that we have global frame settings and the value moved there, the asset is not rebuilt and as a consequence the values need to be updated at runtime every frame as soon as the global settings are queried. Fortunately the logic is not much. 

This fixes the issue reported by @pierre-unity  here https://unity.slack.com/archives/C6Y79CZM0/p1627905741038800?thread_ts=1627905530.037600&cid=C6Y79CZM0 

**What did I test**: checked that when changing said value on the global settings, the exposure reacts accordingly immediately.


sebastienlagarde:
@FrancescoC-unity  if this issue is due to move to global settings, it is for master, can you retarget? thanks

sebastienlagarde (approved):

sebastienlagarde:
Hmm maybe you will need that I merge Hd/bugfix to master first

sebastienlagarde (approved):

PR #5283, opened 2021-08-03
Fix for FXAA when rendertarget has alpha channel

adrien-de-tocqueville:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1344618/
https://fogbugz.unity3d.com/f/cases/1351502/

Fix regression introduced by https://github.com/Unity-Technologies/Graphics/pull/2869.

The PR mentionned above was setting keepalpha to false automatically if the render target doesn't have an alpha channel. But the code that check for alpha presence wasn't taking into account render textures

---
### Testing status
Tested both repro projects


iM0ve (approved):
Small change, testing the repro projects should be sufficient

sebastienlagarde (approved):

PR #5284, opened 2021-08-03
Fixed an issue where bloom filtering was wrong when using multiple cameras

JulienIgnace-Unity:
### Purpose of this PR
Fixed an issue where bloom filtering was wrong when using multiple cameras
Fixes this issue: https://fogbugz.unity3d.com/f/cases/1351723/

---
### Testing status
Tested the repro project and ran tests locally


iM0ve (approved):
Testing should be sufficient for the scope of the change

sebastienlagarde (approved):

PR #5286, opened 2021-08-03
[Ready for Merge] Hdrp template iet

pierre-unity:
### Purpose of this PR
This PR adds a brand new In-Editor Tutorial (IET) to the HDRP Scene Template.

More than a dozen tutorials cover a wide range of HDRP basics and features, from Sky to Fog, as well as Light Baking and Decals.

image

The tutorials are meant to be fairly short. Additional resources are linked at the end of each section, for users to dive deeper. Again, it's not meant to be an in-depth tutorial.

image

Other noticeable changes:
- Decals have been repositioned/optimized
- The IET framework package has been added to the package.manifest
- Excluded all VFX from TAA
- Fixed doubles materials in FR_Platform_01.prefab
---
### Comments to reviewers

To access the tutorials, in case the window doesn't already show up, go to the main toolbar at the top of Unity and pick **Tutorials > Show Tutorials**.

To recreate a first experience for the user, choose **Tutorials >  Debug > ClearInitCodeMarker**. This should ensure the tutorial window is loading properly (as a popup).

Check that the instructions are clear enough, easy to follow. 

Check that all the links provided are pointing to the right place. (I did notice that the "latest" manual links often points to HDRP 11 doc pages, I suppose that's normal, as HDRP 12 is still in beta)

Each section should be taking no less than a couple of minutes to complete.

Something important, is to make sure that the template window shows up when opening the tutorial for the first time. I do not know exactly how this works for this scene however (startup code?).

Check as well the tutorial on both Windows and Mac (or other Unix platforms), because the parser used for the text may result in slightly different results. So it you see odd spacing, or strange line returns, this might be a bug due to some hidden Unix characters that are not interpreted correctly on Windows, such as "\r".


com.unity.template-hd/Assets/SampleSceneAssets/Tutorials/Settings/TutorialCallbacks.cs
```diff
@@ -0,0 +1,61 @@
+﻿using UnityEngine;
```
RSlysz, line 1: I dont get the purpose of this script. Callbacks are never called anywhere. There is a ScriptableObject created from it but its GUID is nowhere which means it is used in no scene/prefab/assets. Is it some dead code to remove ?

RSlysz (commented):
There is a lot of change.

Ihave not reviewed all datas. But there is perhaps something on scripts (see below)

Also I tried the tutorial on 2022.1.0a4 and I see this strange scrollbar. It serve no purpose as there is nothing at right
image

com.unity.template-hd/Assets/SampleSceneAssets/Tutorials/Settings/TutorialCallbacks.cs
pierre-unity: @mikaisomaa, is this TutorialCallbacks file absolutely necessary?

I did re-generate it, because upgrading from the preview version of the IET package to the new one created errors on this script. However, it doesn't indeed seem to be necessary at all.

Is it safe to remove?

pierre-unity:
I've added a default layout for tutorial, to make sure that the tutorial window is docked on the very right, rather than floating around (users will most likely be annoyed at it an close it as a result...).

It should look like this:
image

iM0ve (commented):
**Testing status:** Complete

**Issues:**
1) **FIXED** No Welcome page error. We can create the welcome page and assign it to the tutorial or not use it at all ?

https://user-images.githubusercontent.com/36502659/128168534-837d3f55-2070-4362-a0c7-3b9b59be3c7d.mp4

2) Memory leak error

Reproduction:
1) Go through the fist tutorial
2) Open the second
3) Error

https://user-images.githubusercontent.com/36502659/128316821-04b5e870-f2c2-4f74-ad2e-75d6fad582ba.mp4

3) **FIXED**  Light Cookies tutorial is broken, file needs updating. Cant proceed to the end.

https://user-images.githubusercontent.com/36502659/128328951-6c04ba04-e689-4c68-97ed-a543b88cc973.mp4

4) Light layer names are lost 

image


**Suggestions:**
1)  Add among the tutorials a link to feedback. Can be a quiz or forum page. To know what people liked it, what they disliked etc. Data for improvement

image

2) **FIXED**  Add a step mentioning that you need to also go to Lighting tab and select the Static Lighting Sky or use Dynamic mode. This is arguably the most confusing part of HDRP and users often bump into it. Maybe worth explaining how to "get" the ambient lighting

image

3) Do we want this on to increase performance?

image

iM0ve (changes requested):
Updated the review

com.unity.template-hd/Assets/SampleSceneAssets/Tutorials/Settings/TutorialCallbacks.cs
pierre-unity: By the way, the scroll bar is a bug that the IET team will investigate.

pierre-unity:
New Welcome Page was added:

image

iM0ve (changes requested):
Tried to have another look, but having new issues with the first tutorial

https://user-images.githubusercontent.com/36502659/128492490-9f6aabc5-5189-4922-b5fe-0cedaf123f4b.mp4

pierre-unity:
> Tried to have another look, but having new issues with the first tutorial
> 
>  K0BR54gx7Q.mp4

Thanks, yes these errors will popup at random times. The IET team is working on them, or trying to find out at least. It's a known issue with the IET package.

iM0ve (approved):
I updated my review. All issue fixed, 2 concerns are still active, but not blocking the PR

PR #5287, opened 2021-08-03
[HDRP] Fix incorrect light list indexing when TAA is enabled

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1352444/

When indexing into the light list, we need to take into account the jittered projection matrix when TAA is enabled.

Before this PR, some tiles in the planar reflection do not have the correct number of reflection probes and part of the reflection is missing (and flickers because on every frame the error is different based on the TAA jitter):
image

After this PR, all tiles in the planar reflection have the correct number of reflection probes and the reflection is not flickering:
image

---
### Testing status
The ticket does not include a repro scene, so I've built a very simple scene as you see above.
After this PR, the reflection does not flicker.

Note: I have not tested XR, which might be affected by this PR.

### DIFF
--- a/.yamato/_latest_editor_versions_2021.2.metafile
+++ b/.yamato/_latest_editor_versions_2021.2.metafile
@@ -1,5 +1,5 @@
 '2021_2':
-  updated_at_UTC: '2021-07-29 23:13:58.166338'
+  updated_at_UTC: '2021-08-02 23:14:11.327280'
   changeset:
-    id: 2055a14f48ed1c1a48cc5b0a7fefa05d6f6701a4
-    date: '2021-07-29T14:01:49'
+    id: ba17e22e6705c63f1dfb5f84a30bea78e27b0e91
+    date: '2021-08-02T17:44:16'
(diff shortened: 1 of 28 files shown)


JulienIgnace-Unity (commented):
I see we are only changing the DeferredUseComputeAsPixel version of deferred. Isn't there similar code in other versions?

FrancescoC-unity (approved):

FrancescoC-unity:
> I see we are only changing the DeferredUseComputeAsPixel version of deferred. Isn't there similar code in other versions?

I think you might have been tricked by github diff viewer, the change is the prepare light function

sebastienlagarde (approved):

sebastienlagarde:
good catch!

JulienIgnace-Unity:
> I think you might have been tricked by github diff viewer, the change is the prepare light function

I have been tricked by my own bad eye sight :D
All is good then!

JulienIgnace-Unity (approved):

sebastienlagarde:
For QA: merging, test should be enough

PR #5289, opened 2021-08-03
Fix for bad light list access and MSAA mismatch 

FrancescoC-unity:
This should fix a GPU hang that was happening on SSGI on console even though it wasn't directly SSGI fault, but rather weird usage of light lists. 

This PR: 
- Renames LightList from tile and cluster differently
- Bind them locally where possible. 

Also there was an issue with MSAA settings, that would end up sanitize frame settings as MSAA was ON if the frame settings where pointing to Asset and asset had None as MSAA quality. 

This also required a change in the test. 

I kicked yamato but before the test change so it might need rekicking, but you can see how PS5 already is not hanging anymore https://yamato.cds.internal.unity3d.com/jobs/339-ScriptableRenderPipelinePrivate/tree/HDRP%2Ffrancesco-test2/.yamato%2Fhdrp-ps5-playstation5.yml%23HDRP_PS5_PlayStation5_Standalone_il2cpp_Linear_2021.2/7958473/job/(log:Execution)

### DIFF
--- a/TestProjects/HDRP_RuntimeTests/Assets/Scenes/005-DistortCloudsParallax.unity
+++ b/TestProjects/HDRP_RuntimeTests/Assets/Scenes/005-DistortCloudsParallax.unity
@@ -652,7 +652,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 172515602e62fb746b5d573b38a5fe58, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  isGlobal: 1
+  m_IsGlobal: 1
   priority: 0
   blendDistance: 0
   weight: 1
@@ -1819,6 +1819,7 @@ MonoBehaviour:
   taaAntiFlicker: 0.5
   taaMotionVectorRejection: 0
   taaAntiHistoryRinging: 0
+  taaBaseBlendFactor: 0.875
   physicalParameters:
     m_Iso: 200
     m_ShutterSpeed: 0.005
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -816,7 +816,7 @@ internal static void Sanitize(ref FrameSettings sanitizedFrameSettings, Camera c
             //MSAA only supported in forward and when not using ray tracing.
             if (sanitizedFrameSettings.litShaderMode != LitShaderMode.Forward || pipelineSupportsRayTracing)
                 sanitizedFrameSettings.msaaMode = MSAAMode.None;
-            bool msaa = (sanitizedFrameSettings.msaaMode == MSAAMode.FromHDRPAsset && renderPipelineSettings.msaaSampleCount != MSAASamples.None) || sanitizedFrameSettings.msaaMode != MSAAMode.None;
+            bool msaa = sanitizedFrameSettings.msaaMode == MSAAMode.FromHDRPAsset ? renderPipelineSettings.msaaSampleCount != MSAASamples.None : sanitizedFrameSettings.msaaMode != MSAAMode.None;
```
sebastienlagarde, line 819: @JulienIgnace-Unity please review this one

sebastienlagarde (approved):
Note to self: add an upgrade guide for this change

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
FrancescoC-unity: We did talk about this with Julien, but worth giving the thing a re read yeah

JulienIgnace-Unity (approved):

PR #5294, opened 2021-08-03
Optimizing PrepareLightsForGPU (ProcessVisibleLights) with burst

kecho:
### Purpose of this PR
Optimization of PrepareLightsforGPU (ProcessVisibleLights) on HDRP using burst, in a ECS friendly way.

This PR includes:
* Introduction of HDLightEntityCollection - which is a master singleton render side entity manager.
* Replacing the flat ProcessVisibleLights loop with a parallel job, using the HDVisibleLightEntities helper.
* Reduction of dependencies to HDAdditionalLights on the ProcessVisibleLight loop
* Burstification / parallelization of ProcessVisibleLights
* New sorting functions, performance oriented. There are now 3 differnet sorting algorithms. Insertion sort, which performs best between 1 and 31 lights, Merge sort which performs best between 32 and 199 lights, and finally radix sort which has best performance 200+ lights more.
* This PR has a bit of regression in performance on HDAdditionalLightData creation, and modification of several properties. This perf regression does justify the gains on the main pipeline. At some point, DOTs lights should fix any performance issues inside HDAdditionalLightData. This PR's mission is to optimize the HDRP side, not the game object side. 

NOTE:
* The convert to GPU format path is still sub optimal, since its tightly coupled with shadow processing / allocation. The next round of optimizations will focus on shadow allocation.
* This PR adds an extra cost to HDAdditioanLight having

---
### Testing status
Seeing a reduction of around 50%-70% on PrepareLightsForGPU.
image
Comparisons are done after compiling with IL2CPP at max optimization level, against master on ps4. 
Similar performance comparisons must be done in Windows: create a player in windows with il2cpp with maximum optimization level, and compare master vs this branch.

Testing performed
* Shuriken lights - Created shuriken particle object, added generation of lights. Ensured that lights work, fixed also a crash on master when lights  in shuriken have shadows enabled.
* Tested live editing of HDAdditionalLightData (intensity and spot light inner/outer angles)
* Tested live editing of area light shadows and geometry generation
* Tested scene live edit / and general parameters for shadows (spot light shadows, tinting)
* Tested deleting adding many lights. 

Additional testing required:
* Testing of light cookies
* Testing of shadow allocation (live editing + stress testing many lights with shadows)
* IES profile cutoff percent and IES profile live editing / runtime
* light dimmer parameter
* fade distance parameter
* volumetric fade distance parameter
* affect diffuse / affect specular for all light types parameter
* shape width / shape height for lights parameter
* shape radius, softness scale, light range attennuation
* area light cookie
* Ray traced shadows for lights (if available)
* Ray tracing using local lights for reflections
* Screen space light shadows 
* Interacting light with sky (interactsWithSky parameter)
* Shadow tint / penumbra tint for all light types
* flare falloff, flare tint and flare size parameters
* Light baking and shadow baking of lights. 

---
### Technical Information

The following is a technical writeup, intended to give a high level overview of the work done on this PR. Before we can dive in, I need to explain how exactly HDRP does light culling and why it is expensive.
Originally the light loop in HDRP is divided into 3 big steps.

#### Culling in HDRP

**Step 1**
first step is frustum culling coming from the SRP, which converts Light game objects (refered as GOs) to VisibleLight structs. VisibleLight struct contains a copy of the visible light, and is all performed in the c++, providing these structs (burst accessible) for the SRP to access

**Step 2**
Next, HDRP has a HDAdditionalLightData GO for additional PBR options not present in the regular Light component. This component is self managed, and the UI obfuscates everything, creating a nice GUI that lets HDRP have a more advanced light definition.
The second step basically took the VisibleLight objects, and processed them with the HDAdditioanlLightData and did some extra quick rejection. This produced an intermediate list called ProcessedLightData.
The ProcessedLightData then gets sorted using a recursive quicksort, which is unnacceptably heavy on the CPU.

**Step 3** 
Last step then converts ProcessedLightData into a GPU ready struct light light list.

The 2nd and 3rd step were all done linearly, in the main thread, accessed a lot the HDAdditionalData and were huge perf bottle necks. The sort was as well gigantic.

#### Optimization Parts

**Part A**
Added the concept of a render entity (HDRenderLightEntity) and a global manager for these entities. This basically mantains light data (from HDAdditionalLightData GO ) as struct. The GO has an entity (handle) to this render data and is responsible to update it from the Sim side. The GO then just becomes an interface to the UI / serialization / artist. Its not used anymore by the render pipeline (with the exception of shadow procesisng, which will require another separate PR).
This effectively (almost) decouples the rneder pipeline from the GO. This allows the render pipeline then to access the HDLightRenderLightEntityData as a struct, thus unlocking burstification.

**Part B**
Moved the step 2 mentioned above to burst, with the exception of shadow allocation. ProcessedLightData gets accumulated using an atomic increment across all threads. This is a gigantic speedup on the CPU. With the exception of shadows, shadow lights still get processed 'patched' in the main thread.

**Part C**
Rewrote the sorts to be done only on UInts, and on pointers for maximum speed.
Added 3 sort algorithms:
Insertion sort for a light count <32
Non recursive merge sort with auxiliary data for >= 32  and <= 256
Radix sort with auxiliary data with 8 bit radix for 256 >= and above.
This is effectively almost 5x faster, and can also scale massively. The sort is still done in the main thread (there were no good gains on it being on burst)

**Part D** 
Move the Step 3 above to burst. Keep the shadow pieces in the main thread.

#### Conclusion and Next steps
Huge perf gains, new sort loops available are now in Core, for next work  we have to optimize the shadow atlas allocation, matrix calculation. Investigate reducing the 3 step culling algorithm to a single step.

This makes it now possible for DOTS as well to inject its own 'sim side' light definition and wrap the render side entity. Shadow work will finalize the decoupling of HDAdditionalLightData GO and enable this final perk.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
+++ b/com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
 
 namespace UnityEngine.Rendering
@@ -148,6 +149,233 @@ namespace UnityEngine.Rendering
             var c = Mathf.Min(count, list.Length);
             for (int i = 0; i < c; ++i)
                 UnsafeUtility.WriteArrayElement<T>(dest, i, list[i]);
+        }
+
+        private static void CalculateRadixParams(int radixBits, out int bitStates)
+        {
+            if (radixBits != 2 && radixBits != 4 && radixBits != 8)
+                throw new Exception("Radix bits must be 2, 4 or 8 for uint radix sort.");
+            bitStates = 1 << radixBits;
+        }
+
+        private static int CalculateRadixSupportSize(int bitStates, int arrayLength)
+        {
+            return bitStates * 3 + arrayLength;
+        }
+
+        private static unsafe void CalculateRadixSortSupportArrays(
+            int bitStates, int arrayLength, uint* supportArray,
+            out uint* bucketIndices, out uint* bucketSizes, out uint* bucketPrefix, out uint* arrayOutput)
+        {
+            bucketIndices = supportArray;
+            bucketSizes = bucketIndices + bitStates;
+            bucketPrefix = bucketSizes + bitStates;
+            arrayOutput = bucketPrefix + bitStates;
+        }
+
+        private static unsafe void MergeSort(uint* array, uint* support, int length)
+        {
+            for (int k = 1; k < length; k *= 2)
+            {
+                for (int left = 0; left + k < length; left += k * 2)
+                {
+                    int right = left + k;
+                    int rightend = right + k;
+                    if (rightend > length)
+                        rightend = length;
+                    int m = left;
+                    int i = left;
+                    int j = right;
+                    while (i < right && j < rightend)
+                    {
+                        if (array[i] <= array[j])
+                        {
+                            support[m] = array[i++];
+                        }
+                        else
+                        {
+                            support[m] = array[j++];
+                        }
+                        m++;
+                    }
+                    while (i < right)
+                    {
+                        support[m] = array[i++];
+                        m++;
+                    }
+                    while (j < rightend)
+                    {
+                        support[m] = array[j++];
+                        m++;
+                    }
+                    for (m = left; m < rightend; m++)
+                    {
+                        array[m] = support[m];
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Merge sort - non recursive
+        /// </summary>
+        /// <param name="arr">Array to sort.</param>
+        /// <param name="sortSize">Size of the array to sort. If greater than array capacity, it will get clamped.</param>
+        /// <param name="supportArray">Secondary array reference, used to store intermediate merge results.</param>
+        public static unsafe void MergeSort(uint[] arr, int sortSize, ref uint[] supportArray)
+        {
+            sortSize = Math.Min(sortSize, arr.Length);
+            if (arr == null || sortSize == 0)
+                return;
+
+            if (supportArray == null || supportArray.Length < sortSize)
+                supportArray = new uint[sortSize];
+
+            fixed (uint* arrPtr = arr)
+            fixed (uint* supportPtr = supportArray)
+                CoreUnsafeUtils.MergeSort(arrPtr, supportPtr, sortSize);
+        }
+
+        /// <summary>
+        /// Merge sort - non recursive
+        /// </summary>
+        /// <param name="arr">Array to sort.</param>
+        /// <param name="sortSize">Size of the array to sort. If greater than array capacity, it will get clamped.</param>
+        /// <param name="supportArray">Secondary array reference, used to store intermediate merge results.</param>
+        public static unsafe void MergeSort(NativeArray<uint> arr, int sortSize, ref NativeArray<uint> supportArray)
+        {
+            sortSize = Math.Min(sortSize, arr.Length);
+            if (!arr.IsCreated || sortSize == 0)
+                return;
+
+            if (!supportArray.IsCreated || supportArray.Length < sortSize)
+                supportArray.ResizeArray(arr.Length);
+
+            CoreUnsafeUtils.MergeSort((uint*)arr.GetUnsafePtr(), (uint*)supportArray.GetUnsafePtr(), sortSize);
+        }
+
+        private static unsafe void InsertionSort(uint* arr, int length)
+        {
+            for (int i = 0; i < length; ++i)
+            {
+                for (int j = i; j >= 1; --j)
+                {
+                    if (arr[j] >= arr[j - 1])
+                        break;
+
+                    var tmp = arr[j];
+                    arr[j] = arr[j - 1];
+                    arr[j - 1] = tmp;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Insertion sort
+        /// </summary>
+        /// <param name="arr">Array to sort.</param>
+        /// <param name="sortSize">Size of the array to sort. If greater than array capacity, it will get clamped.</param>
+        public static unsafe void InsertionSort(uint[] arr, int sortSize)
+        {
+            sortSize = Math.Min(arr.Length, sortSize);
+            if (arr == null || sortSize == 0)
+                return;
+
+            fixed (uint* ptr = arr)
+                CoreUnsafeUtils.InsertionSort(ptr, sortSize);
+        }
+
+        /// <summary>
+        /// Insertion sort
+        /// </summary>
+        /// <param name="arr">Array to sort.</param>
+        /// <param name="sortSize">Size of the array to sort. If greater than array capacity, it will get clamped.</param>
+        public static unsafe void InsertionSort(NativeArray<uint> arr, int sortSize)
+        {
+            sortSize = Math.Min(arr.Length, sortSize);
+            if (!arr.IsCreated || sortSize == 0)
+                return;
+
+            CoreUnsafeUtils.InsertionSort((uint*)arr.GetUnsafePtr(), sortSize);
+        }
+
+        private static unsafe void RadixSort(uint* array, uint* support, int radixBits, int bitStates, int length)
+        {
+            uint mask = (uint)(bitStates - 1);
+            CalculateRadixSortSupportArrays(bitStates, length, support, out uint* bucketIndices, out uint* bucketSizes, out uint* bucketPrefix, out uint* arrayOutput);
+
+            int buckets = (sizeof(uint) * 8) / radixBits;
+            uint* targetBuffer = arrayOutput;
+            uint* inputBuffer = array;
+            for (int b = 0; b < buckets; ++b)
+            {
+                int shift = b * radixBits;
+                for (int s = 0; s < 3 * bitStates; ++s)
+                    bucketIndices[s] = 0;//bucketSizes and bucketPrefix get zeroed, since we walk 3x the bit states
+
+                for (int i = 0; i < length; ++i)
+                    bucketSizes[((inputBuffer[i] >> shift) & mask)]++;
+
+                for (int s = 1; s < bitStates; ++s)
+                    bucketPrefix[s] = bucketPrefix[s - 1] + bucketSizes[s - 1];
+
+                for (int i = 0; i < length; ++i)
+                {
+                    uint val = inputBuffer[i];
+                    uint bucket = (val >> shift) & mask;
+                    targetBuffer[bucketPrefix[bucket] + bucketIndices[bucket]++] = val;
+                }
+
+                uint* tmp = inputBuffer;
+                inputBuffer = targetBuffer;
+                targetBuffer = tmp;
+            }
+        }
+
+        /// <summary>
+        /// Radix Sort
+        /// Radix sort or bucket sort, stable and non in place.
+        /// </summary>
+        /// <param name="arr">Array to sort.</param>
+        /// <param name="sortSize">Size of the array to sort. If greater than array capacity, it will get clamped.</param>
+        /// <param name="supportArray">Array of uints that is used for support data. The algorithm will automatically allocate it if necessary.</param>
+        /// <param name="radixBits">Number of bits to use for each bucket. Can only be 8, 4 or 2.</param>
+        public static unsafe void RadixSort(uint[] arr, int sortSize, ref uint[] supportArray, int radixBits = 8)
+        {
+            sortSize = Math.Min(sortSize, arr.Length);
+            CalculateRadixParams(radixBits, out int bitStates);
+            if (arr == null || sortSize == 0)
+                return;
+
+            int supportSize = CalculateRadixSupportSize(bitStates, sortSize);
+            if (supportArray == null || supportArray.Length < supportSize)
+                supportArray = new uint[supportSize];
+
+            fixed (uint* ptr = arr)
+            fixed (uint* supportArrayPtr = supportArray)
+                CoreUnsafeUtils.RadixSort(ptr, supportArrayPtr, radixBits, bitStates, sortSize);
+        }
+
+        /// <summary>
+        /// Radix Sort
+        /// Radix sort or bucket sort, stable and non in place.
+        /// </summary>
+        /// <param name="arr">Array to sort.</param>
+        /// <param name="sortSize">Size of the array to sort. If greater than array capacity, it will get clamped.</param>
+        /// <param name="supportArray">Array of uints that is used for support data. The algorithm will automatically allocate it if necessary.</param>
+        /// <param name="radixBits">Number of bits to use for each bucket. Can only be 8, 4 or 2.</param>
+        public static unsafe void RadixSort(NativeArray<uint> array, int sortSize, ref NativeArray<uint> supportArray, int radixBits = 8)
+        {
+            sortSize = Math.Min(sortSize, array.Length);
+            CalculateRadixParams(radixBits, out int bitStates);
+            if (!array.IsCreated || sortSize == 0)
+                return;
+
+            int supportSize = CalculateRadixSupportSize(bitStates, sortSize);
+            if (!supportArray.IsCreated || supportArray.Length < supportSize)
+                supportArray.ResizeArray((int)supportSize);
+
+            CoreUnsafeUtils.RadixSort((uint*)array.GetUnsafePtr(), (uint*)supportArray.GetUnsafePtr(), radixBits, bitStates, sortSize);
         }
 
         /// <summary>
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.Types.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.Types.cs
@@ -198,9 +198,9 @@ namespace UnityEngine.Rendering.HighDefinition
 
     public partial class HDAdditionalLightData
     {
-        //Private enum to differentiate built-in LightType.Point that can be Area or Point in HDRP
+        //Internal enum to differentiate built-in LightType.Point that can be Area or Point in HDRP
         //This is due to realtime support and culling behavior in Unity
-        private enum PointLightHDType
+        internal enum PointLightHDType
         {
             Punctual,
             Area
@@ -245,14 +245,17 @@ namespace UnityEngine.Rendering.HighDefinition
                         case HDLightType.Directional:
                             legacyLight.type = LightType.Directional;
                             m_PointlightHDType = PointLightHDType.Punctual;
+                            HDLightRenderDatabase.instance.EditLightDataAsRef(lightEntity).pointLightType = m_PointlightHDType;
                             break;
                         case HDLightType.Spot:
                             legacyLight.type = LightType.Spot;
                             m_PointlightHDType = PointLightHDType.Punctual;
+                            HDLightRenderDatabase.instance.EditLightDataAsRef(lightEntity).pointLightType = m_PointlightHDType;
                             break;
                         case HDLightType.Point:
                             legacyLight.type = LightType.Point;
                             m_PointlightHDType = PointLightHDType.Punctual;
+                            HDLightRenderDatabase.instance.EditLightDataAsRef(lightEntity).pointLightType = m_PointlightHDType;
                             break;
                         case HDLightType.Area:
                             ResolveAreaShape();
@@ -288,6 +291,7 @@ namespace UnityEngine.Rendering.HighDefinition
                     return;
 
                 m_SpotLightShape = value;
+                HDLightRenderDatabase.instance.EditLightDataAsRef(lightEntity).spotLightShape = m_SpotLightShape;
 
                 // If the current light unit is not supported by this spot light shape, we change it
                 var supportedUnits = GetSupportedLightUnits(type, value);
@@ -309,6 +313,8 @@ namespace UnityEngine.Rendering.HighDefinition
                     return;
 
                 m_AreaLightShape = value;
+                HDLightRenderDatabase.instance.EditLightDataAsRef(lightEntity).areaLightShape = m_AreaLightShape;
+
                 if (type == HDLightType.Area)
                     ResolveAreaShape();
                 UpdateAllLightValues();
@@ -318,6 +324,7 @@ namespace UnityEngine.Rendering.HighDefinition
         void ResolveAreaShape()
         {
             m_PointlightHDType = PointLightHDType.Area;
+            HDLightRenderDatabase.instance.EditLightDataAsRef(lightEntity).pointLightType = m_PointlightHDType;
             if (areaLightShape == AreaLightShape.Disc)
             {
                 legacyLight.type = LightType.Disc;
@@ -466,6 +473,28 @@ namespace UnityEngine.Rendering.HighDefinition
             return allowedUnits.Any(u => u == unit);
         }
 
+        internal static HDLightType TranslateLightType(LightType lightType, PointLightHDType pointLightType)
+        {
+            switch (lightType)
+            {
+                case LightType.Spot: return HDLightType.Spot;
+                case LightType.Directional: return HDLightType.Directional;
+                case LightType.Point:
+                    switch (pointLightType)
+                    {
+                        case PointLightHDType.Punctual: return HDLightType.Point;
+                        case PointLightHDType.Area: return HDLightType.Area;
+                        default: return HDLightType.Point;
+                    }
+                case LightType.Disc:
+                    return HDLightType.Area;
+                case LightType.Rectangle:
+                    return HDLightType.Area;
+                default:
+                    return HDLightType.Point;
+            }
+        }
+
         //To use in render loop instead of type as we can add on the fly an
         //HDAdditionalLightData that is not really added to the GameObject
         //In this case, the type property will return a false value as this will
@@ -476,39 +505,23 @@ namespace UnityEngine.Rendering.HighDefinition
             if (attachedLight == null)
                 return HDLightType.Point;
 
-            switch (attachedLight.type)
-            {
-                case LightType.Spot: return HDLightType.Spot;
-                case LightType.Directional: return HDLightType.Directional;
-                case LightType.Point:
-                    switch (m_PointlightHDType)
-                    {
-                        case PointLightHDType.Punctual: return HDLightType.Point;
-                        case PointLightHDType.Area: return HDLightType.Area;
-                        default:
-                            //Debug.Assert(false, $"Unknown {typeof(PointLightHDType).Name} {m_PointlightHDType}. Fallback on Punctual");
-                            return HDLightType.Point;
-                    }
-                case LightType.Disc:
-                    return HDLightType.Area;
-                case LightType.Rectangle:
-                    // not supported directly. Convert now to equivalent if not default HDAdditionalLightData:
-                    if (this != HDUtils.s_DefaultHDAdditionalLightData)
-                    {
-                        legacyLight.type = LightType.Point;
-                        m_PointlightHDType = PointLightHDType.Area;
-                        m_AreaLightShape = AreaLightShape.Rectangle;
+            HDLightType hdLightType = TranslateLightType(attachedLight.type, m_PointlightHDType);
 
-                        //sanitycheck on the baking mode first time we add additionalLightData
+            if (attachedLight.type == LightType.Rectangle && this != HDUtils.s_DefaultHDAdditionalLightData)
+            {
+                legacyLight.type = LightType.Point;
+                m_PointlightHDType = PointLightHDType.Area;
+                HDLightRenderDatabase.instance.EditLightDataAsRef(lightEntity).pointLightType = m_PointlightHDType;
+
+                m_AreaLightShape = AreaLightShape.Rectangle;
+
+                //sanitycheck on the baking mode first time we add additionalLightData
 #if UNITY_EDITOR
-                        legacyLight.lightmapBakeType = LightmapBakeType.Realtime;
+                legacyLight.lightmapBakeType = LightmapBakeType.Realtime;
 #endif
-                    }
-                    return HDLightType.Area;
-                default:
-                    Debug.Assert(false, $"Unknown {typeof(LightType).Name} {attachedLight.type}. Fallback on Point");
-                    return HDLightType.Point;
             }
+
+            return hdLightType;
         }
     }
 }
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using JetBrains.Annotations;
 using UnityEngine;
 using UnityEngine.Rendering;
+using UnityEngine.Assertions;
 #if UNITY_EDITOR
 using UnityEditor;
 #endif
@@ -122,6 +123,9 @@ namespace UnityEngine.Rendering.HighDefinition
 
         #region HDLight Properties API
 
+        [ExcludeCopy]
+        internal HDLightRenderEntity lightEntity = HDLightRenderEntity.Invalid;
+
         [SerializeField, FormerlySerializedAs("displayLightIntensity")]
         float m_Intensity;
         /// <summary>
@@ -194,6 +198,9 @@ namespace UnityEngine.Rendering.HighDefinition
 
                 m_InnerSpotPercent = Mathf.Clamp(value, k_MinSpotInnerPercent, k_MaxSpotInnerPercent);
                 UpdateLightIntensity();
+
+                if (lightEntity.valid)
+                    HDLightRenderDatabase.instance.EditLightDataAsRef(lightEntity).innerSpotPercent = m_InnerSpotPercent;
             }
         }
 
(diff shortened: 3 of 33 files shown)


LaserYGD:
Hi, 

Is this still planned? PrepereLightsForGPU takes +1ms in our project.

kecho:
> Hi,
> 
> Is this still planned? PrepereLightsForGPU takes +1ms in our project.

Still planned. Had to work on other things this week. Which platform are you seeing this and how many lights? (How many have shadows etc?)

LaserYGD:
> > Hi,
> > Is this still planned? PrepereLightsForGPU takes +1ms in our project.
> 
> Still planned. Had to work on other things this week. Which platform are you seeing this and how many lights? (How many have shadows etc?)

Around 160 pointlights, all have shadows disabled & volumetrics disabled. For curiosity In one scene I enabled shadows and got +300ms in PrepereLightsForGPU. I do realize how expensive point light shadows are, just curious if this is expected results from PrepereLightsForGPU. 

I thought it might be a misconfiguration or something project related, so I downloaded an asset demo and tried it out there with a large number of point lights, got the same result. Same issue occurs when shadows are set to on enable. 

This is the profiler when shadows are enabled: 
image
It's still 

I'm not really sure if this is to be expected/normal. PrepereLights gets bad even with 20-40 point lights with shadows enabled.

kecho:
> > > Hi,
> > > Is this still planned? PrepereLightsForGPU takes +1ms in our project.
> > 
> > 
> > Still planned. Had to work on other things this week. Which platform are you seeing this and how many lights? (How many have shadows etc?)
> 
> Great to hear that.
> 
> Around 170 pointlights, all have shadows disabled & volumetrics disabled. For curiosity In one scene I enabled shadows and got +300ms in PrepereLightsForGPU. I do realize how expensive point light shadows are, just curious if this is expected results from PrepereLightsForGPU.
> 
> I thought it might be a misconfiguration or something project related, so I downloaded an asset demo and tried it out there with a large number of point lights, got the same result. Same issue occurs when shadows are set to on enable.
> 
> This is the profiler when shadows are enabled:
> image
> 
> Maybe GC related? PrepereLights gets bad even with 20-40 point lights with shadows enabled.
> Project and demo test is on 2020.3.13, Windows.

Shadows is doing some crazy cpu work at the moment to allocate them and get info about it's projection. There are many opportunities to optimize it, but that's gonna be a second phase. Your 130 point lights should be reduced highly hopefully. 
Another question I have, is what CPU are you doing this on?

LaserYGD:
> > > > Hi,
> > > > Is this still planned? PrepereLightsForGPU takes +1ms in our project.
> > > 
> > > 
> > > Still planned. Had to work on other things this week. Which platform are you seeing this and how many lights? (How many have shadows etc?)
> > 
> > 
> > Great to hear that.
> > Around 170 pointlights, all have shadows disabled & volumetrics disabled. For curiosity In one scene I enabled shadows and got +300ms in PrepereLightsForGPU. I do realize how expensive point light shadows are, just curious if this is expected results from PrepereLightsForGPU.
> > I thought it might be a misconfiguration or something project related, so I downloaded an asset demo and tried it out there with a large number of point lights, got the same result. Same issue occurs when shadows are set to on enable.
> > This is the profiler when shadows are enabled:
> > image
> > Maybe GC related? PrepereLights gets bad even with 20-40 point lights with shadows enabled.
> > Project and demo test is on 2020.3.13, Windows.
> 
> Shadows is doing some crazy cpu work at the moment to allocate them and get info about it's projection. There are many opportunities to optimize it, but that's gonna be a second phase. Your 130 point lights should be reduced highly hopefully.
> Another question I have, is what CPU are you doing this on?

laptop: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz, 2601 Mhz, 4 Core(s), 8 Logical Processor(s)

kecho:
> laptop: Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz, 2601 Mhz, 4 Core(s), 8 Logical Processor(s)

Thank you for your info @LaserYGD 
Give me a couple of weeks, I should have something more functional. Id love it if you could test my changes once they are more functional locally in your project. My email is Kleber.Garcia@unity3d.com so feel free to send me any extra info that way. Thank you again, this is great info, confirms my exploration so far.

kecho:
hey @LaserYGD , I have finished the first round of burstification. Moved the entire 'Preprocess lights' logic to burst.
Locally, I am seeing an improvement of around 70% on CPU wall time. This degrades though when we do shadows. I still have to work on shadow on CPU.
Could you try this branch in your project if you could? Would be great to see the impact of it on your project.

LaserYGD:
@kecho 
Hey, so i did 2 tests,  on a big scene with lots of lights, and another on the same scene with reduced point lights. 
Camera/lights are identical between burst and before tests. 
All tests were done in the editor. All shadows off. 
Note that I didn't average the numbers to get an approximate ~, I simply eyed it over 20-50 frames and picked an estimate. They aren't exact, but they're good enough to get an understanding 😅 

### bigger test (200 point lights, not all in view): 


### With burst: 
PrepereLightsForGPU(total): 0.70ms ~ 0.90ms
PrepereGPUlightdata: 0.40 - 0.45ms
**Visible lights: 0.35 - 0.45ms**

### without burst, latest beta: 
PrepereLightsForGPU(total): 1.00ms ~ 1.10ms 
PrepereGPUlightdata: 0.34 - 0.45ms
**Visible lights: 0.60ms ~ 0.70ms**

-------------------------------------------
### Smaller test (67 point lights, not all in view):


### Burst: 

PrepereLightsForGPU(total): 0.50ms ~ 0.60ms
**PrepareGPULightdata: 0.25ms ~ 0.30ms**
VisibleLights: 0.16ms ~ 0.23ms 
  
### Without burst:
PrepereLightsForGPU(total): 0.40 ~ 0.47
**PrepareGPULightdata: 0.16ms ~ 0.20ms**
VisibleLights: 0.20ms ~ 0.25ms 

It seems that PrepareGPULightdata is a bit slower in this pull. 
VisibleLights improvements is obvious especially when you get to high numbers of lights.

kecho:
> @kecho
> Hey, so i did 2 tests, on a big scene with lots of lights, and another on the same scene with reduced point lights.
> scene is identical between burst and before tests.
> All tests were done in the editor. All shadows off.
> Note that I didn't average the numbers to get an approximate ~, I simply eyed it over 20-50 frames and picked an estimate. They aren't exact, but they're good enough to get an understanding 😅
> 
> burst = this pull
> 
> ### bigger test (200 point lights~):
> ### With burst:
> PrepereLightsForGPU(total): 0.70ms ~ 0.90ms
> PrepereGPUlightdata: 0.40 - 0.45ms
> **Visible lights: 0.35 - 0.45ms**
> 
> ### without burst, latest beta:
> PrepereLightsForGPU(total): 1.00ms ~ 1.10ms
> PrepereGPUlightdata: 0.34 - 0.45ms
> **Visible lights: 0.60ms ~ 0.70ms**
> 
> ### Smaller test (67 point lights, not all in view):
> ### Burst:
> PrepereLightsForGPU(total): 0.42ms~
> **PrepareGPULightdata: 0.17ms ~ 0.25ms**
> VisibleLights: 0.13ms ~ 0.16ms
> 
> ### Without burst, latest beta:
> PrepereLightsForGPU(total): 0.38ms~
> **PrepareGPULightdata: 0.12ms ~ 0.14ms**
> VisibleLights: 0.19ms ~ 0.23ms
> 
> **It seems that PrepareGPULightdata is a bit slower in this pull.**
> VisibleLights improvements is obvious especially when you get to high numbers of lights.
> 
> **_I'm a bit surprised that I'm not getting as big of an improvement as your testing, did you do your tests in the editor? Smaller test seems to do better without burst. There might be something wrong on my side so in that case I'll send the scene over to you. to do your own tests if you want. I will do this test again in a public asset scene so I can share it. Will edit this post when I'm done._**
> 
> edit: sent an email to you with the scene download

Seriously thank you so much! I think the overhead cost is making it a bit more expensive for low count of lights.  It also depends on the CPU. I am seeing big perf gains on console. There are some maths that are incomplete, especially PrepareGPULightdata, I had to add some code to keep it working. I'll try to bundle that optimization on this same PR and then we should see an overall gain.

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
```diff
@@ -150,6 +151,233 @@ public static void CopyTo<T>(this T[] list, void* dest, int count)
                 UnsafeUtility.WriteArrayElement<T>(dest, i, list[i]);
         }
 
+        private static void RadixCalculateRadixParams(int radixBits, out int bitStates)
+        {
+            bitStates = 1 << radixBits;
+            if (radixBits != 2 && radixBits != 4 && radixBits != 8)
```
RSlysz, line 157: Perhaps it is just me but I'm used to check before using. What if it is negative or even NaN then ?

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
```diff
@@ -150,6 +151,233 @@ public static void CopyTo<T>(this T[] list, void* dest, int count)
                 UnsafeUtility.WriteArrayElement<T>(dest, i, list[i]);
         }
 
+        private static void RadixCalculateRadixParams(int radixBits, out int bitStates)
```
RSlysz, line 154: To some extend, each method "calculate" or "compute" something.
Plus here there is two time "Radix". And the order of words are not compliant with standard english. AFAIK we neither really do prefixing this way.
Proposition: `GetRadixParams` / `ComputeRadixParams` / `CalculateRadixParams`

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
```diff
@@ -150,6 +151,233 @@ public static void CopyTo<T>(this T[] list, void* dest, int count)
...
+        {
+            bitStates = 1 << radixBits;
+            if (radixBits != 2 && radixBits != 4 && radixBits != 8)
+                throw new Exception("Radix bits must be 2, 4 or 8 for uint radix sort.");
+        }
+
+        private static int RadixCalculateSupportSize(int bitStates, int arrayLength)
```
RSlysz, line 161: Proposition: `GetRadixSupportSize` / `ComputeRadixSupportSize` / `CalculateRadixSupportSize`

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -3506,6 +3475,60 @@ ShadowMapType shadowMapType
...
+            lightRenderData.barnDoorAngle = m_BarnDoorAngle;
+            lightRenderData.affectDiffuse = m_AffectDiffuse;
+            lightRenderData.affectSpecular = m_AffectSpecular;
+            lightRenderData.applyRangeAttenuation = m_ApplyRangeAttenuation;
+            lightRenderData.penumbraTint = m_PenumbraTint;
+            lightRenderData.interactsWithSky = m_InteractsWithSky;
+            lightRenderData.surfaceTint = m_SurfaceTint;
+            lightRenderData.shadowTint = m_ShadowTint;
+            lightRenderData.flareTint = m_FlareTint;
+        }
```
RSlysz, line 3518: If we add another field, I'm pretty sure we will forget to update the list here as we forgot to amend the Copy in the past.
You should add a test on that. It is not simple but I think you can do it the same way the Copy is tested (using the attributes and some reflection, see Editor tests)

RSlysz (approved):
Ok but:
- check functions names. We can do better on some.
- try to add test on the list of transferred fields in UpdateRenderEntity or expect several FB popping time to time on this.

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
kecho: Done! renamed all these to follow the convention you suggested.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
kecho: This one is interesting, since these fields shouldnt be read back by anything but the rendering pipeline in burst. Adding a field here arbitrarily could potentially hurt performance (since this is what is read on the burst side of things). I will add a comment explaining this. Basically, if a field is not used in the GPU, this should not be added for a field.

Thus, I think I should:
* Add a test indeed that checks the state of the GPU side array
* Add a comment explaining what to do, and when to add a field.
* Perhaps rename the HDLightCollection to something like HDLightRenderDatabase or HDLightGpuDatabase. Something along those lines.

Let me know your thoughts. I left more comments in slack.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2764,30 +1896,21 @@ void PrepareGPUProbeData(CommandBuffer cmd, HDCamera hdCamera, CullingResults cu
...
                         CreateBoxVolumeDataAndBound(localVolumetricFogList.bounds[i], LightCategory.LocalVolumetricFog, featureFlags, worldToViewCR, 0.0f, out LightVolumeData volumeData, out SFiniteLightBound bound);
-                        m_lightList.lightsPerView[viewIndex].lightVolumes.Add(volumeData);
-                        m_lightList.lightsPerView[viewIndex].bounds.Add(bound);
+
+                        m_GpuLightList.AddLightBounds(viewIndex, bound, volumeData);
                     }
                 }
 
-                m_TotalLightCount = m_lightList.lights.Count + m_lightList.envLights.Count + decalDatasCount + m_LocalVolumetricFogCount;
+                m_TotalLightCount = m_GpuLightList.lightsCount + m_lightList.envLights.Count + decalDatasCount + m_LocalVolumetricFogCount;
```
kecho, line 1904: @sebastienlagarde as explained in slack, this is a very nice thing that we also get in CPU.
This entire copy is removed, since this implementation has the light bounds and lights per view bounds in a single array (not using c# List<>).
As a consequence I have to pass an extra offset  in the GPU (the offset must be different from the light count, because is the light count that comes from the internal culler, the visibleLightCount). THis is because the processed lights could do extra culling.
Anyway let me know your thoughts and please review the GPU updates of this offse.t

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
```diff
@@ -150,6 +151,233 @@ public static void CopyTo<T>(this T[] list, void* dest, int count)
                 UnsafeUtility.WriteArrayElement<T>(dest, i, list[i]);
         }
 
+        private static void CalculateRadixParams(int radixBits, out int bitStates)
```
JulienIgnace-Unity, line 154: Since we are adding some Sorting functions to Core I think it would be good to have at least some very basic unit tests written for these.

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
```diff
@@ -150,6 +151,233 @@ public static void CopyTo<T>(this T[] list, void* dest, int count)
...
+                        array[m] = support[m];
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Merge sort - non recursive
+        /// </summary>
+        /// <param name="arr">array to sort.</param>
```
FrancescoC-unity, line 222: Super minor nitpick, but descriptions should all start capitalized (e.g. **A**rray to sort). 

Same applies to lot of things in this file, gonna only mark this here :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.Jobs.cs
```diff
@@ -0,0 +1,910 @@
...
+            private void IncrementCounter(HDGpuLightsBuilder.GPULightTypeCountSlots counterSlot)
+            {
+                unsafe
+                {
+                    int* ptr = (int*)gpuLightCounters.GetUnsafePtr<int>() + (int)counterSlot;
+                    Interlocked.Increment(ref UnsafeUtility.AsRef<int>(ptr));
+                }
+            }
+
+            public static void ConvertLightToGPUFormat(
```
FrancescoC-unity, line 151: For review sake, is it safe to assume the code this is essentially a move of the code previously in LightLoop.cs? Or is it better to look at the full code again?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.Jobs.cs
```diff
@@ -0,0 +1,910 @@
...
+            private void ConvertLightToGPUFormat(
+                int outputIndex, int lightIndex,
+                LightCategory lightCategory, GPULightType gpuLightType, LightVolumeType lightVolumeType)
+            {
+                var light = visibleLights[lightIndex];
+                var processedEntity = processedEntities[lightIndex];
+                var lightData = new LightData();
+                ref HDLightRenderData lightRenderData = ref GetLightData(processedEntity.dataIndex);
+
+                ConvertLightToGPUFormat(
```
FrancescoC-unity, line 342: I might be missing something (will need to chat via slack when back), but isn't most of what is computed below on lightData already computed by this function on the value passed as ref?

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
kecho: Sounds good.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDLightRenderDatabase.cs
```diff
@@ -0,0 +1,354 @@
...
+        private GameObject[] m_AOVGameObjects = null;
+        private HDAdditionalLightData[] m_HDAdditionalLightData = null;
+        private TransformAccessArray m_LightTransforms;
+
+        #endregion
+
+        private void ResizeArrays()
+        {
+            if (m_HDAdditionalLightData == null)
+                m_HDAdditionalLightData = new HDAdditionalLightData[m_Capacity];
```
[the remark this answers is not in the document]
kecho: Switched these to dynamic arrays.

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
kecho: Done!

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.Jobs.cs
kecho: Its a bit more than that. I will chat with you tomorrow over to look structurally at what was done.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.Jobs.cs
kecho: Yes good catch. When I moved the contents into the function i forgot to delete this. Good catch!

com.unity.render-pipelines.core/Runtime/Common/CoreUnsafeUtils.cs
kecho: Unit tests done! I had no idea we had functional tests :) will use them more often now.

kecho:
@LaserYGD this PR is now on its final form. I have tested your scene on my I9 PC and i can see gains.
To test on your side I recommend:
* Compare between master and this branch. Have burst always enabled on this branch. 
* Dont compare this branch with burst on/off. That is something I will certainly investigate later.  I am more interested in this branch vs vanilla master. 
* Let me know what sort of gains you see on your side.
* If you want to further see performance gains, I recommend creating a player using IL2CPP with highest performance optimizations and compare.

LaserYGD:
Tests are done on github master and this branch
Numbers are eye averaged over +30 frames, so will not have literal accuracy
Tests done on HDRP default template, and the medieval scene. 
HDRP template tests have very few lights, tested at spawn without moving. 
The focus is more on the medieval scene. 

Red = master branch. 
Green = pull branch. 
```diff
- Medieval scene with master (201 point lights)**: 
total: 1.00ms
ProcessVisibleLight: 0.6ms
PrepareGPUlightdata: 0.35ms

+ Medieval scene with pull (201 point lights)**: 
total: 0.47ms!
ProcessVisibleLights: 0.30ms (2x improvement)!
PrepareGPULightData: 0.15ms (2x improvement)!

- Medieval scene with master (67 point lights)**: 
total: 0.4ms
processVisibleLight: 0.2-0.25ms
PrepareGPUlightdata: 0.14ms

+ Medieval scene with pull (67 point lights)**: 
total avg: 0.25ms (1.6x improvement)!
processVisibleLights: 0.12 
PrepareGPULightData: 0.06ms

- Template with master**: 
total: 0.55-0.65ms
ProcessVisibleLight: 0.23ms
PrepereGPUlightData: 0.12ms
PrepareGPUProbeData: 0.9-11ms

+ Template with pull**: 
total: 0.55-0.65ms 
(rest is about the same)

Misc tests: 
The following tests are most likely not relevant as you mentioned before and you know about it, but alas, I had to 😅 

- Medieval scene with master (201 point lights. SHADOWS ENABLED for 134 point lights)**: 
total: 87ms
processVisibleLight: 85ms
PrepareGPUlightdata: 1.1-1.2ms

+ Medieval scene with pull (201 point lights. SHADOWS ENABLED for 134 point lights)**: 
total: 93ms 
Process shadows: 92.5ms
```

Overall, it's looking very nice if I say so myself. Especially if your project has more than a few dynamic lights, the difference becomes fairly big. Unfortunately I didn't get to do il2cpp tests, but I just might a day or two later.

kecho:
I believe I pass all the tests, there are some known instabilities:
* HDRP_GraphicTestRunner.1225_Lit_SpeedTree8SG
* Crash in WorkerManagerASIO::HandleWorkerEvent at shutdown
* Mac Metal: Invalid Pipeline state on Texture copy

Apart from that we are all green

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.Types.cs
```diff
@@ -245,14 +245,17 @@ public HDLightType type
                         case HDLightType.Directional:
                             legacyLight.type = LightType.Directional;
                             m_PointlightHDType = PointLightHDType.Punctual;
+                            HDLightRenderDatabase.instance.GetLightData(lightEntity).pointLightType = m_PointlightHDType;
```
FrancescoC-unity, line 248: More of a discussion point rather than a review comment per se, but should we maybe have the getter for a _modifiable_ pointer named differently? I wonder if we can find a way to have this more explicitly showing that we are actually modifying underlying data. 

I don't think how it is now is necessarily bad, but I wonder if we could make this more explicit in showing that it could lead to modifications too.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1296,6 +1355,8 @@ public float shadowDimmer
                     return;
 
                 m_ShadowDimmer = Mathf.Clamp01(value);
+                if (lightEntity.valid)
```
FrancescoC-unity, line 1358: Putting here as a random place in code. Given how often we do the validity check, what are the situations in which the lightEntity is invalid, but the lighting data is perfectly valid? 
Can we enforce validity as early as possible and get rid of all the checks?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDLightRenderDatabase.cs
```diff
@@ -0,0 +1,339 @@
...
+    //Light rendering entity. This struct acts as a handle to set / get light render information into the database.
+    internal struct HDLightRenderEntity
+    {
+        public int entityIndex;
+        public static readonly HDLightRenderEntity Invalid = new HDLightRenderEntity() { entityIndex = HDLightRenderDatabase.InvalidDataIndex };
+        public bool valid { get { return entityIndex != HDLightRenderDatabase.InvalidDataIndex; } }
+    }
+
+    //Data of the lights inside the database.
+    internal struct HDLightRenderData
```
FrancescoC-unity, line 20: As we discussed yesterday, at some point, after this first round, would be worth evaluating the ordering of these fields if not moving to a SoA pattern (though that is likely much more work)

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.Jobs.cs
```diff
@@ -0,0 +1,748 @@
...
+            public float airScaleHeight;
+            [ReadOnly]
+            public float aerosolScaleHeight;
+            [ReadOnly]
+            public Vector3 airExtinctionCoefficient;
+            [ReadOnly]
+            public float aerosolExtinctionCoefficient;
+            #endregion
+
+            #region input light entity SoA data
```
FrancescoC-unity, line 81: This is a AoS though? Is the region name a sort-of TODO or I am missing something?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.Types.cs
FrancescoC-unity: I think even GetLightDataRef() could be a bit better? Though not a great name :D

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.LightLoop.cs
```diff
@@ -0,0 +1,483 @@
...
+                m_LightsPerView.ResizeArray(m_LighsPerViewCapacity);
+            }
+
+            m_LightsPerViewCount = viewCounts;
+
+            int totalBoundsCount = maxLightCount * viewCounts;
+            int requestedBoundsCount = Math.Max(totalBoundsCount, 1);
+            if (requestedBoundsCount > m_LightBoundsCapacity)
+            {
+                m_LightBoundsCapacity = Math.Max(Math.Max(m_LightBoundsCapacity * 2, requestedBoundsCount), ArrayCapacity);
```
FrancescoC-unity, line 28: Is the *2 here for the two eyes? I think we in theory support arbitrary view counts for XR

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.LightLoop.cs
```diff
@@ -0,0 +1,483 @@
...
+
+                int dataIndex = visibleLights.visibleLightEntityDataIndices[lightIndex];
+                if (dataIndex == HDLightRenderDatabase.InvalidDataIndex)
+                    continue;
+
+                HDAdditionalLightData additionalLightData = lightEntities.hdAdditionalLightData[dataIndex];
+                if (additionalLightData == null)
+                    continue;
+
+                //We utilize a ray light data pointer to avoid copying the entire structure
```
FrancescoC-unity, line 431: raw and not ray I assume

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
```diff
@@ -159,7 +160,7 @@ static void ClearLightLists(BuildGPULightListPassData data, CommandBuffer cmd)
                 // Also, we clear all the lists and to be resilient to changes in pipeline.
                 if (data.runBigTilePrepass)
                     ClearLightList(data, cmd, data.output.bigTileLightList);
-                if (data.lightList != null) // This can happen for probe volume light list build where we only generate clusters.
+                if (data.canClearLightList) // This can happen for probe volume light list build where we only generate clusters.
```
FrancescoC-unity, line 163: Note mostly for myself: we don't have a probe volume light list build, so is this needed? If so, update comment.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.Types.cs
kecho: Replaced this with EditLightDataAsRef(). Added another one called GetLightDataAsRef().
Because c# cant have readonly refs, at the least put these methods for perf & convension.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
kecho: So this is very very tricky.
A render entity gets created on OnEnable, and destroyed during OnDisable. This has carefully been done, otherwise would require figuring out when to delete the entity (cant be done at destruction since this doenst clear the buffers). 
Because of OnEanble/OnDisable, say during upgrade scripts, or someone modifying a disabled gameobject through a script could potentially have lightEntity being invalid. These functions are rarely hit though for many many lights. I recommend paying for the cost here for that reason.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDLightRenderDatabase.cs
kecho: Indeed, I have added a comment to express that.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.Jobs.cs
kecho: Renamed this.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
kecho: Commend updated

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
FrancescoC-unity: The functions might be actually be called a lot of times in user land as they are the way the user can drive light by script and so I expect multiple of them might be called per light. The overhead of the branch is not going to be massive (especially in the grand scope of things and our current state), but still it'd be nice to think of it at some point. 

I do see the point, but worth thinking about it if we ever reach a stage of needing a further push with a collection of micro-optimizations. 
I assume on Awake() and OnDestroy() with some checks on whether the light is enabled or not when populating the processed list could work? You say OnDestroy() wouldn't clear the buffer, what buffers? 

Regardless this is def. not a blocking point for me, just discussion :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.LightLoop.cs
FrancescoC-unity: Missing this ^ @kecho

FrancescoC-unity (approved):
LGTM with the various discussed points as future steps (non-blocking for this PR)

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDGpuLightsBuilder.LightLoop.cs
kecho: on it!!

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
kecho: The buffers I talk about are the NativeArray<> holding the entity datas.
These need to live outside the HDRP Pipeline lifetime. Because the pipeline gets instantiated and destroyed outside OnAwake / OnDestroy etc.
Thus, we Dispose the global NativeArray<> containing all the entity data when the last HDAdditionalLightObject gets destroyed.
Every object calls OnDestroy always, even if you kill the editor / reload scripts.

tlaedre:
Great work! Really wish someone would slap the backport-to-2021.2 label on this one 🤞

berniegp:
@LaserYGD check out the [Profile Analyzer](https://docs.unity3d.com/Packages/com.unity.performance.profile-analyzer@1.1/manual/index.html) to get all those stats averaged for you over a range of frames. It also does comparisons of captures.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.Types.cs
```diff
@@ -466,49 +473,55 @@ public static bool IsValidLightUnitForType(HDLightType type, SpotLightShape spot
...
-                    // not supported directly. Convert now to equivalent if not default HDAdditionalLightData:
-                    if (this != HDUtils.s_DefaultHDAdditionalLightData)
-                    {
-                        legacyLight.type = LightType.Point;
-                        m_PointlightHDType = PointLightHDType.Area;
-                        m_AreaLightShape = AreaLightShape.Rectangle;
-
-                        //sanitycheck on the baking mode first time we add additionalLightData
-#if UNITY_EDITOR
-                        legacyLight.lightmapBakeType = LightmapBakeType.Realtime;
```
sebastienlagarde, line 504: what the reaons to remove this part? or it have move somewhere else?

sebastienlagarde: ah seen it

PR #5297, opened 2021-08-04
[URP] Automatic Intermediate Texture handling

pbbastian:
### Purpose of this PR
Previously, URP would force rendering to go through an intermediate renderer if the Renderer had any Renderer Features active. On some platforms, this has significant performance implications. Due to that, Renderer Features are now expected to declare their inputs using `ScriptableRenderPass.ConfigureInput`. This information is used to decide automatically whether rendering via an intermediate texture is necessary.

For compatibility reasons, a new property **Intermediate Texture** has been added to the Universal Renderer. This allows for either using the new behaviour, or to force the use of an intermediate texture. The latter should only be used if a Renderer Feature does not declare its inputs using `ScriptableRenderPass.ConfigureInput`.

All existing Universal Renderer assets that were using any Renderer Features (excluding those included with URP) are upgraded to force the use of an intermediate texture, such that existing setups will continue to work correctly. Any newly created Universal Renderer assets will default to the new behaviour.

Screenshot 2021-08-04 at 15 23 49

---
### Testing status
Verified the behaviour of the option and the asset upgrades locally.

---
### Comments to reviewers
Please do test the upgrading, and see if it acts as expected.


TheoWong-pixel:
Would suggest that we have the explanation about Always for the docs. we normally just have a generalised description of the feature and what it does, maybe speak more to that and explain the performance implications more without referencing the options.

thomas-zeng (approved):
LGTM! Thank you!

thomas-zeng:
We will need this PR for 2021.2(requested by msft hololens). The branch off for 2021.2 is scheduled for Sep 1st and it's likely the PR will land afterward. So we will need to backport this PR to 2021.2 branch once it lands.

ernestasKupciunas (approved):
Tested the upgradability, works as expected.
Also ran Boat Attack and URP Rendering Examples builds in the Editor and MacStandalone build with _Always_ and _When Needed_, seems to work as intended too.

Device under testing:
MacBook Pro (15-inch, 2017), 2,8 GHz Quad-Core Intel Core i7, 16 GB 2133 MHz LPDDR3, Radeon Pro 555 2 GB, macOS Big Sur 11.5.1 (20G80)

Unity used for testing:
Version: 2022.1.0a8.1153 Personal
Revision: trunk a6d7f27874c0
Built: Sat, 04 Sep 2021 19:12:24 GMT

Good job with the PR description and attached images!

phi-lira (approved):

TheoWong-pixel (approved):

oleks-k (approved):
Looks good.

pbbastian:
I've gone through all the failing jobs, and verified that they're failing in the same way on master.

PR #5305, opened 2021-08-05
[URP] Move depth copy pass after the transparents pass

manuele-bonanno:
### Purpose of this PR
Currently, when the "Depth Texture" option is selected, the scene depth is copied to _CameraDepthTexture after the Opaques pass. Because of this Unity is unable to merge compatible render passes (opaque, skybox, transparents) using a lot of extra bandwidth due to having to store/load RTs of intermediate passes.

This PR adds the option to copy after the Transparents pass, allowing opaque,skybox and transparents pass to be merged ina. single one, keeping all the intermediate RTs results in tile memory.

image

**NOTE:**
To not potentially break existing projects, the Copy Depth Mode is set to AfterOpaques for existing UniversalRendererData assets. New assets will default to the optimized AfterTransparents setting.

This PR also adds depth resolve support, currently disabled because relies on 22.1 trunk features. So it will be fully enabled after the 22.1 branching in a separate PR.

---
### Testing status
Local (forward, deferred, native render pass) on iOS
Yamato: ABV

---
### Comments to reviewers

Before (Opaques+Skybox+Transparents):
image

After (merged passes and depth resolved. Only the resolved targets are stored, MSAA'd ones are discarded):
image

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRendererData.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRendererData.cs
@@ -89,7 +89,7 @@ namespace UnityEngine.Rendering.Universal
 
         public ShaderResources shaders = null;
 
-        const int k_LatestAssetVersion = 1;
+        const int k_LatestAssetVersion = 2;
         [SerializeField] int m_AssetVersion = 0;
         [SerializeField] LayerMask m_OpaqueLayerMask = -1;
         [SerializeField] LayerMask m_TransparentLayerMask = -1;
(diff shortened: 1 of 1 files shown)


TheoWong-pixel (approved):

jonuuukas (approved):

manuele-bonanno:
the test failures are the same as the master branch

simon-engelbrecht-soerensen (approved):
Tested and working well. One thought is to add to the documentation that passes/renderer features that happen before transparent should be moved to after, if using depth? Maybe :)

phi-lira (approved):

PR #5307, opened 2021-08-05
Highlighter for URP and HDRP settings

adrien-de-tocqueville:
### Purpose of this PR

This PR adds a button to jump to correct HDRP asset section to enable some setting if needed, this is easier than having to navigate through all settings and search manually.

5307_highlight

On URP side, the feature has been implemented for the decal component. See URP QA doc:
https://docs.google.com/document/d/16pDMcIwfqQkFF6YUmsO04KsacB57y1DPmWsyEdKE2mA/edit?usp=sharing

### Testing status

HDRP coverage is done, all helpbox that could benefit from this feature have been updated.


RSlysz (approved):
Seams ok to me beside small code convention issue.
But I'm not UX.

sebastienlagarde (approved):

remi-chapelain (approved):
I really like this change. 🎈 
It will prevent everyone from wasting time trying to find a setting among billions checkboxes :). 

**_I allowed myself to fix some tiny mistakes to be able to approve it._**

I tested all the properties touched (which I believe could have been mentioned in the PR description): 
- Decal Projector ✔️ 
- SSAO ✔️ 
- SSR ✔️ 
- SSGI (has been forgotten but now fixed) ✔️
- Raytracing mixed tracing mode ✔️
- Volumetric fog ✔️ 
- No HDRP Asset in Graphic Setting (led you to quality instead of graphics, now fixed) ✔️ 
- Material inspector missing diffusion profile in HDRP asset ✔️ 

I tested with the project settings windows opened and closed ✔️ 

I also tested removing the HDRP asset to see if there was some kind of warning. 
There is one saying the property couldn't been found.
It could be clearer as to "why" but it may be a generic warning, so I guess it's fine. 
image

Even though, I will approve this PR, I believe we can go further in another PR, 
---
Here's a few thoughts: 
- Since we can do that now, I believe we could do the same (or just a warning instead of a red error) if the **Default Frame Settings** is not enabled for this property (ex: SSGI is off by default in the frame settings) to help users even more. This is a common mistake that can be easily avoided with a warning and a way to fix it quickly.   
- I noticed there's at least 3 features which you can activate even if they are disabled in the HDRP asset (**Custom pass**, **Dynamic Resolution**, **Subsurface Scattering** volume override). I believe this can lead to users thinking they can enable something but since it's already off in the asset, they won't know why it's not working. Implementing this info box for those would help here too. 
- This one can be debatable, but there is some option/feature we don't display on the inspector if they are disabled in the asset (which is kind of the opposite of the previous point, i.e displaying and being able to enable feature that are already disabled); For example, we hide **Volumetric Fog**, **Lights Layers**, or **Screen Space Shadows** if they are not enabled. This could probably benefit from some coherence.

remi-chapelain:
> won't a button that just turns Ambient Occlusion On be faster/simpler?
faster/simpler yes. But with this window open > highlight etc.. we're not just giving users a fish, we're teaching them how to catch the fish and I think that has some real value as well (if that makes sense)

adrien-de-tocqueville:
Here's my take on @TheoWong-pixel's comment:

> Given that this is a great idea, this is not consistent with any error messaging that we have in the editor. if we implemented this here we would have to make sure that we have a pass on all similar situations and implement the same feature. This would mean looking across SRPs. This is especially important with URP and HDRP alignment.

Indeed, this is currently just a proposition, but if the system is approved we can apply it to everywhere needed (at least for SRPs)

> Most commonly we include in the error message that says "enabled in the render pipeline asset" so users know where to enable this.

I'd say the problem here is that the render pipeline asset is so big with so many sections that it doesn't help much to say "look in the render pipeline asset" 

> Also with this even if they clicked 'open' if the setting is not on screen, users would miss the flash no?

No, the correct window will be opened, the relevant foldouts unfolded, and the view scrolled to the correct place. Then the highlight will last for a few seconds (see the gif in the description)

> Also the flash is inconsistent. with how we normally high items in the UI.

I used the builtin editor highlight feature so not sure what you refer to with "how we normally high items in the UI", maybe i missed something ?

> Similar to this, the URP side already has the 'Fix' Button that is already in. (also not approved by UX). if we are looking to have a similar interaction i would advise that they are shared on both URP and HDRP and have his live in Core so that when UX does have a chance to have a full pass on error messaging we can change them in one location.

Just to be clear, this is not a "fix" button, often the user will want to have some features disabled on purpose.
We could have an "enable" button, but i don't see it as a better solution (although maybe more consistent with the rest of UX). It feels like black magic and doesn't intruct user on how to turn it back off or find related settings in the hdrp asset

TheoWong-pixel:
Ah, okay, so  long as we are doing a full pass of this in both SRPs then i am fine with this implementation. Please include the relevent people in this like @Verasl and @martint-unity To help advise where this needs to be implemented. I would assume that this replaces the "Fix me" buttons in URP. I agree that not 'Fix'-ing with auto magic is much better. 

I was imagining the flash as we use for the 'Additional Settings' maybe @JulienIgnace-Unity or @alex-vazquez would know. 

Love it! Thanks for the hard work!

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/QualitySettingsPanel.cs
```diff
@@ -97,6 +103,17 @@ public void OnGUI(string searchContext)
...
+                    if (Highlighter.active)
+                    {
+                        if (Highlighter.activeVisible)
+                        {
+                            if (Time.realtimeSinceStartup - m_HighlightStart > s_HighlightDuration)
+                                Highlighter.Stop();
+                        }
+                        else
+                            m_HighlightStart = Time.realtimeSinceStartup;
+                    }
```
alex-vazquez, line 116: I wonder if we could extract this functionality somewhere on CoreEditorUtils, It might be interesting to have it easily for the future.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/QualitySettingsPanel.cs
```diff
@@ -70,6 +72,10 @@ public void OnActivate(string searchContext, VisualElement rootElement)
             {
                 m_HDRPAssets.Clear();
                 PopulateHDRPAssetsFromQualitySettings(m_HDRPAssets);
+
+                var currentAsset = QualitySettings.GetRenderPipelineAssetAt(QualitySettings.GetQualityLevel());
```
jenniferd-unity, line 76: `QualitySettings.GetRenderPipelineAssetAt(QualitySettings.GetQualityLevel());` is equivalent to `QualitySettings.currentPipeline;`
However it may not be the one in use if null.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDEditorUtils.cs
```diff
@@ -282,6 +282,26 @@ internal static void HandlePrefixLabelWithIndent(Rect totalPosition, Rect labelP
...
+        internal static void QualitySettingsHelpBox(string message, MessageType type, HDRenderPipelineUI.Expandable uiSection, string propertyPath)
+        {
+            if (CoreEditorUtils.HelpBoxWithButton(message, type))
+            {
+                SettingsService.OpenProjectSettings("Project/Quality/HDRP");
+                HDRenderPipelineUI.Inspector.Expand((int)uiSection);
+                Highlighter.Highlight("Project Settings", propertyPath, HighlightSearchMode.Identifier);
+                GUIUtility.ExitGUI();
+            }
+        }
```
alex-vazquez, line 295: I would highly recommend to create a generic method in core.

something like: 

```
public static void DrawHelpBoxWithHighligt(string message, MessageType type, CoreEditorDrawer<T>, int section, string panelPath, string propertyPath)
```

And implement them here with the `Quality/HDRP` and `Graphics/HDRP Global Settings`

alex-vazquez (changes requested):
I think you did a great job, but I think this could be really beneficial for URP so, I think we should think about that.

I put some comments about the API, I think the way we had with the lamdas, was easier to use and I would continue using the same approach, so feel free to modify the current API or add another one with the same behaviour as we did.

About the HDEditorUtils methods, I would like to include them somehow on core, and be generic as possible, at the end we are referencing things from the CoreEditorDrawers, so I believe it should not be hard to change it.

I think all my comments should be tackled on this PR, as this might introduce public API, that we might potentially change in the future.

Many thanks for the hard work.

alex-vazquez:
@TheoWong-pixel , the flash we use at the additonal properties is a custom background that is fade over time. Do we want to have the same system across everywhere?

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -129,6 +114,51 @@ public static void DrawFixMeBox(string text, MessageType messageType, Action act
         {
             DrawFixMeBox(EditorGUIUtility.TrTextContentWithIcon(text, CoreEditorStyles.GetMessageTypeIcon(messageType)), action);
         }
+        
+        /// <summary>Draw a help box with a button.</summary>
+        /// <param name="message">The message text.</param>
+        /// <param name="type">The type of the message.</param>
+        /// <param name="buttonLabel">The button text.</param>
+        /// <returns>True if the button is clicked</returns>
+        public static bool HelpBoxWithButton(string message, MessageType type, string buttonLabel = "Open")
```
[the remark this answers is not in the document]
adrien-de-tocqueville: Ok, i wanted to have a similar signature as the `EditorGUILayout.HelpBox` function, and I also personally prefer checking to lambdas but i don't really mind, I can change it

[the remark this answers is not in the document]
alex-vazquez: thank you :)

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDEditorUtils.cs
```diff
@@ -282,6 +282,26 @@ internal static void HandlePrefixLabelWithIndent(Rect totalPosition, Rect labelP
         internal static bool IsPresetEditor(UnityEditor.Editor editor)
         {
             return (int)((editor.target as Component).gameObject.hideFlags) == 93;
+        }?
+        internal static void QualitySettingsHelpBox(string message, MessageType type, HDRenderPipelineUI.Expandable uiSection, string propertyPath)
```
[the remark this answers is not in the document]
adrien-de-tocqueville: actually after all I think it's better that way, because that function is just a helper to expand the quality settings foldouts, so we indeed want it to be tied to the type of expandables supported by the quality settings

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDEditorUtils.cs
adrien-de-tocqueville: if it's specific to HDRP i am not sure we should put it in core ? i propose to leave it here for now and see what can be shared once URP support is added

TheoWong-pixel:
If possible yes

> @TheoWong-pixel , the flash we use at the additonal properties is a custom background that is fade over time. Do we want to have the same system across everywhere?

If possible yes.

adrien-de-tocqueville:
@TheoWong-pixel  Sadly I can't really do the exact same highlight as the volume, that's what I managed to do, is it ok ?
highlight

Volume highlight for reference:
highlight_volume

TheoWong-pixel:
Yup thats fine, similar is good :)

TheoWong-pixel (approved):

adrien-de-tocqueville:
@TheoWong-pixel This is in 🙂

adrien-de-tocqueville:
@TheoWong-pixel Also what would you think of a context menu item like that when right clicking some properties ? I guess this can help understand why some settings are disabled or don't work. Maybe we could also have another item that points to the default framesettings.
Maybe that's getting a bit specific but i think it can lead to a lot of time saving
image

TheoWong-pixel:
@adrien-de-tocqueville So this would introduce a very new behavior and interaction into Unity. This might not be very discoverable for a lot of users as well as strange if it is only applied to certain things that are applicable to SRPs but not anywhere else in the editor. We have greyed out fields in almost every corner of the editor. If we introduce this and users get used to this, then we would have to ensure that we provide this menu item that solves problems for any setting that has been greyed out and in some cases we cant provide a link. 

if we only apply this for HDRP it would further segregate HDRP from the rest of Unity.

adrien-de-tocqueville:
@TheoWong-pixel  Yes I have the same concerns, and I am not saying it should be part of this PR because that would add too much work, but I think there should be a way to make links between properties and HDRP settings without having to rely on docs/tooltips or big help boxes.
I guess i just got tired of looking through the bunch of different settings menu and foldouts we have so I am trying things 😄

phi-lira:
Adding @martint-unity for awareness.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -89,45 +89,153 @@ public static Texture2D CreateColoredTexture2D(Color color, string textureName)
...
+        {
+            DrawFixMeBox(EditorGUIUtility.TrTextContentWithIcon(message, CoreEditorStyles.GetMessageTypeIcon(messageType)), buttonLabel, action);
+        }
 
-                    GUILayout.FlexibleSpace();
-                }
-            }
-            GUILayout.Space(5);
+        /// <summary>Draw a help box with the Fix button.</summary>
+        /// <param name="message">The message with icon if need.</param>
```
martint-unity, line 198: if needed

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -89,45 +89,153 @@ public static Texture2D CreateColoredTexture2D(Color color, string textureName)
...
+        /// <param name="action">When the user clicks the button, Unity performs this action.</param>
+        public static void DrawFixMeBox(GUIContent message, Action action)
+        {
+            DrawFixMeBox(message, "Fix", action);
         }
 
         /// <summary>Draw a help box with the Fix button.</summary>
-        /// <param name="text">The message text.</param>
-        /// <param name="messageType">The type of the message.</param>
+        /// <param name="message">The message with icon if need.</param>
```
martint-unity, line 206: if needed

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -89,45 +89,153 @@ public static Texture2D CreateColoredTexture2D(Color color, string textureName)
...
+        static float s_HighlightStart = -1.0f;
+        static Texture2D s_HighlightBackground;
+        static object s_View;
+
+        static FieldInfo k_ViewInfo = typeof(Highlighter).GetField("s_View", BindingFlags.Static | BindingFlags.NonPublic);
+        static FieldInfo k_HighlightStyleInfo = typeof(Highlighter).GetField("s_HighlightStyle", BindingFlags.Static | BindingFlags.NonPublic);
+        static FieldInfo k_WindowBackendInfo = Type.GetType("UnityEditor.GUIView,UnityEditor").GetField("m_WindowBackend", BindingFlags.NonPublic | BindingFlags.Instance);
+        static EventInfo k_GUIHandlerInfo = Type.GetType("UnityEditor.UIElements.DefaultEditorWindowBackend,UnityEditor").GetEvent("overlayGUIHandler", (BindingFlags)(-1));
+        static MethodInfo k_Repaint = Type.GetType("UnityEditor.GUIView,UnityEditor").GetMethod("Repaint", (BindingFlags)(-1));
+
```
martint-unity, line 102: According to this, afaik 
static == s_
const == k_
https://ono.unity3d.com/unity-extra/unity-meta/files/@/ReferenceSource/CSharp/Assets/CSharpReference.cs
- Prefixes: m_ = instance field, s_ = static readwrite field, k_ = const

So it looks a bit confusing to me

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -89,45 +89,153 @@ public static Texture2D CreateColoredTexture2D(Color color, string textureName)
...
+        /// <param name="message">The message with icon if need.</param>
+        /// <param name="buttonLabel">The button text.</param>
         /// <param name="action">When the user clicks the button, Unity performs this action.</param>
-        public static void DrawFixMeBox(string text, MessageType messageType, Action action)
+        public static void DrawFixMeBox(GUIContent message, string buttonLabel, Action action)
         {
-            DrawFixMeBox(EditorGUIUtility.TrTextContentWithIcon(text, CoreEditorStyles.GetMessageTypeIcon(messageType)), action);
+            EditorGUILayout.BeginHorizontal();
+
+            float indent = EditorGUI.indentLevel * 15 - EditorStyles.helpBox.margin.left;
```
martint-unity, line 213: what is this 15? 
We tend to try to not have magic numbers but make a variable out of them and name them properly, like const  k_IndentMargin =15; Or something similar.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
```diff
@@ -89,45 +89,153 @@ public static Texture2D CreateColoredTexture2D(Color color, string textureName)
...
+
+            if (Event.current.type == EventType.Repaint)
+                EditorStyles.helpBox.Draw(boxRect, false, false, false, false);
+
+            Rect labelRect = new Rect(boxRect.x + 4, boxRect.y + 3, rect.width - 8, rect.height);
+            EditorGUI.LabelField(labelRect, message, CoreEditorStyles.helpBox);
+
+            var buttonRect = leftRect;
+            buttonRect.x += rect.width - 2;
+            buttonRect.y = rect.yMin + (rect.height - EditorGUIUtility.singleLineHeight) / 2;
```
martint-unity, line 231: Seems to be a few of those magic numbers here that might maybe be better if they are variables with a nice name.

martint-unity (changes requested):
I have a few comments.

com.unity.render-pipelines.core/Editor/CoreEditorUtils.cs
adrien-de-tocqueville: there's no real name i could give to those numbers, they are just padding to fix the appearance. I could call them paddingBoxLeft, Right and so on but that would make it much better i think

martint-unity (approved):
LGTM

martint-unity:
So now we need a task to use this in URP whenever this lands.

ernestasKupciunas (approved):
Full testing doc: https://docs.google.com/document/d/16pDMcIwfqQkFF6YUmsO04KsacB57y1DPmWsyEdKE2mA/edit?usp=sharing

There is one place that I've found in the URP that in my opinion probably would make sense to add the highlighter. Once the decal is created, it requires the Decal renderer feature added to the URP asset:
Screenshot 2021-09-24 at 11 10 00
Screenshot 2021-09-24 at 11 10 13

theo-pnv:
Hi, on the SDET side there seems to be merge conflicts on our files. It is weird given the changes this PR wants to introduce are already in master. Could we make sure this won't override anything? Thanks.

adrien-de-tocqueville:
@theo-pnv sorry about that, i messed something while merging. You don't need to review the PR it doesn't touch anything related to SDET

martint-unity (approved):
Code looks good to me. Would be good to add URP QA to this

phi-lira (approved):

PR #5316, opened 2021-08-06
[Universal] [fix] delete entry issue on Renderer List

martint-unity:
---
### Purpose of this PR
This is a fix for https://fogbugz.unity3d.com/f/cases/1347464/
For some reason there was a check that seemed a bit unnecessary. So only if the list entry had a Null entry it would need to update the Renderer Index. But I believe we should always check the index. 
We could do the index check if it is lower or higher directly as well if we want to. 

---
### Testing status
Yamato
---
### Comments to reviewers
Would be good to Double check the reason why we added this check in the first place.


alex-vazquez (approved):

ellioman (approved):

Verasl (approved):
LGTM, shipit!

PR #5317, opened 2021-08-09
Fix invalid layer registration with PPv2

fredericv-unity3d:
---
### Purpose of this PR
Fixes case [1345869](https://fogbugz.unity3d.com/f/cases/1345869/)

Issue: 
1. The PPv2 relies on the `GameObject.layer` field to find the layer of `PostProcessVolume`
2. The `PostProcessManager` create an index of the volumes per layer lazily
3. The user cam change the `GameObject.layer` at anytime

This result in the index of the `PostProcessManager` to be incorrect if the user changes the `GameObject.layer` following specific sequences.

About the previous fix:
It does not solve the above issue, the following code does break it again:

```csharp
using UnityEngine;
using UnityEngine.Rendering;
using UnityEngine.Rendering.PostProcessing;

public class GrabSetOnLaterUpdate : MonoBehaviour
{
    public int LayerIndex = 8;
    
    void LateUpdate()
    {
        var cam = Camera.main;

        // change the layer
        gameObject.layer = LayerIndex; 

        // Change layermask of this layer
        var layer = cam.GetComponent<PostProcessLayer>();
        layer.volumeLayer = 1 << LayerIndex;
        
        // Trigger an update of the volume system
        layer.UpdateVolumeSystem(cam, new CommandBuffer());
        
        // change the layer
        gameObject.layer = LayerIndex + 1; 

        Destroy(gameObject);
    }
}
```

Solution:
* Refactoring the PostProcessVolume database in a separate class and using a copy of the layer used at registration to perform the lazy indexing

---
### Testing status
Manual test:
1. Followed the original steps to reproduce (case 1 and case 2)
2. >Observed there are no error

1. Used the above code on the master code and observed an error
2. Used it on the new code and observed no errors

Ran the automated tests on both master and the new branch:
- the failure are the same on both branches (0207, 0208, 0211, 0214, 0221)

---

### DIFF
--- a/com.unity.postprocessing/PostProcessing/Runtime/Utils/PostProcessVolumeDatabase.cs
+++ b/com.unity.postprocessing/PostProcessing/Runtime/Utils/PostProcessVolumeDatabase.cs
@@ -33,12 +33,12 @@ namespace UnityEngine.Rendering.PostProcessing
             }
         }
 
-        public static bool operator==(in PostProcessVolumeMeta l, in PostProcessVolumeMeta r)
+        public static bool operator ==(in PostProcessVolumeMeta l, in PostProcessVolumeMeta r)
         {
             return l.Equals(r);
         }
 
-        public static bool operator!=(in PostProcessVolumeMeta l, in PostProcessVolumeMeta r)
+        public static bool operator !=(in PostProcessVolumeMeta l, in PostProcessVolumeMeta r)
         {
             return !l.Equals(r);
         }
@@ -89,7 +89,7 @@ namespace UnityEngine.Rendering.PostProcessing
 
         public Exception Add(PostProcessVolume volume, PostProcessVolumeMeta meta)
         {
-            var entry = new Entry {volume = volume, meta = meta};
+            var entry = new Entry { volume = volume, meta = meta };
             var mask = 1 << meta.layer;
             m_Volumes.Add(entry);
 
@@ -199,6 +199,7 @@ namespace UnityEngine.Rendering.PostProcessing
                     RemoveEntryFromIndex(entry);
                     AddEntryInIndex(entry, newMeta.layer);
                 }
+
                 // We use else if as a Adding and Removing will update the priority
                 // ReSharper disable once CompareOfFloatsByEqualityOperator
                 else if (entry.meta.priority != newMeta.priority)


RemyUnity (approved):
Tests mentioned in the PR description seems sufficient to me, 🟢.

com.unity.postprocessing/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [3.1.3] -
```
sebastienlagarde, line 7: We haven't publish the 3.1.2 version, this is where the changelog should go. It is a bit misleading like this but it is requried to have this format: ## [3.1.2] - 2021-02-06 to pass Yamato test. When we produce a new version we bump it

adrien-de-tocqueville (approved):

com.unity.postprocessing/CHANGELOG.md
fredericv-unity3d: I moved the changelog to 3.1.2

com.unity.postprocessing/PostProcessing/Runtime/PostProcessManager.cs
```diff
@@ -116,52 +113,60 @@ public void GetActiveVolumes(PostProcessLayer layer, List<PostProcessVolume> res
...
-                    continue;
-
-                // Global volume always have influence
-                if (volume.isGlobal)
+                Exception e;
+                if ((e = m_Volumes.FindSortedVolumesByLayerMask(mask, volumes)) != null)
                 {
-                    results.Add(volume);
-                    continue;
+                    Debug.LogException(e);
```
jenniferd-unity, line 121: Should we have this in the final PR?

jenniferd-unity (changes requested):
I worry that this is implemented with new features on a maintenance-mode package. Is there a more hacky version of the fix?
Note: I see some Debug.Log that should also be removed in final version of this code?
Also please add a URP reviewer, they may have a similar issue if I recall.

com.unity.postprocessing/PostProcessing/Runtime/PostProcessManager.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine.Assertions;
+using UnityEngine.Pool;
```
Chman, line 5: Needs a version bump in the package info as it's currently set to 2019.4.19f. AFAIK that package was introduced in 2021? That's quite a bump for a bug fix :(

com.unity.postprocessing/PostProcessing/Runtime/PostProcessManager.cs
fredericv-unity3d: I prefer to log an error if any occurs instead of silently fail. So I would keep it in the final PR

com.unity.postprocessing/PostProcessing/Runtime/PostProcessManager.cs
fredericv-unity3d: I can workaround the use of `UnityEngine.Pool` to avoid the version bump.

fredericv-unity3d: Fixed in 8349088

jenniferd-unity (changes requested):
I would not introduce a refactor in a bugfix on a package in maintenance mode.
Also, I would encourage that we don't introduce exception handling in a code that can be used at runtime. We already have performance issues with the Volume system, I worry that the use of Dictionary, exception, and Linq is slowing things down.

Additionally, I do not understand the issue you are trying to solve. I would expect that if a user is changing a layer priority without explicitly updating the Volume system then it is expected for the system to not be aware of the change. I would consider this issue as a by-design issue. What do you think @Thomasch-unity3d ?

fredericv-unity3d:
@jenniferd-unity Regarding performance with exception:
- The `catch` statement does introduce a lot of overhead (tracking the stack)
- The `finalize` statement introduce a very small overhead, which is usually affordable (no stack tracked)

Here, I don't use any try/catch/finalize statement, so I do not introduce any performance overload. The `Exception` is returned like any other value and used only as an error message.

PR #5318, opened 2021-08-09
Fixed the clouds missing in the ambient probe and in the static and dynamic sky.

anisunity:
This PR adds the support of the volumetric clouds in the baking of the static sky, dynamic sky and ambient probe. It also generated a second probe without the clouds that is used to light the clouds themselfs as an environement lighting.
The clouds can be evaluated either at full resolution or quarter resolution and that can be controlled through the frame settings  of the camera.

Testing status
- Added new tests to cover the new functionnality.
- Updated the screenshots with the new changed.
- Made sure that the resolution of the dynamic sky correctly appears based on the frame setting.
- Made sure that the tint of the ambient probe is now correct on top and at the bottom of the clouds


RemyUnity:
Isn't it also worth adding a graphic test ?

anisunity:
@RemyUnity it's a draft, so the PR is not ready for QA review

anisunity:
image
Now the interpolation between top and bottom of the sky works just fine

anisunity:
image
The static and ambient probes work just fine

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricCloudsAccumulation.cs
```diff
@@ -166,7 +166,7 @@ VolumetricCloudsParameters_Accumulation PrepareVolumetricCloudsParameters_Accumu
         }
 
         static void TraceVolumetricClouds_Accumulation(CommandBuffer cmd, VolumetricCloudsParameters_Accumulation parameters,
-            RTHandle colorBuffer, RTHandle depthPyramid, TextureHandle motionVectors, TextureHandle volumetricLightingTexture, TextureHandle scatteringFallbackTexture,
+            RTHandle colorBuffer, RTHandle depthPyramid, RTHandle motionVectors, RTHandle volumetricLightingTexture, RTHandle scatteringFallbackTexture,
```
JulienIgnace-Unity, line 169: Why the change to RTHandle here?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricCloudsFullResolution.cs
```diff
@@ -77,7 +78,7 @@ VolumetricCloudsParameters_FullResolution PrepareVolumetricCloudsParameters_Full
         }
 
         static void TraceVolumetricClouds_FullResolution(CommandBuffer cmd, VolumetricCloudsParameters_FullResolution parameters,
-            RTHandle colorBuffer, RTHandle depthPyramid, TextureHandle volumetricLightingTexture, TextureHandle scatteringFallbackTexture,
+            RTHandle colorBuffer, RTHandle depthPyramid, RTHandle volumetricLightingTexture, RTHandle scatteringFallbackTexture,
```
JulienIgnace-Unity, line 81: Same question as above.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricCloudsSky.cs
```diff
@@ -0,0 +1,360 @@
...
+                RTHandles.Release(m_IntermediateCloudsLightingCube0Buffer);
+                RTHandles.Release(m_IntermediateCloudsLightingCube1Buffer);
+            }
+        }
+
+        void InitializeVolumetricCloudsStaticTextures()
+        {
+            if (m_Asset.currentPlatformRenderPipelineSettings.supportVolumetricClouds)
+            {
+                int skyResolution = (int)m_Asset.currentPlatformRenderPipelineSettings.lightLoopSettings.skyReflectionSize;
```
JulienIgnace-Unity, line 30: This buffers are only used to render the sky in order to compute ambient probe for the clouds right?
If that's the case, I think we should not use skyReflectionSize here because it can be huge (2k or 4k) and it would be very wasteful, even when this goes through render graph.
There are plans to add a new resolution setting for ambient only when possible but in the meantime, can we use a much smaller hardcoded fixed size here?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricCloudsAccumulation.cs
anisunity: it was a mistake, should have been rthandle all long

RemyUnity (approved):
I did a test scene to check if the clouds were visible in the ambiant probe, as well as in the sky reflections of the different rendering objects (camera/static&dynamic reflection probes/plana probes), and I all these I checked if the full res option was working.

✔

anisunity:
Managed to fix the metal issue, however an other PR needs to be merged first for us to generate the new screenshots.

JulienIgnace-Unity (approved):

anisunity:
Updated all the screenshots on all 5 platforms (OSX, Linux, WinDx11, WinDx12, WinVulkan) Standalone and playmode

PR #5319, opened 2021-08-09
[HDRP] Fix HD template (particularly visible in path tracing)

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1339852

Before: 2 of the 3 lamps would appear too dark, because light bulbs cast shadows (see screenshots in the corresp. ticket).
Now: bulbs are set as non-shadow casters, which makes no difference in raster mode, but makes all 3 lamps look brighter in path tracing.


JulienIgnace-Unity (approved):

iM0ve (changes requested):
The fix works, but may I suggest a different approach, changing the scene file is always very painful due to conflicts. I know there is currently HDRP Tutorial branch that is also working on the scene file so the merge to master will be complicated. 

What if we just updated the prefab instead?

https://user-images.githubusercontent.com/36502659/129024430-57f5e434-279a-4fd9-9095-77dd04121a20.mp4


com.unity.template-hd/Assets/SampleSceneAssets/Meshes/Props/Lights/Prefab/CeilingLamp_Lamp_01.prefab

eturquin:
> What if we just updated the prefab instead?
Sounds like a better, cleaner way to do it; let's do that instead. :)

iM0ve (approved):

PR #5321, opened 2021-08-09
[HDRP] [LightExplorer] Fix missing refreshwhen editing property with a running paused editor.

RSlysz:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1354129/
(Missing refresh of scene view when changing light in LightExplorer and editor is in run and paused.

FB is for ContactShadow but the same also apply to affectDiffuse and affectSpecular (also fixed here)

---
### Testing status
Manually tested (now we see the change in sceneview while editing from the LightExplorer). Tested with the directional light (sun) in the hd template while looking at the tree.

---
### Comments to reviewers
Note: only appears when running and pausing the game.


RemyUnity (approved):
Look good to me ! 🟢

PR #5323, opened 2021-08-10
Hd/fix framesettings ordering

RSlysz:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1355836/

Several PR have landed on the ProbeVolume. One of it added an hidding mechanisme. I am not fond of it as it may be more difficult now to add a new framesettings as we have some dynamically hidden. And this is also not compatible with the current UX workflow. But I understand the needs for this as it is experimental.

Then another FrameSettings was added with a dependency on the hidden one. But it was not hidden itself. So usual case, when the feature flag is disable, you have the dependant shown indented under somecing that have no relation with it (see below)
image

The fix here do:
- Make also dependency hidden the same way (see below)
- Add extra warning in the comment against the usage of hideUI
- Add code documentation on the AmmendInfo to clarify each argument usage as they are more and more here
- Grouped FrameSetting in code per group again. Initially, each group had an allotted page of bit to use but with time, devs haven't really used it this way. So only regroup them per group for now.

image or image

---
### Testing status
Manually tested the hidding (see screenshot above)

---
### Comments to reviewers
@FrancescoC-unity Have your PRs been backported ?

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -359,6 +359,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed incorrect light list indexing when TAA is enabled (case 1352444).
 - Fixed Additional Velocity for Alembic not taking correctly into account vertex animation
 - Fixed wrong LUT initialization in Wireframe mode.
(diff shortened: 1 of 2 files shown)


RemyUnity (approved):
Thank for the screenshots, I don't really see what we could test more.
LGTM 🟢

JulienIgnace-Unity (approved):

PR #5324, opened 2021-08-10
Fix first frame of SSR PRB Accumulation

skhiat:
**Purpose of this PR**

Improve the transition between state changing on SSR_PBR_Accumulation.
Keep the result from first hit one more frame to not converge by starting from black texture.

Before PR:
Before

After PR:
After

---
### Comments to reviewers


iM0ve (changes requested):
I have tried the repro project with this branch and the issue is still the same

https://user-images.githubusercontent.com/36502659/129000807-30366662-6758-489d-8757-dbc76f5d546f.mp4

With my limited understanding, I think changing the hd.Camera.cameraFrameCount might not help in this case because the issue is with cinemachine cameras and not regular cameras. Cinemachine uses only one camera that is always active and just teleports it around when switching to other virtual cameras, so frame count is not reset?

skhiat:
@iM0ve Should be compared with playmode.
At least improve the SSR, if that do not fix the fogbugz could you re-assigned to me?

iM0ve (changes requested):
I understand this can be an improvement in some cases. If we want to merge this improvement we could, but need to remove/change the changelog entry.

As for the original fogbuz that is mentioned in the changelog - it is not fixed with this PR, I have reassigned it. Need to investigate if its possible do to something or close by design.

Testing: I launched the attached project and used the recorder to capture frames. Both in play mode. Before and after the PR the problem is identical, after cinemachine virtual camera changes, first frame has no geometry

**Before PR:**

Frame 8
image_003_0008

Frame 9
image_003_0009

Frame 10
image_003_0010

**After PR:**

Frame 8
image_002_0008

Frame 9
image_002_0009

Frame 10
image_002_0010

sebastienlagarde:
any update?

skhiat:
@iM0ve 
Could you confirm it's still the case.
I tested this PR with the project/scene attached on this fogbugz:
https://fogbugz.unity3d.com/f/cases/1349132/
Before PR:
BeforePR
After PR:
AfterPR

Tested with timeline and tested with Approximation and PRB Accumulation.

iM0ve (changes requested):
Please request review again when new changes are implemented. Currently there is still no geometry for the first frame, can be seen when captured with recorder. Marking request changes to clean up backlog.

iM0ve (approved):
This PR doesnt fix the first frame not having reflections, but it alleviates the problem. Approving. There are plans to investigate the original issue, the fogbuz is still active.

sebastienlagarde (approved):

PR #5325, opened 2021-08-10
Updated minDistance epsilon URP decal

Jonasmortensen:
### Purpose of this PR
This is a URP port of this HDRP PR: https://github.com/Unity-Technologies/Graphics/pull/5199 
It fixes an issue caused by a large epsilon value.
Images are from URP:
Before
DecalEpsilonBefore
After
DecalEpsilonAfter


lukaschod:
Could you check just in case it is not updating always now, just to be sure as it would be performance regression.

Jonasmortensen:
> Could you check just in case it is not updating always now, just to be sure as it would be performance regression.

Confirmed that the decals are only updated when moving the transform.
Tested by logging a message after the early return in `Execute()` of the `UpdateTransformsJob`

lukaschod (approved):

aleksandrasdzikia (approved):
As talked to Jonas, no tests need, approving on behalf of gfx-urp team.

phi-lira (approved):

PR #5326, opened 2021-08-10
Create VFX Editor window's UI on the correct callback

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1357007

---
### Testing status
Tested by switching branch between this branch and a temporary branch which have the same fix and a change that causes the issue.

---
### Comments to reviewers
You can use this temporary branch `vfx/tmp/change-uxml-structure` for testing there should be no issue.
If you switch from the temporary branch to master you'll have an issue (because the fix is not on master)
If you switch from master to this PR or the temporary branch there should not be any issue

### DIFF
--- a/com.unity.visualeffectgraph/CHANGELOG.md
+++ b/com.unity.visualeffectgraph/CHANGELOG.md
@@ -76,7 +76,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Inspector group headers now have a better indentation and alignment
 - Zoom and warning icons were blurry in the "Play Controls" and "Visual Effect Model" scene overlays
 - Random crash using subgraph [Case 1345426](https://issuetracker.unity3d.com/product/unity/issues/guid/1345426/)
-- Graph Editor now creates the window's UI on the correct callback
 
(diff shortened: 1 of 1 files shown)


VitaSkruibyte (approved):

VitaSkruibyte:
Ah, speaking of bug description you probably wanted to link 1357007 instead of 1347399?

com.unity.visualeffectgraph/CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Inspector group headers now have a better indentation and alignment
 - Zoom and warning icons were blurry in the "Play Controls" and "Visual Effect Model" scene overlays
 - Random crash using subgraph [Case 1345426](https://issuetracker.unity3d.com/product/unity/issues/guid/1345426/)
+- Graph Editor now creates the window's UI on the correct callback
```
julienf-unity, line 79: No need for a changelog in that case. It does not bring anything to users and there's no bug assicoated with it

julienf-unity (approved):

PR #5329, opened 2021-08-10
[Fogbugz 1352407] Fixing TAAU and DLSS resolutions on custom post process

kecho:
### Purpose of this PR
Fixing TAAU and DLSS resolution by passing a custom resolution to custom post processes.
The problem is that RTHandle system has a monolithic rtHandleProperties. Thus, we cant just give a texture its specific resoltuion. For this to work with the blitter and the ecosystem, I am introducing a small API that allows to have a custom handle properties into the desired target. 

Fogbugz: https://fogbugz.unity3d.com/f/cases/1352407/
---
### Testing status
Tested with custom post process from example dx11, also hardware DRS on dx12.

---
### Comments to reviewers
I tried many ways of fixing it, all of them incurring into heavy complexity on HDRP. The real solution should be implementing proper support of multi resolutions in RTHandle.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
@@ -1417,10 +1417,10 @@ namespace UnityEngine.Rendering.HighDefinition
                                         // In the case of DLSS and TAAU, the post process viewport and size for the color target have changed, thus we override them here.
                                         // When these upscalers arent set, behaviour is still the same (since the post process scale is the same as the global rt handle scale). So for simplicity, we always take this code path for custom post process color.
                                         var newProps = srcRt.rtHandleProperties;
-                                        newProps.rtHandleScale = passData.postProcessScales;
-                                        newProps.currentRenderTargetSize = passData.postProcessViewportSize;
-                                        newProps.previousRenderTargetSize = passData.postProcessViewportSize;
-                                        newProps.currentViewportSize = passData.postProcessViewportSize;
+                                        newProps.rtHandleScale = data.postProcessScales;
+                                        newProps.currentRenderTargetSize = data.postProcessViewportSize;
+                                        newProps.previousRenderTargetSize = data.postProcessViewportSize;
+                                        newProps.currentViewportSize = data.postProcessViewportSize;
                                         srcRt.SetCustomHandleProperties(newProps);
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -1397,19 +1399,39 @@ bool DoCustomPostProcess(RenderGraph renderGraph, HDCamera hdCamera, ref Texture
...
+                                        // which require the rtHandleProperties to set the viewport and sample scales.
+                                        // In the case of DLSS and TAAU, the post process viewport and size for the color target have changed, thus we override them here.
+                                        // When these upscalers arent set, behaviour is still the same (since the post process scale is the same as the global rt handle scale). So for simplicity, we always take this code path for custom post process color.
+                                        var newProps = srcRt.rtHandleProperties;
+                                        newProps.rtHandleScale = passData.postProcessScales;
+                                        newProps.currentRenderTargetSize = passData.postProcessViewportSize;
+                                        newProps.previousRenderTargetSize = passData.postProcessViewportSize;
+                                        newProps.currentViewportSize = passData.postProcessViewportSize;
```
FrancescoC-unity, line 1422: Uhm waiting for @alelievr as I don't know, but can custom PP have source that is not full size? If so, will override src RTHandle stuff create problems?

FrancescoC-unity: Also can user use anything but the HD Blit that creates problems? (Not that I have a solution, but worth knowing a bit more the problem space)

JulienIgnace-Unity (approved):
As discussed previously, this is a fine stop gap solution until we can do something better.
I think it's worth waiting for Antoine's answer to Francesco's comment though.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
kecho: yes. The documentation we explain that the color must use m_PostProcessSize. Apart from that, the user could just use the rtHandleProperties now blindly and anything would work.

alelievr:
@FrancescoC-unity technically yes

sebastienlagarde (approved):

FrancescoC-unity (approved):

PR #5330, opened 2021-08-10
Added synonyms to nodes for better searchability

bencloward:
This change adds synonyms to a set of nodes.  The synonyms are used for searches so that users don't have to know the exact name of the nodes to find what they are looking for.

### Purpose of this PR
Many of the nodes have slightly unexpected names or alternate names to the industry standard - so this allows users to enter search terms with names that are different and still come up with the correct results.  It makes searching more "intelligent"

---
### Testing status
For each of the synonyms, I have typed them in the Create Node search field to make sure that the proper nodes are displayed as search results.

---
### Comments to reviewers
A complete list of all of the synonyms I have added can be found in this document:
https://docs.google.com/document/d/1eYeCXM-Qre1O5tTrPglkSuuynLmLtS-t_b_9d7V3XUM
Note that the document also contains some symbols as synonyms such as +-/* etc - but symbols don't currently work as search terms, so I left those out of the code changes.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Nodes/HDSceneColorNode.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Nodes/HDSceneColorNode.cs
@@ -17,6 +17,7 @@ namespace UnityEditor.Rendering.HighDefinition
         public HDSceneColorNode()
         {
             name = "HD Scene Color";
+            synonyms = new string[] { "screen buffer" };
             UpdateNodeAfterDeserialization();
         }
 
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - Added Tessellation Option to PositionNode settings, to provide access to the pre-displaced tessellated position.
   - Added visible errors for invalid stage capability connections to shader graph.
   - Added a ShaderGraph animated preview framerate throttle.
+  - Added many node synonyms for the Create Node search so that it's easier to find nodes.
 
 ### Changed
 - Properties and Keywords are no longer separated by type on the blackboard. Categories allow for any combination of properties and keywords to be grouped together as the user defines.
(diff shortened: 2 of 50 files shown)


jessebarker (approved):
LGTM.

xiaoxicici (approved):
Awesome additions! Thanks for implementing these.

GrantLamb-Unity (changes requested):
1. Synonyms which have the string as existing node names or other synonyms will highlight/select the first option in alphabetical order.
For Example, if you search for Divide, it will highlight Blend over the Divide node. So when a user types Divide and immediately hits enter they will get Blend.

Nodes with synonym name conflicts.
Blend - Divide, Multiply, Negate, Screen, Subtract, overwrite
Negate - Invert
OneMinus - Invert
Saturate - Clamp
Lerp - Blend

2. Searcher doesn't seem to support just numbers, the following do not find the node.
Float - 1
Vector2 - 2
Vector3 - 3
Vector4 - 4

3. Are Sawtooth Wave Node and Triangle Wave Nodes interchangable? They look very similar and are synonyms of each other. Is this intentional? Would it make sense to remove one rely on synonyms in this case? Or do we need to keep them as is so we don't break existing projects?

4. Probably unrelated and probably a searcher bug. When you search for something then attempt to use the up or down arrows a "Events cannot be dispatched more than once." exception will trigger.
EventsCannotBeDispatched

These aren't all required changes, I'd just like some more information before approving.

bencloward:
Thank you for the thorough testing, Grant!  This is great feedback.
1. I don't think it's a problem to have more than one result come up with a search term as long as each if the results are relevant - so in your list, I wouldn't called those conflicts, I would call them alternate results. Each of them actually does do the thing that the search term describes. If you feel this is a significant issue, I'd be happy to remove them, but I think it's better to give more results than fewer.
2. For the unsupported numbers, our options are to either remove the numbers as synonyms or leave them in and get a task in the back log to get numbers and symbols supported.  I have a whole list of symbols I'd also like to use as search terms as well, but those don't work either.  In this case, I don't think it's hurting anything to have the numbers in there for now and I would definitely like to have them work at some point.
3. Both sawtooth wave and triangle wave are triangular shaped waves - so I think it's fine that both come up for the other's search term.
4. Could you file this as a separate issue? The change I made doesn't cause that issue as far as I know.

bencloward:
@GrantLamb-Unity I talked with Marc and Sai about item 1 this morning.  Sai just submitted a change to make all search results sort by best match instead of alphabetical.  Going forward, Sai has plans to make node name results always come before synonym results.

GrantLamb-Unity:
Thanks for all the feedback! Your points makes sense to me and I'm glad we were able to adjust it so it finds the best match instead of the first alpha option.

I went ahead and wrote up the exception bug I found here: https://fogbugz.unity3d.com/f/cases/1358016/

GrantLamb-Unity (approved):

PR #5339, opened 2021-08-12
Add API to completely disable Runtime Debug UI

arttu-peltonen:
### Purpose of this PR
This PR is intended to mitigate issues raised in this forum thread: https://forum.unity.com/threads/renderpiplines-debugupdater-creates-duplicate-eventsystem-and-more.1154456/

#### Problem: 
In order to support touch & mouse for Runtime Debug UI, we need to have `EventSystem` & corresponding `*InputModule` components present in the scene. To make this work out-of-the-box, `DebugUpdater` has been adding them automatically at startup, unless already present. This approach is problematic because some scenes (splash screen) might not have these components, but a subsequent scene would. Unity throws an error when more than one of these components are present, and this is what users are seeing.

#### Solution: 
This PR makes it possible to disable DebugUpdater altogether using `DebugManager.enableRuntimeUI` (which replaces the `enableWindowHotkey` variable added in an [earlier PR](https://github.com/Unity-Technologies/Graphics/pull/5059), meant to solve input conflicts). With this public API, users will be able to write scripts to initially disable the runtime UI, and also re-enable it later (at which point the UI will reuse existing `EventSystem` & `*InputModule` components if present).

This is not a perfect solution because some users will have to write custom scripts to make the runtime UI compatible with their game. The problem could be avoided if it were possible to have uGUI input (keyboard, touch, mouse, gamepad...) work without having to instantiate these components to the scene. AFAIK this is not currently possible (see [slack discussion](https://unity.slack.com/archives/C09Q7LYP9/p1628756059100500)), so this is the best I have come up with.

---
### Testing status
I used [this test script](https://github.com/Unity-Technologies/Graphics/files/6980797/TestDisableRuntimeUI.zip) in editor playmode to simulate users changing `enableRuntimeUI` value back and forth from script at runtime. Seems to work correctly - runtime UI disappears when value is disabled, and can be reopened when it is enabled again. Also when the value is disabled in `Awake()`, the `DebugUpdater` GameObject is never instantiated in the first place, thus avoiding any potential initialization overhead.

Additionally, tested that Debug UI & input still works as expected with these projects & platforms:
- HDRP Sample (Windows) - Editor Playmode, both with Input Manager (Old) and Input System Package (New)
- URP Sample (Windows) - both in Editor Playmode & Player build, both with Input Manager (Old) and Input System Package (New)
- URP Sample (Android) - Player build, both with Input Manager (Old) and Input System Package (new)

I would appreciate some testing with "real projects" that have multiple scenes (ideally ones using input and others not using it), to see that Runtime Debug UI behaves sensibly when scene is changed.


com.unity.render-pipelines.core/Runtime/Debugging/DebugManager.cs
```diff
@@ -86,15 +86,30 @@ public ReadOnlyCollection<DebugUI.Panel> panels
         /// </summary>
         public bool displayEditorUI => m_EditorOpen;
         /// <summary>
-        /// Controls whether the hotkey for opening and closing the debug window is enabled.
-        /// </summary>
-        public bool enableWindowHotkey = true;
```
arttu-peltonen, line 91: I assume it's OK to change public API here because this was just added a couple of weeks ago.

com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Debug-Window.md
```diff
@@ -16,10 +16,12 @@ The Render Pipeline Debugger window.
...
+To open the Render Pipeline Debugger window in the Editor, go to  **Window > Analysis > Rendering Debugger**. You can also open this window at runtime in Play Mode, or in the standalone Unity Player on any device on **Development build**. Use the keyboard shortcut Ctrl+Backspace (Ctrl+Delete on macOS) or press L3 and R3 (Left Stick and Right Stick) on a controller to open the window.
 
 You can display read-only items such as the FPS counter independently of the **Render Pipeline Debugger** window. This means that when you disable the **Render Pipeline Debugger** window, they are still visible in the top right corner of the screen. This is particularly useful if you want to track particular values without cluttering the screen.
 
+You can disable the runtime UI entirely by using the [enableRuntimeUI](https://docs.unity3d.com/Packages/com.unity.render-pipelines.core@12.0/api/UnityEngine.Rendering.DebugManager.html#UnityEngine_Rendering_DebugManager_enableRuntimeUI) property.
```
alex-vazquez, line 23:
```suggestion
You can disable the runtime UI entirely by using the [enableRuntimeUI](https://docs.unity3d.com/Packages/com.unity.render-pipelines.core@latest/api/UnityEngine.Rendering.DebugManager.html#UnityEngine_Rendering_DebugManager_enableRuntimeUI) property.
```

Or a relative path maybe? ../api/UnityEngine.Rendering.DebugManager.html#UnityEngine_Rendering_DebugManager_enableRuntimeUI

alex-vazquez (approved):
Minor comment about the docs

com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Debug-Window.md
arttu-peltonen: Good catch, thanks! I did what you suggested, relative path won't work here because it's referring to core package and this is HD docs.

JulienIgnace-Unity (changes requested):
Looks good to me.
Might be good to add a small section to the what's new page to point to the new doc for more information regarding this new option.

JulienIgnace-Unity (approved):

com.unity.render-pipelines.core/Runtime/Debugging/DebugManager.cs
sebastienlagarde: yes

sebastienlagarde (approved):

simon-engelbrecht-soerensen (approved):
The changes work fine. Tested with boat attack and changing scenes with/without eventsystem/input.

Verasl (approved):
LGTM, a much needed improvement though it definitely still needs an automated solution in the near future as users will still encounter the issue still unless they add the extra code and systems to disable it thus will still feel compelled to send in bug reports.

RemyUnity:
Looks like @Unity-Technologies/gfx-qa-hdrp review is not needed anymore, but I still tested this PR with the spaceship demo project using [this script](https://github.com/Unity-Technologies/Graphics/files/7085692/DebugUIController.zip) .
I was able to reproduce the behavior described in the description, LGTM 👍

RemyUnity (approved):

PR #5345, opened 2021-08-13
Fixed the earth curvature not being properly taken into account when evaluating the sun attenuation (case 1357927).

anisunity:
https://fogbugz.unity3d.com/f/cases/1357927/
We used to evalute the sun at the entrace of the cloud layer intersection, which is cheaper, but doesn't give that caracteristic spherical earth look at dawn and dusk. This samples twice, which does the trick (ideal would be to sample at each step, but that would be way too expensive).

Testing status
Played extensively with the sun after the change, looks fine.
SunAttenuationpng
AfterFix
BeforeFix


iM0ve (approved):
Testing looks sufficient, thank you for the showcase

JulienIgnace-Unity (approved):

pierre-unity (approved):
It's working well for the most part. Some of my issues were related to the aerosol that was too high and was badly affecting the sun light as a result. 

There is however still some issues potentially regarding the PBsky exposure itself, but this is most likely not related to the clouds.

PR #5348, opened 2021-08-16
Make some volumetric clouds properties additional to reduce the number default parameters (case 1357926).

anisunity:
https://fogbugz.unity3d.com/f/cases/1357926/
image
image
Some properties in the volumetric clouds were moved to additional, and an extra empty line was removed

Testing status
Checked the volume's appearence in normal and additional property mode, all good.


anisunity:
Trivial changes, do not require running the tests

JulienIgnace-Unity (approved):

remi-chapelain (approved):
Thanks for adding screenshots with and without additional properties. 
Change simple enough to approve without testing !

PR #5349, opened 2021-08-16
Fixed the fallback sun for volumetric clouds having a non null intensity (case 1353955).

anisunity:
https://fogbugz.unity3d.com/f/cases/1353955/
The volumetric clouds require a sun to work, previously a fallback sun was created, vertical and had an intensity of 1 (pre-exposed). Users prefer to have a black sun, let's do that.

Testing status
Checked that it works, trivial enough
sunBlack


remi-chapelain (approved):
Change makes sense and is more inline with what users expect when there's no sun. ✔️

JulienIgnace-Unity (approved):

PR #5354, opened 2021-08-16
Added a parameter to control the vertical shape offset of the volumetric clouds (case 1358528)

anisunity:
https://fogbugz.unity3d.com/f/cases/1358528/
Added a slider to control the vertical shape offset.
Changed the name of the label of cloud offset to cloud map offset without renaming the variable, so no migration issue.

Testing status.
Made sure the label is correct
Played with the new slider, made sure moving it positively, makes the clouds go up (and aswell as the opposite).
shapeOffsetY
image
image


anisunity:
Changed Cloud Tiling to Cloud Map Tiling aswell

anisunity:
Trivial enough, no need for the tests, just the docs approval

remi-chapelain (approved):
- Tested that the vertical offset looped correctly when on whole positive and negative number. (-1,0,1..etc).
- Tested that already setup project updated properly. 


For now, there's still going to be 3 distinct variable to avoid creating upgrade issue. 
But as a improvement, theses offsets could be merged into a Vector3 with 3 inline float field to save space since there's already a lot of parameters (like the tiling / offset above) 

image

JulienIgnace-Unity (approved):

PR #5361, opened 2021-08-17
Removed DLSS keyword from settings' search (case 1358409)

fredericv-unity3d:
---
### Purpose of this PR
Fixes case [1358409](https://fogbugz.unity3d.com/f/cases/1358409)

Removed DLSS keyword in settings search to prevent pinging hidden DLSS properties when the NVIDIA package is not installed.

---
### Testing status
Manual testing:
* Searched DLSS without NVIDIA package installed: observed no result where returned
* Searched DLSS with NVIDIA package installed: observed 2 results and DLSS highlighted properties

---


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/RenderPipelineSettingsUtilities.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections.Generic;
+using System.Linq;
+
+namespace UnityEditor.Rendering.HighDefinition
+{
+    static class RenderPipelineSettingsUtilities
+    {
+        public static IEnumerable<string> Case1358409Workaround(IEnumerable<string> keywords)
```
JulienIgnace-Unity, line 8: Do we really need a name like this? I guess there is no other way to do that right? So it's not really necessary to reference a bug case IMO.
So having something explicit like "RemoveDLSSKeywordIfNeeded" would be fine by me.

remi-chapelain (approved):
While this solution is not "generic", sadly, as discussed with @fredericv-unity3d , we cannot do something more elegant that would include future hidden properties too. 

This still fixes the issue though ✔️

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/RenderPipelineSettingsUtilities.cs
fredericv-unity3d: As discussed offline, I'll wait for Julien to decide the naming

JulienIgnace-Unity (approved):

PR #5363, opened 2021-08-17
Reduced the volumetric clouds pattern repetition frequency (case 1358717)

anisunity:
https://fogbugz.unity3d.com/f/cases/1358717/
image
Changed the way the shaping noise is used to break the repetition pattern.

Testing status
Tested with various shape factors, the improvement is clear.
Will break all the volumetric clouds tests.


remi-chapelain (approved):
Nice improvement. ✔️ 
Approving but clouds tests needs to be updated before merging.

JulienIgnace-Unity (approved):

PR #5364, opened 2021-08-17
Bugfix 1357311: Fixed old ray tracing material conversion

JarkkoUnity:
### Purpose of this PR
Fixed material conversion crash where material used old ray tracing rendering queue (3900)

---
### Testing status
Tested the project associated with the bug report that the conversion works properly

---
### Comments to reviewers
The original Fogbugz case: https://fogbugz.unity3d.com/f/cases/1357311/


JulienIgnace-Unity (approved):

remi-chapelain (approved):
I can't think of anything more to test than making sure the repro is not happening again. 
Trusting @JarkkoUnity on this ✔️

PR #5365, opened 2021-08-17
Bugfix 1358480: Fixed missing tooltip for "Tessellation Mode"

JarkkoUnity:
### Purpose of this PR
Fixed missing tooltip for "Tessellation Mode"

---
### Testing status
Tested the tooltip shows in editor

---
### Comments to reviewers
The original Fogbugz case: https://fogbugz.unity3d.com/f/cases/1358480/

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -367,7 +367,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed the volumetric clouds having no control over the vertical wind (case 1354920).
 - Fixed case where the SceneView don't refresh when using LightExplorer with a running and Paused game (1354129)
 - Fixed wrong ordering in FrameSettings (Normalize Reflection Probes)
-- Fixed missing tooltip for "Tessellation Mode"
 
 ### Changed
 - Changed Window/Render Pipeline/HD Render Pipeline Wizard to Window/Rendering/HDRP Wizard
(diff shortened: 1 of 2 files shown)


alex-vazquez (approved):

JulienIgnace-Unity (approved):

remi-chapelain (changes requested):
While this technically fixes the issue indeed. I feel like this does not help users a lot, especially since it's not self explanatory which option does what. 

What do you think of adding two quick lines (summed up from [documentation](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@12.0/manual/Tessellation.html)) to help users a bit more here ? 

"Specifies the method HDRP uses to tessellate the mesh. 
None only use Displacement Map to tessellate the mesh. 
Phong tessellation applies vertex interpolation to make geometry smoother."

image

JarkkoUnity:
I don't know how much of the documentation is usually pulled to tooltips. The problem I see is the maintenance, i.e. if options are removed/added, then in addition to the docs the tooltips needs to be updated as well to reflect the change, which I guess is easy to miss. How is this with other dropdown menus?

remi-chapelain:
The best answer I can give is, it depends :). Some dropdown tooltips are brief when it's self explanatory and some are way longer. 

I personally prefer having a longer tooltip as soon as I consider there can be confusion or questioning for the user as to what these select options do. And it saves lots of time searching for details in the documentation. 

As for your concerns about maintenance, I would say, @Unity-Technologies/gfx-qa-hdrp is already very sensitive about tooltips. HDRP is already not easy to learn, better take all our chances :)

Here's a few example from materials and overrides: 

image

image

image

JarkkoUnity:
Sure no problem to add it if it's already a policy to do so. Ideally the editor would pull the tooltips from the doc files instead of having them embedded in the code,

remi-chapelain (approved):

sebastienlagarde (approved):

PR #5367, opened 2021-08-17
XPipeline- Improvements of Light UI inspector code.

alex-vazquez:
### Purpose of this PR
XPIPELINE-247

- Improve the code for Light UI
- Add light temperature preset to URP (https://unity.slack.com/archives/G010YL29H8F/p1626351559009800?thread_ts=1626349381.004000&cid=G010YL29H8F)
- Refactor of the Temperature slider.

HDRP and URP look the same now:

image

---
### Testing status
- Check all the icons
- Check presets work on URP

---
### Comments to reviewers
Moving code to Core from HDRP and using it on URP.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
@@ -1136,9 +1136,9 @@ namespace UnityEditor.Rendering.HighDefinition
 
                 if (serialized.settings.isMixed)
                 {
-					bool enabled = HDRenderPipeline.currentAsset.currentPlatformRenderPipelineSettings.supportShadowMask;
-					if (Lightmapping.TryGetLightingSettings(out var settings))
-						enabled &= settings.mixedBakeMode == MixedLightingMode.Shadowmask;
+                    bool enabled = HDRenderPipeline.currentAsset.currentPlatformRenderPipelineSettings.supportShadowMask;
+                    if (Lightmapping.TryGetLightingSettings(out var settings))
+                        enabled &= settings.mixedBakeMode == MixedLightingMode.Shadowmask;
                     using (new EditorGUI.DisabledScope(!enabled))
                     {
                         Rect nonLightmappedOnlyRect = EditorGUILayout.GetControlRect();


arttu-peltonen (commented):
Found some code that could potentially get removed?

arttu-peltonen:
General comment: I haven't found a great way to get a sense of "what's changed" when reviewing refactoring PRs such as this one, so any tips with this are appreciated. Lots of things have just moved to new files, but sometimes there are subtle changes that are hard to notice. Some ideas that might help with this issue in the future:

1. List non-functional changes separately from the functional changes in the PR description.
2. Make separate commits where things are only moved around, and functional changes in other commits. This would allow me to view the functional changes in isolation.

Any thoughts?

com.unity.render-pipelines.high-definition/Editor/Lighting/LightUnit/HDLightUnitSlider.cs
```diff
@@ -0,0 +1,290 @@
...
+                UpdatePiece(ref piece, internalValue);
+                value = SliderToValue(piece, internalValue);
+            }
+        }
+    }
+
+    /// <summary>
+    /// Formats the provided descriptor into a punctual light unit slider with contextual slider markers, tooltips, and icons.
+    /// </summary>
+    class HDPunctualLightUnitSlider : PiecewiseLightUnitSlider
```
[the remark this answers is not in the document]
alex-vazquez: yeah, I forgot to delete the file, thank you :)

martint-unity:
Why do we have a drop-down for only one option?
And also the length of the drop-down is wider than that it needs to be. SO it sticks out on the right hand side. Just looks a bit off.
image

com.unity.render-pipelines.core/Editor/Lighting/LightUI.Drawers.cs
```diff
@@ -0,0 +1,69 @@
...
+    public partial class LightUI
+    {
+        /// <summary>
+        /// Draws the color UI for a given light
+        /// </summary>
+        /// <param name="serialized"><see cref="ISerializedLight"/>The light to draw the color UI</param>
+        /// <param name="owner"><see cref="Editor"/>The editor that is handling the light</param>
+        public static void DrawColor(ISerializedLight serialized, Editor owner)
+        {
+            using (var changes = new EditorGUI.ChangeCheckScope())
```
martint-unity, line 19: var changes are not used and can be removed

com.unity.render-pipelines.core/Editor/Lighting/LightUI.Skin.cs
```diff
@@ -0,0 +1,42 @@
...
+            public static readonly GUIContent generalHeader = EditorGUIUtility.TrTextContent("General");
+            public static readonly GUIContent shapeHeader = EditorGUIUtility.TrTextContent("Shape");
+            public static readonly GUIContent renderingHeader = EditorGUIUtility.TrTextContent("Rendering");
+            public static readonly GUIContent emissionHeader = EditorGUIUtility.TrTextContent("Emission");
+            public static readonly GUIContent shadowHeader = EditorGUIUtility.TrTextContent("Shadows");
+
+            public static readonly GUIContent lightLayer = EditorGUIUtility.TrTextContent("Light Layer", "Specifies the current Light Layers that the Light affects. This Light illuminates corresponding Renderers with the same Light Layer flags.");
+
+            // in casse that you want to keep the indentation but have nothing to write
+            public static readonly GUIContent empty = EditorGUIUtility.TrTextContent(" ");
```
martint-unity, line 24: Is this the place for this empty style. Isn't this a more global thing? Or maybe I am missing something.

com.unity.render-pipelines.core/Editor/Lighting/LightUnit/LightUnitSliderSettings.cs
```diff
@@ -0,0 +1,238 @@
...
+        public static class LightUnitIcon
+        {
+            // Note: We do not use the editor resource loading mechanism for light unit icons because we need to skin the icon correctly for the editor theme.
+            // Maybe the resource reloader can be improved to support icon loading (thus supporting skinning)?
+            static Texture2D GetLightUnitIcon(string name)
+            {
+                return CoreEditorUtils.LoadIcon(@"Packages/com.unity.render-pipelines.core/Editor/Lighting/Icons/LightUnitIcons", name, ".png");
+            }
+
+            // TODO: Move all light unit icons from the package into the built-in resources.
```
martint-unity, line 195: Do we have a Jira ticket for this task?

com.unity.render-pipelines.high-definition/Editor/Lighting/LightUnit/HDLightUnitSlider.cs
```diff
@@ -0,0 +1,289 @@
...
+
+            // Internally, slider is always in terms of lumens, so that the slider is uniform for all light units.
+            k_PunctualLightUnitSlider = new HDPunctualLightUnitSlider(LightUnitSliderDescriptors.LumenDescriptor);
+
+            // Exposure is in EV100, but we load a separate due to the different icon set.
+            k_ExposureSlider = new HDPiecewiseLightUnitSlider(LightUnitSliderDescriptors.ExposureDescriptor);
+        }
+
+        // Need to cache the serialized object on the slider, to add support for the preset selection context menu (need to apply changes to serialized)
+        // TODO: This slider drawer is getting kind of bloated. Break up the implementation into where it is actually used?
```
martint-unity, line 241: is this TODO tracked as a JIRA somewhere?

martint-unity (changes requested):
Have a few comments

com.unity.render-pipelines.core/Editor/Lighting/LightUI.Drawers.cs
arttu-peltonen: Nice, I didn't know you can omit the variable assignment. Apparently compiler introduces a hidden variable in this case so the using block still works as expected. Good to know!

com.unity.render-pipelines.core/Editor/Lighting/LightUnit/LightUnitSliderSettings.cs
alex-vazquez: XPIPELINE-247

com.unity.render-pipelines.high-definition/Editor/Lighting/LightUnit/HDLightUnitSlider.cs
alex-vazquez: not that I know

com.unity.render-pipelines.core/Editor/Lighting/LightUI.Drawers.cs
RSlysz: Question though: As it is not used, do we need the using line at all?

RSlysz (approved):

com.unity.render-pipelines.core/Editor/Lighting/LightUI.Drawers.cs
arttu-peltonen: The using-scope ensures that the finalizer is called at the end of the scope (when the compiler-generated hidden variable falls out of scope), so it's still needed I think.

RSlysz: AFAIK the finalizer of this scope only call EndChangeCheck() as `changes.changed` was never called. Also the balanceness of Begin-End is ensured by the scope mechanism but also checked in case of.
So here it is the same than doing:
```
BeginChangeCheck();
//all instruction in the using
if (EndChangeCheck())
    ; //nothing to do
```
And I don't get the point to do that.

arttu-peltonen: You might be right, but it's not obvious to me if it would be safe to remove the ChangeCheckScope. What if some code inside of the scope touches `GUI.changed`? If the scope is removed, then that "changed status" leaks out of this function, correct? It maybe that I don't yet fully understand how this all works so I'd trust your judgement on this though :) I guess my argument is that since Begin/EndChangeCheck touch global state, you need to be careful to not unintentionally change behavior.

RSlysz: It leaks anyway. The way it is done is a push/pop mechanisme similar to OpenGL push/pop transformation matrix.
So when you `BeginChangeCheck()`, it add a layer above existing data with `GUI.changed` at false.
And when you `EndChangeCheck()`, it `GUI.changed` from the layer you are poping but also pass the result this way:
> GUI.changed |= s_ChangedStack.Pop();

arttu-peltonen: Ah I see. In this case you are then most likely right and the change scope is not needed when the scope variable is unused.

alex-vazquez: this was tackled in another PR

com.unity.render-pipelines.core/Editor/Lighting/LightUI.Skin.cs
alex-vazquez: moved in another PR

arttu-peltonen (approved):

martint-unity (changes requested):

RSlysz (approved):

martint-unity (approved):

erikabar (approved):
Developer already tested this PR, I'm approving without additional testing

PR #5369, opened 2021-08-17
[URP] DepthCopy texture as a color

jonuuukas:
---
### Purpose of this PR
---
This PR changes the behaviour of URP's DepthCopy pass to output R32_SFloat texture instead of a native depth texture.
Why? Because Native RenderPasses support only a single depth buffer and this copy buffer usually is the main offender on why RenderPasses get split.
This should also enable some features(f.e. SSAO, Decals and etc.) to be able to utilise RenderPass API as well, as a lot of them rely on additional depth texture. This would also mean that the depth could technically be framebuffer-fetched using this newly changed texture

### Testing status
Yamato is as green as it can get (it seems, not sure), but QA pass definitely should be done, as this affects the whole URP. Manually checked all of the URP test suites and ran locally on mac editor.
---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/VFXGraph/Shaders/VFXPasses.template
+++ b/com.unity.render-pipelines.universal/Editor/VFXGraph/Shaders/VFXPasses.template
@@ -86,7 +86,7 @@ void frag(ps_input i
 #if VFX_PASSDEPTH == VFX_PASSDEPTH_ACTUAL
     #if defined(WRITE_NORMAL_BUFFER)
     , out float4 outNormalBuffer : SV_Target0
-    , out float4 outDepthColor : SV_Depth
+    , out float4 outDepthColor : SV_Target1
     #else
     , out float4 outDepthColor : SV_Target0
     #endif


com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -43,14 +43,15 @@ public void Setup(RenderTargetHandle source, RenderTargetHandle destination)
         public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderingData)
         {
             var descriptor = renderingData.cameraData.cameraTargetDescriptor;
-            descriptor.colorFormat = RenderTextureFormat.Depth;
-            descriptor.depthBufferBits = 32; //TODO: do we really need this. double check;
+            // descriptor.colorFormat = RenderTextureFormat.Depth;
+            descriptor.graphicsFormat = GraphicsFormat.R32_SFloat;
+            descriptor.depthBufferBits = 0; //TODO: do we really need this. double check;
```
manuele-bonanno, line 48: this looks weird, not sure is needed

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -43,14 +43,15 @@ public void Setup(RenderTargetHandle source, RenderTargetHandle destination)
         public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderingData)
         {
             var descriptor = renderingData.cameraData.cameraTargetDescriptor;
-            descriptor.colorFormat = RenderTextureFormat.Depth;
-            descriptor.depthBufferBits = 32; //TODO: do we really need this. double check;
+            // descriptor.colorFormat = RenderTextureFormat.Depth;
```
manuele-bonanno, line 46: the commented out line can be removed

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -31,6 +31,7 @@ public DepthOnlyPass(RenderPassEvent evt, RenderQueueRange renderQueueRange, Lay
             base.profilingSampler = new ProfilingSampler(nameof(DepthOnlyPass));
             m_FilteringSettings = new FilteringSettings(renderQueueRange, layerMask);
             renderPassEvent = evt;
+            useNativeRenderPass = false;
```
manuele-bonanno, line 34: why is this pass not RP compatible anymore?

manuele-bonanno (approved):
looks good! Some minor comments

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -108,7 +108,7 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
                         cameraTarget,
                         RenderBufferLoadAction.Load,
                         RenderBufferStoreAction.Store,
-                        ClearFlag.None,
+                        ClearFlag.Depth,
```
phi-lira, line 111: I don't follow why we are changing this. Could this maybe introduce regression for users that might read depend on depth after final blit pass?

phi-lira (approved):

phi-lira:
Adding @PaulDemeulenaere and @thomas-zeng for area review.

manuele-bonanno:
please also update the changelog

thomas-zeng (approved):
Approve for the XR related changes. Please address other reviewer's feedback :)

PaulDemeulenaere (approved):
Yamato is green on VFX: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fdepth-copy-as-color/.yamato%252Fall-vfx.yml%2523VFX_trunk/9097612/job

PR #5375, opened 2021-08-17
Fixed Bug That Prevented Sclera Iris Blend node from being added

bencloward:
The Sclera Iris Blend node had the same name as the Sclera Limbal Ring node, so when you selected the Sclera Iris Blend node from the Create Node menu, the Sclera Limbal Ring node was added instead.  This made it impossible to add a Sclera Iris Blend node to the graph.

### Purpose of this PR
Fixed the name field of the Sclera Iris Blend node so that it can be added to a graph via the Create Node menu.

---
### Testing status
After making this change, I was able to add the node correctly in my local branch.

---
### Comments to reviewers


anisunity (approved):
thanks

jessebarker (approved):
LGTM

PR #5384, opened 2021-08-18
[URP] Decal fixes

lukaschod:
---
### Purpose of this PR
Fix cases:
- Fixed decals to work with native render pass [case 1353141]
- Fixing decals rendering to be correct with render scale [1353885]

Depends on https://github.com/Unity-Technologies/Graphics/pull/5369

---
### Testing status


ellioman (approved):

erikabar (approved):
approving without additional testing as developer already verified the fixes

PR #5388, opened 2021-08-18
Misc fixes for APV

JulienIgnace-Unity:
### Purpose of this PR
Collection of small fixes for APV
- With a null ref exception when applying virtual offset on unloaded scenes.
- Fixed an issue causing failure of the bake when multiple scenes are loaded at the same time.
- Fixed debug probes after domain relaod when using L1
- Fixed per probe culling.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
@@ -620,6 +620,7 @@ namespace UnityEngine.Experimental.Rendering
             InitializeDebug(parameters.probeDebugMesh, parameters.probeDebugShader);
             InitProbeReferenceVolume(kProbeIndexPoolAllocationSize, m_MemoryBudget, m_SHBands);
             m_IsInitialized = true;
+            m_NeedsIndexRebuild = true;
             sceneBounds = parameters.sceneBounds;
 #if UNITY_EDITOR
             if (sceneBounds != null)
@@ -736,6 +737,7 @@ namespace UnityEngine.Experimental.Rendering
                 }
             }
 
+            // |= because this can be called more than once before rebuild is done.
             m_NeedsIndexRebuild |= m_Index == null || m_PendingInitInfo.pendingMinCellPosition != minCellPosition || m_PendingInitInfo.pendingMaxCellPosition != maxCellPosition;
 
             m_PendingInitInfo.pendingMinCellPosition = minCellPosition;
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.VirtualOffset.cs
```diff
@@ -17,6 +17,9 @@ private static void AddOccluders()
             for (int sceneIndex = 0; sceneIndex < UnityEngine.SceneManagement.SceneManager.sceneCount; ++sceneIndex)
             {
                 UnityEngine.SceneManagement.Scene scene = UnityEngine.SceneManagement.SceneManager.GetSceneAt(sceneIndex);
+                if (!scene.isLoaded)
```
FrancescoC-unity, line 20: Oh, weird that a scene is in the scene manager but is not loaded, did not know that was possible :D

com.unity.render-pipelines.high-definition/Runtime/Debug/ProbeVolumeDebug.shader
```diff
@@ -65,7 +65,7 @@ Shader "Hidden/HDRP/ProbeVolumeDebug"
             float4 position = float4(UNITY_MATRIX_M._m03_m13_m23, 1);
             int brickSize = UNITY_ACCESS_INSTANCED_PROP(Props, _IndexInAtlas).w;
 
-            if (distance(position.xyz, _WorldSpaceCameraPos.xyz) > _CullDistance ||
+            if (distance(position.xyz, GetCurrentViewPosition()) > _CullDistance ||
```
FrancescoC-unity, line 68: Uhm I am not sure I understand this. The position above should be absolute world position and as such we could compare with absolute position whereas this change will compare against 0 when camera relative is on.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.VirtualOffset.cs
JulienIgnace-Unity: Had that by just clicking on a scene in the hierarchy and click "Unload".

com.unity.render-pipelines.high-definition/Runtime/Debug/ProbeVolumeDebug.shader
JulienIgnace-Unity: We can talk about this one when you're back. Was actually quite easy to repro.

FrancescoC-unity (approved):

PR #5389, opened 2021-08-18
X pipeline - When editing a light preset, the light type was disabled

alex-vazquez:
### Purpose of this PR
- Fix [1358733](https://fogbugz.unity3d.com/f/cases/1358733/)

The scope was handling a property from the additional camera data

---
### Testing status
Check light presets save, apply and edit. Also check normal light inspector.

---
### Comments to reviewers
@Unity-Technologies/gfx-qa-hdrp Test the preset edition of the lights, and the light inspector

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
@@ -692,7 +692,7 @@ namespace UnityEditor.Rendering.HighDefinition
                     temperatureSliderRect.x += EditorGUIUtility.labelWidth + k_ValueUnitSeparator;
                     temperatureSliderRect.width -= EditorGUIUtility.labelWidth + k_ValueUnitSeparator;
                     TemperatureSliderUIDrawer.Draw(serialized.settings, serialized.serializedObject, serialized.settings.colorTemperature, temperatureSliderRect);
-                    
+
                     // Value and unit label
                     // Match const defined in EditorGUI.cs
                     const int k_IndentPerLevel = 15;


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -126,23 +138,52 @@ static HDLightUI()
...
+            }
+
+            void IDisposable.Dispose()
+            {
+                lightTypeScope.Dispose();
+                pointLightScope?.Dispose();
+            }
+        }
+
+        static bool s_LightTypeChanged = false;
```
arttu-peltonen, line 183: I'm not a fan of storing state in a static variable here, and using it in the next function call. I wonder if there's a better way to do this. To be fair, I don't really understand what's happening here.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -126,23 +138,52 @@ static HDLightUI()
...
-        static void DrawGeneralContent(SerializedHDLight serialized, Editor owner)
+        // This scope is here mainly to keep pointLightHDType isolated
+        public struct LightTypeEditionScope : IDisposable
+        {
+            EditorGUI.PropertyScope lightTypeScope;
+            EditorGUI.PropertyScope pointLightScope;
+
+            public LightTypeEditionScope(Rect rect, GUIContent label, SerializedHDLight serialized, bool isPreset)
+            {
+                lightTypeScope = new EditorGUI.PropertyScope(rect, label, serialized.settings.lightType);
```
arttu-peltonen, line 168: I noticed lightTypeScope and pointLightScope are in different order here vs the old code. Not sure if it matters at all.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
alex-vazquez: Changed ;)

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
alex-vazquez: I changed just in case

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
```diff
@@ -808,6 +808,27 @@ void SwitchEnabler(TData data, Editor owner)
...
+        /// <summary>
+        /// Conditioned drawer that will render the normal content if the editor is not editing a preset
+        /// </summary>
+        /// <param name="normalContent">The content of the group</param>
+        /// <param name="presetContent">The content of the group when the editor is preset</param>
+        /// <returns>A IDrawer object</returns>
+        public static IDrawer GroupWithPreset(IDrawer normalContent, IDrawer presetContent)
+        {
+            return Group(
+                    Conditional((s, e) => !IsPresetEditor(e), normalContent),
```
jenniferd-unity, line 828: what do s and e variables represent here?
also, here we are adding a new public API to have explicit Presets change. is this what we want? isnt it exposing a workaround to our users?

jenniferd-unity (changes requested):
This PR integrates the flow of Presets as something permanent. I am worried about the impact of integrating a workaround as something not temporary (even if we aim to remove it and never know when it will land).
Any fields for Preset are supposed to be supported. The previous PR aims to rectify our UI to show that some fields are not properly supported without having to check the doc or file a bug because of a preset not being fully applied/working.

com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
alex-vazquez: Ok , i see that this was a work arround and we do not want to mantain any kind of public api for this

martint-unity (approved):

remi-chapelain (approved):
Issue fixed ! All good ✔️

alex-vazquez:
image
Yamato is green despite the fact that the CI is not being updated on the PR.

PR #5391, opened 2021-08-18
Exclude templates from formatting checks

theopnv:
### Purpose of this PR

Similarly to already excluded TestProject folders and to all projects that can be run through Unity, the templates folders should be excluded from the formatting checks (done by adding a regex to the .repoconfig file).
That will avoid scanning temporary, library and log files, resulting in shorter formatting time.

---


jessebarker (approved):
LGTM

PR #4601, opened 2021-05-21
[VFX] HDRP Decal Output

ludovic-theobald:
# **[VFX] HDRP Decal Output**

# **[Design Doc](https://docs.google.com/document/d/1W519ngUEqaXSchEMFkWFFhY67tq-Cc3LsgNlHfyAR0w/edit#)**

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [x] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [x] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
We want to be able to output decals that affect not only the base color, but also the other physical properties of the affected material (Normal, Smoothness, Metallic, AO). 
The goal is to match as much as possible the features available for the HDRP decals.
This feature will be HDRP-exclusive.

---
### Testing status
Extensive testing by QA can be found here : https://docs.google.com/document/d/1lAurKH3M2Qbwtq3lb-37yQkn4DWsGsAHKkDTlHEaBZE/edit

---
### Comments to reviewers

At this stage, you need to enable Decal Layers in the HDRP asset for the feature to work. (temporary)

PR #5401, opened 2021-08-19
[VFX/URP] Fix decal with overlay & scale

PaulDemeulenaere:
---
### Purpose of this PR
- Fix wrong sampling when there is a renderscale : https://fogbugz.unity3d.com/f/cases/1343674/
- ~Fix regression (not logged) on overlay introduced by https://github.com/Unity-Technologies/Graphics/pull/4601 (see this change : https://github.com/Unity-Technologies/Graphics/pull/5401/commits/776655f803c5638b474864a590cafd3d18cc6f52)~ 
~We opened a dedicated fogbugz for this URP issue https://fogbugz.unity3d.com/f/cases/1368762/~ => Fixed 🟢 

---
### Testing status
I pushed the debug scene.
*Before*
image

*After*
image

However, I *can't* use this scene for graphicTest because the framework doesn't correctly save the result when there is an overlay :
image
⚠️ The overlay fix has been reverted and will be independently fixed. ⚠️ 

Also, testing manually with renderscale update modification:
_renderscale_update

---
### Comments to reviewers
This PR follows up with several fixes done in this change : https://github.com/Unity-Technologies/Graphics/pull/4733
@Unity-Technologies/gfx-qa-vfx This change also impacts newly introduced HDRP decals but, afaik, HDRP is never using rendering **without** offscreen texture.


com.unity.render-pipelines.high-definition/Editor/VFXGraph/Shaders/VFXDecal.template
```diff
@@ -33,8 +33,7 @@ if(_EnableDecalLayers)
 	float3 clipPos;
 	clipPos.xy = (i.pos.xy / _ScreenParams.xy) * 2.0f - 1.0f;
 	clipPos.z = VFXSampleDepth(i.pos);
-	if (_ProjectionParams.x < 0)
```
ludovic-theobald, line 36: Just had a second thought about that, and I believe this shader code is correct, and that the issue comes from URP not setting projectionParams.x properly, and/or not setting UNITY_MATRIX_I_VP properly.
I'm confused about where these are defined exactly, because they seem to be set from C# first (in ScriptableRenderer.cs l.210, and in RenderingUtils.cs, l.138), and later from C++ (GfxDevice.cpp).

PaulDemeulenaere: Mhh, you mean the _ProjectionParams.x should not have been inverted in overlay.
I made a renderdoc capture and indeed, the only difference is about `_ProjectionParams.x`
[_Base_CS.csv](https://github.com/Unity-Technologies/Graphics/files/7032430/_Base_CS.csv)
[_Overlay_CS.csv](https://github.com/Unity-Technologies/Graphics/files/7032431/_Overlay_CS.csv)
image

So, maybe we can check if commenting this `if` : https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Runtime/RenderingUtils.cs#L195 changes something ⏳

ludovic-theobald: Yes, when looking at a Renderdoc capture, both Base Camera and Overlay use the flipped texture, but Overlay has _ProjectionParams.x to +1.

When debugging at this line https://github.com/Unity-Technologies/Graphics/blob/70f78edee0ed1e28d62a6584a59bcee5822ccb6d/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L211, `projectionFlipSign` is set to +1 even for the Base Camera, but subsequently (in C++ from what I understand), it gets set to -1.

PaulDemeulenaere: The inverse matrix seems correctly computed. I'm wondering if it isn't a wrong consideration here : https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L209
**Before**
`bool isOffscreen = cameraData.targetTexture != null;`
**After**
`bool isOffscreen = cameraData.targetTexture != null || cameraData.renderType == CameraRenderType.Overlay;`

ludovic-theobald: I agree it doesn't feel right..

PaulDemeulenaere: Following up in this [conversation](https://unity.slack.com/archives/GBJ7AU4V8/p1629731647074000)

julienf-unity: Hi I don't have access to the conversation but to me projectionparams is needed. I would say it should be directly `clipPos.y *= _ProjectionParams.x`
If there's a issue with the overlay mode not setting the value correctly, this should be fixed in that code as part of this PR

PaulDemeulenaere: I opened an independant fogbugz about the overlay issue: https://fogbugz.unity3d.com/f/cases/1368762/

ludovic-theobald (approved):
Looks good to me!

aleksandrasdzikia (approved):

VitaVFX (approved):
Thank you for the fix, Paul. Couldn't find any outstanding issues related to PR. 🟢 
URP: Retested Render Scale, tweaked various camera settings, built standalone
HDRP: Checked Transparent Render Queue, tweaked Dynamic Resolution settings, built standalone (used HDRP decal output for tests)

PaulDemeulenaere:
Double check at https://github.com/Unity-Technologies/Graphics/pull/5401/commits/19b5a3670ba31b3d47e514036cc13d708c6b31b8 the behavior is still valid:
_RenderScale
And also still needed (reverting locally changes in Pass.template):
_RenderScale_fail

julienf-unity (approved):

PR #5414, opened 2021-08-19
[Yamato] Include extended reporting of UnityLog

kivistikliis:
---
### Purpose of this PR
- Adds initial patterns for instabilities and functionality to parse UnityLog.txt
- Rewrites the parsing script to be able to accommodate any other logs (aside from UTR and Unity log) into the solution with ease

---
### Testing status
Script was ran locally on several execution logs pointing to failures either in UTR or Unity log, or both (or none). 

---
### Comments to reviewers
It is difficult to test it in Yamato, as jobs rarely fail due to instabilities, and failed tests are not part of the extended reporting.

### DIFF
--- a/.yamato/script/log_scripts/utils/constants.py
+++ b/.yamato/script/log_scripts/utils/constants.py
@@ -1,0 +1,4 @@
+# Following constants correspond to which log parser class to use
+UTR_LOG = 'utr_log'
+UNITY_LOG = 'unity_log'
+EXECUTION_LOG = 'execution_log'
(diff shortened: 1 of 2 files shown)


JanBorupJakobsen (commented):
Replace strings used multpiple places like "####....####" with a single string and refer to that instead.

JanBorupJakobsen:
Maybe we should have a walkthrough of the code before submitting it?

JanBorupJakobsen (approved):

PR #5416, opened 2021-08-23
[VFX] Fix new SG integration & Sample Texture/Buffer

PaulDemeulenaere:
### Purpose of this PR
Fix case : https://fogbugz.unity3d.com/f/cases/1359391/

This kind of graph was generating a compilation error with the new shaderGraph integration :
image

---
### Testing status
- Added a Graphic Test case in `35_ShaderGraphGenerationFTP`
image
It covers the most obvious usage.

- Tested manually with skinned mesh sampling : we can't update this test right now because the data is shared with URP.
https://user-images.githubusercontent.com/22494706/130441108-557b1fc7-ae78-4874-948c-bbbe4c7f074c.mp4 (require change from 674b463b40b053bc858d48806d4dcde6e5685c35)

---
### Comments to reviewers
Alternatively, we could have remove texture from this list : https://github.com/Unity-Technologies/Graphics/blob/3ba2b545667f23607f8d098292dee254666dfe0a/com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs#L514
However, it should only be done in case of new SG integration and we *need* this default GPU expression to setup not exposed textures (if we are filtering here, we will always use the default texture fallback).

I noticed there is an issue with scene selection pass, it doesn't seems related to this PR :
_selection_outline_issue


VladNeykov (commented):
Works great with SampleTexture, SampleMesh, and SampleSkinnedMesh, tried editor and player build, using both per-particle attributes and inline values.

As discussed on slack, the SampleGraphicsBuffer issue ([1355257](https://fogbugz.unity3d.com/f/cases/1355257/)) still reproduces at the moment.

PaulDemeulenaere:
> Works great with SampleTexture, SampleMesh, and SampleSkinnedMesh, tried editor and player build, using both per-particle attributes and inline values.
> 
> As discussed on slack, the SampleGraphicsBuffer issue ([1355257](https://fogbugz.unity3d.com/f/cases/1355257/)) still reproduces at the moment.

Good catch, the fix is available at https://github.com/Unity-Technologies/Graphics/commit/963ad8a2cfcdbfaec2f2432b5994b688db864500

VladNeykov (approved):
Looking great, thanks for the fix!

Tested with Sample Texture, Mesh, Skinned Mesh, and Graphics Buffer in editor and player. Checked also the old SG integration and ran FTP, and tried both lit and unlit shaders; no issues found.

p.s. We can probably also resolve [this](https://fogbugz.unity3d.com/f/cases/1355257/) FB case as it's also fixed by this PR?

com.unity.visualeffectgraph/Editor/Compiler/VFXCodeGenerator.cs
```diff
@@ -362,10 +362,16 @@ static private void SubstituteMacros(StringBuilder builder)
             blockCallFunctionContent = blockCallFunction.builder.ToString();
         }
 
-        internal static void BuildParameterBuffer(VFXContextCompiledData contextData, out string parameterBufferContent)
+        internal static void BuildParameterBuffer(VFXContextCompiledData contextData, IEnumerable<string> filteredOutTextures, out string parameterBufferContent)
```
julienf-unity, line 365: I find that strange to have a parameter called filteredOutTextures that end up being written in the function...

PaulDemeulenaere: This helper function is really specific to the new SG integration.
I can potentially move the logic from https://github.com/Unity-Technologies/Graphics/pull/5416/files#diff-7174c990ca8df27503bf5973a949e19bc00cf239832d30b20b2a69167667272eR288 into this "BuildParameterBuffer".
But afterwards, it would be the same, the `WriteTextures` function have a filter parameter...

The problem here is to handle texture only used by the VFX, the textures used in shaderGraph will be declared by shaderGraph generation.

Alternatively, I could have alter the shaderGraph code generation behavior to remove the exposed texture properties (admitting they are declared by vfx)

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubTarget.cs
```diff
@@ -285,8 +285,16 @@ enum VFXAttributeType
             }
             defineSpaceDescriptor = new AdditionalCommandDescriptor("VFXDefineSpace", defineSpaceDescriptorContent);
 
+            //Texture used as input of the shaderGraph will be declared by the shaderGraph generation
+            //However, if we are sampling a texture (or a mesh), we have to declare them before the VFX code generation.
+            //Thus, remove texture used in SG from VFX declaration and let the remainder.
+            var filteredTextureInSG = context.inputSlots.Where(o =>
+            {
+                return VFXExpression.IsTexture(o.property.type);
+            }).Select(o => o.property.name);
```
PaulDemeulenaere, line 294: N.B.: Check double texture sampling (in SG & VFX)

PaulDemeulenaere: Checked locally, it works as expected @julienf-unity (I suppose we are duplicating the mapping in that case, the same texture is bound twice in the material) :
_read_texture_from_same_texture

julienf-unity (approved):

PaulDemeulenaere:
Unrelated issue while running test in edit mode, the `LogAssert.Expect` didn't behave as expected (see [here](https://github.com/Unity-Technologies/Graphics/blob/master/TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXComponentTest.cs#L605))
```
[07:49:07.413 Information] [486/1147] TestStart 'UnityEditor.VFX.Test.VisualEffectTest.CreateComponent_And_CheckDimension_Constraint'

[07:49:41.974 Information] LogEntry
	severity: Error
	message: Error assigning 3D texture to 2D texture property 'exposed_test_tex2D': Dimensions must match
	stacktrace: UnityEditor.EditorApplication:Internal_CallUpdateFunctions ()
```

PaulDemeulenaere:
Ready to be merged once formatting has been processed ⏳

PR #5383, opened 2021-08-18
Added an option to render screen space global illumination in half resolution to achieve real-time compatible performance in high resolutions (case 1353727).

anisunity:
https://fogbugz.unity3d.com/f/cases/1353727/
A new volume parameter "Full Resolution" was added to the ray marching based GI to be able to tune down the cost of the effect a very high rendering resolutions. The cost of the effect is approximately reduced by 4 when disabling the full resolution mode.
image

Testing status
Made sure the rendering looked correct in both full resolution and half resolution modes.

PR #5417, opened 2021-08-24
[HDRP][DXR] Fix for punctual and directional ray traced shadow broken temporal denoiser 

remi-chapelain:
### Purpose of this PR
Fixes [this issue](https://fogbugz.unity3d.com/f/cases/1360132/). 
[This recently merged PR](https://github.com/Unity-Technologies/Graphics/pull/5383) added the possibility to use half resolution denoiser for SSGI.
In HDTemporalFilter.cs, in the Denoiser function the new parameter fullResolution was set via the filterParams. 
However, the full resolution parameter was not set in the temporal denoiser function leading to rendering the temporal part in half resolution thus breaking Punctual and Directional ray traced shadows denoiser. (Area light were not impacted)

**Before fix** 
beforeFix

**After fix**
afterFix

---
### Testing status
- Tested all light type with ray traced shadows. ✔️ 
- Tested multiple RTShadows at the same time ✔️ 
- Also tested the options in RTShadows (Color shadows, Semi Transparent shadows, Distance based denoising) ✔️ 
- Tested in play mode with moving objects and lights to ensure accumulation is working as expected ✔️ 
- Tested that SSGI full and half resolution was working correctly ✔️ 
- All DXR tests are now green locally. ✔️ 

---
### Comments to reviewers
Fix seems "trivial" to me but since I'm not familiar with the codebase, I'd love to have some feedback if the fix makes sense to someone who actually knows how it has been written. Might be a more clean / elegant / robust fix.


remi-chapelain:
[Yamato's green ](https://yamato.cds.internal.unity3d.com/jobs/1485-Graphics/tree/HDRP%252Ffix-1360132/.yamato%252Fall-hdrp_dxr.yml%2523PR_HDRP_DXR_trunk/8321096/job/pipeline)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDTemporalFilter.cs
```diff
@@ -507,6 +510,7 @@ class TemporalFilterArrayPassData
                             ctx.cmd.SetComputeTextureParam(data.temporalFilterCS, data.outputHistoryKernel, HDShaderIDs._IntermediateDenoiseOutputTexture, data.intermediateSignalOutput);
                             ctx.cmd.SetComputeTextureParam(data.temporalFilterCS, data.outputHistoryKernel, HDShaderIDs._DenoiseOutputArrayTextureRW, data.intermediateSignalOutput);
                             ctx.cmd.DispatchCompute(data.temporalFilterCS, data.outputHistoryKernel, numTilesX, numTilesY, data.viewCount);
+                            CoreUtils.SetKeyword(ctx.cmd, "FULL_RESOLUTION_INPUT", true);
```
sebastienlagarde, line 513: note: this code mean FULL_RESOLUTION_INPUT will always be true, even if we take another path later, is it what we want?

sebastienlagarde:
currently test are green, does it mean that we miss test that cover this cases? should we add one?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDTemporalFilter.cs
remi-chapelain: I do not fully grasp the reason why this keyword is set to true after dispatch command. But, If I understand correctly, that's the default behavior (same is happening in the denoise() function above in the same file). Default is full res, unless we explicitly set to false before with the function parameter. The parameter (data.fullResolution) is only here for eventual future use but we always denoise ray traced shadows at full resolution anyway.

remi-chapelain:
> currently test are green, does it mean that we miss test that cover this cases? should we add one?

There are tests that cover this and what worries me is that locally all the tests using denoiser for punctual and directional fail but they seem to pass on Yamato for some reasons... :/ 

image

iM0ve (approved):
If the implementation is good code wise, no issues from QA side.

anisunity:
Incorrect, i'll change the code

anisunity:
The issue will only be triggered no RTGI, SSGI or RTAO is rendered before the the first shadow.

anisunity:
Fixed

anisunity (approved):

sebastienlagarde (approved):

JulienIgnace-Unity (approved):

PR #5421, opened 2021-08-24
[HDRP] Mitigate ghosting artifacts when TAA and physically-based DoF are enabled

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1340541/

The recent TAAU PR exposed the base blend factor parameter in the TAA settings
image
allowing to finetune the tradeoff between anti-aliasing and ghosting. As expected, this parameter helps a lot with ghosting when physically based DoF is enabled, because TAA runs before and after DoF in this case. 

This PR expands the range of blend factor values internally for the post-DoF TAA pass, allowing better mitigation of ghosting.

---
### Testing status
I have used the following test scene, where:
image
 
- there are high-contrast pixels at the edges of the letters in the in-focus areas and the out-of-focus background 
- the dark-green column in the defocused near region has a color that is inside the (very wide) color neighborhood of the background pixels.

In this case, if the camera "strafes" from left-to-right and the column passes over the letters, then the edges around the letters get a bit "smeared" with ghosting artifacts. Using a base blend factor of 0.6 or close to that helps to mitigate this issue and reduces ghosting significantly. 

[Note that speed rejection also helps with the ghosting, but it cannot completely solve the issue in the DoF case because the defocused pixels do not have 100% correct motion vectors]


FrancescoC-unity (approved):

sebastienlagarde (approved):

sebastienlagarde:
Please run Yamato

iM0ve (commented):
I have tested the PR, didnt find bugs. But also it was hard for me to see the improvement, maybe due to my setup not being perfect (I tried to mimic the gif from demo team).

We pair tested with Pavlos and considered to make some comparisons of images before and after the fix of his scene. So in theory it could be merged because its a very isolated minor fix, but if possible would be good to wait for Pavlos comparison results

sebastienlagarde:
@pmavridis  waiting for your conclusion, otherwise PR is ready to merge.

pmavridis:
It was a bit tricky to capture the improvement, the best way to see it is if you flip between the before/after images.

First, here is a GIF that compares the image using a TAA base blend factor of 0.8 (old) vs 0.6 (new) - here we see the improvement from both this PR and the TAAU PR:
Hysteresis_veryold_new

This is zoomed-in to show there is a reduction in the "motion trail" that makes the lines thicker. This is and "edge case" which is caused because the dark color in the foreground and the background is in the same color neighborhood (so TAA's neighborhood clipping will not work 100%). 

And here is a comparison if we compare 0.6 (old) vs 0.6 (new) - strictly the improvement from this PR:
Hysteresis_old_new

So overall, most of the improvement comes from the already merged TAAU PR that introduces the ability to tweak the TAA blend factor / hysteresis. This PR slightly refines the parameter to slightly reduce ghosting a bit more, as shown in the second gif. 

It's not a huge improvement (will depend on the scene), but since the change is also small, I think we should merge it.

PR #5423, opened 2021-08-24
Fix APV error on first time it is enabled and improve debug probe perf by 2x 

FrancescoC-unity:
This PR has two things in it: 

- Fix an error that was kicking when APV is enabled for the first time. The reason is that the cleanup seems to happen with new settings instead of old, so we try to cleanup when not init. Fix is simple enough to just check if enabled. 

- Significantly decimated the Sphere fbx (though still to a decent level). It was way too detailed to be a debug mesh for such usage. This improved the performance of the debug probe view more than 2x (still fairly slow if a lot of probes). A nightmare case I had went from 180ms per frame to 81ms


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -373,7 +373,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed the earth curvature not being properly taken into account when evaluating the sun attenuation (case 1357927).
 - Reduced the volumetric clouds pattern repetition frequency (case 1358717).
 - Fixed the clouds missing in the ambient probe and in the static and dynamic sky.
-- Fixed lens flare not rendering correctly with TAAU or DLSS
+- Fixed lens flare not rendering correctly with TAAU or DLSS.
+- Fixed assert failure when enabling the probe volume system for the first time.
+- Significantly improved performance of APV probe debug.
```
sebastienlagarde, line 378: note for later: APV isn't an official feature (yet) so we don't need to put changelog.

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -772,7 +772,7 @@ void DisposeProbeCameraPool()
 #endif
             }
 
-            if (IsAPVEnabled())
+            if (IsAPVEnabled() && ProbeReferenceVolume.instance.isInitialized)
```
JulienIgnace-Unity, line 775: Shouldn't we be able to test isInitialized inside the cleanup function and avoid having this public?
If we do it this way it means that any SRP needs to replicate this not too obvious pattern

FrancescoC-unity: I did think about that, but the problem is that if cleaning up happens _anywhere else_ but this specific use case, then if the system is not initialized something is probably wrong. 

Now, one could argue in that case we could just not cleanup, but I wondered if that "hides" potential issues.

com.unity.render-pipelines.high-definition/Runtime/RenderPipelineResources/Mesh/Sphere.fbx.meta
```diff
@@ -1,17 +1,11 @@
 fileFormatVersion: 2
-guid: 9e0af751bc36ea146940ba245193e28c
+guid: f7f013a81640a284ea38df61c509ff9a
```
PaulDemeulenaere, line 2: This change of `guid` will lead to a lost references: https://github.com/Unity-Technologies/Graphics/search?q=9e0af751bc36ea146940ba245193e28c 
See this [conversation](https://unity.slack.com/archives/G1BTWN88Z/p1630578902026800).

FrancescoC-unity: Yup, reverting it as we speak (cannot open the conversation tho :) )

PR #5125, opened 2021-07-13
[VFX] Additional Sorting Keys [22.1]

ludovic-theobald:
# **[VFX] Additional Sorting Keys**

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [ ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
This PR introduces new criteria to sort the particles of an output context. Prior to this PR, if sorting was enabled, particles were sorted by their distance to the camera. We now propose a total of 5 sorting criteria : 

- By Distance (previously, the only option)
- Youngest in front
- Oldest in front
- By Depth
- Custom
    - the Custom criterion creates a new slot in the Output context, where the user can plug a sorting key which will be used to sort the particles.
SortKeySlot

The sorting criterion can be chosen in the inspector if the **Sort** is set to On or Auto.
SortCriterionMenu
FilteredOut_SortCriterionMenu

Notes: 

- We cannot sort particles from different outputs, meaning that if you have two outputs to a system, one will appear completely in front or completely behind, depending on their sort priority.
- When possible (outputs sharing the same sort criterion), only one global sort is issued for all the outputs. If outputs have different sorting criteria, they will require their own sort pass, incurring an additional cost in performance.
- This [bug](https://fogbugz.unity3d.com/f/cases/1203011/) causing flickering when there are more than 4096 particles is fixed in this PR.

### [JIRA story](https://jira.unity3d.com/browse/GVFXG-128)

This PR required modification of the C++ code.
### Link to C++ PR : https://ono.unity3d.com/unity/unity/pull-request/130921/_/graphics/vfx/feature/new-sortkeys

---
### Testing status
What has been tested : 
- Sorting 1 output (simple quads) with all the sorting criterion
[...]

PR #5425, opened 2021-08-24
[VFX] Unexpected public API

PaulDemeulenaere:
---
### Purpose of this PR
Fix unexpected public API for VFXHDRPSettingsUtility introduced at https://github.com/Unity-Technologies/Graphics/pull/4601

---
### Testing status
Only tested if it compiles locally.
Yamato ⏳ 

---
### Comments to reviewers
See also this [conversation](https://unity.slack.com/archives/GHD5LADU7/p1629821667137800)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/VFXGraph/Utility/VFXHDRPSettingsUtility.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/VFXGraph/Utility/VFXHDRPSettingsUtility.cs
@@ -3,7 +3,7 @@ using UnityEditor.VFX.UI;
 
 namespace UnityEditor.VFX.HDRP
 {
-    public static class VFXHDRPSettingsUtility
+    static class VFXHDRPSettingsUtility
     {
         public static void RefreshVfxErrorsIfNeeded(ref bool needRefreshVfxErrors)
         {


com.unity.render-pipelines.high-definition/Editor/VFXGraph/Utility/VFXHDRPSettingsUtility.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace UnityEditor.VFX.HDRP
 {
-    public static class VFXHDRPSettingsUtility
+    static class VFXHDRPSettingsUtility
     {
         public static void RefreshVfxErrorsIfNeeded(ref bool needRefreshVfxErrors)
```
ludovic-theobald, line 8: Do we also need to set RefreshVfxErrorsIfNeeded() to internal or changing the parent class in enough ?

ludovic-theobald (approved):
Hi Paul, thanks for this PR, I also did these changes in [that PR](https://github.com/Unity-Technologies/Graphics/pull/5125), but it's cleaner if it is a PR on its own :)

com.unity.render-pipelines.high-definition/Editor/VFXGraph/Utility/VFXHDRPSettingsUtility.cs
PaulDemeulenaere: No, otherwise, this function won't be accessible by any user code, we have :
```internal class VFXHDRPSettingsUtility
{
 public //reachable in HDRP
}
```

ludovic-theobald (approved):

PR #5432, opened 2021-08-25
[URP] Lit Shader Detail Fixes

lukaschod:
### Purpose of this PR
- Added warning for lit shader detailed abledo, if texture is not linear. [1342011]
- Fixed lit detail correctly upgraded from standard shader. [1323725]

---
### Testing status
Locally tested cases


com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -168,6 +168,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed artifacts in Speed Tree 8 billboard LODs due to SpeedTree LOD smoothing/crossfading [case 1348407]
 - Support undo of URP Global Settings asset assignation (case 1342987).
 - Removed unsupported fields from Presets of Light and Camera [case 1335979].
+- Fixed an issue where Unlit and ParticlesUnlit shaders did not have HDR color selection for albedo [case 1283767](https://issuetracker.unity3d.com/issues/built-in-unlit-particle-shader-has-hdr-color-selection-for-albedo-urp-unlit-particles-do-not)
```
phi-lira, line 171: This seems merge issue?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
```diff
@@ -391,9 +392,22 @@ public static void UpdateStandardSpecularMaterialKeywords(Material material)
             CoreUtils.SetKeyword(material, "_METALLICSPECGLOSSMAP", material.GetTexture("_SpecGlossMap"));
             CoreUtils.SetKeyword(material, "_SPECULAR_SETUP", true);
             UpdateSurfaceTypeAndBlendMode(material);
+            UpdateDetailScaleOffset(material);
             BaseShaderGUI.SetupMaterialBlendMode(material);
         }
 
+        static void UpdateDetailScaleOffset(Material material)
```
phi-lira, line 399: minor: it would be good to explain the upgrade behavior in comment.

phi-lira (approved):

com.unity.render-pipelines.universal/Editor/ShaderGUI/ShadingModels/LitDetailGUI.cs
```diff
@@ -20,6 +21,7 @@ public static class Styles
                 "Designates a Normal Map to create the illusion of bumps and dents in the details of this Material's surface.");
 
             public static readonly GUIContent detailAlbedoMapScaleInfo = EditorGUIUtility.TrTextContent("Setting the scaling factor to a value other than 1 results in a less performant shader variant.");
+            public static readonly GUIContent detailAlbedoMapFormatError = EditorGUIUtility.TrTextContent("This texture is not in linear space.");
```
eh-unity, line 24: I wonder if this is clear enough for our users. I think this might be confusing for some users, perhaps expand it a bit.

I think a good error message (is hard :) ) gives you the error and context (runtime parameters) and then points/guides you towards a likely fix.

1. What is wrong
2. What is the context (how we got there)
3. What is the effect/end result (for example, detail texturing is skipped until fixed).
4. What is the correct/expected state
5. Suggestion how to fix it.

Can we give the current format of the texture in the message? so that you know what the format is when it's wrong and what it should be.

These were just my thoughts on this. No specific implementation requirements.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
```diff
@@ -391,9 +392,22 @@ public static void UpdateStandardSpecularMaterialKeywords(Material material)
...
+        {
+            // In URP details tile/offset is tied to base map tile offset
+            var baseScale = material.GetTextureScale("_BaseMap");
```
eh-unity, line 402: Is this specific to our shaders? _BaseMap is attributed as [MainTexture] for us and default is of course _MainTexture.

eh-unity (approved):
Some review comments to think about. In general looks pretty good to me.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
lukaschod: Added more info

com.unity.render-pipelines.universal/CHANGELOG.md
lukaschod: fixed

com.unity.render-pipelines.universal/Editor/ShaderGUI/ShadingModels/LitDetailGUI.cs
lukaschod: This is infobox, so usually we do not go so deep in details for the sake of UX. I kinda copied this from normal texture info box. Do you have particular suggestion into what it should change for better readability?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
lukaschod: It is specific from builtin->urp shader where tile/offset needs to be recalculated to retain same visual

com.unity.render-pipelines.universal/Editor/ShaderGUI/ShadingModels/LitDetailGUI.cs
Jonasmortensen: I also think some users will struggle with connecting linear space to the sRGB tick box. Maybe "This texture is marked as sRGB (Gamma space)"?
I think its ok to be a bit more clear since the "fix now" button is not available.

Jonasmortensen (approved):
Looks good 👍 
Will need to be updated when a new workflow for property highlighting lands.

lukaschod:
All failing issues happens on master too, merging.

PR #2146, opened 2020-10-09
Adding missing HDR tags in Unlit and ParticlesUnlit shaders

ellioman:
# Purpose of this PR
Fixes [case 1283767](https://issuetracker.unity3d.com/issues/built-in-unlit-particle-shader-has-hdr-color-selection-for-albedo-urp-unlit-particles-do-not).

# Solution
Add [HDR] to the color parameters in Unlit.shader and ParticlesUnlit.shader.

# What has been tested
Confirmed that HDR color picker is now available in the inspector

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fadding-hdr-to-unlit-colors

PR #5434, opened 2021-08-25
[Master] Adding [Gamma] tag to the MainColor property inside ParticlesUnlit and Unlit shaders

ellioman:
# Purpose of this PR
In #2146 I introduced a regression, causing the main color to change in materials using Unlit and Particles Unlit shaders.
This affected several of our CI tests.

The fix is to add the [Gamma] tag to those properties.

# Yamato
https://unity-ci.cds.internal.unity3d.com/job/8350051


Verasl (approved):
LGTM

PR #5439, opened 2021-08-25
[HDRP] Fix spot shadow sampling when using custom spot angle

alelievr:
### Purpose of this PR
Fix an issue regarding the shadow map sampling of spotlights with the custom spot angle option enabled

https://forum.unity.com/threads/custom-shadow-angle-for-spot-light-having-wrong-wrap-mode.1139824/

---
### Testing status
Checked that there is no more shadow leaking from the side of the shadow map:

Before:
image

After:
image

I also tested the pyramid spot light shape

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Ffix%252Fspot-shadow-sampling/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/8370632/job


RemyUnity (approved):
LGTM, thanks for the screenshots 🟢

PR #5441, opened 2021-08-25
[ShaderGraph] [2022.1] Fix for exception thrown when changing packages with embedded subgraphs

cdxntchou:
### Purpose of this PR
Fixing an issue I ran into, where sometimes the subgraph asset is destroyed, and we don't remove it from the cached list before attempting to access it.  This PR just adds a check to ignore the cache entry if it has been destroyed.

As far as I can tell, this is triggered when I have a package mapped to a repo, that contains a subgraph, and I sync the repo (with a subgraph asset deletion) while Unity is running.  ShaderGraph seems to run this code before the removing the deleted asset from the list, resulting in an exception being thrown when the asset is accessed -- the exception borks ShaderGraph, making it unoperational.

---
### Testing status
Tested against the repro steps above (changing package version).  The PR fixes the issue.

Yamato:
ShaderGraph PR Job:  https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/sg%252Ffix%252Fsearch-window-exception/.yamato%252Fall-shadergraph.yml%2523PR_ShaderGraph_2021.2/8407608/job

---
### Comments to reviewers


jessebarker (approved):
LGTM

PR #5444, opened 2021-08-26
Adding a null check as safeguard

martint-unity:
---
### Purpose of this PR
There could be cases in where material was null and the converter would stop and throw errors. So this will let it continue.


ellioman (approved):

ellioman:
This PR does not need run run through Yamato so I'm merging it.

PR #5450, opened 2021-08-26
[WIP] Instance indexing in shader

ludovic-theobald:
# **[WIP] Instance indexing in shader.**

---
### Overview of the changes 

- Added helper functions to retrieve the particleIndex/instanceIndex/indexInInstance etc.
- Binary search and indirection to find the instance in a vertex shader (requires some additionnal input from the runtime)
- The aligned capacity is set as a compile time constant

---
### Testing status
Tested on particle quads, particle meshes, particle strip. But limited because I didn't emulate multiple instances in a batch.

---
### Comments to reviewer
UpdateStrips.compute (!= VFXUpdate.template with HAS_STRIPS) hasn't been taken into account yet.

### DIFF
--- a/com.unity.visualeffectgraph/Shaders/VFXInstancing.hlsl
+++ b/com.unity.visualeffectgraph/Shaders/VFXInstancing.hlsl
@@ -28,7 +28,7 @@ uint GetIndexInAttributeBuffer(uint instanceIndex, uint indexInInstance)
 
 uint GetInstanceIndexFromGroupID(uint3 groupId,uint nbThreadPerGroup, uint dispatchWidth)
 {
-    return (groupId.x + dispatchWidth * groupId.y) * nbThreadPerGroup / ALIGNED_SYSTEM_CAPACITY;
+    return (1 + (groupId.x + dispatchWidth * groupId.y) * nbThreadPerGroup) / max(ALIGNED_SYSTEM_CAPACITY, nbThreadPerGroup);
 }
 
 void SetIndexAndInstanceIndex(inout VFXIndices vfxIndices)
@@ -39,32 +39,20 @@ void SetIndexAndInstanceIndex(inout VFXIndices vfxIndices)
     vfxIndices.index = GetIndexInAttributeBuffer(vfxIndices.instanceIndex, vfxIndices.particleIndex);
 }
 
-// The methods for extracting the indices are different in Compute and Output.
-// In Compute shader, it relies on the SV_GroupId and DispatchWidth which are not relevant in Output
-void VFXSetComputeInstancingIndices(
-                            uint nbParticlesPerInstance,
-                            uint3 groupId,
-                            uint nbThreadPerGroup,
-                            uint dispatchWidth,
-                            inout VFXIndices vfxIndices)
-{
-    vfxIndices.instanceIndex = GetInstanceIndexFromGroupID(groupId, nbThreadPerGroup, dispatchWidth);
-    vfxIndices.particleIndex = GetIndexInInstance(vfxIndices.index, vfxIndices.instanceIndex, nbParticlesPerInstance);
-    SetIndexAndInstanceIndex(vfxIndices);
-}
-void VFXSetOutputInstancingIndices(
-                            #if VFX_INSTANCING_VARIABLE_SIZE
-                            uint nbInstancesInDispatch,
-                            #else
-                            uint nbParticlesPerInstance,
-                            #endif
-                            inout VFXIndices vfxIndices)
+void VFXSetInstancingIndices(
+                        #if !VFX_INSTANCING_VARIABLE_SIZE
+                        uint nbParticlesPerInstance,
+                        #endif
+                        inout VFXIndices vfxIndices
+    )
 {
     #if VFX_INSTANCING_VARIABLE_SIZE
+        uint nbInstancesInDispatch, stride;
+        prefixSumInstances.GetDimensions(nbInstancesInDispatch, stride);
         vfxIndices.particleIndex = BinarySearchPrefixSum(vfxIndices.index, prefixSumInstances, nbInstancesInDispatch,vfxIndices.instanceIndex);
     #else
-        vfxIndices.instanceIndex = vfxIndices.index / nbParticlesPerInstance;
         vfxIndices.particleIndex = GetIndexInInstance(vfxIndices.index, vfxIndices.instanceIndex, nbParticlesPerInstance);
     #endif
     SetIndexAndInstanceIndex(vfxIndices);
 }
+
--- a/com.unity.visualeffectgraph/Shaders/VFXParticleCommon.template
+++ b/com.unity.visualeffectgraph/Shaders/VFXParticleCommon.template
@@ -324,25 +324,26 @@ float3 GetStripTangent(float3 currentPos, uint relativeIndex, const StripData st
 ${VFXEnd}
 
 ${VFXBegin:VFXSetOutputInstancingIndices}
-    uint nbInstancesInDispatch = 1; //TODO :nbInstancesInDispatch is supposed to be a uniform in the future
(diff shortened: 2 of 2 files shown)


com.unity.visualeffectgraph/Editor/Compiler/VFXCodeGenerator.cs
```diff
@@ -604,6 +604,7 @@ static private StringBuilder Build(VFXContext context, string templatePath, VFXC
             string renderPipeCommon = context.doesIncludeCommonCompute ? "Packages/com.unity.visualeffectgraph/Shaders/Common/VFXCommonCompute.hlsl" : renderRuntimePipePath + "/VFXCommon.hlsl";
             perPassIncludeContent.WriteLine("#include \"" + renderPipeCommon + "\"");
             perPassIncludeContent.WriteLine("#include \"Packages/com.unity.visualeffectgraph/Shaders/VFXCommon.hlsl\"");
+            perPassIncludeContent.WriteLine("#include \"Packages/com.unity.visualeffectgraph/Shaders/Common/VFXInstancing.hlsl\""); //Should it go in Global Includes ?
```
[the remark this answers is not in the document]
julienf-unity: if it does not rely on per pass / SRP defines then yes probably

com.unity.visualeffectgraph/Shaders/Common/VFXInstancing.hlsl
```diff
@@ -0,0 +1,56 @@
...
+{
+    return 0; //Not implemented
+}
+
+uint GetInstanceIndexFromThreadId(uint threadId, uint nbMax)
+{
+    //nbMax != ALIGNED_SYSTEM_CAPACITY for two reasons :
+    // * nbMax can be equal to capacity, not aligned_capacity
+    // * nbMax can be m_TotalSpawnCount in the case of immortal particles (without GPU events)
+    return threadId / nbMax;
```
julienf-unity, line 45: That does not sound correct, as as you said nbMax is potentially not fixed per instance

julienf-unity: But my other concern is that instanceIndex must be scalar whenever possible, so basically depends only on scalar inputs (like SV_GroupId), here it depends on per thread input, so will use VGPR.

com.unity.visualeffectgraph/Shaders/VFXOutputUpdate.template
```diff
@@ -72,13 +72,16 @@ ${VFXDeclareGetStripTangent}
 void CSMain(uint3 groupId		: SV_GroupID,
 			uint3 groupThreadId	: SV_GroupThreadID)
 {
-	uint id = groupThreadId.x + groupId.x * NB_THREADS_PER_GROUP + groupId.y * dispatchWidth * NB_THREADS_PER_GROUP;
+	uint id = GetThreadId(groupId, groupThreadId, NB_THREADS_PER_GROUP, dispatchWidth);
+	uint instanceIndex = GetInstanceIndexFromThreadId(id, nbMax);
+    uint indexInInstance = GetIndexInInstance(id, instanceIndex, nbMax);
+	uint index = GetIndexInAttributeBuffer(instanceIndex, indexInInstance);
```
julienf-unity, line 78: Maybe having a struc to hold those variable will be clearer?

com.unity.visualeffectgraph/Shaders/VFXParticleHeader.template
```diff
@@ -40,6 +40,27 @@ ByteAddressBuffer elementToVFXBufferPrevious;
 
 #endif
 
+#define VFX_USE_INSTANCING_INDIRECTION (!VFX_HAS_INDIRECT_DRAW)
+#if VFX_USE_INSTANCING_INDIRECTION
+StructuredBuffer<uint> instancePrefixSum;
+StructuredBuffer<uint> instanceIndirectionBuffer;
```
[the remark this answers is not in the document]
julienf-unity: Beware because on other gfx devices it will be undefined behaviour

julienf-unity (commented):
All the reformatting makes things a bit more difficult to review. Is it legit btw?

ludovic-theobald:
> All the reformatting makes things a bit more difficult to review. Is it legit btw?
Not sure what's the deal with the reformatting, AFAIK they shouldn't be concerned with the formatting hooks.

Following this morning's conversation, I separated the computation of the indices in Compute and in Output, since they rely on different assumptions. 
In the compute shaders, instanceIndex is extracted using SV_GroupID, assuming that the dispatches will be of size multiple of NB_THREADS_PER_GROUP. In vertex shader, instanceIndex will still use a VGPR.

The defines (VFX_USE_INSTANCING and others) are set globally from VFXCodeGenerator.cs, but this is only for testing. The logic of how these defines are set and where is not tackled in this PR.

com.unity.visualeffectgraph/Shaders/VFXInstancing.hlsl
```diff
@@ -0,0 +1,79 @@
...
+
+    #if VFX_INSTANCING_INDIRECTION
+        vfxIndices.instanceIndex = indirectionBufferInstances[vfxIndices.instanceIndex];
+    #endif
+    vfxIndices.index = GetIndexInAttributeBuffer(vfxIndices.instanceIndex, vfxIndices.particleIndex,alignedSystemCapacity);
+}
+
+
+void VFXSetOutputInstancingIndices(
+                            #if VFX_INSTANCING_VARIABLE_SIZE
```
gabrieldelacruz, line 53: I think you should try to unify this 2 functions, it will be easier to maintain

gabrieldelacruz (changes requested):
As discussed over the call, still some adjustments are needed to remove redundancy.
It looks good, everything seems to be already there.
Thanks!

com.unity.visualeffectgraph/Editor/Compiler/VFXCodeGenerator.cs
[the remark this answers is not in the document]
ludovic-theobald: It has been moved inside of VFXCommon.hlsl.

com.unity.visualeffectgraph/Shaders/VFXInstancing.hlsl
```diff
@@ -0,0 +1,70 @@
...
+                            uint dispatchWidth,
+                            inout VFXIndices vfxIndices)
+{
+    vfxIndices.instanceIndex = GetInstanceIndexFromGroupID(groupId, nbThreadPerGroup, dispatchWidth);
+    vfxIndices.particleIndex = GetIndexInInstance(vfxIndices.index, vfxIndices.instanceIndex, nbParticlesPerInstance);
+    SetIndexAndInstanceIndex(vfxIndices);
+}
+void VFXSetOutputInstancingIndices(
+                            #if VFX_INSTANCING_VARIABLE_SIZE
+                            uint nbInstancesInDispatch,
```
gabrieldelacruz, line 57: nbInstancesInDispatch should be a uniform as well or, even better, extracted from the buffer with StructuredBuffer::GetDimensions. No need to pass it as an argument

com.unity.visualeffectgraph/Shaders/VFXInstancing.hlsl
gabrieldelacruz: This is still not a single function. I understand the difficulty of unifying the instanceIndex obtained from the index and the optimized form that uses the group id, but it is just a matter of taking the specific code to the definitions in VFXParticleCommon.template. In the end, it is still a division, it just takes different values.

gabrieldelacruz (approved):
Looks good!
There will be more changes to this code, specially regarding nbMax that will depend on the instanceIndex, but it is pointless to keep iterating now until we have a better picture.
Good job, thanks!

PR #5455, opened 2021-08-27
Universal/package samples 2

Jonasmortensen:
### Purpose of this PR
With 21.2 comes a lot of new features to URP. These new features should be represented in the package samples so users can quickly interact and play with them. 

All samples now also have some kind of movement to make them interesting in play mode.

The existing samples also had some issues that are addressed in this PR (case 1360832).

In the PR the sample has also been renamed from "Teaching Samples" to " URP Package Samples"

So far the samples have supported building them to mobile. This is not longer actively supported. It is still doable but because of the limits of the package sample system its up to the user to setup project settings correctly (this was previously handled badly by sample scripts).

Samples that are added in this iteration:
**3D Skybox**
image

**Blob Shadow**
image

**Proxy Lighting**
image

**Paint Splat**
image

**Sun Flare**
image

**Lens Flare Showroom**
image

**Keep Frame**
image

**Trail Effect**
image

---
### Testing status

---
### Comments to reviewers
Should not be tested on mobile as support is deprecated.

To QA: Would be good to get some feedback on the descriptions (both typos and clarity) and of course whether the scenes load and behave as they should. 

To Dev: As QA will try the scenes in Editor, I only ask for a review of the code in the `.cs` files in the PR. Specifically the `KeepFramesFeature.cs` renderer feature as that code is expected to be looked at by users.

These errors might print on some samples :
image
This is a known issue being handled in another branch (case 1359269)

Some samples might also throw a "is missing renderer features" error. This is caused by the package manager not compiling all scripts before running the check and needs to be fixed by package manager team (they are aware of it)
Error2


hdb-unity (approved):
Manual QA testing complete.

Checked the project imported to a built-in project, and a URP project.

Some actions remaining, these are tracked in the test report and we'll keep discussing those.

No shipstoppers found in testing.

Thanks for the clear and detailed PR description, and images of each new test scene!

PR #5457, opened 2021-08-27
[HDRP] Adding a new button in HDRP+DXR Wizard to check scene content for Ray Tracing issues

remi-chapelain:
### Purpose of this PR

image

This PR adds a "Current Scene" section (**only in the HDRP+DXR Wizard**). 
This section contains a "check" button that triggers the "Check Scene Content For HDRP Ray Tracing" Edit > Rendering menu. 
The goal of this is to inform users that this check exists and could help fix a bad setup. 
It also helps to understand what this menu really checks instead of searching for clues in the doc. 

https://user-images.githubusercontent.com/57442369/131122611-79a7b1f5-9a29-4a8f-8e8e-3bba52b85375.mp4

I believe we could create other functions like to benefit HDRP only setup as well because some things in the checks (like verifying if meshes has null materials, renderer in LODs... etc could also benefit **Non DXR** HDRP setup.

---
### Testing status
Tested opening the wizard with any of the tabs selected by default ✔️ 
Tested clicking on "Check" with and without issue to see if console was giving information properly ✔️ 
Tested that the menu was still working properly ✔️ 

---
### Comments to reviewers
I'm open for suggestions about: 
- Better place to put this check instead of creating another section. 
- Better wording of what the check is doing. 
- Better way to ensure that the button will always stays rather than creating a function (_CheckForRTX()_) that always returns false.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Wizard/HDWizard.Window.cs
@@ -418,7 +418,8 @@ namespace UnityEditor.Rendering.HighDefinition
             m_BaseUpdatable.Add(globalScope);
             m_BaseUpdatable.Add(currentQualityScope);
 
-            var addCurrentScene = new HiddableUpdatableContainer(() => m_Configuration == Configuration.HDRP_DXR);
+            var addCurrentScene = new HiddableUpdatableContainer(
+                () => m_Configuration == Configuration.HDRP_DXR);
             addCurrentScene.Add(currentSceneScope);
             addCurrentScene.Init();
             m_BaseUpdatable.Add(addCurrentScene);


RSlysz:
I dislike this idea A LOT.

Wizard was always planed and designed only for PROJECT configuration. Now we open the door to scene by scene change check.
Why this is wrong:
- Wizard is not meant to be open each time we create a new scene so there is no easy way for user to check the new scene he create comply some constraint. Each time he should open back the wizard or have it always opened somewhere taking unneeded space.
- There is many edge case you need to check now: What hapens if we are currently viewing a prefab stage instead of a scene? What about multiple opened scene?
- In the end, what is added is just a checker not a fixer. We introduced checker to some corner case that we cannot easily change because no scripting API was available. Or because user can want to do it in some context. It is not meant to be generalized more than this : the more we add check in Wizard, the more we dilute information to solve project issue.

Please discuss this change with UX.

RSlysz (changes requested):
Please check with UX (see above)

RSlysz:
So currently it is already a menu item.
Do we have other similar action we want to check onto the scene? I would prefer having a new dedicated small Scene Configuration Checker window with really few line that can be docked somewhere as it take almost no space in this case. Still be sure to add checks on the edges cases listed above.

remi-chapelain:
Thanks for the feedbacks ! 

> Wizard was always planed and designed only for PROJECT configuration. Now we open the door to scene by scene change check.

I agree this is debatable indeed. The initial goal was to use the "notoriety" of the wizard to bring users attention to this unknown/unused feature for raytracing setup. But maybe that's not the proper location to put it. 

> Checker window with really few line that can be docked somewhere as it take almost no space in this case.

I believe having a new "checker window" with this check in it, will suffer from the same problem the command menu is already suffering from: discoverability. So that's not really a solution either.

> What happens if we are currently viewing a prefab stage instead of a scene? 

The check (as was the command menu already doing) is still going to check the current open scene. 

> What about multiple opened scene?

If you have multiple scene opened, the check will iterate through all the items in the current opened scenes and treat the multiple scene as one big opened scene. 

> Please check with UX (see above)

Only comment from @TheoWong-pixel from now is to make sure that the console will open if it's closed to be sure users won't miss messages which I believe also makes sense. 

---
**However**, with all your concerns, don't really know if that makes sense to still move forward with this :). 
What does the rest of the QA thinks about it?

sebastienlagarde:
I would advice to not go further with more windows that do check, and I am agree with RemiS.

The Wizard is already a huge investment on our side and something very specific to HDRP. I don't want to make it more complex or have more functionnality, it should stay really simple. We have already enough issue reported with it, don't want to have more :D

remi-chapelain:
Let's close it then :)

PR #5461, opened 2021-08-30
Fix an issues of flickering in URP's deferred rendering mode

Seongdae:
### Checklist for PR maker

---
### Purpose of this PR
To fix flickering issues in screen space shadows with deferred rendering mode

---
### Testing status

---


com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceShadows.cs
```diff
@@ -34,7 +34,7 @@ public override void Create()
 
             LoadMaterial();
 
-            m_SSShadowsPass.renderPassEvent = RenderPassEvent.AfterRenderingPrePasses;
+            m_SSShadowsPass.renderPassEvent = RenderPassEvent.AfterRenderingGbuffer;
```
Seongdae, line 37: @phi-lira @ellioman not be able to get access to rendering mode. In forward rendering mode, this pass's event is 'AfterRenderingPrePasses', but to fix flickering in deferred mode, I have to change the event to 'AfterRenderingGbuffer'. this fix will work on both of rendering mode. I wonder if URP provides the rendering mode, RendererFeature will be effective on any rendering mode.

ellioman: Take a look at `ScreenSpaceAmbientOcclusion.cs` 
There we pass the renderer to the `Setup() `function in the pass, and have `isRendererDeferred` check.

Seongdae: @ellioman thanks. It will help me :). What if anyone can get the rendering mode outside URP package? for example users are  working on their own renderer feature, they are not be able to get it? I'm just asking you.

ellioman: @Seongdae Good question. I'm not sure why it was decided to put the the rendering mode as Internal, maybe @kaychang-unity can help us there. My feeling is that we did not want to expose it too soon in case of any issues popping up. 

But you're absolutely right, that this might be a pain for users creating their own Rendering Features.

Seongdae: @ellioman Thank you :)

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceShadows.cs
```diff
@@ -49,11 +49,17 @@ public override void AddRenderPasses(ScriptableRenderer renderer, ref RenderingD
                 return;
             }
 
+            bool isDeferredRenderingMode = renderer is UniversalRenderer && ((UniversalRenderer)renderer).renderingMode == RenderingMode.Deferred;
+
```
ellioman, line 53: Shouldn't we move this check so it's inside the` if (shouldEnqueue)` as that's the only place it's used?

Seongdae: Ok, let me do it

ellioman (approved):

eh-unity (approved):
LGTM.

com.unity.render-pipelines.universal/Runtime/RendererFeatures/ScreenSpaceShadows.cs
kaychang-unity: I made the rendering mode internal as safety measure, initially. It could be made public.

PR #5462, opened 2021-08-30
Bugfix 1357740: Black pixels in Fabric material

JarkkoUnity:
### Purpose of this PR
To prevent users from feeding out-of-bounds values to CottonWool AO & smoothness strength material inputs.

---
### Testing status
Tested the attached Kiki sample

---
### Comments to reviewers
The real issue is in data that is out-of-bounds, but I added couple of saturate nodes (for AO & smoothness strength inputs) to protect against such cases in the future. AO strength was set to 1.51 for the detail material (even though 0-1 is the valid range) that resulted in negative AO values and cascading issues.


sebastienlagarde (approved):

remi-chapelain (approved):
Aside from the UX issue from shader graph forcing you to clamp twice (once for inspector and once again in the graph), didn't manage to repro issue while fiddling with inputs. 

Fabric tests are still green in HDRP and HDRP_DXR test projects ✔️

PR #5464, opened 2021-08-30
Adding Depth Texture option for Overlay Cameras

jonuuukas:
---
### Purpose of this PR
---

Fix for https://fogbugz.unity3d.com/f/cases/1296770/ - basically only way to have depth in Overlay camera is to expose this setting. This was discussed in triage meeting/slack and etc. and i was asked to implement this. Only problem i found was the y-flip which i fixed already it seems ( though it could maybe still happen in some scenarios? hard to say)
### Testing status
tested manually - according to the repro project in the fogbugz case
---
### Comments to reviewers

Previous behaviour:
Screenshot 2021-08-30 at 17 32 39

New behaviour (CopyDepth pass now appears on Overlay camera):
Screenshot 2021-08-30 at 17 32 21

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
@@ -136,7 +136,6 @@ namespace UnityEngine.Rendering.Universal.Internal
                 {
                     // Blit has logic to flip projection matrix when rendering to render texture.
                     // Currently the y-flip is handled in CopyDepthPass.hlsl by checking _ProjectionParams.x
-                    // However special exception is done by Overlay camera which doesn't have _ProjectionParams.x set up
                     // If you replace this Blit with a Draw* that sets projection matrix double check
                     // to also update shader.
                     // scaleBias.x = flipSign


ellioman (approved):

ernestasKupciunas (approved):
Approving without testing. The developer has tested the basic functionality and FB case.
Thanks for adding the screenshots "before" and "after"!

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -136,6 +136,7 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
                 {
                     // Blit has logic to flip projection matrix when rendering to render texture.
                     // Currently the y-flip is handled in CopyDepthPass.hlsl by checking _ProjectionParams.x
+                    // However special exception is done by Overlay camera which doesn't have _ProjectionParams.x set up
```
phi-lira, line 139: The overlay camera should setup the correct params though. I think the reason this might not be set is because this is set on SetupCameraProperties which today is only called for base camera. I'd argument that at some point we remove the dependency for SetupCameraProperties and setup all camera properties consistently for both base and overlay cameras. cc @lukaschod for some thoughts as I think you last touched the SetupCameraProperties.
 https://sourcegraph.com/github.com/Unity-Technologies/Graphics/-/blob/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs?L682-693

phi-lira (approved):
Approving but it misses changelog.

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
lukaschod: _ProjectionParams should be setup on overlay, @jonuuukas could u check maybe it is wrong by chance. As I had to recreate logic in C# which is not 100% accurate to c++ side

jonuuukas: Yeah, i was wondering whether we should call SetupCameraProperties for Overlay camera, but took the easy way for the time being. I could change this if you'd like

jonuuukas: so i fixed this in 8e9f70829cbcb3cfa65eb50962240f79b1b0d115 - previously the check in SetPerCameraShaderVariables wasn't 100% match with what we do in c++ (IsCameraProjectionMatrixFlipped matches though). Nothing much changes though - instead of only checking whether targetTexture != null, we'll also check if it's rendering to not-backbuffer (which requires flip)

lukaschod (approved):

Verasl (approved):

PR #5465, opened 2021-08-30
Changed Ambient Mode to Dynamic by default

JulienIgnace-Unity:
### Purpose of this PR
Redoing PR https://github.com/Unity-Technologies/Graphics/pull/5350 on master

Change the defaults of the Visual Environment component's Ambient mode to "Dynamic".
The idea is that by default it will provide users a more "out of the box" experience where ambient lighting works without any manual action from them.

This should not break any existing project apart from one very specific case: If the user's project does not have a default volume profile already containg a visual environment component, a new one will be created by HDRP automatically in memory and will then use Dynamic Ambient contrary to default static.
I feel that this is not too much of a problem because very few projects will have that issue since we always create one by default.

Issue raised in this case https://fogbugz.unity3d.com/f/cases/1357625/

---
### Testing status
- Created a new Visual Env component and checked that Ambient was indeed Dynamic
- Tested that existing components were not changed.


sebastienlagarde (approved):

sebastienlagarde:
@JulienIgnace-Unity PR have been reverted

PR #5466, opened 2021-08-30
Fix slope scale depth bias when depth offset is ON

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1360709/ (though the bug didn't have a repro, so I must assume this was the issue). 

The problem is that when explicitly outputting the depth, the slope scale depth bias is not properly added via HW, so we now add manually. 

**What did I test**: can be seen here
Before

Unity_xmAVNQZd2W

After 

Unity_WlKhBbjOEA

Without Depth Offset

Unity_5WnjvvSqD5


sebastienlagarde (approved):

sebastienlagarde:
FYI @francesco: the discussion on slack about this issue was here: https://unity.slack.com/archives/GHD5LADU7/p1629901034150000

sebastienlagarde:
@FrancescoC-unity  / @Unity-Technologies/gfx-qa-hdrp : I poked Alix who reported the issue and this fix correctly works with her use case: https://unity.slack.com/archives/GHD5LADU7/p1629901034150000

iM0ve (approved):
Looks good, simple fix while tested both by the developer and the artist

PR #5469, opened 2021-08-30
Merge shape offset parameters

adrien-de-tocqueville:
### Purpose of this PR

fix https://fogbugz.unity3d.com/f/cases/1358715
image

If one of the three shape offset parameters were overriden, the new parameter is marked overriden. Otherwise it still copies the values but is not overriden.

Replace the three shapeOffsetX, shapeOffsetY and shapeOffsetZ by one parameter shapeOffset, which causes an API change but I believe it's not problematic

---
### Testing status
Tested migration of a volume profile

---
### Comments to reviewers
@anisunity  for awareness


RemyUnity (approved):
LGTM 🟢

sebastienlagarde (approved):

sebastienlagarde:
please, launch yamato on trunk (it works)

PR #5471, opened 2021-08-30
[XR][URP] Fix issue with vignette location in XR

maloyer-unity:
Vignette center needs to respect the real view center since in XR, with asymmetric FOV, the center of the render taget is not always the center of the view. We deduce the real center via the projection matrix data and use that to offset the vignette's center.

### Checklist for PR maker

---
### Purpose of this PR
Fix the vignette post-process to respect the XR eye center (which is not the texture center because of AsymmetricFov). This caused the vignette effect to look like its overlapping between the eyes ([case 1358336](https://issuetracker.unity3d.com/issues/xr-sdk-urp-vignette-post-processing-effect-is-overlapping-between-eyes))

---
### Testing status
Tested on 
 - PC with Quest via Oculus Link
 - Quest 2 AndroidPlayer 64bit IL2CPP

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/XR/XRPass.cs
+++ b/com.unity.render-pipelines.universal/Runtime/XR/XRPass.cs
@@ -463,7 +463,7 @@ namespace UnityEngine.Rendering.Universal
             result.y = views[0].eyeCenterUV.y - centerDeltaY;
             if (singlePassEnabled)
             {
-                // With single-pass XR, we need a new variable to contain the data for the 2nd view
+                // With single-pass XR, we need to add the data for the 2nd view
                 result.z = views[1].eyeCenterUV.x - centerDeltaX;
                 result.w = views[1].eyeCenterUV.y - centerDeltaY;
             }


com.unity.render-pipelines.universal/Runtime/XR/XRPass.cs
```diff
@@ -46,6 +47,13 @@ internal XRView(Matrix4x4 proj, Matrix4x4 view, Rect vp, int dstSlice)
             viewport = vp;
             occlusionMesh = null;
             textureArraySlice = dstSlice;
+
+            var projectionParameters = proj.decomposeProjection;
+            float left = Math.Abs(projectionParameters.left);
+            float right = Math.Abs(projectionParameters.right);
+            float top = Math.Abs(projectionParameters.top);
+            float bottom = Math.Abs(projectionParameters.bottom);
+            eyeCenterUV = new Vector2(left / (right + left), top / (top + bottom));
```
fabien-unity, line 56: Can we add a method to avoid code duplication : computeEyeCenterUV(Matrix4x4 proj) ?

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
```diff
@@ -191,6 +201,10 @@ Shader "Hidden/Universal Render Pipeline/UberPost"
             UNITY_BRANCH
             if (VignetteIntensity > 0)
             {
+            #ifdef USING_STEREO_MATRICES
+                const float2 VignetteCenter = unity_StereoEyeIndex == 0 ? VignetteCenterEye0 : VignetteCenterEye1;
```
fabien-unity, line 205: In general, we'd like to be able to support more than 2 views. Can we evaluate moving this code to an array with indexing instead ?

com.unity.render-pipelines.universal/Runtime/XR/XRPass.cs
```diff
@@ -38,6 +38,7 @@ internal struct XRView
         internal readonly Rect viewport;
         internal readonly Mesh occlusionMesh;
         internal readonly int textureArraySlice;
+        internal readonly Vector2 eyeCenterUV;
```
fabien-unity, line 41: Quick note : this will create a merge conflict with https://github.com/Unity-Technologies/Graphics/pull/3557
Not a huge deal but let's discuss the order of PRs.

com.unity.render-pipelines.universal/Runtime/XR/XRPass.cs
maloyer-unity: Yes, we sure can !

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
maloyer-unity: Since this use case is very specific to eyes, I don't know if we would even need that for "more than two" displays. 
And, as far as I know, we don't have 3-eyes human, yet ;)

Also, for more than 2 display scenarios, like domes, super panoramic views or stuff like that we would need to change the whole vignette post-process to have a center in world space to be able to sync all views to work together, a screen-space effect might not work well in that case without being fully custom.

maloyer-unity: Update 2 : Returning to 2 views to keep the code simple and we'll add more when we need to.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -1251,6 +1251,32 @@ void SetupVignette(Material material)
                 m_Vignette.smoothness.value * 5f
             );
 
+#if ENABLE_VR && ENABLE_XR_MODULE
+            if (xrPass != null && xrPass.enabled)
+            {
+                var v3 = new Vector4(0.0f, 0.0f, 0.0f, 0.0f);
```
fabien-unity, line 1257: I'm being very picky but you can replace this by Vector4.zero :)

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -1251,6 +1251,32 @@ void SetupVignette(Material material)
...
+                float centerDeltaX = 0.5f - center.x;
+                float centerDeltaY = 0.5f - center.y;
+
+                XRView view0 = xrPass.views[0];
+                v2.x = view0.eyeCenterUV.x - centerDeltaX;
+                v2.y = view0.eyeCenterUV.y - centerDeltaY;
+                if (xrPass.views.Count == 1)
+                {
+                    v3.z = v2.x;
```
fabien-unity, line 1266: is this part ever used ? if we're in multipass mode (where views.Count == 1) then the _Vignette_Params3 will never be read in the shader.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -1251,6 +1251,32 @@ void SetupVignette(Material material)
...
+                    v3.w = v2.y;
+                }
+                else
+                {
+                    XRView view1 = xrPass.views[1];
+                    v3.x = view1.eyeCenterUV.x - centerDeltaX;
+                    v3.y = view1.eyeCenterUV.y - centerDeltaY;
+                }
+
+                material.SetVector(ShaderConstants._Vignette_Params3, v3);
```
fabien-unity, line 1276: Even this part should really only be set if we're in multiview/spi situation (viewCount > 1)

maloyer-unity: Yes, good point, I'll make that part of the code only used in single-pass mode and run some tests.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
mikesfbay: One of our partners is moving away from asymmetric frustums. We should keep this fix as simple as possible. What about extracting the center from the current stereo projection matrix in either vertex or pixel shader directly? Then there's no need to pass in per eye centers or changing script code.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -1502,6 +1521,7 @@ static class ShaderConstants
             public static readonly int _Chroma_Params = Shader.PropertyToID("_Chroma_Params");
             public static readonly int _Vignette_Params1 = Shader.PropertyToID("_Vignette_Params1");
             public static readonly int _Vignette_Params2 = Shader.PropertyToID("_Vignette_Params2");
+            public static readonly int _Vignette_Params3 = Shader.PropertyToID("_Vignette_Params3");
```
eh-unity, line 1524: Maybe call this "_Vignette_ParamsXR" since that's what it is and it's easier to read later.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -1251,6 +1251,25 @@ void SetupVignette(Material material)
                 m_Vignette.smoothness.value * 5f
             );
 
+#if ENABLE_VR && ENABLE_XR_MODULE
```
eh-unity, line 1254: Maybe move the XR specific block into a function, easier to read than bunch of defines. 
That function could even be in xrPass. 
Something like
vec4 xrPass.applyXREyeOffset(vec2 center); which packs XR eye data into Vec4. Or xrPass.applyEyeOffset( ref vec2 centerAsLeft, out vec2 right);

But just moving this block to SetupVignetteXR() would be tiny bit better IMO.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
```diff
@@ -68,7 +71,12 @@ Shader "Hidden/Universal Render Pipeline/UberPost"
         #define LensDirtIntensity       _LensDirt_Intensity.x
 
         #define VignetteColor           _Vignette_Params1.xyz
+    #ifdef USING_STEREO_MATRICES
+        #define VignetteCenterEye0      _Vignette_Params2.xy
```
eh-unity, line 75: Since we're already differentiating with center and Eye0, Eye1 we could just pack all the XR Eye data to Params3 (or ParamsXR). I think it would be a tiny bit cleaner as it would avoid this interleaving
`params2( standard.xy, XR0.xy), params3( XR1.xy,  --/standard.xy? )` 
pattern, and be 
`params2( standard.xy, --/standard.xy? ), params3( XR0.xy, XR1.xy ) or paramsXR( XR0.xy, XR1.xy )`
as it doesn't matter where the 2 extra floats are.

eh-unity (commented):
Looks good, but added some improvement suggestions.

ernestasKupciunas (approved):
Approving without testing. The developer has already tested his changes on Ques2. I have only Quest2 at home too, so there is no point testing twice on the same platform.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
maloyer-unity: I don't expect this change to prevent symmetric FOV from being used, it will only allow asymmetric FOV to work correctly. This change will work with both symmetric and asymmetric FOV, the center calculated for symmetric FOV will be (0.5, 0.5) for both eyes since in that case leftFov == rightFov && topFov == bottomFov. 

Extracting the data from the shader (Vertex or Fragment) will make this change more complicated than it currently is since this is a post process and uses a generic Vertex Shader which does not have information about the projection. Changing that will introduce code duplication and as a side effect, more possible point of failures.

I believe that the eye center should be calculated on the CPU since it is something that will likely be used by other post process or shaders that rely on knowing where the center of the view is (any type of radial post-process, some screen space effects or more), it is also something that I expect some games will want to know too (I used these values for a few things in StarWars Squadrons VR). 

In fact, I was very surprised that the XR view center coordinates was not already available in our XR C# code !

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
maloyer-unity: Yes, we can do that

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
maloyer-unity: I think we can extract a good amount of code to the XRPass class, but we still need to do some part of the work in SetupVignette() since when we are rendering in Multi-pass (each eye are rendered completely separately), the center still need to be corrected for each eyes and then placed in v2.xy since AFAIK the shader does not have a variation for "XR enabled, but not single-pass". 

I'll see what I can do to reduce the amount of XR specific code in here.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
maloyer-unity: Sure, I'll get the XR single-pass data to its own variable then to group it all together

fabien-unity (approved):

maloyer-unity:
Hi everyone, I would like to close this bug, since there is no more action items on my side, is it possible to resume the review ?

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
mikesfbay: Ok, the points sound reasonable.

mikesfbay:
Need XR QA to do some regression testing on this PR.

eh-unity (approved):

fcoulombe (approved):
LGTM

JasonCostanza (approved):
Version: 2022.1.0a10.1326
Revision: trunk 48c8cfa2acec
Built: Fri, 17 Sep 2021 07:51:06 GMT

commit 85c33751f57794f18789571a4b2e1149c886b22c (HEAD -> urp/xr/fix_vignette_xr_center, origin/urp/xr/fix_vignette_xr_center)

From what I can tell, this appears to look good to me. I spoke to Marc-Andre on slack and compared notes and I believe what I'm seeing is the desired effect. Basically, the vignette tunnels appear to align in the eyes creating the illusion that only one is being drawn. From my tests, the vignette felt correct and natural and I couldn't spot any sign of misalignment.

maloyer-unity:
@mikesfbay, Ryan was not available to do a QA pass, but Jason was able to have look, is there anything else that is needed before we can close this issue?

@phi-lira, As soon we have the final approval, the fix will be ready to merge into the new URP staging branch, I'm just not sure how we can change the upstream target of a PR.

mikesfbay (approved):

PR #5472, opened 2021-08-30
Fix darkening in SSR fade 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1340425/ 

The issue was that SSR was premultiplied by its alpha, but before applying SSR results on screen we multiplied alpha again, causing a darkening of the value. 

This fixes the issue by applying the multiply at reading only when Raytracing is enabled.


sebastienlagarde (approved):

remi-chapelain (approved):
Small fix, great improvement for SSR quality :)
Had a quick look at DXR reflection tests to see if that changes anyting there. All good ✔️

PR #5474, opened 2021-08-30
Surface ReflectionTypeLoadExceptions in HDUtils.GetRenderPipelineMaterialList()

pastasfuture:
Without surfacing these exceptions, developers cannot act on any underlying reflection errors in the HDRP assembly.

https://unity.slack.com/archives/CE7DZN2H1/p1630106304084900

An example error message that this will now generate (from the case I hit while developing, which previously silently failed):
`Encountered an exception while attempting to reflect the HDRP assembly to extract all RenderPipelineMaterial types.
This exception must be fixed in order to fully initialize HDRP correctly.
Generic class cannot have explicit layout.
UnityEngine.Rendering.HighDefinition.AtlasAllocatorDynamic`1/AtlasNode`
To be clear - the example error will not fire on master - that is just an example from code changes I was making on a branch.

Outside of the scope of this PR: I wonder if it would be better to not rely on reflection here, and instead just manually populate the list as mentioned in the comment with all RenderPipelineMaterial types - the list is fairly short, and I would expect this reflection the be the cause of maybe > 1 second during initialization time.


sebastienlagarde:
Adding Remi to get his though.

Note that we can have Material define outside of HDRP package (it is the case for automotive), thus why we rely on the reflection mechanism.

RSlysz:
> Adding Remi to get his though.
> 
> Note that we can have Material define outside of HDRP package (it is the case for automotive), thus why we rely on the reflection mechanism.

I agree. We cannot do it with hard coded list. But perhaps we can do it more efficiently using the TypeCache instead of C# reflection.

RSlysz (approved):
Proposed version is better than what we currently have.

But we should plan some improvement on it later:
- Is there a similar mechanisme in URP? If yes, we should handle this in CoreSRP directly.
- Using TypeCache will only gather type that are already loaded. Thus it already handled the load exception. It would also be more efficient this way than using reflection.

sebastienlagarde (approved):

sebastienlagarde:
> But we should plan some improvement on it later:
> 
> * Is there a similar mechanisme in URP? If yes, we should handle this in CoreSRP directly.

Nope

PR #5485, opened 2021-08-31
[HDRP] Improve path traced subsurface scattering for transmissive surfaces

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1329403/

There was a problem when computing subsurface scattering for transmissive thin/open surfaces

This is what the rasterized scene looks like:
image

Path tracer BEFORE this PR:
image

Path tracer AFTER this PR:
image

(The text is missing from the path traced images because the emitted UI geometry is not part of the acc structure)

---
### Testing status
Tested using the scene that was attached to the bug ticket, as you see above

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -378,6 +378,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed case where the SceneView don't refresh when using LightExplorer with a running and Paused game (1354129)
 - Fixed wrong ordering in FrameSettings (Normalize Reflection Probes)
 - Fixed a null ref exception when no opaque objects are rendered.
+- Fixed issue with depth slope scale depth bias when a material uses depth offset.
+- Fixed shadow sampling artifact when using the spot light shadow option 'custom spot angle'
+- Fixed issue with fading in SSR applying fade factor twice, resulting in darkening of the image in the transition areas.
 - Fixed path traced subsurface scattering for transmissive surfaces (case 1329403)
 
 ### Changed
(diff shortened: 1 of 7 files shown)


RemyUnity (approved):
LGTM, thanks for the comparison screenshots 🟢

eturquin (approved):

remi-chapelain:
How about adding this as a test scene in the DXR Test project ?

pmavridis:
> How about adding this as a test scene in the DXR Test project ?

That's a good idea, I will followup with another PR, after I finish with some other work.

remi-chapelain:
> That's a good idea, I will followup with another PR, after I finish with some other work.

Thanks 🙏

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingBSDF.hlsl
```diff
@@ -607,6 +607,16 @@ bool RandomWalk(float3 position, float3 normal, float3 diffuseColor, float3 mean
         result.exitNormal = intersection.value;
     }
 
+// if the material is transmissive, pick a side for the normal randomly to properly handle thin surfaces (case 1329403)
+#ifdef _MATERIAL_FEATURE_TRANSMISSION
```
sebastienlagarde, line 611: I will accept the PR for the fix, but let's have in mind that _MATERIAL_FEATURE_TRANSMISSION isn't a standard name for "transmission". This is define only per Material and Material don't have obligation to reuse this name (on raster side). We may need to revisit how we handle transmission from a general point of view later.

sebastienlagarde (approved):

sebastienlagarde:
As mention by Remi, please add a test in another PR

PR #1475, opened 2020-07-31
Fixing gaussian dof/bokeh dof/ssao/fxaa/motion blur to work with dynamic resolution

lukaschod:
# **Please read;**
## **PR Workflow for the Graphics repository:**
* **All PRs must be opened as draft initially**
* Reviewers can be added while the PR is still in draft
* The PR can be marked as “Ready for Review” once the reviewers have confirmed that **no more changes are needed**
* Tests will start automatically after the PR is marked as “Ready for Review”
* **Do not use [skip ci]** - this can break some of our tooling
* Read the [Graphics repository & Yamato FAQ](http://go/graphics-yamato-faq).

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ]  Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [x] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Fix for the case 1225467 (Enterprise Support). The fixed isssue is that depth of field shaders depended on _TexelSize property for effect calculation. However the property does not account for dynamic resolution, which resulted in incorrect calculations. With the fix now its passing custom property _SourceSize.
Fix other post processing effects to work with dynamic resolution (fxaa, ssao, motion blur).

---
### Testing status

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [x] Built a player
- [x] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in developer mode, you have a button at end of resources that check the paths)
- Other: 

**Automated Tests**: Added gfx test 133_DepthOfField_DynamicResolution
image

[...]

PR #5487, opened 2021-08-31
[2022.1] Fixing Dynamic Resolution with Render Textures

ellioman:
# Purpose of this PR
This PR fixes issue 1361679.

The problem was that when we calculated the normalized UV's in our shaders, the code was using `_ScaledScreenParams`. The problem with that is when Dynamic Resolution is enabled and a camera renders to a render texture, the render texture has already been affected by the Dynamic Resolution, so the UV's were incorrectly calculated. 

The solution is to use `_SourceSize` which was introduced in #1475. This PR does that change and also exposes a function for getting `_SourceSize` and updates various files, in URP and ShaderGraph, so they call that function instead of declaring the parameter and using it directly. 

# Yamato


eh-unity (commented):
Looks ok, but it's getting confusing with ScreenParams, ScaledScreenParams and SourceSize. It could be DynamicScaledScreenParams or DynamicScreenParams for example.

I wonder if we should do a URP specific struct instead that can be extended and where things are clearly named. Pass the other sizes too but with clear naming and without the 1+1/width thing. Keep the old vars for backwards compatibility for now.

SourceSize is not really consistently named. Thats both good and bad. Good because it's not defined the same way. Bad that it's almost the same thing but completely different, so harder to find.

Also Source is ambiguous and at a global level. As in "source of what?". This is the main reason I'm not liking it too much.

Is it consistent with HDRP?

phi-lira:
Adding @sandy-carter-unity as reviwer as he's working on RTHandle and scaling.

PR #5499, opened 2021-09-01
Prep work for APV cell streaming

JulienIgnace-Unity:
### Purpose of this PR
Preparation work for cell streaming:
- Refactor to split properly loading cell from the asset and uploading data to GPU
- Renamed things a bit: Add for adding a cell from asset and Load to upload it to GPU
- Handled out of memory (both index and texture) more gracefully: Cell isn't uploaded but still available in CPU memory
- Cell debug is now red for cells that aren't "loaded" (uploaded to GPU)
- Small refactor of debug gizmo to unify APV debug
- Precompute of cell chunk cost
- Added an update (empty for now) function to handle streaming update

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.cs
@@ -294,12 +294,8 @@ namespace UnityEngine.Experimental.Rendering
                             if (ShouldCullCell(cell.position, prv.GetTransform().posWS))
                                 continue;
 
-                            var positionF = new Vector3(cell.position.x, cell.position.y, cell.position.z);
-                            var center = positionF * profile.cellSizeInMeters + profile.cellSizeInMeters * 0.5f * Vector3.one;
-                            var positionF = new Vector3(cell.position.x, cell.position.y, cell.position.z);
-                            var center = positionF * cellSizeInMeters + cellSizeInMeters * 0.5f * Vector3.one;
                             var positionF = new Vector4(cell.position.x, cell.position.y, cell.position.z, 0.0f);
-                            var center = positionF * profile.cellSizeInMeters + profile.cellSizeInMeters * 0.5f * Vector4.one;
+                            var center = positionF * cellSizeInMeters + cellSizeInMeters * 0.5f * Vector4.one;
                             center.w = cellInfo.loaded ? 1.0f : 0.0f;
                             yield return center;
                         }
@@ -308,13 +304,9 @@ namespace UnityEngine.Experimental.Rendering
 
                 foreach (var center in GetVisibleCellCentersAndState())
                 {
-                    Gizmos.DrawCube(center, Vector3.one * profile.cellSizeInMeters);
-                    cellGizmo.AddWireCube(center, Vector3.one * profile.cellSizeInMeters, new Color(0, 1, 0.5f, 1));
-                    Gizmos.DrawCube(center, Vector3.one * cellSizeInMeters);
-                    cellGizmo.AddWireCube(center, Vector3.one * cellSizeInMeters, new Color(0, 1, 0.5f, 1));
                     bool loaded = center.w == 1.0f;
-                    m_MeshGizmo.AddWireCube(center, Vector3.one * profile.cellSizeInMeters, loaded ? new Color(0, 1, 0.5f, 1) : new Color(1, 0.0f, 0.0f, 1));
-                    m_MeshGizmo.AddCube(center, Vector3.one * profile.cellSizeInMeters, loaded ? new Color(0, 1, 0.5f, 0.2f) : new Color(1, 0.0f, 0.0f, 0.2f));
+                    m_MeshGizmo.AddWireCube(center, Vector3.one * cellSizeInMeters, loaded ? new Color(0, 1, 0.5f, 1) : new Color(1, 0.0f, 0.0f, 1));
+                    m_MeshGizmo.AddCube(center, Vector3.one * cellSizeInMeters, loaded ? new Color(0, 1, 0.5f, 0.2f) : new Color(1, 0.0f, 0.0f, 0.2f));
                 }
             }
 


alelievr (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1121,6 +1121,8 @@ protected override void Render(ScriptableRenderContext renderContext, Camera[] c
 #endif
             Terrain.GetActiveTerrains(m_ActiveTerrains);
 
+            ProbeReferenceVolume.instance.Update();
```
FrancescoC-unity, line 1124: We already have two sort of updates already. 

We have the pending operation stuff (called every frame) and we have a SortPendingCells in the Lightloop. We should integrate the streaming in one of the two. I am fine with having the `SortPendingCells ` into the new `Update` but also I think it should live in the lightloop as it is lighting related only.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -295,20 +295,21 @@ internal class Cell
             public SphericalHarmonicsL2[] sh;
             public float[] validity;
             public int minSubdiv;
-
-            [System.NonSerialized]
-            public int flatIdxInCellIndices = -1;
-
-            [System.NonSerialized]
-            public bool loaded = false;
+            public int indexChunkCount;
```
FrancescoC-unity, line 298: Do we need these 2 serialized?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
```diff
@@ -63,7 +63,7 @@ internal void Cleanup()
 
         const int kMaxPoolWidth = 1 << 11; // 2048 texels is a d3d11 limit for tex3d in all dimensions
 
-        int m_AllocationSize;
+        static int m_AllocationSize;
```
FrancescoC-unity, line 66: It's a bit weird this is static given it is added in the constructor of the pool itself

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
```diff
@@ -140,8 +146,8 @@ internal void Allocate(int numberOfBrickChunks, List<BrickChunkAlloc> outAllocat
             {
                 if (m_NextFreeChunk.z >= m_Pool.depth)
                 {
-                    Debug.Assert(false, "Cannot allocate more brick chunks, probevolume brick pool is full.");
-                    break; // failure case, pool is full
+                    Debug.LogError("Cannot allocate more brick chunks, probe volume brick pool is full.");
```
FrancescoC-unity, line 149: Ok for now, but we should remember to remove the error when streaming is in. Pool full should just mean that a load/unload operation needs to happen.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeIndexOfIndices.cs
```diff
@@ -57,6 +57,8 @@ internal void Pack(out uint[] vals)
             }
         }
 
+        static uint[] s_PackedValues = new uint[kUintPerEntry];
```
FrancescoC-unity, line 60: While I agree with this change, can we move it inside IndexMetaData ? So that we don't confuse it for something else elsewhere.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -660,36 +657,45 @@ public int GetVideoMemoryCost()
...
-                    m_CellToBricks.Remove(cell);
-                }
-            }
+            ReleaseBricks(cellInfo);
+
+            cellInfo.loaded = false;
+            cellInfo.updateInfo = new ProbeBrickIndex.CellIndexUpdateInfo();
 
-            cell.loaded = false;
+            ClearDebugData();
```
FrancescoC-unity, line 678: Why clearing the whole debug data?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -921,52 +927,29 @@ void LoadPendingCells(bool loadAll = false)
...
-                    return;
-                }
+                m_CellsToBeAdded.RemoveAt(0);
             }
         }
 
         /// <summary>
         /// Perform all the operations that are relative to changing the content or characteristics of the probe reference volume.
         /// </summary>
         /// <param name ="loadAllCells"> True when all cells are to be immediately loaded..</param>
```
FrancescoC-unity, line 946: Need to update doc here

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -1123,9 +1124,11 @@ RegId AddBricks(List<Brick> bricks, ProbeBrickPool.DataLocation dataloc, ProbeBr
                 }
             }
 
-            // Update the pool and index and ignore any potential frame latency related issues for now
+            // Update textures with incoming SH data and index and ignore any potential frame latency related issues for now
```
FrancescoC-unity, line 1127: The comment seems less precise now. 

What textures? We have intermediate textures. Also without commas now it looks like we are updating textures with index data. 

Maybe
> Update pool textures with incoming SH data and ignore any potential frame latency related issues for now. 

And move the index part straight above the m_Index.AddBricks

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -1135,29 +1138,42 @@ RegId AddBricks(List<Brick> bricks, ProbeBrickPool.DataLocation dataloc, ProbeBr
...
+            m_Index.AddBricks(id, bricks, ch_list, ProbeBrickPool.GetChunkSize(), m_Pool.GetPoolWidth(), m_Pool.GetPoolHeight(), cellUpdateInfo);
+
+            // Update CellInfo
+            cellInfo.regId = id;
+            cellInfo.updateInfo = cellUpdateInfo;
+            cellInfo.chunkList = ch_list;
+            cellInfo.loaded = true;
+
+            // Update indirection buffer
+            m_CellIndices.UpdateCell(cellInfo.flatIdxInCellIndices, cellUpdateInfo);
```
FrancescoC-unity, line 1150: It is a bit weird here that we are updating cell info and the cell only indirection buffer in something called AddBricks, I'd prefer if we have this Cell level changes outside.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
JulienIgnace-Unity: I agree and it's true that it's dangerous but I think that this actually uncovers an issue.
This allocation size is set at runtime init which means that you can theoretically change it after baking.
It's currently hard coded in HDRP init code. Would you mind if I made it a constant like we have for probe indices? Or is this something that we want to really expose to users later on?
In any case, the reason why it's static is because I wanted to precompute that at bake time because it feels like the right place to do that and from there I don't have access to runtime pool etc.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
JulienIgnace-Unity: Yup, that's the plan. Just wanted to have something explicit in the meantime

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeIndexOfIndices.cs
JulienIgnace-Unity: I'll do that yeah.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: I currently set them at bake time. I feel like it should stay there to be able to keep compatibility if we change values etc.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: Currently this only clears probe debug (that's the only persistant part) but semantically yeah, for now we clear the whole thing.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: Good catch

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: Agreed (not even sure why I changed that tbh)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
JulienIgnace-Unity: Didn't really think too hard on this one so I can change that.
The only constraint we have is that it needs to be called once for all cameras at some point because we'll need to factor N cameras to derive the proper streaming needs for a frame.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: It's mostly because iirc it was already kinda split like that. What I did was move back pool update in the same function AddBrick. I still intend to clean this further at some point as it's a bit messy right now. Also I'd like to find a better name than "AddBricks" because IMO it does not really convey what this does (aka, allocate and update indices and probe sh data).

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
FrancescoC-unity: I think is fine to move for now as constant, I don't see much need to have it as settable for user for now, can always change later.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
FrancescoC-unity: Before the cell level stuff was done in LoadingPendingCell _after_ the add bricks. 

And yes the name is not great indeed, but let's keep it for this PR and change it after.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
JulienIgnace-Unity: So I took a closer look and both SortPendingCells and the pending loading stuff are called per camera. So I can't use that for stream. I don't see any other way then.

FrancescoC-unity: Currently is done per camera because it is a quick hack, it should follow the same heuristic of (what's the highest priority cell) that streaming is using. 

The sorting happens on the monolithic system still, so 2 camera stomp each other (it is a rough and very temp system before streaming has a good heuristic)

JulienIgnace-Unity: Hm all right. Ok then, I'm removing this one and I'll probably have to change the existing in my next PRs then.

FrancescoC-unity (approved):

JulienIgnace-Unity:
Work will be continued from https://github.com/Unity-Technologies/Graphics/pull/5731

PR #5503, opened 2021-09-01
Fix instability detection tagging

kivistikliis:
---
### Purpose of this PR
- Removes adding a redundant `unknown` tag, if any pattern from any log is matched (except for successful-retry)
- Overwrites matched pattern conclusion (which is usually `failure`) to be `success` for all successful-retries
- Enforces having either `successful-retry` or `instability` tag, never both (having both would complicate distinguishing between successful retries, failed instabilities, and failed retries which are not instabilities. Refer to docs for more elaborate explanation)
- Some other smaller rewrites

---
### Testing status
Tested locally on several execution logs, as well as the new ones from [this ](https://unity-ci.cds.internal.unity3d.com/job/8476277) and [this ](https://unity-ci.cds.internal.unity3d.com/job/8476213) job, which brought these issues to attention.

---
### Comments to reviewers
This is properly tested via Yamato, but as instabilities occur by random, the best way is to investigate the output after nightlies.

### DIFF
--- a/.yamato/script/log_scripts/utils/constants.py
+++ b/.yamato/script/log_scripts/utils/constants.py
@@ -2,3 +2,9 @@
 UTR_LOG = 'utr_log'
 UNITY_LOG = 'unity_log'
 EXECUTION_LOG = 'execution_log'
+
+
+# tags
+TAG_SUCCESFUL_RETRY = 'successful-retry'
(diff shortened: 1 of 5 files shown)


kivistikliis:
Merging without explicit approval, as this was discussed and approved internally.

PR #5502, opened 2021-09-01
Do not open a VFX Graph editor when an asset is assigned in the inspector

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1347399/
and https://fogbugz.unity3d.com/f/cases/1349251/ (see bottom of this comment)

When assigning a VFX asset to VisualEffect GameObject through the inspector an new empty VFX Graph editor was opened.
The fix consist to prevent this empty editor to be opened (no editor is opened at all now).

https://user-images.githubusercontent.com/85612173/131680812-655522c4-8fd2-48d1-b096-01d0892cec61.mp4

---
### Testing status
I only tested the scenario described in the fogbugz.

---
### Comments to reviewers
It happens that this bug fix also fixes this other issue: https://fogbugz.unity3d.com/f/cases/1349251/
This is because when modifying any value in the inspector of the VisualEffect would give focus to the current VFX editor window, preventing the user from continuing typing in the input field.

But even worse, if no VFX editor was opened it was opening an empty one.

PR #5505, opened 2021-09-01
Vfx/fix/1349251 random seed edition improved

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1349251/
User could not type in multiple numbers in a row to change the random seed value because each new character moves the focus out of the input field because it's applied immediately.

Now the user can type in any number and validate the value with Enter
HMZf7xYPST

---
### Testing status
N/A

---
### Comments to reviewers
The behavior is a bit different since before the value with immediately taken into account (losing the focus). Now the value is taken into account only after the user press Enter.


VitaSkruibyte:
Hi, thanks for the fix! A few concerns:
- Is it intended that Random Seed is no longer displayed in real numbers? 
3ZCJibwOtM

- [Unrelated to PR] Reseed on play toggle and entering a new value leads to VFX window being focused. Could this be looked into in this PR? 
LMBsDGmDgy

julienamsellem:
This PR is abandoned since the issue has been fixed by this PR https://github.com/Unity-Technologies/Graphics/pull/5502

PR #5507, opened 2021-09-01
[10.x] Define ENABLE_HYBRID_RENDERER_V2 by default for Hybrid Renderer 0.14.0 and newer [V1]

vincent-breysse:
Enable HRV2 by default for DOTS 0.5 since HRV1 will be removed.
---
Unfortunately, we noticed a bit too late that more asmdef files had to be updated. The V2 can be found here https://github.com/Unity-Technologies/Graphics/pull/5635
### Testing status
This change is supposed to be tested with HRV1 removed on a project not defining `ENABLE_HYBRID_RENDERER_V2` explicitly in the ProjectSettings. 
Since HRV1 removal has not landed on DOTS master yet, I tested by locally adding the same asmdef change to the file `com.unity.rendering.hybrid/Unity.Rendering.Hybrid/Unity.Rendering.Hybrid.asmdef`. This is to ensure I wasn't testing with the define enabled globally for the project, which would be pointless.

- **2020.3/partner/staging** (735bfa915849a78b9b58864119288b6e2f200704)
- **DOTS master** (459ac3fdb66af053a90fec111bf6edcea836ff46)

### com.unity.rendering.hybrid
- Added: Define `ENABLE_HYBRID_RENDERER_V2` by default for Hybrid Renderer 0.14.0 and newer

### DIFF
--- a/com.unity.shadergraph/Editor/Utilities/Unity.ShaderGraph.Utilities.asmdef
+++ b/com.unity.shadergraph/Editor/Utilities/Unity.ShaderGraph.Utilities.asmdef
@@ -10,5 +10,11 @@
     "precompiledReferences": [],
     "autoReferenced": false,
     "defineConstraints": [],
-    "versionDefines": []
+    "versionDefines": [
+        {
+            "name": "com.unity.rendering.hybrid",
+            "expression": "0.14.0-preview.0",
+            "define": "ENABLE_HYBRID_RENDERER_V2"
+        }
+    ]
 }


joeldevahl (approved):

com.unity.shadergraph/Editor/Utilities/Unity.ShaderGraph.Utilities.asmdef
```diff
@@ -10,5 +10,11 @@
     "precompiledReferences": [],
     "autoReferenced": false,
     "defineConstraints": [],
-    "versionDefines": []
+    "versionDefines": [
+        {
+            "name": "com.unity.rendering.hybrid",
+            "expression": "0.14.0-preview.0",
+            "define": "ENABLE_HYBRID_RENDERER_V2"
```
cdxntchou, line 17: I assume from the way this is set up that this is only defined when the given package with the given version number is present?

Not an expert on the asmdef syntax, but if that is true, then this looks correct.  :)

cdxntchou (approved):

com.unity.shadergraph/Editor/Utilities/Unity.ShaderGraph.Utilities.asmdef
vincent-breysse: Yeah almost. It is meant to define `ENABLE_HYBRID_RENDERER_V2` when the hybrid renderer package version `0.14.0-preview.0` _or newer_ is present. So it should apply to 0.15 and etc for example. 

In case you want the sources, it's from the Unity doc :

```
As a shortcut, you can enter a single version without range brackets to indicate that the expression includes that version or later:

2.1.0-preview.7 evaluates to x >= 2.1.0-preview.7
```

https://docs.unity3d.com/2021.2/Documentation/Manual/ScriptCompilationAssemblyDefinitionFiles.html

PR #5508, opened 2021-09-01
Add Test Coverage for Custom Velocity and Precomputed Velocity

johnpars:
### Purpose of this PR

This PR adds Test 9304, adding further motion vector coverage for the recently added Custom Velocity shader graph input, as well as Precomputed Velocity from Alembic. 

9304_MotionVectorsPrecomputedAndCustomVelocity

Various information about this test:
- Forces HDRP into Motion Vector Debug View. 
- Top Row: Custom Velocity + Vertex Animation
- Middle Row: Precomputed Velocity (Alembic) + Vertex Animation
- Bottom Row: Custom Velocity + Precomputed Velocity + Vertex Animation
- Covers all primary HDRP material types (Lit, Fabric, etc.).
- Since we are using the debug view, contextual text was not added.

---
### Comments to reviewers
**Very important note**: I made a slight addition to **HDCamera.cs** that allows internal assemblies to override the GPU / material animation time. This mimics the behavior of [this](https://docs.unity3d.com/ScriptReference/EditorUtility.SetCameraAnimateMaterialsTime.html) editor API utility, which sadly does not work for HDRP. It was necessary to add this to achieve stable, deterministic results for this test (specifically stability with vertex animation motion vectors).


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -1125,6 +1129,13 @@ unsafe internal void UpdateShaderVariablesGlobalCB(ref ShaderVariablesGlobal cb,
             float ct = time;
             float pt = lastTime;
 #if UNITY_EDITOR
+            // Apply editor mode time override if any.
+            if (animateMaterials)
```
johnpars, line 1133: Important addition to HDCamera to achieve deterministic motion vector results for vertex animation.

com.unity.testing.hdrp/TestRunner/OverrideTime.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+using UnityEngine;
+using UnityEngine.Rendering.HighDefinition;
+
+// Utility for achieving deterministic motion vector results with vertex animation in play mode.
+public class OverrideTime : MonoBehaviour
```
johnpars, line 6: Override is utilized by this utility, part of internal HDRP testing assemblies.

sebastienlagarde (approved):

PR #5509, opened 2021-09-02
Moving WS view direction from vertex to fragment to lower varying size

skerys:
---
### Purpose of this PR
These optimizations were discussed with Apple.

mobile optimizations to lower bandwidth:
completely removed viewDirectionWS variable declaration from StructField.Varyings struct and also the usage of the variable from the shader template as well.

For WorldSpaceViewDirection that is used in the ShaderGraph node, replaced the calculation from:

normalize(input.viewDirectionWS)

to:

normalize(GetWorldSpaceViewDir(input.positionWS)

Also changed the field dependency of the WorldSpaceViewDirection node to now requiring StructFields.Varyings.positionWS variable.

---
### Testing status
local comparison test for the View Direction Node in editor and on an iOS device
all URP/all Shadergraph on Yamato

---


skerys:
All the failures from the last run are in master too and shouldn't be related to this PR

cdxntchou (approved):
Looks good!
HDRP seems to already use this approach, so only URP has the changes...

phi-lira (approved):
LGTM.

simon-engelbrecht-soerensen (approved):
Did a comparison test and it seems to work. 👍

GrantLamb-Unity (commented):
@simon-engelbrecht-soerensen 's comparison tests should be enough for this change 👍

phi-lira (approved):

PR #5514, opened 2021-09-02
Do not recompile the VFX when property min/max value change

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1307648/

The VFX Graph get recompile every time the min or max values of a property is changed
Compiling-issue

---
### Testing status
I tested the scenario described in the fogbugz

---
### Comments to reviewers
When _number_ property is set as a range the user can manipulate a slider to pick a value.
This slider has a minimum and a maximum value.
While testing I notice a few annoying behaviors related to this:
- ❌When the mode is switched to "Range" the min, max and value get the same value
- ❌The user can easily set a minimum > maximum or the opposite, nothing is done against that 
- ✅When changing the min or the max value the value is set to the max


com.unity.visualeffectgraph/Editor/Controls/VFXSliderField.cs
```diff
@@ -130,9 +131,10 @@ public Vector2 range
                     m_Slider.lowValue = m_Range.x;
                     m_Slider.highValue = m_Range.y;
 
-                    m_Slider.value = m_Range.x;
-                    m_Slider.value = m_Range.y;
-                    m_Slider.value = ValueToFloat(this.value);
+                    if (m_Slider.value < m_Slider.lowValue || m_Slider.value > m_Slider.highValue)
```
julienamsellem, line 134: Only change the value if it's out of bounds

VitaSkruibyte (approved):
Looks neat! Tested:
- Slider in Blackboard and Inspector of exposed property on vfx (float, arcshapes (+gizmos))
- Auto compilation/manual compilation

PaulDemeulenaere (approved):
Looks good to me, the usage of settings invalidation only when it's need is more elegant.
Thanks for the fix.

PR #5519, opened 2021-09-02
[Quality Audit] Make MenuItem dynamically hiddeable for LookDev and RenderGraphWindow [OnHold]

RSlysz:
### Purpose of this PR
MenuItem are not dynamically hideable normally. Here we added this possibility.
For LookDev and RenderGraph, we can find which RP can use them from CoreRP. So there is dedicated rules to them. This is a change from first attempt ([MenuItemForRenderPipeline]). As the attribute is no use right now, it is not in the final version of the PR. (It can still be added later though)

This is a requirement for Quality Audit in order to not have so many state (currently it can be "not here", "here but disabled", "here and enabled")

---
### Testing status
Tested Manually:
- Persistent through DomainReload (meanings the recomputation give same result)
- Tested LookDev and RenderGraph both show up in HDRP and not in Universal nor Built-in

---
### Comments to reviewers
- Also adding missing namespace on the RenderGraphViewer

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
+++ b/com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
@@ -14,8 +14,6 @@ namespace UnityEditor.Rendering.Utilities
         {
             InitMethodsWithReflection();
 
-            ConstructMenuItemForRenderPipeline();
-
             RenderPipelineManager.activeRenderPipelineTypeChanged += Update;
 
             // PR on Hold due to this:
@@ -29,9 +27,6 @@ namespace UnityEditor.Rendering.Utilities
         {
             UpdateRenderGraph();
             UpdateLookDev();
-
-            // should always be updated last
-            UpdateMenuItemForRenderPipeline();
         }
 
         #region Reflection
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        public MenuItemForRenderPipeline(string path, bool validate = false, int priority = 0, params Type[] renderPipelineAssetTypes)
+        {
+            if (renderPipelineAssetTypes == null || renderPipelineAssetTypes.Length == 0)
+                throw new ArgumentNullException("Argument cannot be null. Use MenuItem instead if you don't need to filter per RenderPipeline.", "renderPipelineTypes");
+
+            // Note: rpAssetType == null possible and should mean built-in
+            // but this is not possible as a key in s_MenuMap so we use the forbiden RenderPipelineAsset type for that
+            foreach (Type type in renderPipelineAssetTypes)
+                if (!typeof(RenderPipelineAsset).IsAssignableFrom(type) && type != null)
```
alex-vazquez, line 149:
```suggestion
                if (type != null && !typeof(RenderPipelineAsset).IsAssignableFrom(type))
```
I think is better to check if it is null before

com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
```diff
@@ -0,0 +1,159 @@
...
+                    throw new ArgumentException("Argument must be a RenderPipelineAsset child type", "renderPipelineTypes");
```
alex-vazquez, line 150:
```suggestion
                    throw new ArgumentException("Argument must be a RenderPipelineAsset child type", nameof(renderPipelineTypes));
```

com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            this.renderPipelineAssetTypes = renderPipelineAssetTypes
+                .Select(rpat => rpat != null ? rpat : typeof(RenderPipelineAsset)).ToArray();
```
alex-vazquez, line 153: I think we should throw an exception if the typeof is null or if the typeof is RenderPipelineAsset. This will mean invalid data. I think attribute must only be used to specify

com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
```diff
@@ -0,0 +1,159 @@
...
+
+    [AttributeUsage(AttributeTargets.Method)]
+    public class MenuItemForRenderPipeline : Attribute
+    {
+        public readonly Type[] renderPipelineAssetTypes;
+        public readonly string path;
+        public readonly int priority;
+        public readonly bool validate;
+
+        public MenuItemForRenderPipeline(string path, bool validate = false, int priority = 0, params Type[] renderPipelineAssetTypes)
```
alex-vazquez, line 141: Are we sure we want the asset and not the pipeline?

com.unity.render-pipelines.core/Editor/RenderGraph/RenderGraphViewer.cs
```diff
@@ -1,690 +1,691 @@
 using UnityEngine;
```
alex-vazquez, line 1: Is this file from this PR?

com.unity.render-pipelines.high-definition/Editor/ToolMenus.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace UnityEditor.Rendering.HighDefinition
+{
+    class ToolMenus
+    {
+        [MenuItemForRenderPipeline("Window/Rendering/Look Dev", false, 10001, typeof(HDRenderPipelineAsset))]
+        static void OpenLookDev()
+            => LookDev.LookDev.Open();
+
+        [MenuItemForRenderPipeline("Window/Analysis/Render Graph Viewer", false, 10006, typeof(HDRenderPipelineAsset))]
```
alex-vazquez, line 12: Priority should be the same,  right? 10001 vs 10006

alex-vazquez (changes requested):
Minor comments, I think we should use the Pipeline instead of the Asset though.

com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
RSlysz: This is only to support case where we want something only in builtin. Builtin is null (render pipeline asset used) but null cannot be used as a key

com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
RSlysz: Jenny seamed to say that the only thing we are certain at editor start is GraphicsSettings.currentRenderPipeline which is the Asset. Did I miss something @jenniferd-unity ? I would also prefer checking the RenderPipeline and not the asset if possible

com.unity.render-pipelines.core/Editor/RenderGraph/RenderGraphViewer.cs
RSlysz: Yes sorry I added missing name space here.
Only modification is that I remove the menu item for the RenderGraphWindow

com.unity.render-pipelines.high-definition/Editor/ToolMenus.cs
RSlysz: Priority for RenderGraph (Viewer) was 1006. I keep what was used.

RSlysz:
@alex-vazquez Did you do any test for other attributes somewhere? I think I perhaps need to add some for this one. (I dunno exactly how)

com.unity.render-pipelines.core/Editor/Utilities/MenuUtilities.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            // remove
+            foreach (var menuItemData in s_MenuMap.Values.SelectMany(mid => mid))
+                UnregisterMenu(menuItemData); //do nothing if not registered before
+
+            // add if required
+
+            // Note: GraphicsSettings.currentRenderPipeline == null possible and should mean built-in
+            // though we cannot use null as key so it was early replaced by typeof(RenderPipelineAsset)
+            Type currentSRPType = GraphicsSettings.currentRenderPipeline?.GetType() ?? typeof(RenderPipelineAsset);
```
jenniferd-unity, line 124: you can use `RenderPipelineManager.currentPipeline` if you want the RenderPipeline type (our APIs are very confusing I know).
The problem I can see here is that this can be called from a InitializeOnLoad where we cannot guarantee that the ADB is ready - which means we cannot guarantee accessibility to the SRP asset of the current Quality level.
So I guess your solution works in some circumstances but not all. We definitely need to push for a callback to guarantee those scenarios. (OnAssetDatabaseReady?)

jenniferd-unity (changes requested):
See my comments about InitializeOnLoad and not being able to know which SRP is going to be used for rendering.

alex-vazquez (changes requested):
Use the MenuManager :)

RSlysz:
Update: This PR is OnHold and will remain as we still don't have a good lifecycle for render pipeline nd this is required to land it. Removing remaining reviewer.

RSlysz:
Closing the PR for now

PR #5521, opened 2021-09-03
[Shadergraph] GatherTexture2D and TexelSize support all shader stages

BenGraterUnity:
---
### Purpose of this PR
GatherTexture2D and TexelSize nodes no longer restricted to just fragment and now supports all shader stages.

Example:
image

image
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-object-texture2d-gather
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/textureGather.xhtml
https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf

---
### Testing status
Tested changes locally in editor 2021.2.0b9 Windows
- Observed in Shader Graph that I am able to connect `GatherTexture2D` and `TexelSize` nodes to vertex stage (image above).
- Applied test shader to object in scene, observed a visible effect and was able to `Build and Run Windows 64-bit` without errors.


russoh:
Looks good to me.

russoh (approved):

joshua-davis (approved):
This seems fine according to the different specs, but no test would currently catch this (they only use texture gather in the fragment stage). It would be good to get a test added to SG for vertex at the least.

PR #5523, opened 2021-09-03
Fix how EventSystem component gets added to scene by DebugUpdater

arttu-peltonen:
### Purpose of this PR
This PR fixes issue reported in [case 1361901](https://fogbugz.unity3d.com/f/cases/1361901), as well as related converse case:
- Problem case 1: After entering playmode, if the EventSystem component disappears from scene (e.g. because a different scene was loaded), Runtime Debug UI mouse/touch input stops working.
- Problem case 2: After entering playmode, if an EventSystem component gets added (e.g. because a different scene was loaded, or UI was instantiated at runtime like happens in BoatAttack), log is spammed with a warning ("more than 1 EventSystem component in scene").

This PR solves the issue by changing when DebugUpdater adds its own EventSystem component. Before it was added always after initialization; **now it is always added/removed whenever Rendering Debugger Runtime UI is opened/closed**. This is okay because EventSystem (+ related input component) is only required when interacting with the runtime UI with mouse/touch. The design aims to minimize conflicts caused by potentially duplicate EventSystems - if debug UI is not opened, there is no chance of conflict. 

#### Known issue
The current logic doesn't support a situation where game logic code adds/removes the EventSystem component _while Rendering Debugger Runtime UI is open_. Even in this case, the situation can be fixed by closing and reopening the runtime UI.  This corner case cannot be solved properly without doing expensive calls like `FindObjectsOfType<EventSystem>()` in an Update method, because there is no API callback informing when EventSystems are added/removed, so I'm ignoring it for now. Ideally, none of this component peddling should be necessary, but until we switch to UITK this is is needed.

---
### Testing status
Tested switching back and forth between scenes using the test project & test script in [case 1361901](https://fogbugz.unity3d.com/f/cases/1361901) (both directions: from scene with no EventSystem to scene that has it, and in the other direction). Tested that keyboard + mouse input in debug UI works as expected in Windows Editor (both when using old and new input systems), and unexpected log spam is not produced.

Tested BoatAttack project (demo_island alone, and main_menu -> demo_island, using new input system) in Windows Editor, verified that mouse input works after scene transition from menu to game. No log spam produced when starting demo_island scene alone, mouse input works.

Tested Android build to ensure touch input still works, even after switching back and forth between scenes (both old & new input systems).

### DIFF
--- a/com.unity.render-pipelines.core/CHANGELOG.md
+++ b/com.unity.render-pipelines.core/CHANGELOG.md
@@ -83,6 +83,7 @@ The version number for this package has increased due to a version update of a r
 - Fixed Lens Flare 'radialScreenAttenuationCurve invisible'
 - Fixed Lens Flare rotation for Curve Distribution
 - Fixed potentially conflicting runtime Rendering Debugger UI command by adding an option to disable runtime UI altogether (1345783).
+- Fixed issues caused by automatically added EventSystem component, required to support Rendering Debugger Runtime UI input. (1361901)
 
 ### Changed
 - Improved the warning messages for Volumes and their Colliders.


com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -83,6 +83,7 @@ The version number for this package has increased due to a version update of a r
 - Fixed Lens Flare 'radialScreenAttenuationCurve invisible'
 - Fixed Lens Flare rotation for Curve Distribution
 - Fixed potentially conflicting runtime Rendering Debugger UI command by adding an option to disable runtime UI altogether (1345783).
+- Fixed issues caused by automatically added EventSystem component, required to support Rendering Debugger Runtime UI input. (1361901)
```
alex-vazquez, line 86:
```suggestion
- Fixed issues caused by automatically added EventSystem component, required to support Rendering Debugger Runtime UI input. (1361901).
```

com.unity.render-pipelines.core/Runtime/Debugging/DebugUpdater.cs
```diff
@@ -4,22 +4,27 @@
...
+        static DebugUpdater s_Instance = null;
+
         ScreenOrientation m_Orientation;
+        bool m_RuntimeUiWasVisibleLastFrame = false;
 
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.AfterSceneLoad)]
         static void RuntimeInit()
         {
-            if (!Debug.isDebugBuild || !DebugManager.instance.enableRuntimeUI || FindObjectOfType<DebugUpdater>() != null)
+            if (!Debug.isDebugBuild)
```
alex-vazquez, line 23: In another PR you have another system with some defines to check if it is a debug build, do we want to have this here like this?

com.unity.render-pipelines.core/Runtime/Debugging/DebugUpdater.cs
```diff
@@ -32,20 +37,90 @@ internal static void SetEnabled(bool enabled)
 
         static void EnableRuntime()
         {
+            if (s_Instance != null)
+                return;
+
             var go = new GameObject { name = "[Debug Updater]" };
-            var debugUpdater = go.AddComponent<DebugUpdater>();
+            s_Instance = go.AddComponent<DebugUpdater>();
```
alex-vazquez, line 44: It is weird to see the initialization of the SIngleton outside the `instance` property. Can we move it to that property?

alex-vazquez (approved):
Nice PR, Some really minor comments.

com.unity.render-pipelines.core/Runtime/Debugging/DebugUpdater.cs
arttu-peltonen: I think I'll just reconcile these changes into the other PR, I'll need to do a merge there anyway.

com.unity.render-pipelines.core/Runtime/Debugging/DebugUpdater.cs
arttu-peltonen: Hmm.  I don't think I could move it to the property, because it's a component so we need a gameobject to initialize it. This is not a traditional singleton in the sense that it has explicit lifetime (i.e. it will be created _and_ destroyed at will, depending on how `SetEnabled(bool)` has been called), so I think it's more clear to keep the create/destroy here. Moving that logic to the property might give the false impression that it's always available, which is not the case.

JulienIgnace-Unity (approved):

Verasl (approved):
LGTM

Tested on boat attack(worst case due to dynamic loading unloading of event systems) on both editor and iOS build, worked great.

erikabar (approved):
from the description this PR seems to be well covered, approving without additional testing

PR #5525, opened 2021-09-03
[URP][UI-UX] Removed the cookie foldout and moved into emission

martint-unity:
---
### Purpose of this PR
Removing the cookielight foldout in URP and move the cookie property into emission foldout group

image

---
### Testing status
Check that the cookie property still works and are now under the emission foldout in the inspector


ellioman (approved):

eh-unity (approved):
Seems good.

aleksandrasdzikia (approved):

PR #5527, opened 2021-09-03
ShadowCaster now using varyings normalWS for shader graph

Nzollner:
### Purpose of this PR
This PR is needed in order to use changed normals in the shadow caster if they are updated in the fragment shader. Otherwise will the normal bias break the shadows.

---
### Testing status
I have tested the fix with the project given in the bug and tested if the universal graphics tests worked locally.

---
### Comments to reviewers


pbbastian (approved):
LGTM

simon-engelbrecht-soerensen (approved):
LGTM

PR #5529, opened 2021-09-03
New UI for subdivision controller on probe volume

FrancescoC-unity:
This changes the UX as discussed in slack, should hopefully be a bit more clear.

We still have a fairly poor nomenclature around subdivision terms, I logged a task to look into it.

Video to show it: 

https://user-images.githubusercontent.com/43168857/132036455-971843f0-b2a4-4040-8a07-bca3ad817612.mp4


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
```diff
@@ -138,23 +138,47 @@ static void Drawer_VolumeContent(SerializedProbeVolume serialized, Editor owner)
...
+
+            EditorGUILayout.LabelField("Subdivision Overrides", EditorStyles.boldLabel);
+            EditorGUI.indentLevel++;
+            int value = serialized.highestSubdivisionLevelOverride.intValue;
+
+            // We were initialized, but we cannot know the highest subdiv statically, so we need to resort to this.
+            if (serialized.highestSubdivisionLevelOverride.intValue < 0)
+                serialized.highestSubdivisionLevelOverride.intValue = maxSubdiv;
+
+            serialized.highestSubdivisionLevelOverride.intValue = EditorGUILayout.IntSlider("Highest Subdivision Level", serialized.highestSubdivisionLevelOverride.intValue, 0, maxSubdiv);
```
alelievr, line 173: Missing `EditorGUI.BeginProperty` and EndProperty to correctly handle presets, prefabs, undo, etc...
(you can take the code before your PR as a reference to add it)

alelievr (commented):
LGTM except for the property thing in the UI. 

One thing to note though is that users will lose their valuer of the old subdivision controller but I guess it's fine since it's still experimental.

alelievr (approved):

FrancescoC-unity:
NOTE: There is an issue (we always had, before  too) with what we display. Looking at it.

FrancescoC-unity:
Fixed the info boxes, now we give proper info (before upper range in profile was wrong and the lower bound wrong on the probe volume itself)

JulienIgnace-Unity (approved):

PR #5530, opened 2021-09-03
Change something in the TAAU docs

FrancescoC-unity:
I was in vacation when changes were made so I didn't review them post @Vic-Cooper changes. 

A point was changed to something that might sound misleading, I replaced it with my original phrase, but please let me know if you prefer different wording :-)


sebastienlagarde (approved):
change is small enough. If there is any mistake Vic please correct in another PR. thanks

PR #5532, opened 2021-09-06
Bugfix 1359282: Emissive decal silhouette issue

JarkkoUnity:
### Purpose of this PR
Emissive projected decals suffer from artifacts in object silhouettes because UV discontinuities and how GPU uses the UV screen derivatives to calculate texture LOD. Fogbugz: https://fogbugz.unity3d.com/f/cases/1359282/

---
### Testing status
Added 1715 graphics test and got expected results on different platforms

---
### Comments to reviewers
Calculates texture LOD manually and clamps to given max based on depth screen derivatives. This is not perfect solution but reduces the issue. This solution also requires GetDimensions() to be available (#ifdeffed in shader) and otherwise samples from LOD 0 (which suffers from projected emissive texture aliasing but at least doesn't have the artifact). Only platform  that doesn't support GetDimensions() is Metal, which graphics test shows more texture aliasing artifacts on the floor. There also seems to be some issue with depth derivatives that required clamping the device depth to a small but larger-than-0 value in shader, which was showing artifacts in sphere silhouette against sky background. The graphics test camera is oriented so that shows the issue in the sphere silhouette if this depth clamp isn't present.


com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalData.hlsl
```diff
@@ -31,7 +31,17 @@ void GetSurfaceData(FragInputs input, float3 V, PositionInputs posInput, float a
 #ifdef _MATERIAL_AFFECTS_EMISSION
     surfaceData.emissive = _EmissiveColor.rgb * fadeFactor;
     #ifdef _EMISSIVEMAP
-    surfaceData.emissive *= SAMPLE_TEXTURE2D(_EmissiveColorMap, sampler_EmissiveColorMap, texCoords).rgb;
+    float2 emissiveColorMapSize;
+    float emissiveColorMapLODs;
+    _EmissiveColorMap.GetDimensions(0, emissiveColorMapSize.x, emissiveColorMapSize.y, emissiveColorMapLODs);
```
sebastienlagarde, line 36: Hey, so GetDimensions isn't correctly supported on various platform. To retrieve the number of LOD you should use our function uint GetMipCount(TEXTURE2D_PARAM(tex, smp)) declare in common.hlsl. It isn't implemented for all platform but return 0 at least. Which in your context is not ideal but should be ok (in particular for Metal). Otherwise you need to do it a different way without trying to access to current number of LOD or by "fixing" it.

However for the size of the texture itself I have no idea if we have something cross platforms. From the comment of the function GetMipCount, it is possible that retrieving the size works. If we want to validate all this, best is to create a test scene with the artifact and add it to our project suite, then run it on Yamato to see what is happening on author platforms.

JarkkoUnity: There's also an issue I haven't yet figured out, where the background that's not rendered to G-buffer has some emissive dots in the silhouette. It's less of an issue than before, but it's  something I need to investigate

sebastienlagarde:
Yamato is good (error from master). Ready for test

iM0ve (approved):
I have verified the fix. It doesnt cure the issue 100% but its a lot better. See the results below:

Before the PR:

wBc7s3bZy9

After the PR:

yoVbUh7gdY

To completely get rid of the outline there are a few options:
* Enable Always Refresh + Temporal anti-aliasing
* Have geometry behind the object (not sky as background)

https://user-images.githubusercontent.com/36502659/134011703-8771242f-2d7d-4d86-981b-9a69761da53d.mp4

PR #5536, opened 2021-09-06
Improved the performance of the volumetric clouds in non-local mode when large occluders are on screen.

anisunity:
When in non local mode, we used to to render volumetric clouds  fo all pixels even if we had a wall in front of the camera covering the whole screen (which is a waste of performance). This PR uses the maxZMask texture to reduce in a conservative way to number of pixels that trace cloud rays.
image

Testing status
Tested that no new artifacts appeared when moving fast in non local mode at the edges of geometry.


com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricCloudsSky.cs
```diff
@@ -274,7 +276,7 @@ internal void RenderVolumetricClouds_Sky(CommandBuffer cmd, HDCamera hdCamera, M
                         parameters.commonData.cloudsCB._CloudsPixelCoordToViewDirWS = pixelCoordToViewDir[faceIdx];
 
                         // Render the face straight to the output cubemap
-                        RenderVolumetricClouds_Sky_High(cmd, parameters, m_MpbClouds, m_IntermediateCloudsLighting0Buffer, m_IntermediateCloudsLighting1Buffer, m_IntermediateCloudsDepthBuffer, skyboxCubemap);
+                        RenderVolumetricClouds_Sky_High(cmd, parameters, m_MpbClouds, m_IntermediateCloudsLighting0Buffer, m_IntermediateCloudsLighting1Buffer, m_IntermediateCloudsDepthBuffer, skyboxCubemap, TextureXR.GetBlackTexture());
```
FrancescoC-unity, line 279: Passing black here, won't make the check in here https://github.com/Unity-Technologies/Graphics/pull/5536/files#diff-897ad540cb7032a41aaf5e63dce2b5256798820e918184f5ac4db09ac32c70f9R847 always early out? Is that ok?

anisunity: You are right, doesn't work (the tests also failed). Not sure how to pas a 1e10 texture tho haha.

anisunity:
Review fixes done

FrancescoC-unity (approved):

remi-chapelain (approved):
Tested with multiple type of occluders to see if there wasn't any regression artefacts. 
(Opaque / Transparent / Opaque alpha clipped / Double sided) ✔️ 

Tested performance improvement on a very simple scene with camera pointed up and an occluder covering half de screen in 4K. 
Performance **gain** on the GPU side was about 40% and around 8% on the frame time overall compared to before the PR were all the clouds were still evaluated behind the occluder. 

image

sebastienlagarde (approved):

PR #5537, opened 2021-09-06
[QualityAudit] [SRP Workflow] Upgrade volume priority float to int [OnHold]

RSlysz:
### Purpose of this PR
Priority is int everywhere but in Volume. This PR fix this ( https://jira.unity3d.com/browse/HDRP-908 )
There is no good conversion function that works all the time. So we provided a script for the user to edit all Volume of its project with a new conversion function if there is any collision/out of capacity.

---
### Testing status
Manually tested the upgrade.
Manually tested the script provided for custom conversion. (Volume in scene, Prefab Instance in scene, Prefab asset, Prefab Variant, Embedded prefab asset)

---

### DIFF
--- a/com.unity.render-pipelines.core/Documentation~/whats-new-13.md
+++ b/com.unity.render-pipelines.core/Documentation~/whats-new-13.md
@@ -1,0 +1,151 @@
+# What's new in SRP Core version 13 / Unity 2022.1
+
+This page contains an overview of new features, improvements, and issues resolved in version 13 of the Core Render Pipeline package, embedded in Unity 2022.1.
+
+## Improvements
+
+### Volume
+
+The Volume framework have evolve a little to be more similar in the way it handle priorities to the remaining of the editor. Now it will be int based. The transition is done automatically in an internal upgrade. Though changing from float to int is not a trivial task and some project can have priority that collide.
+
+If you encounter any colliding priority, you may have Volume that didn't apply change anymore. So you should resolve any conflict. The upgrade just take former priority times 1000 to keep first 3 digits of the float as relevant for the new int priority. It is fine most of the time.
+
+But in some case you want to apply a different float to int adaptation. Here is a script that allow you to pass on every Volume of your project and apply your own re-range.
+
+```
+using System;
+using System.Linq;
+using UnityEditor;
+
+namespace UnityEngine.Rendering
+{
+    class VolumePriorityUpgrader : Editor
+    {
+        [MenuItem("Project/Upgrade Project")]
+        static void UpgradeProject()
+        {
+            // Load all prefabs
+            string[] guids = AssetDatabase.FindAssets("glob:\"*.prefab\"");
+            for (int i = 0; i < guids.Length; ++i)
+            {
+                string assetPath = AssetDatabase.GUIDToAssetPath(guids[i]);
+                GameObject gameObject = AssetDatabase.LoadAssetAtPath<GameObject>(assetPath);
+                if (RecursiveApplyOnVolumeInHierarchy(gameObject.transform, Upgrade))
+                {
+                    EditorUtility.SetDirty(gameObject);
+                    AssetDatabase.SaveAssetIfDirty(gameObject);
+                }
+            }
+
+            // Load all scenes
+            guids = AssetDatabase.FindAssets("glob:\"*.unity\"");
+            for (int i = 0; i < guids.Length; ++i)
+            {
+                string assetPath = AssetDatabase.GUIDToAssetPath(guids[i]);
+
+                //prevent scene modification in readonly packages
+                UnityEditor.PackageManager.PackageInfo packageInfo = UnityEditor.PackageManager.PackageInfo.FindForAssetPath(assetPath);
+                if (!(packageInfo == null
+                    || packageInfo.source == UnityEditor.PackageManager.PackageSource.Local
+                    || packageInfo.source == UnityEditor.PackageManager.PackageSource.Embedded))
+                    return;
+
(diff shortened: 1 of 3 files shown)


alex-vazquez (approved):
Looks good. Is cool that you have implemented the migration for volumes. Thanks

alex-vazquez (changes requested):
Use the  IVersionable and MigrationSteps system.

JulienIgnace-Unity (approved):

com.unity.render-pipelines.core/Runtime/Volume/Volume.cs
```diff
@@ -9,26 +9,39 @@ namespace UnityEngine.Rendering
...
-    public class Volume : MonoBehaviour, IVolume
+    public partial class Volume : MonoBehaviour, IVolume
     {
         [SerializeField, FormerlySerializedAs("isGlobal")]
         private bool m_IsGlobal = true;
 
         /// <summary>
         /// Specifies whether to apply the Volume to the entire Scene or not.
         /// </summary>
-        [Tooltip("When enabled, the Volume is applied to the entire Scene.")]
```
jenniferd-unity, line 20: why is that tooltip removed?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LookDev.cs
```diff
@@ -109,7 +109,7 @@ void IDataProvider.FirstInitScene(StageRuntimeInterface SRI)
             volumeGO.name = "StageVolume";
             Volume volume = volumeGO.AddComponent<Volume>();
             volume.isGlobal = true;
-            volume.priority = float.MaxValue;
```
jenniferd-unity, line 112: i am not sure about the intrication of this since float.maxValue != int.MaxValue.
since this is only for initialization I think it has low impact but could you confirm?

jenniferd-unity (changes requested):
Good PR in general. Just a couple of questions for clarifying purposes.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LookDev.cs
RSlysz: We just want to have a top priority volume to override any kind of volume when using the LookDev. This one only apply to the "scene" of the LookDev. And the volume contains (normally) light and shadowing information.

com.unity.render-pipelines.core/Runtime/Volume/Volume.cs
RSlysz: A tooltip on a property have no usage AFAIK. So it is dead code at the moment.
And in the Inspector the Volume have the Mode field to set it as global or not and it use a different tooltip than this one.

jenniferd-unity: ok thanks!

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LookDev.cs
jenniferd-unity: sounds good!

jenniferd-unity (approved):

com.unity.render-pipelines.core/Documentation~/whats-new-13.md
```diff
@@ -0,0 +1,151 @@
...
+## Improvements
+
+### Volume
+
+The Volume framework have evolve a little to be more similar in the way it handle priorities to the remaining of the editor. Now it will be int based. The transition is done automatically in an internal upgrade. Though changing from float to int is not a trivial task and some project can have priority that collide.
+
+If you encounter any colliding priority, you may have Volume that didn't apply change anymore. So you should resolve any conflict. The upgrade just take former priority times 1000 to keep first 3 digits of the float as relevant for the new int priority. It is fine most of the time.
+
+But in some case you want to apply a different float to int adaptation. Here is a script that allow you to pass on every Volume of your project and apply your own re-range.
+
```
martint-unity, line 14: @oleks-k please help with the wording here.

martint-unity (approved):
Looks OK. Need a Docs person to review the wall of text :) So I am adding @oleks-k 
Will also add URP QA if not already added.

RSlysz:
> 
> 
> Use the IVersionable and MigrationSteps system.

This is not easilly doable right now. It is planed here: https://jira.unity3d.com/browse/XPIPELINE-326
No ETA at the moment.

Meanwhile, we can land it as is and adapt later to unify migration framework.

alex-vazquez (approved):

JMargevics:
Hey, found no regressions using this PR on an empty project. However upgrading existing template results in `Library\PackageCache\com.unity.cinemachine@2.8.0\Runtime\PostProcessing\CinemachineVolumeSettings.cs(266,34): error CS0266: Cannot implicitly convert type 'float' to 'int'. An explicit conversion exists (are you missing a cast?)` 
Would it be something for Cinemachine team to fix?

RSlysz:
> Would it be something for Cinemachine team to fix?

Erf no. I tested a hdrp project migration to a newer version. But I didn't expected it to broke on other packages. I need to counter check this.

RSlysz:
Checking with UX if this should land at all.
Reason: 
- Many package impacted
- Cannot automigrate in package (readonly)
- User can have difficulties to find collision, especially as they can be not owning the prefab or the code that create the volume.

RSlysz:
After discussing it with UX, we finally drop this alignment for now as it is too much work and add more problems than it solve at the moment. We will try to find another way later or at least bulk it with other breaking change at major version change of Unity.

PR #5546, opened 2021-09-07
Fixed regression in the ambient probe intensity for volumetric clouds.

anisunity:
When adding the code that changes which ambient probe is used. I removed pieces of the code that i thought were optional (spoiler alert, they are not). This adds them back and restores a look that is closer to what we expect.
image

@pierre-unity  is going to update the LUT so I have to wait before updating the screenshots

Testing status
Tested on the samples, looks good now.
This will break a bunch of clouds tests


anisunity:
This PR now also includes:
- An update of the presets by @pierre-unity
- A fix for the sun leaking behind the clouds
- A fix for an artifact appearing when in non local mode and lowest cloud altitude very low.

pierre-unity (approved):
tested locally

anisunity:
Merged https://github.com/Unity-Technologies/Graphics/pull/5556 into this PR

anisunity:
In this PR i've added a fix to this bug https://fogbugz.unity3d.com/f/cases/1364152/

TomasKiniulis (approved):
Checked with Anis on listed concerns, looks all good assuming CI is green.

anisunity:
Updated all screenshots and relaunched yamato

PR #5551, opened 2021-09-07
[HDRP][Path Tracing] SSS transmissive mode now triggered by double-sidedness + transmission

eturquin:
Changes how thin mode for SSS is enabled in Path Tracing:

A recent change made it so that it is turned on when the "transmission" sss option is set on the material.
However, activating this option when not needed is to be avoided, as half the samples will be lost (on closed objects, they stay "inside" and will have a null contribution)... and since the "transmission" option is on by default on the Lit familty of materials, this would change the results of previously set up scenes, or by default on new scenes even if this behaviour is unwanted in most scenarios.

What is more, the transmissive SSS mode only makes full sense for thin, double-sided surfaces (leafs, sheet of paper or fabric, etc.). We could make it so that it requires both "double-sided" and "transmission" on, but then these options become redundant, and there is no practical case in path tracing where you'd want one without the other. For that reason, it's simpler to make this thin surface behaviour triggered by double-sidedness only, and "transmission" does not have any effect (as was the case originally).

Documentation has been added on the path tracing page to cover this behaviour.

EDIT: after further deliberations, it has been decided to keep dependency on the transmission toggle too: while in most situations it is redundant, there are some corner cases for which it can be useful to rely on both sidedness and this extra toggle.


remi-chapelain (approved):
While I don't really like the fact that path tracer doesn't match again with what happens in raster (because transmission keywords is now basically ignored with path tracer), there's no perfect solution and even requiring double sided and transmission will create discrepancies. 

Kudos for adding documentation to explain those behaviors. 🥳  

To better comprehend the fixed problem, I added a graphic test in the DXR project with some bit of explanation for the future. (because it's far from intuitive). 

**Not the actual reference image**
_composite of raster / multi sample path tracer to better see what's tested_
image

I tested that the test fail if double sided was checked in SSS Single material and if double sided unchecked in SSS Double ✔️ 
[Launched yamato](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Fpt_sss_thin/.yamato%252Fall-hdrp_dxr.yml%2523HDRP_DXR_trunk/8610701/job).

remi-chapelain (approved):
- Updated test for fabric that [this previous PR](https://github.com/Unity-Technologies/Graphics/pull/5485) broke
- Updated new added test to be more robust for regressions

image

All PT test are green locally. 

[Relaunched yamato](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Fpt_sss_thin/.yamato%252Fall-hdrp_dxr.yml%2523HDRP_DXR_trunk/8630319/job). All green 🟢

sebastienlagarde:
Todo: merge master and re-target to master

pmavridis (approved):

sebastienlagarde (approved):

sebastienlagarde:
@Vic-Cooper I merged this PR but please if you can track it and review the doc. THanks

PR #5552, opened 2021-09-07
Updated template due to URP RendererData changes

Verasl:
### Purpose of this PR
Template update for the fixes in this PR:https://github.com/Unity-Technologies/Graphics/pull/5410 it also resolves [this bug](https://fogbugz.unity3d.com/f/cases/1357377/)
Some additional files had general updated to serialisation from current URP, I have added these also as they are low risk and will be experienced by the user regardless when opening the template.

---
### Testing status
Tested manually by QA

---
### Comments to reviewers


com.unity.template-universal/ProjectSettings/QualitySettings.asset
```diff
@@ -93,7 +93,7 @@ QualitySettings:
     skinWeights: 255
     textureQuality: 0
     anisotropicTextures: 1
-    antiAliasing: 0
+    antiAliasing: 2
```
phi-lira, line 96: Is this change intentional?

phi-lira (approved):

com.unity.template-universal/ProjectSettings/QualitySettings.asset
Verasl: my guess is yes, I didn't change any AA settings but my assumption here is that the Template hasn't had this file updated since URP started to set this depending on it's state, we still haven't separated the logic here right?

PR #5555, opened 2021-09-07
SRP Workflows - New Additional Data Interface to be able to identify additional data's in Core Package

alex-vazquez:
---
### Purpose of this PR
From generic code on Core is imposible to know where an Additional Data component is from URP or HDRP.

There are some PR's comming and improving some x-pipeline workflows, so this PR is a shared between them.

---
### Testing status
Test the compilation.

---
### Comments to reviewers
Some other PR will use this interface.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Common/IAdditionalData.cs
+++ b/com.unity.render-pipelines.core/Runtime/Common/IAdditionalData.cs
@@ -1,6 +1,9 @@
 
 namespace UnityEngine.Rendering
 {
+    /// <summary>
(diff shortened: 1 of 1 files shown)


RSlysz (approved):
This will help us find all additional data.

When refactor of them will land, User can still want to have additional data for component we don't think of. So adding it will still be needed. But perhaps we should need to move it to trunk with the refactor to add it on all extensions.

martint-unity (approved):
LGTM

RSlysz:
Github bot test @arttu-peltonen

PR #5558, opened 2021-09-07
[HDRP] APV Baking and Settings window

alelievr:
### Purpose of this PR
Add a new window to manage settings for APV. The new window is located in **Window > Rendering > Probe Volume Settings (Experimental)**

image

---
### Testing status
- Tested undo/redo on all the UI
- Tested the bake / load scene / clear buttons
- Multi-scene baking
- Tested that parameters in the UI actually works

### DIFF
--- a/.yamato/_latest_editor_versions_trunk.metafile
+++ b/.yamato/_latest_editor_versions_trunk.metafile
@@ -1,5 +1,5 @@
 trunk:
-  updated_at_UTC: '2021-09-09 23:29:33.229714'
+  updated_at_UTC: '2021-09-10 23:29:14.667211'
   changeset:
-    id: 2da9b43a3f3fbe7adef2aeb55d07239f96b5814c
-    date: '2021-09-09T03:55:47'
+    id: 5a1ba9fbc92df5c92bab6a470d4ff080834af3bd
+    date: '2021-09-10T05:09:36'
--- a/.yamato/script/log_scripts/log_parser.py
+++ b/.yamato/script/log_scripts/log_parser.py
@@ -6,9 +6,9 @@ import sys
 import json
 import traceback
 import re
-from utils.execution_log import Execution_log
-from utils.utr_log import UTR_log
-from utils.unity_log import Unity_log
+from patterns.execution_log import Execution_log
+from patterns.utr_log import UTR_log
+from patterns.unity_log import Unity_log
 from utils.shared_utils import *
 from utils.constants import *
 
@@ -29,9 +29,7 @@ To run it locally, specify
     --execution-log "<path to execution log file>"
 '''
 
-
-
-def parse_failures(execution_log, logs, local):
+def parse_failures(execution_log, logs, args):
     '''Parses each command in the execution log (and possibly UTR logs),
     recognizes any known errors, and posts additional data to Yamato.'''
     for cmd_key in logs.keys():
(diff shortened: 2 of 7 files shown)


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -320,6 +320,9 @@ internal static void PerformDilation()
 
             if (dilationSettings.dilationDistance > 0.0f)
             {
+                // Make sure all pending unloads are done before we reset everything through ForceSHBands.
```
JulienIgnace-Unity, line 323: @FrancescoC-unity can you validate this fix?
Basically some scenes are still in the pending unload list, we then reset the ref volume to a different size which makes subsequent unloading fail (out of bounds for flat array access).

FrancescoC-unity: Uhm this is potentially dangerous I think, as we are flushing also the loading that are pending while we want to flush them only after we force the sh bands. 

We might need something that only flushes the unloading.

JulienIgnace-Unity: Right, that makes sense. Shouldn't we be flushing everything at the start of baking anyway?
There's probably a better place for this I guess.

FrancescoC-unity: Yes we can try to do that, we can try at the start of OnAdditionalProbesBakeCompleted , worth testing this a bit as I wonder if there are unwanted side effects we are not thinking about.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeBakingWindow.cs
```diff
@@ -0,0 +1,525 @@
...
+            m_ScenesInSet = new ReorderableList(set.sceneGUIDs, typeof(string), true, true, true, true);
+            m_ScenesInSet.drawHeaderCallback = (rect) => EditorGUI.LabelField(rect, "Scenes", EditorStyles.largeLabel);
+            m_ScenesInSet.multiSelect = true;
+            m_ScenesInSet.drawElementCallback = (rect, index, active, focused) =>
+            {
+                var guid = set.sceneGUIDs[index];
+                // Find scene name from GUID:
+                var scene = FindSceneData(guid);
+
+                // TODO: Add a label on the first scene to say the the lighting settings of this scene will be used for baking
```
FrancescoC-unity, line 225: We do have it now, so can get rid of this TODO.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeBakingWindow.cs
```diff
@@ -0,0 +1,525 @@
...
+                    path = path,
+                    guid = s
+                };
+            }).ToList();
+        }
+
+        SceneData FindSceneData(string guid)
+        {
+            // TODO: replace m_ScenesInProject list by a dictionary
+            var data = m_ScenesInProject.FirstOrDefault(s => s.guid == guid);
```
FrancescoC-unity, line 199: Do we want to this TODO before getting this through?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeBakingWindow.cs
```diff
@@ -0,0 +1,525 @@
...
+            };
+
+            m_BakingSets.onRemoveCallback = (list) =>
+            {
+                if (m_BakingSets.count == 1)
+                {
+                    EditorUtility.DisplayDialog("Can't delete baking set", "You can't delete the last Baking set. You need to have at least one.", "Ok");
+                    return;
+                }
+                if (EditorUtility.DisplayDialog("Delete selected baking set?", $"Do you really want to delete the baking set '{sceneData.bakingSets[list.index].name}'?", "Yes", "Cancel"))
```
FrancescoC-unity, line 158: Delete _the_ selected baking set

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeBakingWindow.cs
alelievr: I guess it's fine

PR #5561, opened 2021-09-08
SRP Workflows - Fix invalid cast exception

alex-vazquez:
---
### Purpose of this PR
https://jira.unity3d.com/browse/XPIPELINE-275

---
### Testing status
Check the invalid cast is not longer happening

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
@@ -633,15 +633,11 @@ namespace UnityEngine.Rendering.HighDefinition
 
         private void QueueSHBaking()
         {
-            var hdrp = (HDRenderPipeline)RenderPipelineManager.currentPipeline;
-            if (hdrp != null)
-            {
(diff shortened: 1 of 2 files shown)


TomasKiniulis (approved):
Thanks for the fix! Works like a charm

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -633,15 +633,11 @@ private Vector3 ComputeCapturePositionWS()
...
         private void QueueSHBaking()
         {
-            var hdrp = (HDRenderPipeline)RenderPipelineManager.currentPipeline;
-            if (hdrp != null)
-            {
-                if (!hdrp.currentPlatformRenderPipelineSettings.supportProbeVolume) return;
-            }
-            else
-            {
+            if (!(RenderPipelineManager.currentPipeline is HDRenderPipeline hdrp))
```
jenniferd-unity, line 636: Maybe we can simplify with `HDRenderPipeline.asset != null` and/or `HDRenderPipeline.currentPlatformRenderPipelineSettings != null` ?

jenniferd-unity (approved):
I put suggestions but otherwise it is good to go for me :)

PR #5562, opened 2021-09-08
VFX now properly receive APV

FrancescoC-unity:
Because of missing pragmas for the PROBE_VOLUME the VFX were always falling back on ambient probe. 

**Before**
Unity_xFTxZzwy1Q

**After**
Unity_RPTyl6EJr5

All is good now


FrancescoC-unity:
Side note for later: particles should probably sample this per vertex and not per pixel :) But I guess it is an optimization for later on.

sebastienlagarde (approved):

JulienIgnace-Unity (approved):

alelievr (approved):

PaulDemeulenaere (approved):
Only built in lit output should be affected, maybe we should check the new shaderGraph integration (but multicompile are already handled by SG generation in that case)
LGTM

PR #5564, opened 2021-09-08
Prevent error when saving a scene with no profile being set anywhere

FrancescoC-unity:
@iM0ve this is a fix for the error you sent me. 

There could be a situation in which we don't have any profile set for a given scene and the update scene bounds do require a profile to have been set at some point. 

The warning message is intentionally vague now w.r.t where to find the settings as they are soon to be moved.


JulienIgnace-Unity (approved):

alelievr (approved):

iM0ve (approved):
Thanks

PR #5565, opened 2021-09-08
SSR clear coat improvement

anisunity:
https://fogbugz.unity3d.com/f/cases/1352424/
A double contribution of the clear cloat was found when SSR/RTR was enabled on the Lit and Stack Lit shader. This fixes that behavior.

Testing status.
The solution was validated by the user.
Multiple tests will need screenshot updates


TomasKiniulis (approved):
Looks good assuming updated reference images pass, since it's already validated by user.

anisunity:
Fixed formatting and launching yamato

patrickp-unity3d:
Should this fix also be applied to AxF?

com.unity.render-pipelines.high-definition/Runtime/Material/StackLit/StackLit.hlsl
```diff
@@ -4351,6 +4367,10 @@ IndirectLighting EvaluateBSDF_Env(  LightLoopContext lightLoopContext,
         if( (i == (0 IF_FEATURE_COAT(+1))) && _DebugEnvLobeMask.y == 0.0) continue;
         if( (i == (1 IF_FEATURE_COAT(+1))) && _DebugEnvLobeMask.z == 0.0) continue;
 #endif
+
+        // If we are going to process the clear coat and it is flagged as akready processed, skip.
+        if (i == COAT_LOBE_IDX && COAT_NB_LOBES == 1 && preLightData.clearCoatIndirectSpec == 0.0) continue;
```
anisunity, line 4372: Remove this

sebastienlagarde:
Added Stephane for awareness

anisunity:
An other solution to avoid the inout would be to add a parameter, but that would mean breaking the shader api... so

sebastienlagarde:
> An other solution to avoid the inout would be to add a parameter, but that would mean breaking the shader api... so

yeah, was afraid that inout may currently braek the API but it doesn, we will go with it. Don't like it too much but I haven't better alternative to propose without breaking the API so let's use that

sebastienlagarde (approved):

sebastienlagarde:
There is various failure with Stacklit on test apparently

anisunity:
Checking that

patrickp-unity3d:
Looking at AxF.hlsl, it really looks like the fix should be done there too

anisunity:
Fixed stack lit, relaunching the tests

sebastienlagarde:
there is still one test failing: 1302_StackLitSG_PixarLM.png

anisunity:
After looking at the screenshots, the new screenshots make more sense than the previous ones. Updating them.

PR #5567, opened 2021-09-08
[HDRP][Path Tracing] Directional light color now affected by physical sky

eturquin:
This addresses: https://fogbugz.unity3d.com/f/cases/1363584/

Before:
In path tracing, when sky model was set to physical, directional light emission color wasn't affected by its direction

Now:
Emission color is affected by direction (warmer colors) at grazing angles, then zero under horizon (as is the case in rasterization).


sebastienlagarde:
Please trigger Yamato.
Note to self: redirect PR to master

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingLight.hlsl
```diff
@@ -262,15 +262,37 @@ float3 GetPunctualEmission(LightData lightData, float3 outgoingDir, float dist)
...
 
-float3 GetDirectionalEmission(DirectionalLightData lightData, float3 outgoingVec)
+float3 GetDirectionalEmission(DirectionalLightData lightData, float3 position)
 {
     float3 emission = lightData.color;
 
+#if SHADEROPTIONS_PRECOMPUTED_ATMOSPHERIC_ATTENUATION
+    // Nothing to do here
+#else
+    // Physical sky emission color code, adapted from EvaluateLight_Directional()
```
sebastienlagarde, line 272: Sanity check, is this code call for any kind of sky? what's happen with an HDRI sky?

eturquin: It was the same in the original function, I think it comes from the dynamic if() condition right after this (or at least that's what I saw when checking its effect: if the "affect sky" on the light is off, or if there is no physical sky, then the if condition is false).

eturquin:
> Please trigger Yamato.

DXR tests broken atm, we should merge https://github.com/Unity-Technologies/Graphics/pull/5551 before this one, so that we get clean tests again. It is likely that this new PR will break one test too, will have to be updated after this other merge.

iM0ve (approved):
**What's tested:**
* Repro case with physical sky
* Verified that using HDRI the sun angle doesnt impact color
* As bonus, did a smoke test for all Physical sky properties with Pathtracer

**Showcase** <3

https://user-images.githubusercontent.com/36502659/133593504-28f0186b-7d12-45a0-8b3f-542788fd8a9d.mp4

PR #5577, opened 2021-09-08
Change the way we handle emissive with RTGI/SSGI/Mixed and APV + Remove ForceForwardEmissive code

sebastienlagarde:
### Purpose of this PR
This PR is a big cleanup PR and also change the way we handle emissive to works correctly with RTGI/SSGI/Mixed and APV.

This PR:
- Remove all the code introduce for ForceForwardEmissive including documentation, Receive SSGI, stripping code etc...
- Cleanup remaining code from ProbeVolume V1
- Cleanup stencil related code of Receive SSGI in SSGI and RTGI code
- Uniformize the code use for RTGI/SSGI/Mixed and APV regarding lightloop, AO and SampleGI
- Try to simplify code path, and optimize where possible
- Added an alias EvaluateLightProbe for the function EvaluateAmbientProbe for more clarity

Explanation about how APV and SSGI/RTGI/Mixed effects will handle emissve with this PR and the steps.

All effects will output only Emissive inside the lighting buffer (gbuffer3) in case of deferred (For APV this is done only if we are not a lightmap). The Lightmaps/Lightprobes contribution is 0 for those cases. Code enforce it in SampleBakedGI(). The remaining code of Material pass (and the debug code) is exactly the same with or without effects on, including the EncodeToGbuffer.  builtinData.isLightmap is used by APV to know if we have lightmap or not and is harcoded based on preprocessor in InitBuiltinData()

 AO is also store with a hack in Gbuffer3 if possible. Otherwise it is set to 1.
In case of regular deferred path (when effects aren't enable) AO is already apply on lightmap and emissive is added on to of it. All is inside bakeDiffuseLighting and emissiveColor is 0. AO must be 1
 
When effects are enabled and for APV we don't have lightmaps, bakeDiffuseLighting is 0 and emissiveColor contain emissive (Either in deferred or forward) and AO should be the real AO value.
Then in the lightloop in below code we will evalaute APV or read the indirectDiffuseTexture to fill bakeDiffuseLighting.
We will then just do all the regular step we do with bakeDiffuseLighting in PostInitBuiltinData()
No code change is required to handle AO, it is the same for all path.

Note: With current approach, Decals Emissive aren't taken into account by RTGI and by the Mixed method.
Note2: With current approach, Transparent Emissive works with SSGI and RTGI but can cause artifact with Mixed (RayMarch part could get a hit and thus avoid Raytrace path)
Note3: Forward opaque emissive work in all cases. The current code flow with Emissive store in GBuffer3 is only to manage the case of Opaque Lit Material with Emissive in case of deferred
Note4: Only APV can handle backFace lighting, all other effects are front face only.
        
TODO: complete documentation

---
### Testing status

TODO (no test done yet) Test Plan:

Scenario are: SSGI, RTGI, Mixed, APV, SSGI with APV with forward and deferred case

With scene that have emissive decal, emissive transparent unlit,  emissive opaque unlit lit opaque emissive.

Test all debug lighting mode: lighting only, emissive only, override emissive for all scenario
Test with probe volume on/off when APV is enable
Test indirect diffuse multiplier for all scenario
Test that AO behave correctly in all scenario

---
### Comments to reviewers

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/2505_Area_Light_ShadowMask_Baking.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/2505_Area_Light_ShadowMask_Baking.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:296d3596c3504a3be38fdb0459df9aa0415aa4279cc6ab8eabd79b1f2aead821
-size 136986
+oid sha256:78988b202e4aaf08d27ca213efa45bc2621d7485a53d43ef9ad5c0f50c2549d7
+size 137217


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -391,73 +391,91 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+                        R,
+                        V,
+                        posInput.positionSS,
+                        tempBuiltinData.bakeDiffuseLighting,
+                        tempBuiltinData.backBakeDiffuseLighting,
+                        lightInReflDir);
+                }
+                else // If probe volume is disabled we fallback on the ambient probes
+                {
+                    tempBuiltinData.bakeDiffuseLighting = EvaluateAmbientProbe(bsdfData.normalWS);
```
sebastienlagarde, line 457: @FrancescoC-unity I change the code of APV which handle the case of disable the APV. Now nothing is done in Material Pass and all is done in Lightloop. This was a requirement to be compliant with the new system but it is also cleaner.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceGlobalIllumination.compute
```diff
@@ -264,16 +264,14 @@ void REPROJECT_GLOBAL_ILLUMINATION(uint3 dispatchThreadId : SV_DispatchThreadID,
     {
         BuiltinData apvBuiltinData;
         ZERO_INITIALIZE(BuiltinData, apvBuiltinData);
-        SetAsUninitializedGI(apvBuiltinData.bakeDiffuseLighting);
```
sebastienlagarde, line 267: Guess this code was useless

com.unity.render-pipelines.high-definition/Runtime/Material/Builtin/BuiltinData.cs
```diff
@@ -58,6 +58,9 @@ public struct BuiltinData
             public float distortionBlur;           // Define the color buffer mipmap level to use
 
             // Misc
+            [SurfaceDataAttributes("Is Lightmap")]
+            public uint isLightmap; // Currently only use with APV
```
sebastienlagarde, line 62: @FrancescoC-unity new attribute used to identify if we are a lightmap or not for APV. Encoded into the high bit of specular occlusion when APV is enabled.

com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinGIUtilities.hlsl
```diff
@@ -19,40 +19,15 @@ real3 EvaluateAmbientProbe(real3 normalWS)
...
 }
 
+// Alias to make code less confusing as the same code is used to evaluate Ambient Probe and also to evaluate builtin light probe (which include ambient probe + local lights)
+#define EvaluateLightProbe EvaluateAmbientProbe
+
 #if defined(PROBE_VOLUMES_L1) || defined(PROBE_VOLUMES_L2)
 #include "Packages/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl"
-
-// y channel is reserved for potential payload information to carry alongside the unintialized flag.
-#define UNINITIALIZED_GI float3((1 << 11), 1, (1 << 10))
```
sebastienlagarde, line 26: All this code is no longer needed, rely on isLightmap now

com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinGIUtilities.hlsl
```diff
@@ -123,11 +98,12 @@ void EvaluateLightProbeBuiltin(float3 positionRWS, float3 normalWS, float3 backN
 {
     if (unity_ProbeVolumeParams.x == 0.0)
     {
-        bakeDiffuseLighting += EvaluateAmbientProbe(normalWS);
-        backBakeDiffuseLighting += EvaluateAmbientProbe(backNormalWS);
+        bakeDiffuseLighting += EvaluateLightProbe(normalWS);
```
sebastienlagarde, line 101: Just reuse an atlas name for this function for clarity

com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinUtilities.hlsl
```diff
@@ -49,12 +49,19 @@ void InitBuiltinData(PositionInputs posInput, float alpha, float3 normalWS, floa
...
-    // Sample lightmap/probevolume/lightprobe/volume proxy
+    // Sample lightmap/lightprobe/volume proxy
     builtinData.bakeDiffuseLighting = 0.0;
     builtinData.backBakeDiffuseLighting = 0.0;
     SampleBakedGI(  posInput, normalWS, backNormalWS, builtinData.renderingLayers, texCoord1.xy, texCoord2.xy,
                     builtinData.bakeDiffuseLighting, builtinData.backBakeDiffuseLighting);
 
+    builtinData.isLightmap =
+#if defined(LIGHTMAP_ON) || defined(DYNAMICLIGHTMAP_ON)
+        1;
```
sebastienlagarde, line 60: setup isLightmap

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -96,40 +96,40 @@ static const uint kFeatureVariantFlags[NUM_FEATURE_VARIANTS] =
     // Precomputed illumination (no dynamic lights) with standard, SSS and transmission
     /*  1 */ LIGHTFEATUREFLAGS_SKY | LIGHTFEATUREFLAGS_ENV | LIGHTFEATUREFLAGS_SSREFLECTION | MATERIALFEATUREFLAGS_LIT_SUBSURFACE_SCATTERING | MATERIALFEATUREFLAGS_LIT_TRANSMISSION | MATERIALFEATUREFLAGS_LIT_STANDARD,
     // Precomputed illumination (no dynamic lights) for all material types
-    /*  2 */ LIGHTFEATUREFLAGS_SKY | LIGHTFEATUREFLAGS_ENV | LIGHTFEATUREFLAGS_PROBE_VOLUME | LIGHTFEATUREFLAGS_SSREFLECTION | MATERIAL_FEATURE_MASK_FLAGS,
```
sebastienlagarde, line 99: @FrancescoC-unity would like to chat with you about this change. My understanding is that LIGHTFEATUREFLAGS_PROBE_VOLUME was added for V1. Which mean that with V2 this code is currently inefficient and force to always used the worst case variant. So this change here should currently bring back performance. Let's chat to validate it is true

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -685,32 +688,31 @@ void EncodeIntoGBuffer( SurfaceData surfaceData
...
+    // If this is 0 it mean that both bakeDiffuseLighting and emissiveColor are 0 and we are potentially in case of one effect
+    // so store AO instead. It doesn't matter if it is a false positive as result will be correct, this is to reduce code divergence
+    // Note: We assume that having non black baseColor * AO * lighting + Emissive is uncommon / rare (We expect mostly baseColor * AO * lighting or Emissive or baseColor * lighting + Emissive)
+    // and use this information as a tradeoff to improve quality in all others cases
+    if (all(outGBuffer3 == 0.0))
     {
-        outGBuffer3 = float4(builtinData.bakeDiffuseLighting * surfaceData.ambientOcclusion + builtinData.emissiveColor, 0.0);
-        // Pre-expose lighting buffer
-        outGBuffer3.rgb *= GetCurrentExposureMultiplier();
+        outGBuffer3.xz = AO_IN_GBUFFER3_TAG.xz;
```
sebastienlagarde, line 709: @FrancescoC-unity . I finally decided to do some kind of hack to store AO. See the comment, but short: if we don't have emissive, then we store AO. Should solve 90% of our use case and don't require extra buffer.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
```diff
@@ -236,16 +235,6 @@ static void BuildPerTileLightList(BuildGPULightListPassData data, ref bool tileF
...
                     }
 
-                    if (data.probeVolumeEnabled)
-                    {
-                        // If probe volume feature is enabled, we toggle this feature on for all tiles.
-                        // This is necessary because all tiles must sample ambient probe fallback.
-                        // It is possible we could save a little bit of work by having 2x feature flags for probe volumes:
-                        // one specifiying which tiles contain probe volumes,
-                        // and another triggered for all tiles to handle fallback.
-                        baseFeatureFlags |= (uint)LightFeatureFlags.ProbeVolume;
```
sebastienlagarde, line 246: guess it is obsolete code from V1

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRenderPipeline.RaytracingIndirectDiffuse.cs
```diff
@@ -256,70 +254,7 @@ class UpscaleRTGIPassData
                 return passData.outputBuffer;
             }
         }
-
-        class AdjustRTGIWeightPassData
```
sebastienlagarde, line 260: @AnisB To check it is correct to remove all this code

anisunity (approved):
Approved after changes

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -391,74 +385,99 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+        // No code change is required to handle AO, it is the same for all path.
+        // Note: Decals Emissive and Transparent Emissve aren't taken into account by RTGI/Mixed.
+        // Forward opaque emissive work in all cases. The current code flow with Emissive store in GBuffer3 is only to manage the case of Opaque Lit Material with Emissive in case of deferred
+        // Only APV can handle backFace lighting, all other effects are front face only.
+
+        // If we use SSGI/RTGI/Mixed effect, we are fully replacing the value of builtinData.bakeDiffuseLighting which is 0 at this step.
+        // If we are APV we only replace the non lightmaps part.
+        bool replaceBakeDiffuseLighting = false;
+#if defined(PROBE_VOLUMES_L1) || defined(PROBE_VOLUMES_L2)
+        float3 lightInReflDir = float3(-1, -1, -1); // This variable is used with APV for tinting reflection probe - see code for LIGHTFEATUREFLAGS_ENV
```
FrancescoC-unity, line 407: Tiny note but I wouldn't refer to the normalization as tinting as it might be confusing to the reader of the code.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
FrancescoC-unity: Let's chat about it yes, the thing was indeed added for PV1

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -685,32 +682,31 @@ void EncodeIntoGBuffer( SurfaceData surfaceData
...
+        outGBuffer3.y = surfaceData.ambientOcclusion;
     }
 
 #ifdef LIGHT_LAYERS
     // Note: we need to mask out only 8bits of the layer mask before encoding it as otherwise any value > 255 will map to all layers active
+    // If light layers is available, we take the opportunity to store AO in it to improve quality of indirect lighting with APV/SSGI/RTGI/Mixed
```
FrancescoC-unity, line 709: This comment seems wrong given what is done above?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
FrancescoC-unity: Yeah sounds like a valid solution indeed

remi-chapelain (approved):
This PR simplifies greatly the setup and need for comprehension when having a combinaison of real-time/baked global illumination technique with emissive materials. 

Before, in deferred, users needed to disable receive SSGI flag or enable Force Forward Emissive per material which created much confusion. 
Now, everything is supposed to work properly out of the box without having to enable anything new. 

PR have been tested thoroughly to ensure no unexpected regressions. 
What's tested: 
- Upgrade a project with SSGI and Force Forward Emissive ✔️ 
- SSGI / RTGI / Mixed still replaces lightmaps ✔
- Check transparent emissive still not contributing except when in pre-refraction for SSGI ✔️ 
- Decal Emissive still not contributing in RTGI or Mixed mode ✔️ 
- Building APV and relying on probe data to fill SSGI artefacts with emissive objects in the scenes ✔️ 
- Material with Diffuse, AO, Emissive with SSGI and/or APV  ✔️
_(in Deferred, with SSGI or APV, we lose material AO contribution, but not with Lightmaps)_
_(in Forward, everything works in all cases)_
- SSAO/RTAO still contributes as expected ✔️
- Can still multiply indirect lighting contribution for Lightmaps / APV / SSGI / RTGI / Mixed ✔️
- Emissive still works with SSS ✔️ 
- Emissive still works with RT SSS 🔴 (not related to this PR > reported [here](https://fogbugz.unity3d.com/f/cases/1364456/))
- DXR Tests are all green locally ✔️ 
- Performance (roughly same frame time in 4K in a scene with SSGI with lots of emissive / ao) ✔️

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
sebastienlagarde: udpated

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
sebastienlagarde: good catch, old code :)

sebastienlagarde:
@Vic-Cooper : Sorry I realize I forget to add you for doc checking on this PR. Could you add it in your backlog? Thanks a lot

PR #5578, opened 2021-09-09
Vfx/fix/1359426 shift context on new block

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1359426/
When inserting a new node in a context the context could then overlap the context below.
With the fix, each context below is pushed down to avoid overlapping (it's not moved if not needed).
[Before]
Before

[After]
After

---
### Testing status
I tested inserting new nodes with the filter window or with copy/paste an existing node.
I tested `Spawn`, `Init` and `Update` contexts.

---
### Comments to reviewers
If the contexts are side by side the behavior is the same even if it's not absolutely required the graph is still more readable if contexts are pushed down.


VitaSkruibyte:
Looking solid, but one question: When having _non-linear_ layout, contexts shift instead of staying where they are (attached gif as a showcase). Is there any way to identify if nothing is behind (for example there's no Update context below Init) so contexts would remain in the same location when having non-linear layout?
5tXiPWat3g

Blocks/contexts shift when and are clearly visible when:
- Adding blocks manually to spawn, init, update, output contexts
- Copy/pasting single and multiple blocks
- Having disconnected systems
- In Subgraphs

julienamsellem:
I changed a bit the behavior so that contexts are not shifted if they are side by side.
Until the approach is done differently there will still be some corner cases with unexpected behavior.
Today we try to adjust contexts by following the connections of the context that has increased.
But a more general approach could be a global graph analysis and fix any overlapping context/node.
But even that solution is not ideal because we should not change the layout too much in the risk of breaking the user's own graph organisation.

On edge case I found with the current fix is if two contexts are swapped vertically (output above update for example).

VladNeykov (approved):
Looking solid! Tested: 
- Adding, duplicating, copy-pasting, and moving single or groups of blocks
- Moving blocks to previous or next contexts
- Outputs connected to Update and Outputs connected directly to Initialize
- Multiple outputs _**(only those under the previous context will move down)**_
- Smoke-tested subgraph blocks added to contexts, as well as within the subgraph itself
- Different arrangements/placement of contexts when doing the above
- Different default templates
- Undo/Redo

Thanks for the fix! :)

PaulDemeulenaere (approved):
Side note, internal on GrowContext was actually already implicit
Otherwise, the code looks good and it's clear.
Thanks for this improvement. 🟢

PaulDemeulenaere:
Has been backported in https://github.com/Unity-Technologies/Graphics/pull/6444, updating flag

PR #5586, opened 2021-09-09
Fix previous matrix shader compilation errors with Hybrid Renderer

vincent-breysse:
### Purpose of this PR
Fixes previous matrix shader compilation errors with hybrid renderer. Hybrid renderer tests with SRP have been disabled for 2021 because hybrid renderer did not support 2021 Unity versions. So it seems that someone broke our code without noticing, since part of the CI was disabled.

The main issues were that :
- Previous matrices metadata wasn't defined with HDRP, which is needed by hybrid renderer
- `unity_MatrixPreviousM` and `unity_MatrixPreviousMI` were redefined, which caused issues when expanding the (recently added) macros `UNITY_PREV_MATRIX_M` and `UNITY_PREV_MATRIX_I_M`. Those were added on July the 28th as part of the `Hd/new-draw-api` branch : `0305bdfa86979df4d785f5c16687c03f618e3c84`
---
### Testing status
Testing here is a bit tricky because hybrid renderer has no support for SRP 13. We are working on it, but it hasn't landed yet. Although it's in a shape that is probably good enough for testing. So what I ended up doing is testing with :
- C++ `graphics/hrnext` (`6abba03edd2c2278475bc4c7ec876a4b74ea947e`)
- DOTS `master` (`45d8045f4474d4b4c5301d8a30741a14621c6cd4`)
- Graphics `hybrid/BRGTestProject` (`f1822e3e16ebf7ee3280c938fdba0ea05410471d`)

This branch `2022.1/fix-prev-matrix-hybrid-renderer` only contains the commit needed to fix shader compilation errors. It has been cherry-picked from the dev branch `hybrid/BRGTestProject`.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
@@ -460,9 +460,11 @@ float4x4 GetRawUnityPrevWorldToObject() { return unity_MatrixPreviousMI; }
 #define UNITY_PREV_MATRIX_M    ApplyCameraTranslationToMatrix(GetRawUnityPrevObjectToWorld())
 #define UNITY_PREV_MATRIX_I_M  ApplyCameraTranslationToInverseMatrix(GetRawUnityPrevWorldToObject())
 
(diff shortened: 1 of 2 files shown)


vincent-breysse:
This Hybrid Renderer fix will land at the same time as full Hybrid Renderer support for SRP 13. Keep the PR around because it will need to be ported for DOTS 0.5 to SRP 11/12

JussiKnuuttila (commented):
Looks good to me!

PR #5591, opened 2021-09-09
[2022.1] Fix for Volume Updates

ellioman:
# Purpose of this PR
In #4242 we introduced a setting to control the updates URP does to the Volume Framework.
There was however an issue when switching between `EveryFrame` and `ViaScripting` in the URP Asset after moving the camera. The issue is that the camera would use an old stack that had been saved earlier instead of an updated one right before the switch to ViaScripting. The reason for this was that the `volumeStack` is only updated if it was null or when the setting is changed on the camera. 

The solution is to destroy the `volumeStack` if the Volume Update Setting is set to `EveryFrame`. That guarantees that the volumeStack will be created and updated when the switch between `EveryFrame` and `ViaScripting` happens. This PR also also makes sure we do not update the stack when switching between `ViaScripting` and `UsePipeline` on the camera when the URP Asset is set to `ViaScripting`.

# Testing
### What has been tested
- Tested in the Editor (Mac) with three different cameras and two volumes

# Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/universal%252Fvolume-update-fix

### DIFF
--- a/TestProjects/UniversalGraphicsTest_PostPro/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/200_Lens_Flare_Sanity.png
+++ b/TestProjects/UniversalGraphicsTest_PostPro/Assets/ReferenceImages/Linear/Android/OpenGLES3/None/200_Lens_Flare_Sanity.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:cae48508af526f9db9a70ec309210d69b500c964d191707279d05d84e17283e0
+size 245521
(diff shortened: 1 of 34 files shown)


Verasl (approved):
LGTM simple path that was overlooked, this solves the issue, we might need to revisit a lot of this when we have time to re-do the volume system.

erikabar (approved):
Tested with URP-validation-project 07-VolumeFramework scene. Both Editor and Mac Standalone build are working as expected.

ellioman:
Failures on Yamato not related to this PR.

PR #5597, opened 2021-09-09
Vfx/fix/1315493 disable paste if not allowed

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1315493/

The context menu for "Contexts" in the VFX Graph had no `Paste` option.
Now the paste option is added and also to avoid confusion the option is disabled when we detect that the paste action is not allowed (previously there was an error message in the console)

---
### Testing status
- Paste block from one context to the same context or other context.
- Paste block on the graph or an operator (not allowed)
- Paste an operator on a context (the operator is pasted on the graph)

---
### Comments to reviewers


VladNeykov:
Hi @julienamsellem ,
Thanks for the fix, it looks good! Tried pasting blocks on blocks, contexts (works), nodes and the graph itself (doesn't, as expected).

I found a related issue which is not caused by your PR, but might be addressable here. I wonder if it's feasible to disable pasting a block subgraph onto itself as it leads to a crash:
https://user-images.githubusercontent.com/7230872/132842122-52dccc0b-100f-4b74-8934-f24713deccfd.mp4

julienamsellem:
> I found a related issue which is not caused by your PR, but might be addressable here. I wonder if it's feasible to disable pasting a block subgraph within itself when opened, as it leads to a crash:
> https://user-images.githubusercontent.com/7230872/132842122-52dccc0b-100f-4b74-8934-f24713deccfd.mp4
I'd prefer to open a new issue for this case so that this PR stays "atomic" and then easier to backport or revert :)

VladNeykov:
> I'd prefer to open a new issue for this case so that this PR stays "atomic" and then easier to backport or revert :)

Sounds good! :) In that case, no outstanding issues, I've logged the above-mentioned one here: https://fogbugz.unity3d.com/f/cases/1364480/

VladNeykov (approved):
Works as expected, no PR-specific issues found.
Thanks for the fix!

com.unity.visualeffectgraph/Editor/GraphView/Views/VFXPaste.cs
```diff
@@ -47,6 +47,27 @@ public static void PasteBlocks(VFXViewController viewController, object data, VF
...
+        {
+            try
+            {
+                var serializableGraph = JsonUtility.FromJson<SerializableGraph>(data.ToString());
+                if (serializableGraph.blocksOnly)
+                {
+                    var selectedContexts = view.selection.OfType<VFXContextUI>();
+                    var selectedBlocks = view.selection.OfType<VFXBlockUI>();
+
+                    return selectedBlocks.Any() || selectedContexts.Count() == 1;
```
julienamsellem, line 60: This is the  same check that is done when really doing the paste

com.unity.visualeffectgraph/Editor/GraphView/Views/VFXPaste.cs
```diff
@@ -85,7 +106,7 @@ void PasteBlocks(VFXView view, ref SerializableGraph serializableGraph, List<VFX
             }
             else
             {
-                Debug.LogError(m_BlockPasteError.text);
+                Debug.LogWarning(m_BlockPasteError.text);
```
julienamsellem, line 109: This is not a error, rather an action that is not allowed. But with my changes it should not happen anymore

PaulDemeulenaere (approved):
LGTM! 🟢

PR #5604, opened 2021-09-09
Multiple Scattering Approximation for Hair

johnpars:
### Purpose of this PR

This PR introduces an improved multiple scattering approximation to the HDRP Hair shading model. 

image

Currently we rely on the Kajiya-Kay diffuse component to approximate this, which generally is good enough for high-absorbing (dark) hair, but far from what is necessary to properly depict low-absorbing (light) hair. This is due to the multiple scattering most prominent in low-absorbent hair which gives (for example) blonde hair its voluminous look. 

Thus, this PR offers a new advanced scattering mode, and is based on the [dual scattering](http://www.cemyuksel.com/research/dualscattering/dualscattering.pdf) approximation. 

### Usage
Users access this new scattering mode through the following steps:
- Create a new Hair Shader Graph asset and open the graph.
- Navigate to the Graph Inspector window in the graph.
- Change the Material Type from Kajiya-Kay to Marschner.
- Change the Geometry Type from Cards to Strands.
- Change the Scattering Mode from Approximate to Advanced. 

The Advanced mode is based off of pre-integrated tables derived from the physically principled Marschner BSDF and currently is only confirmed to support strand geometry, which is why currently we only support the Advanced mode when the Marschner model is chosen. 

### Parameterization

Enabling the Advanced Scattering Mode will add only one new input block to the Shader Graph:

| Parameter  | Description |
| :-------------: | ------------- |
| _Strand Count Probe_  | L0 and L1 spherical harmonic coefficients that represent the number of strands on the sphere of directions at the shading point.  |

The only other terms that govern the multiple scattering are the _Base Color_ (defining the absorption) and the _Smoothness_ (defining the scattering) parameters. 

Comparison between Approximate and Advanced scattering modes.
image

### Performance

As expected, performance will come at roughly a ~15% extra cost as compared to the normal Marschner model with simplified Kajiya-Kay diffuse component. Below are some performance metrics captured on a standalone Windows build with a NVIDIA Quadro RTX 8000, at 3840x2160 resolution, no MSAA.

**Scenario #1**: Three screen-aligned quads (Kajiya, Marschner with and without dual scattering), each taking up a third of the viewport. Illuminated by one point, one directional, and two environment light evaluations (HDRI and realtime reflection probe).

| Model  | Forward Opaque Time Per-Quad (ms) |
| Kajiya-Kay  | 0.61  |
| Marschner (No Dual Scattering)  | 0.75 |
| Marschner (Dual Scattering)  | 0.85  |

image

Side note, there is an emissive red ball in the reflection probe causing the red result for the Marschner quads, which evaluate environment lighting different from Kajiya. Additionally, the test quad doesn't have any meaningful strand count probe as input to the model, so the visual result is not relevant in this case. 

### Light Loop Integration

| Light Type |   |
|:---:|---|
| _Analytic_  |  All analytic lights in HDRP (point, spot, directional) support advanced multiple scattering. |
| _Area_  | For now, like Kajiya, we must fall back to GGX for area lights. Thus they do not currently support advanced multiple scattering. |
| _Environment_  | Environment lighting currently does not support advanced multiple scattering. It falls back to the Kajiya-Kay diffuse component.  |

### Testing 

Updated the hair graph test to cover the new multiple scattering case, as well as the fixed environment lighting evaluation.
I only added a general singular case as the other test permutations already have good coverage from the initial Kajiya and Marschner tests. 

image

### Comments to reviewers

The algorithm is dependent on the number of strands between the shading position and light. This information is typically computed in real-time via deep opacity maps, however for now we retrieve this information via an L1 spherical harmonic "strand probe" that allows us to quickly decode the strand count for analytic lights. This probe field is readily generated by the demo team's hair package. 

This PR also fixes an issue with environment light evaluation. Previously we utilized the light probe lighting and treated it as a directional light; I forgot to transpose this concept to the actual environment evaluation (and not from the diffuse probes), and thus it has been corrected. Environment lighting should appear correct now. Note that we use the lowest mip of the environment instead of a spherical harmonic representation, but it might be possible in the future to do that latter with APV. 

It's important to note that the dual scattering approach is currently biased to depict straight to wavy hair, due to the assumptions it makes about hairs being oriented similarly in the shading neighborhood and toward the light. Sebastien Lachambre wrote a great [document ](https://docs.google.com/document/d/1zSxEvHnru2osUiatAvvN9irBWxei55KRMyMU5PDQRNw/edit?usp=sharing) going into further detail on hair types and parameterization. It is worth improving our multiple scattering approximation in the future to ensure we have good representation for all hair types. Below is a table describing the hair type classifications (this PR should be good to approximate 1-3, but might break down in quality for 3-7 in low-absorbent hair). 

 image

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Hair/MultipleScattering/HairMultipleScattering.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Hair/MultipleScattering/HairMultipleScattering.hlsl
@@ -28,7 +28,7 @@ float DecodeHairStrandCount(float3 L, float4 strandCountProbe)
         HALF_SQRT_3_DIV_PI * L.x
     );
 
-    return dot(strandCountProbe, Ylm);
+    return abs(dot(strandCountProbe, Ylm));
 }
 
 float GetDirectFraction(BSDFData bsdfData, float strandCount)


com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/ShaderPass.template
```diff
@@ -19,6 +19,7 @@ Pass
 
     // Pragmas
     $splice(PassPragmas)
+    #pragma enable_d3d11_debug_symbols
```
sebastienlagarde, line 22: miss comment

sebastienlagarde (approved):

PR #5611, opened 2021-09-10
Prevent errors when loading incompatible assets in APV

FrancescoC-unity:
Up to now we didn't have any check on whether we were loading assets that are incompatible. With the refactor of the backend of the workflow I did recently (not relying on the panel, hence why the PR targets master and not that branch), I added information in the asset we can use to detect the case.

This change leverage said information and prevents bad scenarios to happen. Before we'd just catastrophically fail, now we log an error and stop loading incompatible assets. 

https://user-images.githubusercontent.com/43168857/132894772-fccacca3-3cd5-4565-b2c3-ff11fb01dbff.mp4

This caused some serious issues also with the baking panel workflow as if we bake back-to-back sets that are different it will be spawning a lot of errors (hence why in the gif I tested with the panel). 

**NOTE:** This reference a way to fix the issue that depends on the panel to go through, so even if this PR is independent, we should merge the panel first.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
+++ b/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbePlacement.cs
@@ -607,8 +607,9 @@ namespace UnityEngine.Experimental.Rendering
                     // Compute the max size of a brick that can fit in the smallest dimension of a probe volume
                     float minSizedDim = Mathf.Min(pvAABB.size.x, Mathf.Min(pvAABB.size.y, pvAABB.size.z));
                     float minSideInBricks = Mathf.CeilToInt(minSizedDim / ProbeReferenceVolume.instance.MinBrickSize());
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -101,6 +101,7 @@ internal void QueueAssetLoading()
                     refVol.dilationValidtyThreshold = refVol.sceneData.GetBakeSettingsForScene(gameObject.scene).dilationSettings.dilationValidityThreshold;
                 }
 #endif
+                m_PreviousState = m_CurrentState;
```
FrancescoC-unity, line 104: Note for @JulienIgnace-Unity : this is the change we discussed the other day for the double-loading, since we now spawn an error on wrong loading, having it spawn twice is annoying :D

JulienIgnace-Unity (approved):

alelievr (approved):

PR #5614, opened 2021-09-10
[ShaderGraph][2022.1] Fix direction transform from view to world space

jessebarker:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1362034/

Transform direction from view to world was effectively doing a position transform rather than direction.  This change fixes that.

---
### Testing status
Manual testing according to the repro case in the bug (confirmed that the transform from world->view and then back from view->world yields vec3(0, 0, 0) when subtracted from the original vec3.

Screen Shot 2021-09-10 at 2 32 02 PM

Yamato details below.

Here are the built-in test results that got "erased" when the ShaderGraph test project reference images got updated (reference images from TestProjects/ShaderGraph have no impact on the two built-in target test projects):
- [PR ShaderGraph BuiltIn Foundation on trunk](https://unity-ci.cds.internal.unity3d.com/job/8663624)
- [PR ShaderGraph BuiltIn Lighting on trunk](https://unity-ci.cds.internal.unity3d.com/job/8663841)

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Nodes/Math/Vector/TransformNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Math/Vector/TransformNode.cs
@@ -191,7 +191,7 @@ namespace UnityEditor.ShaderGraph
                 {
                     transformString = string.Format(conversionType == ConversionType.Direction ?
                         "mul(UNITY_MATRIX_I_V, $precision4({0}, 0)).xyz" :
-                        "mul(UNITY_MATRIX_I_V, $precision4({0}, 1)).xyz", inputValue) ;
+                        "mul(UNITY_MATRIX_I_V, $precision4({0}, 1)).xyz", inputValue);
                 }
                 else if (conversion.to == CoordinateSpace.Object)
                 {


jessebarker:
Backports on this one?  @marctem

com.unity.shadergraph/Editor/Data/Nodes/Math/Vector/TransformNode.cs
```diff
@@ -189,7 +189,7 @@ public void GenerateNodeCode(ShaderStringBuilder sb, GenerationMode generationMo
             {
                 if (conversion.to == CoordinateSpace.World)
                 {
-                    transformString = string.Format("mul(UNITY_MATRIX_I_V, $precision4({0}, 1)).xyz", inputValue);
```
joshua-davis, line 192: Won't this make TransformViewToWorld wrong for position? I believe this needs a conditional like the otehr cases (see ViewToObject below).

joshua-davis (approved):

com.unity.shadergraph/Editor/Data/Nodes/Math/Vector/TransformNode.cs
jessebarker: Good catch.  Fixed.  I think this also means the test images don't need to be updated, but I'm re-running to confirm.

joshua-davis: Actually, if the images don't need to be updated, this implies we don't have a test for both scenarios. Maybe adding new nodes might be good.

PR #5617, opened 2021-09-12
Backport of 2d/fix for self shadow 

unity-cchu:
### Checklist for PR maker
- [X] Have you updated the changelog? Each package has a `CHANGELOG.md` file.

---
### Purpose of this PR
This is a backport of https://github.com/Unity-Technologies/Graphics/commit/0d1d0f89318c244137f1c555168112be8b7ad6c3

### DIFF
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [10.10.0]
+## [Unreleased]
 
 ### Fixed
 - Fixed an issue where ShadowCaster2Ds were sometimes being rendered twice in the editor while in playmode.


com.unity.render-pipelines.universal/Runtime/2D/Shadows/ShadowCasterGroup2DManager.cs
```diff
@@ -2,15 +2,36 @@
...
 
+#if UNITY_EDITOR
+        static ShadowCasterGroup2DManager()
+        {
+            EditorApplication.playModeStateChanged += OnPlayModeStateChanged;
+        }
+
+        private static void OnPlayModeStateChanged(PlayModeStateChange state)
+        {
+            if (state == PlayModeStateChange.ExitingEditMode || state == PlayModeStateChange.ExitingPlayMode)
```
kennytann, line 29: You should add a null check for s_ShadowCasterGroups here as it was causing a NullReferenceException

kennytann (approved):

com.unity.render-pipelines.universal/Runtime/Materials/ParticlesUnlit.mat
```diff
@@ -26,12 +26,13 @@ Material:
   m_LightmapFlags: 4
   m_EnableInstancingVariants: 0
   m_DoubleSidedGI: 0
-  m_CustomRenderQueue: 3050
+  m_CustomRenderQueue: 3000
```
phi-lira, line 29: Maybe this was due to an upgrader, but at this point we should not backport such changes to LTS. 
Please revert these material changes.

phi-lira (changes requested):

com.unity.render-pipelines.universal/Runtime/Materials/ParticlesUnlit.mat
unity-cchu: Removed changed files

PR #5619, opened 2021-09-13
Disable shadowmask option when not enabled in the Lighting settings

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1361956/

If the Lighting Mode is not set to ShadowMask in the Lighting settings, the shadowmask option in the Light Inspector is now disabled.

Also updated the tooltip to explain why it is disabled (because t also needs to be enabled in the HDRP asset so that might be confusing)

---
### Testing status

Tested to change the lighting mode


TomasKiniulis (approved):
Checked if option gets disabled/enabled changing HDRP asset Shadowmask support.
Switching between assets in Graphics and Quality settings and changing Lighting Modes. Looking good!

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.Skin.cs
```diff
@@ -45,7 +45,7 @@ sealed class Styles
...
             public readonly GUIContent affectSpecular = EditorGUIUtility.TrTextContent("Affect Specular", "When disabled, HDRP does not calculate specular lighting for this Light. Does not increase performance as HDRP still calculates the specular lighting.");
-            public readonly GUIContent nonLightmappedOnly = EditorGUIUtility.TrTextContent("Shadowmask Mode", "Species the behavior of  the shadowmask when using Mixed lighting. Distance Shadowmask: HDRP uses real-time shadows to Shadow Distance and baked shadows after. Shadowmask: Static shadow casters always use baked shadows.");
+            public readonly GUIContent nonLightmappedOnly = EditorGUIUtility.TrTextContent("Shadowmask Mode", "Species the behavior of  the shadowmask when using Mixed lighting. Distance Shadowmask: HDRP uses real-time shadows to Shadow Distance and baked shadows after. Shadowmask: Static shadow casters always use baked shadows.\nOnly available if Shadowmask is enabled in the HDRP asset and Lighting Mode is set to Shadowmask in the Lighting Settings.");
```
radishface, line 48: Please modify the tooltip text with the following:

> Determines Shadowmask functionality when using Mixed lighting. Distance Shadowmask casts real-time shadows within the Shadow Distance, and baked shadows beyond. In Shadowmask mode, static GI contributors always cast baked shadows.
> Enable Shadowmask support in the HDRP asset to make use of this feature. Only available when Lighting Mode is set to Shadowmask in the Lighting window.

radishface (changes requested):
Please change the tooltip text, as per my recommendation. Otherwise, changes look good!

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.Skin.cs
adrien-de-tocqueville: thanks for the rewrite, done

radishface (approved):
Looks good! Thank you for implementing the change so quickly!

sebastienlagarde (approved):

PR #5620, opened 2021-09-13
planet center position support

emilybrown1:
In response to: https://fogbugz.unity3d.com/f/cases/1364180/

Added a note to say that volumetric clouds and cloud layer do not support planet center position

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/clouds-in-hdrp.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/clouds-in-hdrp.md
@@ -7,4 +7,4 @@ The High Definition Render Pipeline (HDRP) includes two cloud solutions that you
 
 You can use both solutions together. In this case, you can use Volumetric Clouds for closer, interactable (receives fog and volumetric light) clouds, and Cloud Layer for distant clouds that are part of the skybox.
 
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Documentation~/clouds-in-hdrp.md
```diff
@@ -6,3 +6,5 @@ The High Definition Render Pipeline (HDRP) includes two cloud solutions that you
 - [Volumetric Clouds](Override-Volumetric-Clouds.md)
 
 You can use both solutions together. In this case, you can use Volumetric Clouds for closer, interactable (receives fog and volumetric light) clouds, and Cloud Layer for distant clouds that are part of the skybox.
+
+__Note:__ Neither Cloud Layer or Volumetric Clouds support Planet Center Position.
```
Vic-Cooper, line 10: Neither should be paired with "nor" but seeing as that's kind of archaic language (and "either" doesn't work here) I recommend the following: 

__Note:__ Cloud Layer and Volumetric Clouds do not support Planet Center Position.

Vic-Cooper (changes requested):
Added a grammatical suggestion here.

sebastienlagarde (approved):

PR #5621, opened 2021-09-13
Fixed the volumetric clouds cloud map not being centered over the world origin (case 1364465).

anisunity:
This makes sure that the cloud map's center is above (0,0,0)
https://fogbugz.unity3d.com/f/cases/resolve/1364465/

### Testing status
Made sure the world center was at the middle of the pattern in the repro case.
image
I don't think any test will fail (not sure)


JMargevics:
Subtracting 0.5 seem a pretty straightforward fix and if repro case works correctly now, no additional testing is required ✔

JMargevics (approved):

PR #5624, opened 2021-09-13
nest and path fixup

esmelusina:
---
### Purpose of this PR
This change nests the editor folder in preparation for breaking future work down into additional editor assemblies.

---
### Testing status
Existing automated tests should be all that's necessary.

### DIFF
--- a/com.unity.shadergraph/ValidationConfig.json
+++ b/com.unity.shadergraph/ValidationConfig.json
@@ -3,7 +3,7 @@
 	{
 		"IgnoreFilenames":
 		[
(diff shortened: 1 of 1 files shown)


jessebarker:
Note to other reviewers (I learned a new git thing today - sorry if this is otherwise well known):

'git show' accepts the command line options for 'git diff'.  So, 'git show --diff-filter=r <commit>' means "exclude renames".  Very useful when you know that most of your 1700 file PR is just straight renames (there are also options to change what git considers a rename by percentage of difference).

PR #5630, opened 2021-09-13
Added multi-edit support to urp asset

Nzollner:
### Purpose of this PR
The purpose of this PR is to add multi-edit to the universal render pipeline asset, described in this [fogbugz case](https://fogbugz.unity3d.com/f/cases/1364966/).

This is only a UI change and because of that, I have not started any testing since there should be no changes to the result.

image

---
### Testing status
I have tested the multi-edit on my laptop(Windows 10, using dx11).
I tested the multi-edit by checking if I could change the values and if the values change affected the view and if the actions were undoable which they were.

---
### Comments to reviewers
When you test this check if the values work as expected and if the changes correspond to the visual changes which should be applied with the changed values.


alex-vazquez (approved):
Looks good to me

martint-unity (approved):
LGTM

PR #5632, opened 2021-09-13
Fix APV debug using profile and not baked data + small probe volume init

FrancescoC-unity:
This PR has a fix for the debug view using the wrong cell size for baked data that lead to wrong visualization. 

Instead we grab the actually loaded info if we are displaying data for baked data, otherwise we use the info from the profile for realtime subdivision debug view. 

The PR has also another small check for probe volume UX if the feature gets disabled.

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
+++ b/com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
@@ -111,7 +111,7 @@ namespace UnityEditor.Experimental.Rendering
 
         static void Drawer_VolumeContent(SerializedProbeVolume serialized, Editor owner)
         {
-            if (!ProbeReferenceVolume.instance.isInitialized)
+            if (!ProbeReferenceVolume.instance.isInitialized || !ProbeReferenceVolume.instance.enabledBySRP)
             {
                 var renderPipelineAsset = UnityEngine.Rendering.RenderPipelineManager.currentPipeline;
                 if (renderPipelineAsset != null && renderPipelineAsset.GetType().Name == "HDRenderPipeline")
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
```diff
@@ -111,7 +111,7 @@ static void Drawer_ToolBar(SerializedProbeVolume serialized, Editor owner)
 
         static void Drawer_VolumeContent(SerializedProbeVolume serialized, Editor owner)
         {
-            if (!ProbeReferenceVolume.instance.isInitialized)
+            if (!ProbeReferenceVolume.instance.isInitialized || !ProbeReferenceVolume.instance.enabledBySRP)
```
JulienIgnace-Unity, line 114: When can this happen?
Start with APV enabled in the asset then switch it off?
Sounds like we should cleanup the APV in this case no?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.cs
```diff
@@ -173,22 +173,28 @@ bool IsResponsibleToDrawGizmo()
 
         internal bool ShouldCullCell(Vector3 cellPosition, Vector3 originWS = default(Vector3))
         {
-            var profile = ProbeReferenceVolume.instance.sceneData.GetProfileForScene(gameObject.scene);
-            if (profile == null)
-                return true;
+            var cellSizeInMeters = ProbeReferenceVolume.instance.MaxBrickSize();
```
JulienIgnace-Unity, line 176: Shouldn't we factor this piece of code inside MaxBrickSize() instead? This way it's always the good value that we get.

FrancescoC-unity: No, because MaxBrickSize should always be the one loaded. The real time subdiv is coming from profile  because is a special case.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Drawer.cs
FrancescoC-unity: It can be valid but not enabled by current frame settings

JulienIgnace-Unity: Ah right, forgot about the frame settings and thought it was only the global settings.

JulienIgnace-Unity (approved):

PR #5635, opened 2021-09-13
[10.x] Define ENABLE_HYBRID_RENDERER_V2 by default for Hybrid Renderer 0.14.0 and newer [V2]

vincent-breysse:
Enable HRV2 by default for DOTS 0.5 since HRV1 will be removed. 
--------
This PR is completing (or rather reworking) the V1 that can be found here https://github.com/Unity-Technologies/Graphics/pull/5507
Unfortunately, we noticed a bit too late that more asmdef files had to be updated

### Testing status
This change is supposed to be tested with HRV1 removed on a project not defining `ENABLE_HYBRID_RENDERER_V2` explicitly in the ProjectSettings. 
Since HRV1 removal has not landed on DOTS master yet, I tested by locally adding the same asmdef change to the file `com.unity.rendering.hybrid/Unity.Rendering.Hybrid/Unity.Rendering.Hybrid.asmdef`. This is to ensure I wasn't testing with the define enabled globally for the project, which would be pointless.

- **2020.3/partner/staging** (bc0db7ac94bdb504a4d9cb8e0ab8f891b5fc25ec)
- **DOTS master** (82663ca7a46d2bceaa346e5b4a34468276a826b3)

### com.unity.rendering.hybrid
- Added: Define `ENABLE_HYBRID_RENDERER_V2` by default for Hybrid Renderer 0.14.0 and newer


com.unity.render-pipelines.core/Runtime/Unity.RenderPipelines.Core.Runtime.asmdef
```diff
@@ -30,6 +30,11 @@
             "name": "com.unity.rendering.hybrid",
             "expression": "0.6.0-preview.0",
             "define": "HYBRID_0_6_0_OR_NEWER"
+        },
+        {
+            "name": "com.unity.rendering.hybrid",
+            "expression": "0.14.0-preview.0",
+            "define": "ENABLE_HYBRID_RENDERER_V2"
```
joeldevahl, line 37: I don't think this is actually needed in the end. No harm in keeping it for now though

joeldevahl (approved):

com.unity.render-pipelines.core/Runtime/Unity.RenderPipelines.Core.Runtime.asmdef
vincent-breysse: Yes, it seems it shouldn't be needed, but since one file is using it in the assembly I just added it here as well.
The file in question for the record : https://github.com/Unity-Technologies/Graphics/blob/10.x.x/hrv2-default/com.unity.render-pipelines.core/Runtime/AssemblyInfo.cs

vincent-breysse:
Needs to be forward ported once we can test DOTS with SRP 11/12. Currently, the latest editor version compatible with DOTS is `2020.3/partner/staging`, which is only compatible with SRP 10

com.unity.render-pipelines.high-definition/Editor/Unity.RenderPipelines.HighDefinition.Editor.asmdef
```diff
@@ -30,6 +30,11 @@
             "name": "com.unity.xr.management",
             "expression": "1.0.0",
             "define": "ENABLE_XR_MANAGEMENT"
+        },
+        {
+            "name": "com.unity.rendering.hybrid",
+            "expression": "0.14.0-preview.0",
+            "define": "ENABLE_HYBRID_RENDERER_V2"
         }
```
PaulDemeulenaere, line 38: It impacts the usage of 
- [DecalSubTarget](https://github.com/Unity-Technologies/Graphics/blob/10.x.x/release/com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalSubTarget.cs#L491)
- [HDTarget](https://github.com/Unity-Technologies/Graphics/blob/10.x.x/release/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDTarget.cs#L717)
It looks good to me, also discussed about this with @alelievr

PaulDemeulenaere (approved):
LGTM 🟢

cdxntchou (approved):

PR #5636, opened 2021-09-13
Fixed the emissive being overriden by ray traced sub-surface scattering (case 1364456).

anisunity:
https://fogbugz.unity3d.com/f/cases/1364456/
The emissive contribution was killed by the RTSSS. The emissive contribution will be slightly lower with RTSSS as we multiply it by the diffuse signal in the RTSSS case.

Testing status


remi-chapelain (approved):
Has been discussed and new behavior has been "pair-validated" ✔️ 
Officially, we still don't support it (hence the doc), but the behavior will still work for most scenarios, so still an improvement.

sebastienlagarde (approved):

PR #5637, opened 2021-09-13
Fix terrain grass shader alpha write.

eh-unity:
### Checklist for PR maker

---
### Purpose of this PR
Fix terrain grass shader.

Issue:
035_Shader_TerrainShaders test was unstable and often failed with white grass. Turns out the white was actually transparent black.

The issue was something like this:
1. Background is cleared, alpha == 0.
2. Terrain writes alpha 1, write Z buffer.
3. Grass drawn on top of terrain and background, but does not write alpha. Writes z buffer.
4. Sky rendered, writes alpha 1, but is occluded by Z i.e. the grass on background. 
5. Grass parts are left transparent alpha == 0. 

Grass shaders had masked alpha writes, this PR removes those.
The issue then is that alpha on the edges of grass blades can make holes (alpha < 1) to a Render Texture target.
So instead alpha mask, the shaders write 1 to alpha always, since the shader isn't doing blending anyway, just discard.

For the 2nd case I added additional test that renders the terrain into a render texture and then that to a quad with another red quad behind that. So if the render texture has any "holes" (i.e. grass overwrote terrain alpha) those will show as red pixels.

Also removed the Shader Lab `AlphaTest` command from the shaders as it is not used on SRPs.

---
### Testing status
Ran test locally. Added a new test for Terrain grass in Render Texture.

This will need some additional manual QA for terrain grass.

---
### Comments to reviewers


TestProjects/UniversalGraphicsTest_Terrain/Assets/CommonAssets/ZeroVariantColorTextureAlphaOnly.shader
```diff
@@ -0,0 +1,125 @@
...
+        {
+            Name "ZeroVariantColorTexAlphaOnly"
+
+            HLSLPROGRAM
+                #pragma vertex Vert
+                #pragma fragment Frag
+            ENDHLSL
+        }
+
+        /*Pass
```
TimAidleyAtUnity, line 95: Should these commented out sections just be removed?

eh-unity: Yes. I will remove those.

ellioman (approved):

PR #5644, opened 2021-09-13
[ShaderGraph][2021.2][Backport] Fix for 1365186 - correct direction transform from view to world space

jessebarker:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1365186/

Transform direction from view to world was effectively doing a position transform rather than direction. This change fixes that.
---
### Testing status
Without these changes, transforming (direction) from world->view->world and subtracting the result from the original vector yields non-0 value in Z:
image

With these changes:
Screen Shot 2021-09-10 at 2 32 02 PM

---
### Comments to reviewers


jessebarker:
The draft reminder check/action is failing, but it looks like it is set up wrong (trying to do something on 'master').  Everything else is passing.

PR #5647, opened 2021-09-14
Fixed disabled menu item for volume additional properties.

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1365278/

"Show All Additional Properties" was disabled when the volume didn't have any additional properties, whereas it should always be enabled

---
### Testing status

tested that the menu is present in volumes with and without additional properties


alex-vazquez (approved):
Thanks

JMargevics (approved):
What's tested is sufficient, approving as is 👍✔

sebastienlagarde (approved):

PR #5649, opened 2021-09-14
Match UI and doc for lit shaders advanced options

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1361637/

Removed "motion vector for vertex animation" field on layered lit shaders, as far as i understand it, it's only valid for shadergraphs.
Moved "global illumination" field from "advanced options" to "emission inputs" to make it match between all lit shaders.

Updated the docs and used snippets to remove duplicates


com.unity.render-pipelines.high-definition/Editor/Material/LayeredLit/LayeredLitGUI.cs
```diff
@@ -22,8 +22,6 @@ internal struct SerializeableGUIDs
     class LayeredLitGUI : HDShaderGUI
     {
         const LitSurfaceInputsUIBlock.Features commonLitSurfaceInputsFeatures = LitSurfaceInputsUIBlock.Features.LayerOptions;
-        // Layered lit shaders don't support emission (realtime or baked)
```
adrien-de-tocqueville, line 25: @sebastienlagarde  is that an out of date comment ? because emission field is currently displayed

JMargevics (approved):
UI change makes sense and doesn't need any more testing from our side. Approving as is ✔👍

com.unity.render-pipelines.high-definition/Editor/Material/LayeredLit/LayeredLitGUI.cs
sebastienlagarde: yes

sebastienlagarde (approved):

PR #5663, opened 2021-09-15
[HDRP][DXR] Fix recursive rendering transmittance over the sky

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1323945/

Transmittance was not tinting the sky correctly. The reason was that the travel distance for missed rays was left to zero (the initialization value) and not the _"max length"_, so it's a rather trivial fix.

---
### Testing status
Using the repro scene from the ticket.

Before:
image

After:
image

---
### Comments to reviewers
Alternatively we could have the miss shader write the max_length. I opted to initialize the variable with the correct value, since this does not add any extra instructions in the ray tracing inner loop.


pmavridis:
Note: The but ticket also mentions that enabling the "Transparent write motion vectors" flag adds extra tint to the image. 
This happens because in this case the rasterized version of the effect is computed on top of the ray traced one, so this is unrelated to the bug that is fixed in this PR (and we should fix in another PR).

anisunity (approved):
Looks simple enough as a fix, hope it doesn't have any side effects.

sebastienlagarde (approved):

iM0ve (approved):
I have verified the fix too to better understand. In terms of regression testing, I opened HDRPRayTracingScenes project to compare with master. No issues found, I think should be safe if graphic tests are passing

PR #5672, opened 2021-09-15
[ShaderGraph][2022.1] Fix 1356725 - incorrect appearance of warning message in Node Settings

jessebarker:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1356725/

The behavior for the HelpBoxRow changed and this instance did not get updated.  This change updates the one use (that I could find) that doesn't handle the label setup correctly.  With this change, we match the "without" image in the bug report - i.e., the text is both in the correct color and the text wraps so that the whole message is viewable.

Fixed (labeled "without" in the bug report):
image

Broken (labeled "with" in the bug report):
image

---
### Testing status
Manual testing according to the repro steps in the bug report.  I could not find any other uses of HelpBoxRow with message type warning elsewhere in the code base, so, contrary to the claim in the bug, I could not make a node with an error (e.g., an invalid custom function node) generate a similar warning in Node Settings.

For when the checks inevitably get erased (due to resolving changelog conflicts):
[PR ShaderGraph on trunk](https://unity-ci.cds.internal.unity3d.com/job/8761331)

---
### Comments to reviewers


Nightmask3 (approved):
Looks good to me!

PR #5673, opened 2021-09-15
Cube, 2D and 3D Texture Parameters for Volume system

adrien-de-tocqueville:
### Purpose of this PR

Fix for https://fogbugz.unity3d.com/f/cases/1365393

There was a TextureParameter in the volume framework to have textures in volume components but it was not able to make any distinction based on the texture type. A CubemapParameter class was recently added for cubemaps.
This PR adds also Texture2DParameter and Texture3DParameter to complete.

Updated various components to use the correct parameter type (which introduce an API change so can't be backported)
For Texture2D:
- Bloom
- Vignette
- CloudLayer
- HDRISky
- ChromaticAberration
- Exposure
- FilmGrain

For Texture3D:
- Tonemapping

The Texture2DParameter only accept Texture2D and Render Textures of dimensions 2D (and same for 3D).
Assigning anything else is impossible to input and will have no effect.
Also, less incompatible textures will be shown in the object picker. All render textures are still displayed, even with wrong dimensions. And for some reason the cubemaps also show in the 2D texture field, not sure why.

For HDRI Sky:
image

---
### Testing status

Tested inputting correct and wrong textures into the three types of parameters.
Also tested drag and dropping.


alelievr (approved):

remi-chapelain:
Kudos for trying and fixing on overrides other than bloom too ! 👍

com.unity.render-pipelines.core/Editor/Volume/Drawers/TextureParameterDrawer.cs
```diff
@@ -0,0 +1,132 @@
...
+        static readonly MethodInfo k_ValidateObjectFieldAssignment = Type.GetType("UnityEditor.EditorGUI,UnityEditor").GetMethod("ValidateObjectFieldAssignment", BindingFlags.Static | BindingFlags.NonPublic);
+        static readonly MethodInfo k_EditorGUI_DoObjectField = Type.GetType("UnityEditor.EditorGUI,UnityEditor")
+            .GetMethod("DoObjectField", BindingFlags.Static | BindingFlags.NonPublic, null,
+                    new Type[] { typeof(Rect), typeof(Rect), typeof(int), typeof(Object), typeof(Object), typeof(Type), typeof(Type), typeof(SerializedProperty), k_ObjectFieldValidator, typeof(bool), typeof(GUIStyle) }, null);
+
+        internal delegate Object ObjectFieldValidator(Object[] references, Type objType, SerializedProperty property, int options);
+
+        internal static Object ValidateObjectFieldAssignment(Object[] references, Type objType, SerializedProperty property, int options)
+        {
+            return (Object)k_ValidateObjectFieldAssignment.Invoke(null, new object[] { references, objType, property, options });
```
RSlysz, line 21: [Performence] You should compile once using linq tree at static construction instead of researching the method on the object each time with Invoke. Here it is a static one. Not 100% sure there is perf to gain beside the GC of the object array.

com.unity.render-pipelines.core/Editor/Volume/Drawers/TextureParameterDrawer.cs
```diff
@@ -0,0 +1,132 @@
...
+            DoObjectField(r, property, label, type1, type2, validator);
+        }
+
+        internal static void DoObjectField(Rect r, SerializedProperty property, GUIContent label, Type type1, Type type2, Delegate validator)
+        {
+            int id = GUIUtility.GetControlID(FocusType.Keyboard, r);
+            r = EditorGUI.PrefixLabel(r, id, label);
+
+            var parameters = new object[] { r, r, id, null, null, type1, type2, property, validator, false, EditorStyles.objectField };
+            k_EditorGUI_DoObjectField.Invoke(null, parameters);
```
RSlysz, line 41: [Performence] You should compile once using linq tree at static construction instead of researching the method on the object each time with Invoke. Here it is a static one. Not 100% sure there is perf to gain beside the GC of the object array.

RSlysz (approved):
Nice work. Thanks for adding also other texture types.

Can you check the GC optim on editor side though?

Also, @martint-unity for awarness. I dunno if there is such texture in URP volume that can profite of it.

JMargevics:
image
Hey, could we treat Weight Texture Mask this way too?. Otherwise I've also checked all other components, everything accepts only correct textures now 👍

com.unity.render-pipelines.core/Editor/Volume/Drawers/TextureParameterDrawer.cs
adrien-de-tocqueville: for this one I didn't manage to do it because the function DoObjectField takes an argument internal to the engine
So I can't define a delegate at compile time for this function signature (or I don't know how to)

adrien-de-tocqueville:
@JMargevics Itaht's fixed, as well as another texture in filmgrain 🙂

com.unity.render-pipelines.core/Editor/Volume/Drawers/TextureParameterDrawer.cs
adrien-de-tocqueville: actually it's the same issue as the one below for this function, I reverted the change

JMargevics (approved):
All good now 👍✔

sebastienlagarde (approved):

PR #5674, opened 2021-09-15
Enable renderer sorting for mesh decals

adrien-de-tocqueville:
### Purpose of this PR

This PR is enabling sorting based on the meshrenderer priority value for mesh decals.
5674_priority

It can also sort based on the draw order parameter of the decal material by making it impact the renderqueue.
5674_renderqueue

I made the draw order to be clamped between -50 and +50, which is not necessary, but shouldn't break existing project as it doesn't modify existing values until they are changed. But having the clamp looks nicer in the UI

note: not sure it should be backported, maybe it's only for master

---
### Testing status

tested the sorting with two mesh decals


JMargevics (approved):
Tried to upgrade a project containing mesh decals with previously set Mesh Renderer priorities and everything upgraded and got sorted correctly ✔

Tried sorting manually, undo, redo, also no issues. ✔

Looks good to me 👍

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderQueue.cs
```diff
@@ -110,6 +110,8 @@ internal static RenderQueueType MigrateRenderQueueToHDRP10(RenderQueueType rende
         public const int sortingPriortyRange = 50;
         public static int ClampsTransparentRangePriority(int value) => Math.Max(-sortingPriortyRange, Math.Min(value, sortingPriortyRange));
 
+        public const int meshDecalPriortyRange = 50;
```
sebastienlagarde, line 113: there is a typo: meshDecalPriortyRange +> meshDecalPriorityRange  typo was already in sortingPriortyRange :)

sebastienlagarde: note that here add a comment to say that it is a RendererPriority unlike sortingPriortyRange  which is a RenderQueue priority

sebastienlagarde (approved):

PR #5678, opened 2021-09-16
[ShaderGraph][2021.2][Backport] Fix for 1365780 - node settings warning message formatting

jessebarker:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1365780/

The behavior for the HelpBoxRow changed and this instance did not get updated.  This change updates the one use (that I could find) that doesn't handle the label setup correctly.  With this change, we match the "without" image in the bug report - i.e., the text is both in the correct color and the text wraps so that the whole message is viewable.

Fixed (labeled "without" in the bug report):
image

Broken (labeled "with" in the bug report):
image

---
### Testing status
Manual testing according to the repro steps in the bug report.  I could not find any other uses of HelpBoxRow with message type warning elsewhere in the code base, so, contrary to the claim in the bug, I could not make a node with an error (e.g., an invalid custom function node) generate a similar warning in Node Settings.

Yamato jobs:
[PR ShaderGraph on 2021.2](https://unity-ci.cds.internal.unity3d.com/job/8771099)
N.b., the dependency failure on this was an out of date minimum version on VFX and nothing to do with this PR.

---
### Comments to reviewers


jessebarker:
Accidentally use the wrong base on creation.  Change only impacts ShaderGraph, so I've removed the reviewers from the other teams.

Nightmask3 (approved):
Approving!

PR #5682, opened 2021-09-16
Public API for URP DebugDisplaySettings

arttu-peltonen:
### Purpose of this PR
This main point of this PR is to expose public DebugDisplaySettings API. This is useful for both internal and external users, who want to create custom UI for certain debug modes. This was the exact use case for [Lookdev Studio](https://blog.unity.com/technology/try-the-new-lookdev-studio-prototype-today-0) team. This PR will bring URP to parity with HDRP, where an equivalent API is already public.

#### Public DebugDisplay API
Users can now control DebugDisplay from script. Example usage:
```
public void Update()
{
    var debugData = UniversalRenderPipelineDebugDisplaySettings.Instance;

    debugData.materialSettings.materialDebugMode = DebugMaterialMode.Albedo;
    debugData.lightingSettings.lightingDebugMode = DebugLightingMode.ShadowCascades;
    debugData.renderingSettings.wireframeMode = DebugWireframeMode.SolidWireframe;
}
```
The PR renames the now public properties to be consistent and use `camelCase` as described in [coding standards](https://ono.unity3d.com/unity-extra/unity-meta/raw/@/ReferenceSource/CSharp/Assets/CSharpReference.cs).

#### Added AlbedoDebugValidationPreset.Custom
As an incidental change, a new preset has been added to Material Albedo validation presets: `AlbedoDebugValidationPreset.Custom`. This makes sense from API point of view, and also useful from user perspective: In the images above, the custom color has been set to yellowis-orange, which matches the albedo of helmet and power tool, and gives error colors for other albedos:

image
image

---
### Testing status
Tested locally by creating a simple script ([custom-debugui-v2.zip](https://github.com/Unity-Technologies/Graphics/files/7278297/custom-debugui-v2.zip)) that uses the public API. This simulates the usecase that users of the API would have.

custom-debug-ui

This should demonstrate that the API works as intended - feel free to download the attached script and try it out.

In terms of other testing, please do a quick test for rendering debugger to see that nothing has been broken by the renaming.


com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettings.cs
```diff
@@ -15,45 +15,45 @@ public class DebugDisplaySettings : IDebugDisplaySettingsQuery
...
 
         /// <summary>
         /// Returns true if lighting is active for current state of debug settings.
         /// </summary>
-        public bool IsLightingActive => MaterialSettings.IsLightingActive &&
-        RenderingSettings.IsLightingActive &&
-        LightingSettings.IsLightingActive;
+        public bool IsLightingActive => materialSettings.IsLightingActive &&
+        renderingSettings.IsLightingActive &&
+        lightingSettings.IsLightingActive;
```
alex-vazquez, line 56: It wont be better to iterate over `m_Settings`?, in this way any time we add another `IDebugDisplaySettingsData`. We do not need to touch this properties, and methods

alex-vazquez (approved):

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettings.cs
arttu-peltonen: Fair point, I've changed it to do as you suggested.

simon-engelbrecht-soerensen (approved):
Seems to be working well, didn't discover any issues withthe rendering debugger.

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsCommon.cs
```diff
@@ -103,11 +103,13 @@ public SettingsPanel()
...
-        public bool AreAnySettingsActive => debugDisplaySettings.AreAnySettingsActive;
-        public bool IsPostProcessingAllowed => debugDisplaySettings.IsPostProcessingAllowed;
-        public bool IsLightingActive => debugDisplaySettings.IsLightingActive;
-        public bool TryGetScreenClearColor(ref Color color) => debugDisplaySettings.TryGetScreenClearColor(ref color);
+
+        // All common settings are owned by another panel, so they are treated as inactive here.
+
+        public bool AreAnySettingsActive => false;
+        public bool IsPostProcessingAllowed => true;
+        public bool IsLightingActive => true;
```
ellioman, line 111: Feels a bit inconsistent that these properties are Upper Camel Case. But I guess this was already done before.

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
```diff
@@ -134,50 +142,100 @@ struct AlbedoDebugValidationPresetData
...
+        /// Current maximum threshold value for metallic validation.
+        /// Any metallic values above this value will be considered invalid and will appear blue on screen.
+        /// </summary>
+        public float metallicMaxValue = 0.9f;
 
-        internal DebugMaterialValidationMode MaterialValidationMode;
+        /// <summary>
+        /// Current material validation mode.
+        /// </summary>
+        public DebugMaterialValidationMode materialValidationMode;
```
ellioman, line 226: Would it make sense to have these public variables as Properties?

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
```diff
@@ -358,7 +426,7 @@ public bool TryGetScreenClearColor(ref Color color)
             return false;
         }
 
-        public IDebugDisplaySettingsPanelDisposable CreatePanel()
+        IDebugDisplaySettingsPanelDisposable IDebugDisplaySettingsData.CreatePanel()
```
ellioman, line 429: Do we not need to deprecate this function if we're changing from public to private?

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsRendering.cs
```diff
@@ -310,10 +363,10 @@ public bool TryGetScreenClearColor(ref Color color)
 
                 default:
                     throw new ArgumentOutOfRangeException(nameof(color));
-            }       // End of switch.
+            }
         }
 
-        public IDebugDisplaySettingsPanelDisposable CreatePanel()
+        IDebugDisplaySettingsPanelDisposable IDebugDisplaySettingsData.CreatePanel()
```
ellioman, line 369: Same question here regarding deprecating a public function...

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
arttu-peltonen: This is not a public API change in my view, because this class was internal until now. So I don't think this needs to go through deprecation.

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsRendering.cs
arttu-peltonen: Same response here. Let me know if you agree!

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
arttu-peltonen: Thanks, I think this is a good idea. I'll update this

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsCommon.cs
arttu-peltonen: Yeah the reason here was that I didn't want to go through deprecation process just to change the first letter to be lowercase, as this is implementing an interface that was already public (even though this & other implementing classes have been internal up until now). If you'd prefer consistency here, I can do it. What do you think?

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
ellioman: Ahh I see. Just wanted to be sure 👍

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsMaterial.cs
ellioman: 👍

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsCommon.cs
ellioman: I personally think it's worth it as consistency decreases the WTF! moment users will get. 
If it's not too much trouble I would love to get that change but it can also be done in a followup PR.

ellioman (approved):

unity-cchu (approved):

lukaschod (approved):

Verasl (approved):
LGTM

PR #5685, opened 2021-09-16
Improvements to the DynamicArray class

JulienIgnace-Unity:
### Purpose of this PR
This PR adds a few utility APIs to the DynamicArray class. The goal is to provide a serviceable replacement for List when allocations are not allowed. ICollection interface allocates all over the place, in foreach, Sort etc and in some critical code that can be called every frame, we can't afford that.

---
### Testing status
Automated tests have been added for all APIs.

---
### Comments to reviewers
Added @arttu-peltonen for a Core SRP perspective on this.
Also, the QuickSort was implemented as a separate static function to avoid the requirement of IComparable for ALL dynamic arrays. Not super happy with that so if anyone has a more elegant solution I'm all for it :)

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Visual Environment ambient mode is now Dynamic by default.
 - Surface ReflectionTypeLoadExceptions in HDUtils.GetRenderPipelineMaterialList(). Without surfacing these exceptions, developers cannot act on any underlying reflection errors in the HDRP assembly.
+- Improved the DynamicArray class by adding several utility APIs.
 
 ## [12.0.0] - 2021-01-11
 


com.unity.render-pipelines.core/Runtime/Common/DynamicArray.cs
```diff
@@ -78,21 +212,88 @@ public int Add(in T value)
...
+
+                if (pivot >= 1)
+                    QuickSort(data, left, pivot);
+
+                if (pivot + 1 < right)
+                    QuickSort(data, pivot + 1, right);
+            }
+        }
+
+        public static void QuickSort<U>(DynamicArray<U> array) where U : IComparable<U>, new()
```
arttu-peltonen, line 294: You should be able to achieve better API by making this an [extension method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods). For that, it's needs to be declared in a separate static non-nested class, naming convention here would be `public static class DynamicArrayExtensions`. Signature something like this: `public static void QuickSort<T>(this DynamicArray<T> array) where T : IComparable<T>, new()`. That way you can keep the stricter type requirements for this function only, but calling it looks like just calling a normal method: `m_DynamicArray.QuickSort();`

arttu-peltonen (changes requested):
Looks great! I recommend just making the QuickSort change to keep the API consistent and make the method more discoverable via autocomplete.

com.unity.render-pipelines.core/Runtime/Common/DynamicArray.cs
JulienIgnace-Unity: I did not think of that. I knew it was a good idea to add you as a reviewer :)
Will try to implement that.

JulienIgnace-Unity: Updated the PR with the extension. Looks much cleaner now.

com.unity.render-pipelines.core/Runtime/Common/DynamicArray.cs
```diff
@@ -70,6 +81,129 @@ public int Add(in T value)
...
 
+        /// <summary>
+        /// Adds the elements of the specified collection to the end of the DynamicArray.
+        /// </summary>
+        /// <param name="array">The array whose elements should be added to the end of the DynamicArray. The array itself cannot be null, but it can contain elements that are null, if type T is a reference type.</param>
+        public void AddRange(DynamicArray<T> array)
+        {
+            Reserve(size + array.size);
+            for (int i = 0; i < array.size; ++i)
+                Add(array[i]);
```
FrancescoC-unity, line 92: Quick performance note, given that we do the reserve before, there should be no need of doing the full Add that does range check,  unless I missed something.

FrancescoC-unity (approved):
LGTM except a small perf improvement that can be done

com.unity.render-pipelines.core/Runtime/Common/DynamicArray.cs
JulienIgnace-Unity: Good point. Updated the PR.

sebastienlagarde (approved):

PR #5688, opened 2021-09-16
[HDRP] Add documentation section about editing shader at runtime

remi-chapelain:
### Purpose of this PR
Following this [forum post](https://forum.unity.com/threads/emissive-textures-generated-at-runtime-dont-apply-unless-inspected.1070798/), I propose adding this bit of information to the documentation. 
 
Even though, the recommended route to take is to use the Lit master node in shader graph instead of using HDRP/Lit materials. 
I still think users will try it (cf forum post) and thus makes sense to include it. 

---
### Comments to reviewers
Text is "draft", feel free to correct anything according to proper English grammar and documentation standards :)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/editing-shader-properties-at-runtime.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/editing-shader-properties-at-runtime.md
@@ -1,6 +1,6 @@
 # Editing shader properties at runtime
 
-Each combination of shader feature [Keyword](https://docs.unity3d.com/Manual/SL-MultipleProgramVariants.html) used in a shader produces a different shader code to compile. This is called a shader variant. Unity tries to reduce the number of shader variants to compile by only considering the sets of kewords used in a material. Until a specific shader feature is used on a material, the corresponding shader variant does not exist. 
+Each combination of shader feature [Keyword](https://docs.unity3d.com/Manual/SL-MultipleProgramVariants.html) used in a shader produces a different shader code to compile. This is called a shader variant. Unity tries to reduce the number of shader variants to compile by only considering the sets of keywords used in a material. Until a specific shader feature is used on a material, the corresponding shader variant does not exist. 
 
 As a result, if one tries to edit a material at runtime and change properties that affects which keywords is enabled, Unity is not able to find the shader variant and the rendering of this shader feature fails. 
 This is because the correct keywords are not enabled in the shader. 


JMargevics (approved):
Topic and the addition make perfect sense. Approving ✔👍

com.unity.render-pipelines.high-definition/Documentation~/editing-shader-properties-at-runtime.md
```diff
@@ -0,0 +1,11 @@
+# Editing shader properties at runtime
+
+For High Definition Render Pipeline (HDRP) non-Shader Graph shaders, such as the [Lit](Lit-Shader.md), [Unlit](Unlit-Shader.md) shaders, changing a property at runtime, in some cases, does not have any effect. This is because if the property is not enabled before runtime, the specific shader variant associated with this property is not included in the shader (and is also not included when bulding a player). To include the shader variant, the proper keywords need to be enabled on the material before editing.
```
sebastienlagarde, line 3: There is various incorrect statement here. Probe exist also for shader graph.

Here is a proposal:

Each shader feature (Keyword) use in a shader (link to the doc) produce a different shader code to compile. Each of this shader code is call a shader variant. Unity try to reduce the number of shader variant to compile by only considering the sets of kewords use in a Material. Other shader variant don't exist. When building a Player, it is the union of the all shader variant based on keyword used that is compiled.

If you try to create a Material at runtime and change proerties which affect which keywords is enabled. In case this combination was not existing when building the player, this shader variant will not exist and the rendering will fail.
This can happen if you use for example the HDRP/Lit shader and try to setup a iridecence material but don't have any ididescence properties enabled in an already existing Material.


^ So this is not good but this is what I think we should express. We should be clear that 1) This is for Unity in general, not only HDRP. Builtin have same issue. And that this is also for shader graph if you add keyword in shader graph.
Then the goal is to explain that users must enable the right keyword at runtime and that this shader variant need to exist. 

this statement can be share with URP documentation.

remi-chapelain: Thanks, will update with a new proposal with this info !

remi-chapelain:
Updated the version based on what has been discussed with @sebastienlagarde

com.unity.render-pipelines.high-definition/Documentation~/editing-shader-properties-at-runtime.md
```diff
@@ -0,0 +1,20 @@
+# Editing shader properties at runtime
+
+Each combination of shader feature [Keyword](https://docs.unity3d.com/Manual/SL-MultipleProgramVariants.html) used in a shader produces a different shader code to compile. This is called a shader variant. Unity tries to reduce the number of shader variants to compile by only considering the sets of kewords used in a material. Until a specific shader feature is used on a material, the corresponding shader variant does not exist.
```
sebastienlagarde, line 3: typo: kewords  -> keywords

sebastienlagarde (approved):

Vic-Cooper:
This is large enough to need editorial review so I'm working on this in [a google doc.](https://docs.google.com/document/d/19kAQ0sI5tZStSlAo0rpcnrZZuqe3F2NirYZKSUk1R6c/edit#heading=h.xi84r4kar5n3). 

@remi-chapelain I'll invite you for a tech review when I've made changes :).

Vic-Cooper:
@sebastienlagarde This doc has been absorbed into the shader docs. See the [draft google doc](https://docs.google.com/document/d/19kAQ0sI5tZStSlAo0rpcnrZZuqe3F2NirYZKSUk1R6c/edit#heading=h.xi84r4kar5n3) for input from Kerry. Please close and do not merge this PR.

remi-chapelain:
Closing this PR because the information have been included in this [more recent one](https://github.com/Unity-Technologies/Graphics/pull/6376).

PR #5689, opened 2021-09-16
fixed render settings setting incorrect cascade value, causing the wh…

Verasl:
### Purpose of this PR
This PR fixes a stall in the render settings converter when the user has no cascades checked in the Graphics tiers.

---
### Testing status
Ran the converter on the Kart Racing learn project, going from builtin to URP with the disabled cascades

---
### Comments to reviewers


martint-unity (approved):

martint-unity:
Merged it since we do not have any tests to run for this converter framework.

PR #5693, opened 2021-09-16
Fixed specular anti aliasing for layeredlit shader

adrien-de-tocqueville:
### Purpose of this PR

Noticed that layeredlit was missing the shader feature for specular anti aliasing. It was removed during the keyword refactor (git blame gives this PR #2996)

### Testing status

verified that it doesn't work on master.
verified that it works with this PR


sebastienlagarde (approved):
great catch, thanks for the fix!

PR #5694, opened 2021-09-16
Fixed a warning when enabling Tile/Cluster debug.

JulienIgnace-Unity:
### Purpose of this PR
With current version, enabling Tile/Cluster debug in the Lighting debug panel caused a warning to be displayed.
This PR fixes this.

The root cause was the use of the wrong overload of DrawFullScreen caused by an unnamed parameter being wrongly interpreted as a RenderTargetIdentifier. Making the parameter explicit fixes the issue.

---
### Testing status
Tested that the warning does not appear anymore and the debug still worked.


sebastienlagarde (approved):

JMargevics (approved):
Testing done is sufficient, approving 👍✔

PR #5698, opened 2021-09-16
PPV2 - Fix VR/XR detection when using XR plugins

maloyer-unity:
This detection was only working at run-time or with a pre-2020.1 Editor... So any recent Editor that fully rely on XR system will not work.
With XR extensions, we can query the XR Management system to see if there is any active XR extension.

---
### Purpose of this PR
Fix for : https://fogbugz.unity3d.com/f/cases/1302489 

Warnings were not showing up when XR plugins are used (was working with old VR module).
 * Lens distortion
 * Motion blur 
 * Bloom's "dirt texture"

This PR fix the detection of VR/XR in the editor (run-time detection was already working).

---
### Testing status
Checked that the warnings appear when XR is active and have a XR plugin selected, if no XR plugin is selected or XR manager is not installed, the warnings will not show.

### DIFF
--- a/com.unity.postprocessing/PostProcessing/Editor/Utils/EditorUtilities.cs
+++ b/com.unity.postprocessing/PostProcessing/Editor/Utils/EditorUtilities.cs
@@ -370,12 +370,14 @@ namespace UnityEditor.Rendering.PostProcessing
             get
             {
 #if ENABLE_VR_MODULE && ENABLE_VR
-#if !UNITY_2020_1_OR_NEWER
-                return UnityEditorInternal.VR.VREditor.GetVREnabledOnTargetGroup(BuildPipeline.GetBuildTargetGroup(EditorUserBuildSettings.activeBuildTarget));
-#elif UNITY_2020_1_OR_NEWER && ENABLE_XR_MODULE && XR_MANAGEMENT_4_0_1_OR_NEWER
+#if ENABLE_XR_MODULE && XR_MANAGEMENT_4_0_1_OR_NEWER
+                // If XR manager extension is available, we can query it to know if any XR extension is currently active
                 var buildTargetSettings = XRGeneralSettingsPerBuildTarget.XRGeneralSettingsForBuildTarget(BuildTargetGroup.Standalone);
                 return (buildTargetSettings != null && buildTargetSettings.AssignedSettings != null &&
                         buildTargetSettings.AssignedSettings.activeLoaders.Count > 0);
+#elif !UNITY_2020_1_OR_NEWER
+                // This will only work with 2019.3 and older since it rely on the old VR module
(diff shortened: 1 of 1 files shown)


maloyer-unity:
Not sure if the failed tests are valid or if the test even work at all... My code does not change any rendering features, only Editor code was changed.

fcoulombe (approved):
LGTM

com.unity.postprocessing/PostProcessing/Editor/Unity.Postprocessing.Editor.asmdef
```diff
@@ -1,9 +1,11 @@
 {
     "name": "Unity.Postprocessing.Editor",
+    "rootNamespace": "",
     "references": [
-        "Unity.Postprocessing.Runtime"
+        "Unity.Postprocessing.Runtime",
+        "Unity.XR.Management",
```
fabien-unity, line 6: Do we need these new deps here ? Can we avoid them ?

maloyer-unity: I'll double check if we can avoid new dependency.

mikesfbay:
@maloyer-unity Have you tested this with 2018.4 LTS or 2019.4 LTS using old VR module? How far back will this be backported?
We need to add QA to the loop to test with both 19.4 and 2020+.

com.unity.postprocessing/PostProcessing/Editor/Unity.Postprocessing.Editor.asmdef
maloyer-unity: We need these references, otherwise the code that query the XR Management module can't work.

But it does not add a strict dependency to XR Management extension, it will only use it if the package is added to the project by the user (detected in code via the preproc `#if ENABLE_XR_MODULE && XR_MANAGEMENT_4_0_1_OR_NEWER`), and if its not there, the part of the code that require it is disabled.

maloyer-unity:
@mikesfbay, I re-tested the fix with 2019.4.29f1, 2020.3.24f1 and 2022.1.0a17(trunk).
On 2020.x and higher the code will only try to use the XR Management package if the package is installed by the user.
On 2019.4 it will use XR Management package if the package is installed by the user, otherwise it will try via the old VR extension (both works as expected now).
I could not test with 2018.x, but it will use the same path as 2019.4 with the old VR extension code (which was the code used before my change, so I expect it will keep on working)

mikesfbay:
@maloyer-unity Sounds good, looping in QA @ryanhy-unity on postfx v2.

ryanhy-unity (approved):
Changes are editor only and seem pretty low risk and isolated to post processing.  Verified fix with user project and saw the proper errors displayed after updating to package with these changes.

mikesfbay (approved):

mikesfbay:
@maloyer-unity Fyi, we should let Sebastian's team do the final merge to master since he has ownership of the package and is making the releases.

maloyer-unity:
@mikesfbay, good point, I'll do that next time I have changes for this package

PR #5700, opened 2021-09-16
[ShaderGraph][2022.1] Fix for 1354463 - subgraph node inputs should match property list on subgraph

jessebarker:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1354463/

When the blackboard categories were added, the ordering of properties and subgraph node inputs got decoupled.  So, if a user wants to reorder the properties on the subgraph, a shader graph using that subgraph as a node would see a different ordering.  This PR makes it so that the property ordering is preserved (both on the input to the node, and in the generated code).

---
### Testing status
Manual testing according to the repro case in the bug report.  Also, the same basic steps, but with more properties.

Just in case this goes away from "checks"
[PR ShaderGraph on trunk](https://unity-ci.cds.internal.unity3d.com/job/8825214l)
---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Importers/ShaderSubGraphImporter.cs
+++ b/com.unity.shadergraph/Editor/Importers/ShaderSubGraphImporter.cs
@@ -19,7 +19,7 @@ using UnityEngine.Pool;
 namespace UnityEditor.ShaderGraph
 {
     [ExcludeFromPreset]
-    [ScriptedImporter(28, Extension, -905)]
+    [ScriptedImporter(29, Extension, -905)]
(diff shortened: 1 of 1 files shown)


Nightmask3 (approved):
looks good to me!

cdxntchou (approved):

sebastienlagarde:
hey, the PR is tagged as need backport for 21.2 but have title ruby, expected?

jessebarker:
> hey, the PR is tagged as need backport for 21.2 but have title ruby, expected?

Sorry, forgot to update the tag to 'backported-2021.2'.  Just did that.  Backport PR is #5716

PR #5340, opened 2021-08-12
Rework toolbar design

julienamsellem:
### Purpose of this PR
Jira
https://jira.unity3d.com/browse/GVFXG-143
Design docs
https://docs.google.com/document/d/1oG7-tKzlEDqQCRnw0KVsRqldVUyTY5gBfVuF-5tPx2Q/edit
https://docs.google.com/document/d/1VS4LKjZmfx40kutvb39h3sIL6UCSM9m92IxuluxGTeI/edit#

Before
old-toolbar
After
new-toolbar

---
### Testing status
Tested on dark and light themes with 100% and 150% DPI screens

---
### Comments to reviewers
**Known bug**: when DPI is different than 100% the drop down popups can be position at a "not ideal" position.
The bug is most probably on UIToolkit side. I warned the [#dev-uitoolkit](https://unity.slack.com/archives/C3414V4UV/p1628498431112200) channel about it

PR #5705, opened 2021-09-17
Revert "Rework toolbar design"

julienamsellem:
Reverts Unity-Technologies/Graphics#5340


VitaSkruibyte (approved):
Autoapproving, as we need to revert accidentally merged PR

PR #5707, opened 2021-09-17
[HDRP] Adding a transparency setup samples 

remi-chapelain:
### Purpose of this PR
The idea behind theses transparency setup samples is to share to the users the knowledge in [that document](https://docs.google.com/document/d/1yftE_RM-L7RG05K7Bz5xnWIgnDfxqrAH0lE-eprkNFc/). 

In HDRP, different transparent setups are needed depending on which pipeline we use: Raster, Path Traced or Raster and Ray Tracing. Each of theses setup are not trivial nor intuitive to users. 

image

For now, this sample only includes the scenes about shadows. Each part contains 3 scenes, one raster, one path traced, one mixing raster and ray tracing with details and instructions on the tiles. 
It would be interesting to add some of theses infos (what's on the tiles) to the docs too and a link how to get the samples. 

2cf657a1215438712183267df3548ee0

### What will be added in later PRs

Informations and examples about:
- Stacking
- Sorting
- Refraction models
- Transparent pre-pass
- Distortion
- Shader graph implementation.  

**WIP**
wip_stacking

wip-refraction

---
### Testing status
- Tried to import the sample without and with other samples ✔️
- Imported the sample in a fresh new HDRP project ✔️ 
- Opened the 3 scenes independently ✔️ 

---
### Comments to reviewers
Added @Vic-Cooper to have feedback on documentation and text that is inside the scene. Also I would like to know if you think it's interesting to include some of those bit in the documentation itself for redundancy. 
For simplicity purpose, here's a [gdoc](https://docs.google.com/document/d/1l2d739WxW8hmz4liTjZWDDgbn9Vjp2TMxFH6zbJZJvo/edit?usp=sharing) with all the text.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Samples~/TransparencySamples/Prefabs/Shadows.prefab
+++ b/com.unity.render-pipelines.high-definition/Samples~/TransparencySamples/Prefabs/Shadows.prefab
@@ -5951,7 +5951,7 @@ MonoBehaviour:
   m_OnCullStateChanged:
     m_PersistentCalls:
       m_Calls: []
-  m_text: Light types
+  m_text: A. Light types
   m_isRightToLeft: 0
   m_fontAsset: {fileID: 11400000, guid: 629ab6f6948c603418a378b3e3210d77, type: 2}
   m_sharedMaterial: {fileID: 2180264, guid: 629ab6f6948c603418a378b3e3210d77, type: 2}


iM0ve:
Not sure what Im doing wrong but getting a ton of warnings and almost no text is rendering. 

image

image

JMargevics:
https://user-images.githubusercontent.com/50581102/134149551-e27dd652-027a-40a6-9fb8-d162dc363ae8.mp4

Yes looks like TMP fonts and shader got dropped. Also weird that only one mobile shader is packaged with this. Could be a bug with TMP not playing nicely with imported packages, will try to repro 👍

JMargevics:
This looks like a great addition however I have one issue with the current content: it's very shadow casting oriented but the sample pack is called "Transparency Setup" implying there will be more general cases :) 

Some things I'd like to see:

- Practical examples on how to setup sorting and it's limitations. Like usage of transparent depth prepass or show that we can't sort refractive surfaces
- Practical examples of Transparent unlit with distortion enabled. I think these deserve a lot more love :) 
- Practical examples of using different refraction models. Sphere model is still an enigma for me and it never looks right (in your example too)
image

I understand that transparency is a massive subject and adding all these will result in a way too big sample pack especially taking DXR variations into consideration but pretty sure there can be a middle ground 😊

AlixMi:
Hello I have too warnings about font not being found, making text not appear. It's great that you are doing this thank you. Indeed the sphere refraction model isn't looking too good, I noticed that there is some discrepancies between the behavior of sphere and box model with their behavior around thickness, this is stuff I hope we will tackle with transparency improvements work. As I said before to you I will gladly contribute to this transparency sample by giving models and/or shader.

remi-chapelain:
Quick update: 
- Rebased master
- Fixed Text Mesh Pro issues (was using the wrong TMP material, the one in material samples)
- Fixed a few typos
- Issue with sphere refraction model disappeared
- Added documentation about the sample itself (stayed vague to not promised things that are not there yet)
- After documentation review, should be ready to merge and then iterations will continue on the rest of work in progress.

PR #5710, opened 2021-09-17
[Universal][converter][fix] Reloading previously loaded scene

martint-unity:
---
### Purpose of this PR
This fix, makes sure that the user saves the current open scene. This is so that things works as it should. The user might have things in the current open scene that the converters might need to know about. So having the user first save the current scene is the right thing to do.

---
### Testing status
Testing this locally since we do don't have tests.

---
### Comments to reviewers
Have a scene that is dirty (changes in it) then run the converter 
Windows->Rendering->Render Pipelines Converter
Initialize or /and run converters

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
@@ -471,20 +471,7 @@ namespace UnityEditor.Rendering.Universal.Converters
 
         void InitializeAllActiveConverters(ClickEvent evt)
         {
-            Scene currentScene = SceneManager.GetActiveScene();
-            if (currentScene.isDirty)
-            {
-                if (EditorUtility.DisplayDialog("Scene is not saved.",
-                    "Current scene is not saved. Please save the scene before continuing.", "Save and Continue",
-                    "Cancel"))
-                {
-                    EditorSceneManager.SaveScene(currentScene);
-                }
-                else
-                {
-                    return;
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
```diff
@@ -469,6 +471,21 @@ void CheckAllConvertersCompleted()
...
+                    "Cancel"))
+                {
+                    EditorSceneManager.SaveScene(currentScene);
+                }
+                else
+                {
+                    return;
+                }
+            }
+
```
ellioman, line 488: Can we extract this in to a function that takes in the Scene, since it's the same code in two places?

com.unity.render-pipelines.universal/Editor/Converter/RenderPipelineConvertersEditor.cs
```diff
@@ -614,6 +631,23 @@ void UpdateInfo(int stateIndex, RunItemContext ctx)
 
         void Convert(ClickEvent evt)
         {
+            // Ask to save save the current open scene and after the conversion is done reload the same scene.
+            Scene currentScene = SceneManager.GetActiveScene();
+            string currentScenePath = currentScene.path;
```
ellioman, line 636: Seems like getting the path is only used in lines 694 and 696. 
If you extract this scene dirty check in to a function, then you can move the line for the path down there.

ellioman (approved):

PR #5713, opened 2021-09-17
Make Android tests use windows images to build android player

tomzig16:
### Purpose of this PR
This PR makes android tests to use windows image to build the android player. Reason for that being Unity using old SDK from android image, which at the moment throws an error. We download SDK together with Unity, so there's no need to use that image. However, we still use the same android image for running tests as it is the easiest way to download ADB and other needed tools.

---
### Testing status
ABV
Manually built project on Yamato with this image.

---
### Comments to reviewers
None


kivistikliis (approved):
LGTM, thanks!

PR #5715, opened 2021-09-17
Vfx/fix/1289612 filter texture by dimension

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1289612/

When an operator, a node or even a property exposes a Texture object, the picker lists all kind of textures (2D, 3D, Cubemap ...).
The list is then much longer than needed and the user is bothered with non relevant textures.
With the fix, the picker lists only relevant textures.

[Before]
Unity_GvXsSUPpfD

[After]
Unity_qZPKnwpnhu

---
### Testing status
The change will apply to any object field in the VFX graph editor, so I checked
- The blackboard
- Operator (Sample SDF)
- Nodes (Position Mesh and Output Quad main texture)

---
### Comments to reviewers
The scope of the change is quite large because not only texture picker is affected, any kind of object picker (mesh, skinmesh, texture ...).
The UI on this part is completely new, even though I tried to make it the same as before so you may also check for regression on this part.

**Note**
The object icon is a bit improved since now it's more accurate than before. For example if your object should be a cubemap, you see a cubemap icon, not the generic texture icon.
[Before]
Unity_hSfvG5kTSz

[After]
Unity_XWzxDs2bkV


julienamsellem:
I forgot one aspect of the issue: the texture picker should also include RenderTextures objects (if possible of the correct dimension).
I'll add more commits to support this feature.

VitaSkruibyte:
I am happier than a seagull with a French fry 🍟 when I see such detailed and informative PR descriptions, thank you! Keep it up! 

Tested:
1. VFX Graph, Subgraph Operator, Blackboard, Inspector
2. Sample Attribute Map, Sample Texture Cube, Sample Texture 3D, Sample Texture 2D, Load Texture 2D Array, Load Texture 3D, Load Texture 2d, Sample Multiple Position Map, Sample Signed Distance Field, Cubemap, Cube map array nodes, Shader Graph, Main texture inputs in Output context
3. .jpg, .png, .tga, .exr formats
4. Drag texture to the texture slot, assign texture via object picker
5. Setting texture to none, deleting it from project browser, undo/redo, copy/paste
6. Tried tweaking Texture type and shape settings

**Issues found:**

1. Texture can not be dragged to texture slot (not reproducible on master), (Does not work with inline nodes, Blackboard properties, works with Inspector)
uTlGUzRMWt


2. When the texture's shape changes, the texture remains assigned to the field instead of being removed. For example: Changing textures shape from 2D to Cubemap.
image

3. As you already mentioned, RT support is missing

iTris666:
Have you tested that a render texture with the right dimension should also be regarded as compatible ?

VladNeykov:
Hi @julienamsellem ,
Really useful fix! I've jotted down some observations (WIP) in the [test doc](https://docs.google.com/document/d/1wHv4xj8yQYc2Oc0iZrDYudztP6iLqHWt2eEoD-uWOjs/edit?usp=sharing)

VladNeykov (approved):
Looking good, no outstanding issues found.
[Test doc](https://docs.google.com/document/d/1wHv4xj8yQYc2Oc0iZrDYudztP6iLqHWt2eEoD-uWOjs/edit) for reference.
Thanks for the fix! :)

PR #5727, opened 2021-09-20
Fixed error in SSGI when disabling decals (case 1365521).

anisunity:
The SSGI was using a texture it had already bound, but under and other identifier. Already fixed in master and more recent versions


iM0ve (changes requested):
Hello, 
If this requires qa to verify - please describe what testing was already made or any tips what to test.
Or if its only for awareness - let us know in the description.
Thank you

sebastienlagarde (approved):
>@AnisB  does it only happen on 10.x? do we need to forward port on master?
answer yes

sebastienlagarde:
Note: this will be for 10.8 package (so in a month)

anisunity:
Only for QA awareness.

iM0ve (approved):
Thanks

PR #5731, opened 2021-09-20
[APV] Cell streaming system

JulienIgnace-Unity:
### Purpose of this PR
This PR introduces cell streaming to the APV.

This includes:
- Refactor to split properly loading cell from the asset and uploading data to GPU
- Renamed things a bit: Add for adding a cell from asset and Load to upload it to GPU
- Handled out of memory (both index and texture) more gracefully: Cell isn't uploaded but still available in CPU memory
- Cell debug is now red for cells that aren't "loaded" (uploaded to GPU)
- Precompute of cell chunk cost
- Added name of 3D textures allocated for APV
- First version of streaming of cell. Only based on distance to camera for now.
- Debug option to "freeze" streaming in order to assess the current state
- Removed all GCAlloc for load/unload code as it is now called regularly.
- Disabled by default. Added a debug entry to enable it.
- Actual data upload is done in "chunks" to avoid allocating for worse case scenario. Currently set at 8 chunks (500KB)
- Index GPU Buffer update is now done partially to improve performance.

Still needs to be done:
- Handle multiple cameras
- Handle fragmentation
- Base streaming on more than just camera position.

What has been tested:
- Automated tests are green
- Baking one/Multiple scenes
- Tested add/remove of scenes after baking
- Tested that streaming actually worked while under budget (with the help of cell debug)
- Tested that being under budget is handled gracefully with streaming disabled (no error, just stops loading after being oom)


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -305,13 +306,14 @@ internal static void PerformDilation()
                 for (int iterations = 0; iterations < dilationSettings.dilationIterations; ++iterations)
                 {
                     // Make sure all is loaded before performing dilation.
-                    ProbeReferenceVolume.instance.PerformPendingOperations(loadAllCells: true);
+                    ProbeReferenceVolume.instance.PerformPendingOperations();
```
FrancescoC-unity, line 309: We need to make sure all cells are loaded in before performing dilation. 

If they don't all fit in memory we should split in multiple dilation passes, but data needs to be forced in, otherwise it won't work.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -479,7 +481,10 @@ static void OnAdditionalProbesBakeCompleted()
                     cell.validity[i] = validity[j];
                 }
 
-                probeRefVolume.cells[cell.index] = cell;
+                cell.indexChunkCount = probeRefVolume.GetNumberOfBricksAtSubdiv(cell, out var minValidLocalIdxAtMaxRes, out var sizeOfValidIndicesAtMaxRes);
```
FrancescoC-unity, line 484: This has the side effect of assigning the chunks in the index, we call this during LoadCell, so calling this here like this will allocate indices when it really shouldn't

FrancescoC-unity: If we need this here, we need to call a version without the side-effect.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
```diff
@@ -49,26 +51,48 @@ struct ReservedBrick
...
+        ObjectPool<VoxelMeta> m_VoxelMetaPool = new ObjectPool<VoxelMeta>(x => x.Clear(), null, false);
 
         int m_VoxelSubdivLevel = 3;
 
         bool m_NeedUpdateIndexComputeBuffer;
+        int m_UpdateMinIndex = int.MaxValue;
+        int m_UpdateMaxIndex = int.MinValue;
+
+        // Static variable required to avoid allocations inside lambda functions
+        static Cell g_Cell = null;
```
FrancescoC-unity, line 95: yikes, can we use the cell index instead to do the checks?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
```diff
@@ -111,10 +136,18 @@ internal ProbeBrickIndex(ProbeVolumeTextureMemoryBudget memoryBudget)
...
+            return m_AvailableChunkCount;
+        }
+
         internal void UploadIndexData()
         {
-            m_PhysicalIndexBuffer.SetData(m_PhysicalIndexBufferData);
+            var count = m_UpdateMaxIndex - m_UpdateMinIndex + 1;
+            m_PhysicalIndexBuffer.SetData(m_PhysicalIndexBufferData, m_UpdateMinIndex, m_UpdateMinIndex, count);
             m_NeedUpdateIndexComputeBuffer = false;
+            m_UpdateMaxIndex = int.MinValue;
```
FrancescoC-unity, line 149: This looks dangerous to me, we should set this to default values like this straight before we start computing the new valid indices or if not possible at least do some range check before actually setting data to avoid some weird numerical explosion (in count)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
```diff
@@ -7,6 +7,8 @@ namespace UnityEngine.Experimental.Rendering
 {
     internal class ProbeBrickPool
     {
+        const int kProbeIndexPoolAllocationSize = 128;
```
FrancescoC-unity, line 10: The word Index here is _very_ confusing, this is just the Pool. 

I'd just call this kProbePoolChunkSize

FrancescoC-unity: (I know was there before, but it was bad :P , might as well change it )

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
JulienIgnace-Unity: Good point. I'll take a look on how to do that properly. Did not think of that.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -1163,75 +1093,93 @@ internal void Clear()
...
+            int brickChunksCount = (bricks.Count + chunkSize - 1) / chunkSize;
+            cellInfo.chunkList.Clear();
+
+            // Try to allocate texture space
+            if (!m_Pool.Allocate(brickChunksCount, cellInfo.chunkList))
+                return false;
+
+            // In order not to pre-allocate for the worse case, we update the texture by smaller chunks with a preallocated DataLoc
+            int chunkIndex = 0;
+            while (chunkIndex < cellInfo.chunkList.Count)
```
FrancescoC-unity, line 1114: This looks correct, but is delicate, have you stress tested this a bit to see if all looks fine?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -295,37 +295,41 @@ internal class Cell
...
-            internal string sourceAsset;
-            internal Cell cell;
-            internal float distanceToCamera = 0;
-            internal Vector3 position;
+            public Cell cell;
+            public List<Chunk> chunkList = new List<Chunk>();
+            public int flatIdxInCellIndices = -1;
+            public bool loaded = false;
+            public ProbeBrickIndex.CellIndexUpdateInfo updateInfo;
+            public int sourceAssetInstanceID;
```
FrancescoC-unity, line 310: Any reason for this change? It was very useful to use a string as a ID to debug

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -691,36 +660,65 @@ public int GetVideoMemoryCost()
...
-                if (m_ChunkInfo.ContainsKey(cell.index))
-                    m_ChunkInfo.Remove(cell.index);
+                if (cellInfo.loaded)
+                    m_LoadedCells.Remove(cellInfo);
+                else
+                    m_UnloadedCells.Remove(cellInfo);
 
-                if (cell.flatIdxInCellIndices >= 0)
-                    m_CellIndices.MarkCellAsUnloaded(cell.flatIdxInCellIndices);
+                UnloadCell(cellInfo);
```
FrancescoC-unity, line 672: Can we move this in the if (cellInfo.loaded) side of the branch? Would read more logically I think

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -691,36 +660,65 @@ public int GetVideoMemoryCost()
...
         }
 
-        void AddCell(Cell cell, List<Chunk> chunks)
+        internal void AddCell(Cell cell, int assetInstanceID)
         {
-            cell.loaded = true;
-            cells[cell.index] = cell;
+            // The same cell can exist in more than one asset
+            // Need to check existence because we don't want to add cells more than once to streaming structures
+            // TODO: Check perf if relevant?
```
FrancescoC-unity, line 699: Is it? :D

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
JulienIgnace-Unity: Hm, I don't see the side effect you are talking about. Aren't you talking about GetCellIndexUpdate ?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -691,36 +660,65 @@ public int GetVideoMemoryCost()
...
+            if (!cells.ContainsKey(cell.index))
+            {
+                var cellInfo = m_CellInfoPool.Get();
+                cellInfo.cell = cell;
+                cellInfo.flatIdxInCellIndices = m_CellIndices.GetFlatIdxForCell(cell.position);
+                cellInfo.sourceAssetInstanceID = assetInstanceID;
+                cells[cell.index] = cellInfo;
+
+                m_UnloadedCells.Add(cellInfo);
```
FrancescoC-unity, line 708: m_UnloadedCells makes it sound to me that the cell has been unloaded, not that it is not loaded (subtle difference, but I guess a tad confusing like this)

FrancescoC-unity: Pending loading maybe?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
JulienIgnace-Unity: We can but I don't think that would make any difference, would it? It's just comparing references.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
FrancescoC-unity: Yes, I need to drink my coffee :D The VS find failed me

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
FrancescoC-unity: would we still need to have the global static thing? I guess?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
JulienIgnace-Unity: I don't think we can reinit them just before computing new valid indices because this happens every time we add/remove bricks, so we would not be accumulating the right range of indices.
I can add range check before sending data though. Not sure if it should be an assert or a no-op+error message in case of error though... (in all cases, result will be broken)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
JulienIgnace-Unity: I agree with the change. Will do.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
FrancescoC-unity: I think we should assert here as it means we fucked something up really bad :D

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
```diff
@@ -272,38 +315,55 @@ public static DataLocation CreateDataLocation(int numProbes, bool compressed, Pr
...
-
-
-            if (bands == ProbeVolumeSHBands.SphericalHarmonicsL2)
-            {
-                L2_0_locData = new Color[loc.width * loc.height * loc.depth];
-                L2_1_locData = new Color[loc.width * loc.height * loc.depth];
-                L2_2_locData = new Color[loc.width * loc.height * loc.depth];
-                L2_3_locData = new Color[loc.width * loc.height * loc.depth];
-            }
+            ValidateTemporaryBuffers(loc, bands);
```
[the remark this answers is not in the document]
JulienIgnace-Unity: I'm not sure I'm following here. Regardless of the temp buffer size we always copy the right amount.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: Perf reason. I need to check that everytime an asset is streamed in/out to remove existing cells from internal arrays. Comparing strings did not feel good at all.
I might be overthinking this though, for now we don't even have asset streaming so...

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: Agreed. The little bit of code above was added later so I did not think of that.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: Kind of the same as above. This is relevant in term of perf when we have asset streaming. Right now it is expected that this happens during loading only so it's probably fine as is.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: I agree but I'm not a fan of "pending" either. It's just a bunch of cells that are not loaded _right now_. It does not mean that all of them are destined to be loaded so pending feels wrong as well.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
JulienIgnace-Unity: I tested as much as I could but it could be interesting to test with bigger/more complex scenes as well. As it is, it's hard to force streaming activity in a relevant way :/

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickIndex.cs
JulienIgnace-Unity: Yeah, the static is mandatory because we use lambda for finding stuff. So it would capture the variable and generate gcalloc without it.
I guess we can make functors and set the ref in it to avoid this but it's much more verbose for not a lot of gain (I don't expect this to be run multi-threaded)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
FrancescoC-unity: ToBeLoaded like before?

PR #5733, opened 2021-09-20
UI fix for emissive

adrien-de-tocqueville:
### Purpose of this PR

Clamping the emissive intensity was introduced in a previous PR but didn't work because of a swap between log2 and log10
Also, made the unit picker smaller because it's anormally large

emission

---
### Testing status

launched yamato and verified clamping works


sebastienlagarde (approved):

sebastienlagarde:
@adrien-de-tocqueville can you retrieve the PR, guess it have been backport up to 10.x?

JMargevics:
https://user-images.githubusercontent.com/50581102/134169906-2f8d88aa-6f89-4805-a200-3429ce3c4b3b.mp4

Question: should we clamp just the input value or the final result after Exposure Weight contribution? It breaks the scene lighting in a similar manner before previous clamping PR.

adrien-de-tocqueville:
Preventing nans is pretty much impossible with just the clamp in the slider because there are too many ways to exceed max float value. But having such high intensity should never happen so i don't think it's a problem, the goal here is to not have _infinity_ in the field

JMargevics (approved):
Clamping works fine, as expected ✔👍

PR #5751, opened 2021-09-21
Fixed Realtime lightmap not working correctly in player with various Lit shader

sebastienlagarde:
### Purpose of this PR
Fix issue https://fogbugz.unity3d.com/f/cases/1360021/

the root cause of the issue is that with various Lit shader there is no uv2 defined. uv2 are defined only if DYNAMICLIGHTMAP_ON is defined for vertex shader.

In our lit shader we have this:
            #pragma multi_compile_fragment _ DIRLIGHTMAP_COMBINED
            #pragma multi_compile _ DYNAMICLIGHTMAP_ON

which mean

For vertex shader: nothing is defined
For fragment: DIRLIGHTMAP_COMBINED and DYNAMICLIGHTMAP_ON are defined.

Nothnig is define for vertex shader as there is a hardcoded behavior about not defining DYNAMICLIGHTMAP_ON if DIRLIGHTMAP_COMBINED  is defined.

Whether we want to change the C++ behavior or ont is a separte discussion. this PR simply remove the _fragment frequency for DIRLIGHTMAP_COMBINED so  it fix the scene provided.

Note that the issue don't happen with ShaderGraph as we always defined UV2
---
### Testing status

Tested the repro project. Don't think there is any QA pass to do here more than that.

---
### Comments to reviewers


FrancescoC-unity (approved):

JulienIgnace-Unity (approved):

iM0ve:
Okay, makes sense. Testing the repro should be enough.

PR #5752, opened 2021-09-21
Render Pipeline Converter docs.

oleks-k:
Render Pipeline Converter docs.

### DIFF
--- a/com.unity.render-pipelines.universal/Documentation~/upgrade-guide-2021-2.md
+++ b/com.unity.render-pipelines.universal/Documentation~/upgrade-guide-2021-2.md
@@ -10,7 +10,9 @@ For information on converting assets made for a Built-in Render Pipeline project
 
 * The Universal Renderer asset contains the property **Rendering Path** that lets you select the Forward or the Deferred Rendering Path.
 
-* The method `ClearFlag.Depth` does not implicitely clear the Stencil buffer anymore. Use the new method `ClearFlag.Stencil`.
+* The method `ClearFlag.Depth` does not implicitly clear the Stencil buffer anymore. Use the new method `ClearFlag.Stencil`.
+
+* URP 12 implements the [Render Pipeline Converter](features/rp-converter.md) feature. This feature replaces the asset upgrade functions that were previously available at **Edit > Render Pipeline > Universal Render Pipeline > Upgrade...**
 
 ## Upgrading from URP 10.0.x–10.2.x
 


martint-unity (changes requested):
me and @TheoWong-pixel went through this together and we have some feedback

TheoWong-pixel:
**For rp-converter.md:**

**Line 43:** 
Yellow icon: the conversion went with a warning, you might need to do some action related to the element.
Red icon: the conversion failed.

- Yellow icon: The conversion may not succeed, there may be some manual action required. (This state is only available pre-conversion).
- Red icon: the conversion failed. The original file was not changed. 

**Line 47**
6. Close the Render Pipeline Converter window. If the conversion went with warnings, inspect the elements with the warnings.

- When you close the window now, nothing is saved so we might want to advise them to check out their conversions with the converter window still open, so to keep reference. 

**Line 89**
Upgrades assets from earlier URP versions to the current URP version.

- this is false, because we dont have any URP upgraders in this convertor. 

**Line 107**
This converter indexes the project and creates the temporary `.index`, which might take a significant time.

- "...creates a temporary .index file. This might take a significant time."

**For upgrading-your-shaders.md:**
Line 20/19: The table below shows which URP shaders the Unity built-in shaders convert to when you run the shader upgrader.

- Sound be material upgrader, we dont have a shader upgrader.

oleks-k:
Thanks for the review, I've updated the descriptions, see the new commit.

com.unity.render-pipelines.universal/Documentation~/features/rp-converter.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Upgrade 2D URP Assets
+
+This conversion type upgrades assets of a 2D project from an earlier URP version to the current URP version.
+
+Available converters:
+
+* **Parametric to Freeform Light Upgrade**
+
+    This converter converts all parametric lights to freeform lights.
```
phi-lira, line 77: @martint-unity @unity-cchu question, should URP to URP upgrade be handled in this converter?

com.unity.render-pipelines.universal/Documentation~/TableOfContents.md
```diff
@@ -18,6 +18,7 @@
     * [Decal](renderer-feature-decal.md)
       * [Decal Shader Graph](decal-shader.md)
 * [Upgrade guides](upgrade-guides.md)
+  * [Render Pipeline Converter](features/rp-converter.md)
   * [Upgrading to URP 12.0.x](upgrade-guide-2021-2.md)
```
phi-lira, line 22: Imo there should also be an entry in the upgrading to 12.0.x that mentions the old menu is gone and users shoudl use the new rp converter.

phi-lira (approved):

com.unity.render-pipelines.universal/Documentation~/TableOfContents.md
oleks-k: Thanks, I've added the point (and also a reference at the top of the page).

com.unity.render-pipelines.universal/Documentation~/features/rp-converter.md
TheoWong-pixel: URP to URP is still a conversion, so it makes sense. With all this tooling centralised it makes it more predictable for users in the future to know where to find upgraders and converters in the future. 

we built this framework in a way that is flexable enough to encompass things like upgrades, conversions and even maybe setting up a scene.

TheoWong-pixel (approved):

oleks-k (commented):
I've implemented the changes from Martin, Theo, Felipe.

phi-lira (approved):

erikabar (approved):
Looks good.

PR #5754, opened 2021-09-21
[HDRP][Path Tracing] Improved volumetric scattering light sampling

eturquin:
This PR addresses the following tickets, all related to volumetric scattering in path tracing:
- Noisy result when multiple point/area lights overlap  https://fogbugz.unity3d.com/f/cases/1358777
- Exaggeratedly strong contribution from directional lights when the ray hits a surface afterwards https://fogbugz.unity3d.com/f/cases/1304688
- Fog color is now taken into account in path-traced volumetric scattering, albeit in a slightly different way than in its raster counterpart https://fogbugz.unity3d.com/f/cases/1346105 https://fogbugz.unity3d.com/f/cases/1358783

The new version shows better convergence speed, and fixes the directional light intensity issue.
It also uses the fog color to produce a result similar to the non-volumetric one, and then adds volumetrically scattered contributions on top of it (which is IMHO a better behaviour than what the raster version does, whose computations I couldn't make sense of).

This has been tested on the scenes of the respective tickets, and on an additional test scene which mixes all kinds of lights (see screenshot below).


eturquin:
volscat

sebastienlagarde (approved):

sebastienlagarde:
For QA: this PR is also part of it and merged as well, so when you validate it include this one as well: https://github.com/Unity-Technologies/Graphics/pull/5798

PR #5759, opened 2021-09-22
Universal/bugfix/1362345

Nzollner:
### Purpose of this PR
This PR fixes a bug where motion vectors get in front of other objects which should stay on top described [here](https://fogbugz.unity3d.com/f/cases/1362345/).

---
### Testing status
I added ```207_Motion_Vectors_Cube_In_Front.scene``` to UniversalGraphicsTest_Foundation.

image
[Image of the new test scene.]

image
[Image of the game view when running.]

The new test scene results in a black scene. This makes sure motion vectors no longer gets in front of other objects.
The test scene is the same as the given scene for showing the bug.

I have UniversalGraphicsTest_Foundation on Windows 10 DX11 locally without any error. I didn't test the other graphics' tests because the fix only affects the motion vectors which are tested in UniversalGraphicsTest_Foundation.

---
### Comments to reviewers


phi-lira (approved):

Nzollner:
This solves the bug but creates a new one in Windows 10 using Vulkan.
206_Motion_Vectors

simon-engelbrecht-soerensen (approved):
Looks like this has been tested thoroughly, and with a test scene I feel comfortable in approving.

PR #5760, opened 2021-09-22
Fix for case 1366011 debug spew when switch target to non RT platform

peterjohnlong:
---
### Purpose of this PR
This PR is to attempt to fix fogbugz case https://fogbugz.unity3d.com/f/cases/1366011/
Where Realtime Raytracing (Preview) checkbox is ticked and then build target is switched to non raytrace console platform,
there are a continuous stream of errors output in the editor console window about missing compute kernels etc.
---
### Testing status
Locally tested that it fix the issue in the bug report. Local test tested that HDRP_DXR test still work as expected on a console target where raytrace is supported (tested console build was built under both DX11 and DX12 editor API).

---

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/AmbientOcclusionEditor.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/AmbientOcclusionEditor.cs
@@ -76,7 +76,7 @@ namespace UnityEditor.Rendering.HighDefinition
                 return;
             }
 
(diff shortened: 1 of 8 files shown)


anisunity:
Unfortunately, this behavior breaks the pc behavior. I'll change the fix to fix both platforms.

anisunity:
I still need to test on a ps4 to make sure it's okay

sebastienlagarde (approved):

sebastienlagarde:
For QA: Anis have done the QA pass on this PR, so no need for further testing

PR #5769, opened 2021-09-22
Added new synonyms for texture sampler nodes

bencloward:
### Purpose of this PR
Users will be able to find the texture sampler nodes in the library using the dx9 HLSL terms for them - tex2d, tex2dlod, tex2darray, and tex3d.  This will make the sampler nodes easier to discover.

---
### Testing status
Used each of the search terms to find the corresponding nodes in the Create Node menu.


cdxntchou (approved):
Woo!

PR #5771, opened 2021-09-22
Increasing HDRP fined pruned light tile count to 63.

kecho:
### Purpose of this PR
Increasing the default maximum value for a fine prune light tile from 23 to 63.
Added as well a configuration on ShaderConfig.cs to go back to 23 if needed.

NOTE: this configuration could also be done at runtime. Might be worth testing since it could add some extra cost.

---
### Testing status

* Tested on a scene with many lights of many types. Flickering does occur once you exceed the limit of 64 of a *big tile* 
* Tested the debug viewer
* See performance for additional tests.

scene with 50 point lights

Before:
(notice missing lights and tile artifacts due to 23 light limit per tile)
image

After:
(All lights represented corretly)
image

---
### Performance
Profiling done in  PS4 base with 2 scenes.
For these perf tests, I disabled async compute in the camera frame settings.

The only performance degradation observed was additional transient gpu memory utilized by the fine pruned light list buffer (in which case, at worst, it doubles) and additional cost from deferred lighting.

#### Scene 1 ####
Scene with justs 20 lights, looking for performance cost of just increasing the per pruned tile capacity.
image

Before
* Build Light list (2.48ms)
    * GenerateLightAABBs - 0.05ms
    * BigTilePrepass - 0.05ms
    * BuildPerTileLightList - 0.7ms
    * VoxelLightListGeneration  - 1.3ms
    * BuildDispatchIndirectArguments - 0.13ms
* Deferred lighting - 8.75ms

After
* Build Light list (2.48ms)
    * GenerateLightAABBs - 0.05ms
    * BigTilePrepass - 0.05ms
    * BuildPerTileLightList - 0.7ms
    * VoxelLightListGeneration  - 1.3ms
    * BuildDispatchIndirectArguments - 0.13ms
* Deferred lighting - 8.75ms

The profiling shows barely any fluctiation in costs of production / consumption of the fined pruned light list.

#### Scene 2 ####
Scene with 150 lights, looking for performance cost of additional lights and effects on deferred lighting.
image

Before
* Build Light list (4.63ms)
    * GenerateLightAABBs - 0.043ms
    * BigTilePrepass - 0.11ms
    * BuildPerTileLightList - 1.33ms
    * VoxelLightListGeneration  - 2.88ms
    * BuildDispatchIndirectArguments - 0.11ms
* Deferred lighting - 6.43ms

After
* Build Light list (4.59ms)
    * GenerateLightAABBs - 0.061ms
    * BigTilePrepass - 0.110ms
    * BuildPerTileLightList - 1.33ms
    * VoxelLightListGeneration  - 2.82ms
    * BuildDispatchIndirectArguments - 0.10ms
* Deferred lighting - 10.71ms (+4 ms from more lights processed)

The fluctuation in Build light list can be attributed to noise. The fluctuation in deferred lighting is direclty caused by processing more lights in the light loop.

---
### Notes for reviewers
Waiting on confirmation from Morten Mikkelsen to see if anything could be regressing on perf.

---
### Recommended QA test plan
The goal is to answer the question "Is the new light count more stable than the previous one?"
Stable is a bit of a subjective word because both light counts (the old one) and the new one have hard limits.
Before, at 23 lights you would start seeing missing lights or flickering artifacts in the tiles.
I would like to know how more stable is the new light count on a low light count (with many categories), with a medium count, and an extreme count. 

Old light count max was 23 on each tile.
New one is a max of 63 on each tile.

Please test both deferred and volumetric lighting.

* Test volumetric lighting, and figure out at which light count does it break?
* Is the new limit more forgiving with light count for volumetric / transparent and opaque? or does it behave less forgiving
* Test for light square artifacts on deferred / forward on scenes with transparent and opaque.
* Test ray tracing as well, check for light limits that cause flicker, compare with light limits that we had before.
* Please mix many light types, spot / area / light probe volume / GI probes


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2997,7 +3000,7 @@ void GetContactShadowMask(HDAdditionalLightData hdAdditionalLightData, BoolScala
             // If contact shadows are not enabled or we already reached the manimal number of contact shadows
             // or this is not rasterization
             if ((!hdAdditionalLightData.useContactShadow.Value(contactShadowEnabled))
-                || m_ContactShadowIndex >= LightDefinitions.s_LightListMaxPrunedEntries
+                || m_ContactShadowIndex >= ShaderConfig.FPTLMaxLightCount
```
sebastienlagarde, line 3003: hmm... I haven't think about this one. Guess contact shadow must stay at 32 max. So we need a separate definition. The contact shadow code write in an RGBA8 render target.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -107,9 +107,12 @@ class LightDefinitions
 
         // light list limits
         public static int s_LightListMaxCoarseEntries = 64;
-        public static int s_LightListMaxPrunedEntries = 24;
         public static int s_LightClusterMaxCoarseEntries = 128;
 
+        // We have room for ShaderConfig.FPTLMaxLightCount lights, plus 1 implicit value for length.
+        // We allocate only 16 bits per light index & length, thus we divide by 2, and store in a word buffer.
+        public static int s_LightDwordPerFplTile = ((ShaderConfig.FPTLMaxLightCount + 1)) / 2;
```
sebastienlagarde, line 114: typo: s_LightDwordPerFplTile => s_LightDwordPerFptlTile

com.unity.render-pipelines.high-definition-config/Runtime/ShaderConfig.cs
```diff
@@ -4,6 +4,13 @@
 
 namespace UnityEngine.Rendering.HighDefinition
 {
+    internal enum FPTLMaxLightSizes
+    {
+        Small = 23,
```
sebastienlagarde, line 9: I don't see any reasons to have this number configurable if there is no impact on perf. 63 is fine. If this is related to memory saving then we should evaluate the memory cost but then there is no difference between 31 and 63 so we should remove 23

sebastienlagarde (commented):
The PR only increase light number for opaque (fptl) but you need to also handle it for cluster otherwise when switching to MSAA we will go done to 23.

kecho:
@sebastienlagarde I have now pushed the change that correctly allocates up to 63 lights as well for clustered.

com.unity.render-pipelines.high-definition-config/Runtime/ShaderConfig.cs
kecho: As discussed, we want to keep this flexible so the user can do intense 8k rendering with a high light count cluster (>128 lights on a single spot).

sebastienlagarde: note. ok for 31 and 63 but 23 should be remove

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
kecho: Done

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
kecho: Done, made this a nice enum. We keeping it to 24 because we have the fade and bits slots set to 24.
This means that the maximum number of lights that can have contact shadows are 24 per pixel.

com.unity.render-pipelines.high-definition-config/Runtime/ShaderConfig.cs
kecho: Removed 23

kecho:
Running yamato, once I get the tests stable will remove the draft state, and send it to qa for testing.

RemyUnity:
Thank you for the detailed description and the QA guidelines ! The PR has been added for testing and review on our Jira board.

FrancescoC-unity:
While I don't necessarily think it will be more expensive, but have you tried to capture the overhead at sampling time of the increased size words per tile? 
(To do so I'd comment out the actual evaluation of the light or make it extremely trivial, so the cost is just grabbing the right index)

kecho:
> While I don't necessarily think it will be more expensive, but have you tried to capture the overhead at sampling time of the increased size words per tile?
> (To do so I'd comment out the actual evaluation of the light or make it extremely trivial, so the cost is just grabbing the right index)

Hey, I measured the deferred light loop at it had absolutely 0 changes in performance. Lmk if theres anything else I should look at.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
```diff
@@ -343,16 +346,17 @@ EnvLightData FetchEnvLight(uint index)
...
 }
 
-// In the first 8 bits of the target we store the max fade of the contact shadows as a byte
+// In the first bits of the target we store the max fade of the contact shadows as a byte.
+//By default its 8 bits for the fade and 24 for the mask, please check the LightLoop.cs definitions.
 void UnpackContactShadowData(uint contactShadowData, out float fade, out uint mask)
 {
-    fade = float(contactShadowData >> 24) / 255.0;
-    mask = contactShadowData & 0xFFFFFF; // store only the first 24 bits which represent
+    fade = float(contactShadowData >> CONTACT_SHADOW_MASK_BITS) / ((float)CONTACT_SHADOW_FADE_MASK);
```
FrancescoC-unity, line 353: So yeah contact shadow are gonna be awkward :D I wonder if we should change a bit how we do it later on. 

Now if we have only one light that has contact shadow, but that happens to have an index bigger than 24 in the list it will not show up. 

Even worse if that light changes depending on culling from say 20 and 25 as index, it will "flicker" contact shadows :/

FrancescoC-unity (approved):
LGTM to me overall, I am afraid tho the contact shadow issue might pop up as weird bugs soon I am afraid :(

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
sebastienlagarde: I don't remind exactly but aren't we sorting light by shadow casting? i.e the first set of light in the list currently cast shadow? increasing the likelihood that we have shadow on the first 24 light index. Guess something to test

kecho: So there is no sorting by shadow casting as far as I can tell. We sort by light type on the entire frame, then we sort per light index per tile.
The way it works is that we allocate all contact shadows from scratch every frame on all lights on the view. Not on a per tile basis.

The very first 24 lights that *Have* shadows enabled get it, and the allocation is on the CPU.  Note that we can have 1 light with contact shadows followed by 32 that dont, and then 24 more that have contact shadows. In this case it will be perfectly fine. 

The only way shadow allocation can vary is if a new light enters the view (passes culling).

For example, we have in the frame 512 lights in view. The first one and last 23 ones have contact shadows enabled.
 * In this situation we will never get flickering
 
 Now lets assume we have 1 more light with contact shadow, now we have a total of 25.
 If all lights are on view, it wont flicker either because the sorting in the CPU remains deterministic.
 
 Hope this makes sense.

FrancescoC-unity: Yeah stuff changed and forgot, used to be that each bit of the 24 bit mapped to the index of the light in the tile (and there 24 was convenient for that :P )

kecho:
@sebastienlagarde changes from Fast PrepareLightsForGPU are now merged. Tested locally. We should be good.

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Added public API to edit materials from script at runtime.
+- Added new configuration ShderOptions.FPTLMaxLightCount in ShaderConfig.cs for maximum light count per fine pruned tile.
```
RemyUnity, line 48: Typo "ShderOptions.FPTLMaxLightCount" :)

com.unity.render-pipelines.high-definition-config/Runtime/ShaderConfig.cs
```diff
@@ -36,6 +42,13 @@ public enum ShaderOptions
 
         /// <summary>Support to apply a global mip bias on all texture samplers of HDRP.</summary>
         GlobalMipBias = 1,
+
+        /// <summary>
+        /// Maximum number of lights for a fine pruned light tile. This number can only be 3 prespecified possibilities, 23, 31 or 63.
+        /// The reserved index is there for the count of the fined prunned tile list. The lower the count, some memory savings will occur for the tile light list.
+        /// Note: For any rendering bigger than 4k (in native) it is recommended to use Medium or Small light counts per tile, to avoid possible artifacts.
```
RemyUnity, line 49: This is confusing. In the doc there is mention of two possible values (32 & 64), but here you're mentioning 3 and recommend using 23 or 31 in big resolutions. And the Enum only has two. What is the right information ?

com.unity.render-pipelines.high-definition-config/Runtime/ShaderConfig.cs
```diff
@@ -4,6 +4,12 @@
 
 namespace UnityEngine.Rendering.HighDefinition
 {
+    internal enum FPTLMaxLightSizes
+    {
+        Small = 31,
+        High = 63
+    }
```
RemyUnity, line 11: Can't the be declared in the HDRP package to avoid users changing these values ?
Accordingly the comments lines 47 don't have to mention the values, only the enum "text".

Also, nitpicking about the semantics, but if one is "Small", shouldn't the other be "Big" ? Or "Low" / "High" ?
We are referring to a "count" here, so maybe the enum name is also not best chosen, something like "FPTLMaxLightBudget" would be a better fit ?

RemyUnity (changes requested):

RemyUnity (changes requested):
With further testing, I found an other issue. It looks like this PR introduces some instability when a lot of lights are displayed on screen. Aside from the expected tiles lacking lights (visible on the tog of the sphere in the captures), some lights globally flicker when it was "stable" before the PR.
Before the PR :

https://user-images.githubusercontent.com/32760367/138247765-2d6fe907-4731-495e-8539-2ac95724ef22.mp4

This PR, with 31 lights limit (behaviour similar with 63):

https://user-images.githubusercontent.com/32760367/138247881-fc787026-93bf-48da-a4cf-aeef96b8fa89.mp4

[Test project is here](https://drive.google.com/file/d/1kNsHFpiVa3nP78h0YW5VtgLBuaM_421Y/view?usp=sharing).

com.unity.render-pipelines.high-definition/CHANGELOG.md
kecho: Done

com.unity.render-pipelines.high-definition-config/Runtime/ShaderConfig.cs
kecho: Changed it to low as recommended.

kecho:
As far as the bug goes, unsure if we should make this a pr blocker sinec it was introduced by the other PR. I recommend creating a bug for it. Ill get to it right away though.

PR #5428, opened 2021-08-24
[ShaderGraph] [bugfix 1352662] [2022.1] Use VPOS for fragment stage screenspace calculations

cdxntchou:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1352662/

2022.1:
 + https://github.com/Unity-Technologies/Graphics/pull/5428
 + https://github.com/Unity-Technologies/Graphics/pull/5784

2021.2 backport:
 + https://github.com/Unity-Technologies/Graphics/pull/5817

The issue is that the calculation for screenspace position is calculated from world space position via applying the matrix transforms.  However, the calculation is prone to floating point error, that is exacerbated greatly when the camera is far from the origin and the pipeline is not using camera-relative "world" space.

You can see in the examples below that the error can easily grow to be larger than a pixel.

This PR fixes this by changing the calculation in the fragment shader, to be based on the value returned by the VPOS fragment input.  This value is not based on the projection matrices, independent of camera position, and much more accurate overall.

Docs ticket: https://jira.unity3d.com/browse/GSG-593
Console Test PR: https://github.cds.internal.unity3d.com/unity/ScriptableRenderPipelinePrivate/pull/239

Test example:
Left sphere visualizes frac(screenspace pixel coordinate) using a gradient (red, black, green) where 0.5 is black.
Right sphere shows a screenspace checkerboard alphatest pattern

Before:  (at 20000 distance from origin) -- quite a bit of inaccuracy
image

After:   (at 20000 distance from origin) -- perfect 0.5 centered pixel coordinates everywhere!
image

Implementation Details:
We introduce two new global values that ShaderGraph can produce, and nodes can request, in addition to ScreenPosition.
The values are defined as:
PixelPosition                 2D pixel coordinate of the current fragment or vertex. Pixels are centered at 0.5.
NDCPosition                2D coordinate (0,1) across the viewport; equal to PixelPosition / ScreenResolution.
ScreenPosition (raw)     4D Clip Space homogenous coordinates of the current fragment or vertex.

Both PixelPosition and NDCPosition are calculated from VPOS in the fragment stage, and are calculated from ScreenPosition in other stages.

---
### Testing status

Tested in both HDRP and URP:
* Tested that Screen Position node and slots work correctly in both fragment and vertex shader.
* Validated that "Raw" mode works identically to previous, returning clip space ZW.
* Tested that the other modes return exact values in the fragment shader, based off of the VPOS value.
* Tested that VFX screen position works.

[...]

PR #5784, opened 2021-09-23
[VFX] Fix inconsistent NeedsPositionWorldInterpolator

PaulDemeulenaere:
---
### Purpose of this PR
Fix issue introduced at https://github.com/Unity-Technologies/Graphics/pull/5428
```
Unhandled log message: '[Error] Shader error in 'Hidden/VFX/SampleScene/System (1)/Output Particle Quad': 'VFXGetPositionRWS': no matching 1 parameter function at line 1647 (on d3d11)
```
~~The condition which generates VFX_VARYING_POSWS wasn't consistent between defines & code generation~~
Extract code which computes the ScreenPosition/PositionNDC from the WorldPosition dependency, it saves interpolator & computation when it isn't actually needed.

---
### Testing status
Locally 🟢 
Yamato ⏳ 

---
### Comments to reviewers
See this [conversation](https://unity.slack.com/archives/G1BTWN88Z/p1632390484205400)

### DIFF
--- a/com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
+++ b/com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
@@ -749,7 +749,8 @@ namespace UnityEditor.VFX
         private static bool NeedsPositionWorldInterpolator(GraphCode graphCode)
         {
             return graphCode.requirements.requiresPosition != NeededCoordinateSpace.None
-                    || graphCode.requirements.requiresViewDir != NeededCoordinateSpace.None;
+                    || graphCode.requirements.requiresViewDir != NeededCoordinateSpace.None
+                    || graphCode.requirements.requiresScreenPosition;
         }
 
         public override IEnumerable<KeyValuePair<string, VFXShaderWriter>> additionalReplacements
@@ -866,24 +867,22 @@ namespace UnityEditor.VFX
                                 if ((graphCode.requirements.requiresViewDir & NeededCoordinateSpace.Tangent) != 0)
                                     callSG.builder.AppendLine("INSG.TangentSpaceViewDirection = mul(tbn, V);");
                             }
-                        }
-
-                        if (graphCode.requirements.requiresScreenPosition
-                            || graphCode.requirements.requiresNDCPosition
-                            || graphCode.requirements.requiresPixelPosition)
-                        {
-                            callSG.builder.AppendLine("{");
-                            if (graphCode.requirements.requiresNDCPosition || graphCode.requirements.requiresPixelPosition)
-                                callSG.builder.AppendLine("float2 localPixelPosition = i.VFX_VARYING_POSCS.xy;");
-                            if (graphCode.requirements.requiresNDCPosition)
-                                callSG.builder.AppendLine("float2 localNDCPosition = localPixelPosition.xy / _ScreenParams.w;");
 
                             if (graphCode.requirements.requiresScreenPosition)
-                                callSG.builder.AppendLine("INSG.ScreenPosition = ComputeScreenPos(localPixelPosition, _ProjectionParams.x);");
+                            {
+                                //ScreenPosition is expected to be the raw screen pos (float4) before the w division in pixel (SharedCode.template.hlsl)
+                                callSG.builder.AppendLine("INSG.ScreenPosition = ComputeScreenPos(VFXTransformPositionWorldToClip(i.VFX_VARYING_POSWS), _ProjectionParams.x);");
+                            }
+                        }
+
+                        if (graphCode.requirements.requiresNDCPosition || graphCode.requirements.requiresPixelPosition)
+                        {
+                            callSG.builder.AppendLine("{");
+
                             if (graphCode.requirements.requiresNDCPosition)
-                                callSG.builder.AppendLine("INSG.NDCPosition = localPixelPosition / _ScreenParams.xy;");
+                                callSG.builder.AppendLine("INSG.NDCPosition = i.VFX_VARYING_POSCS.xy / _ScreenParams.xy;");
                             if (graphCode.requirements.requiresPixelPosition)
-                                callSG.builder.AppendLine("INSG.PixelPosition = localPixelPosition;");
+                                callSG.builder.AppendLine("INSG.PixelPosition = i.VFX_VARYING_POSCS.xy;");
                             callSG.builder.AppendLine("}");
                         }
 


com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
```diff
@@ -748,6 +746,15 @@ public override void EndCompilation()
...
         }
 
+        private static bool NeedsPositionWorldInterpolator(GraphCode graphCode)
+        {
+            return graphCode.requirements.requiresPosition != NeededCoordinateSpace.None
+                    || graphCode.requirements.requiresViewDir != NeededCoordinateSpace.None
+                    || graphCode.requirements.requiresScreenPosition
+                    || graphCode.requirements.requiresNDCPosition
+                    || graphCode.requirements.requiresPixelPosition;
+        }
```
julienf-unity, line 756: I'm not that familiar with SG but why would a world pos interpolator be required in some of those cases? SV_Position seems enough.

julienf-unity: I mean isnt it exactly what the incriminated PR was fixing?

PaulDemeulenaere: The issue has been introduced precisely at this change: https://github.com/Unity-Technologies/Graphics/commit/d74237b8a3f6e41bfa091ed6bff60014efe32d13#diff-12c1990cad6d2bf349a05118c2befeb419e8bb047aed21d85c30600aac5c554f

The problem is about the condition of VFX_NEEDS_POSWS_INTERPOLATOR define, it used this set of test :
```
graphCode.requirements.requiresPosition != 
NeededCoordinateSpace.None 
|| graphCode.requirements.requiresScreenPosition 
|| graphCode.requirements.requiresViewDir != NeededCoordinateSpace.None;
```

While the condition for generating at call to `VFXGetPositionRWS(i.VFX_VARYING_POSWS);` is now :
```
graphCode.requirements.requiresPosition != NeededCoordinateSpace.None 
|| graphCode.requirements.requiresViewDir != NeededCoordinateSpace.None 
|| graphCode.requirements.requiresScreenPosition 
|| graphCode.requirements.requiresNDCPosition 
|| graphCode.requirements.requiresPixelPosition
```

If `VFX_NEEDS_POSWS_INTERPOLATOR` is defined, then `VFX_VARYING_POSWS` is available.

The computation of old .ScreenPosition still uses a not optimal approach:
```
INSG.ScreenPosition = ComputeScreenPos(VFXTransformPositionWorldToClip(i.VFX_VARYING_POSWS), _ProjectionParams.x);
```

julienf-unity: Ok but the correct fix would be the same as #5428 ie use SV_Position instead of WorldPosition where relevant. It does not make sense to compute screenspace pos from World pos in fragment when it's available in SV_Position already. It's more expensive and introduce errors. I didnt know it was made like that in SG in the first place.

PaulDemeulenaere: 👍 I will extract the code which doesn't use `VFX_VARYING_POSWS` and remove `graphCode.requirements.requiresNDCPosition || graphCode.requirements.requiresPixelPosition` from NeedsPositionWorldInterpolator condition.

PaulDemeulenaere:
I'm switching back to draft while I'm improving this implementation: https://github.com/Unity-Technologies/Graphics/pull/5784#discussion_r714894355

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
PaulDemeulenaere: Fixed at c78ff305c8caf4800881c3a318a6b9e7607fc7f4

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
```diff
@@ -880,6 +868,25 @@ public override void EndCompilation()
...
+                            || graphCode.requirements.requiresPixelPosition)
+                        {
+                            callSG.builder.AppendLine("{");
+                            if (graphCode.requirements.requiresNDCPosition || graphCode.requirements.requiresPixelPosition)
+                                callSG.builder.AppendLine("float2 localPixelPosition = i.VFX_VARYING_POSCS.xy;");
+                            if (graphCode.requirements.requiresNDCPosition)
+                                callSG.builder.AppendLine("float2 localNDCPosition = localPixelPosition.xy / _ScreenParams.w;");
+
+                            if (graphCode.requirements.requiresScreenPosition)
+                                callSG.builder.AppendLine("INSG.ScreenPosition = ComputeScreenPos(localPixelPosition, _ProjectionParams.x);");
```
julienf-unity, line 882: This seems incorrect. computeScreenPos expects HClip pos as input and you're passing float2 pixel pos. It should not even compile. Do we have test coverage for this?

julienf-unity (approved):

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
```diff
@@ -880,6 +868,25 @@ public override void EndCompilation()
...
 
+                        if (graphCode.requirements.requiresScreenPosition
+                            || graphCode.requirements.requiresNDCPosition
+                            || graphCode.requirements.requiresPixelPosition)
+                        {
+                            callSG.builder.AppendLine("{");
+                            if (graphCode.requirements.requiresNDCPosition || graphCode.requirements.requiresPixelPosition)
+                                callSG.builder.AppendLine("float2 localPixelPosition = i.VFX_VARYING_POSCS.xy;");
+                            if (graphCode.requirements.requiresNDCPosition)
+                                callSG.builder.AppendLine("float2 localNDCPosition = localPixelPosition.xy / _ScreenParams.w;");
```
julienf-unity, line 879: Does this make sense? It the variable even used and what does it represents?

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
PaulDemeulenaere: There is a special definition of this in `ShaderGraphLibrary/Functions.hlsl#L20`
https://github.com/Unity-Technologies/Graphics/blob/0de4625e8b5604e3e573b4b681b84460ef77d791/com.unity.shadergraph/ShaderGraphLibrary/Functions.hlsl#L20
This code path is only used with shaderGraph and it's equivalent to this, in URP:
https://github.com/Unity-Technologies/Graphics/blob/75f4b421fa774d43000d703f2bbf05d7a1ca6606/com.unity.render-pipelines.universal/Editor/ShaderGraph/Templates/SharedCode.template.hlsl#L56
And in HDRP:
https://github.com/Unity-Technologies/Graphics/blob/75f4b421fa774d43000d703f2bbf05d7a1ca6606/com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/Templates/Pixel.template.hlsl#L39

julienf-unity: So screenpos is still requiring worldpos in SG then?

PaulDemeulenaere: > Do we have test coverage for this?
We have two test which covers this.
1) SampleScene:
image
This one is run on HDRP & URP.

2) ShadergraphUnlit
image
This one is HDRP only.

PaulDemeulenaere: Oh but it's float2 vs float4 😶
It should at least generate a warning.

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
julienf-unity: I bet it's a typo between .w and .xy. But still, this line seems uneeded as ndc is computed a few lines below

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
julienf-unity: But do we really test ScreenPosition in PS? How could the test pass if it's the case?

PaulDemeulenaere: Indeed, I double check, I though "ScreenPosition" was used in "Default" Mode, but actually, it's correspond to the "Raw" value.
image
I'm fixing this case 🔧

PaulDemeulenaere: Resolved at 1db0279e862a495577f27d9a9f7e79bdedc61048 (moving back to ScreenPosition computed from WorldPos as before)

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
PaulDemeulenaere: I removed the localNDCPosition & localPixelPosition for code generation (localPixelPosition == i.VFX_VARYING_POSCS.xy;). It prevents a missing condition and improve readability.

PaulDemeulenaere: I also quickly tested all cases from ScreenPosition matching between VFX & SG: 
https://user-images.githubusercontent.com/22494706/134672858-30a2a786-7d90-46fd-a273-706be48b6114.mp4

PaulDemeulenaere: However, @cdxntchou About this https://github.com/Unity-Technologies/Graphics/blob/75f4b421fa774d43000d703f2bbf05d7a1ca6606/com.unity.render-pipelines.universal/Editor/ShaderGraph/Templates/SharedCode.template.hlsl#L56

Do you know why `ScreenPosition` (so, if I understand correctly the exposed "Raw" settings from the ScreenPosition node in ShaderGraph) always uses a TransformWorldToHClip(worldPos) ?
Can't we recompute the initial clip position using SV_Position (multiplying by w to return into the initial homogenous space) ? I think I'm missing an obvious constraint which prevent us to remove this matrix mulplication in pixel.

cdxntchou (approved):

com.unity.visualeffectgraph/Editor/ShaderGraph/VFXShaderGraphParticleOutput.cs
PaulDemeulenaere: N.B.: We discussed internally about this possibility, it's a potential improvement but we should check the consistency among platform first.
See this [conversation](https://unity.slack.com/archives/C02G8EDKLLQ/p1632505097008800)

PaulDemeulenaere:
This change 11417d30e2ebf958431796c3eee10595d3c5369f must be backported right after or preferably within the same backport PR.
See https://github.com/Unity-Technologies/Graphics/pull/5784

PaulDemeulenaere:
Has been backported with https://github.com/Unity-Technologies/Graphics/pull/5817

PR #5786, opened 2021-09-23
Add calls to AssetDatabase.StartAssetEditing and AssetDatabase.StopAs…

peterhall-unity3d:
…setEditing around the loops which reimport all materials, significantly speeds up this process.

---
### Purpose of this PR
When upgrading HDRP/URP version a material reimport is triggered.  This operation refreshes the asset database for each material.  On larger projects this can take ~10 seconds each time we refresh, and with 100s or 1000s of materials this leads to a reimport time of maybe a couple of hours.  By placing calls to AssetDatabase.StartAssetEditing and AssetDatabase.StopAssetEditing around the import loop we can avoid doing the refresh operation, cutting the import time from (in a real world case) 3 hours to 3 minutes.  In the HDRP sample project the import time is roughly cut in half from 12 seconds to 6 seconds.

---
### Testing status
Manually tested upgrading HDRP version with and without this change in the HDRP test project and a large real world project.


sebastienlagarde:
Thanks for doing this PR. Could you add a changelog? (in hdrp and urp).
Also can you confirm that those call to asset database exist in previous version and that we can backport the PR (we will handle it for you)? thanks

sebastienlagarde:
note yamato is green error come from master

RSlysz (approved):
LGTM

peterhall-unity3d:
> Thanks for doing this PR. Could you add a changelog? (in hdrp and urp).
> Also can you confirm that those call to asset database exist in previous version and that we can backport the PR (we will handle it for you)? thanks

Yes, these calls exist back to 2018.4 (and beyond) so backporting should be fine:
https://docs.unity3d.com/2018.4/Documentation/ScriptReference/AssetDatabase.StartAssetEditing.html
https://docs.unity3d.com/2018.4/Documentation/ScriptReference/AssetDatabase.StopAssetEditing.html

Have added changelog entries and also wrapped the calls with a try/finally block as suggested by latest docs on those functions.

sebastienlagarde:
Waiting Felipe  and urp qa approval.

phi-lira (approved):

sebastienlagarde:
For QA: merging to be sure this optimization is in the official release of 21.2

iM0ve:
I have tested by upgrading two projects - Amalienborg and HDRP Template. Did not encounter any related warnings/errors. All materials imported correctly

Images after upgrade
Unity_D3JZtE04rD

HwmeL6YLa5

PR #5790, opened 2021-09-23
[Universal] Fixed changing camera position in beginCameraRendering won't work for XR

thomas-zeng:
---
### Purpose of this PR
- Fixed fogbugz case 1365000

---
### Testing status
- The repro project in fogbugz case is for 2021.2. I have created a new project in 2022.1 with the beginCameraRendering script from the repro case. Tested with new project and verified the issue is fixed using mock provider.
- Also backported the changes to 2021.2/staging branch locally and tested the repro project against the local branch. Verified the issue is fixed there.


phi-lira (approved):
We should add a test to capture this bug.

fabien-unity (approved):

thomas-zeng:
Adding XR QA ryan as reviewer

ryanhy-unity:
This change appears to break color texture sampling in my shader graph shader in my test project.  When I use master URP the render texture samples the color correctly but when I use this code that texture is mostly black.  I'm using the URP E2E test project from https://github.cds.internal.unity3d.com/unity/xr.xrautomatedtests.

ryanhy-unity (changes requested):
This change seem to break a shader I created with Shader Graph to sample background color.  My depth sampling shader seems a bit off too but I'm not as certain there.  Let me know if you'd like more details, render doc caps or whatnot.

ryanhy-unity:
Apparently the issues I was seeing while testing this were due to issues in master and not this PR.  I did actually test master during this to make sure that wasn't the case but I think I failed to pull latest master and so missed it.  Here's the bug if you're interested.  https://fogbugz.unity3d.com/f/cases/1370540/

Given that the issue I found was not a regression in this PR and the bug was indeed fixed I think this PR is good.  Approved.

ryanhy-unity (approved):

PR #5793, opened 2021-09-24
[VFX] Fix usage on texture internally in SG

PaulDemeulenaere:
---
### Purpose of this PR
Fix regression introduced at https://github.com/Unity-Technologies/Graphics/pull/5416 
https://fogbugz.unity3d.com/f/cases/1367167/
We shouldn't only list texture declared as input slot but also internal texture listed in `textureInfos`
It only concerns the new SG integration.

---
### Testing status
- Tested the repro case from the [fogbugz](https://fogbugz.unity3d.com/f/cases/1367167/)
- I slightly modified one of our graphicTest to use the two possibilities (without changing the expected rendering)

**In URP** We use an internal texture declaration (which ends in textureInfos)
image

**In HDRP** We use an exposed texture (which ends in InputSlot).
image
- Yamato🟢: The failure aren't related to this PR

---
### Comments to reviewers
N/A


julienf-unity (approved):

VitaSkruibyte (approved):
Looking great, thank you!

Tested:
1. Gather Texture 2D/3D,Texture 2D Array, Load/Sample Texture 2D Array, Sample/Load Texture 2D/3D, cubemap
2. Ran FTP
3. Checked URP and HDRP

PaulDemeulenaere:
Already backported in https://github.com/Unity-Technologies/Graphics/pull/5914

PR #5795, opened 2021-09-24
Generate a material as a subasset for ShaderGraphs

adrien-de-tocqueville:
### Purpose of this PR

This PR adds an automatically generated material subasset on ShaderGraphs:

material_subasset
graph inspector

I bumped the shadergraph importer version, so for existing projects every graph will be reimported and the material willl be generated.

Benefits:
* It avoids having to manually create a material each time a shadergraph is created
* When the shadergraph is saved, the material is regenerated, so the default values for properties are updated on the material
* When changing the default values in the blackboard, the material is updated and therefore changes can be seen instantly in the scene view.

Supported properties:
* float/vector2/vector3/vector4
* color
* boolean
* texture 2D/3D/Array/Cube
* keyword Enum/Boolean

That's essentially all properties that can be exposed, I don't think I am missing one

This is part of this JIRA task https://unity3d.atlassian.net/browse/UR-1450 

---
### Testing status

See the gif above.
Tested undo/redo of shadergraph properties.
Tested closing shadergraph window with saving or discarding changes.
Tested that the generated material works on a mesh renderer in a standalone build.

The feature has also been tested by several artists in the context of material variants
I will add a graphic test.


com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -311,6 +311,18 @@ public string path
             }
         }
 
+        Material m_Material;
```
cdxntchou, line 314: What is this material used for?   I would think we don't need to store the material in the source asset, just in the imported artifacts...

cdxntchou (changes requested):

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
cdxntchou: Ah, I see, it's a non-serialized cache of the runtime generated material, to enable the instant update..

cdxntchou: Being part of graph data makes it potentially re-entrant, which is not ideal.   In the sense that it should not be used by any part of import as it's data from the previous import.  It's basically a short-circuit to modify the output results, in situations where we know the changes only modify the material.   I think it would be better to move this cache to another location in ShaderGraph...

cdxntchou: or at minimum name the public getter something scary, like "defaultMaterialArtifact" or something...

cdxntchou: And add a comment that this is to do immediate updates of the material, skipping the import job.

cdxntchou: Also:  how do we back out these changes if the user doesn't save the graph?

com.unity.shadergraph/Editor/Drawing/Controllers/ShaderInputViewController.cs
```diff
@@ -50,33 +50,43 @@ void ChangePropertyValue(GraphData graphData)
             {
                 case BooleanShaderProperty booleanProperty:
                     booleanProperty.value = ((ToggleData)newShaderInputValue).isOn;
+                    if (graphData.material) graphData.material.SetFloat(shaderInputReference.referenceName, booleanProperty.value ? 1.0f : 0.0f);
```
cdxntchou, line 53: How do we undo this if the user closes the graph, and doesn't save it?
We would have to reset the material back to defaults in that case, to keep it in sync with the actual state of the graph on disk.

cdxntchou: (and there's also the problem if unity crashes, it could leave the Material in a non-synced state... that wouldn't get cleaned up until you force-reimported the shadergraph)

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
adrien-de-tocqueville: I moved the property to the GraphObject, maybe it makes a bit more sense. Also renamed it to materialArtifact with comment.

You're right, I didn't think about discarding changes, it's fixed

com.unity.shadergraph/Editor/Drawing/Controllers/ShaderInputViewController.cs
adrien-de-tocqueville: I tested the case with a crash and it works fine, because the generated material is never saved so when reopenning the project after a crash it will be in the state it was at the end of import

sebastienlagarde:
>For existing projects, the material will not be generated until the SG is saved. It can be done by bumping the importer version but I don't think it's necessary

yes we should bump the importer version.

>Should the user be able to disable this feature in the graph settings ?
no

sebastienlagarde (commented):
What is happening if a users setup this on the fly material to an object and build. does it works in player? (i.e is the build step considering it is a true material and store it correctly in build data? Because shader graph don't exist at runtime.

adrien-de-tocqueville:
> What is happening if a users setup this on the fly material to an object and build. does it works in player? (i.e is the build step considering it is a true material and store it correctly in build data? Because shader graph don't exist at runtime.

It works fine, actually it's the same process as for the shader itself, which is an artifact generated by the shadergraph importer that is available at runtime

com.unity.shadergraph/Editor/Drawing/Controllers/ShaderInputViewController.cs
cdxntchou: nice!   Did you test doing "modify property default", Ctrl+S save project, then crash?

cdxntchou: I wonder if we can mark the material as "don't save this, runtime only modifications present" when we change it...

cdxntchou (approved):
Looks good!

As we discussed, there might be an edge case if some script comes along and forces AssetDatabase to save the dynamic changes to the default material artifact.   Not sure if that is even possible, but that should be an issue with the script if so.

com.unity.shadergraph/Editor/Drawing/Controllers/ShaderInputViewController.cs
adrien-de-tocqueville: Given the public API of the assetdatabase, even if we could do it internally I am pretty sure it's impossible for users to save changes to the material

alindmanUnity (approved):
UX approval

GrantLamb-Unity (approved):
Ran some manual tests with the following exposed properties:
- float
- Texture 2D
- Color

Also ran tests to ensure naming conflicts don't cause issues with this auto generated material. No issues were encountered during these tests.

sebastienlagarde (commented):
do we have any what's new for shader graph?

PR #5799, opened 2021-09-24
[HDRP][URP] update manifest of SRP templates

sebastienlagarde:
### Purpose of this PR
This slack thread: https://unity.slack.com/archives/CV9P2CNP4/p1632419133062300 mention that we need to update manifest of urp and hdrp template as we will not have package by default anymore. so we do it.

Also fix obsolete material on hdrp template

---
### Testing status

---
### Comments to reviewers

### DIFF
--- a/com.unity.template-hd/Packages/com.unity.template.hd/ValidationExceptions.json
+++ b/com.unity.template-hd/Packages/com.unity.template.hd/ValidationExceptions.json
@@ -2,7 +2,7 @@
   "ErrorExceptions": [
     {
       "ValidationTest": "Primed Library Validation",
-      "PackageVersion": "13.0.0"
+      "PackageVersion": "13.1.0"
     }
   ],
   "WarningExceptions": []
--- a/com.unity.template-universal/Packages/com.unity.template.universal/ValidationExceptions.json
+++ b/com.unity.template-universal/Packages/com.unity.template.universal/ValidationExceptions.json
@@ -2,7 +2,7 @@
   "ErrorExceptions": [
     {
       "ValidationTest": "Primed Library Validation",
-      "PackageVersion": "13.0.0"
+      "PackageVersion": "13.1.0"
     }
   ],
   "WarningExceptions": []


com.unity.template-universal/Packages/com.unity.template.universal/package.json
```diff
@@ -7,6 +7,50 @@
   "host": "hub",
```
nicolasvidil, line 7: It's an update for 21.2 but the minimum version is set to 22.1.
Was it updated using the 22.1 version?

com.unity.template-universal/Packages/com.unity.template.universal/package.json
```diff
@@ -7,6 +7,50 @@
...
+    "com.unity.modules.unitywebrequest": "1.0.0",
+    "com.unity.modules.unitywebrequestassetbundle": "1.0.0",
+    "com.unity.modules.unitywebrequestaudio": "1.0.0",
+    "com.unity.modules.unitywebrequesttexture": "1.0.0",
+    "com.unity.modules.unitywebrequestwww": "1.0.0",
+    "com.unity.modules.vehicles": "1.0.0",
+    "com.unity.modules.video": "1.0.0",
+    "com.unity.modules.vr": "1.0.0",
+    "com.unity.modules.wind": "1.0.0",
+    "com.unity.modules.xr": "1.0.0"
```
nicolasvidil, line 54: We don't need to add the modules. These will still be injected by default. You can remove all this

com.unity.template-universal/Packages/com.unity.template.universal/package.json
```diff
@@ -7,6 +7,50 @@
...
+	"com.unity.2d.sprite": "1.0.0",
+    "com.unity.2d.tilemap": "1.0.0",
+    "com.unity.ads": "3.7.1",
+    "com.unity.analytics": "3.6.11",
+    "com.unity.collab-proxy": "1.7.1",
+    "com.unity.ext.nunit": "1.0.6",
+    "com.unity.feature.development": "1.0.1",
+    "com.unity.purchasing": "3.2.2",
+    "com.unity.render-pipelines.universal": "13.0.0",
+    "com.unity.test-framework": "1.1.29",
```
nicolasvidil, line 19: We can remove test-framework, it's already included in the engineering feature set

com.unity.template-universal/Packages/com.unity.template.universal/package.json
```diff
@@ -7,6 +7,50 @@
...
   "dependencies": {
-    "com.unity.render-pipelines.universal": "13.0.0"
+	"com.unity.2d.sprite": "1.0.0",
+    "com.unity.2d.tilemap": "1.0.0",
+    "com.unity.ads": "3.7.1",
+    "com.unity.analytics": "3.6.11",
```
nicolasvidil, line 13: Are you sure you want these packages in there?

com.unity.template-universal/Packages/com.unity.template.universal/package.json
```diff
@@ -7,6 +7,50 @@
...
+    "com.unity.collab-proxy": "1.7.1",
+    "com.unity.ext.nunit": "1.0.6",
+    "com.unity.feature.development": "1.0.1",
+    "com.unity.purchasing": "3.2.2",
```
nicolasvidil, line 17: Isn't this a mobile template? is it needed?

com.unity.template-universal/Packages/manifest.json
```diff
@@ -6,9 +6,7 @@
     "com.unity.analytics": "3.6.11",
```
nicolasvidil, line 6: Same as in the package.json file, I don't think the first 4 packages should be there. Is this intended?

com.unity.template-hd/Packages/com.unity.template.hd/package.json
```diff
@@ -7,9 +7,47 @@
   "unity": "2022.1",
```
nicolasvidil, line 7: Shouldn't it be 21.2 ?

com.unity.template-hd/Packages/com.unity.template.hd/package.json
```diff
@@ -7,9 +7,47 @@
...
+    "com.unity.modules.unitywebrequest": "1.0.0",
+    "com.unity.modules.unitywebrequestassetbundle": "1.0.0",
+    "com.unity.modules.unitywebrequestaudio": "1.0.0",
+    "com.unity.modules.unitywebrequesttexture": "1.0.0",
+    "com.unity.modules.unitywebrequestwww": "1.0.0",
+    "com.unity.modules.vehicles": "1.0.0",
+    "com.unity.modules.video": "1.0.0",
+    "com.unity.modules.vr": "1.0.0",
+    "com.unity.modules.wind": "1.0.0",
+    "com.unity.modules.xr": "1.0.0"
```
nicolasvidil, line 51: no modules needed, we can remove all this

com.unity.template-universal/Packages/com.unity.template.universal/package.json
```diff
@@ -7,6 +7,50 @@
...
+    "com.unity.render-pipelines.universal": "13.0.0",
+    "com.unity.test-framework": "1.1.29",
+    "com.unity.textmeshpro": "3.0.6",
+    "com.unity.timeline": "1.6.2",
+    "com.unity.ugui": "1.0.0",
+    "com.unity.xr.legacyinputhelpers": "2.1.8",
```
nicolasvidil, line 23: It feels to me this one doesn't belong here either. Is this intended?

com.unity.template-hd/Packages/com.unity.template.hd/package.json
```diff
@@ -7,9 +7,47 @@
...
-    "com.unity.cinemachine": "2.8.0",
+    "com.unity.collab-proxy": "1.9.0",
+    "com.unity.feature.development": "1.0.1",
+    "com.unity.textmeshpro": "3.0.6",
+	"com.unity.cinemachine": "2.8.0",
+    "com.unity.feature.development": "1.0.1",
     "com.unity.inputsystem": "1.1.0-pre.5",
     "com.unity.learn.iet-framework": "2.0.0",
+    "com.unity.learn.iet-framework.authoring": "1.0.0",
     "com.unity.render-pipelines.high-definition": "13.0.0"
```
nicolasvidil, line 18: You're missing a coma `,` at the end of the line, Yamato is broken because of it

com.unity.template-universal/Packages/manifest.json
```diff
@@ -6,9 +6,7 @@
...
     "com.unity.collab-proxy": "1.7.1",
     "com.unity.ext.nunit": "1.0.6",
-    "com.unity.ide.rider": "3.0.7",
-    "com.unity.ide.visualstudio": "2.0.9",
-    "com.unity.ide.vscode": "1.2.3",
+    "com.unity.feature.development": "1.0.1",
     "com.unity.purchasing": "3.2.2",
     "com.unity.render-pipelines.universal": "13.0.0",
     "com.unity.test-framework": "1.1.29",
```
nicolasvidil, line 12: No need for this one, and we should remove the modules as well

nicolasvidil (changes requested):
I know it's still in draft mode, but I figured I would give your some feedback

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: this template update is for master. The backport will be for 21.2

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: done

com.unity.template-hd/Packages/com.unity.template.hd/package.json
sebastienlagarde: PR is for master a backport to 21.2 will be perform after

com.unity.template-hd/Packages/com.unity.template.hd/package.json
sebastienlagarde: done

com.unity.template-universal/Packages/manifest.json
sebastienlagarde: We can try, but last time we have try to remove test-framework nothing was working... (on yamato side)

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: I have simply copy/paste existing manifest package list, I have no idea what is really required :)

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: I have simply copy/paste existing manifest package list, I have no idea what is really required :)

com.unity.template-universal/Packages/manifest.json
sebastienlagarde: I have simply copy/paste existing manifest package list, I have no idea what is really required :)

sebastienlagarde:
Yamato test pass and the PR is no ready for review and Qa valiation (@joeksy ) thanks

com.unity.template-universal/Packages/com.unity.template.universal/package.json
```diff
@@ -7,6 +7,14 @@
   "host": "hub",
   "description": "The Universal Render Pipeline Template configures Project settings for Projects where performance, wide platform support, and ease of customizing graphics are the primary considerations.\n\nThis Template uses the Universal Render Pipeline (URP) and Shader Graph. URP is prebuilt Scriptable Render Pipeline that is quick and easy to customize, and lets you create optimized graphics across a wide range of platforms. URP also includes an optimized 2D renderer complete with 2D lights and pixel perfect rendering, and an integrated post-processing solution. Shader Graph is a tool that allows you to create shaders using a visual node editor instead of writing code.\n\nThis template contains a sample Scene that contains examples of how to configure lighting settings, Materials, Shaders, and post-processing effects in URP, several preconfigured Universal Render Pipeline Assets that let you quickly swap between graphics quality levels, and Presets that have been optimized for use with URP.",
   "dependencies": {
-    "com.unity.render-pipelines.universal": "13.0.0"
+    "com.unity.2d.sprite": "1.0.0",
+    "com.unity.2d.tilemap": "1.0.0",
```
vicky-voin, line 11: I see these packages have been in the `manifest.json` since before this PR, but the `package.json` didn't have them before and they are not injected by default - so they technically haven't been part of URP Sample before, is this an intentional change?

com.unity.template-universal/Packages/com.unity.template.universal/package.json
```diff
@@ -7,6 +7,14 @@
...
+    "com.unity.collab-proxy": "1.7.1",
+    "com.unity.feature.development": "1.0.1",
+    "com.unity.render-pipelines.universal": "13.0.0",
+    "com.unity.textmeshpro": "3.0.6",
+    "com.unity.timeline": "1.6.2",
+    "com.unity.ugui": "1.0.0",
+    "com.unity.xr.legacyinputhelpers": "2.1.8"
```
vicky-voin, line 18: Same as my comment about the 2d packages above

vicky-voin (changes requested):

com.unity.template-hd/Packages/com.unity.template.hd/package.json
```diff
@@ -7,9 +7,16 @@
...
   "dependencies": {
+    "com.unity.collab-proxy": "1.9.0",
+    "com.unity.feature.development": "1.0.1",
+    "com.unity.textmeshpro": "3.0.6",
     "com.unity.cinemachine": "2.8.0",
+    "com.unity.feature.development": "1.0.1",
     "com.unity.inputsystem": "1.1.0-pre.5",
     "com.unity.learn.iet-framework": "2.0.0",
-    "com.unity.render-pipelines.high-definition": "13.0.0"
+    "com.unity.learn.iet-framework.authoring": "1.0.0",
```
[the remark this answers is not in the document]
sebastienlagarde: I have no idea. Guess it is require to work with IET no? so any change to iet will be required. Guess we should let him otherwise it will be painful to work each time dev wants to do a change in iet? 
cc @iM0ve  @pierre-unity

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: @phi-lira  / @Verasl  any thought?

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: @phi-lira  / @Verasl

com.unity.template-hd/Packages/com.unity.template.hd/package.json
```diff
@@ -7,9 +7,16 @@
   "unity": "2022.1",
   "description": "This template utilizes the High Definition Render Pipeline. Making it a good starting point for people focused on high-end graphics that want to develop games for platforms that support Shader Model 5.0 (DX11 and above). \n\nBeyond being tuned for High End visuals this project includes\n\u25aa Shadergraph \n\u25aa Visual Effect Graph  \n\u25aa Presets \n\u25aa Example content",
   "dependencies": {
+    "com.unity.collab-proxy": "1.9.0",
+    "com.unity.feature.development": "1.0.1",
+    "com.unity.textmeshpro": "3.0.6",
     "com.unity.cinemachine": "2.8.0",
+    "com.unity.feature.development": "1.0.1",
```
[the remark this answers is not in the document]
sebastienlagarde: thanks fixed

com.unity.template-universal/Packages/com.unity.template.universal/package.json
Verasl: Ah weird, I think they should be there, will make much more sense once we have the 2D renderer being used in the new Template, but for now I think this should indeed be added.

com.unity.template-universal/Packages/com.unity.template.universal/package.json
Verasl: This is will need someone more familiar with XR to chime in, looking a the name, it should be removed, but @sandy-carter-unity or @thomas-zeng will know for sure.

sebastienlagarde:
note: we have also bump version to 13.1.0 for master as a 13.0.0 have been publish and so we must not collide with it

com.unity.template-universal/Packages/com.unity.template.universal/package.json
phi-lira: I don't follow why we have some package dependencies duplicated in both `manifest.json` and `packages.json` is that expected? I'd expect that the dependencies in package.json should not be in manifest.json as the URP template package will import them indirectly anyway. If this is correct, the duplicated packages in manifest.json are redundant and should be removed.

com.unity.template-universal/Packages/com.unity.template.universal/package.json
phi-lira: If it should be removed it should be removed from manifest.json as well as otherwise that package will be imported.

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: packages.json is used when we develop,  manifest.json is used when we bundle the template into a packages and is the real thing that matter for users. But the test currently check that both are in sync

com.unity.template-hd/Packages/com.unity.template.hd/package.json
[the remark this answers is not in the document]
vicky-voin: You can keep this package in manifest.json only, this way internal developers can still make changes to IET without problems. Including it in package.json will include it in the final template, so end users will have access to it and will be able to make changes to the tutorial. The authoring package does not need to be included in package.json if the intention is only to present tutorials to end users.

com.unity.template-universal/Packages/com.unity.template.universal/package.json
vicky-voin: I believe it's the opposite re: package.json & manifest.json - manifest.json is essentially used for development and package.json is what the end users will see. This is why I wanted to double check that we wanted to add 2D packages, since they haven't been included in package.json before and therefore have not been a part of the template from the looks of it.

Template tests are checking that the packages are duplicated between manifest.json and package.json like Sebastien mentioned (with the exception of preview packages - these are only allowed in manifest.json), so it is expected for the packages to be duplicated between the two files.

com.unity.template-universal/Packages/com.unity.template.universal/package.json
fabien-unity: Not sure if com.unity.xr.legacyinputhelpers is required now. @ryanhy-unity  ?

com.unity.template-hd/Packages/com.unity.template.hd/package.json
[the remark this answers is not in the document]
sebastienlagarde: I see, ok

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: let those package for universal then, it don't harm and many people start from the template for new project (yes I know there is blank template now :))

phi-lira: I'll approve as I see no risk here, but would like to follow up on package.json and manifest.json and why it needs to be duplicated.

phi-lira (approved):

Verasl (approved):
LGTM

com.unity.template-universal/Packages/com.unity.template.universal/package.json
thomas-zeng: That package is still being used in URP XR. It's a very bad name since it has Legacy in it, but yeah it's something needed for setting up an XR Rig if you aren't using the XR Interaction Toolkit. The main behavior that package has is the TrackedPoseDriver.

thomas-zeng (approved):

com.unity.template-universal/Packages/com.unity.template.universal/package.json
ryanhy-unity: I think Thomas answered this in that this package can allow XR HMD head tracking when not using the new input system.  Its name is weird in that it uses legacy which is a bit of a misnomer I think.

vicky-voin (approved):

com.unity.template-universal/Packages/com.unity.template.universal/package.json
sebastienlagarde: thanks for the info.

PR #5800, opened 2021-09-24
[HDRP][Path Tracing] Slight robustness improvement for sample rescaling

eturquin:
When rescaling a sample for later reuse (in path tracing), there could be situations where, due to numerical imprecision, the sample value would reach 1.0, potentially causing issues in subsequent computations (NaNs, zeros, etc.). Such issues were spotted for instance in volumetric scattering. This PR addresses that problem.


remi-chapelain (changes requested):
For the fix in the PR, couln't find any issues after clamping the value ✔️ 

Using this PR to give feedback on top of [theses fixes](https://github.com/Unity-Technologies/Graphics/pull/5754). 
Theses improvements helps a lot what happens with volumetric fog in raster !

There is still some discrepencies, not linked to specific issues but just unsuported features of the volumetric fog override by the pathtracer. 
What I would like is to include those in the documentation. (Adding a Volumetric Fog section, under sub surface scattering section for example, [here](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@13.0/manual/Ray-Tracing-Path-Tracing.html)) 

Here's a list of what is currently not supported. 
Some are linked to how they are rendered in raster so would be expected (like the quality stuff) but others are just unsupported but could make sense in pathtracing pipeline:
- Density volume, now called **Local Volumetric Fog** (plays a large role is discrepencies in project where there is one setup)
- Shadow dimmer in Volumetric options in lights is ignored
- Ambient light probe dimmer
- Volumetric fog distance
- Slice deformation uniformity  
- Denoising mode
- All the quality controls 
- Directionnal light only checkbox
- Anisotropy


image

image

image

eturquin:
@sebastienlagarde, @remi-chapelain, I'll add a couple of these missing features that are easy to add (light and shadow dimmers, directional-only mode) to this PR, do not merge it just yet.

eturquin:
I added support for
- light/shadow dimmers (volumetric and non-volumetric)
- directional lights only volumetric scattering
The PR is now ready for review.

remi-chapelain (changes requested):
Thanks for the additions, theses are great to better match what happens in raster ! 🥳 
I tested and couldn't find any issues ✔️ 

However, would be great to just add a section in the doc of the pathtracer just to note the rest of the things unsupported for now : 
- Local volumetric fog (aka old density volume)
- Ambient light probe dimmer
- Volumetric fog distance
- Anisotropy
- And raster quality settings (denoising, quality control, slice deformation uniformity.. etc)

eturquin:
@remi-chapelain, the doc will be updated in another PR.

remi-chapelain (approved):

PR #5832, opened 2021-09-29
Fixes for Hair Reference and Multiple Scattering Approximation

johnpars:
### Purpose of this PR

This PR includes various fixes for our physically based hair shading, namely:
- Pathtraced hair reference is now validated against a ground truth ([PBRT](https://pbrt.org/)).
- Pathtraced hair reference is now importance sampled.
- Pathtraced hair reference now correctly supports both near and far field models (by default using near).
- Adds a furnace test to test for energy conservation of the hair reference. 
- Adds two new helpful color parameterization modes (Explicit absorption and melanin concentration)
- Fixes various subtle issues in our dual scattering.

image

This work began with the intent to fix some discrepancies I observed between our dual scattering approximation and our pathtraced reference. However I realized that even our reference itself had some issues, which made it an unreliable multiple scattering reference. As such, I began a process of validating our reference, so that we may ultimately validate our multiple scattering approximation. This PR is a culmination of everything that was necessary to achieve that goal. 

#### Ground Truth
The pathtraced reference was fixed with the help of [this](https://pbrt.org/hair.pdf) PBRT supplemental listing which shared helpful implementation details on the same model we use. Following that paper, we now are in good agreement with PBRT:

image

#### Importance Sampling
Importance sampling was also added, drastically reducing the variance we otherwise experience from uniform sampling. 

**64 SPP, BM = 0.8, BN = 0.7**
image

#### Near/Far
Fixed an issue where the offset fiber term from tube intersection was being incorrectly calculated, causing incorrect results of the near field model. The D'Eon far field model was also removed and we "fake" a far field by choosing a random offset for every sample (instead of integrating over the fiber width). 

image

#### Furnace Test
I added a furnace test to confirm that the reference is now energy conserving. The test is performed by setting up a theoretical fiber with zero absorption and confirming that the outgoing energy is close to 1 for uniform incident lighting. This is done for varying intervals of roughness. I added this test via our test runner framework but I leave it disabled for now (only enabled it locally for testing).

image

#### Color Modes
I added some other helpful color parameterization modes. Specifically: *Absorption* and *Melanin Concentration*. The absorption mode was added so that we can directly submit absorption coefficients to the model to reference our results against other papers. The melanin mode was added as an alternative but intuitive way to define the absorption in terms of pigmentation. I will keep these modes hidden in the UI from now, but can easily be enabled after some further discussion with artists (it will default to the normal color mode).

Color Modes

#### Dual Scattering Fixes 

Ultimately the above work made it easier to iterate on our multiple scattering approximation and easily determine if there were issues. Numerous issues were discovered and corrected. Overall the approximation should be much closer in agreement with the reference. 

Replacement

There are however still some unresolved problems: 
- Apparent energy loss at low roughness values, compared to the reference. It is almost necesarry to tune the strand count to something lower to counter act it. 
- Slightly different highlight width compared to reference, this may simply be due to the gaussian we use to approximate the longitudinal scattering. 

---
### Testing status
~

---
### Comments to reviewers
~

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
@@ -610,7 +610,7 @@ void GetHairAngleWorld(float3 V, float3 L, float3 T, inout HairAngle angles)
     // Projection onto the normal plane, and since phi is the relative angle, we take the cosine in this projection.
     float3 VProj = V - angles.sinThetaO * T;
     float3 LProj = L - angles.sinThetaI * T;
-    angles.cosPhi = dot(LProj, VProj) * rsqrt(dot(LProj, LProj) * dot(VProj, VProj) + 0.0001); // zero-div guard
+    angles.cosPhi = dot(LProj, VProj) * rsqrt(dot(LProj, LProj) * dot(VProj, VProj) + 1e-5); // zero-div guard
     angles.phi = FastACos(angles.cosPhi);
 
     // Fixed for approximate human hair IOR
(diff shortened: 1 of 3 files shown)


sebastienlagarde (approved):
@johnpars could you also update the documentation. In your various PR description there is a lot of informatoin which should be also in documentation. Could you add a new pages in doc about the path tracing hair and how to test the white furnace for example.

i.e overall we need users to be able to understand what the model provide :) can be done in a separate PR, thanks

sebastienlagarde:
For QA: the testing is covered by demo team.

PR #5837, opened 2021-09-29
[HDRP] Update eyes sample to show different scale setup

remi-chapelain:
### Purpose of this PR
This PR fixes [this issue](https://fogbugz.unity3d.com/f/cases/1368665/). The setup was fine but the text was misleading. 
This PR now shows different scale setup with correctly adjusted world scale diffusion profile. 

The text has been changed accordingly and it has been added that this graph is made only for this model and at this import scale factor to avoid possible confusions.  

image

> Eyeballs have a diameter of roughly 25mm. To use the diffusion profiles at a realistic scale, the world scale in the diffusion profiles needs to be adjusted accordingly. In this sample, the diffusion profiles world scale accommodates for eyeballs that have different diameters.
> 
> This graph is specifically made for this model with an import scale factor of one. A different graph needs to be created when using different models or scale factor.
> 
> To create lighting that works with small objects, use punctual lights instead of directional lights.

---
### Testing status
Loaded the sample into a fresh project. ✔️ 
Checked the other sample scenes visually to see if something was different. ✔️ 

---
### Comments to reviewers
🤷


anisunity (approved):
Pair validated during the change

sebastienlagarde (approved):

PR #5839, opened 2021-09-29
add tag on all shader

sebastienlagarde:
This PR add missing tag Tags{ "RenderPipeline"="HDRenderPipeline" } on all HDRP shaders. Required to improver cross pipeline project  build time


alex-vazquez (approved):
Many thanks

PR #5840, opened 2021-09-29
Fix potential NullReferenceException in URP/HDRP/SG material import

arttu-peltonen:
### Purpose of this PR
This PR fixes following cases:
- https://fogbugz.unity3d.com/f/cases/1349298/
- https://fogbugz.unity3d.com/f/cases/1359625/

The issue seems to be that sometimes `AssetDatabase.LoadAssetAtPath(asset, typeof(Material));` can produce a null material, even if `asset` is not null. This can happen for instance for assets that have the ".mat" suffix but are not actually of type Material. 

### Testing status
The NullReferenceException was not reproduced locally, so the fix is not verified. Should be minimal risk.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/AssetProcessors/MaterialPostProcessor.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/AssetProcessors/MaterialPostProcessor.cs
@@ -334,6 +334,8 @@ namespace UnityEditor.Rendering.HighDefinition
                 // Materials (.mat) post processing:
 
                 var material = (Material)AssetDatabase.LoadAssetAtPath(asset, typeof(Material));
+                if (material == null)
+                    continue;
 
                 if (MaterialReimporter.s_ReimportShaderGraphDependencyOnMaterialUpdate && GraphUtil.IsShaderGraphAsset(material.shader))
                 {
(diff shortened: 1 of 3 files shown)


RSlysz (approved):
lgtm

com.unity.render-pipelines.high-definition/Editor/AssetProcessors/MaterialPostProcessor.cs
```diff
@@ -334,6 +334,8 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
                 // Materials (.mat) post processing:
 
                 var material = (Material)AssetDatabase.LoadAssetAtPath(asset, typeof(Material));
+                if (material == null)
+                    continue;
```
phi-lira, line 338: I wonder if we should threat this an exception or log error. It feels risky we silently fail to process a material.

arttu-peltonen: Fair point, I'm not sure. As I don't have a repro case, I'm working under the impression that these are not really materials, but some other assets that happen to have a .mat extension. But I'm not sure if that's even allowed or makes sense in any valid use case. Does unity have strict 1:1 mapping between asset file extensions and asset types? Perhaps this needs some deeper investigation...

arttu-peltonen:
@Unity-Technologies/gfx-qa-urp I've returned this PR to draft status as per @phi-lira's comment I'm not sure anymore if this is a safe fix, if it might silently leave some materials unprocessed. Could you help me find someone to reproduce the original issue? I was unable to find a SG QA person to contact since the original reporter has left.

arttu-peltonen:
Upon further examination and discussion with @RSlysz, it seems like the fix proposed here is not the correct one. The reason for `LoadAssetAtPath` returning null seems to be related to AssetDatabase internals, and Rémi has separate ongoing work concerning that, so I'm going to close this PR and pass the bug to him. The issue seems to be connected to this bug: https://fogbugz.unity3d.com/f/cases/1347209/.

PR #5847, opened 2021-09-29
[Dots Deformations] Fix issue in linear blend skinning vertex shader deformations

svens-unity:
---
### Purpose of this PR
Zero init out data for deformed vertices to ensure correct end result. Not all platforms do this which results in the mesh disappearing. 

---
### Testing status
Test in windows built, behavior is unchanged.
Also tested on device, before the mesh isn't visible. After the mesh is visible and deforms as expected.

---
### Comments to reviewers


tomtoyeunity (approved):
Tested and LBS mesh now works correctly.
Testing doc here available here - https://docs.google.com/document/d/14doRmEqivPOPl3yxeS8PFEmxaXB5btz_7sKckWGW6NY/edit?usp=sharing

HedvigAxelsson (approved):

nigeljw-unity (approved):
LGTM

PR #5850, opened 2021-09-29
[URP] [Doc] Fix XR SPI full screen blitting example 

thomas-zeng:
---
### Purpose of this PR
- Addressed forum feedback reported here: https://forum.unity.com/threads/scriptable-render-feature-breaks-ui-masks-on-a-world-space-canvas.1175291/

---
### Testing status
- Manually tested the repro project(from forum post) with the fix and verified the UI mask is working as expected

---
### Comments to reviewers
- Example code is missing ConfigureTarget call. By calling ConfigureTarget, URP knows the feature's final target and will bind targets for the following passes if needed.


oleks-k (approved):
Looks good!

simon-engelbrecht-soerensen (approved):

eh-unity (approved):

PR #5855, opened 2021-09-30
[HDRP][Path Tracing] Added selection of light types for Unlit shadow mattes.

eturquin:
Addresses: https://fogbugz.unity3d.com/f/cases/1335487/
You can now select which light type(s) you want to compute mattes for (point/spot, area, directional).


patrickp-unity3d (approved):
Awesome, thanks! (inputs make sense, didn't review algo)

sebastienlagarde (approved):

PR #5859, opened 2021-09-30
Add missing DisallowMultipleComponent attribute in AdditionalData components

arttu-peltonen:
### Purpose of this PR
Fixes https://fogbugz.unity3d.com/f/cases/1365879/.

AdditionalData components are associated with components that themselves only allow one instance of the component per gameobject (Camera, Light and Reflection Probe) . Therefore the AdditionalData components need to follow the same rule - the attribute was missing from `HDAdditionalReflectionData` and `HDAdditionalLightData`, so this PR adds them.

---
### Testing status
Checked that after the fix, you can no longer add multiple instances of these additionaldata components via "paste component as new". 

Also I checked that the attribute were already present for `HDAdditionalCameraData` component, as well as `UniversalAdditionalCameraData` and `UniversalAdditionalLightData`. Therefore this should be low risk.


RSlysz (approved):
~~Did you check also the Camera?~~
LGTM :)

sebastienlagarde (approved):

alex-vazquez (approved):

PR #5013, opened 2021-06-30
Fix Lens Flare GLES2

skhiat:
---
### Purpose of this PR
Fix regression on GLES2 only.
Move a function where HLSLcc will convert to available instruction.
https://fogbugz.unity3d.com/f/cases/1343061/
https://fogbugz.unity3d.com/f/cases/1343443/

---
### Testing status
Have to do a testing on GLES2 hardware.

PR #5864, opened 2021-09-30
Fixing 16 bit issues with Universal RP and lens flare occlusion

skhiat:
Fixed a critical issue on android devices & lens flares.
16 bit texture not supported.

Add more comment on code.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
+++ b/com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
@@ -3,15 +3,6 @@
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Random.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Sampling/Sampling.hlsl"
 
-struct AttributesLensFlare
-{
-    uint vertexID : SV_VertexID;
-
-#ifndef FLARE_PREVIEW
-    UNITY_VERTEX_INPUT_INSTANCE_ID
-#endif
-};
-
 #if SHADER_API_GLES
 struct AttributesLensFlare
 {
@@ -23,6 +14,16 @@ struct AttributesLensFlare
 #endif
 };
 #else
+struct AttributesLensFlare
+{
+    uint vertexID : SV_VertexID;
+
+#ifndef FLARE_PREVIEW
+    UNITY_VERTEX_INPUT_INSTANCE_ID
+#endif
+};
+#endif
+
 struct VaryingsLensFlare
 {
     float4 positionCS : SV_POSITION;
@@ -33,7 +34,6 @@ struct VaryingsLensFlare
     UNITY_VERTEX_OUTPUT_STEREO
 #endif
 };
-#endif
 
 TEXTURE2D(_FlareTex);
 SAMPLER(sampler_FlareTex);


kecho (approved):

com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
```diff
@@ -125,10 +112,10 @@ float GetOcclusion(float ratio)
         if (all(pos >= 0) && all(pos <= 1))
         {
             float depth0 = GetLinearDepthValue(pos);
-#ifdef UNITY_REVERSED_Z
-            if (_ScreenPosZ < depth0)
+#ifdef SHADER_API_GLES3 // GLES3.hlsl do not define UNITY_REVERSED_Z properly
```
phi-lira, line 115: This seems a breaking change. GLES2 and GLCore also doesnt have reversed z. 
  What's the issue with REVERSED_Z in GLES3 file? This seems strange as we GLES3 automation on features that depend on reading depth in URP.

com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
```diff
@@ -3,17 +3,6 @@
...
-struct AttributesLensFlare
-{
-    float4 positionCS       : POSITION;
-    float2 uv               : TEXCOORD0;
-
-#ifndef FLARE_PREVIEW
-    UNITY_VERTEX_INPUT_INSTANCE_ID
-#endif
-};
-#else
```
phi-lira, line 16: What is the reason to remove this? it seems this was introduced to fix GLES2 compilation error.

phi-lira (changes requested):
Requesting changes due to the OpenGL platform questions.

com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
skhiat: This PR is merged:
https://github.com/Unity-Technologies/Graphics/pull/5013
So GLES2 is not supported for LensFlareSRP (cf. Discussion with Nigel on Slack).
The define "UNITY_REVERSED_Z" is not defined on GLES3.hlsl (should be 0 or 1).
With this code with URP GLES3:
rnJ9B5NkcE
With URP GLES2 and only this if:
if (COMPARE_DEVICE_DEPTH_CLOSER(depth0, _ScreenPosZ))
QY77dQMAbR

L.et me know If I missed something.

com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
skhiat: [Fixed]

com.unity.render-pipelines.core/Runtime/PostProcessing/LensFlareCommonSRP.cs
```diff
@@ -8,15 +8,34 @@ public sealed class LensFlareCommonSRP
...
+        /// If this value change that could implies an implementation modification on:
+        /// com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/LensFlareMergeOcclusionDataDriven.compute
+        /// </summary>
         public static int maxLensFlareWithOcclusionTemporalSample = 8;
 
+        /// <summary>
+        /// Set to 1 to enable temporal sample merge.
+        /// Set to 0 to disable temporal sample merge (must support 16 bit textures, and the occlusion merge must be written in the last texel (vertical) of the lens flare texture.
+        /// </summary>
         public static int mergeNeeded = 1;
```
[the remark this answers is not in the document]
skhiat: Here to fix the "missing doc issue"

com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
phi-lira: I wonder if we made a mistake here then, the define was removed intentionally so we have both #if and #ifdef usages working in SRP. https://unity.slack.com/archives/C20E70CAC/p1602760055081400

Either way, you should not check for GLES3 as GLCore also needs this, if there's problem with define the GLES3, GLES2 and GLCORE api defines should be fixed instead. 

Let's have a chat over that.

skhiat: [Fixed]

com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
```diff
@@ -3,28 +3,26 @@
...
-    float2 uv               : TEXCOORD0;
+    uint vertexID : SV_VertexID;
 
 #ifndef FLARE_PREVIEW
     UNITY_VERTEX_INPUT_INSTANCE_ID
 #endif
 };
-#else
+
+#if SHADER_API_GLES
```
phi-lira, line 15: This change is still wrong, you will have double declaration of AttributeLensFlare and no declaration of Varyings in GLES.
 
 master is correct to me: https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl#L6-L25

phi-lira (changes requested):

phi-lira (approved):

com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/LensFlareCommon.hlsl
skhiat: [Fixed]

phi-lira (approved):

aleksandrasdzikia (approved):

kecho (approved):

PR #5866, opened 2021-09-30
Improving performance of FPTL algorithm by 0.3 ms on console.

kecho:
### Purpose of this PR
This PR contains 2 micro optimizations:
1 - Internal loop of FPTL is now dynamic, and earlies out. This generates nicer assembly on consoles (ps4) and saves around 0.2 ms with just 20 lights. Savings are higher with more lights.

2- Throwing the volume and coarseIdx into LDS to reduce ALU cost and some bandwidth cost. Lights are only loaded once per wave, this saves around 0.1 ms

Gains:
 * Reduced VALU cost by 5%
 * Reduced total memory access
 * Increased occupancy of FPTL by 2 waves

Total savings .3ms.

### Testing
This change should be completely safe and I expect no difference in visuals.

For QA: not sure if any further testing is needed, other than a sanity check with a few ligth types on a scene: we gotta make sure the tiles make sense and that we dont get any checkerboard artifacts (unless of course you exceed the light limit)

Before: 
image

After:
image

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/lightlistbuild.compute
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/lightlistbuild.compute
@@ -398,6 +398,22 @@ int GetCoarseLightIndex(int l, int iNrCoarseLights)
     return l < iNrCoarseLights ? GenerateLightCullDataIndex(coarseList[l], g_iNrVisibLights, unity_StereoEyeIndex) : 0;
 }
 
+groupshared uint s_lightVolumesCache[LIGHT_LIST_MAX_COARSE_ENTRIES];
+
+void StoreLightVolumeCache(int lightIndex, int coarseIndex, uint volumeType)
+{
+    // 3 bits for the volume type, in case we have a corrupted one we can early out of the switch statement.
+    // 29 bits for a coarse light index.
+    s_lightVolumesCache[lightIndex] = (volumeType & 0x7) | (coarseIndex << 3);
+}
+
+void LoadLightVolumeCache(int lightIndex, out int coarseIndex, out int volumeType)
+{
+    int data = s_lightVolumesCache[lightIndex];
+    coarseIndex = data >> 3;
+    volumeType = data & 0x7;
+}
+
 // initializes ldsNrLightsFinal with the number of accepted lights.
 // all accepted entries delivered in prunedList[].
 #if PIXEL_PER_THREAD == 4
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/lightlistbuild.compute
```diff
@@ -414,12 +435,97 @@ void FinePruneLights(uint threadID, int iNrCoarseLights, uint2 viTilLL, float vL
...
+        StoreLightVolumeCache(threadID, idxCoarse, uLightVolume);
+    }
+
+#if NR_THREADS > PLATFORM_LANE_COUNT
+    GroupMemoryBarrierWithGroupSync();
+#endif
+
+    //When using LDS to cache the volume data, this produces the best most optimal code.
+    //Doing a manual loop like the one below adds an extra cost of .1 ms on ps4 if we use LDS.
+    for (int l = 0; l < iNrCoarseLights; ++l)
```
kecho, line 453: I dont understand why, but the wave compiler  generates much better code when we do the loop naively on top of LDS. 
The biggest win so far has been the uVal extra check in the PIXEL_PER_THREAD, not just the early out, but also the fact that it forces the loop to be dynamic, which is consistent with the loop below.
I tried having the loop below (the old one) wiht the dynamic inner loop + on top of LDS, and it generates slower code. So this is a win of .1 ms, on top of the .2 ms that we get from the dynamic inner loop.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/lightlistbuild.compute
```diff
@@ -26,6 +26,7 @@
 
 #define FINE_PRUNING_ENABLED
 #define PERFORM_SPHERICAL_INTERSECTION_TESTS
+#define USE_LIGHTVOLUME_CACHE
```
kecho, line 29: Should I disable this for switch? since I have only tested on msft/sony consoles.

FrancescoC-unity: We don't support switch

FrancescoC-unity: I would delete the old code path, or if we want to keep it for other reasons (Don't see any if you tested on consoles and it is always faster :) ) we should factor out the processing of the light given that it looks the same as before, just the outer part of the loop is different and at any change we'd have to maintain two copies of the same code. 

But then again, I vote for deleting old path instead :D

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/lightlistbuild.compute
FrancescoC-unity: Did you check what gets generated for Xbox? FXC can lead to very different results than wave often

kecho: Checked xbox. Neutral. Deleting the old code.

TomasKiniulis (changes requested):
Did a quick check on Editor and Windows standalone with Spot(Cone, Pyramid, Box), Point, Area lights, there's no visual issues indeed, but doing player build throws "lightlistbuild" warnings which doesn't occur on master:
image (16)

kecho:
@TomasKiniulis fix for compiler warning is now in. Thanks again. Let me know if you find anything else.

TomasKiniulis (approved):
Thanks for the fix @kecho! It's perfect now

mmikk (approved):
Nice improvement on wave and register count! It looks good to me and thank you for doing the extra perf tests we discussed too.

PR #5868, opened 2021-10-01
Hair Documentation Pass

johnpars:
### Purpose of this PR
This PR adds updated documentation for the upcoming hair Physical model.

Section explaining the two models.
image
image
image

Section explaining geometry type.
image

Section explaining hair in path tracing. 
image

One additional change was to the nomenclature of the Hair "material types". They are now:
- Kajiya-Kay -> **Approximate**
- Marschner -> **Physical**

---
### Testing status
I built the docs with the DocTools and verified everything was formatted correctly.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPropertyBlocks.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPropertyBlocks.cs
@@ -17,7 +17,7 @@ namespace UnityEditor.Rendering.HighDefinition.ShaderGraph
     {
         class Styles
         {
-            public static GUIContent materialType = new GUIContent("Material Type", "TODO");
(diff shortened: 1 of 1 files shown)


johnpars:
Adding images shortly

merpheus-dev:
It is a convenient simplification! Although, I would propose putting a side note somewhere to describe what Basic and Physical would refer to as their hair shading model.
Along with that, the term "Basic" might get a bit misleading for the general audience.

johnpars:
Thanks @merpheus-dev, I will incorporate this feeback!

johnpars:
We decided to switch the nomenclature to **Approximate / Physical** for both the **Material Type** and the **Scattering Mode** for consistency.

merpheus-dev:
Awesome!

johnpars:
@sebastienlagarde yamato is now green - I forgot to update the reference image for OSX in the previous PR. It has been updated and the test re-ran.

sebastienlagarde:
@Vic-Cooper merging this PR as it include change in the feature that need to be in 21.2. Could you do a correction pass in a separate PR? thanks

PR #5880, opened 2021-10-04
Enable iris normal for Eye shader

adrien-de-tocqueville:
### Purpose of this PR

Fix https://fogbugz.unity3d.com/f/cases/1367911/

Iris normal was not enabled in the graph options, now it is to match the Eye samples.

Also renamed the properties to have spaces in their names, and moved them to categories

before:
before
after:
after

---
### Testing status

verified that iris normal now works by default with the eye shadergraph.
verified that shader properties didn't change name in the shader
ran yamato


remi-chapelain (changes requested):
Great change ! 
But unfortunately, importing the material samples and opening the eye scene does not include the change. I think it's because the sample uses another graph bundled with the sample called [SG_Eye](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.high-definition/Samples~/MaterialSamples/Shadergraphs/SG_Eye.shadergraph) (which is indentical execept your latest changes). 

What would be best ? To make the same changes in this graph or try to use the graph you changed in RenderPipelineResources for the samples ?

adrien-de-tocqueville:
I would say using the HDRP Eye shader n the sample is the best, to avoid issues like the one in the fogbugz appear again.
But maybe it's important for users to have an editable graph in the sample, in which case we should duplicate it. Do you have an idea if it's the case ?

remi-chapelain (approved):
All good ! ✔️

anisunity:
@adrien-de-tocqueville think to add me on the review next time you change something related to the eye shader! Thanks

anisunity (approved):

PR #5882, opened 2021-10-04
[SRP] Fix issue when changing volume profiles at runtime with a script

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1364256/

Some volume components of a volume profile can be destroyed externally (for example by the editor when exiting play mode, if they were created at runtime).  In this PR I have added a method to check and remove the "zombies".  

---
### Testing status
I have tested using the repro scene. The nullref error does not repro anymore after my change.


JulienIgnace-Unity (approved):

remi-chapelain (approved):
Fix makes sense and testing seems sufficient given the specific use case. ✔️

sebastienlagarde (approved):

PR #5883, opened 2021-10-04
Srp workflows/align overdraw workflow

fredericv-unity3d:
### Purpose of this PR
This align the overdraw debugging workflows for both URP and HDRP. (Initiated by Fogbugz [1365666 ](https://fogbugz.unity3d.com/f/cases/1365666/))

HDRP:
- Updated the color feedback
Unity_pn9UtJDGKE

URP:
- Added Overdraw Debug Mode and Overdraw Max Pixel Cost
image
- Overdraw Debug Mode All
Unity_OCZdUodJ0d
- Overdraw Debug Mode Transparent Only
Unity_kjIFla6isC
- Overdraw Debug Mode Opaque Only
Unity_GlDR0TOnzz

---
### Testing status
Manual Test:
- Created a set of transparent and opaque cube overlapping each other, one per renderer (total of 4 sets)
- Activated all overdraw mode, one by one to make sure the color feedback is correct

---
### Comments to reviewers

HDRP: I updated the debug shader to use the Shader Core Library for the feedback color. The change is minimal here.

URP: I added a specific code to render the overdraw debug mode in several place. Can you check that it fit appropriately in URP?


com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsRendering.cs
```diff
@@ -40,6 +40,30 @@ bool overdraw
...
+
+        int m_MaxPixelCost = 10;
+
+        internal int maxPixelCost
+        {
+            get => m_MaxPixelCost;
+            set
+            {
+                m_MaxPixelCost = value;
+                UpdateDebugSceneOverrideMode();
```
RSlysz, line 63: Not sure why you need this as it will only check wireframeMode which is not what you changed here.

com.unity.render-pipelines.universal/ShaderLibrary/Debug/DebugViewEnums.cs
```diff
@@ -113,6 +113,19 @@ public enum DebugSceneOverrideMode
...
+    /// </summary>
+    public enum DebugOverdrawMode
+    {
+        /// <summary>Debug overdraw of opaque only.</summary>
+        Opaque,
+        /// <summary>Debug overdraw of transparent only.</summary>
+        Transparent,
+        /// <summary>Debug overdraw of everything only.</summary>
+        All
+    }
```
arttu-peltonen, line 127: The convention for other debug modes is to have the first entry here be "None". This would tell whether the mode is enabled, whereas now you kept the separate boolean `overdraw` for this purpose. I would prefer to do it this way instead to be consistent with other debug modes. The UI will be a little simpler too (you would display the overdraw dropdown always, and the `isHidden` condition for the `max cost` sub-option would be `overdrawMode == DebugOverdrawMode.None`.

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsRendering.cs
arttu-peltonen: Agree, this seems redundant. The one above you'd need to keep, if you get rid of the separate `overdraw` boolean.

com.unity.render-pipelines.universal/Runtime/Debug/DebugHandler.cs
```diff
@@ -147,7 +149,7 @@ internal void SetupShaderProperties(CommandBuffer cmd, int passIndex = 0)
             {
                 case DebugSceneOverrideMode.Overdraw:
                 {
-                    cmd.SetGlobalColor(k_DebugColorPropertyId, new Color(0.1f, 0.01f, 0.01f, 1));
+                    cmd.SetGlobalColor(k_DebugColorPropertyId, new Color(1f, 1f, 1f, 1));
```
arttu-peltonen, line 152: `Color.white` would probably work here

arttu-peltonen (changes requested):
Looks great to me, I love the new color scheme and especially the handy legend bar. I would just prefer the separate `overdraw` boolean be removed on URP, to simplify both code & UI.

RSlysz (approved):

com.unity.render-pipelines.universal/Runtime/Debug/DebugDisplaySettingsRendering.cs
```diff
@@ -57,11 +45,7 @@ internal DebugOverdrawMode overdrawMode
         internal int maxPixelCost
         {
             get => m_MaxPixelCost;
-            set
-            {
-                m_MaxPixelCost = value;
-                UpdateDebugSceneOverrideMode();
-            }
+            set => m_MaxPixelCost = value;
         }
```
arttu-peltonen: Tiny suggestion :) not sure but I think this + backing field could now be simplified to just `internal int maxPixelCost { get; set; } = 10;`

arttu-peltonen (approved):
Nice!

JulienIgnace-Unity (approved):

Nzollner (approved):
LGTM

martint-unity:
My question would be for @TheoWong-pixel is the color scheme a Unity Standard?

arttu-peltonen:
FYI there is a set of 12 UX-verified colorblind-friendly colors in [Debug.hlsl](https://sourcegraph.ds.unity3d.com/github.com/Unity-Technologies/Graphics/-/blob/com.unity.render-pipelines.core/ShaderLibrary/Debug.hlsl?L4). Though I don't think they help in this case where you need a variable number of colors, and the colors are calculated analytically.

oleks-k:
> @oleks-k does "Max Pixel Cost" make sense for the user?

I can't answer this question right now, I need to understand better what it does first (I would say we should try to be more direct, try to replace "cost" with something). I need to focus on remaining 2021.2 tasks, and I created a JIRA ticket for this task:
https://jira.unity3d.com/browse/DOCR-378

fredericv-unity3d:
As discussed offline, this PR requires legends to add meaning to the color bar. I will work on it before validation.

sebastienlagarde:
Hey, any update on this PR?

arttu-peltonen:
@sebastienlagarde the author is on leave for another 3 weeks so this will probably have to wait.

phi-lira (approved):

phi-lira:
@TheoWong-pixel is this PR ready to go? @fredericv-unity3d can you make sure to run URP PR?

TheoWong-pixel:
Still waiting for some UI to indicate the range for the colours.

TheoWong-pixel (approved):

PR #5884, opened 2021-10-04
Fixed the clouds not taking properly into account the fog when in distant mode and with a close far plane (case 1367993).

anisunity:
https://fogbugz.unity3d.com/f/cases/1367993/
The fog was not properly applied to volumetric clouds when local mode was unchecked and the far plane was closer than the position of the clouds. This fixes it and the behavior is closed to what is expected.

Testing status:
Tested the various combinations (local/distance, far/close far plane). Looks fine now.


anisunity:
@sebastienlagarde not sure which version we need to backport this to as clouds are not in all versions

remi-chapelain (approved):
Quickly tested the initial issue and made sure that volumetric fog was also included in this fix ✔️

anisunity:
Fixed the formatting, will launch the tests.

PR #5887, opened 2021-10-04
Revert to non-C#8 switch case syntax

theopnv:
### Purpose of this PR
Breaking formatting when vendoring to core 
```
check_complex_statements(889): Packages\com.unity.render-pipelines.core\Editor\Debugging\DebugUIDrawer.Builtins.cs, orig_line is 612, Error: Expected '(', got '{' for 'SWITCH'
ERROR: Error formatting file: Packages/com.unity.render-pipelines.core/Editor/Debugging/DebugUIDrawer.Builtins.cs

** IMPORTANT: one or more formatters hit fatal parse errors (see above). Usually this is caused by badly constructed macros in C++, but it can also be caused by bugs or missing features in a formatter. Check #devs-format for support if you can't figure it out.
```
https://unity.slack.com/archives/GAULB1M5J/p1633343834137400

### DIFF
--- a/com.unity.render-pipelines.core/Editor/Debugging/DebugUIDrawer.Builtins.cs
+++ b/com.unity.render-pipelines.core/Editor/Debugging/DebugUIDrawer.Builtins.cs
@@ -614,7 +614,7 @@ namespace UnityEditor.Rendering
                 case DebugUI.MessageBox.Style.Info:
                     type = MessageType.Info;
                     break;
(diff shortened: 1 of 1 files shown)


arttu-peltonen (approved):

theo-pnv:
Merging while keeping the tests running - so that I can prepare the srp2core PR. At least it will avoid the formatting error and allow preparing the PR. If there are problems with the tests we can fix them on top of the srp2core branch.

PR #5888, opened 2021-10-04
Fix APV issue spewing asserts when baking with max subdiv level of 2

FrancescoC-unity:
I found this issue while repro another problem, took a while to find out what was going on. 
Essentially if putting simplification level to 2, then the calculations to relative brick index/positions are not behaving as the new index algorithm expects. 

To solve this, we make sure the "voxel" (unit used to update index buffer) is never bigger than the maximum brick we can represent.


FrancescoC-unity:
Oh I need to revert one commit, one sec.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #5889, opened 2021-10-04
Fixed initial decal position

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1368987/

When projectors are created, they are moved to be visible in the current scene view.
Changed the behavior to match builtin for regular components. It now depends on if the object has a parent and if the "Create Objects at Origin" in the editor preferences is enabled.

---
### Testing status

tested that decal projector position is the same as when creating a point light under various cnoditions (preference option, parent or not)


remi-chapelain (approved):
Don't think anymore testing is needed for this one ! ✔️

sebastienlagarde (approved):

PR #5892, opened 2021-10-04
Fix range compression factor clamping

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1365707/

color from reflection probes was clamped after undoing the range compression factor, which didn't result in the original value before range compression.
Moved the uncompression to be done after clamping, which should still prevent inf from propagating

Now:
image

### Testing status
tested the repro case
High emissive values with high enough compression factor on the probe now match between the real object and the reflected object


FrancescoC-unity (approved):

remi-chapelain (approved):
All good now that we can make the reflection match with a high enough compression factor.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
```diff
@@ -152,15 +152,18 @@ float4 SampleEnv(LightLoopContext lightLoopContext, int index, float3 texCoord,
             color.rgb = SAMPLE_TEXTURECUBE_ARRAY_LOD_ABSTRACT(_EnvCubemapTextures, s_trilinear_clamp_sampler, texCoord, _EnvSliceSize * index + sliceIdx, lod).rgb;
         }
 
+        // Planar and Reflection Probes aren't pre-expose, so best to clamp to max16 here in case of inf
+        color.rgb = ClampToFloat16Max(color.rgb);
```
sebastienlagarde, line 156: That's wrong, you must clamp after 

color.rgb *= rangeCompressionFactorCompensation;

 // Planar and Reflection Probes aren't pre-expose, so best to clamp to max16 here in case of inf
        color.rgb = ClampToFloat16Max(color.rgb);
        
        otherwise rangeCompressionFactorCompensation become fully useless... rangeCompressionFactorCompensation is to currentlty allow to fit large range.

sebastienlagarde (approved):

PR #5893, opened 2021-10-04
Fix another issue with probe volume subdiv multiplier when changing to a lower subdiv level

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1367914/ though it also needs this https://github.com/Unity-Technologies/Graphics/pull/5888  

The issues were two, one is the one fixed in the other PR, one is due to the fact that if a probe volume had a max subdiv override set to a val that is higher than the maximum subdiv allowed, multipliers could go above 1, leading to all sorts of problems.


sebastienlagarde (approved):

JulienIgnace-Unity (approved):

FrancescoC-unity:
Yup I kinda expected this, will fixup

alelievr (approved):

PR #5901, opened 2021-10-04
Fixed bad performance when manipulating probe shape in editor.

JulienIgnace-Unity:
### Purpose of this PR
Fixed an issue where probe being forced realtime would cause perf to degrade greatly when manipulating them in editor.
The root cause is that we were forcing the builtin (not used in HDRP) "mode" to realtime under the hood. This caused massive slowdown due to forcing re-rendering the whole probe each frame when manipulating the probe gizmos.

The reason we were doing this is to fix a culling issue there was in A2. I was not able to reproduce any culling issue in my test scenes but there might be more complicated cases. It might also be that we fixed that some other way since then and the workaround is not required anymore.

This is a fix for https://fogbugz.unity3d.com/f/cases/1368553/

---
### Testing status
Tested the repro project and ran graphics tests.

---
### Comments to reviewers
@fredericv-unity3d do you have the original fogbugz case regarding the culling issue in A2? Might be helpful checking for regression.
@Unity-Technologies/gfx-qa-hdrp it might be good to thoroughly test culling of reflection probes in various configuration.

If no culling regression is found by QA (Waiting on Jans for A2 testing) we can probably safely merge this fix.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/SerializedHDReflectionProbe.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/SerializedHDReflectionProbe.cs
@@ -39,9 +39,6 @@ namespace UnityEditor.Rendering.HighDefinition
 
         internal override void Apply()
         {
-            // Force the mode to real time so its influence is properly culled by the camera.
-            legacyMode.intValue = 1;
-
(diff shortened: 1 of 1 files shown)


fredericv-unity3d:
Sorry, I don't find the fogbugz again

PR #5902, opened 2021-10-04
[Fogbugz # 1365687] Fixing depth pyramid with multicamera on hardware drs.

kecho:
### Purpose of this PR
AO and SSR in Hardware DRS is currently broken when we have scene view + gameview camera. To reproduce ensure that we have a proper dx12 scene.

This fix makes it so that the pyramid depth is done per camera, and not globally. The issue was coming from multicamera support, stomping the state of the previous camera, since the depth pyramid was global.

Fogbugz: https://fogbugz.unity3d.com/f/cases/1365687/

Before:
image

After:
image

---
### Testing status

Tested on a small test scene. Enabled / disabled drs and saw the fix. Its important to have dx12 and also gameview + scene view up at the same time.

#### Testing for QA
For qa the following testing points are suggested:

* For drs please have dx12 enabled and use hardware DRS. Ensure too to have both scene view and gameview displayed at the same time.
* Please reproduce and test fix on the HDRP template. Dont test refraction, will address that separately
* Test ambient occlusion
* Test screen space reflections
* Test screen space GI
* Test ray tracing AO / reflections

Anything that is dependant on pyramid depth should be regressed. The above suggestions are the minimal things I'd expect to be tested, feel free to add more. There might be problems still with the Pyramid Color but these will be worked on in a separate bug / PR.

---
### Comments to reviewers
NOTE: color pyramid might also be a problem, but will address it in a separate PR.


FrancescoC-unity (approved):

kecho:
@remi-chapelain we need a separate bug for SSGI since I dont think is related to hardware DRS / depth pyramid. Id say we close this PR and move to separate bugs for these, just like we did for color pyramid.

PR #5904, opened 2021-10-05
Depth Texture clearing fix for Overlay Camera

jonuuukas:
---
### Purpose of this PR
---
Removing SetGlobalTexture to black/white for Overlay cameras, as it's main intention is to reclear it every frame and it's guaranteed Base camera will render first and do this. For Overlay cameras - there is specific checkbox that forces clear of the depth.

This also changes behaviour for failing tests, as it seems like depth is recreated from scratch and might create ordering problems (needs more investigation, but it's hard due to no repro on local devices) so disabling Depth Texture for the overlay cameras in URP_PostPro tests 145 as they only need previous camera's depth and no need to recreate it

### Testing status
---
URP_PostPro is green, tested manually, running rest of Universal_Split

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
@@ -722,6 +722,7 @@ namespace UnityEngine.Rendering.Universal
             }
 
             // Set the depth texture to the far Z if we do not have a depth prepass or copy depth
+            // Don't do this for Overlay cameras to not lose depth data in between cameras (as Base is guaranteed to be first)
             if (cameraData.renderType == CameraRenderType.Base && !requiresDepthPrepass && !requiresDepthCopyPass)
             {
                 Shader.SetGlobalTexture(m_DepthTexture.id, SystemInfo.usesReversedZBuffer ? Texture2D.blackTexture : Texture2D.whiteTexture);


com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -722,7 +722,7 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             }
 
             // Set the depth texture to the far Z if we do not have a depth prepass or copy depth
-            if (!requiresDepthPrepass && !requiresDepthCopyPass)
+            if (cameraData.renderType == CameraRenderType.Base && !requiresDepthPrepass && !requiresDepthCopyPass)
```
martint-unity, line 725: Could you add to the comment above the reasoning behind this?

martint-unity (changes requested):

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
jonuuukas: added comment in e6049c62a06b9f1374cc18d24c8bf3e1a4bb16a2

martint-unity (approved):
LGTM

ernestasKupciunas (approved):
Approving without testing. The developer has successfully tested his changes.

PR #5910, opened 2021-10-05
Fixed removal of depth buffer binding

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1372920/

A recent PR removed the depth buffer binding for the cloud layer, which resulted in cloudlayer appearing on top of geometry.

---
### Testing status

Tested that cloudlayer doesn't appear on top of geometry anymore
Updated test ase to include opaque objects


remi-chapelain (changes requested):
Issue introduced by this PR should have been caught by automation. 
Discussed a bit with @adrien-de-tocqueville and HDRP cloud Layer test will be update to have geometry between clouds and camera to prevent a future regression.

PR #5911, opened 2021-10-05
Fix shadow mask fade and optimize it at same time 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1369324/ 

The fade calculations were a bit off given that it was relying on culling spheres and not the max shadow distance.   What we were using before was also extremely slow for what it was. 

**What did I test**: the repro (adding a border fade too to notice an actual fade going on :P )

**Before**: 

https://user-images.githubusercontent.com/43168857/136002862-e55bc299-b1f7-41fe-9401-3710a259957e.mp4

**After:**

https://user-images.githubusercontent.com/43168857/136002897-b2b0d8ba-3453-42d8-b5ab-5a081e4a8d8b.mp4


remi-chapelain (approved):
Thanks for adding before/after video. Really help to understand the issue quickly. 
Looks good to me. ✔️

radishface:
Looks good to me. Thanks for fixing this!

PR #5912, opened 2021-10-05
Fix wrong format for contact shadows and shadow atlas default textures 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1355108/  

+ other D3D validation layers errors because default textures set when a shadow atlas is not rendered to were not shadow textures and those are not compatible with shadow samplers leading to a lot of validation errors. 

**What did I test**: I checked with D3D validation layer, I had the warnings before but not after. 
**FOR QA:** Could you try the original repro on Mac?

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed ambient occlusion strenght incorrectly using GTAOMultiBounce
 - Fixed range compression factor being clamped. (case 1365707)
 - Fixed tooltip not showing on labels in ShaderGraphs (1358483).
+- Fixed and optimize distance shadowmask fade.
 - Fix API warnings in Matcap mode on Metal.
 - Fix D3D validation layer errors w.r.t shadow textures when an atlas is not used.
 
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
@@ -95,6 +95,8 @@ namespace UnityEngine.Rendering.HighDefinition
         [SurfaceDataAttributes(precision = FieldPrecision.Real)]
         public Vector4 shadowMaskSelector;      // Used with ShadowMask feature
 
+        public Vector2 cascadesBorderFadeScaleBias;
+
         public float diffuseDimmer;
         public float specularDimmer;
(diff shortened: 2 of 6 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
```diff
@@ -40,6 +40,15 @@ static void BindGlobalLightingBuffers(in LightingBuffers buffers, CommandBuffer
             cmd.SetGlobalTexture(HDShaderIDs._ScreenSpaceShadowsTexture, buffers.screenspaceShadowBuffer);
         }
 
+        static void BindDefaultTexturesLightingBuffers(RenderGraphDefaultResources defaultResources, CommandBuffer cmd)
```
JulienIgnace-Unity, line 43: I'm not sure I'm a fan of binding everything globally when the plan is to remove all globals like this.
Can't we just return the proper default texture from the right passes as we do some other places?

FrancescoC-unity: This just mirrors the others that are bound globally (see a bit above in the file), the problem is that the set global and all the passes producing the outputs are skipped during debug. 

The passes already return the right default it is just that they are never run :D

JulienIgnace-Unity: Fair enough. I'll take a deeper look when I actually do the thing.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -109,6 +109,8 @@ public partial class HDRenderPipeline
                     // For alpha output in AOVs or debug views, in case we have a shadow matte material, we need to render the shadow maps
                     if (m_CurrentDebugDisplaySettings.data.materialDebugSettings.debugViewMaterialCommonValue == Attributes.MaterialSharedProperty.Alpha)
                         RenderShadows(m_RenderGraph, hdCamera, cullingResults, ref shadowResult);
+                    else
+                        HDShadowManager.BindDefaultShadowGlobalResources(m_RenderGraph);
```
johnpars, line 113: For awareness, we added this extra coverage here (scene lighting disabled caused more validation errors).

PR #4034, opened 2021-03-30
Allow copy depth pass when in game view with gizmos

nigeljw-unity:
Copy depth pass was being skipped in game view when rendering gizmos so draw order was not being respected.

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1302504/

---
### Testing status
Tested against sample provided in case 1302504 and ABV testing

---
### Comments to reviewers
Follow up change for https://github.com/Unity-Technologies/Graphics/pull/3272

PR #5913, opened 2021-10-05
Fix error when using gizmos with camera stack in editor

tomzig16:
### Purpose of this PR
This PR fixes: https://fogbugz.unity3d.com/f/cases/1347472/
Previously there was an assert that was used as camera stack was not implemented, but now it works with it. When PR was made to change the if statement, assertion was forgotten hence it causes the regression. After removing it everything works correctly.

---
### Testing status
Tested locally + automated ABV tests

---
### Comments to reviewers
Although bug report tells that it is not a regression, we found that the issue appeared indirectly after the changes made in #4034
I will check for a need of backports as this issue was introduced through a different change.


tomzig16:
Not sure, but it seems like this PR breaks PostPro tests, looking into it.

PR #5915, opened 2021-10-05
Fixes for light anchor

adrien-de-tocqueville:
### Purpose of this PR

fix for https://fogbugz.unity3d.com/f/cases/1362809/
Added a property when an anchor override is selected to offset the anchor position (in the anchor object space)
light_anchor

at the same time, fix for https://fogbugz.unity3d.com/f/cases/1362802/
Added a max distance value clamp to 10000, because the math instability looked very linked with how the component works

---
### Testing status

Tested the repro steps in both bugs


skhiat (approved):
Look good to me for now.
We'll discuss the "clamp" on slack.
Thanks

PR #5921, opened 2021-10-05
Minor performance improvements to SSGI (case 1367144).

anisunity:
https://fogbugz.unity3d.com/f/cases/edit/1367144/
Unfortunately as expected all the suggestions by the user do no improve the performance at all. I've improved the performance a bit, but it is towards for intrinsic based platforms (ps4/xbox1).

Testing status
Tested the performance improvement on console and on PC. The difference is more important on console and remains meh on PC unfortunately. tested with full/half res effect and half res denoiser. We need to run the tests mainly to validate.


com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/BilateralUpsample.compute
```diff
@@ -10,78 +9,103 @@
...
 // The half resolution texture that needs to be upscaled
 TEXTURE2D_X(_LowResolutionTexture);
 
-// Constant buffer where all variables should land
-CBUFFER_START(UnityScreenSpaceGlobalIllumination)
-    float4 _HalfScreenSize;
-    float2 _DepthPyramidFirstMipLevelOffset;
-CBUFFER_END
+// LDS that store the half resolution data
+groupshared float3 gs_cacheLighting[36];
```
anisunity, line 20: Compressing the result to 11 11 10 makes the kernel overall slower unfortunately.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceGlobalIllumination.compute
```diff
@@ -135,16 +135,27 @@ void TraceReflectionProbes(PositionInputs posInput, float3 normalWS, float3 rayD
     GetCountAndStart(posInput, LIGHTCATEGORY_ENV, envLightStart, envLightCount);
     totalWeight = 0.0f;
 
+    uint envStartFirstLane;
```
anisunity, line 138: Force scalarization makes the kernel twice faster on ps4 as it is VGPR bound by the reflection probe code.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/Denoising/DiffuseDenoiser.compute
```diff
@@ -117,15 +118,7 @@ void BILATERAL_FILTER(uint3 dispatchThreadId : SV_DispatchThreadID, uint2 groupT
...
-        float2 newSample;
-        newSample.x = GetLDSequenceSampleFloat(sampleIndex + sampleOffset, 0);
-        newSample.y = GetLDSequenceSampleFloat(sampleIndex + sampleOffset, 1);
-
-        // Convert the sample to a local unit disk
-        newSample = SampleDiskCubic(newSample.x, newSample.y);
-
-        // Distribute them according a square profile
-        newSample *= denoisingRadius;
+        float2 newSample = _PointDistribution[sampleIndex] * denoisingRadius;
```
anisunity, line 121: Improves a bit the performance to save those two texture reads.

FrancescoC-unity (approved):

remi-chapelain (commented):
- Tested if SSGI was working the same ✔️ 
- Tested quickly performance of the SSGI markers and found improvement between 5% and 10% 

If possible, it'd be great the remove the marker SSGIConvert which apprently does not make sense anymore since the Forward Emissive rework. 
Also, the marker SSGIDenoise always show **zero** since SSGI is using the ray tracing denoiser.

PR #5922, opened 2021-10-05
Fix auto exposure flickering on the right eye

fcoulombe:

### Checklist for PR maker

---
### Purpose of this PR
Fix auto exposure flickering on the right eye https://fogbugz.unity3d.com/f/cases/1336238/
multipass mode renders with a single eye since it does it in two passes.
using xrActiveEye will ping pong between 0 and 1 but here we always want to pass 0 since we have a single eye.

---
### Testing status
trunk, 2020.3.11f1, 2019.4.30f1
https://unity-ci.cds.internal.unity3d.com/job/9175200
https://unity-ci.cds.internal.unity3d.com/job/9175186
https://unity-ci.cds.internal.unity3d.com/job/9175557

---
### Comments to reviewers

### DIFF
--- a/com.unity.postprocessing/PostProcessing/Runtime/PostProcessLayer.cs
+++ b/com.unity.postprocessing/PostProcessing/Runtime/PostProcessLayer.cs
@@ -1210,6 +1210,7 @@ namespace UnityEngine.Rendering.PostProcessing
             context.uberSheet = uberSheet;
             context.autoExposureTexture = RuntimeUtilities.whiteTexture;
             context.bloomBufferNameID = -1;
+
             if (isFinalPass && context.stereoActive && context.stereoRenderingMode == PostProcessRenderContext.StereoRenderingMode.SinglePassInstanced)
                 uberSheet.EnableKeyword("STEREO_INSTANCING_ENABLED");
 
@@ -1299,6 +1300,7 @@ namespace UnityEngine.Rendering.PostProcessing
             if (context.bloomBufferNameID > -1) cmd.ReleaseTemporaryRT(context.bloomBufferNameID);
 
             cmd.EndSample("BuiltinStack");
+
             return tempTarget;
         }
 


com.unity.postprocessing/PostProcessing/Runtime/Effects/AutoExposure.cs
```diff
@@ -121,12 +121,14 @@ void CheckTexture(int eye, int id)
 
         public override void Render(PostProcessRenderContext context)
         {
+            // multi pass and singlepass only has a single eye
+            int eye = (context.numberOfEyes>1) ? context.xrActiveEye : 0;
```
mikesfbay, line 125: The current eye is already passed into RenderBuiltins(context, ..., eye) before calling autoExposure.Render(), which we can use to determine whether in multipass or SPI mode. Can you try passing in current eye to autoExposure.Render() instead of calculating it here?

fcoulombe: FYI i was debating between changing the Render() signature to receive an eye or to store it in the context.

since xrActiveEye was already in the conext, it felt right to put it next to it so that if someone needs one eye and it's not the right one... he's not gonna have issue finding the other eye.

it also avoided modifying the signature of every PPFX

com.unity.postprocessing/PostProcessing/Runtime/PostProcessLayer.cs
```diff
@@ -1182,6 +1182,7 @@ int RenderBuiltins(PostProcessRenderContext context, bool isFinalPass, int relea
             context.uberSheet = uberSheet;
             context.autoExposureTexture = RuntimeUtilities.whiteTexture;
             context.bloomBufferNameID = -1;
+            context.currentEye = eye;
```
mikesfbay, line 1185: For single-pass we can just set context.xrActiveEye = eye here instead of creating another eye variable (currentEye). Otherwise, there'll be 3 eye variables if we count the one in PostProcessLayer's for-eye loop. 
You can save and restore old value of xrActiveEye here if they are different from camera.

com.unity.postprocessing/PostProcessing/Runtime/PostProcessLayer.cs
```diff
@@ -1182,6 +1182,9 @@ int RenderBuiltins(PostProcessRenderContext context, bool isFinalPass, int relea
             context.uberSheet = uberSheet;
             context.autoExposureTexture = RuntimeUtilities.whiteTexture;
             context.bloomBufferNameID = -1;
+            int xrActiveEyeBackup = context.xrActiveEye;
+            if (context.stereoRenderingMode == PostProcessRenderContext.StereoRenderingMode.MultiPass)
+                context.xrActiveEye = eye;
```
mikesfbay, line 1187: Since depth-of-field and TAA also use xrActiveEye, can you also test/double-check those effects to make sure they still work in multipass with this change?

fcoulombe: I changed the code so that it's only applied to autoexposure. 
It shouldn't affect t'as and depth of field

mikesfbay:
Can you add a QA to verify (Jason or Ryan)? Otherwise, LGTM

PR #5929, opened 2021-10-06
[HDRP] Change RenderGraph Begin/Execute function pattern to avoid leaks

alelievr:
### Purpose of this PR
Fix an error where render graph resources were not released because of partial execution of the render graph when switching to a platform that doesn't support compute shaders (like WebGL).

This PR is part of the fix for https://fogbugz.unity3d.com/f/cases/1367363/.

---
### Testing status
Switch platformed from Windows to WebGL and looked for render graph release resource errors: no issue after the fix.

https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/core%252Ffix%252Frendergraph-exception-handling/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/9297640/job


remi-chapelain (approved):
Testing seems sufficient for this one. 
The rest of the fix is done in [this PR](https://github.com/Unity-Technologies/Graphics/pull/5918)

PR #5938, opened 2021-10-06
Add camera type mask to Render Objects

pbbastian:
### Purpose of this PR
Case [1361557](https://fogbugz.unity3d.com/f/cases/1361557/) ([public](https://issuetracker.unity3d.com/product/unity/issues/guid/1361557/)) identified an issue with the layer of objects in previews. This is being changed such that preview objects use layer 0 (Default) instead of layer 1 (TransparentFX). This means that renderer features might become active in the material preview. To allow for control over this, this PR is adding a mask for camera type to Render Objects. The feature will only run if the camera matches the mask.

image

---
### Testing status
Tested functionality locally


TheoWong-pixel:
We should have analytics added to this renderer feature, in general, for all the fields. We may do another pass in the future to simplify the workflow / hide some of the settings under "additional settings" 

analytics will be in a later PR.

TheoWong-pixel (approved):

Verasl (commented):
I feel like this should be under the 'event' property as the filter group is more for filtering of the objects themselves, less so the filtering/settings of the pass itself

aleksandrasdzikia (approved):

PR #5941, opened 2021-10-06
[SRP backport 21.2] [HDRP][URP] update manifest of SRP templates (#5799)

sebastienlagarde:
### Purpose of this PR
Backport https://github.com/Unity-Technologies/Graphics/pull/5799

---
### Testing status
To be done

---
### Comments to reviewers

For reviewver, this PR is using package 12.1.0 of SRP. We need to wait that the SRP2Core PR which include this bump land. Then we will merge this PR and publis/promote the template.


com.unity.template-hd/Packages/manifest.json
```diff
@@ -1,14 +1,14 @@
...
   "dependencies": {
     "com.unity.cinemachine": "2.8.0",
-    "com.unity.ide.rider": "3.0.7",
-    "com.unity.ide.visualstudio": "2.0.11",
-    "com.unity.ide.vscode": "1.2.3",
+    "com.unity.collab-proxy": "1.9.0",    
+    "com.unity.feature.development": "1.0.1",
     "com.unity.inputsystem": "1.1.0-pre.5",
     "com.unity.learn.iet-framework": "2.0.0",
     "com.unity.render-pipelines.high-definition": "12.0.0",
```
vicky-voin, line 8: The version in package.json seems to be different, it is likely going to be flagged by Yamato when it runs

com.unity.template-universal/Packages/manifest.json
```diff
@@ -2,15 +2,10 @@
   "dependencies": {
     "com.unity.2d.sprite": "1.0.0",
     "com.unity.2d.tilemap": "1.0.0",
-    "com.unity.ads": "3.7.5",
+    "com.unity.ads": "3.7.1",
```
vicky-voin, line 5: Don't know if I missed it on the previous PR, do we need this package? (it is not one of the default packages we need to add)
If yes, it should be included in package.json as well and if not - would be great if we can remove it from the manifest.

vicky-voin (changes requested):

com.unity.template-universal/Packages/manifest.json
sebastienlagarde: looks like a bad merge from me, thanks for pointing it

com.unity.template-hd/Packages/manifest.json
sebastienlagarde: good catch

vicky-voin (approved):

PR #5945, opened 2021-10-06
Fixed references to probes not cleared when unloading a scene

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1357459/

HDRP stores probes in a global array, but when a probe is removed, the reference is not removed from the array.
Also the probe was added to a lightmapper callback at OnEnable but not removed during OnDisable

---
### Testing status

tested the repro steps, the probes don't appear anymore in the memory profiler

For QA: HDRP also has a global array for planar probes and lights. So I cleared references for planar probes, I suppose they have the same issue but didn't try to test. I didn't manage to repro anything similar for lights though, so didn't change anything for them


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1610,6 +1610,16 @@ void PrepareGPULightdata(CommandBuffer cmd, HDCamera hdCamera, CullingResults cu
             }
         }
 
+        void ClearUnusedProcessedReferences(CullingResults cullResults, HDProbeCullingResults hdProbeCullingResults)
+        {
+            for (int i = cullResults.visibleLights.Length; i < m_ProcessedLightData.size; i++)
```
sebastienlagarde, line 1615: severity: Error
	message: C:\build\output\Unity-Technologies\Graphics\com.unity.render-pipelines.high-definition\Runtime\Lighting\LightLoop\LightLoop.cs(1615,64): error CS0103: The name 'm_ProcessedLightData' does not exist in the current conte
	
	With @kecho change this don't exist anymore. @kecho  can you look at this PR know what we need to change?

TomasKiniulis (changes requested):
Did a quick check to validate lights, reflection probes and planar probes, no repro indeed with lights and cube reflection probes anymore, but planar probes fill up Scene Memory -> Render Texture section when switching scenes.

https://user-images.githubusercontent.com/50582134/140769209-a1311a00-3a65-4e00-959a-0ed159d021e4.mp4

TomasKiniulis (approved):
Thanks for the fix! Not encountering any leaks anymore shuffling through the test project with additional lights and planar probes

PR #5953, opened 2021-10-07
[HDRP][Path Tracing] Added proper support for interleaved tiling

eturquin:
Modified camera ray generation slightly to accomodate interleaved tiling, following an internal request from Verticals.


patrickp-unity3d:
Actually, @eturquin I think I must use the viewproj matrix because many shaders depend on the screen space position. Chromatic aberration as well as the street in the above image need it to be in full resolution. If I don't offset the viewproj matrix these will not use the correct pixel. Unless I am very wrong, I think we either need to do as discussed originally or permit both strategies.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingMain.raytrace
```diff
@@ -90,68 +89,90 @@ void MissMaterial(inout PathIntersection pathIntersection : SV_RayPayload)
...
-        float focusDistance = _PathTracedDoFConstants.y;
-        float t = focusDistance / dot(directionWS, _ViewMatrix[2].xyz);
-        float3 focusPointWS = GetPrimaryCameraPosition() - t * directionWS;
+        // Use planar clipping, to match rasterization
+        float dotDirection = dot(cameraDirection, ray.Direction);
+        ray.TMin /= dotDirection;
 
-        // The new ray direction should pass through the focus point
-        directionWS = normalize(focusPointWS - cameraPosWS);
+        ApplyDepthOfField(pixelCoord, dotDirection, ray.Origin, ray.Direction);
```
pmavridis, line 147: Should we use here "pixelCoord * tileCount + tileIndex" instead of "pixelcoord" to avoid correlation in lens sampling? Or perhaps I missed something?

eturquin: No you're right, thanks! I'll fix that.

patrickp-unity3d (approved):
Looks good, I tested depth of field as well

Here's some detail at the transition (screengrab via Parsec at 150%)
image

sebastienlagarde (approved):

eturquin:
> Hi! So if I understand correctly this will be in HDRP 12.2 and 13.2?

Yes, in the first patch for 12.2.

PR #5961, opened 2021-10-08
[2022.1][URP] Decal fixes

lukaschod:
---
### Purpose of this PR
Fix cases:
- Fixed decals to work with native render pass [case 1353141]
- Fixing decals rendering to be correct with render scale [1353885]

Depends on https://github.com/Unity-Technologies/Graphics/pull/5369

---
### Testing status


phi-lira:
Reopening this got closed by a mistake that caused the universal/staging to be deleted.

simon-engelbrecht-soerensen (approved):

PR #5976, opened 2021-10-11
[VFX] Fix Verify_Orphan_Dependencies_Are_Correctly_Cleared

PaulDemeulenaere:
---
### Purpose of this PR
Fix case [1370626](https://fogbugz.unity3d.com/f/cases/1370626/) : missing parent assignation on VFXData.
We were only relying on the sanitize process which assign lazily the parent https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.visualeffectgraph/Editor/Data/VFXData.cs#L118

---
### Testing status
Yamato 🟢 
Tested locally:
image

---
### Comments to reviewers
See also this [conversation](https://unity.slack.com/archives/G1BTWN88Z/p1633362452272000)


com.unity.visualeffectgraph/Editor/Data/VFXData.cs
```diff
@@ -141,6 +141,16 @@ public virtual bool CanBeCompiled()
...
         }
 
+        public void OnRegisteredContextHasBeenAdded(VFXContext context)
+        {
+            if (!m_Owners.Contains(context))
+                throw new InvalidOperationException(string.Format("Unexpected OnRegisterContextHasBeenAdded from context {0} to {1}", context, this));
+
+            var newGraphParent = context.GetGraph();
+            if (newGraphParent != null && m_Parent != newGraphParent)
+                m_Parent = newGraphParent;
```
julienf-unity, line 151: What if the data has other contexts that were not (yet) moved?
Probably we should move only if the data has no parent yet?
There still the potential issue of having contexts from different graph linked together. We should enfore it's not possible

com.unity.visualeffectgraph/Editor/Data/VFXData.cs
```diff
@@ -108,15 +107,17 @@ protected internal override void Invalidate(VFXModel model, InvalidationCause ca
...
         public override void Sanitize(int version)
         {
             base.Sanitize(version);
-
-            if (m_Parent == null)
-            {
-                string assetPath = AssetDatabase.GetAssetPath(this);
-                m_Parent = VisualEffectResource.GetResourceAtPath(assetPath).GetOrCreateGraph();
-            }
+            m_Parent = null;
```
julienf-unity, line 120: Are we sure that having a null parent has no implication at all?

julienf-unity: To me, it means invalidation happening on the VFXData won't be propagated to parent graph anymore, which can cause issues

PaulDemeulenaere: The Invalidate is already iterating over owners and it doesn't need to send an invalidation to the main graph (see [here](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.visualeffectgraph/Editor/Data/VFXData.cs#L102). 
So far, I didn't find access to m_Parent in our current implementation
Let me double check this behavior (beside the implicit GetGraph implementation) ⏳

PaulDemeulenaere: I have double checked:

- m_Parent is acceded while calling AddChild/RemoveChild on VFXModel (see [here](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.visualeffectgraph/Editor/Models/VFXModel.cs#L157)), it will never be used for VFXData
- m_Parent is directly returned by [GetParent](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.visualeffectgraph/Editor/Models/VFXModel.cs#L196), beside editor test, it's used with [context](https://github.com/Unity-Technologies/Graphics/blob/fcffae0178a4cc07a9b600e35b2eb9723b69b631/com.unity.visualeffectgraph/Editor/GraphView/Elements/Controllers/VFXContextController.cs#L186)), in [VFXDataAnchorController](https://github.com/Unity-Technologies/Graphics/blob/fcffae0178a4cc07a9b600e35b2eb9723b69b631/com.unity.visualeffectgraph/Editor/GraphView/Elements/Controllers/VFXDataAnchorController.cs#L413) (we aren't expecting rendered slot on VFXData) and [VFXParameterNodeController](https://github.com/Unity-Technologies/Graphics/blob/fcffae0178a4cc07a9b600e35b2eb9723b69b631/com.unity.visualeffectgraph/Editor/GraphView/Elements/Controllers/VFXParameterNodeController.cs#L351) (which call it from a slot)

Then, we have a GetParent access in VFXModel with [GetFirstOfType](https://github.com/Unity-Technologies/Graphics/blob/fcffae0178a4cc07a9b600e35b2eb9723b69b631/com.unity.visualeffectgraph/Editor/Models/VFXModel.cs#L201) which is called either on owner of [VFXData](https://github.com/Unity-Technologies/Graphics/blob/f4079dcbbca4d0f847224140ac5644ff273306c1/com.unity.visualeffectgraph/Editor/Data/VFXData.cs#L285) or [VFXSlot](https://github.com/Unity-Technologies/Graphics/blob/e2cb5e6f60f2ef2047f0fd317ee512162c88360f/com.unity.visualeffectgraph/Editor/Compiler/VFXCodeGenerator.cs#L346)

All in all, the having a not declared parent is wrong approach, I think the main issue was about the GetGraph() behavior but it now returns the first graph of the first owner (see [here](https://github.com/Unity-Technologies/Graphics/pull/5976/files/a673f5c30fb34d307a22511ed836f571c980e4d2#diff-fe97c1cc58fb20b96676af540e5d17f8f1839b44cba93e99b2b11a8bdd180208R110))

PaulDemeulenaere:
Move to https://github.cds.internal.unity3d.com/unity/unity/pull/15053

PR #5982, opened 2021-10-12
Add API to completely disable Runtime Debug UI (#5339)

arttu-peltonen:
### Purpose of this PR

Adding backport of https://github.com/Unity-Technologies/Graphics/pull/5339 (case https://fogbugz.unity3d.com/f/cases/1361811/).

---
### Testing status
Tested manually that adding a MonoBehaviour script with following code correctly disables all hotkeys & access to Debug UI:

```
void Awake ()
{
    DebugManager.instance.enableRuntimeUI = false;
}
```

---
### Comments to reviewers

The fogbugz ticket was erroneously marked as resolved, even though the fix only got submitted into master and nothing was backported. 

Question: Should I add 21.1 backport too? The original report (https://fogbugz.unity3d.com/f/cases/1345783/) only has backport requests for 21.2 and 20.3.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Debug-Window.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Debug-Window.md
@@ -21,6 +21,8 @@ To open the Render Pipeline Debug window in the Editor, go to  **Window > Render
 
 You can display read-only items such as the FPS counter independently of the **Render Pipeline Debug** window. This means that when you disable the **Render Pipeline Debug** window, they are still visible in the top right corner of the screen. This is particularly useful if you want to track particular values without cluttering the screen.
 
+You can disable the runtime UI entirely by using the [enableRuntimeUI](https://docs.unity3d.com/Packages/com.unity.render-pipelines.core@10.8/api/UnityEngine.Rendering.DebugManager.html#UnityEngine_Rendering_DebugManager_enableRuntimeUI) property.
+
 ### Navigation at runtime
 
 To change the current active item:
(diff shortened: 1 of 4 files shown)


com.unity.render-pipelines.high-definition/Documentation~/Render-Pipeline-Debug-Window.md
```diff
@@ -21,6 +21,8 @@ To open the Render Pipeline Debug window in the Editor, go to  **Window > Render
 
 You can display read-only items such as the FPS counter independently of the **Render Pipeline Debug** window. This means that when you disable the **Render Pipeline Debug** window, they are still visible in the top right corner of the screen. This is particularly useful if you want to track particular values without cluttering the screen.
 
+You can disable the runtime UI entirely by using the [enableRuntimeUI](https://docs.unity3d.com/Packages/com.unity.render-pipelines.core@10.8/api/UnityEngine.Rendering.DebugManager.html#UnityEngine_Rendering_DebugManager_enableRuntimeUI) property.
```
arttu-peltonen, line 24: Is this URL OK - it doesn't exist right now but I assume it should get generated automatically? Is 10.8 the correct version to begin with?

sebastienlagarde: That's a good question. I don't know if there is better way to reference API doc, @Vic-Cooper  any thought?
here the question is how to get the latest doc of a given version (10.x). For this particular case here there is no risk as this API will not change in future 10.x version, but would be good to know if it is something doable.

arttu-peltonen: Checked with Vic and the original link is okay. @sebastienlagarde this is good to go on my side.

arttu-peltonen: Got it, makes sense. So sounds like xref links would have been OK in this case too. Bit of a bummer though that it's untestable before it goes live.

PR #5988, opened 2021-10-12
Acceleration Structure setup using RayTracingAccelerationStructure.CullInstances

INedelcu:
Use RayTracingAccelerationStructure.CullInstances function (latest trunk - 2022.a13)  to filter Renderers and populate the acceleration structure used for ray tracing (DXR). This greatly increases CPU performance on the main thread when using ray tracing effects.

On my PC with i9 with 12 cores, the performance improvement can vary between 8X and 30X or more, depending on scene size.
On a quad core laptop the CPU speedup for acceleration structure setup on the HDRP template scene is around 16x.

More information about testing and CPU profiling is on this PR which was merged already to trunk : https://ono.unity3d.com/unity/unity/pull-request/132659/_/graphics/raytracing/rtas-cull-instances

HDRP DXR tests on trunk and this branch is green: https://unity-ci.cds.internal.unity3d.com/job/9294644

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.HDRTASManager.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.HDRTASManager.cs
@@ -158,7 +158,7 @@ namespace UnityEngine.Rendering.HighDefinition
 
         public RayTracingInstanceCullingResults Cull(HDCamera hdCamera, in HDEffectsParameters parameters)
         {
-            // The list of instanceTestArray needs to be cleared every frame as the list depends on the active effects and their parameters. 
+            // The list of instanceTestArray needs to be cleared every frame as the list depends on the active effects and their parameters.
             instanceTestArray.Clear();
 
             // Set up the culling flags


anisunity (approved):

remi-chapelain (approved):
Here's what I tried: 
- Adding pathtracer and testing the dirtiness of most change (transform, materials, volume...) to reset accumulation ✔️ 
- Checked performance in the HDRP template (not much difference, makes sense since building RTAS is probably not the bottleneck here) 
- Tested performance (FPS) with huge number of objects ✔️ (details below)
- Let the project run in player and editor a bit to see if performance was staying the same ✔️ 
- Checked RaytracingBuildAccelerationStructure new markers timings. Going from 52.88ms to 4.02ms so a 13x decrease ! ✔️ 

Before
image

After
image

_10k emissive sphere in a box._
Without PR: average frame time **225ms**
With PR: average frame time **136ms**
40% decrease ! ✔️ 
https://user-images.githubusercontent.com/57442369/137454155-6b7b8a30-434e-4650-9904-9e85028bf6f1.mp4

__Complex static scene lighted only with emissive objects__
Without PR: average frame time 196ms
With PR: average frame time: 99ms
50% decrease ! ✔️ 
RemiCHAPELAIN_Direct Lighting + Probe + RTGI + RTR

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingManager.HDRTASManager.cs
```diff
@@ -0,0 +1,270 @@
...
+    };
+
+    class HDRTASManager
+    {
+        public RayTracingAccelerationStructure rtas = null;
+        public RayTracingInstanceCullingConfig cullingConfig = new RayTracingInstanceCullingConfig();
+        public List<RayTracingInstanceCullingTest> instanceTestArray = new List<RayTracingInstanceCullingTest>();
+
+        // Culling tests
+        RayTracingInstanceCullingTest ShT_CT = new RayTracingInstanceCullingTest();
```
sebastienlagarde, line 34: Sorry but from the comment I don't understand why this code is here? 
But I guess "Test" here doesn't mean "Testing" right? i.e not part of our test framework

anisunity: yeah not part of the test framework, it's just the set of tests that are used to inject into the RTAS @sebastienlagarde

PR #5990, opened 2021-10-12
[VFX] Activation slots in blocks

julienf-unity:
### Purpose of this PR
This PR implements the activation ports in block to allow a better/more convenient way to enable/disable blocks with graph.
Activation expressions can be evaluated on both CPU and GPU and works for spawner blocks. Also for compile time constant, an optimized path is used.

activationslot

In addition to the core feature, this PR also:
- Fixes some useless recompilation triggered when dragging an edge out of an output slot connected to some compilable system
- Add bool <> uint, int, float casts
- Optimize the reduction of logical expressions

TODOs:
- Some minor UI issues (small offset when collapsing/uncollapsing, separation mark still visible when collapsed...)
- Still have to optimize a bit underneath (remove shader constants...)
- Some graphics / editor tests
- Doc update
- [EDIT] activation slots work on blocks within subgraph but not on subgraph itself atm

One of the drawback of the PR is that it adds one serialized slot per block which will make VFX assets bigger

---
### Testing status
Tested locally.
Some automated tests required

---
### Comments to reviewers
Requires C++ to work: https://katana.ds.unity3d.com/projects/Unity/builders?hide_unstable=true&unity_branch=graphics%2Fvfx%2Ffeature%2Fspawner-activation-ports&sort=1-asc

As this PR changes some thing at serialization level, extensive QA pass needs to test copy/paste, subgraphs and stuff like that.


PaulDemeulenaere:
About `Still have to optimize a bit underneath (remove shader constants...)`, if you are referring to GPU constant which ends in flatten expression graph, you can potentially retrieve changes from : https://github.cds.internal.unity3d.com/unity/vfx-graphics/pull/157

peeweek (changes requested):
So far I've tested the system and it feels simple, intuitive and useful! Some remarks following (these should be addressed in order to provide a streamlined editing experience)

- Currently, the graph evaluation of the block toggle boolean ports triggers a recompile of the graph, (unless it's an exposed property) which makes the editing quite cumbersome : changing numeric values in float/uint fields trigger numerous unwanted recompilations, similarly to the experience in forced runtime mode. While in edit mode the editing of these expressions should not trigger a recompile and would also ease the toggling on/off of blocks without requiring recompiling the graph.

- Bug/unimplemented: subgraph block toggles don't work currently (note: it's still WIP as discussed with @julienf-unity )

PR #5996, opened 2021-10-12
[HDRP] Physically based depth of field optimizations

pmavridis:
### Purpose of this PR

The purpose of this PR is to optimize the performance of the physically based DOF implementation. In particular, DOF regions that are out-of-focus are now computed at lower resolution, while in-focus regions retain the full resolution. For many scenes this results in significant speedup, without any visible reduction in image quality.

It the following image:  
- green correspond to perfectly focused tiles (CoC < 1 pixel). These are not processed at all by DoF.
- blue correspond to "fast tiles" (CoC > 2.5 pixels), these are processed at half res since they are "blurry" by definition.
- red corresponds to tiles with both in-focus and out-of-focus regions. These are processed at full res, as before.

image
So the speedup comes from the green and blue regions.

And here is how the image looks (note that there areno discontinuities in tile boundaries)
image

The PR also switches to a faster noise function, since the previous blue noise one was causing a measurable perf drop. 
(Using defines in the code we can switch between 4 different noise functions, I enabled the fastest one)

---
### Testing status
- I have run the graphics tests and updated the reference images related to this mode of DoF (everything looks ok)
- I have tested the optimized pb DoF when moving around in the template scene.

### Performance numbers
Using these settings:
- With optimizations, total pp time: **~3.8 ms**
- Without the optimizations, total pp time: **~10.3ms**

Measured with the Rendering Debugger while in play mode using 7 near/far samples (which is rather high):
image

---
### Comments to reviewers
Many DoF-related functions where moved from the DoFGather shader to a new DoFGatherUtils file, in order to be shared with the new combine shader. The functions themselves did not change in a significant way.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFGatherUtils.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DoFGatherUtils.hlsl
@@ -99,14 +99,15 @@ CTYPE GetColorSample(float2 sampleTC, float lod)
 {
 #ifndef FORCE_POINT_SAMPLING
     float texelsToClamp = (1u << (uint)ceil(lod)) + 1;
-    float2 uv = ClampAndScaleUVPostProcessTexture(ResScale * sampleTC * _PostProcessScreenSize.zw, _PostProcessScreenSize.zw, texelsToClamp);
+    float2 uv = ClampAndScaleUVPostProcessTexture(ResScale * (sampleTC + 0.5) * _PostProcessScreenSize.zw, _PostProcessScreenSize.zw, texelsToClamp);
 
     // Trilinear sampling can introduce some "leaking" between in-focus and out-of-focus regions, hence why we force point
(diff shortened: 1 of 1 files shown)


sebastienlagarde (approved):

pmavridis:
@remi-chapelain border issue should be fixed now, can you have a quick look again?

remi-chapelain (approved):
Found another issue that I'll report later (unrelated to the PR) but the issue with the black border is fixed. Approving ✔️

remi-chapelain (changes requested):
Approved too fast, works with a set resolution but not in free aspect 

https://user-images.githubusercontent.com/57442369/140491602-85c9cdd7-c788-4099-86f4-555714b5f15d.mp4

remi-chapelain (approved):
Approving ✔️ 

Previous issue is a very specific edge case and we did not manage to repro it on several other machines. 
We also found another very specific edge case that will be invesitgated later but don't want to block the PR for this, I'll make a report.

PR #6001, opened 2021-10-13
Fixed a regression in the transparent SSR color pyramid usage.

anisunity:
While doing this PR https://github.com/Unity-Technologies/Graphics/pull/5898 I made a mistake because i thought we had a least one color pyramid ready before the SSR transparent which is not the case. This fixes the issue and the tests.

Testing status
The result looks fine and the tests are now passing.


remi-chapelain:
all good, already discussed with anis.

PR #6007, opened 2021-10-13
Fix issues with shadows and volumetric fog on PS5 (potentially PS4 too)

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1347303/ 

As discussed with @peterjohnlong , the issue is due to the fact that the light loop in volumetric lighting is not guaranteed to produce scalar light reads, as such we can't force the scalarization at shadow reading time. 

@peterjohnlong  could you please verify if this fixes the repro project?


peterjohnlong (approved):
I confirm that I tested these changes on PS5 applied to 10.x.x and they fixed the issue in the customers 2020.3 repo project provided in https://fogbugz.unity3d.com/f/cases/1347303/

PR #6008, opened 2021-10-13
Add missing shader stages in shader keyword copying 

FrancescoC-unity:
As reported by @olivierprat , the shader stages were not copied alongside the keyword. This one liner fixes just that


sebastienlagarde:
Yamato is green with same issue as on master; PR is simple enough and have been validated by Olivier. Merging.

PR #6015, opened 2021-10-14
fix for freeze in renderpipeline converter

martint-unity:
---
### Purpose of this PR
Fix for this FB Case
https://fogbugz.unity3d.com/f/cases/1353885/
When converting using renderer settings converter it froze if there were more than one setting to convert. This fixes that.

---
### Testing status
Manual tested the bug retro steps from the fogbugz case.


ellioman (approved):

RSlysz (commented):
Well, I don't get why this solve your issue. But it seams to help.

@bastienunity can I have your feedback on this change? AFAIK it should speed up having the assetdatabase be in asset editing mode in order to batch change written on disc. It seams in some case it can cause issue.

phi-lira (approved):

PR #6019, opened 2021-10-14
Enable FrameTimingManager tests in standalone builds

arttu-peltonen:
### Purpose of this PR

EDIT: I had misunderstood the code, the UI change is not necessary because iOS works the same as other platforms. I've repurposed this PR to just be about improving the tests.

1) ~Add actionable warning message that informs user why they are seeing only 0 in each value. This is because FrameTimingManager is not available.~
~**Note:** This message is be rarely visible because for most platforms, FrameTimingManager is always enabled in development builds (and Rendering Debugger is only accessible in development builds). However, the message should appear on **iOS development builds when Frame Timing Stats is disabled in Player Settings**. What it should look like:~

2) Enable RT profiler unit test in player builds, now that there is player-visible API for checking whether FrameTimingManager is enabled. Before this API was added, these tests had to be disabled in player builds because it was not possible to query if frame timings are available.

---
### Testing status

EDIT: UI changes are removed, so manual testing is not necessary here, I will focus on ensuring Yamato is fine with the test changes.

~Please verify the following things using an **iOS development build**:~
~1) The warning message is **shown** and the frame stats are 0, when Frame Timing Stats is **disabled** in Player Settings.
2) The warning message is **not shown** and the frame stats show valid numbers, when Frame Timing Stats is **enabled** in Player Settings.~

~**NOTE: Testing requires building latest version of trunk (feature relies on new trunk API)**~

### DIFF
--- a/.yamato/_abv.yml
+++ b/.yamato/_abv.yml
@@ -75,7 +75,7 @@ Safety_net_trunk:
         TEST_FILTER: .*
         CACHE_ARGS: '{{cache.flags}}'
     triggers:
-        expression: pull_request.target eq "master" AND NOT pull_request.draft AND pull_request.changes.any match [".yamato/**/*", "com.unity.*/**/*", "TestProjects/**/*"]
+        expression: pull_request.target eq "master" AND NOT pull_request.draft AND pull_request.changes.any match [".yamato/**/*", "com.unity.*/**/*", "TestProjects/**/*"] AND NOT pull_request.push.changes.all match ["*.md", "**/*.md", "doc/**/*", "**/Documentation*/**/*"]
     dependencies:
       -  path: .yamato/shadergraph-win-dx11.yml#ShaderGraph_Win_DX11_playmode_mono_Linear_trunk
          rerun: on_new_revision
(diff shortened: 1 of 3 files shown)


alex-vazquez (approved):
Looks good :)

ernestasKupciunas (commented):
It doesn't work as intended. Even the FrameTimingStats checkbox was disabled in the Player Settings, I was still able to see the stats and no warning appeared.
Tested with:

2022.1.0a15.2209
Revision: trunk 49120632d05b
Built: Tue, 09 Nov 2021 06:50:39 GMT

arttu-peltonen:
@ernestasKupciunas thanks for checking. I've investigated and it appears I've misunderstood, and there's no need for this warning after all, iOS works the same as other platform which is great. I'll remove the UI change from this, and repurpose this PR to just be about improving the unit test, which I still want to do.

arttu-peltonen:
This PR is now ready, in the end I'm just modifying the runtimeprofiler tests:
- Changed to "yield return null" because WaitForEndOfFrame is not supported in batchmode.
- Tests disabled on Linux due to https://fogbugz.unity3d.com/f/cases/1370861/

Checked that the test is now skipped on Linux and passes on other platforms.

alex-vazquez (approved):
Looks, good. We should create a task to reenable the test for Linux Platform?

arttu-peltonen:
"Restoring test on Linux" tracked here: https://jira.unity3d.com/browse/XPIPELINE-451

PR #6030, opened 2021-10-15
Mixed cached shadows for directional lights

FrancescoC-unity:
This PR implements mixed cached shadows (allow always draw dynamic) for directional lights too.  Turned out to be less trivial than I originally thought :D  It was a bit tricky to get the 2 set of rendered data to agree with the fact that they are drawn with different view positions. To solve this  we warp the dynamic ones to be rendered in the camera space of the cached shadows. 

In this Pic, the shadows are set to OnDemand, the dancing avatar and a spinning cube are dynamic so the shadow is updated. The other spinning cube is marked as static and so it doesn't update/render. 

https://user-images.githubusercontent.com/43168857/137485227-14c4f1a7-b088-449e-9080-bd7b5f63a72a.mp4

This PR fixes also an issue we had with area light and mixed cached shadows. 

Also, given that it comes with an extra memory cost, it is an option in the HDRP asset. 

For QA: Please give a spin with other light types for the OnDemand/OnEnable with/without Always Draw Dynamic option.  Also a  bit of playing with normal shadows in directional wouldn't hurt.

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/2319_Mixed_Cached_ShadowMap_Area.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/2319_Mixed_Cached_ShadowMap_Area.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7bb0ed024eeb6ff35f2cdc32ef6ee30cb34835a88decac27a0ab9cead780279b
-size 158963
+oid sha256:d62d6b0d9eed0c4eef444aa04b982d97f8f7b190f361610749cd6ec608b8816a
+size 131574


sebastienlagarde (commented):
>This PR fixes also an issue we had with area light and mixed cached shadows.
can you make a separte backport PR for 21.2 for this fix? thanks

remi-chapelain:
Tested again the feature from scratch to made sure I understand. 
Seems to be working the same as with punctuals and area lights ✔️ 

Only thing I'd like changed is the typo on the tooltip on the HDRP Asset. Here's the proposed version: 
_"When enabled, HDRP allows mixed cached shadows for directional lights, it will incur additional memory costs."_

sgribushin:
Hi, this is a feature we definitely need in our game to hit our performance target. I have tried back-porting it to 2021.2.0f1 and it works but only in the scene edit view. 
In playmode in both the game and scene view it renders badly (the static shadows are shifted randomly everywhere, dynamic ones draw normally) depending on which view you use, it changes. It looks like the issue is a wrong transformation matrix or something similiar since they kind of keep their shape but their position is skewed. 
We are using HDRP 12.1.

Do you have an idea about where the problem could be?

FrancescoC-unity:
> Hi, this is a feature we definitely need in our game to hit our performance target. I have tried back-porting it to 2021.2.0f1 and it works but only in the scene edit view. In playmode in both the game and scene view it renders badly (the static shadows are shifted randomly everywhere, dynamic ones draw normally) depending on which view you use, it changes. It looks like the issue is a wrong transformation matrix or something similiar since they kind of keep their shape but their position is skewed. We are using HDRP 12.1.
> 
> Do you have an idea about where the problem could be?

The video captured in this PR is in playmode, however when it comes to cached shadows and directional lights is really important to know that the shadows are view dependent. For the cached component this is really important as when switching the view a lot the cached component is invalidated. 

I assume you have scene and game view together at the same time? (This will confuse one of the two as the view) 

Generally the cached component needs to be updated every time you have significant change on the main view. 

If you have only one view and no other change in view, it might be that your backport was not successful. One thing that is fundamental to get things working properly is this https://github.com/Unity-Technologies/Graphics/pull/6030/files#diff-f649de66adbf0e482453589d1bb42e446bdb7df4af577f41355a71dc72caca50R333 in case you missed this step.

sgribushin:
More info: In the build it works but only up to a certain world position, then the shadows get "noodlified" as i move and stop working as i move further, if i go back it "de-noodlifies" and works again. In the editor in playmode it is not right no matter what i do. I did check the changes you linked to (even before posting here as it was what i thought too).

Could it be that im on an older version of HDRP (2021.2 + 12.1)? This was tested on 2022 alpha right?

PR #6032, opened 2021-10-15
[DeadSoon] Transparent QoF (part 1): RefractNode and Default DistortionBlur to 0

skhiat:
Transparent Quality Of Life (part 1)

- RefractNode:
Parametrize with eta which is the ratio of both IOR (Index Of Refraction) from incoming_medium (n_i) and outgoing_medium (n_o) eta == n_i/n_i. For instance from Air to Glass that would be 1.0/1.3333.
If it's air-to-air interaction eta == 1:
image
Otherwise:
Air-glass interaction:
image

Note: the parameter incident is a vector which is from "source-to-surface" (for instance from eye-to-pixel):
image

We have 2 mode:
- Safe: Protect against NaN when a sqrt of a negative number which happen when we have glass-air interaction, at a specific angle we have no refraction but pure reflection which should be handle by Fresnel.
- CriticalAngle: Not protected the user should care of the negative case.

- Default DistortionBlur to 0:
Double check:
- "Check what's is happening with existing project, does it change their value as well"

https://jira.unity3d.com/browse/HDRP-1599

### DIFF
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/MathNodes.png.meta
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/MathNodes.png.meta
@@ -6,7 +6,7 @@ TextureImporter:
   serializedVersion: 11
   mipmaps:
     mipMapMode: 0
-    enableMipMap: 1
+    enableMipMap: 0
     sRGBTexture: 1
     linearTexture: 0
     fadeOut: 0
@@ -21,7 +21,7 @@ TextureImporter:
     heightScale: 0.25
     normalMapFilter: 0
     flipGreenChannel: 0
-  isReadable: 0
+  isReadable: 1
   streamingMipmaps: 0
   streamingMipmapsPriority: 0
   vTOnly: 0
@@ -40,7 +40,7 @@ TextureImporter:
     wrapU: 0
     wrapV: 0
     wrapW: 0
-  nPOTScale: 1
+  nPOTScale: 0
   lightmap: 0
   compressionQuality: 50
   spriteMode: 0
@@ -71,7 +71,7 @@ TextureImporter:
     maxTextureSize: 2048
     resizeAlgorithm: 0
     textureFormat: -1
-    textureCompression: 1
+    textureCompression: 0
     compressionQuality: 50
     crunchedCompression: 0
     allowsAlphaSplitting: 0
--- a/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/WindowsPlayer/Direct3D11/None/MathNodes.png
+++ b/TestProjects/ShaderGraph/Assets/ReferenceImages/Linear/WindowsPlayer/Direct3D11/None/MathNodes.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5b19559db9b64d6a2fde88cdb0ca688292dee46b3501b9f5d1c60030fe1287d5
-size 153341
+oid sha256:f403d2d27a2af926d8ba482d3f97324fc48f5b6edec7cd188fc11dd039e21014
+size 247602


kecho (approved):

com.unity.shadergraph/Documentation~/Refract-Node.md
```diff
@@ -0,0 +1,42 @@
+# Refract Node
+
+## Description
+
+Generates a refraction [Ibn Sahl-Snell-Descartes's Law](https://en.wikipedia.org/wiki/Snell%27s_law). Based on an **Incident** vector, the **Normal** of the surface and **Eta** we produce the refracted vector.
```
sarah-welton, line 5: Are all of the names really necessary, if it's more commonly known as just "Snell's law"? Even the link you're pointing to is just "Snell's Law". I can see this being unnecessarily confusing. 

> Generates a refraction using [Snell's Law](https://en.wikipedia.org/wiki/Snell%27s_law) and produces a new refracted vector.

com.unity.shadergraph/Documentation~/Refract-Node.md
```diff
@@ -0,0 +1,42 @@
...
+
+Based on Ibn Sahl's Law, for a given refractive index of the medium, we can reach an angle where all the surface behave like a perfect mirror, that's why in **Safe** Mode generate a null vector, this implementation is here to not double the reflection intensity which is handle by Fresnel.
```
sarah-welton, line 7: This needs some work. If I'm understanding correctly... 

1. For any refractive index for a given medium, there is an angle where the surface behaves like a perfect mirror. 
2. There is a "Safe Mode" for this node that will return a null vector. 

I don't know what you mean by "this implementation is here to not double the reflection intensity which is handle by Fresnel"

com.unity.shadergraph/Documentation~/Refract-Node.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Ports
+
+| Name        | Direction           | Type  | Binding | Description |
+|:------------ |:-------------|:-----|:---|:---|
+| Incident      | Input | Vector | None | Incident normalied vector to the surface (From surface to source, for instance from pixel to eye). |
```
sarah-welton, line 13: I'm not sure what this is supposed to mean. Is this the vector that represents the direction of the light as it hits the surface and reflects back, or what?

com.unity.shadergraph/Documentation~/Refract-Node.md
```diff
@@ -0,0 +1,42 @@
...
+| Normal      | Input | Vector | None | Normal normalized of the surface |
```
sarah-welton, line 14: Normal normalized what of the surface? 

I'm not sure what you mean.

com.unity.shadergraph/Documentation~/Refract-Node.md
```diff
@@ -0,0 +1,42 @@
...
+| Eta      | Input | Float    | None | The ratio of refractive index (Eta == (source refractive index)/(medium refractive index)), For instance for an interaction from air to glass **Eta** will be 1.0/1.5 and for an interaction from glass to air **Eta** will be 1.5/1.0. |
```
sarah-welton, line 15: What are we referring to by the "source" refractive index here? The air around the item before it hits the medium of the shader? 

Also, is "Eta" the best term to use here? Is this supposed to be an acronym for something? If it is an acronym, it should be capitalized as ETA - but then I'd argue that there should be a better name that would avoid confusion with "estimated time of arrival".

com.unity.shadergraph/Documentation~/Refract-Node.md
```diff
@@ -0,0 +1,42 @@
...
+| Out | Output      |  Vector | None | Refracted vector |
```
sarah-welton, line 16: > The refracted vector.

com.unity.shadergraph/Documentation~/Refract-Node.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Controls
+
+| Name        | Type           | Options  | Description |
+|:------------ |:-------------|:-----|:---|
+| Mode      | Dropdown | CriticalAngle, Safe | Select if the implementation will protect against critical angle refraction which will produce a NaN (Square root of a negative number), if the Safe mode is used when we reach the critical angle a null vector is returned. |
```
sarah-welton, line 22: I don't understand what these two modes are supposed to do. Does selecting "CriticalAngle" allow for returning the NaN result, and "Safe" prevents this?

sarah-welton (changes requested):
Please address these comments before final submission. Feel free to reach out to me on Slack @sarah.welton to discuss :)

com.unity.shadergraph/Documentation~/Refract-Node.md
skhiat: Safe mode is here to not have NaN and avoid the square root of a negative number.

"this implementation is here to not double the reflection intensity which is handle by Fresnel":
I could change the code to return the ray that implies the reflection (from total reflection), but I decided to not do that, because that behavior should be handle by the reflection from a better fresnel code.

com.unity.shadergraph/Documentation~/Refract-Node.md
skhiat: That represents what we want to be refracted, one use case would be light to pixel/surface, or eyes to surface/pixel.
But it can be used as the user wants for multi bounce inside a bottle for instance.

com.unity.shadergraph/Documentation~/Refract-Node.md
skhiat: A vector can have a norm/length larger than one, here I underline the fact that vector should be normalized.
Like the node View Vector/View Direction, one is normalized and the other one not.

com.unity.shadergraph/Documentation~/Refract-Node.md
skhiat: Eta is for the greek letter eta η which define the ratio from the source (where the light is coming from) and the medium IOR.
For example an Air-Glass Interation would have an eta of 0.666 (== 1.0/1.5)
And a Glass-Air Interaction would have an eta of 1.5 (== 1.5/1.0) with 1.0 as the IOR of Air and 1.5 the IOR of glass.

com.unity.shadergraph/Documentation~/Refract-Node.md
skhiat: Yes that's it, the Safe one have a small extra cost.
For example if you always do Air-Medium interaction you'll never have NaN, sor "cheaper" "CritialAngle" is the way to go.
But for Medium-Air it's either the user check the value to switch or always use the "Safe".

com.unity.shadergraph/Documentation~/Refract-Node.md
skhiat: It's Snell's Law only on anglo-saxon world, I learnt this as "Descartes's Law".
I'll push an simpler update.
Wikipedia:
"The law eventually named after Snell was first accurately described by the Persian scientist Ibn Sahl at the Baghdad court in 984"
It's to be fair, Sahl Law is 600 years before Snell.
Ref:
https://twitter.com/wkjarosz/status/1443562779701547012?s=20

com.unity.shadergraph/Documentation~/Refract-Node.md
sarah-welton: >The Refract node generates a refraction using an **Incident** vector, a normalized **Normal** vector of the surface, and the refractive index ratio, or **Eta** and produces a new refracted vector. The node uses the principles described in [Sahl-Snell's Law](https://en.wikipedia.org/wiki/Snell%27s_law). 

> Based on Sahl-Snell's Law, for a medium's given refractive index, there is an angle where the surface behaves like a perfect mirror. When you set the Refract node's Mode to **Safe**, the node generates a null vector when it reaches this critical angle. 

Please change the text here to the above.

com.unity.shadergraph/Documentation~/Refract-Node.md
sarah-welton: > The normalized vector of what should be refracted to the surface causing the refraction. For example, this could be from a light source to a pixel or the surface, or from the eye to the pixel or surface. 

Please change the text to the above.

com.unity.shadergraph/Documentation~/Refract-Node.md
sarah-welton: > The normalized normal of the surface that should cause the refraction. 

Please change the text to the above.

com.unity.shadergraph/Documentation~/Refract-Node.md
sarah-welton: > The refractive index ratio, which is defined by the Source Refractive Index (where the light is coming from) divided by the Medium Refractive Index (the medium causing the refraction). For example, the **Eta** for an air source and a glass medium would be 1.0/1/5, while a glass source to an air medium would be 1.5/1.0. 

Please change the text to the above.

com.unity.shadergraph/Documentation~/Refract-Node.md
sarah-welton: > Select **Safe** to prevent the Refract node from returning a NaN result when there is a risk of critical angle refraction. **Safe** will return a null vector result, instead. Select **CriticalAngle** to avoid this check for a potential NaN result. 

Please change the text to the above.

sarah-welton (changes requested):
If I've mis-stated anything in the text I've provided, please let me know - otherwise, please make the requested changes :) Thank you!

sarah-welton (approved):
Wonderful - thank you!

com.unity.shadergraph/Editor/Data/Nodes/Math/Advanced/RefractNode.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                m_RefractMode = value;
+                Dirty(ModificationScope.Graph);
+            }
+        }
+
+        public RefractNode()
+        {
+            name = "Refract";
+            synonyms = new string[] { "refract", "refract safe", "refract critical angle" };
```
bencloward, line 37: synonyms are alternate search terms, so that people can find the node using alternate names in the search - so you don't need to include the node's name among the synonyms.  I recommend using terms like warp, bend, and distort as synonyms for refract.

com.unity.shadergraph/Editor/Data/Nodes/Math/Advanced/RefractNode.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace UnityEditor.ShaderGraph
+{
+    enum RefractMode
+    {
+        CriticalAngle,
+        Safe
+    };
+
+    [Title("Math", "Advanced", "Refract")]
```
bencloward, line 14: I recommend putting this in the Math/Vector category instead of Math/Advanced.  It's very similar to the Reflection node - which is also in the Vector subcategory.

com.unity.shadergraph/Editor/Data/Nodes/Math/Advanced/RefractNode.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        static string Unity_RefractCriticalAngle(
+            [Slot(0, Binding.None, 0, 0, 0, 1)] DynamicDimensionVector Incident,
+            [Slot(1, Binding.None, 0, 1, 0, 1)] DynamicDimensionVector Normal,
+            [Slot(2, Binding.None)] Vector1 Eta,
+            [Slot(3, Binding.None)] out DynamicDimensionVector Out)
+        {
+            return
+@"
+{
```
bencloward, line 56: Hmm, I expected you would use the refract() intrinsic function. How is this different from that?

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDBlockFields.cs
```diff
@@ -32,7 +32,7 @@ public struct SurfaceDescription
             public static BlockFieldDescriptor Distortion = new BlockFieldDescriptor(SurfaceDescription.name, "Distortion", "SURFACEDESCRIPTION_DISTORTION",
                 new Vector2Control(Vector2.zero), ShaderStage.Fragment); // TODO: Lit is Vector2(2.0f, -1.0f)
             public static BlockFieldDescriptor DistortionBlur = new BlockFieldDescriptor(SurfaceDescription.name, "DistortionBlur", "Distortion Blur", "SURFACEDESCRIPTION_DISTORTIONBLUR",
-                new FloatControl(1.0f), ShaderStage.Fragment);
+                new FloatControl(0.0f), ShaderStage.Fragment);
```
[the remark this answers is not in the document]
remi-chapelain: Tested it, it works as expected, meaning we keep the value previously set before upgrading

remi-chapelain (changes requested):
To validate the utility and the behavior of the node, I tried making refraction with a transparent unlit sampling the HDSceneColor or with an opaque unlit sampling reflection probe node. 
Left is HDRP/Lit, right is the graph. This is the kind of thing that I would like to be included in the transparent sample later so that users have a base to start their own transparency pipeline. 

https://user-images.githubusercontent.com/57442369/142402177-a259283e-a299-4265-af89-554f96a464c5.mp4

The only thing I'd like changed and that has already been discussed with the stakeholder is to change "eta" input with a more easily understandable "source IOR" and "medium IOR" to avoid confusion. It can also prevent some mistakes from user of using a medium IOR of 0, resulting in infinity thus breaking the rendering.

com.unity.shadergraph/Editor/Data/Nodes/Math/Advanced/RefractNode.cs
skhiat: "intrinsic" refract will return zero at the critical angle and our case we have 2 modes criticalAngle and safe to handle that.
For the safe-mode we block the rectracted angle to the surface, and for the critical here we return let the math do the rest.

remi-chapelain (approved):
Asked for a new test to be added, 
with the latest changes, looks good to me ! ✔️

remi-chapelain:
Just noticed that the default distortion scale / Distortion blur scale set to 1 by default. So if we change distortion default value to 0 in the SG Lit, we should also change the default for HDRP/Unlit too

https://user-images.githubusercontent.com/57442369/142641082-24dc707e-dcad-4eff-b060-77cce423d7ec.mp4

bencloward (approved):
Looks good - thanks for making this!

com.unity.render-pipelines.core/ShaderLibrary/BSDF.hlsl
```diff
@@ -85,6 +85,20 @@ real3 F_Transm_Schlick(real3 f0, real u)
     return F_Transm_Schlick(f0, 1.0, u);        // sub mul mul mad mad*3
 }
 
+// Compute the cos of critical angle: cos(asin(eta)) == sqrt(1.0 - eta*eta)
+// eta == IORMedium/IORSource
+// If eta >= 1 the it's an AirMedium interation, otherwise it's MediumAir interation
+real CosCritialAngle(real eta)
```
sebastienlagarde, line 91: there is a typo here: CosCritialAngle -> CosCriticalAngle

esmelusina:
Looks like this PR was refactored to be HDRP only and merged, and this new one will move it into ShaderGraph proper: https://github.com/Unity-Technologies/Graphics/pull/6924.

As there is now a replacement PR, I'm going to go ahead and close this.

PR #6033, opened 2021-10-15
Fix APV "Size" probe debug view

FrancescoC-unity:
It got broken at some point (or well not re-implemented as the code just disappeared). 

This fixes the issue

Now the debug view shows a gradient from green (densest subdiv level) to red (coarser subdiv level). 

Unity_E1y2eqBkui

Before this it would be just plain broken.


alelievr (approved):

iM0ve (approved):
Size debug mode is working. Checked other APV debug modes that they are still working

sebastienlagarde (approved):

PR #6035, opened 2021-10-15
Convert SkyManager to use RenderGraph

JulienIgnace-Unity:
### Purpose of this PR
Moved all SkyManager code to proper render graph passes and resources where possible.
- Moved call to SkyManager.UpdateEnvironment to the RecordRenderGraph func
- Refactor Cloud system to reuse render graph passes to generate ambient probe
- Changed the RequiresPreRenderSky API on SkyRenderer class to only provide SkySettings instead of BuiltinSkyParameters because the function could access stale parameters depending on the situation (could happen with old code as well)
- Removed a hack waiting for all async readback during first frame because it broke renderer list when called from a render graph pass. The hack was only there to provide fog with proper ambient on first frame in the template. After discussion it was decided to do another PR before this one where Ambient would be provided directly on the GPU to the fog in order to avoid this hack.

https://jira.unity3d.com/browse/HDRP-330

---
### Testing status
Ran graphics tests and tested locally basic sky setups

---
### Comments to reviewers
In theory, the Cubemap target stored in a SkyRenderContext could be a temporary texture only used to generate ambient probe and sky reflection when needed. For now, we still need it to provide Enlighten and the Baking system with sky data. Once we have proper API and/or drop Enlighten we should be able to save this texture.


TomasKiniulis:
@JulienIgnace-Unity I suppose for QA we need to test various sky types with Dynamic and Static lighting(Enlighten, Progressive lightmap bakes)? Anything else to be aware of?

JulienIgnace-Unity:
@TomasKiniulis yes that's pretty much it. A quick look at multiple camera (with different sky setups) setup would be good too. It's automatically tested too but is probably the things that breaks the most :p

JMargevics (approved):
Things tested: 

- DX11, 12, Vulkan (win), PS4
- Upgrading template
- Empty project
- HDRI, PBR and Gradient skies
- Multiple cameras with different skies rendering onto RT
- Baking lightmaps, APV, reflection probes
- Enabling Enlighten
- Creating env. libraries in LookDev

Found no regressions or visual issues 👍🟢

anisunity:
@JMargevics This needs to be extensively tested on metal as it touches codes that has different behaviors on metal (and especially that our automation doesn't run)

anisunity (commented):
These kind of PRs are quite hard to review, but it looks overall fine. This needs extensive testing on metal as it was a nightmare to make it work and our automation doesn't run properly on metal at the moment (i really don't want to have to fix it again in 2 weeks).

Couple minor code coherence comments were shared directly with @JulienIgnace-Unity

JMargevics:
Did an additional pass on mac with Radeon VII:

- Specifically checked how volumetric clouds mix with PBR skies
- Tried baking probes and lightmaps
- Verified that global ambient probe works correctly
- Built a player

Found no regressions related to this PR 👍🟢

sebastienlagarde (approved):

PR #6057, opened 2021-10-19
Fix reflections with smoothness debug mode on URP

arttu-peltonen:
### Purpose of this PR
Fixes [1374181](https://fogbugz.unity3d.com/f/cases/1374181/). Currently the "Reflections with Smoothness" lighting debug mode on URP doesn't make sense. This PR fixes that. See screenshots for comparison.

To recap, the "Reflection" debug mode describes what the specular portion of lighting would look like using perfectly smooth mirror-like materials. The "Reflection with smoothness" debug mode does the same thing, but doesn't override smoothness to 1 and instead uses the smoothness value from each material.

Debug mode: None:
image

Debug mode: Reflection (before & after):
image

Debug mode: Reflection with smoothness (**before**):
image

Debug mode: Reflection with smoothness (**after**):
image

---
### Testing status
Tested debug mode manually in URP template.


Jonasmortensen (approved):
Looks good!

PR #6071, opened 2021-10-19
DOTS minimal picking support (Graphics part)

vincent-breysse:
### Purpose of this PR
This PR implements URP/HDRP minimal picking support for DOTS. This is intended to be on parity with Hybrid Renderer V2 picking. 

The basic idea here is to define a new shader for URP/HDRP and use that only when rendering the BatchRendererGroup objects for the picking (renders object instanceIDs in a tiny 4x4 texture, then read it back from CPU). 
Of course, it does not fully work with any kind of vertex deformation (including tessellation) and decals. 
To put it simply: Picking "hitbox" = Mesh transformed by LocalToWorld matrix

This PR does not have any impact on the current GameObject picking.

---
### Filtered diff for URP/HDRP reviewers [here](https://github.com/Unity-Technologies/Graphics/pull/6071/files?authenticity_token=auK5pt0IYq0bcIb30n6FmrFhRQHkJODn27yjtoFtHDsVKpvjWQ571gxZZ4TWx3maoNvPFflLPiF0oOAILuJh6w%3D%3D&file-filters%5B%5D=.hlsl&file-filters%5B%5D=.md&file-filters%5B%5D=.shader)

---
### Testing status
- HDRP validated by @TomasKiniulis 
- URP validated by @cinight 
- Yamato HDRP green except for some unrelated failing tests
- Yamato URP green except for some unrelated failing tests
---
### Note to reviewers
There are some changes included in the ShaderGraph and VFX Graph packages, but they are not mine. Those changes are the result of an auto-upgrade performed when opening the HybridURP/HDRP test projects with recent trunk.


joeldevahl (approved):

TomasKiniulis (approved):
In good state on HDRP side, main review here:
https://github.cds.internal.unity3d.com/unity/unity/pull/6293#pullrequestreview-224958

JussiKnuuttila (approved):

JussiKnuuttila (approved):

cinight (approved):
Done URP testing. All good.
Test doc: https://docs.google.com/document/d/108rKkGsKsHEupselMUUtPv5yH5GwU0J6BC89ob2ftl4/edit#

com.unity.render-pipelines.high-definition/Editor/Material/BRGPicking.hlsl
```diff
@@ -0,0 +1,131 @@
...
+    objectToWorld._m03_m13_m23 -= unity_BRGPickingCameraWorldPos.xyz;
+ #endif
+    return objectToWorld;
+}
+
+float4 ComputePositionCS(float3 positionWS)
+{
+    float4x4 viewMatrix = unity_BRGPickingViewMatrix;
+    // HDRP expects no translation in the matrix because of camera relative rendering
+    viewMatrix._m03_m13_m23_m33 = float4(0,0,0,1);
```
sebastienlagarde, line 45: you need to test the define 
#if (SHADEROPTIONS_CAMERA_RELATIVE_RENDERING != 0)
#endif

as it is possible to disable the camera relative in shader with shaderconfig.hlsl

com.unity.render-pipelines.high-definition/Editor/Material/BRGPicking.hlsl
```diff
@@ -0,0 +1,131 @@
...
+
+#define MAX_TESSELLATION_FACTORS 1.0
+
+struct PickingTessellationFactors
+{
+    float edge[3] : SV_TessFactor;
+    float inside : SV_InsideTessFactor;
+};
+
+PickingTessellationFactors HullConstant(InputPatch<PickingMeshToDS, 3> input)
```
sebastienlagarde, line 79: I don't get it, if you define your own tessellation pass, then it will not work with shader graph controlled tessellation. (haven't read the full PR yet, so maybe there is something later). My guess is that this picking pass must be define for all material. Same thing for vertex deformation. Have you tested with it?

com.unity.render-pipelines.high-definition/Editor/Material/BRGPickingFallback.shader
```diff
@@ -0,0 +1,92 @@
...
+            #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
+            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
+            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/FragInputs.hlsl"
+            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPass.cs.hlsl"
+
+            //-------------------------------------------------------------------------------------
+            // variable declaration
+            //-------------------------------------------------------------------------------------
+
+            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitProperties.hlsl"
```
sebastienlagarde, line 41: If you have a generic shader, you shouldn't include LitProperties.hlsl. This file is only mean to be used with Lit.shader. Why do you include it?

sebastienlagarde (changes requested):
Hey, so for HDRP there is something wrong, or maybe I do'nt understand the PR.

The current code don't work with vertex deformation or tessellation in SG or depth offset. 
And overall don't work with SG.

I am not sure to understand why you need to add this file BRGPicking.hlsl with vert and frag definition.

My understanding is what you really want is simply to reuse the current scene picking pass we have declare everywhere, and simply adjust a bit the code with #ifdef DOTS_INSTANCING
to call this specific function LoadObjectToWorldMatrixBRGPicking();

this function is the main change from what I get.
Let's setup a call or a chat will be easier to see how to solve that (or if I haven't understand the purpose)

sebastienlagarde:
>Tessellated object picking is only partially implemented. The picking "hitbox" is basically the same as the non-tessellated >version of the object, which means that trying to pick displaced triangles does not work.

ok I should have read the description, but same problem is true for vertex deformation. Let's talk can be easily fixed.

com.unity.render-pipelines.high-definition/Editor/Material/BRGPicking.hlsl
vincent-breysse: Solved by removing anything related to camera relative rendering. It's not needed at all anymore with an override picking shader (instead of fallback)

com.unity.render-pipelines.high-definition/Editor/Material/BRGPicking.hlsl
vincent-breysse: The picking isn't intended to work with displaced triangles. This is only needed to make sure the shader compiles. For example, the picking pass from the LitTessellation shader declares `#pragma hull Hull` and `#pragma domain Domain`, which means that the tessellation shader stages must be defined at some point. That's what `BRGPicking.hlsl` was doing by just defining passthrough tessellation stages. Although all of this doesn't apply to the new iteration since now it uses an override picking shader instead of a fallback. Now there is no need for any tessellation stage anymore because it can simply be omitted entirely.

com.unity.render-pipelines.high-definition/Editor/Material/BRGPickingFallback.shader
vincent-breysse: That is true. I removed it

TestProjects/HDRP_HybridTests/Assets/SampleScenes/BRGGameObjects/RenderBRG.cs
```diff
@@ -332,12 +340,27 @@ public JobHandle OnPerformCulling(BatchRendererGroup rendererGroup, BatchCulling
...
+        Shader shader = Shader.Find("Hidden/HDRP/BRGPicking");
+        Material material = new Material(shader);
+
+        // Prevent Material unloading when switching scene
+        material.hideFlags = HideFlags.HideAndDontSave;
+
+        return material;
+    }
+
+    Material m_pickingFallbackMaterial;
```
JussiKnuuttila, line 354: Maybe this shouldn't have "fallback" in the name anymore?

TestProjects/UniversalHybridTest/Assets/SampleScenes/BRGGameObjects/RenderBRG.cs
```diff
@@ -332,12 +341,28 @@ public JobHandle OnPerformCulling(BatchRendererGroup rendererGroup, BatchCulling
         return jobHandleOutput;
     }
 
+    public static Material LoadFallbackMaterial()
```
JussiKnuuttila, line 344: "Fallback" still here also

JussiKnuuttila (approved):

TestProjects/UniversalHybridTest/Assets/SampleScenes/BRGGameObjects/RenderBRG.cs
vincent-breysse: Fixed

TestProjects/HDRP_HybridTests/Assets/SampleScenes/BRGGameObjects/RenderBRG.cs
vincent-breysse: Fixed

sebastienlagarde (approved):

phi-lira (approved):
Approved with some comments.

com.unity.render-pipelines.universal/Shaders/BRGPicking.shader
```diff
@@ -0,0 +1,75 @@
...
+            Varyings Vert(Attributes input)
+            {
+                Varyings output = (Varyings)0;
+                UNITY_SETUP_INSTANCE_ID(input);
+                UNITY_TRANSFER_INSTANCE_ID(input, output);
+                UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(output);
+
+                float4 positionWS = mul(UNITY_MATRIX_M, input.positionOS);
+                float4 positionVS = mul(unity_BRGPickingViewMatrix, positionWS);
+                output.positionCS = mul(unity_BRGPickingProjMatrix, positionVS);
```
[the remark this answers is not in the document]
vincent-breysse: So the way the picking works is that when you click, the objects around your cursor are rendered to a tiny 4x4 texture. Each rendered objects output a different index that is related to the instanceID (the `_SelectionID` thingy). The texture is then read back from the CPU to find the instanceID of the object on which you clicked. 

All of this happens in a completly custom render loop which sets the relevant shader constants (i.e view/proj matrices + `_SelectionID`). 
For SRP and built-in shaders, the picking matrices are automatically propagated via some macro definitions. Which is why you can usually just use the `UNITY_MATRIX_V/P` and don't have to use those matrices explicitly.

Now the problem is that whereas it works for GameObjects, there are some issues with `BatchRendererGroup` rendering, which is for example used to render DOTS entities. 

We had to hack the picking render loop to connect it to SRP Batcher to render BRG draw commands, and this causes some interferences with the usual picking global shader constants, namely `unity_matrixV` and `glstate_matrix_projection`. Those constants are set both by the picking render loop and by SRP Batcher at the same time, but in an unsynchronized way. When the picking render loop sets the matrices for some particular camera, SRP Batcher may set the matrices for another random camera. When rendering BRG draw commands, this causes the shader picking passes to render with the wrong view/proj and matrices and the picking doesn't work.

To solve this, I just defined the two new matrices `unity_BRGPickingViewMatrix` and `unity_BRGPickingProjMatrix`, which are only set from the picking render loop when it stumbles across a BRG node.

com.unity.render-pipelines.universal/Shaders/BRGPicking.shader
```diff
@@ -0,0 +1,75 @@
...
+            };
+
+            Varyings Vert(Attributes input)
+            {
+                Varyings output = (Varyings)0;
+                UNITY_SETUP_INSTANCE_ID(input);
+                UNITY_TRANSFER_INSTANCE_ID(input, output);
+                UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(output);
+
+                float4 positionWS = mul(UNITY_MATRIX_M, input.positionOS);
```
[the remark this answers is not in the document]
vincent-breysse: This dance is needed with HDRP because of camera relative rendering. Camera relative rendering is not needed for the picking and causes issues because it uses the `_WorldSpaceCameraPos` constant, which is affected by the interference problem I described in my previous comment https://github.com/Unity-Technologies/Graphics/pull/6071#discussion_r745587459

com.unity.render-pipelines.high-definition/Editor/Material/BRGPicking.shader
```diff
@@ -0,0 +1,74 @@
...
+            struct PickingAttributesMesh
+            {
+                float3 positionOS   : POSITION;
+                UNITY_VERTEX_INPUT_INSTANCE_ID
+            };
+
+            struct PickingMeshToPS
+            {
+                float4 positionCS : SV_Position;
+                UNITY_VERTEX_INPUT_INSTANCE_ID
```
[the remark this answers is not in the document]
vincent-breysse: I don't think it's actually needed by any SRP for the picking. This is likely the result of a copy pasta. I'll just remove it :)

com.unity.render-pipelines.universal/Shaders/BRGPicking.shader
```diff
@@ -0,0 +1,75 @@
+Shader "Hidden/Universal Render Pipeline/BRGPicking"
```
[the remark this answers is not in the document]
vincent-breysse: Yes

com.unity.render-pipelines.universal/Shaders/BRGPicking.shader
```diff
@@ -0,0 +1,75 @@
...
+{
+    SubShader
+    {
+        // Universal Pipeline tag is required. If Universal render pipeline is not set in the graphics settings
+        // this Subshader will fail. One can add a subshader below or fallback to Standard built-in to make this
+        // material work with both Universal Render Pipeline and Builtin Unity Pipeline
+        Tags{"RenderType" = "Opaque" "RenderPipeline" = "UniversalPipeline" "UniversalMaterialType" = "Lit" "IgnoreProjector" = "True" "ShaderModel"="4.5"}
```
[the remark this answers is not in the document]
vincent-breysse: Copy pasta :)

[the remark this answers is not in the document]
vincent-breysse: I removed the `"UniversalMaterialType" = "Lit"` part

PR #6083, opened 2021-10-20
Fix shadow mask fade and optimize it at same time

sebastienlagarde:
Redo the PR here: https://github.com/Unity-Technologies/Graphics/pull/5911 after the lightloop optimization PR have landed as it was not compatible.

Fix for https://fogbugz.unity3d.com/f/cases/1369324/ 

The fade calculations were a bit off given that it was relying on culling spheres and not the max shadow distance.   What we were using before was also extremely slow for what it was. 

**What did I test**: the repro (adding a border fade too to notice an actual fade going on :P )

**Before**: 

https://user-images.githubusercontent.com/43168857/136002862-e55bc299-b1f7-41fe-9401-3710a259957e.mp4

**After:**

https://user-images.githubusercontent.com/43168857/136002897-b2b0d8ba-3453-42d8-b5ab-5a081e4a8d8b.mp4


kecho:
PR looks good, got some console performance numbers? Approving conditionally assuming we sanity checked VGPR cost :)

kecho (approved):
Awesome. Please merge once you are confident yamato is green! thank you for the tests!

PR #6086, opened 2021-10-20
[HDRP] Fix linux vulkan tests

alelievr:
### Purpose of this PR
 Fix linux vulkan tests.
This PR was the cause of the issue: https://github.com/Unity-Technologies/Graphics/pull/5918. It triggered an error on yamato because we moved the compatibility check in HDRP, causing the shaders to load even on VMs that don't have GPU


sebastienlagarde:
Merging to get green Yamato

PR #6091, opened 2021-10-21
Fix order of render targets in shaderpass forward

adrien-de-tocqueville:
### Purpose of this PR

In Lit Shader, there is the following command for forward transparent:
`ColorMask [_ColorMaskTransparentVel] 1`
which disable writing to rendertarget 1. But when virtual texturing is enabled, transparent motion vector is moved to rendertarget 2, and vt feedback is on target 1.

So it means virtual texturing on a transparent object with motion vectors disabled would not work.

I believe this is a very specific case that probably no one will encounter but we may still want ot fix it

---
### Testing status

Launched yamato, but I don't know if it contains any test for virtual texturing.


sebastienlagarde (approved):

sebastienlagarde:
GC alloc issue:
System.Exception : Memory allocation test failed, 2 allocations detected. Steps to find where your allocation is:
                        - Open the profiler window (ctrl-7) and enable deep profiling.
                        - Run your the test that fails and wait (it can take much longer because deep profiling is enabled).
                        -

PR #6106, opened 2021-10-21
[HDRP] Lit coat ssr-rtr: Fix various issues with using SSR lighting with IBL fallback (case 1380351)

slunity:
---
### Purpose of this PR
See https://github.com/Unity-Technologies/Graphics/pull/5565
and https://github.com/Unity-Technologies/Graphics/pull/4968/

1)
SSR lighting was applied using the coatMask in such a way that the bottom layer incorrectly gained coat-traced light as the coatMask went to 0, even prior #5565.
#5565 added the usage of the coat lighting (irrespective of coatMask considerations, ie even when coatMask = 1) when the bottom layer has a roughness close enough to 0, ie to that of the coat lobe for which we traced the indirect light.
However, even when the bottom layer roughess was very different (blendFactor = 1), coatMask could escape that logic as it went to 0.
(In that case, the comment in the code "// On top of this above we also add an additional hack to limit the smoothness used on based layer depending on coat layer for SSR application (See below)" was wrong)

2)
The reflectionHierarchyWeight (again even before #5565) was incorrectly including an F term, while the consumed / used-up light hierarchy weight is orthogonal
to the specific material shading details (reflectance factors, ie pre-integrated FGD or coatF here).
So regardless of if we used the ssr coat lighting for the bottom layer, as coatF grew, we would artificially remove total bottom layer light at grazing angles.
Ergo, this was incorrectly darkening the grazing angle fresnel reflection because of darkening of the bottom layer.

Moreover, with the separate coat hierarchy weight, the envlighting can no longer use the same probe volume intersection weight, and a separate update and capping must be done for the coat reflection hierarchy.

Lit - wrong bottom hierarchyWeight set by SSR evaluation
Lit - wrong bottom hierarchyWeight set by SSR evaluation - corrected

https://fogbugz.unity3d.com/f/cases/1380351/
Lit - wrong bottom hierarchyWeight set by SSR evaluation

(Notice the grazing angle darkening in the A/B compare)

3)
In #5565, the code states we lerp-in coat light contribution for the bottom layer when its smoothness is above 0.9 and lerp-out from 0.9 to 0.8, but the code was actually fading from 0.9 to 0.7, not 0.8. I assumed the code behavior was desired (since eyeballed) and used the same ranges but made the code cleaner.

4)
More importantly (vs 3), the behavior since #5565 could actually erroneously eliminate the coat lobe and use only the bottom lobe with the coat-traced light: this happened when the bottom roughness was close
to the coat roughness, as in that case we had a "blendfactor" that would lerp from coat F to bottom FGD, instead of actually combining both lobes with their respective wanted contributions.

Additional video showing coat mask behavior (eg for point 1 above):

Before:
https://drive.google.com/file/d/1z0i81gTs9PqZnZ2VnRV6XGQsF26GvhwJ/view?usp=sharing
https://drive.google.com/file/d/1f5e_Kus8WaXdE2UwQ-rb9DnBeQTO4YyI/view?usp=sharing

After:
https://drive.google.com/file/d/1ifHDZUlfdByzSL4ocN0jMYjiB8b_EuD4/view?usp=sharing

---
### Testing status

Manual test scene, see note below for graphics tests.

---
### Comments to reviewers

Note that other incoming PRs might also change graphic test results, might be worth it to bundle them together instead of changing screenshots for each.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed Xbox Series X compilation issue with DoF shader
 - Fixed references to reflection probes that wouldn't be cleared when unloading a scene. (case 1357459)
 - Fixed issue with Stacklit raytrace reflection
+- Fixed various issues with using SSR lighting with IBL fallback for Lit shader with clear coat(case 1380351)
 
 ### Changed
 - Optimizations for the physically based depth of field.


anisunity (approved):
lgtm

anisunity:
This PR fixes this issue https://fogbugz.unity3d.com/f/cases/1380351/

sebastienlagarde (approved):

PR #6107, opened 2021-10-21
[HDRP] StackLit coat ssr-rtr light hierarchy and IBL fallback fixes.

slunity:
---
### Purpose of this PR
See https://github.com/Unity-Technologies/Graphics/pull/6106

This implements more complex base/coat light reflection hierarchy as was done in https://github.com/Unity-Technologies/Graphics/pull/4968/ but with the addition of the coat-traced light reuse for bottom for similarly-rough lobes on the bottom.

This fixes several issues in the implementation in StackLit since #5565:

- Similar issues fixed for Lit (see other PR for Lit)
- StackLit can have a rough coat, but an arbitrary 0.9 smoothness value was used to decide if the coat-traced light would be re-used for the bottom lobes.
- The wrong bottom lobe roughnesses were used (values that don't include the effect of BRDF vertical layering computations)
- StackLit can have different lobe directions and the light hierarchy now account for this.

---
### Testing status

Tests in local test scene, see note below:

---
### Comments to reviewers

Note that other incoming PRs might also change graphic test results, might be worth it to bundle them together instead of changing screenshots for each.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed references to reflection probes that wouldn't be cleared when unloading a scene. (case 1357459)
 - Fixed issue with Stacklit raytrace reflection
 - Fixed various issues with using SSR lighting with IBL fallback for Lit shader with clear coat(case 1380351)
+- Fixed stackLit coat screen space reflection and raytrace reflection light hierarchy and IBL fallback
 
 ### Changed
 - Optimizations for the physically based depth of field.


com.unity.render-pipelines.high-definition/Runtime/Material/StackLit/StackLit.hlsl
```diff
@@ -1200,6 +1196,25 @@ void ClampRoughness(inout PreLightData preLightData, inout BSDFData bsdfData, fl
...
 }
 
+void GetVLayeredBottomLobesPerceptualRoughnesses(BSDFData bsdfData, PreLightData preLightData, out float basePerceptualRoughness[BASE_NB_LOBES])
+{
+    bool perLightOption = GetRecomputeStackPerLightOption();
+    bool haveAnisotropy = HasFlag(bsdfData.materialFeatures, MATERIALFEATUREFLAGS_STACK_LIT_ANISOTROPY);
+
+    if (perLightOption && haveAnisotropy)
+    {
+        basePerceptualRoughness[0] = preLightData.unmodifiedIblPerceptualRoughness[0];
```
anisunity, line 1206: Maybe use indices to reference coats when in relative space (when cannot be using BASE_LOBEA_IDX and BASE_LOBEB_IDX)? a bit confusing to follow

com.unity.render-pipelines.high-definition/Runtime/Material/StackLit/StackLit.hlsl
```diff
@@ -4215,43 +4269,66 @@ IndirectLighting EvaluateBSDF_ScreenSpaceReflection(PositionInputs posInput,
...
+
+            // We use the coat-traced light according to how similar the base lobe roughness is to the coat roughness
+            // (we can assume the coat is always smoother):
+            //
+            // - The roughness is equal to bsdfData.coatPerceptualRoughness
+            //   We use the fact the clear coat and that base layer lobe have the same roughness and use the SSR as the indirect specular signal.
+            // - The roughness is superior to bsdfData.coatPerceptualRoughness + 0.2.
+            //   We cannot use the SSR for that base layer lobe.
+            // - The roughness is within <= 0.2 away of bsdfData.coatPerceptualRoughness, we lerp between the two behaviors.
+            float coatSSRLightOnBottomLayerBlendingFactor = lerp(1.0, 0.0, saturate( (basePerceptualRoughness[i] - bsdfData.coatPerceptualRoughness) / 0.2 ) );
```
anisunity, line 4309: need to run formatting on the branch i guess

anisunity (approved):
Didn't read in depth, but looks good to me.

com.unity.render-pipelines.high-definition/Runtime/Material/StackLit/StackLit.hlsl
slunity: Yeah I've considered it since the beginning when I made these indices (between having yet another set for BASE_NB_LOBES sized arrays or just or 0-1), right now in the code everywhere where we process only BASE_NB_LOBES it is 0,1 so this addition is in line with that, but I can change it everywhere np.

sebastienlagarde (approved):

PR #6115, opened 2021-10-22
Fix Luminance/EV conversion when changing LightMeterCalibrationConstant

adrien-de-tocqueville:
### Purpose of this PR

Fixed light unit conversion when LightMeterCalibrationConstant is changed (for example by exposure mid gray component)
The variables were static float so they would get recomupted. I changed them to function.

Now there is two exp / two log when calling the conversion function, but doing only one exp/log causes precision issue when converting really high intensities (close to float max)

---
### Testing status

launching yamato, not much to test.
Verified by printing the constant in the console after changing mid gray value in exposure override


JMargevics (approved):
Looks good to me, not much else to test. 👍✔
**Please update the changelog** 👀

sebastienlagarde (approved):

adrien-de-tocqueville:
hum apparently the PR was backported to 10.x.x and 2021.2/staging but not to 2021.1/staging
I don't know if it has implications but just saying it

sebastienlagarde:
PR is not backport yet :).   but there is no more 21.1, the support is stopped

PR #6117, opened 2021-10-22
Made a GPU buffer with ambient probe SH available to shaders doing lighting.

JulienIgnace-Unity:
### Purpose of this PR
The purpose of this PR is to make available the result of the ambient probe convolution done on the GPU to other passes without having to go through CPU readback.
The end goal is to avoid frame latency for effects that can benefit from this.
To achieve that:
- Ambient probe convolution compute shaders can now output up to two more buffers. Originally it would output SH coeffs in a readable way for CPU readback. Now it also outputs packed coefficients for diffuse convolution (directly useable in a shader) and another one for volumetric lighting specific convolution.
- Volumetric Lighting and APV now access their respective ambient probe representation directly from these new buffers instead of constant buffer parameters gotten from CPU readback.
- Removed a hack where we'd wait for all async readbacks on first frame to avoid missing ambient from volumetric fog on first frame.

---
### Testing status
- Ran graphics test
- Tested Ambient probe with APV enabled with a mesh outside of the baked APV
- Tested volumetric fog with gradient sky (red on top and green at bottom) to observe the right color still being applied (be careful to play with the anisotropy fog paremeter to still have directionality in the result).

---
### Comments to reviewers
- Implementation of the volumetric specific convolution is directly copied from the former function SetPreconvolvedAmbientLightProbe. It could be optimized further but given that we don't often execute this code I considered readability to be more important. Also left a comment about some discrepancy I noticed when trying to do so.
- There are some performance consideration to be had here. Now volumetric and APV will sample an additional compute buffer instead of more data in the constant buffer. I don't expect this to be a huge deal but I did not have proper hw to test this. Ideally would need to make some comparison on PS4.


com.unity.render-pipelines.core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -121,6 +121,14 @@ float3 SampleSH9(float4 SHCoefficients[7], float3 N)
     return res;
 }
 
+void GetCornetteShanksPhaseFunction(out float zh[3], float anisotropy)
```
FrancescoC-unity, line 124: Worth looking a bit at the produced code, but often I saw arrays like this compiling into weird stuff. 
Safest bet is to use float3 here though it'd mean slightly changing the code. 

Alternatively, it's fine to keep as is if we check the disasm

FrancescoC-unity (approved):
To discuss the array usage, but minor point anyhow

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
```diff
@@ -98,6 +99,20 @@ groupshared int gs_localLightList[GROUP_SIZE_1D*GROUP_SIZE_1D][MAX_SUPPORTED_LIG
 
 #endif
 
+real3 EvaluateVolumetricAmbientProbe(real3 normalWS)
```
sebastienlagarde, line 102: it miss a comment to say that _VolumetricAmbientProbeBuffer include the cornel shank case. Just so people don't ask themselve why there is a differnt funciton EvaluateAmbientProbe

com.unity.render-pipelines.high-definition/Runtime/Sky/SkyManager.cs
```diff
@@ -881,36 +957,27 @@ internal void RequestStaticEnvironmentUpdate()
...
-                                            CommandBufferPool.Release(cmd);
-                                            renderContext.Submit();
-                                            cmd = CommandBufferPool.Get();
-                                            m_RequireWaitForAsyncReadBackRequest = false;
-                                        }
+                                        cmd.SetComputeBufferParam(m_ComputeAmbientProbeCS, kernel, m_VolumetricAmbientProbeOutputBufferParam, renderingContext.volumetricAmbientProbeBuffer);
+                                        cmd.SetComputeVectorParam(m_ComputeAmbientProbeCS, m_FogParameters, new Vector4(fog.globalLightProbeDimmer.value, fog.anisotropy.value, 0.0f, 0.0f));
                                     }
+                                    cmd.DispatchCompute(m_ComputeAmbientProbeCS, kernel, 1, 1, 1);
+                                    cmd.RequestAsyncReadback(renderingContext.ambientProbeResult, renderingContext.OnComputeAmbientProbeDone);
```
sebastienlagarde, line 979: Do we still have a usage for it?

sebastienlagarde (commented):
don't forget changelog

com.unity.render-pipelines.high-definition/Runtime/Sky/AmbientProbeConvolution.compute
```diff
@@ -1,17 +1,28 @@
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
+#include "Packages/com.unity.render-pipelines.core/ShaderLibrary/EntityLighting.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Sampling/Hammersley.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Sampling/Sampling.hlsl"
 
 #pragma only_renderers d3d11 playstation xboxone xboxseries vulkan metal switch
 
 #pragma kernel AmbientProbeConvolution KERNEL_NAME=AmbientProbeConvolution
+#pragma kernel AmbientProbeConvolutionDiffuse KERNEL_NAME=AmbientProbeConvolutionDiffuse OUTPUT_DIFFUSE
```
sebastienlagarde, line 10: you can do multi compile instead

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/Sky/AmbientProbeConvolution.compute
```diff
@@ -159,5 +232,48 @@ void KERNEL_NAME(uint dispatchThreadId : SV_DispatchThreadID)
         {
             _AmbientProbeOutputBuffer[i] = outputSHCoeffs[i] * ConvolveCosineLobeBandFactor[i % 9] * weight;
         }
+
+        float result[27];
+
+#if OUTPUT_DIFFUSE
+        for (i = 0; i < SH_COEFF_COUNT; ++i)
+        {
+            result[i] = _AmbientProbeOutputBuffer[i];
```
sebastienlagarde, line 241: guess it miss a multiply by factor , lets discuss

com.unity.render-pipelines.high-definition/Runtime/Sky/AmbientProbeConvolution.compute
JulienIgnace-Unity: Not all combinations are valid here. Next PR will introduce the few variants that are actually used with more explicit names

JMargevics (approved):
image
Tested in the following scene with quality set to high, ambient mode set to dynamic and sky update mode set to realtime. Did two GPU frame traces to compare: 

Master 1186.32 μs for UpdateSkyEnvironment, 26,924.96 μs total
master

This branch 1158.82 μs for UpdateSkyEnvironment, 26,511.77 μs total
image

I see no visual or performance regressions with this branch ✅👍

PR #6129, opened 2021-10-25
RP Workflows - Adding the capability to revert a single override of a Volume Component

alex-vazquez:
### Purpose of this PR

JIRA: https://jira.unity3d.com/browse/XPIPELINE-343

Adding a context menu "Revert Property Override" for Property Fields of Volume Components

image

---
### Testing status
Check that the restore of the values works fine.
Check that the override is NOT restored, this should only be applied when Resetting the full Volume Component.

---
### Comments to reviewers

I've used the system from `IApplyRevertPropertyContextMenuItemProvider`, unfortunatelly we could not replace the name of the context menu as this is generated from the "Revert + GetSourceTerm() + override". As this system is already working, and well tested, I think counterproductive to implement a new system just to remove the override, in this scenario makes sense and we won't need to touch trunk code for this PR.


arttu-peltonen (approved):
In that case, it should be fine. Tagging @RSlysz to be aware.

JMargevics:
Hey, found a some generic volume component overrides and some specific ones that don't have a revert function.

image
All color pickers. **I think it is very important to bring the revert feature to color pickers as well!** 

image
All texture pickers. 

image
Cloud Layer. Anything in Layer A section. 

image
Gradient, HDRI, PBR skies. All intensity modes.

And the final issue is that any Revert action **is not undoable.**: if you revert a value you stick to it and there's no way to return to previous one.

alex-vazquez:
@JMargevics I think all parameters can be reset now :)

arttu-peltonen (approved):

JMargevics:
> @JMargevics I think all parameters can be reset now :)

All those fields work perfectly now ✔👍

Reposting my other issue for visibility:

https://user-images.githubusercontent.com/50581102/142857424-3520ef1e-298a-4044-accc-c968633f0f77.mp4

Quality levels and underlying parameters should be connected when we revert them.

JMargevics (approved):
All perfect now 👍✔

PR #6132, opened 2021-10-25
Add scale parameter in eye shader

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1371406/

Added a scale parameter in the HDRP Eye shader graph to counter eye shell import scale. It works by multiplying the object space position by a scale factor
The bug report suggested exposing multiple parameters to manually balance the scale factor.
This can be done more easily by adding a scale parameter, but maybe exposing the other parameters has an interest too ?

Currently the new parameter is visible, we can make it non exposed and leave it for users that want to edit the graph
image

Screnshot with new shader: left and right are now identical
image

---
### Testing status
See screenshot above

Launched yamato

---
### Comments to reviewers

Please give feedback about whether this solution is better than exposing iris size / plane offset
Also, should we make the new parameter exposed by default ?


anisunity:
A geometry section would be nice! @adrien-de-tocqueville

anisunity (changes requested):
A geometry section in the shader would be nice, also didn't open it, so I don't know if the result is 100% the same, would be worth adding a test that checks that.

remi-chapelain (changes requested):
Change is great, it now allows users to have an import scale different than 1 ✔️ 
As @anisunity said, it would be great to have a geometry section since everything is already in a section here. 

I checked and the result is basically identical depending on the scale. 

https://user-images.githubusercontent.com/57442369/139910630-80fc9f80-f2b5-4cef-9780-393d37382c6d.mp4

Now for the changes, that would be great to duplicate those changes in the HDRP Material sample. Basically the graph is duplicated [here](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.high-definition/Samples~/MaterialSamples/Shadergraphs/SG_Eye.shadergraph)

Also, there's already two tests that uses this graph ([one](https://github.com/Unity-Technologies/Graphics/blob/master/TestProjects/HDRP_Tests/Assets/Samples/High%20Definition%20RP/MaterialSamples/Shadergraphs/SG_Eye.shadergraph) in HDRP_Tests and [one ](https://github.com/Unity-Technologies/Graphics/blob/master/TestProjects/HDRP_DXR_Tests/Assets/Samples/High%20Definition%20RP/MaterialSamples/Shadergraphs/SG_Eye.shadergraph)in HDRP_DXR_Tests project. 
Would be cool if those were updated as well

adrien-de-tocqueville:
Moved it to a new geomtry section
image

Updated the 3 duplicates mentionned by rémi

For the test, not sure how to test they are similar, so I updated test 1501 to have one row use a 0.5 scaled mesh with a scaled shader. So the resulting image should not have changed

remi-chapelain (approved):
Thanks for the changes ! 
Looks good to me ✔️

sebastienlagarde (approved):

sebastienlagarde:
@AdrienDeTocqueville Please update documentation

sebastienlagarde (approved):

PR #6137, opened 2021-10-25
Adding FadeDitherNode

johnhableunity:
### Purpose of this PR
PR is for adding a dither node that fades out in a way that works well with TAA. As an example, here is a blue sphere with a noise texture fading out.
image

### Testing status
Testing was performed manually. Steps are:
1. Create a Fade Dither node.
2. Put a texture that has full range (such as a noise texture) into the Texture slot, as well as a UV.
3. Connect the output to the Alpha channel of an alpha blended material graph.
4. Tweak the values. The FadeValue controls the amount of fade (0=fully transparent to 1=fully opaque). FadeSpeed affects the hardness of the edges in the fade, PixelRatio scales the texel size.

---


bencloward (approved):
Thumbs up from me.  Looks good!

PR #6140, opened 2021-10-26
Fix issue with reflection probe normalization via APV

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1373264/ 

Tricky issue to find. The issue was actually a collection of few smaller issues - weirder of which is the need of making prefab accept the changes done programmatically. 

**What did I test** : the repro, tried to open and close the scene and verify that it's all still fine w.r.t normalization pre and post baking. 

--- Note -- 
Normalization data will be gone even if the position of the probe doesn't change but other properties do, this is somewhat undesirable, but it is not a bug, going to add as a new feature we need to do.


sebastienlagarde (approved):

iM0ve (approved):
**What I tested:**
* Checked the repro project to see the issue is indeed gone
* Checked Amalienborg project where realtime probes are used without APV. To see if maybe some unwanted changes were introduced outside APV system
* Checked ClassRoom project rebaking APV and probes to see if general APV issues were introduced

No issues related to the PR were found.

PR #6143, opened 2021-10-26
fix indentation for mask property

martint-unity:
---
### Purpose of this PR
Indented the mask property in LitGUI

image


ellioman (approved):

phi-lira (approved):
LGTM.

PR #6144, opened 2021-10-26
[HDRP][DXR] Path traced objects that are excluded from the visibility mask should not cast shadows

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1375638/

The shadow culling was not taking into account the path tracing visibility mask.

---
### Testing status
- Tested with the repro project, problem does not reproduce anymore
- Tested with the Cast Shadows "on, off, shadows only" lighting options and work as expected (when pt is enabled). 
- DXR tests are green


anisunity (commented):
Thanks for the change, I'll alter the change to make the behavior more explicit.

anisunity:
Also would be great to have a test that catches this specific behavior as the previously catching tests did not cover it when we changed the way we do culling.

sebastienlagarde:
@pmavridis  thanks for the fix, can you updated one test to catch this behavior? thanks

pmavridis:
I have added a test that checks for the cast shadow options (on, off, shadow only) and also includes an "invisible" object to catch this behavior.
5013_PathTracing_ShadowFlags

remi-chapelain:
by the way, fix doesn't need a backport because:
- Layer masks in PT doesn't suffer from this problem (object in hidden layers don't cast shadows), I just checked ✔️ 
- Manual RTAS is not a thing there.

sebastienlagarde (approved):

PR #6145, opened 2021-10-26
Fix NaN on Intel GPU when using PBR Sky and rendering sun disk 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1369376/

This was """fun""" . 

TL;DR: trig operation precision is bad on Intel GPU. 

Longer explanation: 

So in the code we do 

```
float radInner = 0.5 * light.angularDiameter;
float cosInner = cos(radInner);
```

With default parameters (and generally with angular diameter small-ish) radInner is very close to 0 but not really 0 (about 0.008). On NVIDIA cos(0.008) leads correctly to  a non-one value, while on Intel integrated this leads to an evaluation equivalent to cos(0) --> 1. 

Later on in the code we do 

```
float solidAngle = TWO_PI * (1 - cosInner);
...
float  scale = rcp(solidAngle);
```

So solid angle ends up being 0 and scale leads to a divide by zero with the rcp operator dividing by zero --> INF value slips through eventually.  

Now the divide by zero would not happen if the values of cos were calculated precisely hence why it is not happening on NVIDIA.
Using a clamp on the values to make sure we always feed a value that will not lead to bad precision is not ideal as it will inflate the size of the sun disk which is not what we want. 

Thankfully the values that make thing explode can be precomputed on CPU, so that's what this PR is doing, so that now the precision is comparable with all vendors. 

**What did I test**: the repro case and messing with sky parameters. 

_Note_ Adding @sebastienlagarde just for extending the tribal knowledge of "trig on intel == bad"  and @JulienIgnace-Unity  for awareness on the sky change. But I don't think we are blocked waiting for their green tick :)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Sky/PhysicallyBasedSky/PhysicallyBasedSky.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/Sky/PhysicallyBasedSky/PhysicallyBasedSky.shader
@@ -4,7 +4,6 @@ Shader "Hidden/HDRP/Sky/PbrSky"
 
     #pragma vertex Vert
 
-    #pragma enable_d3d11_debug_symbols
     #pragma editor_sync_compilation
     #pragma target 4.5
     #pragma only_renderers d3d11 playstation xboxone xboxseries vulkan metal switch


kecho:
Started yamato just for sanity check

sebastienlagarde (approved):

PR #6062, opened 2021-10-19
Urp/depth priming tests

NicoLeyman:
Re-try of this PR: https://github.com/Unity-Technologies/Graphics/pull/4769
Original PR added a small fix and test coverage for Depth Priming.

The PR had an issue when the test was moved to a different project but that has been fixed.
Some redundant ref images that stuck around after a master merge have been removed.
Master has been merged.

PR #6027, opened 2021-10-15
Re-enabled URP debug view shader code following a HLSLcc fix

arttu-peltonen:
### Purpose of this PR
This bug has been fixed in 22.1.0a13: https://fogbugz.unity3d.com/f/cases/1334240/, so I can now re-enable code that was disabled on VK/GL due to the bug.

---
### Testing status
Tested locally using the repro project on the [ticket](https://fogbugz.unity3d.com/f/cases/1334240/,). With engine build before the fix, the shader error reproduces, and after the fix it no longer occurs.

Note to QA: If you want to test yourself, use latest trunk or at least 2022.1.0a13.

**Note: Needs backport to 2021.2, but backporter must make sure [this ono PR](https://ono.unity3d.com/unity/unity/pull-request/133543/_/2021.2/graphics/shadersystem/minor_fixes_0004#comment-1035075) has landed in 21.2/staging before backporting this PR.**

PR #6149, opened 2021-10-26
Universal/staging

phi-lira:
- Fix URP blending modes and separate preserve specular lighting feature. (#5762)
- fix for freeze in renderpipeline converter (#6015) 
- Re-enabled URP debug view shader code following a HLSLcc fix (#6027) 
- Universal/fix sceneview filter (#6053) 
- GraphicsSettings.useScriptableRenderPipelineBatching is now set only at URP creation and not every frame (case 1368853 fix) (#6094)
- Add adaptive performance decals scaler options. (#6013) 
- Urp/depth priming tests (#6062) 
- [URP] Changing default CopyDepthMode to AfterTransparents (#6058) 
- Fix reflections with smoothness debug mode on URP (#6057)


ernestasKupciunas (approved):
Overall looks good!
Found 1 bug : https://fogbugz.unity3d.com/f/cases/1376269/ (not related to staging branch).
And 3 PRs needed some improvements like better description, root cause analysis(contacted the developers and asked to make those changes).

PR #6151, opened 2021-10-26
Clean up Public API Documentation

johnpars:
### Purpose of this PR
This PR cleans up the API documentation so that we are again passing the documentation tool validation.

Before:
image

After:
image

---
### Testing status
Running Yamato just in case the access level change I made on one function doesn't cause any issues to the tests. 

---
### Comments to reviewers
Adding Anis just for awareness as I update documentation related to acceleration structure / fog.


anisunity:
fair enough

PR #6159, opened 2021-10-28
[URP] Editor Resource Overrides

vbhayden:
### Purpose of this PR
This change allows users to create and use a custom URP Resource Asset instead of the default one.  

Presumably, the creation of this asset was removed as to not confuse users and remove the magical quality about LWRP / URP simply finding an instance to use at runtime, but this resulted in bricking certain kinds of grass bending and other terrain effects due to the hardcoded GUID being in place.  Maybe there was a great reason for this functionality being removed, but I genuinely think it was the magical assignment aspect.  

By allowing these to be an explicitly-assigned property of the URP assets themselves, this adds an additional level of customization to creators using URP.

---
### Testing status
I checked the MenuItem usage and noticed several incorrect paths for the default URP Editor Resources' `[Reload]` attributes.  These were corrected to use the proper paths, even in cases where the names are misleading (Unlit particles being referred to as Lit).

It is difficult to test these changes given how the editor will reload things from cache, though.

---
### Comments to reviewers
First PR, not super sure of all that is required?


vbhayden:
For tagging reviewers, is there an updated list of relevant reviewers?  Teams aren't taggable on GitHub and the Yamato CI link isn't publicly accessible etc.

phi-lira:
@vbhayden we don't want to expose the ability to override resources to URP quite yet. Can you comment a little bit more on what is the issue you are having that you need to override them?

On the paths, that's a good catch. If you do a separate PR with the path fixes we will take it.

vbhayden:
Sure, so something like using an alternate Grass or Detail shader for terrains is probably the biggest one.  I wrote a grass bending system awhile back that involved supplying alternate Grass / Detail textures to the (at the time) LWRP resources and then a compute shader to relay bending influences.  When I was switched things over to URP, I was surprised to see that the resources object was no longer allowing project-level overrides and instead pointed to a hardcoded GUID.
```
public static readonly string editorResourcesGUID = "a3d8d823eedde654bb4c11a1cfaf1abb";
```

For the resources object itself, a few of these shaders are very niche and understandably protected (Autodesk and SpeedTree), but it seems very odd that we are allowed to configure the Terrain's base material but prevented from changing its Grass and Detail shaders.

People have clearly worked out grass and detail bending through other means, with these usually involving some singleton somewhere and specific URP rendering features, so this isn't a dead stop -- just a bit annoying and unexpected for a system with "Scriptable" in its name, yeah?

Alternatively if the Terrain object could expose optional fields for these Grass + Detail shaders (some Advanced dropdown), then that would also allow a bit more control over things.

phi-lira:
@vbhayden I think the resource override is a proper feature to add, but we need to consider all implications of it. This will probably also involve some other teams (f.ex, terrain). I would suggest to add this as an idea in our product board. https://portal.productboard.com/unity/1-unity-platform-rendering-visual-effects/c/452-didn-t-find-what-you-were-looking-for

I will close this PR then.

PR #6174, opened 2021-10-29
Improvements to APV Virtual Offset

tlaedre:
### Purpose of this PR

The primary purpose of this PR is to improve APV probe position Virtual Offset performance. The current implementation runs entirely on the main thread and doesn't scale well to complex scenes, with the VO pass routinely taking longer to complete than actual (GPU) lightmapper raycasting. This PR implements a bursted, concurrent alternative, while also tweaking the heuristics slightly to handle certain cases better (in the scenes tested by the author).

Currently both the old and the new code can be ran in parallel (toggled by the Use Multi-Threaded checkbox). This makes it easy to compare them for validation and tweaks. The idea for a final PR if approved would be to remove the old implementation and put the MT one into its place.

So, performance first: in our production project the new code is ~200x faster than the old (on my local machine), reducing time spent from ~40s to ~200ms. In simpler test scenes the savings are a little more modest, usually coming in around ~75x faster for me.

The old implementation has a couple of cases where it doesn't yield optimal results. One of these is pushing probes that are already out in the open closer to, or in some cases even inside, geometry - the latter in particular reduces lighting quality since those end up being invalid probes that then have to be filled in by dilation.

Various incorrect push results:
image
image
image

Another case is that since there's no virtual offsetting of the origin, and because it purely chooses the best direction based on distance, it picks the first as opposed to the best direction when there's multiple rays with the same distance.

Lastly, because there's no way to get multiple ray-hits per collider, there's no reliable way to understand what a hit of an incoming ray actually represents, and that easily leads to false positives (think small objects or complex, concave shapes).

The primary change made to avoid many of these cases is removing the use of inbound rays. The other is a tweak of the hit surface and direction selection heuristics where surfaces with colinear normals are preferred for ray hits with distances within a very small threshold value.

image
image

### Testing status

Tested in our current production for more than a month.
Tested in various artificial scenes.

### Comments to reviewers

PR is currently not against master as it builds on a a different PR. This would be changed as the other PR lands.
This PR does not add a package dependency on Burst, it just uses it if available. (Which looks to be the case in HDRP/URP now)

### DIFF
--- a/.yamato/_copycat.yml
+++ b/.yamato/_copycat.yml
@@ -9,7 +9,7 @@ nightly_katana_abv_validate:
   variables:
     MANIFEST: .copycat/graphics.json
     RELEASE_BRANCH: trunk
-    SRP_VERSION: "14.0.0"
+    SRP_VERSION: "14.0.1"
   skip_checkout: true
   commands:
     - eval "$COPYCAT_1"
@@ -31,7 +31,7 @@ vendor:
   variables:
     MANIFEST: .copycat/graphics.json
     RELEASE_BRANCH: trunk
-    SRP_VERSION: "14.0.0"
+    SRP_VERSION: "14.0.1"
   skip_checkout: true
   commands:
     - eval "$COPYCAT_1"
--- a/.yamato/_latest_editor_versions_trunk.metafile
+++ b/.yamato/_latest_editor_versions_trunk.metafile
@@ -1,5 +1,5 @@
 trunk:
-  updated_at_UTC: '2022-01-19 23:55:42.859130'
+  updated_at_UTC: '2022-01-25 00:08:20.195908'
   changeset:
-    id: 3f4731843e630a0b1f56ade2a38fb20afb6812c8
-    date: '2022-01-19T23:16:47'
+    id: fc824837782cacc5ffcc55193f29b414734b1cf7
+    date: '2022-01-24T12:07:13'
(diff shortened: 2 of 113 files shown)


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeBakingProcessSettings.cs
```diff
@@ -19,8 +19,12 @@ internal struct ProbeDilationSettings
     internal struct VirtualOffsetSettings
     {
         public bool useVirtualOffset;
-        public float outOfGeoOffset;
-        public float searchMultiplier;
+        [Range(0f, 1f)] public float outOfGeoOffset;
+        [Range(0f, 2f)]public float searchMultiplier;
```
FrancescoC-unity, line 23: Missing space here :) 


Also generally I think these settings might be a tad too advanced for users. I would add them under an advanced foldout in the editor.

tlaedre: Which ones do you want under an advanced foldout? All of them?

FrancescoC-unity: Yes probably

iM0ve:
Im getting the same issue as mentioned here, preventing me from seeing the APV probes in debug. Making it hard to test
https://github.com/Unity-Technologies/Graphics/pull/6169#pullrequestreview-806204059

iM0ve (commented):
**Review status:** Done
The PR brings improvements to Virtual Offset, feature currently can be turned off(checkbox Use Multi-Threaded). While testing I did not find any regressions.

**What's tested:**
* Compared results of APV when feature is enabled vs disable
* Rebaked lighting of Classrom scene

**FIXED 1) Errors.** So far whenever I enable multi-threaded checkbox I get errors and doesn't bake. Not sure if im missing something or there is an issue.

image


**Feature showcase**
The virtual offset gained a visible improvement im my scene.

**Feature turned off**
Unity_Dyw6O1995e

**Feature turned on**
Unity_7RDQkiF9m8

iM0ve (changes requested):
Confirmed its unintended, waiting for changes

iM0ve (approved):
Issues fixed, updated my review.

PR #6175, opened 2021-10-29
Adding workaround for decal and ssao when in intermediate texture mode

lukaschod:
---
### Purpose of this PR
Adding decal and ssao after opaque to force intermediate texture mode rendering in URP.
The main reason why it is workaround and not complete fix:
- We need fast fix as https://github.com/Unity-Technologies/Graphics/pull/5297 uncovered that decal and ssao does not work well in direct back buffer mode. Also this will need backport.
- We need to investigate more deeply our back buffer flip logic. As any additional fix can convulate logic too much resulting in poor quality in the future.
- URP currently depends on builtin view and proj matrix setup which automatically handles flipping, but that does not apply for other matrices like inverted ones.

---
### Testing status
Manually checked if it works correctly now when changing intermediate texture mode in renderer.

---
### Comments to reviewers


pbbastian (approved):

phi-lira:
We decided to not take this PR as it is.

PR #6176, opened 2021-10-29
Sg/fix/set pvt per material

jonaskhalil:
# **Allow PVT stacks to be bound per material**

### Checklist for PR maker
- [ ] ~~Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).~~
- [ ] ~~Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.~~

---
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1372152/:  Procedural VT does not render materials in a batch properly when several materials using the same shader have different VT-stacks bound to them. The issue is that VT properties are not recognized as "per material", and consequently they are set only once per batch.

With this fix, users are now able to set "Global" or "Per Material" shader declarations in PVT node settings. This is the same workflow as with shader properties, only it's not hidden behind an "Override Property Declaration" flag. Trying to `BindToMaterial()` if a stack is declared "Global" might confuse the SRP-batcher, but that's the same with other shader properties as well. Trying to `BindGlobally()` if a stack is declared "Per Material" will not get picked up (i.e., will not render your PVT data).

image

As a bonus, by exposing a property that can be found by the VT-system to handle _per material_, stack names are now also visible in the inspector. Just like any other property, the stack name is needed for binding.
image

---
### Testing status
Tested on Windows PC in HDRP (VT not supported in URP):
- [x] Rendering with material-bound PVT (with a different stack per material) works
- [x] and uses SRP Batcher path
- [x] Rendering with global-bound PVT still works
- [x] and uses SRP Batcher path
- "_Rendering material-bound PVT when the same stack name is also used for binding globally_" does not work
  - The observed behavior is the same as, for instance, having a Color property in global space, setting a color globally and then also setting a specific color for a specific material: the SRP batcher will use the same color for the entire batch (which makes sense as it isn't aware of that color being set per-material)
  - The case of "BindToMaterial() when no PVT is declared as per-material" can easily be detected at native side, in which case we can act on it (e.g., we can block the BindToMaterial-call, we can show warnings, ...)

---
### Comments to reviewers
This work also needs https://ono.unity3d.com/unity/unity/pull-request/133732 to actually fix the issue; without it, these new ProceduralTextureStack properties basically remain unused, they simply aren't picked up by the VT system.


com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/ProceduralVirtualTextureNode.cs
```diff
@@ -21,8 +27,8 @@ public ProceduralVirtualTextureNode()
 
             vtProperty.displayName = "ProceduralVirtualTexture";
             vtProperty.overrideReferenceName = "MyPVT";
-            vtProperty.generatePropertyBlock = false;
             vtProperty.value.procedural = true;
+            vtProperty.value.bindPerMaterial = true;
```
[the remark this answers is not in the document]
debaetsd: I agree that new nodes should indeed have bindPerMaterial=true as that is by far the most useful.
For existing shaders, I guess keeping it as-is is fine (though imho if we clearly document this, changing it is also a valid option).

com.unity.shadergraph/Editor/Data/Graphs/VirtualTextureShaderProperty.cs
```diff
@@ -60,6 +60,25 @@ internal override void AppendPropertyBlockStrings(ShaderStringBuilder builder)
...
+                    }
+                }
+                else
+                {
+                    // adds properties in this format so: [GlobalProceduralTextureStack.MyStack(0)] [NoScaleOffset] MyStack_0("Global Procedural Stack Placeholder", 2D) = "white" {}
+                    for (int layer = 0; layer < value.layers.Count; layer++)
+                    {
+                        builder.AppendLine($"{hideTagString}[GlobalProceduralTextureStack.{referenceName}({layer})][NoScaleOffset]{referenceName}_{layer}(\"{"Global Procedural Stack Placeholder"}\", 2D) = \"white\" {{}}");
+                    }
+                }
```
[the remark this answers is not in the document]
debaetsd: Initiallly I was going to suggest to remove it but that first reason (listed in the properties) is quite valid. 
Having it for "some future optimalisations that are not even planned atm"  is just bloatware imho. 
Unless we can clearly communicate in a different way to the user that the shader is having a PVT stack, I would keep it

debaetsd (approved):

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/ProceduralVirtualTextureNode.cs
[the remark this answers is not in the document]
cdxntchou: You can add a new node version, then on upgrade of old nodes, enable overrideHLSLDeclaration and set hlslDeclarationOverride to HLSLDeclaration.Global on the vtProperty.value.

Can look at Vector1ShaderProperty.cs as an example of versioning (though here you would version the PVT node):

        public override int latestVersion => 1;
        public override void OnAfterDeserialize(string json)
        {
            if (sgVersion == 0)
            {
                // change declaration values here
                ChangeVersion(1);
            }
        }

You will probably need to use OnAfterMultiDeserialize() instead of OnAfterDeserialize() to ensure that the vtProperty has been deserialized before you go changing it -- MultiDeserialize gets called after everything has been deserialized.

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/ProceduralVirtualTextureNode.cs
```diff
@@ -83,6 +91,21 @@ public int layers
             }
         }
 
+        [EnumControl("Bind Method")]
+        public BindMethod bindMethod
```
cdxntchou, line 95: Can we use the HLSLDeclaration instead?

This probably means using a property drawer UI for the node, though, so you can filter out the HLSLDeclaration.PerInstance from the set of possibilities.

But property drawers (that show up in the node inspector) are preferred over adding more controls to the node itself, in general.

You can look at SampleTexture2DNodePropertyDrawer as an example.

And also BuildHLSLDeclarationOverrideFields() in ShaderInputPropertyDrawer.cs for how it constructs a filtered enum control for HLSLDeclarations.

cdxntchou (changes requested):

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/ProceduralVirtualTextureNode.cs
jonaskhalil: Yes, this has been updated to an `internal HLSLDeclaration shaderDeclaration`; internal as it is now exposed in the node inspector rather than on the node itself. I also stuck with "shader declaration" labeling to have it more consistent with other shader properties.

com.unity.shadergraph/Editor/Data/Graphs/SerializableVirtualTexture.cs
```diff
@@ -57,5 +57,8 @@ internal sealed class SerializableVirtualTexture
 
         [SerializeField]
         public bool procedural;
+
+        [SerializeField]
+        public bool bindPerMaterial;
```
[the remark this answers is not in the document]
jonaskhalil: Yes, `SerializableVirtualTexture` now has a `[SerializeField] public HLSLDeclaration shaderDeclaration;`. I also stuck with "shader declaration" labeling to have it consistent with other, "real" shader properties

com.unity.shadergraph/Editor/Data/Graphs/VirtualTextureShaderProperty.cs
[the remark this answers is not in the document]
jonaskhalil: Had my doubts as well, but to me it looks better this way. I've also had a look at the "communication value" of having it here, and realized that, for PVT at least, there's no need to declare a stack layer-by-layer. So, I've updated the code such that we only declare one "stack":
image

The first two props are SVT (just to show that it's not really consistent between SVT and PVT); the 3rd prop is my PVT node that I have in the graph. I think naming makes more sense now, as it is now actually relevant for the user (before, there were those _layer names_ such as `StackA_0` that were only there as non-functional placeholders and were "somehow" derived from the stack name). In this case, `StackA` is the literal name that users need for binding.

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/ProceduralVirtualTextureNode.cs
```diff
@@ -21,8 +21,8 @@ public ProceduralVirtualTextureNode()
 
             vtProperty.displayName = "ProceduralVirtualTexture";
             vtProperty.overrideReferenceName = "MyPVT";
-            vtProperty.generatePropertyBlock = false;
             vtProperty.value.procedural = true;
+            vtProperty.value.shaderDeclaration = HLSLDeclaration.UnityPerMaterial;
```
jonaskhalil, line 25: See [earlier conversation](https://github.com/Unity-Technologies/Graphics/pull/6176/files/63c3accd5db4a90400a363794e895796d3246b53#r739337016), this is still forcing _all_ nodes initially at UnityPerMaterial (both new and upgraded). 

- Should we have the overhead of version tracking if PVT is not really exposed yet (versus just "good communication")?
- Given that it's only used internally; do we prefer
  - "no project changes behavior"
  - or "all nodes start at the most common PerMaterial", potentially breaking projects, and also potentially fixing many projects where it wasn't clear that things were going subtly wrong ...

cdxntchou: I tend to lean towards "no project changes behavior" most of the time.  Only if we are near 100% sure we can tell what the user intended and we can fix the behavior for that intent, then we can do the unilateral behavior change.  In this case though, it's not clear if the users meant to use the Global path or the Material path.

What would happen if a user is using Global path and they upgrade to this change that sets their PVT to PerMaterial?

cdxntchou: I also feel like the communication overhead is larger than the versioning overhead in this case..  :)

cdxntchou: Though it is a bit weird in this case, as previously it kind of worked both for the Global path and the Material path (as long as it wasn't instanced), correct?  So we don't really know if the user intent was Global or PerMaterial.  We only know that Global rendered correctly, and PerMaterial rendered correctly sometimes.

If we default old PVT nodes to Global, would it continue to have this behavior?   Working in Global, and working in non-instanced Material?

cdxntchou: The actual code change for versioning is minimal:
`    public override int latestVersion => 1;
    public override void OnAfterMultiDeserialize(string json)
    {
        if (sgVersion == 0)
        {
            // version 0 was hard-coded to Global (kinda..)
            shaderDeclaration = HLSLDeclaration.Global;
            ChangeVersion(1);
        }
    }`

jonaskhalil: Cool, the code change is indeed minimal :-) Have it running locally already, doesn't seem to need anything else to have existing nodes as `Global` after reimporting while indeed creating new nodes as `Per Material`.

jonaskhalil: > Though it is a bit weird in this case, as previously it kind of worked both for the Global path and the Material path (as long as it wasn't instanced), correct? So we don't really know if the user intent was Global or PerMaterial. We only know that the Global path rendered correctly, and Material path rendered correctly when non-instanced.

We only know the user intent at runtime, based on their call to either `m_Stack.BindToMaterial(mat)` or to `m_Stack.BindGlobally()`. It indeed worked globally, but unless I'm confused, it never worked per-material _in an SRP-compatible way_, right (which was why it was changed to global in the first place)? So, yes, it was bound globally, and this indeed kinda worked for per-material use as well as long as it wasn't instanced ... But afaik that's a SRP-specific issue: I have a scene set up which does the exact same with Color properties instead of PVT, and setting `Shader.SetGlobalColor("MyColor", globalColor)` but then also setting `mat.SetColor("MyColor", localColor)` shows the same problem, all materials rendered in the same batch as `mat` will get `localColor`

jonaskhalil: > If we default old PVT nodes to Global, would it continue to have this behavior? Working in Global, and working in non-instanced Material?

Yes, this is the screenshot of my set-up, and it is the same _before_ making shader changes, and _after_ making changes with your suggested addition to make sure the nodes get upgraded with shaderDeclaration set to Global
PVT-Before

On the left column, we should see red, green and blue, but because the first two rows are batch-rendered, they only get one color. The middle column is rendered as expected (a single global color). And the third column is also binding globally (to magenta) but then binding a local stack for the first quad (to yellow); again, because the first two quads are batch-rendered, they have the same color. **Note** that that third column is not fixed in this PR, as it's SRP-specific, not VT.

If we would decide not to version (so, upgraded nodes are also Per Material), my project upgrades to this:
PVT-After-NoVersioning
The left column is fixed because instancing works properly now, but the globally bound cyan and magenta colors are missing. As said, upgrading as such would break all projects where `BindGlobally()` is used.

jonaskhalil: > I also feel like the communication overhead is larger than the versioning overhead in this case.. :)

Weirdly enough, I'm starting to thing that communication would be even more important if we do versioning. If we don't do versioning, we'd have to communicate "_Hey, you'll see that `BindGlobally()` isn't working anymore, that's because you now have to explicitly set your nodes to `shaderDeclaration.Global` for this to work!_". And even if people didn't get the memo, their project will immediately show artifacts, they'll probably point to VT rather quickly (because they're binding a stack but it doesn't bind anymore) and we'd be able to help out.

If we do versioning, I'm not sure if we can do without proper communication? Even though the project didn't break, there are still bugs in there, so we'd still need to communicate the opposite message "_Hey, there were issues with `BindToMaterial()` specifically when rendered in a single SRP batch, even if your project looked fine; this is fixed for newly created PVT nodes, but for upgraded nodes you'll have to manually set your nodes to `shaderDeclaration.PerMaterial`_". If people missed the memo this time, it could potentially result in bug reports months later (depending on how soon the instancing issues show up, and whether or not devs will immediately think of VT instead of thinking "this is some kind of SRP/shader flickering issue")

Or maybe I'm overthinking this; the fact remains that PVT is not publicly available yet so there cannot be that many projects that we break, either way

cdxntchou: True.  There's not going to be a 100% solution I think.   If you think this approach is best, that works for me.

cdxntchou (approved):

PR #6178, opened 2021-10-30
[URP] Upscaling Filters & FXAA Scaling Logic

gmitrano-unity:
### Purpose of this PR
This change adds a new "upscaling filter" property to the pipeline asset that allows users to control which filtering method is used when upscaling occurs within URP.  In the current implementation, nearest-neighbor and bilinear filtering are supported. In the future, other upscaling methods such as FidelityFX Super Resolution (FSR) could be supported as well.

This change also adds a new upscaling setup shader which performs the FXAA logic when scaling is enabled. This will likely have a performance cost, but it improves quality since FXAA wasn't designed to be used during scaling. The image below shows the original logic at 77% scale on the left and the modified logic with 77% scale on the right:

image

This new pass could also be used to perform color space conversion logic for FSR if it's added in the future.

---
### Testing status
Manual tests on Viking Village URP with FXAA on and off + nearest-neighbor vs bilinear filtering on Windows and Android

---
### Comments to reviewers
This change does not modify any of the Renderer2D logic. How should the upscaling filter property interact with the pixel perfect filtering support in the 2d renderer?

---
### Notes for QA
The changes in this PR will be verified for quality as part of the larger FSR PR here: https://github.com/Unity-Technologies/Graphics/pull/6262

---
### Remaining Tasks
- [x]  Update CHANGELOG.md
- [x]  Update Docs


phi-lira:
Adding Sandy and Manuele for coverage on this PR.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -1391,6 +1387,70 @@ void RenderFinalPass(CommandBuffer cmd, ref RenderingData renderingData)
...
+
+                // When FXAA is enabled in scaled renders, we execute it in a separate blit since it's not designed to be used in
+                // situations where the input and output resolutions do not match.
+                if (isFxaaEnabled)
+                {
+                    m_Materials.upscaleSetup.EnableKeyword(ShaderKeywordStrings.Fxaa);
+
+                    isUpscaleSetupTextureUsed = true;
+
+                    cmd.GetTemporaryRT(ShaderConstants._UpscaleSetupTexture, tempRtDesc, FilterMode.Point);
```
manuele-bonanno, line 1410: when merging to master this (and any other RT management code) needs to be changed to use RTHandles as we have deprecated GetTemporaryRT

cc @sandy-carter-unity

manuele-bonanno (changes requested):
looks good to me. The only requested change would be to replace the GetTemporaryRT with RTHandles (just landed in master) since we deprecated temporaryRT allocs

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
gmitrano-unity: @manuele-bonanno & @sandy-carter-unity do my RTHandle changes in 4900e6e look ok?

manuele-bonanno (approved):
Looks good!

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
```diff
@@ -0,0 +1,50 @@
+Shader "Hidden/Universal Render Pipeline/Upscale Setup"
+{
+    HLSLINCLUDE
+        #pragma multi_compile_local_fragment _ _FXAA
+        #pragma multi_compile _ _USE_DRAW_PROCEDURAL
```
phi-lira, line 5: draw procedural only requires vertex variants
```suggestion
        #pragma multi_compile_vertex _ _USE_DRAW_PROCEDURAL
```

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
```diff
@@ -0,0 +1,50 @@
...
+            float2 positionNDC = uv;
+            int2   positionSS = uv * _SourceSize.xy;
+
+            half3 color = SAMPLE_TEXTURE2D_X(_SourceTex, sampler_PointClamp, uv).xyz;
+
+#if _FXAA
+            color = ApplyFXAA(color, positionNDC, positionSS, _SourceSize, _SourceTex);
+#endif
+
+            return half4(color, 1.0);
```
phi-lira, line 29: Is this Setup the "slow" path you mentioned in the design doc. The doc proposes the colorspace to be done inline with FXAA on the setup pass (just before FSR) but I don't see it here. I guess this will be changed in the FSR PR? (I haven't looked at it yet).

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+    ENDHLSL
+
+    SubShader
```
phi-lira, line 34: What is this Subshader user for? From what I understood if we don't have FXAA we don't need the upscale setup as the upscaling could be done in final post?

It would be nice to have some comments on the subshaders and usage for future maintainance / understanding.

com.unity.render-pipelines.universal/Runtime/Data/PostProcessData.cs
```diff
@@ -69,6 +69,9 @@ public sealed class ShaderResources
             [Reload("Shaders/PostProcessing/LensFlareDataDriven.shader")]
             public Shader LensFlareDataDrivenPS;
 
+            [Reload("Shaders/PostProcessing/UpscaleSetup.shader")]
+            public Shader upscaleSetupPs;
```
phi-lira, line 73: Minor suggestion for consistency with other naming.

```suggestion
            public Shader upscaleSetupPS;
```

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -1265,6 +1281,63 @@ static void CheckAndApplyDebugSettings(ref RenderingData renderingData)
...
+
+                    break;
+                }
+
+                case UpscalingFilterSelection.Point:
+                {
+                    filter = ImageUpscalingFilter.Point;
+
+                    break;
+                }
```
phi-lira, line 1335: I don't immediately understand the benefit for expose this to users, in which cases they would select bilinear or point instead of Auto? Or is this a preparation for FSR scaling mode in the follow up PR?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -15,6 +15,33 @@ public enum MixedLightingSetup
         Subtractive,
     };
 
+    /// <summary>
+    /// Enumeration that indicates what kind of image scaling is occurring if any
+    /// </summary>
+    public enum ImageScaling
```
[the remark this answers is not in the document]
phi-lira: I don't follow why this is public / user facing, shouldn't we derive this from the render scale? 
Is there any public API that will consumes this enum?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -15,6 +15,33 @@ public enum MixedLightingSetup
...
+        Upscaling,
+
+        /// Downscaling to a smaller image
+        Downscaling
+    }
+
+    /// <summary>
+    /// Enumeration that indicates what kind of upscaling filter is being used
+    /// </summary>
+    public enum ImageUpscalingFilter
```
phi-lira, line 36: Same as above, is there any public API consuming this mode or should it be internal only?

com.unity.render-pipelines.universal/Shaders/PostProcessing/FinalPost.shader
```diff
@@ -60,58 +42,15 @@ Shader "Hidden/Universal Render Pipeline/FinalPost"
             float2 positionNDC = uv;
             int2   positionSS  = uv * _SourceSize.xy;
 
-            half3 color = Load(positionSS, 0, 0).xyz;
+            #if _POINT_SAMPLING
+            half3 color = SAMPLE_TEXTURE2D_X(_SourceTex, sampler_PointClamp, uv).xyz;
+            #else
+            half3 color = SAMPLE_TEXTURE2D_X(_SourceTex, sampler_LinearClamp, uv).xyz;
+            #endif
```
phi-lira, line 49: This seems like an overkill to duplicate variants for sampler state. Is there a way to configure sampler / filtering mode on RTHandle?

phi-lira (commented):
LGTM. Made some comments but nothing blocking the PR.
 How do you plan to land docs?

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
gmitrano-unity: Yes, this setup shader is the slow path I mentioned in the FSR doc. The color space conversion is indeed added into this shader in the FSR PR. I actually ended up removing the fast path earlier though based on some feedback from the URP team. When I did the optimization, I didn't realize that custom post processing passes are extremely common and the optimization can't work in the case where they're present. Based on that information, I figured it wasn't worth the additional code complexity cost. I could add it back in if we still want it? It's in this commit here: https://github.com/Unity-Technologies/Graphics/commit/26c523d90b0344af5e0edf1294fe7549d6789d56

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
gmitrano-unity: Linear and Point were both added before I actually had figured out how to dynamically switch between them based on the current screen size and render scale like in Auto. I figured we'd keep them in case users are unhappy with the automatic filter's behavior and want to explicitly control it themselves? I could also see how this might make things more confusing though.

Yes, FSR does slot into this enum in the next PR so that's the main reason for it.

com.unity.render-pipelines.universal/Shaders/PostProcessing/FinalPost.shader
gmitrano-unity: I completely agree. :( Unfortunately I didn't see a way to set sampler state from the command buffer API? It seems like you can only set textures and the sampler state is fused to the texture which means that changing it as part of the final pass could potentially affect other parts of the frame since it's changing persistent state on the object rather than specifying a per-draw parameter?

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
phi-lira: Ok. Can you point me to discussion, I guess I need some context to understand the pros / const of that. Othewise all good.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
phi-lira: ScalingSetup makes sense to me.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
phi-lira: Ok, yeah if you think it might be useful it's ok, I don't fully understand what could be the cases they are unhappy thought. If the difference between the two modes end up being artistic choice then this makes sense, but if they are unhappy because auto is not working as expected that kind of implies to me an issue that should be considered before we make this decision.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
gmitrano-unity: I don't know if there's a written discussion on this one, it came out of a meeting we had earlier in November. I believe @Verasl was the one to point this out. You can see what the complexity cost in the commit. It's not particularly large, but it does cause logical coupling between uber post and final post, though I imagine the benefit of skipping a full screen pass could be pretty big on mobile so it might be worth having anyways even if it's usually skipped in most projects?

com.unity.render-pipelines.universal/Shaders/PostProcessing/FinalPost.shader
phi-lira: Ok. That's unfortunate. Then I guess we should at bare least strip the unused variants from build (look at ShaderPreprocessor.cs)

phi-lira (approved):
LGTM the remaining feedback can be addressed in the next PR if you wish.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
phi-lira: Ah I see. Ignore this comment for some reason I saw 2 subshaders. I'm crazy.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
gmitrano-unity: Yeah my thought was that some people may not artistically like the integer scaling behavior from auto. It's sharper but with more visible pixelation? Some people may just prefer the blur from linear instead. (The pixelation depends on display pixel density though)

We'll want to make sure the Auto filter's behavior is clearly documented for users so they're aware that it jumps to integer scaling when at appropriate scale + resolution combos.

phi-lira:
VFX_URP failure fixed on this PR https://github.com/Unity-Technologies/Graphics/pull/6540

PR #6180, opened 2021-11-01
[URP] make the Vulkan RenderTexture MSAA support fallback work in the same way as the swapchain backend implementation

manuele-bonanno:
### Purpose of this PR
The way RenderTextures handle MSAA fallback when an unsupported sample count of 2 is requested (falling back to numSamples = 1), differs fom the way the fallback is handled when setting up the Vulkan swapchain (rounding up numSamples to 4, if supported). This caused an issue on Mali GPUs which don't support 2x MSAA.
This PR makes sure that on Vulkan the MSAA unsupported fallback behaviour is consistent between RenderTextures and Swapchain.

We are also making sure that depth copy is always disabled on GLES, since it doesn't work on that platform.

---
### Testing status
Tested the repro project of [this case](https://fogbugz.unity3d.com/f/cases/1375673).

---
### Comments to reviewers

we should review how all backends handle MSAA fallbacks and move these implementation details in engine code.

check the implementation of SelectSupportedSampleCount(const VkSampleCountFlags supportedSamplesCounts, const VkSampleCountFlagBits requested) in GfxDeviceVK.cpp for more info


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -572,6 +572,13 @@ static GraphicsFormat MakeRenderTextureGraphicsFormat(bool isHdrEnabled, bool ne
             desc.bindMS = false;
             desc.useDynamicScale = camera.allowDynamicResolution;
 
+            if (SystemInfo.graphicsDeviceType == GraphicsDeviceType.Vulkan)
+            {
+                int requestedMsaaSamples = desc.msaaSamples;
+                if (requestedMsaaSamples == 2 && SystemInfo.GetRenderTextureSupportedMSAASampleCount(desc) == 1)
+                    desc.msaaSamples = 4;
+            }
```
phi-lira, line 580: Fine for this PR but we should add a TODO here. Ideally URP should not have platform specific code anymore and this should be handled either in Core Unity

phi-lira (approved):

hkr (approved):

phi-lira (approved):

pushmatrix:
Are there any details you can share about `We are also making sure that depth copy is always disabled on GLES, since it doesn't work on that platform.` ?

My webgl project worked with depthtexture copy in 2020, and updating to 2021 still works if I uncomment the part where you return `false` inside `CanCopyDepth`. I'm curious about what's not working about depth in GLES

manuele-bonanno:
>  I'm curious about what's not working about depth in GLES

It doesn't work when MSAA is enabled. We noticed that we could still allow it when MSAA is off, so we are pushing a fix for that

while waiting for the fix you can do this locally:

 if (IsGLESDevice() && msaaDepthResolve)
                return false;

PR #6110, opened 2021-10-21
[ShaderGraph] Fullscreen ShaderGraph Feature

alelievr:
### Purpose of this PR

Adding the basis to support for Fullscreen effects and Blit: https://unity.productboard.com/portal/1-unity-platform-rendering-visual-effects/tabs/373d7e5c-077c-421a-b0e8-7c2dbadaf31f/features/4181551/portal/expanded

This PR includes:
- A SubTarget to be implemented in each pipeline to add the support of the fullscreen material mode (FullscreenSubTarget)
- The implementation of the Fullscreen mode for Builtin pipeline with compatibility of the Post processing package: https://docs.unity3d.com/Packages/com.unity.postprocessing@3.1/manual/index.html
- Add a new Target for the Custom Render Textures
- Add a new API in SubTarget.cs to know if a node is allowed in this SubTarget
- Add the isMainTexture bool in Texture2DArrayShaderProperty (allows to generate the [MainTexture] for 2D array type)

Fixes:
- Fix generation issue when overriding the default inspector (see Generator.cs)
- Fix missing Read Mask in the stencil descriptor class

Documentation for ShaderGraph Fullscreen: https://docs.google.com/document/d/1hnuSH4uOZt5GLNNr_i1ZShQOPn9FoKwu6ZU437OijZo/edit#
Documentation for CRT: https://docs.google.com/document/d/1Fh_1VgqWWBR4fRDwdzxlcezD6USJU6dOsxVWGa_4Tcs/edit#

If you have any questions regarding this feature, you can ask in [#devs-fullscreen-master-node](https://unity.slack.com/archives/C029VTRHMV0).

### Note for QA:
This work is split into 3 PRs for ease of reviewing. If you want to test everything (URP, HDRP, and builtin implementation) I'd recommend to use the [sg/fullscreen-target](https://github.com/Unity-Technologies/Graphics/tree/sg/fullscreen-target) branch 

---
### Testing status
All implementations have been tested with both URP and HDRP.

Automated tests have been added for HDRP (in a separate PR).

Custom Render Texture still needs verification, as well as the Builtin custom post processes.

Test document: https://confluence.unity3d.com/plugins/servlet/samlsso?redirectTo=%2Fpages%2Fviewpage.action%3FspaceKey%3DHDRP%26title%3DFullscreen%2BTarget

[...]

PR #6185, opened 2021-11-02
[HDRP] Fix infitnite material import loop materials

alelievr:
### Purpose of this PR
Fix the infinite loop in HDRP material importer when the shader is a ShaderGraph but don't generate any HDMetaData.

---
### Testing status
Manually tested inside the fullscreen feature PR: https://github.com/Unity-Technologies/Graphics/pull/6110


com.unity.render-pipelines.high-definition/Editor/AssetProcessors/MaterialPostProcessor.cs
```diff
@@ -371,6 +372,14 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
                     // we would miss re-importing that dependency.
                     if (MaterialReimporter.CheckHDShaderGraphVersionsForUpgrade("", material.shader, ignoreNonHDRPShaderGraphs: false))
                     {
+                        s_ImportedMaterialCounter.TryGetValue(asset, out var importCounter);
+                        s_ImportedMaterialCounter[asset] = ++importCounter;
+
+                        // CheckHDShaderGraphVersionsForUpgrade always return true if a ShaderGraph don't have an HDMetaData attached
```
sebastienlagarde, line 378: question, can't we check that the shadergraph currently have or not the HDMetaData  data instead of relying on importCounter info?

alelievr: It's complicated because an HD ShaderGraph could not have an HDMetaData if it's old, in this case, the ShaderGraph needs to be reimported along with the material. It's only after this reimport that we can know if the material has or not an HDMetaData (i.e. the second time the material is imported).

remi-chapelain (approved):

PR #6188, opened 2021-11-02
Post-Process Volume Curve Blending

johnhableunity:
### Purpose of this PR

This PR add curve blending in post processing nodes. When post process volumes have a curve (such as for the sky), the current behavior overrides one curve with another. This PR creates a new curve by interpolating the previous curves together.

---

https://user-images.githubusercontent.com/93157899/139954780-14846ba2-dae8-43e8-8a42-55fe49325e88.mp4

### Testing status

Testing was performed manually by creating custom curves and merging between them. I created several curves trying to make cases where both lhs and rhs curves had more/less elements to test all the paths. Also, because of the trickiness with debugging something that only happens in a transition, there are several extra asserts. If there is a bug in the way the curves are traversed, the asserts should pick it up.

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
+++ b/com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
@@ -82,7 +82,7 @@ namespace UnityEngine.Rendering
         /// the first key might have inTangent=3.0f but the actual incoming tangent is 0.0 because the curve is
         /// clamped outside the time domain. So this helper fetches a key, but zeroes out the inTangent of the first
         /// key and the outTangent of the last key.
-        static private Keyframe GetKeyframeAndClampEdge([DisallowNull] Keyframe[] keys, int index)
+        static private Keyframe GetKeyframeAndClampEdge([DisallowNull] NativeArray<Keyframe> keys, int index)
         {
             var lastKeyIndex = keys.Length - 1;
             if (index < 0 || index > lastKeyIndex)
@@ -105,7 +105,7 @@ namespace UnityEngine.Rendering
 
         /// Fetch a key from the keys list. If index<0, then expand the first key backwards to startTime. If index>=keys.length,
         /// then extend the last key to endTime. Keys must be a valid array with at least one element.
-        static private Keyframe FetchKeyFromIndexClampEdge([DisallowNull] Keyframe[] keys, int index, float segmentStartTime, float segmentEndTime)
+        static private Keyframe FetchKeyFromIndexClampEdge([DisallowNull] NativeArray<Keyframe> keys, int index, float segmentStartTime, float segmentEndTime)
         {
             float startTime = Mathf.Min(segmentStartTime, keys[0].time);
             float endTime = Mathf.Max(segmentEndTime, keys[keys.Length - 1].time);
@@ -174,7 +174,7 @@ namespace UnityEngine.Rendering
 
         /// lhsIndex and rhsIndex are the indices in the keys array. The lhsIndex/rhsIndex may be -1, in which it creates a synthetic first key
         /// at startTime, or beyond the length of the array, in which case it creates a synthetic key at endTime.
-        static private Keyframe EvalKeyAtTime([DisallowNull] Keyframe[] keys, int lhsIndex, int rhsIndex, float startTime, float endTime, float currTime)
+        static private Keyframe EvalKeyAtTime([DisallowNull] NativeArray<Keyframe> keys, int lhsIndex, int rhsIndex, float startTime, float endTime, float currTime)
         {
             var lhsKey = KeyframeUtility.FetchKeyFromIndexClampEdge(keys, lhsIndex, startTime, endTime);
             var rhsKey = KeyframeUtility.FetchKeyFromIndexClampEdge(keys, rhsIndex, startTime, endTime);
@@ -213,18 +213,28 @@ namespace UnityEngine.Rendering
 
                 for (int i = 0; i < rhsCurve.length; i++)
                 {
-                    lhsAndResultCurve.AddKey(rhsCurve.keys[i]);
+                    lhsAndResultCurve.AddKey(rhsCurve[i]);
                 }
                 lhsAndResultCurve.postWrapMode = rhsCurve.postWrapMode;
                 lhsAndResultCurve.preWrapMode = rhsCurve.preWrapMode;
             }
             else
             {
-                // Note: If we reached this code, we are guaranteed that both lhsCruve and rhsCurve are valid with at least 1 key
+                // Note: If we reached this code, we are guaranteed that both lhsCurve and rhsCurve are valid with at least 1 key
 
-                // first, figure out the start and end time to include both curves
-                var lhsCurveKeys = lhsAndResultCurve.keys;
-                var rhsCurveKeys = rhsCurve.keys;
+                // create a native array for the temp keys to avoid GC
+                var lhsCurveKeys = new NativeArray<Keyframe>(lhsAndResultCurve.length, Allocator.Temp);
+                var rhsCurveKeys = new NativeArray<Keyframe>(rhsCurve.length, Allocator.Temp);
+
+                for (int i = 0; i < lhsAndResultCurve.length; i++)
+                {
+                    lhsCurveKeys[i] = lhsAndResultCurve[i];
+                }
+
+                for (int i = 0; i < rhsCurve.length; i++)
+                {
+                    rhsCurveKeys[i] = rhsCurve[i];
+                }
 
                 float startTime = Mathf.Min(lhsCurveKeys[0].time, rhsCurveKeys[0].time);
                 float endTime = Mathf.Max(lhsCurveKeys[lhsAndResultCurve.length - 1].time, rhsCurveKeys[rhsCurve.length - 1].time);


com.unity.shadergraph/Documentation~/Fade-Transition-Node.md
```diff
@@ -0,0 +1,34 @@
+# Fade Transition Node
+
+## Description
+
+Fade Transition is a method of adding noise to add variation while a function transitions from on to off. This node takes in a fade value and remaps it using the noise value (usually from a texture). When FadeValue is 0, the output is always 0, and when FadeValue is 1, the output is always exactly 1. In between 0 and 1 the transition will follow the pattern in the noise.
```
sebastienlagarde, line 5: Note: looks like you have started your PR from the branch you previously used. Be careful with that you must start from master branch for reach new feature. Will merge master in this branch

sebastienlagarde (approved):
Looks good to me, but prefer to have the review of Remi (RP Workflow) for this PR

sebastienlagarde:
Note that you will need to trigger Yamato. We can sync on it so I can show you (or kleber can show you)

sebastienlagarde:
Lastly you will need to add an update in the change log of Core  (it is a file name Changelog.md in Core package)

sebastienlagarde:
Thanks for this work!

com.unity.shadergraph/Documentation~/Fade-Transition-Node.md
johnhableunity: Thanks, good to know. I'm a bit of a newb with git.

TomasKiniulis (approved):
Checked Curves Mapping Exposure, Volumetric Clouds in Scene/Game views made sure Post Process Effects are not affected. 

Volumetric Clouds
https://user-images.githubusercontent.com/50582134/140545614-caea1320-bcc0-4e06-b1a8-9a6c33c5e780.mp4

Curve Mapping Exposure
https://user-images.githubusercontent.com/50582134/140542602-b6d9f5c1-3206-4b0d-84cf-dbee0582bbf1.mp4

It works nicely, but it seems Color Curves uses it's own curves system and no blending occurs. Would it be possible to have that implemented in the future?

https://user-images.githubusercontent.com/50582134/140544821-3e5d34bf-da52-47b5-8add-647900cfa9b7.mp4

In general there seems to be a recent issue with Color Curves not updating when values are changed unless Volume is reenabled or local volume is reentered, but does not seem to be this pr related. Reported a bug here:  https://fogbugz.unity3d.com/f/cases/1378596/

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+                Keyframe[] dstKeys = new Keyframe[maxNumKeys];
+
+                int lhsKeyCurr = 0;
+                int rhsKeyCurr = 0;
+
+                while (lhsKeyCurr < lhsCurve.keys.Length && rhsKeyCurr < rhsCurve.keys.Length)
+                {
+                    // the index is considered invalid once it goes off the end of the array
+                    bool lhsValid = lhsKeyCurr < lhsCurve.keys.Length;
+                    bool rhsValid = rhsKeyCurr < rhsCurve.keys.Length;
```
adrien-de-tocqueville, line 1794: I think these two variables are always true given the end condition of the loop. Code below can be simplified

johnhableunity: Hrm, good catch. That actually looks like a bug. It should be OR, not AND. I'll dig into it and update code.

pierre-unity:
Working well with clouds! Couldn't really find any issues for this use case.

https://user-images.githubusercontent.com/39901544/141133685-78f473b0-bb75-4392-955d-fab89dd021c2.mp4

sebastienlagarde:
Alright, all good, waiting for @phi-lira approval and sanity check from @RSlysz

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+            {
+                m_Value = lhsCurve;
+            }
+            else if (t >= 1.0f)
+            {
+                m_Value = rhsCurve;
+            }
+            else
+            {
+                m_Value = new AnimationCurve();
```
adrien-de-tocqueville, line 1771: Also, this assignment is useless, there is another `new AnimationCurve()` at the end of the function that overwrites it.
But shouldn't there be some sort of cache for these interpolated value, it can add several objects to garbage collection every frame, or maybe it's negligible ?

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+                            //     lhsKey[curr-1].time <= rhsKey.time <= lhsKey[curr].time
+                            // so interpolate lhsKey at the rhsKey.time.
+                            lhsKey = EvalKeyAtTime(lhsCurve.keys, lhsKeyCurr - 1, lhsKeyCurr, startTime, endTime, rhsKey.time);
+                            rhsKeyCurr++;
+                        }
+                        else
+                        {
+                            // should never happen, checking here to be safe
+                            Assert.IsTrue(false);
+                        }
```
RSlysz, line 1830: This effectively cannot happen
if(x == y) ... //as time is float, equality of float can be tricky. Try perhaps Mathf.Approximately
else if (x < y) ...
else if (x > y) ... //can remove the condition
else ... //can be removed

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+                        // lhs will be evaluated between the last lhs key and the extrapolated lhs key at the end time
+                        lhsKey = EvalKeyAtTime(lhsCurve.keys, lhsKeyCurr - 1, lhsKeyCurr, startTime, endTime, rhsKey.time);
+
+                        rhsKeyCurr++;
+                    }
+                    else
+                    {
+                        // at least one of lhsValid or rhsValid should always be true
+                        Assert.IsTrue(false);
+                    }
```
RSlysz, line 1856: this also cannot happen
while (a && b) //of even (a || b)
{
   //you cannot have both a and b that are false here
}

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
RSlysz: instead of duplicating the condition check, try perhaps moving this outside of the loop and update it at the end of the loop. Here, first time we do this check two time for no reason.

RSlysz (changes requested):
Please check the potential performence impact as there is object created at each interpolation of curves. (and then GC)

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
johnhableunity: Generally, I like to keep the else in even if it can't happen with an assert, so if someone changes the code later it's more explicitly clear. But if it's easier, I'll just remove the else and put in a comment. The first (x==y) is by design as even if x and y are very close except for a small epsilon, it's still preferable to add two separate points. I thought about trying to merge points based on an epsilon, but it creates a bunch of corner cases where the interpolation might pop in unexpected ways.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+                // if we have any keys at the same time in both curves, there will be extras in the array
+                // so trim them.
+                if (currNumKeys != dstKeys.Length)
+                {
+                    Keyframe[] trimKeys = new Keyframe[currNumKeys];
+                    for (int i = 0; i < currNumKeys; i++)
+                    {
+                        trimKeys[i] = dstKeys[i];
+                    }
+                    dstKeys = trimKeys;
```
alex-vazquez, line 1873: You can use [Array.Copy](https://docs.microsoft.com/en-us/dotnet/api/system.array.copy?view=net-5.0)

```suggestion
Keyframe[] trimKeys = new Keyframe[currNumKeys];
Array.Copy(dstKeys, 0, trimKeys , 0, currNumKeys);
dstKeys = trimKeys;
```

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+                }
+
+                m_Value = new AnimationCurve(dstKeys);
```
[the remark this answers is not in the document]
alex-vazquez: As the overall improvements on the stack the components should be structs, But this garbage will be only created when you are in the middle of two volumes that override the curve, so I do not think is a big deal. Maybe @fredericv-unity3d has other options here.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
RSlysz: I personally found the else confusing and there is an added unneeded test even if the cost is really small.
This is code that we will maintain so please do the change. It will be simpler for us to go back onto it.

For the float equality, do as you prefer for this. I don't know exactly what is your issue here but I believe you have think about it and think it is better this way.

johnhableunity: Sounds good. I'll fix it. My bottleneck is that at the moment, the HDRP sample scene isn't working in my version of the engine, so I'm trying to figure out how to get a sample scene up and running. Then I'll submit the changes.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
RSlysz: We don't allow GC at runtime in Graphics repository. This will create GC at each interpolation so at each frame if such an animation curve exists in the scene. This is normally detected by one of the automated test but I am not sure he will be able to catch it as I dunno if the scene it use use this new interpolation scheme.

johnhableunity: Ok. What's the best solution then. Create a pool?

RSlysz: There is several point that can be improved, like you can work in usafe and alocate your array on the stack (you create an array of keyframe at some point). But ultimately, if you really want it to be an AnimationCurve, then yes. I don't see any solution beside having a pool. 
Alternatively, you need to define an AnimationCurve like struct that is interpolatable and use it here. 

We are taking a deeper look into this PR in the team. There is improvement on the code side (naming, preventing miss-use, etc). We will check if we see any other good way to handle this interpolation without causing GC. I'll have all the answear Monday :)

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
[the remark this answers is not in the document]
johnhableunity: Just looked through the code again. The AnimationCurve object is instantiated, and it has native members for adding keys. So I should be able to call AnimationCurve.AddKey() to add the keys which will not create garbage since the memory is handled in native code.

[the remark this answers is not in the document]
johnhableunity: Actually, after looking at it, the m_Value is not always non-null, so I'll have to wait to see what the plan is for creating a pool of them.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+            Assert.IsTrue(keys.Length >= 1);
+            Assert.IsTrue(startTime <= keys[0].time);
+            Assert.IsTrue(keys[keys.Length - 1].time <= endTime);
+
+            Keyframe lhsKey = FetchKeyFromIndexClamped(keys, lhsIndex, startTime, endTime);
+            Keyframe rhsKey = FetchKeyFromIndexClamped(keys, rhsIndex, startTime, endTime);
+
+            float currValue = 0.0f;
+            float currDeriv = 0.0f;
+            EvalCurveSegmentAndDeriv(ref currValue, ref currDeriv, lhsKey, rhsKey, currTime);
```
RSlysz, line 1754: Can be simplified. No need to declare the floats in advance if using out with this sintax:
`EvalCurveSegmentAndDeriv(out var currValue,    ...`

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+
+        // lhsIndex and rhsIndex are the indices in the keys array. The lhsIndex/rhsIndex may be -1, in which it creates a synthetic first key
+        // at startTime, or beyond the length of the array, in which case it creates a synthetic key at endTime.
+        static private Keyframe EvalKeyAtTime(Keyframe[] keys, int lhsIndex, int rhsIndex, float startTime, float endTime, float currTime)
+        {
+            Assert.IsTrue(keys.Length >= 1);
+            Assert.IsTrue(startTime <= keys[0].time);
+            Assert.IsTrue(keys[keys.Length - 1].time <= endTime);
+
+            Keyframe lhsKey = FetchKeyFromIndexClamped(keys, lhsIndex, startTime, endTime);
```
RSlysz, line 1749:
```suggestion
            var lhsKey = FetchKeyFromIndexClamped(keys, lhsIndex, startTime, endTime);
```

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+
+            float t = Mathf.Clamp(currTime - lhsKey.time, 0.0f, dx);
+
+            dstValue = (t * (t * (t * c0 + c1) + c2)) + c3;
+            dstDeriv = (t * (3.0f * t * c0 + 2.0f * c1)) + c2;
+        }
+
+        // lhsIndex and rhsIndex are the indices in the keys array. The lhsIndex/rhsIndex may be -1, in which it creates a synthetic first key
+        // at startTime, or beyond the length of the array, in which case it creates a synthetic key at endTime.
+        static private Keyframe EvalKeyAtTime(Keyframe[] keys, int lhsIndex, int rhsIndex, float startTime, float endTime, float currTime)
```
RSlysz, line 1743: If you need to add comment onto a method, please add it with /// so it is available when trying to use it (intellisene)

This should also be part of `KeyframeUtility`, with unit tests.

There is missing checks performed here. For instance you can have null array.
Should we handle case where endTime is lower than startTime or if currTime is not in between?
Can indices be negatives?

Can fail (assertion).
Also there is many reason for a fail here and most of them are not explained in the comment/documentation. 
```
Assert.IsTrue(startTime <= keys[0].time);
Assert.IsTrue(keys[keys.Length - 1].time <= endTime);
```

Naming of this method is unclear. I propose this instead
```suggestion
        bool TryEvaluateOnSegment([DisallowNull] Keyframe[] keys, uint startSegmentIndex, uint endSegmentIndex, float startTime, float endTime, float currentTime)
```

Yes it is the same name. Because it basically do the same. It is just an helper to find start and end of segment.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+        {
+            Assert.IsTrue(keys.Length >= 1);
+            Assert.IsTrue(startTime <= keys[0].time);
+            Assert.IsTrue(keys[keys.Length - 1].time <= endTime);
+
+            Keyframe lhsKey = FetchKeyFromIndexClamped(keys, lhsIndex, startTime, endTime);
+            Keyframe rhsKey = FetchKeyFromIndexClamped(keys, rhsIndex, startTime, endTime);
```
RSlysz, line 1750:
```suggestion
            var rhsKey = FetchKeyFromIndexClamped(keys, rhsIndex, startTime, endTime);
```

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+
+            float currValue = 0.0f;
+            float currDeriv = 0.0f;
+            EvalCurveSegmentAndDeriv(ref currValue, ref currDeriv, lhsKey, rhsKey, currTime);
+
+            Keyframe dstKey = new Keyframe(currTime, currValue, currDeriv, currDeriv);
+            return dstKey;
+        }
+
+        public override void Interp(AnimationCurve lhsCurve, AnimationCurve rhsCurve, float t)
```
RSlysz, line 1760: Missing unit test for this and checking 0 GC allocation. We don't had time to fully discuss the GC issue today. But yes pooling can be a solution. (What size?) It can perhaps also be possible to cache per AnimationCurve pairs. So it would only create data the first frame then reuse created object next frame.

Public API should always be fully documented, with /// for intellisens (It will create tech debt for us to make it at release of package if it is not there.)

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+        {
+            if (t <= 0.0f)
+            {
+                m_Value = lhsCurve;
```
RSlysz, line 1764: You can `return;` after this point. It will make code easier to read and will prevent scope indentation

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+            }
+            else if (t >= 1.0f)
+            {
+                m_Value = rhsCurve;
```
RSlysz, line 1768: You can `return;` after this point. It will make code easier to read and will prevent scope indentation

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1642,239 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+            else
+            {
+                m_Value = new AnimationCurve();
+
+                Assert.IsTrue(lhsCurve.keys.Length >= 1);
+                Assert.IsTrue(rhsCurve.keys.Length >= 1);
+
+                // first, figure out the start and end time to include both curves
+                float startTime = Mathf.Min(lhsCurve.keys[0].time, rhsCurve.keys[0].time);
+                float endTime = Mathf.Max(lhsCurve.keys[lhsCurve.length - 1].time, rhsCurve.keys[rhsCurve.length - 1].time);
```
RSlysz, line 1778: Each time you use `lhsCurve.keys`, a copy is made. Use a variable to store it instead. Same for every array returned by any method that is not a ref return.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1643,227 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+
+                        rhsKeyCurr++;
+                    }
+
+                    Keyframe dstKey = LerpKeyframeDirect(lhsKey, rhsKey, t);
+                    dstKeys[currNumKeys] = dstKey;
+                    currNumKeys++;
+                }
+
+                m_Value = new AnimationCurve();
```
RSlysz, line 1859: This still create GC. Try to find a way to cache it.

RSlysz (changes requested):
So @Unity-Technologies/gfx-srp-workflows did a full review pass. There is several point that prevent us to accept this PR in this current state:
- Missing of public API documentation.
- Creation of GC at frame in the render loop if there is a Volume with such an AnimationCurve.
- Miss of clarification in the naming of methods. But also in variables everywhere. As a general rules, using abbreviation is not good for code clarity.
- Miss in Unity code convention regarding usage of `var`.
- Miss of test for this new feature. 

Additionally, we would really like having the management of the Keyframe be in a separate file utility as it can be reusable this way.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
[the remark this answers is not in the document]
adrien-de-tocqueville: i can also suggest using `TextureCurveParameter` instead of `AnimationCurveParameter`. I believe it will make caching much easier because you won't need a pool. Then you will need a migration step to update the volumetric clouds though

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
johnhableunity: PR has been updated. Please take a look. Major changes:
1. Moved to a separate KeyframeUtility.cs file.
2. Added tests.
3. To solve the GC issue, the volumes are always blended with the temporary volume structure on the left side (it's already hardcoded to 32 volumes). So instead of lerping lhs and rhs and returning the result, the new version lerps them, and re-uses the lhs AnimationCurve class.

phi-lira (changes requested):
Is there any limitation that prevents adding the `RenderInterpolateAnimationCurve` test to CoreRP? That would be not ideal, but if the case then test should be added to URP as well.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
johnhableunity: Also, I think Felipe is right about the test location. Moved it to under core (VolumeComponentTests.cs) as that seems like a more logical place for the test to live.

sebastienlagarde:
To the various reviewver, John have adressed the feedback, could you re-reviewed? Thanks (cc @phi-lira  / @RSlysz ).

To John, there is a formatting issue in the code, (check by our auto test), I will fix it for you and explain you on slack what it is.

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
```diff
@@ -0,0 +1,294 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection;
+using Unity.Collections;
+using UnityEngine.Assertions;
+
+namespace UnityEngine.Rendering
+{
+    /// <summary>
+    /// A helper function for interpolating AnimationCurves together. In general, curves can not be directly blended
```
alex-vazquez, line 14: It would be nice if you add an example to this documentation `<example></example>`

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
```diff
@@ -0,0 +1,294 @@
...
+            var currKey = keys[index];
+            if (index == 0)
+            {
+                currKey.inTangent = 0.0f;
+            }
+            if (index == keys.Length - 1)
+            {
+                currKey.outTangent = 0.0f;
+            }
+            return currKey;
```
alex-vazquez, line 75:
```suggestion
            var keysSize = keys.Length -1:
            if (index < 0 || index > keysSize)
               return new KeyFrame();
               
            var currKey = keys[index];
            if (index == 0)
            {
                currKey.inTangent = 0.0f;
            }
            else if (index == keysSize)
            {
                currKey.outTangent = 0.0f;
            }
            return currKey;
```

I think we should think about this, or either return null but this would mean that code handling this should handle null.

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
```diff
@@ -0,0 +1,294 @@
...
+                    bool rhsValid = rhsKeyCurr < rhsCurveKeys.Length;
+
+                    // it's actually impossible for lhsKey/rhsKey to be uninitialized, but have to
+                    // add initialize here to prevent compiler erros
+                    var lhsKey = new Keyframe();
+                    var rhsKey = new Keyframe();
+                    if (lhsValid && rhsValid)
+                    {
+                        lhsKey = GetKeyframeAndClampEdge(lhsCurveKeys, lhsKeyCurr);
+                        rhsKey = GetKeyframeAndClampEdge(rhsCurveKeys, rhsKeyCurr);
```
alex-vazquez, line 226: If you apply my suggestion on the `GetKeyframeAndClampEdge` you will be able to initialize directly `lhsKey` and `rhsKey` by calling the method, and reduce the cyclomatic complexity of the loop.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
alex-vazquez: Can we use `com.unity.render-pipelines.core\Runtime\Common\ObjectPools.cs`?

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
```diff
@@ -180,5 +180,66 @@ public void TestSupportedOnAvoidedIfHideInInspector()
...
         }
+
+        static private void TestAnimationCurveInterp(AnimationCurve lhsCurve, AnimationCurve rhsCurve, float t, float startTime, float endTime, int numSteps, float eps, bool debugPrint)
+        {
+            AnimationCurve midCurve = new AnimationCurve(lhsCurve.keys);
+            KeyframeUtility.InterpAnimationCurve(ref midCurve, rhsCurve, t);
+
+            for (int i = 0; i <= numSteps; i++)
+            {
+                float timeT = ((float)i) / ((float)numSteps);
```
alex-vazquez, line 191: Do you need both casts?

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
```diff
@@ -180,5 +180,66 @@ public void TestSupportedOnAvoidedIfHideInInspector()
...
+
+                float expectedVal = Mathf.Lerp(lhsVal, rhsVal, t);
+
+                float actualVal = midCurve.Evaluate(currTime);
+
+                float offset = actualVal - expectedVal;
+                if (debugPrint)
+                {
+                    Debug.Log(i.ToString() + ": " + offset.ToString());
+                }
```
alex-vazquez, line 205: As this is code from a test, and we will need to put the parameter to true, I think it will be ok to leave the Debug.Log under a comment.

com.unity.render-pipelines.high-definition/Tests/Editor/HDRenderUtilitiesTests.cs
```diff
@@ -38,5 +38,6 @@ public void RenderThrowWhenTargetIsNotARenderTextureForTex2DRendering()
                 Texture2D.whiteTexture
             ));
         }
+
```
alex-vazquez, line 41: Do you need this space? Did you add this file by mistake?

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// function isn't allowed to fail, so in the case where we run out of memory we fall back to returning a single keyframe.
+    /// </summary>
+    public class KeyframeUtility
+    {
+        /// <summary>
+        /// Helper function to remove all control points for an animation curve. Since animation curves are reused in a pool,
+        /// this function clears existing keys so the curve is ready for reuse.
+        /// </summary>
+        /// <param name="curve">The curve to reset.</param>
+        static public void ResetAnimationCurve(ref AnimationCurve curve)
```
alex-vazquez, line 33: Do you need `ref`? You are not doing any assignment to `curve`. Besides this, I think this method can be directly into trunk,  as it seems helpfull for users.

If we decide to not put it into trunk, I would recommend you to create an extension class for animation curve.

```
namespace UnityEngine
{
    public static class AnimationCurveExtensions
    {
        static public void Reset(this AnimationCurve curve)
        {
            int numPoints = curve.length;
            for (int i = numPoints - 1; i >= 0; i--)
            {
                curve.RemoveKey(i);
            }
        }
     }
}
```

In this way we could do something like `animationCurve.Reset();` and if someday someone creates that reset method it will be easy changed.

alex-vazquez (changes requested):
Nice job, some minor things, I would like to simplify some of the code on the loop if possible, this will help us adding test and having a good test coverage.

RSlysz (changes requested):
Some change needed on the test to be able to better find issue if they occurs.

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
johnhableunity: Sure. Adding one now.

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
johnhableunity: The concern is that returning invalid data (due to a programmer bug) is often worse than stopping right away (either through an assert() or an OOB exception). Suppose a programmer introduces a bug, and we silently return an empty keyframe. The user likely wouldn't know, and we would end up with a very difficult bug to track down. I.e. at some point a QA person realizes that there is a small stutter when transitioning between volumes.

Personally, I don't have any strong feelings on the actual code structure. But it is important that if we have a bug, that we can detect it right away instead of silently hiding it. So I'm including your change, but also adding a debug warning to the log before the return statement.

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
johnhableunity: How exactly would that work? GetKeyframeAndClampEdge() needs to adjust the start and end time based on the segment start, whereas the other calls to GetKeyframeAndClampEdge() need to get the key within the current curve (not including segment start and end).

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
johnhableunity: If we only do one of the casts, both casts will stall have to happen since the compiler will silently convert one of the ints to a float. So I wrote it this way to make it more explicit. But I don't have strong feelings either way.

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
johnhableunity: Commenting it out.

com.unity.render-pipelines.high-definition/Tests/Editor/HDRenderUtilitiesTests.cs
johnhableunity: It's a mistake. The test was originally in that file, so it looks like a mistake in reverting the file.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -1641,14 +1643,30 @@ internal override void Interp(VolumeParameter from, VolumeParameter to, float t)
...
+            : base(value, overrideState)
+        {
+        }
 
-        // TODO: Curve interpolation
+        /// <summary>
+        /// Interpolates between two AnimationCurve values. Note that it will overwrite the values in lhsCurve,
+        /// whereas rhsCurve data will be unchanged. Thus, it is legal to call it as:
+        ///     stateParam.Interp(stateParam, toParam, interpFactor);
+        /// However, it should NOT be called when the lhsCurve parameter needs to be preserved.
```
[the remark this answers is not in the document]
johnhableunity: TBH, it's not ideal, and any solution is a bit messy. The optional workaround would be to change the HDRP code to detect the type of the parameter, and if it's a AnimationCurveParameter then call InterpAnimationCurve() directly instead of following the same overloaded Interp() path as everything else. That seemed messier than adding this rule to the animation curve. But if you would prefer to have it that way, I can make the change.

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
```diff
@@ -180,5 +180,66 @@ public void TestSupportedOnAvoidedIfHideInInspector()
...
+
+                Assert.IsTrue(Mathf.Abs(offset) < eps);
+            }
+        }
+
+        [Test]
+        public void RenderInterpolateAnimationCurve()
```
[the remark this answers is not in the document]
johnhableunity: Sure. I'll split them up.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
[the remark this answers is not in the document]
johnhableunity: Actually, thinking it through, that's probably a better way to do it. A lerp that overrides the data in the left side is fundamentally a different operation than blending lhs and rhs to return a new result. Better to make it explicit. Since it's an HDRP-specific exception, does it even make sense to have this functionality inside core?

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
alex-vazquez: I didn't realize that they were two different usages for it :)

com.unity.render-pipelines.core/Runtime/Volume/KeyframeUtility.cs
alex-vazquez: thanks

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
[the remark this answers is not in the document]
RSlysz: Fair point. But to me it seems that it can be usefull in any pipeline to actually interp curves. 

I dig into the code:
A VolumeComponent contains a readonly collection of VolumeParameter that is gathered at construction.
When the VolumeManager Update the volume stack, it start by resetting each one of this VolumeParameter to the one given by default values. Then it calls for each VolumeComponent its Override onto the current stack. This override call the Interp with first argument the one in the stack. 
So I looked at how it is done for a struct like. In FloatParameter, the underlying m_Value is changed. The current call being `stateParam.Interp(stateParam, toParam, interpFactor);` It means that we modify the first argument anyway (because it is also the one at left of the dot method accessor).
So regarding this, I bet that what you did is working in fact. But this comment is missleading. You should say something like "It should NOT be called when the lhsCurve parameter needs to be preserved. But current framework modify it anyway in VolumeComponent.Override for all type of VolumeParameters"

So if you just update this comment and also be sure that data in the scene are not modified, it will be enough in fact.

com.unity.render-pipelines.core/Tests/Editor/Volumes/VolumeComponentTests.cs
```diff
@@ -204,42 +205,101 @@ static private void TestAnimationCurveInterp(AnimationCurve lhsCurve, AnimationC
...
+                testCurve.AddKey(new Keyframe(4.0f, 6.0f, -9.0f, -2.0f));
+                testCurve.AddKey(new Keyframe(7.0f, 5.2f, -3.0f, -4.0f));
+            }
+            else
+            {
+                // Need for the test case where two curves have no overlap
+                testCurve.AddKey(new Keyframe(11.0f, 1.0f, -1.0f, 3.0f));
+                testCurve.AddKey(new Keyframe(14.0f, 6.0f, -9.0f, -2.0f));
+                testCurve.AddKey(new Keyframe(17.0f, 5.2f, -3.0f, -4.0f));
+            }
```
RSlysz, line 246: Note: this could be set in s_AnimationCurveTestPair. You can have like:
```
            new AnimationCurveTestPair
            {
                lhsIndex = 0,
                rhsIndex = 1,
                t = 0.25f,
                testName = "CurveTest 1"
            },
```
should be 
```
            new AnimationCurveTestPair
            {
                lhsCurve = CreateTestCurve(0),
                rhsCurve = CreateTestCurve(1),
                t = 0.25f,
                testName = "CurveTest 1"
            },
```

RSlysz (approved):
Please clarify comment as stated in https://github.com/Unity-Technologies/Graphics/pull/6188/files/1f258dd79e387562e06030c78e59101be66b02a2#r759233222

Beside this it should be mostly ok. Thanks for all your hard work. (pre-approving)

adrien-de-tocqueville (approved):

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
johnhableunity: Closing up the thread on this one. ObjectPools don't seem practical because we have to call Dispose() when each curve is finished per frame. Some of the volumes are in the constant sized stack of 32, and others come from the editor and need to be GC'd normally. That seemed messy enough that we should instead just reuse the LHS, which will always be from the stack of 32.

com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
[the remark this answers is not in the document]
johnhableunity: Sounds good. In my last change, I actually went in and changed the stateParam.Interp() call to check for the type and explicitly call the curve interpolate function when needed, but I'll revert that change and update the comment as you describe.

alex-vazquez (approved):
Thanks

johnhableunity:
> Thanks for all the review/feedback. Now that the PR is ready time to re-trgger Yamato. John as this PR touch both URP and HDRP it required two launch. One for HDRP:
> 
> https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fcurve-blend/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/10271863/job
> 
> And one for URP: https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/HDRP%252Fcurve-blend/.yamato%252Fall-urp.yml%2523PR_URP_trunk/10272009/job
> 
> I triggered them for you, once green we can merge. THanks

Hm, it's failing due to memory allocation. What's the best way to debug this?

sebastienlagarde:
Yamato have same error as on master, merging

PR #6192, opened 2021-11-03
[URP][PostProcessing] Make LensFlare dimensions respect RenderScale setting 

jonuuukas:
---
### Purpose of this PR

---
This PR fixes https://fogbugz.unity3d.com/f/cases/1376820/

pretty simple PR if you check the code - render scale wasn't taken into the screen resolution when executing Lens Flare, so simply multiplying by it fixes the issue

### Testing status
Tested manually on Mac Editor - same behaviour is expected across all platforms, as nothing platform-specific is done here
---
### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -944,15 +944,15 @@ static float GetLensFlareLightAttenuation(Light light, Camera cam, Vector3 wo)
...
         {
+            var camera = cameraData.camera;
             var gpuView = camera.worldToCameraMatrix;
             var gpuNonJitteredProj = GL.GetGPUProjectionMatrix(camera.projectionMatrix, true);
             // Zero out the translation component.
             gpuView.SetColumn(3, new Vector4(0, 0, 0, 1));
             var gpuVP = gpuNonJitteredProj * camera.worldToCameraMatrix;
-
-            LensFlareCommonSRP.DoLensFlareDataDrivenCommon(m_Materials.lensFlareDataDriven, LensFlareCommonSRP.Instance, camera, (float)Screen.width, (float)Screen.height,
+            LensFlareCommonSRP.DoLensFlareDataDrivenCommon(m_Materials.lensFlareDataDriven, LensFlareCommonSRP.Instance, camera, (float)Screen.width * cameraData.renderScale, (float)Screen.height * cameraData.renderScale,
```
gmitrano-unity, line 955: Could we source the scaled screen dimensions from the m_Descriptor variable instead? It seems like several other passes inside this file are already doing this.

```suggestion
            LensFlareCommonSRP.DoLensFlareDataDrivenCommon(m_Materials.lensFlareDataDriven, LensFlareCommonSRP.Instance, camera, (float)m_Descriptor.width, (float)m_Descriptor.height,
```

The dimension scaling logic for the camera descriptor (which is passed in via Setup()) occurs here:
https://github.com/Unity-Technologies/Graphics/blob/aae01d60c1e964e59fe3f3d4e084fce5fd7fd348/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs#L544-L545

gmitrano-unity (changes requested):

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
jonuuukas: @gmitrano-unity - good catch! didn't notice this. updated the PR with your changes (also reverted me passing CameraData instead of Camera)

gmitrano-unity (approved):

nigeljw-unity (approved):

phi-lira (approved):

PR #6193, opened 2021-11-03
Fixed the bad blending between the sun and the clouds (case 1373282).

anisunity:
https://fogbugz.unity3d.com/f/cases/1373282/
This PR improve significantly the integration of the clouds with the sun disk.
image
image
The improved behavior is only applied to cameras (as the sun disk is not rendered in reflection probes, planars or the sky (that would be a double contribution). The behavior is a bit different for MSAA as the solution cannot be applied, this is mentionned in the documentations.

This PR may break some of the tests and will need a screenshot update.

Testing Status
Tested with/without MSAA
Tested the integration with the fog.
Tested with planars, the sky and reflection probes.


sebastienlagarde (approved):

remi-chapelain (changes requested):
Clouds tests fails (I checked locally) which is good thing :p. 
So indeed screenshots needs to be updated for this list:
- 5014_VolumetricCloudsBanding
- 5011_VolumetricClouds

Other than that, the fix has already been pair tested so it shouldn't cause more issues ! ✔️

anisunity:
Launched all the tests to be able to update the screenshots

pierre-unity (changes requested):
The flare size and falloff on the directional light are somehow ignored. However it seems to influence master too.

image

pierre-unity (approved):
Issue above was not linked to this PR. Approving.

sebastienlagarde:
screenshots updated, merging

PR #6194, opened 2021-11-03
Fixed the point distribution for the diffuse denoiser sometimes not being properly intialized.

anisunity:
Due to using the ComputeShader API to dispatch in the init function, sometimes it doesn't behave as expected. This does a runtime initialization that solves the problem.

The repro is quite hard, but @remi-chapelain managed to mimic the repro project.
This should break any test and can't really be covered by our framework.

Testing status
The issue doesn't repro anymore with the fix.


remi-chapelain (approved):
Issue have been discussed with @anisunity

Can't seem to repro the errors on launch in Cathedral scene in RayTracingProject. 
Quickly tested the 3 effects that uses the same denoiser: 
- SSGI ✔️ 
- RTAO ✔️ 
- RTGI ✔️ 
 
DXR Tests are green locally. ✔️

sebastienlagarde (approved):

PR #6196, opened 2021-11-03
Binary APV data with runtime data separated from debug data

tlaedre:
NOTE: This PR depends on native (git) PR#6458 and will stay in draft until this dependency lands.

### Purpose of this PR

This PR addresses a couple of overlapping issues with Probe Volume lighting data storage.

The primary purpose of this PR is to change PV lighting data to use a more performant storage method. The way Unity serialization works, storing large amounts of data (hundreds of MB) in ScriptableObjects adds significant overhead to editor and runtime loading. In a test scene with ~150 MB of baked lighting data, storing the data in ScriptableObjects adds more than 4 seconds of domain reload overhead, whereas storing it in the binary format add no measurable overhead.
(As a more extreme example, having baked data in one of our current production scenes stored as ScriptableObjects adds more than 30 seconds overhead to every domain reload, whereas with the binary method the overhead is negligible).

The second purpose is to separate runtime data from debug data to reduce build size and runtime memory usage. Debug data will be included in builds if the global HDRP settings have the runtime debug shaders checkbox ticked, but otherwise stripped for builds. The Probe Volume debug UI will present options based on which data is available.

Lastly, the probes storage itself is changed in two ways: It no longer always stores three full SH bands but instead stores the format of the actual baked data. The probe coefficients are stored pre-swizzled in the final format they will be uploaded as at runtime (the last bit is up for discussion, it came from a performance investigation that I will open a discussion about separately).

### Testing status

Tested manually in test scenes and our production project.
Tested in editor and player builds.

### Comments to reviewers

Generally speaking, see comments inline in code.

I suppose one of the more controversial changes here might be the decision to change the swizzled format of the probe data. This originated from the fact that runtime loading of cells is quite slow atm and is an overlapping part of ongoing work investigating loading performance (I will bring this as a discussion to Slack before putting further serious work into it). It's easy enough to undo, but at the very least as things stand now, I think it makes sense to store data as close to the runtime usage as possible.


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -63,7 +75,7 @@ partial class ProbeGIBaking
         static bool onAdditionalProbesBakeCompletedCalled = false;
 
         static Dictionary<Vector3Int, int> m_CellPosToIndex = new Dictionary<Vector3Int, int>();
-        static Dictionary<int, Cell> m_BakedCells = new Dictionary<int, Cell>();
+        static Dictionary<int, BakingCell> m_BakedCells = new Dictionary<int, BakingCell>();
```
tlaedre, line 78: BakingCell is now used throughout the entire baking process, only converted to a runtime Cell at the very end of the baking process.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -240,6 +252,7 @@ static void BrickCountInDirections(out Vector3Int cellsInXYZ, float brickSizeInM
 
         // NOTE: This is somewhat hacky and is going to likely be slow (or at least slower than it could).
         // It is only a first iteration of the concept that won't be as impactful on memory as other options.
+#if false // TODO: Make sure we intend to keep this before fixing
```
tlaedre, line 255: This is currently unused, so I didn't update it. Either I fix it or we remove it.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -388,33 +402,32 @@ internal static void PerformDilation()
...
-                                assetCleared.Add(asset.GetSerializedFullPath(), true);
-                                UnityEditor.EditorUtility.SetDirty(asset);
-                            }
-
                             if (cell2Assets[cell.index].Contains(assetPath))
                             {
-                                asset.cells.Add(cell);
+                                if (!assetsCommitted.Contains(assetPath))
+                                {
+                                    WritebackModifiedCellsData(asset);
```
tlaedre, line 422: Because of the way data storage is aliased, dilation now writes directly into the backing storage and we just re-commit that to disk.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -566,26 +580,38 @@ static void OnAdditionalProbesBakeCompleted()
...
                         var profile = probeRefVolume.sceneData.GetProfileForScene(scene);
                         asset.StoreProfileData(profile);
-                        Debug.Assert(profile != null);
                         CellCountInDirections(out asset.minCellPosition, out asset.maxCellPosition, profile.cellSizeInMeters);
                         asset.globalBounds = globalBounds;
                     }
                 }
             }
 
+            // Convert baking cells to runtime cells
```
tlaedre, line 607: We convert all cells of an asset in a single operation since their data is flattened into shared storage.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -608,10 +634,12 @@ static void OnAdditionalProbesBakeCompleted()
                 probeVolume.OnBakeCompleted();
             }
 
-            UnityEditor.AssetDatabase.SaveAssets();
-            UnityEditor.AssetDatabase.Refresh();
+            AssetDatabase.SaveAssets();
+            AssetDatabase.Refresh();
             probeRefVolume.clearAssetsOnVolumeClear = false;
 
+            m_BakingBatch = null;
```
tlaedre, line 641: Clearing this exposes the fact that we still occasionally have multiple callbacks registered for completion (the second one will encounter this null and bail out with a callstack)

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -626,6 +654,154 @@ static void OnLightingDataCleared()
...
+            target[offset + 3] = new Vector4(sh[0, 8], sh[1, 8], sh[2, 8], 1f);
+        }
+
+        static void GetBlobFileNames(string sourceFilename, out string cellDataFilename, out string cellSupportDataFilename )
+        {
+            cellDataFilename = System.IO.Path.ChangeExtension(sourceFilename, "CellData.bytes");
+            cellSupportDataFilename = System.IO.Path.ChangeExtension(sourceFilename, "CellSupportData.bytes");
+        }
+
+        static void WriteBakingCells(ProbeVolumeAsset asset, List<BakingCell> bakingCells)
```
tlaedre, line 678: Flatten all cells into two byte streams: one for runtime data and one for debug data.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -626,6 +654,154 @@ static void OnLightingDataCleared()
...
+                }
+
+                positions.GetSubArray(startCounts.probesCount, cellCounts.probesCount).CopyFrom(bakingCell.probePositions);
+                validity.GetSubArray(startCounts.probesCount, cellCounts.probesCount).CopyFrom(bakingCell.validity);
+                offsets.GetSubArray(startCounts.offsetsCount, cellCounts.offsetsCount).CopyFrom(bakingCell.offsetVectors);
+
+                startCounts.Add(cellCounts);
+            }
+
+            // Need to save here because the forced import below discards the changes.
```
tlaedre, line 740: This feels like a bug to me.. I'll ask around Slack for what would cause this.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeAssetEditor.cs
```diff
@@ -0,0 +1,16 @@
+﻿using UnityEngine.Experimental.Rendering;
+
+namespace UnityEditor.Experimental.Rendering
+{
+    [CustomEditor(typeof(ProbeVolumeAsset))]
+    class ProbeVolumeAssetEditor : Editor
```
tlaedre, line 6: This might be unwanted? I personally really appreciate being able to see the internal stats of the baked data and I don't see the harm in having that visible.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
```diff
@@ -354,13 +356,18 @@ static void SetPixel(DynamicArray<Color> data, int x, int y, int z, int dataLocW
...
 
-        public static void FillDataLocation(ref DataLocation loc, SphericalHarmonicsL2[] shl2, int startIndex, int count, ProbeVolumeSHBands bands)
+        public static void FillDataLocation(ref DataLocation loc, ProbeVolumeSHBands srcBands, NativeArray<Color> shData, int startIndex, int count, ProbeVolumeSHBands dstBands)
         {
+            var probeStride = srcBands == ProbeVolumeSHBands.SphericalHarmonicsL2 ? 7 : 3;
+
+            // Coeff constants that cancel out shader probe decoding
+            var zhalf = new Color(0f, 0f, 0f, 0.5f);
+            var half = new Color(0.5f, 0.5f, 0.5f, 0.5f);
+
```
tlaedre, line 366: So instead of just writing black as default value, these constants cancel out shader decoding of probes data. This means you can load defaults without getting crazy results, or load L1 data into an L2 enabled config and things will still be correctly lit.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
```diff
@@ -354,13 +356,18 @@ static void SetPixel(DynamicArray<Color> data, int x, int y, int z, int dataLocW
             data[index] = value;
         }
 
-        public static void FillDataLocation(ref DataLocation loc, SphericalHarmonicsL2[] shl2, int startIndex, int count, ProbeVolumeSHBands bands)
+        public static void FillDataLocation(ref DataLocation loc, ProbeVolumeSHBands srcBands, NativeArray<Color> shData, int startIndex, int count, ProbeVolumeSHBands dstBands)
```
tlaedre, line 359: Pass in source data bands and desired destination bands separately so we can correctly fill in missing data where applicable.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.Debug.cs
```diff
@@ -170,13 +173,26 @@ void RegisterDebug()
...
 
-            m_DebugItems = widgetList.ToArray();
-            var panel = DebugManager.instance.GetPanel("Probe Volume", true);
-            panel.children.Add(m_DebugItems);
+            if (widgetList.Count > 0)
+            {
+                m_DebugItems = widgetList.ToArray();
+                var panel = DebugManager.instance.GetPanel("Probe Volume", true);
+                panel.children.Add(m_DebugItems);
+            }
```
tlaedre, line 195: Hide the options that don't have the required backing data available.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
```diff
@@ -50,6 +75,86 @@ public string GetSerializedFullPath()
...
+        void OnDisable()
+        {
+            isBlobCreated = false;
+
+            foreach (var bytes in assetToBytes.Values)
+                bytes.Dispose();
+
+            assetToBytes.Clear();
+        }
+        // END TEMP SECTION
```
tlaedre, line 156: So this whole block is basically just temporary scaffolding to be able to test this logic until the mentioned native PR lands allowing direct access to TextAsset data through a non-allocating NativeArray. (TextAsset.bytes makes a copy of the entire data array, which is not great)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -19,50 +12,51 @@ enum ProbeVolumeState
 
     [ExecuteAlways]
     [AddComponentMenu("")] // Hide.
-    internal class ProbeVolumePerSceneData : MonoBehaviour, ISerializationCallbackReceiver
+    public class ProbeVolumePerSceneData : MonoBehaviour, ISerializationCallbackReceiver
```
tlaedre, line 15: This needs to be public so RPs can strip it depending on settings. I've hidden everything else in the class instead, though.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -92,15 +86,18 @@ internal ProbeVolumeAsset GetCurrentStateAsset()
         internal void QueueAssetLoading()
         {
             var refVol = ProbeReferenceVolume.instance;
-            if (assets.ContainsKey(m_CurrentState) && assets[m_CurrentState] != null)
+            if (assets.TryGetValue(m_CurrentState, out var asset) && asset != null)
             {
-                refVol.AddPendingAssetLoading(assets[m_CurrentState]);
+                asset.ResolveCells();
```
tlaedre, line 91: The source data needed to resolve is propagated in OnAfterDeserialize.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -140,5 +137,13 @@ void Update()
                 QueueAssetLoading();
             }
         }
+
+#if UNITY_EDITOR
+        public void StripSupportData()
+        {
+            foreach (var asset in assets.Values)
+                asset.cellSupportDataAsset = null;
```
tlaedre, line 145: Since OnBeforeSerialize reads this is serves as clearing out the use of the support data asset thus omitting it from builds.

com.unity.render-pipelines.high-definition/Editor/Lighting/ProbeVolumeDebugStripper.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            var probeVolumePerSceneDatas = scene.GetRootGameObjects()
+                .Select(go => go.GetComponent<ProbeVolumePerSceneData>())
+                .Where(c => c);
+
+            foreach (var probeVolumePerSceneData in probeVolumePerSceneDatas)
+            {
+                // TODO: Arguably we should just destroy the entire component if 'supportProbeVolumes' is disabled. Otherwise previously baked data
+                //       will still be bundled with players and loaded with the scene even though the entire feature is disabled. Discuss this
+                //       before enabling it, though.
```
tlaedre, line 40: Thoughts?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.Dilate.cs
```diff
@@ -61,35 +62,93 @@ internal void FromSphericalHarmonicsL2(SphericalHarmonicsL2 sh)
...
                 outputProbes = new ComputeBuffer(probeCount, System.Runtime.InteropServices.Marshal.SizeOf<DilatedProbe>());
 
-
                 // Init with pre-dilated SH so we don't need to re-fill from sampled data from texture (that might be less precise).
-                DilatedProbe[] dilatedProbes = new DilatedProbe[probeCount];
+                dilatedProbes = new DilatedProbe[probeCount];
                 for (int i = 0; i < probeCount; ++i)
                 {
-                    dilatedProbes[i].FromSphericalHarmonicsL2(cell.sh[i]);
+                    dilatedProbes[i].FromSphericalHarmonicsShaderConstants(cell.shBands, cell.shData, i);
```
FrancescoC-unity, line 133: The problem I have with this not dilating the whole 3 bands is that this now makes switching between L1 and L2 at runtime not possible. 

Let's assume the data is baked with L1 set as option, then it will be stored as L1 with garbage L2. 
Switching then to L2 will require a rebake as otherwise garbage will be read. 

I see where you are coming from, but I can also this potentially rising issues and annoyances for users.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
FrancescoC-unity: It is going to be used, we just need to discuss the UX for it. 

Also we use it while debugging (it is just a matter of uncommenting out one single line in code and it will pop up in the UI).

We need this to be functional .

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeAssetEditor.cs
FrancescoC-unity: I think it is fine, a bit fringe use case is why it was not made.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.Dilate.cs
FrancescoC-unity: Before I go on and look more at the PR we should discuss this as I think is pivotal for the rest.

FrancescoC-unity:
Will look at this in few minutes, also adding @JulienIgnace-Unity  as this will be likely relevant for streaming

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -477,11 +491,12 @@ static void OnAdditionalProbesBakeCompleted()
 
                 cell.sh = new SphericalHarmonicsL2[numProbes];
                 cell.validity = new float[numProbes];
+                cell.offsetVectors = new Vector3[0]; // pending PR #6169
```
FrancescoC-unity, line 494: Reminder here for me/you to remove this comment when we can

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -566,26 +580,38 @@ static void OnAdditionalProbesBakeCompleted()
...
-                        asset.cells.Add(cell);
+
+                        if (!asset2BakingCells.TryGetValue(asset, out var bakingCellsList))
+                            bakingCellsList = asset2BakingCells[asset] = new();
+
+                        bakingCellsList.Add(cell);
+
                         var profile = probeRefVolume.sceneData.GetProfileForScene(scene);
                         asset.StoreProfileData(profile);
-                        Debug.Assert(profile != null);
```
FrancescoC-unity, line 582: Any reason why removing this assert? Admittedly this should be before the StoreProfileData and might be outdated, but I wonder if there is any specific reason you removed it.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -626,6 +654,154 @@ static void OnLightingDataCleared()
...
+
+                using (var fs = new System.IO.FileStream(cellDataFilename, System.IO.FileMode.Create, System.IO.FileAccess.Write))
+                {
+                    fs.Write(new ReadOnlySpan<byte>(bricks.GetUnsafeReadOnlyPtr(), bricks.Length * UnsafeUtility.SizeOf<Brick>()));
+                    fs.Write(new byte[AlignRemainder16(fs.Position)]);
+                    fs.Write(new ReadOnlySpan<byte>(probes.GetUnsafeReadOnlyPtr(), probes.Length * UnsafeUtility.SizeOf<Vector4>()));
+                }
+                using (var fs = new System.IO.FileStream(cellSupportDataFilename, System.IO.FileMode.Create, System.IO.FileAccess.Write))
+                {
+                    fs.Write(new ReadOnlySpan<byte>(positions.GetUnsafeReadOnlyPtr(), positions.Length * UnsafeUtility.SizeOf<Vector3>()));
```
FrancescoC-unity, line 758: I have a funny feeling both position and validity will be needed as part of the main data soon :P But fair enough for now.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
FrancescoC-unity: Change is good for me, but why zhalf ? It's the 4th channel

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
FrancescoC-unity: I'd add a comment before the function to make sure that the reader of only this function knows that the shData is preswizzled in the format that the 3d texture expects.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
```diff
@@ -377,65 +384,42 @@ public static void FillDataLocation(ref DataLocation loc, SphericalHarmonicsL2[]
                             // We are processing chunks at a time.
                             // So in practice we can go over the number of SH we have in the input list.
                             // We fill with black to avoid copying garbage in the final atlas.
```
FrancescoC-unity, line 386: Comment will need updating a bit to make it more clear that is not black, but the value that will result in black after decoding

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -286,18 +288,33 @@ public partial class ProbeReferenceVolume
...
             public int indexChunkCount;
             public int shChunkCount;
+
+            public ProbeVolumeSHBands shBands;
+
+            public NativeArray<Brick> bricks { get; internal set; }
+
+            // shData is swizzled for direct runtime upload
+            //
+            // L0r0, L0g0, L0b0, L1r0
```
FrancescoC-unity, line 306: It'd be nice to have a very small function to this PR that moves from a probe index to a SphericalHarmonicsL2 (undoing the swizzling) as I am sure it'll be useful for debug and potentially other use cases.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -988,12 +1005,12 @@ void PerformPendingDeletion()
             m_PendingAssetsToBeUnloaded.Clear();
         }
 
-        internal int GetNumberOfBricksAtSubdiv(Cell cell, out Vector3Int minValidLocalIdxAtMaxRes, out Vector3Int sizeOfValidIndicesAtMaxRes)
+        internal int GetNumberOfBricksAtSubdiv(Vector3Int position, int minSubdiv, out Vector3Int minValidLocalIdxAtMaxRes, out Vector3Int sizeOfValidIndicesAtMaxRes)
```
FrancescoC-unity, line 1008: Any reason for not passing the cell here and split the two fields? The cell gives us more info at disposal in case we'd need to change this logic. 

Tbf, don't care much either way, but I wonder if you had a strong reason to

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
```diff
@@ -50,6 +75,86 @@ public string GetSerializedFullPath()
             return m_AssetFullPath;
         }
 
+        internal void ResolveCells()
```
FrancescoC-unity, line 78: As we discussed the stream should be simply float based

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
```diff
@@ -50,6 +75,86 @@ public string GetSerializedFullPath()
...
+                    cell.probePositions = positionsData.GetSubArray(startCounts.probesCount, counts.probesCount);
+                    cell.validity = validityData.GetSubArray(startCounts.probesCount, counts.probesCount);
+                    cell.offsetVectors = offsetsData.GetSubArray(startCounts.offsetsCount, counts.offsetsCount);
+                }
+
+                startCounts.Add(counts);
+            }
+        }
+
+        // TEMP SECTION: Manually extract a NA from bytes until we have access to: https://github.cds.internal.unity3d.com/unity/unity/pull/6458
```
FrancescoC-unity, line 129: Reminder for me/you to remove when we have that PR landed.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
FrancescoC-unity: We then need a minimal doc to make it pass validation on top of the class

com.unity.render-pipelines.high-definition/Editor/Lighting/ProbeVolumeDebugStripper.cs
FrancescoC-unity: Not sure as AFAIK you can potentially swap global settings asset while the player is running (not sure why you would, but it could happen)

FrancescoC-unity (commented):
Added some comments here and there in addition to what we discussed via slack already, let's have another pass when that lands!

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
JulienIgnace-Unity: Yup, in case you didn't know, you can easily check missing API doc by loading the HDRP_Test project and going Window=>Doc Checker

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
tlaedre: zero-half. lazy naming, I'll fix it.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
tlaedre: I removed it because StoreProfileData accesses fields of profile, so it can never trigger, it would always throw a NRE before getting to it.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
tlaedre: Because it is invoked during baking in a location that now only has access to BakingCell before any Cell has been created.

FrancescoC-unity:
Any ETA on when (or if) this can go be updated? 

@adrien-de-tocqueville  is currently working on baking states and we'd need to do some optimization on having just SH data in a separate asset; this PR looks like could provide one way of doing that so but if we do something else might end up conflicting badly with this :-) 

Let us know if you need us to take ownership of the PR if you don't have the time!

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
tlaedre: Now there are three byte streams (the essential data, the optional data - aka L2, and the support data - aka debug) as discussed.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
tlaedre: (did ask, will log a bug for this separately)

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeBrickPool.cs
tlaedre: made these hopefully a bit more meaningful

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
tlaedre: Removed in latest

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.Dilate.cs
tlaedre: As discussed, we currently always store all three bands when baking (i.e. ignore current lighting setting for bake). A new toggle should be introduced for controlling how many bands to store / how many bands to bundle with builds. Loading mixed cells already works.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
tlaedre: These exist:
ProbeGIBaking.WriteToShaderCoeffsL0L1/WriteToShaderCoeffsL2/ReadFromShaderCoeffsL0L1/ReadFromShaderCoeffsL2
DilatedProbe.FromSphericalHarmonicsShaderConstants/ToSphericalHarmonicsShaderConstants

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
tlaedre: Done.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
tlaedre: Done.

com.unity.render-pipelines.high-definition/Editor/Lighting/ProbeVolumeDebugStripper.cs
```diff
@@ -0,0 +1,57 @@
...
+                if (!settings.supportProbeVolumes)
+                {
+                    Debug.Log($"Stripping embedded Probe Volume data from scene '{scene.name}'.");
+                    Object.DestroyImmediate(probeVolumePerSceneData);
+                }
+                else
+#endif
+                if (!settings.supportRuntimeDebugDisplay)
+                {
+                    Debug.Log($"Stripping debug data from Probe Volume in scene '{scene.name}'.");
```
adrien-de-tocqueville, line 51: do we want to keep that log here ?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -626,6 +657,193 @@ static void OnLightingDataCleared()
...
+
+                if (asset.bands == ProbeVolumeSHBands.SphericalHarmonicsL2)
+                {
+                    var probesTargetL2 = probesL2.GetSubArray(startCounts.probesCount * ProbeVolumeAsset.kL2ScalarCoefficientsCount, cellCounts.probesCount * ProbeVolumeAsset.kL2ScalarCoefficientsCount);
+                    for (int j = 0, k = 0; j < cellCounts.probesCount; ++j, k += ProbeVolumeAsset.kL2ScalarCoefficientsCount)
+                    {
+                        ref var sh = ref bakingCell.sh[j];
+                        WriteToShaderCoeffsL2(ref sh, probesTargetL2, k);
+                    }
+                }
```
adrien-de-tocqueville, line 768: Shouldn't the L2 coefs be written regardless of the settings ?

adrien-de-tocqueville (commented):
minor comment, also note that the formatting job is failing

com.unity.render-pipelines.high-definition/Editor/Lighting/ProbeVolumeDebugStripper.cs
tlaedre: I have no strong opinion tbh. I feel like it's somewhat similar to the debug log for shader stripping, in that it's nice to be able to verify in the log that the debug data was indeed stripped.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
tlaedre: I would say no since I don't think we should force ppl to have L2 data. That being said, `asset.bands` is always `SphericalHarmonicsL2` now until we add an explicit option to only bake L0L1 data.

tlaedre:
> minor comment, also note that the formatting job is failing

thanks for pointing that out. I meant to launch the apply formatting job but forgot :(

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
adrien-de-tocqueville: ah thanks i missed that part

com.unity.render-pipelines.high-definition/Editor/Lighting/ProbeVolumeDebugStripper.cs
adrien-de-tocqueville: i guess people assume debug data are stripped so i would remove it, but not very important

adrien-de-tocqueville (approved):

iM0ve (commented):
**Test status:** Done

**Whats tested:**
* Compared baked result of classroom scene Master vs PR
* Checked Dilation still works
* Checked that Virtual Offset still works
* Debug modes still display bricks, cells and probes
* Global and local volumes still work
* Baking from lighting tab and APV tab
* Baking with Spherical Harmonics L1 and L2
* Creating baking sets and using them
* Changing memory budget
* Building player to see APV is there

iM0ve (approved):
I updated my review. I did quite a few tests but as this feature is already used in production I focused on looking for regressions. Have not found any new issues

sebastienlagarde (approved):

sebastienlagarde:
Hey, test are failing for standalone now:
https://yamato-artifactviewer.prd.cds.internal.unity3d.com/0c0a856d-ba3b-437b-bf8f-e8fe946bae14%2Flogs%2FTestProjects%2FHDRP_RuntimeTests%2Ftest-results/TestReport.html

FrancescoC-unity:
Argh yeah this is not compatible with old assets, @tlaedre  can you please rebake that test and update screenshots if needed? 
Let me know if you don't have the time and I can try myself.

tlaedre:
I could, if I had any idea whatsoever how that worked. Is there any docs for how to update test screenshots?

tlaedre:
I guess this update should be part of the PR?

FrancescoC-unity:
Yes, give me 10 min and I can explain in slack

PR #6202, opened 2021-11-04
Fixed an exception happening when the user doesn't set one of the renderers of a given LOD (case 1372965).

anisunity:
https://fogbugz.unity3d.com/f/cases/1372965/
Before 2021.2 No LOD support, so no need for the fix.
After 2021.2, C++ culling, no need for the fix.

Testing status
Pair tested on the repro case with @remi-chapelain


remi-chapelain (approved):
Pair tested and already tested in[ this PR ](https://github.com/Unity-Technologies/Graphics/pull/6016)anyway !

sebastienlagarde (approved):

PR #6203, opened 2021-11-04
Bug fixes for reflection probes normalization

tlaedre:
### Purpose of this PR

Fixes NaNs in lighting when probe positions are invalid or when probes extraction fails.
Fixed missing probe requests after domain reload.

How to repro NaNs:
Make sure probe volumes and reflection probe normalization is enabled in the project. Place a reflection probe inside geometry. Bake it. Lighting is now broken with nans propagating out from the reflection probe.

How to repro missing requests:
Add a new reflection probe. Do not bake. Force a domain reload (code compilation, e.g.). Bake. Observe how the new probe does not have valid reflection normalization available / applied.

(Need to run full probe volume bakes for this to work, not just baking single reflection probes)

### Testing status

Tested in editor HDRP with PV and reflection normalization enabled.
Only affects baked probes.


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
```diff
@@ -121,11 +128,10 @@ private bool ComputeRequestIsFree(Vector3 capturePosition)
         /// <returns>Whether the request for light probe rendering has been fulfilled and sh is valid.</returns>
         public bool RetrieveProbeSH(int requestID, out SphericalHarmonicsL2 sh)
         {
-            if (requestID >= 0 && requestID < m_SHCoefficients.Count
-                && ComputeCapturePositionIsValid(m_RequestPositions[requestID]))
+            if (requestID >= 0 && requestID < m_SHCoefficients.Count && ComputeCapturePositionIsValid(m_RequestPositions[requestID]))
             {
                 sh = m_SHCoefficients[requestID];
-                return true;
+                return m_SHValidity[requestID] == kValidSH;
```
tlaedre, line 134: Not 100% certain if we should require full validity or also allow a threshold of error here.

sebastienlagarde:
FYI there is also this PR which do a fix in probe normalization, bringing here for awareness: https://github.com/Unity-Technologies/Graphics/pull/6140

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
FrancescoC-unity: It really depends on the values returned by the lightmapper on non-completely invalid probe locations; I'd say for now we can force it to be required to be fully valid.

FrancescoC-unity:
> FYI there is also this PR which do a fix in probe normalization, bringing here for awareness: #6140

This is a different issue, it is for when reflection probes are inside or partially inside geometry.

FrancescoC-unity (approved):

JulienIgnace-Unity (approved):

iM0ve (changes requested):
**Test status:** Done
**What's tested:**
* Verified the fixes - reproduced the issue and checked that its no longer occurring with this PR
* Regression test for APV baking process - opened the ClassRoom scene and rebaked lighting to see that APV and reflection probes bake as expected
* Regression test for non-APV features - Opened HDRP template and Amalienborg to check if non-APV probes had been affected.
* Functional testing - Added new reflection probes, baked individually and via lighting panel. Also checked Realtime on enable ones.

**Issues:**
1) **FIXED** I can still reproduce the 1st one with nans. 
https://user-images.githubusercontent.com/36502659/141264276-91712911-8a54-4fc1-abde-9190220e2cab.mp4

iM0ve (approved):
Updated my review

PR #6206, opened 2021-11-04
Add probe volume influence weight parameter

tlaedre:
### Purpose of this PR

Having a PV weight parameter allows fading probe volumes influence down to ambient probe fallback. While no blending or runtime composition exists in the system, this allows for solving some cases of probe set switcharoo by carefully fading to ambient probe, swapping probe volumes data, and fading back into probe volumes.

### Testing status

Tested in current production.
Tested in test scenes.

### DIFF
--- a/com.unity.render-pipelines.core/CHANGELOG.md
+++ b/com.unity.render-pipelines.core/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Context menu on Volume Parameters to restore them to their default values.
 - Linear version of function that sets FSR RCAS shader constants
+- Add probe volume influence weight parameter
 
 ### Fixed
 - Fixed XR support in CoreUtils.DrawFullscreen function.


com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
```diff
@@ -357,6 +363,12 @@ void EvaluateAdaptiveProbeVolume(in float3 posWS, in float3 normalWS, in float3
         bakeDiffuseLighting += apvSample.L0;
         backBakeDiffuseLighting += apvSample.L0;
         lightingInReflDir += apvSample.L0;
+
+        if (_Weight < 1.f)
```
sebastienlagarde, line 367: there is no need for this test here. don't think we save anything with it

tlaedre: the train of thought was that since weight is likely to be 1 99% of the time we could branch if away cheaply based on the uniform here. (I'll admit I haven't actually checked the asm for this assumption yet)

FrancescoC-unity:
I wonder if this should be in the probe volume options at all, is this supposed to add a different fade spatially? 
If it is a global fade I think it'd work better as an API in ref volume.

tlaedre:
> I wonder if this should be in the probe volume options at all, is this supposed to add a different fade spatially? If it is a global fade I think it'd work better as an API in ref volume.

It's a global fade only. The reason I put it into the options is so that you could use volume overrides / blending to drive it, but maybe that's not a valid approach for things that can only work from global volumes? Should I change it to an API instead? I guess a property would be the most appropriate option then.

FrancescoC-unity:
> > I wonder if this should be in the probe volume options at all, is this supposed to add a different fade spatially? If it is a global fade I think it'd work better as an API in ref volume.
> 
> It's a global fade only. The reason I put it into the options is so that you could use volume overrides / blending to drive it, but maybe that's not a valid approach for things that can only work from global volumes? Should I change it to an API instead? I guess a property would be the most appropriate option then.

Yes, I'd prefer if we move to property on the ref volume

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
JulienIgnace-Unity: The branch is static so I guess we do skip the ALUs but I don't think we skip the loads caused by EvaluateAmbientProbe so we'd be adding potential cost all the time for a fringe case here.
Maybe it's ok since we do call EvaluateAmbientProbe in another branch anyway. Might need to look a bit more at the generated code to make sure of this.

iM0ve:
As I understand there will be changes to this PR, so I should wait a little before testing?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolume.hlsl
tlaedre: @JulienIgnace-Unity why would the loads not be skipped by the branch? at least looking at the assembly it looks like they all get gated behind branches.

tlaedre:
> As I understand there will be changes to this PR, so I should wait a little before testing?

Updated @iM0ve 

Possible testing code now that it's no longer a volume override

```
using UnityEngine;

public class PVWeight : MonoBehaviour
{
    public float weight = 1;

    void OnValidate()
    {
        if(ProbeReferenceVolume.instance != null)
            ProbeReferenceVolume.instance.probeVolumesWeight = weight;
    }
}
```

tlaedre:
> > > I wonder if this should be in the probe volume options at all, is this supposed to add a different fade spatially? If it is a global fade I think it'd work better as an API in ref volume.
> > 
> > 
> > It's a global fade only. The reason I put it into the options is so that you could use volume overrides / blending to drive it, but maybe that's not a valid approach for things that can only work from global volumes? Should I change it to an API instead? I guess a property would be the most appropriate option then.
> 
> Yes, I'd prefer if we move to property on the ref volume

Moved @FrancescoC-unity

FrancescoC-unity (approved):

FrancescoC-unity:
Note @tlaedre  I will merge master and make a small change to the cbuffer as it behaves funny on OSX.

sebastienlagarde (approved):

PR #6207, opened 2021-11-04
[HDRP] Merge FSR RCAS Logic into FinalPass

gmitrano-unity:
### Purpose of this PR
This change merges the FSR RCAS logic into the FinalPost shader to avoid
an unnecessary round trip through memory.

This saves about 200us of frame time (compared to the prior implementation) at 4k on PS5 (Spaceship Demo) with scaling forced to 77%.

---
### Testing status
Manual testing locally on PC DX12 & DX11, HDRP DRS tests on DX12 & DX11, PS5 Spaceship Demo with scaling

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
@@ -41,6 +41,8 @@ Shader "Hidden/HDRP/FinalPass"
         #define FSR_INPUT_TEXTURE _InputTexture
         #define FSR_INPUT_SAMPLER s_linear_clamp_sampler
         #if ENABLE_ALPHA
+            // When alpha is in use, activate the alpha-passthrough mode in the RCAS implementation.
+            // When this mode is active, ApplyRCAS returns a four component vector (rgba) instead of a three component vector (rgb).
             #define FSR_ENABLE_ALPHA 1
         #endif
         #include "Packages/com.unity.render-pipelines.core/Runtime/PostProcessing/Shaders/FSRCommon.hlsl"


sebastienlagarde (approved):
Looks good to me, please add a changelog

sebastienlagarde:
@kecho btw we haven't re-enable the dynamic res graphic test right? so we don't have test allowing to check the change?

kecho:
@sebastienlagarde we have only disabled the DRS vulkan tests from what I see
image

We were waiting on trunk to fix this. I think the fix landed. Otherwise, dx12, dx11 and metal should be tested properly.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
```diff
@@ -95,15 +102,18 @@ Shader "Hidden/HDRP/FinalPass"
...
+            uint2 scaledPositionSS = ((input.texcoord.xy * _UVTransform.xy) + _UVTransform.zw) * _ViewPortSize.xy;
 
             // Flip logic
             positionSS = positionSS * _UVTransform.xy + _UVTransform.zw * (_ScreenSize.xy - 1.0);
             positionNDC = positionNDC * _UVTransform.xy + _UVTransform.zw;
 
             #ifdef CATMULL_ROM_4
             CTYPE outColor = UpscaledResult(positionNDC.xy);
+            #elif defined(RCAS)
+            CTYPE outColor = ApplyRCAS(scaledPositionSS);
```
kecho, line 114: ApplyRCAS returns a float3, yet you have a CTYPE. This means that you could potentially be setting alpha 0 and causing a shader compiler warning (float3 casted into float4).

Notice how the other passes use the CTYPE_SWIZZE to address this. Can you dig in a bit into support for alpha? I think this might break this pass.

kecho (changes requested):

kecho (approved):

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
gmitrano-unity: I added a new commit that adds some comments explaining how this is handled:

https://github.com/Unity-Technologies/Graphics/blob/d2d314354490699d5f0f67e43fd301026869c394/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader#L41-L48

Any idea if this functionality tested in the DRS tests?

PR #6212, opened 2021-11-04
Remove redundant probe volume inits

tlaedre:
### Purpose of this PR

The main purpose of this PR is to reduce domain reload / HDRP initialization overhead when Probe Volumes are used.

The change avoids double (or triple) Probe Volumes initialization. This initial initialization will always be re-created at least once. On my laptop, running high memory budget, each extra init adds about 800ms of startup time.

If the initial scene loaded contains ProbeVolumePerSceneData, then the required setup parameters are technically available at init time, but because of the index rebuild flag (set both in Initialize and from AddPendingAssetLoading) the data gets recreated anyway.

If the initial scene does not contain ProbeVolumePerSceneData, InitProbeReferenceVolume will run three times; first from Initialize, then from PerformPendingIndexChangeAndInit, and lastly when any ProbeVolumePerSceneData is actually added (because position/index will change again).

### Testing status
Tested editor, playmode and standalone.


FrancescoC-unity (approved):

JulienIgnace-Unity (approved):

sebastienlagarde:
Note: this PR will be reverted as it introduce an issue

PR #6220, opened 2021-11-05
Fixing crash when creating area light.

kecho:
### Purpose of this PR
Fixing crash when creating area light, caused by missing validity checks on the light entity.
https://fogbugz.unity3d.com/f/cases/1378494/

---
### Testing status
Tested in an empty scene, added a new area light.


sebastienlagarde (approved):

remi-chapelain (approved):
Tested creating multiple types of lights exclusively and with other type of lights !
That seem to fix the issue, all lights are included in the cluster up to the max cluster limit.

PR #6221, opened 2021-11-05
[Fogbugz # 1388541] Re enable vulkan DRS tests

kecho:
---
### Purpose of this PR
We disabled vulkan drs tests because they were busted from the trunk. Aparently its working now. All we need is a green on yamato vulkan and we can merge this :) 

Fogbugz: https://fogbugz.unity3d.com/f/cases/1388541/


kecho:
Tests failing again. Super weird as to why. Will have to investigate.

kecho:
@sebastienlagarde Once we pass the vulkan DRS test in yamato this PR is ready to go. I verified manually as well, hardware DRS is fixed now :)

PR #6222, opened 2021-11-05
Fix immutable assets issue

theopnv:
### Purpose of this PR

From current srp2core PR tests ([link](https://bokken-artifacts-ui.bf.unity3d.com/katana-production/proj0-Test_EditorTests_-_Windows64Editor___WindowsStandaloneSupport/44153430_05_11_2021_10_37_16_+0000/build/ReportedArtifacts/TestReport.html)):
```
The package cache was invalidated and rebuilt because the following immutable asset(s) were unexpectedly altered:
  Packages/com.unity.render-pipelines.universal/Runtime/Materials/Decal.mat
```

This PR is regenerating the material to fix the issue.

---
### Testing status
Rerunning the failing builder using these changes in the srp2core branch.


theo-pnv:
Tested as part of the srp2core PR: the builder that was failing now passes with that update.

phi-lira (approved):

PR #6229, opened 2021-11-05
[URP] Port HDRP Mip Bias Logic to URP

gmitrano-unity:
### Purpose of this PR
This change integrates the mip bias logic from HDRP into URP. This logic
ensures that shaders select their mips based on the final screen
resolution rather than the current render resolution. In cases where
aggressive upscaling is taking place, this can significantly improve
texture detail in the final image.

50% Scaling Before:
image

50% Scaling After:
image

(Best compared in separate browser tabs since the previews here are scaled)

---
### Testing status
Tested before/after on Windows and Android
No clear performance difference on mobile (Adreno 650)

---
### Comments to reviewers
This is pretty much identical to how HDRP handles the mip bias logic with the exception of it not being behind an option in URP. I removed the option since it would have created more shader permutations (as far as I know) and it doesn't seem to negatively impact performance on the mobile hardware I tested. This might not be true for lower end devices.

---
### Notes for QA
The changes in this PR will be verified for quality as part of the larger FSR PR here: https://github.com/Unity-Technologies/Graphics/pull/6262

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
@@ -224,8 +224,7 @@ namespace UnityEngine.Rendering.Universal
             // Calculate a bias value which corrects the mip lod selection logic when image scaling is active.
             // We clamp this value to 0.0 or less to make sure we don't end up reducing image detail in the downsampling case.
             float mipBias = Math.Min((float)-Math.Log(cameraWidth / scaledCameraWidth, 2.0f), 0.0f);
-            cmd.SetGlobalFloat(ShaderPropertyId.globalMipBias, mipBias);
-            cmd.SetGlobalFloat(ShaderPropertyId.globalMipBiasPow2, Mathf.Pow(2.0f, mipBias));
+            cmd.SetGlobalVector(ShaderPropertyId.globalMipBias, new Vector2(mipBias, Mathf.Pow(2.0f, mipBias)));
 
             //Set per camera matrices.
             SetCameraMatrices(cmd, ref cameraData, true);
--- a/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
@@ -221,6 +221,12 @@ namespace UnityEngine.Rendering.Universal
 
             cmd.SetGlobalVector(ShaderPropertyId.screenSize, new Vector4(cameraWidth, cameraHeight, 1.0f / cameraWidth, 1.0f / cameraHeight));
 
+            // Calculate a bias value which corrects the mip lod selection logic when image scaling is active.
+            // We clamp this value to 0.0 or less to make sure we don't end up reducing image detail in the downsampling case.
+            float mipBias = Math.Min((float)-Math.Log(cameraWidth / scaledCameraWidth, 2.0f), 0.0f);
+            cmd.SetGlobalFloat(ShaderPropertyId.globalMipBias, mipBias);
+            cmd.SetGlobalFloat(ShaderPropertyId.globalMipBiasPow2, Mathf.Pow(2.0f, mipBias));
+
             //Set per camera matrices.
             SetCameraMatrices(cmd, ref cameraData, true);
         }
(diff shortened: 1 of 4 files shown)


manuele-bonanno:
@Unity-Technologies/gfx-qa-urp can we check for any mobile regressions? Especially important on low-end, since this would be enabled by default and no option to disable it

manuele-bonanno (approved):
LGTM!
I am requesting a mobile QA pass to make sure low end devices are fine

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -221,6 +221,12 @@ void SetPerCameraShaderVariables(CommandBuffer cmd, ref CameraData cameraData)
 
             cmd.SetGlobalVector(ShaderPropertyId.screenSize, new Vector4(cameraWidth, cameraHeight, 1.0f / cameraWidth, 1.0f / cameraHeight));
 
+            // Calculate a bias value which corrects the mip lod selection logic when image scaling is active.
+            // We clamp this value to 0.0 or less to make sure we don't end up reducing image detail in the downsampling case.
+            float mipBias = Math.Min((float)-Math.Log(cameraWidth / scaledCameraWidth, 2.0f), 0.0f);
+            cmd.SetGlobalFloat(ShaderPropertyId.globalMipBias, mipBias);
+            cmd.SetGlobalFloat(ShaderPropertyId.globalMipBiasPow2, Mathf.Pow(2.0f, mipBias));
```
phi-lira, line 228: I'm curious about if it's worth to bake this into a single vector property. There's still some mid range mobile GPUs out there that are vector based. On the other hand everything modern is scalar based and I guess making vector adds more VGPR.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -103,6 +103,77 @@
...
 #endif
 
+///
+/// Texture Sampling Macro Overrides for Scaling
+///
+/// When mip bias is supported by the underlying platform, the following section redefines all 2d texturing operations to support a global mip bias feature.
+/// This feature is used to improve rendering quality when image scaling is active. It achieves this by adding a bias value to the standard mip lod calculation
+/// which allows us to select the mip level based on the final image resolution rather than the current rendering resolution.
+
+#ifdef PLATFORM_SAMPLE_TEXTURE2D_BIAS
```
phi-lira, line 113: I wonder why we can't move these two Core RP. Both pipelines support _GlobalMipBias and _GlobalMipBiasPow2 right?

phi-lira (approved):
LGTM. It's kind of unfortunate that we don't yet have a way for pipeline to share uniforms / common data setup, all of this PR is duplicated code in both HDRP and URP.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
gmitrano-unity: I did investigate moving this into core a bit, but ultimately we have two choices that I could think of. We can slip it into Common.hlsl and let it inject itself into basically all shaders and gate the override by a define like HDRP does, or we could add a new core hlsl file that's optionally included and HDRP and URP can include it in different spots but it would be opt-in. The problem I see with both of these is the fact that the setup of the value itself from the CPU would be totally different from either SRP variant and if we did the auto-enable version, custom SRPs could forget to set the value entirely.

I think the ideal solution in the long term is probably to revisit our approach for mip bias and to find a solution that isolates material texel fetches rather than using a global override like this. The current implementation affects sampling in post process shaders and any other shaders as well since it's just overriding the main sampling macros. If we could find a way to isolate the bias logic to just material fetches with a new macro like "SAMPLE_MATERIAL_2D(x)", I think it'd make things a lot safer overall. Maybe we could also include the CPU-side constant setup inside Core as part of the dynamic resolution scaling logic?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
gmitrano-unity: I don't think it'll hurt things on modern GPUs so I just updated the diff to combine the constants into a single vector.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
phi-lira: Yes, that makes sense if we plan to revisit this logic then it's best to do it before moving to core. 

> Maybe we could also include the CPU-side constant setup inside Core as part of the dynamic resolution scaling logic?

Yes this makes sense to me. @sandy-carter-unity for awareness.

phi-lira:
PR looks good to me, the failure in VFX_URP is already fixed in another PR in the staging branch. Merging to staging.

PR #6232, opened 2021-11-08
Edited the Depth Priming MVP docs.

oleks-k:
Edited the Depth Priming MVP docs.


GraphineNicolas (changes requested):
"The performance improvement depends on the amount of overlapping pixels" should mention that this only applies to the opaque pass. Something like this:
"The performance improvement depends on the amount of overlapping pixels in the opaque pass"

manuele-bonanno (approved):

oleks-k:
> "The performance improvement depends on the amount of overlapping pixels" should mention that this only applies to the opaque pass. Something like this: "The performance improvement depends on the amount of overlapping pixels in the opaque pass"

Thanks, I've added that phrase.

PR #6233, opened 2021-11-08
Backport of #6232: Edited the Depth Priming MVP docs.

oleks-k:
Backport of #6232: Edited the Depth Priming MVP docs.


NicoLeyman (changes requested):
Duplicating my feedback from the master PR since I managed to comment on the wrong PR and with the wrong account:
"The performance improvement depends on the amount of overlapping pixels" should mention that this only applies to the opaque pass. Something like this:
"The performance improvement depends on the amount of overlapping pixels in the opaque pass"

aleksandrasdzikia (approved):

oleks-k:
> Duplicating my feedback from the master PR since I managed to comment on the wrong PR and with the wrong account: "The performance improvement depends on the amount of overlapping pixels" should mention that this only applies to the opaque pass. Something like this: "The performance improvement depends on the amount of overlapping pixels in the opaque pass"

Thanks, this is done.

NicoLeyman (approved):

PR #6239, opened 2021-11-08
[2022.1][URP] Decal fixes

lukaschod:
https://github.com/Unity-Technologies/Graphics/pull/5961

Purpose of this PR
Fix cases:

Fixed decals to work with native render pass [case 1353141]
Fixing decals rendering to be correct with render scale [1353885]
Depends on #5369

Testing status


phi-lira (approved):
LGTM. 
 
 - Failures are related to 035 terrain test being fixed in another PR.
 - Test 058 failure is due to reference image being wrong, already fixed on master.

PR #6240, opened 2021-11-08
Universal/swapbuffer 2d perf

robin-demoor:
### Checklist for PR maker
- [N/A] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [ N/A] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Why is this PR needed, what hard problem is it solving/fixing? There was a performance regression in the 2D renderer. Due to a faulty "finishPostProcessOnScreen" check there was an extra quad being drawn in some setups. 
---
### Testing status
Lost crypt performance project was used to verify this fix. 
https://unity-ci.cds.internal.unity3d.com/job/9660230


unity-cchu (approved):

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -538,7 +538,7 @@ void Swap(ref ScriptableRenderer r)
                 }
 
                 // With camera stacking we not always resolve post to final screen as we might run post-processing in the middle of the stack.
-                bool finishPostProcessOnScreen = m_ResolveToScreen;
+                bool finishPostProcessOnScreen = m_UseSwapBuffer ? m_ResolveToScreen : cameraData.resolveFinalTarget || (m_Destination == cameraTargetHandle || m_HasFinalPass == true);
```
phi-lira, line 541: What is the performance issue? There's this section of code that mentions a TODO on 2D renderer that it will avoid a ping pong once 2D renderer supports swap buffer, so insn't the proper fix to add swap buffer support for 2D renderer? https://github.com/Unity-Technologies/Graphics/blob/30830663574703625f8006f48b0c2fc7d599db54/com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs#L560-L573
 
 This approach here makes the logic more convoluted.

phi-lira: Also just a few lines above this line you are changing there's already a check for m_UseSwapBuffer that could initialize the finish post on screen variable. Adding another check for m_UseSwapBuffer makes the logic more convoluted for now reason imo.

phi-lira (changes requested):

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
robin-demoor: I will change it so that it is all set in one check because that will make it more clear indeed. The performance regression was there because the logic was incorrect and sometimes did another blit while the result was already in the correct RenderTarget. Hence doing the "ping pong" effect without it being needed. 

I'll have to discuss with the team on whether I can actually implement it for the 2D renderer right now as that will take some extra time and mostly a lot of QA to make sure everything still works.

phi-lira (approved):

PR #6241, opened 2021-11-08
Temporarily remove HDLightUtils_IESProfileAPI()

arttu-peltonen:
The test fails on Katana (assumes it's being run under HDRP_Tests project) and blocks srp2core.


arttu-peltonen:
This commit was cherrypicked by @theo-pnv to the srp2core batch. I'm merging this to master now to avoid the same issue in future srp2core batches, if the problem isn't fixed by then. @adrien-de-tocqueville please remember to re-enable the test at some point :)

PR #6243, opened 2021-11-08
Add universal and hdrp to smoke project testables 

theopnv:
### Purpose of this PR

In order to catch more (currently uncaught) issues, this PR is adding the render pipelines to the smoke test project's testables
More context there: https://jira.unity3d.com/browse/ANT-299

---
### Testing status
Smoke tests + safety net: https://unity-ci.cds.internal.unity3d.com/project/902/branch/ant%2Fadd-render-pipelines-to-smoke-project-testables/jobDefinition/.yamato%2Fall-smoke_test.yml%23Smoke_Test_trunk

### DIFF
--- a/TestProjects/SRP_SmokeTest/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/SampleScene.png
+++ b/TestProjects/SRP_SmokeTest/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/None/SampleScene.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:142fdb99a6bde81789d56ee49f0a6c4a8a2363336206298adf32aef53a0e61b5
-size 4409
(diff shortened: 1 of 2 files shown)


arttu-peltonen:
I checked this locally and I don't think it's solving the issue discussed on slack. Here's what the test results look like when running the SRP_Smoketest locally:

image

So URP tests fail as expected, but HDRP tests pass. This is because SRP_SmokeTest is configured to have HDRP as active pipeline: 

image

To replicate the same failures we had on Katana, we would need a test project that is empty, and doesn't have a Scriptable Render Pipeline Asset configured here.

theo-pnv:
Only the editmode job is failing, which is what we want: now we have the same behavior than on Katana.
When @arttu-peltonen 's PR https://github.com/Unity-Technologies/Graphics/pull/6247 is merged the failing tests will be skipped if no render pipeline is set. So the editmode job will be green.

arttu-peltonen (approved):

PR #6246, opened 2021-11-08
[Master] Fixing "RenderingCommandBuffer: invalid pass index" errors with SSAO (Case 1374215)

ellioman:
# Purpose of this PR
[Fixing issue 1374215](https://issuetracker.unity3d.com/issues/urp-renderingcommandbuffer-invalid-pass-index-errors-when-opening-urp-sample-scenes)

When opening up a project with a clean library, the SSAO shader would throw an error if Async Shader Compilation was enabled. This was happening because the Renderer Feature is trying to draw while shader was compiling.

The solution is to disable asynchronous shader compilation for the SSAO shader.


gmitrano-unity:
I've seen similar errors in the console when turning on SMAA for the first time in a project while the game view is open. Is this happening because the "stand-in" loading material only has a pass 0? If that's the case, I'd expect all draws with the loading material which target passes other than pass 0 to fail. Should the engine just disable async editor compilation for any shader with multiple passes?

Jonasmortensen (approved):
Tested on package sample import and issue is fixed 👍

ellioman:
@gmitrano-unity 

> I've seen similar errors in the console when turning on SMAA for the first time in a project while the game view is open.
I tried reproducing that error but was unable to. If you can give me a repro, please let me know.

> Is this happening because the "stand-in" loading material only has a pass 0? 
Yes, the loading shader has only one pass and SSAO was trying to call the 4 passes it needs while compiling. 

> If that's the case, I'd expect all draws with the loading material which target passes other than pass 0 to fail. Should the engine just disable async editor compilation for any shader with multiple passes?
I don't think such a blanket thing would work. But we need to be careful with features that have multiple shader passes that need compiling because of variants etc. We're already doing this with the Blit shaders and the ShadowMidtonesHighlights.

phi-lira (approved):

phi-lira:
Approving this PR, without much context this seems a problem that might bite us back in the future. It seems something that could be improved on the asynchronous shader compilation.

PR #6253, opened 2021-11-09
Focus distance in path traced depth of field now takes into account the focus mode setting

pmavridis:
### Purpose of this PR
The path traced depth of field was not taking into account the focus mode setting (volume vs camera).

---

### Testing status
I have added a new graphics test to test the new camera mode.
Also renamed the old graphics scene to make clear that we are testing the volume mode (as suggested by Remi).


eturquin (approved):

sebastienlagarde (approved):

remi-chapelain (approved):
Pair tested ✔️

PR #6256, opened 2021-11-09
Reenable HDLightUtils test

adrien-de-tocqueville:
### Purpose of this PR

fix test that was disabled by #6241 cause it was relying on an asset in HDRP_Tests project.
Changed it to generate an ies profile from script.

### Test status

green in local
launching yamato

@arttu-peltonen @theo-pnv so apparently katana doesn't launch test in the same context as yamato. It should be fine now since i am creating assets, and not relying on assets from a project, but is there a way I can test in the katana context, or we'll see for next srp2core PR ?


arttu-peltonen (approved):
LGTM

sebastienlagarde (approved):

PR #6259, opened 2021-11-09
Fix hightligher for volumetric clouds

adrien-de-tocqueville:
### Purpose of this PR

really minor fixes for highlighter
volumetric clouds defined a useless field in expandables, addin @anisunity in case but i guess it was a mistake.
Also fix when an HDRP asset is not assigned in the quality settings

---
### Testing status
tested in the editor


anisunity (approved):
lgtm

sebastienlagarde (approved):

PR #6260, opened 2021-11-09
Make USE_DRAW_PROCEDURAL variation only for vertex programs in UberPost shader

tomzig16:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1377405/
UberPost has a lot of fragment shader variants that use _USE_DRAW_PROCEDURAL. In fact, removing this keyword (again, only in fragment shaders) with the URP template project reduced its number of variants from 1279 to 639 in 2020.3.18. The change reduces the shader's memory usage from ~12MB to ~6MB. So this PR simply makes _USE_DRAW_PROCEDURAL compile only for vertex shaders by changing its declaration in  UberPost shader from `#pragma multi_compile` to `#pragma multi_compile_vertex`.

---
### Testing status
Manually looked if results are visually correct and affected (before and after change) in  URP Post Processing project + editor
Yamato tests on Post Processing and XR

---
### Comments to reviewers
OSX Metal test fails on master branch with 137 error code.


com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
```diff
@@ -10,7 +10,7 @@ Shader "Hidden/Universal Render Pipeline/UberPost"
         #pragma multi_compile_local_fragment _ _DITHERING
         #pragma multi_compile_local_fragment _ _LINEAR_TO_SRGB_CONVERSION
         #pragma multi_compile_local_fragment _ _USE_FAST_SRGB_LINEAR_CONVERSION
-        #pragma multi_compile _ _USE_DRAW_PROCEDURAL
+        #pragma multi_compile_vertex _ _USE_DRAW_PROCEDURAL
```
phi-lira, line 13: Seems like there are other shaders that could benefit from this same improvement. F.ex FinalPost, SMAA, Bloom etc. 

I recommend checking them.

tomzig16: Addressed in ac3031c

phi-lira (approved):

PR #6262, opened 2021-11-09
[URP] FSR Upscaling Filter

gmitrano-unity:
### Purpose of this PR
Add FSR Upscaling Filter to URP

This change adds a new upscaling filter to URP that uses FidelityFX
Super Resolution. It is only available on platforms that support
shader target 4.5 or above. On unsupported platforms, it will behave
as if the user selected the "Auto" filter in the pipeline asset.

Unlike other filters, this filter will execute even when the renderer
is not configured to scale. HDRP has the same behavior. This allows the
FSR shaders to improve visual quality even when scaling is not active.

77% Scaling with Bilinear Filter:
image

77% Scaling with FSR Filter:
image

100% Scaling (Native):
image

All images were rendered with 1080p as their final resolution and include the mip bias changes currently out for review in a different PR.

---
### Testing status
Tested on Windows with Vulkan, Android with Vulkan. Still needs XR testing and more comprehensive testing in general.

---
### Comments to reviewers
This PR currently contains commits from two other outstanding PRs. This PR should not be merged until those commits are merged into master.

---
### Notes for QA
Please test each filter type at 0.5 resolution scale, and 0.77 resolution scale with FXAA on and off (AA should either be FXAA or MSAA). A fixed viewport size should be used for this (such as 1920x1080) to make sure the integer scaling path is triggered during Auto.

---
### TODO
- [x] Changelog updates
- [ ] Documentation updates
- [x] Automated graphics tests
- [x] Re-enable 16-bit mode before merging (Fix black-screen issue in RCAS)


LaserYGD:
Will there be a sharpening value parameter exposed to users?

m0nsky:
Hi @gmitrano-unity, great work, I hope it's alright to leave some feedback. I've had good results trying this branch on desktop. I've also tried this branch in XR since you mentioned it needs further testing, and I am mainly interested in using this tech in XR.

Currently it seems to result in a black screen in XR (using both mock hmd and on the actual device itself)

Example:

urp fsr xr

I've also set up a plug & play repository with everything pre-configured to hopefully make things easier. Keep in mind I have done two slight namespace changes to the URP package for 2021.2 compatibility, but these are unrelated to FSR.

Repository:

https://github.com/m0nsky/urp_fsr_xr

1. Open Scenes/SampleScene
2. Hit play
3. Open Settings/UniversalRP-LowQuality and set the Upscaling Filter to FidelityFX Super Resolution 1.0

gmitrano-unity:
> Will there be a sharpening value parameter exposed to users?

That's a good question, I'll ask some folks internally and see what the plan is.

gmitrano-unity:
@m0nsky Feedback is much appreciated! :)

I just pushed a new commit that disables the 16-bit FSR implementation since I've seen it cause similar black-screen results on some hardware/software configs. Could you please let me know which GPU hardware and graphics API you were using when you saw the black screen and if the most recent commit fixes your issue? I've seen the issue on Nvidia + DX11, but only when the project target is Android.

gmitrano-unity:
@m0nsky Nevermind, I just tried it out in the Mock HMD locally and the 16-bit change doesn't fix the issue. I'll investigate this. Thanks!

gmitrano-unity:
@m0nsky I just tested my latest commit in your example xr project and it seems to fix the black screen issue for me. I also tried in Viking Village with the XR mock HMD enabled and that now works as well.

m0nsky:
So, the black screen fix (latest) worked for Mock HMD in the editor initially. I tried building for Oculus Quest and got some errors regarding FXAA (probably related to your other PR [#6178](https://github.com/Unity-Technologies/Graphics/pull/6178)).

fxaa err

I temporarily worked around this by commenting out the ApplyFXAA calls. After doing this, the build completed successfully and it seems like FSR is running native on the Oculus Quest. (screenshots taken on device)

fsr on quest

Another issue I found, after switching from OpenXR to Oculus in the XR Plug-In Management, the black screen in editor Mock HMD returned (only when using FSR, not when using bilinear or nearest-neighbor). When I tried switching back to OpenXR, the black screen persisted (even though I am still running your latest commit locally), even after restarting the editor.

gmitrano-unity:
> **Edit** Fixed it locally by converting ApplyFXAA, FXAALoad & FXAAFetch to accept TEXTURE2D_X as inputs, will see what you come up with for an official fix, I could probably learn something here.

I think that's the correct fix actually. :)
Once I turned on FXAA locally, the issue reproduced once the mock HMD was set to use single pass instancing mode.
Adding the "_X" suffixes fixes the issue by adding handling for the 2d array case which is exclusive to single pass instance mode (afaik). I've added that change to this PR and the other FXAA PR.

> I haven't had any more issues with both commits merged locally, would love to see 16-bit + FXAA working.

Great! Thanks for all the help with testing!

You can try out the 16-bit FSR implementation if you'd like by removing the comment sections I added in this commit: https://github.com/Unity-Technologies/Graphics/pull/6262/commits/2ced1604df285b8d88f3e3cf83b5f01b28de090c
I just disabled it for now to avoid causing further confusion relating to compatibility since there are some known issues. I'm hoping to have those resolved before this PR gets merged though.

LaserYGD:
Thank you for exposing sharpness value
I've been asking the same for HDRP for sometime but no luck, it would be great if HDRP would follow URP with that bit. 

Thanks again 😄

com.unity.render-pipelines.core/Runtime/Utilities/FSRUtils.cs
```diff
@@ -86,25 +86,42 @@ public static void SetEasuConstants(CommandBuffer cmd, Vector2 inputViewportSize
...
+        /// <summary>
+        /// The maximum sharpness value in stops before the effect of RCAS is no longer visible.
+        /// This value is used to map between linear and stops.
+        /// </summary>
+        internal const float kMaxSharpnessStops = 2.5f;
+
+        /// <summary>
+        /// AMD's FidelityFX Super Resolution integration guide recommends a value of 0.2 for the RCAS sharpness parameter when specified in stops
+        /// </summary>
+        public const float kDefaultSharpnessStops = 0.2f;
```
manuele-bonanno, line 98: this could be internal as well?

manuele-bonanno:
since we discussed that the benefits of this on mobile are not worth the cost most likely, should the option still be available on those platforms at all? Or at least adding some toll tips/documentation advice about on which platforms this filter is recommended

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -221,6 +221,12 @@ void SetPerCameraShaderVariables(CommandBuffer cmd, ref CameraData cameraData)
 
             cmd.SetGlobalVector(ShaderPropertyId.screenSize, new Vector4(cameraWidth, cameraHeight, 1.0f / cameraWidth, 1.0f / cameraHeight));
 
+            // Calculate a bias value which corrects the mip lod selection logic when image scaling is active.
+            // We clamp this value to 0.0 or less to make sure we don't end up reducing image detail in the downsampling case.
+            float mipBias = Math.Min((float)-Math.Log(cameraWidth / scaledCameraWidth, 2.0f), 0.0f);
+            cmd.SetGlobalFloat(ShaderPropertyId.globalMipBias, mipBias);
+            cmd.SetGlobalFloat(ShaderPropertyId.globalMipBiasPow2, Mathf.Pow(2.0f, mipBias));
+
```
manuele-bonanno, line 229: are we still landing the mipBias PR as a separate PR? Asking since these changes are the same

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -15,6 +15,36 @@ public enum MixedLightingSetup
         Subtractive,
     };
 
+    /// <summary>
+    /// Enumeration that indicates what kind of image scaling is occurring if any
+    /// </summary>
+    public enum ImageScaling
+    {
```
manuele-bonanno, line 22: @sandy-carter-unity how does this fit into the dynamic scaling work in progress?

manuele-bonanno (approved):

com.unity.render-pipelines.universal/Shaders/PostProcessing/EdgeAdaptiveSpatialUpsampling.shader
```diff
@@ -0,0 +1,59 @@
+Shader "Hidden/Universal Render Pipeline/Edge Adaptive Spatial Upsampling"
+{
+    HLSLINCLUDE
+        #pragma multi_compile _ _USE_DRAW_PROCEDURAL
```
phi-lira, line 4:
```suggestion
        #pragma multi_compile_vertex _ _USE_DRAW_PROCEDURAL
```

com.unity.render-pipelines.universal/Shaders/PostProcessing/EdgeAdaptiveSpatialUpsampling.shader
```diff
@@ -0,0 +1,59 @@
...
+
+        TEXTURE2D_X(_SourceTex);
+        float4 _SourceSize;
+
+        #define FSR_INPUT_TEXTURE _SourceTex
+        #define FSR_INPUT_SAMPLER sampler_LinearClamp
+
+        // Only enable 16-bit instructions when the underlying hardware supports them
+        // Note: There are known issues on DX11 drivers so we don't enable 16-bit mode when DX11 is detected
+        #if HAS_HALF && !defined(SHADER_API_D3D11)
```
phi-lira, line 19: This is interesting, Looking at the code in common.hlsl it seems we should use REAL_IS_HALF and you should use `real` instead of `half` in any case that you want fp16. This should work ™️  but well, I know expectations are hard when it comes to drivers 🤣. Either way, I think this kind of check should not be done at URP level, URP should consume REAL_IS_HALF and if there's any kind of specific platform issue / driver issue the exception for DX11 should be handled at the definion of REAL_IS_HALF 
https://sourcegraph.com/github.com/Unity-Technologies/Graphics/-/blob/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl?L85

com.unity.render-pipelines.universal/Shaders/PostProcessing/EdgeAdaptiveSpatialUpsampling.shader
```diff
@@ -0,0 +1,59 @@
...
+
+            // Convert back to linear color space before this data is sent into RCAS
+            color = Gamma20ToLinear(color);
+
+            return half4(color, 1.0);
+        }
+
+    ENDHLSL
+
+    SubShader
```
phi-lira, line 42: Would be nice to have a comment to explain this subshader.

com.unity.render-pipelines.universal/Shaders/PostProcessing/FinalPost.shader
```diff
@@ -24,6 +25,19 @@ Shader "Hidden/Universal Render Pipeline/FinalPost"
         float4 _Grain_TilingParams;
         float4 _Dithering_Params;
 
+        #if SHADER_TARGET >= 45
+            #define FSR_INPUT_TEXTURE _SourceTex
+            #define FSR_INPUT_SAMPLER sampler_LinearClamp
+
+            // Only enable 16-bit instructions when the underlying hardware supports them
+            // Note: There are known issues on DX11 drivers so we don't enable 16-bit mode when DX11 is detected
+            #if HAS_HALF && !defined(SHADER_API_D3D11)
```
phi-lira, line 34: Same comment as the REAL_IS_HALF here.

com.unity.render-pipelines.universal/Shaders/PostProcessing/FinalPost.shader
```diff
@@ -149,6 +102,25 @@ Shader "Hidden/Universal Render Pipeline/FinalPost"
 
     ENDHLSL
 
+    SubShader
```
phi-lira, line 105: Please add comment to explain this shader.

com.unity.render-pipelines.universal/Shaders/PostProcessing/UpscaleSetup.shader
```diff
@@ -0,0 +1,57 @@
...
+            half3 color = SAMPLE_TEXTURE2D_X(_SourceTex, sampler_PointClamp, uv).xyz;
+
+#if _FXAA
+            color = ApplyFXAA(color, positionNDC, positionSS, _SourceSize, _SourceTex);
+#endif
+
+#if _GAMMA_20
+            // EASU expects the input image to be in gamma 2.0 color space so perform color space conversion
+            // while we store the pixel data from the setup pass.
+            color = LinearToGamma20(color);
```
phi-lira, line 33: What does happen in non HDR cases? I guess that's no common in regular 3D games, but could be case of 2D unlit ones. Not even sure if FSR would be benficial in those cases as the render pipeline is simple anyway. Should we make requirement that FSR can only run when HDR is enabled?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
gmitrano-unity: Yep, that commit was pulled into this PR temporarily so it'd represent final-like quality. It'll be removed when this becomes non-draft.

com.unity.render-pipelines.core/Runtime/Utilities/FSRUtils.cs
gmitrano-unity: I don't think I can make these internal since they're located in core and used by URP? I get accessibility compiler errors when I try.

manuele-bonanno: ah ok, I just looked at the diffs here and it looked like it wasn't used by other packages. Then public is fine, thanks for checking :)

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAsset/UniversalRenderPipelineAssetUI.Skin.cs
```diff
@@ -37,6 +37,9 @@ public static class Styles
             public static GUIContent hdrText = EditorGUIUtility.TrTextContent("HDR", "Controls the global HDR settings.");
             public static GUIContent msaaText = EditorGUIUtility.TrTextContent("Anti Aliasing (MSAA)", "Controls the global anti aliasing settings.");
             public static GUIContent renderScaleText = EditorGUIUtility.TrTextContent("Render Scale", "Scales the camera render target allowing the game to render at a resolution different than native resolution. UI is always rendered at native resolution.");
+            public static GUIContent upscalingFilterText = EditorGUIUtility.TrTextContent("Upscaling Filter", "Controls the type of filter used for upscaling when render scale is lower than 1.0.");
+            public static GUIContent fsrOverrideSharpness = EditorGUIUtility.TrTextContent("Override FSR Sharpness", "Overrides the FSR sharpness value for the render pipeline asset.");
+            public static GUIContent fsrSharpnessText = EditorGUIUtility.TrTextContent("FSR Sharpness", "Controls the intensity of the sharpening filter used by FidelityFX Super Resolution.");
```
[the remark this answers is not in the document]
gmitrano-unity: The reason for the additional toggle is for backwards compatibility. It allows us to distinguish between the case where the user wants to use Unity's default value, and the case where the user wants their own specific value. If Unity's default value were to change, we wouldn't be able to automatically migrate existing projects to use the new default without this checkbox since we wouldn't be able to tell if users had selected the old value intentionally or if it was just the default initialized value saved into the project.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -1391,6 +1399,122 @@ void RenderFinalPass(CommandBuffer cmd, ref RenderingData renderingData)
...
+                {
+                    if (isFxaaEnabled)
+                    {
+                        m_Materials.upscaleSetup.EnableKeyword(ShaderKeywordStrings.Fxaa);
+                    }
+
+                    if (isFsrEnabled)
+                    {
+                        m_Materials.upscaleSetup.EnableKeyword(ShaderKeywordStrings.Gamma20);
+                    }
```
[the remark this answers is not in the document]
gmitrano-unity: See: https://github.com/Unity-Technologies/Graphics/pull/6178#discussion_r762097338

Also, I did add the fast path back into this commit. It may be worth keeping since it does seem to trigger on some projects like Viking Village by default.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -1265,6 +1287,76 @@ static void CheckAndApplyDebugSettings(ref RenderingData renderingData)
...
+
+                    break;
+                }
+
+                case UpscalingFilterSelection.Point:
+                {
+                    filter = ImageUpscalingFilter.Point;
+
+                    break;
+                }
```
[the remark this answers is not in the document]
gmitrano-unity: See discussion here: https://github.com/Unity-Technologies/Graphics/pull/6178#discussion_r761865870

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -10,11 +10,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Renderer Features can now use the HelpURLAttribute to specify a documentation URL to be used in the inspector.
 - Added inspector documentation URLs to the SSAO, Decal, and Render Objects renderer features.
 - Changed "_USE_DRAW_PROCEDURAL" to be used only in vertex shader in Post Processing related shaders as they are not needed for fragment shaders. In result we now generate less shader variants.
+- Added support for upscaling filter selection.
```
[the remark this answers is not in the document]
gmitrano-unity: Updated the changelog for the FSR filter, and added a separate line for the other filters in the previous fxaa change.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAsset/UniversalRenderPipelineAssetUI.Drawers.cs
```diff
@@ -148,6 +148,21 @@ static void DrawQuality(SerializedUniversalRenderPipelineAsset serialized, Edito
             EditorGUILayout.PropertyField(serialized.hdr, Styles.hdrText);
             EditorGUILayout.PropertyField(serialized.msaa, Styles.msaaText);
             serialized.renderScale.floatValue = EditorGUILayout.Slider(Styles.renderScaleText, serialized.renderScale.floatValue, UniversalRenderPipeline.minRenderScale, UniversalRenderPipeline.maxRenderScale);
+            EditorGUILayout.PropertyField(serialized.upscalingFilter, Styles.upscalingFilterText);
+            if (serialized.asset.upscalingFilter == UpscalingFilterSelection.FSR)
```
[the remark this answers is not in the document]
gmitrano-unity: Hmmm, so would this mean that in order for users to see the FSR override checkbox, they'd need to select FSR upscaling, and then additionally change the render scale to something at or below 1.0? I could change it to work this way, but when dynamic resolution comes into play, do you think we'd end up changing it again since the render scale slider would become dynamic and users might not change it manually anymore?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -1265,6 +1287,76 @@ static void CheckAndApplyDebugSettings(ref RenderingData renderingData)
...
+                    float pixelScale = (1.0f / renderScale);
+                    bool isIntegerScale = ((pixelScale - Mathf.Floor(pixelScale)) == 0.0f);
+
+                    if (isIntegerScale)
+                    {
+                        float widthScale = (imageSize.x / pixelScale);
+                        float heightScale = (imageSize.y / pixelScale);
+
+                        bool isImageCompatible = (((widthScale - Mathf.Floor(widthScale)) == 0.0f) &&
+                                                  ((heightScale - Mathf.Floor(heightScale)) == 0.0f));
```
sandy-carter-unity, line 1324: There's a risk of float imprecision using `a - b == 0.0f`, prefer using `Math.Abs(a - b) < epsilon`

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
sandy-carter-unity: The dynamic scaling work is downscaling only.
It uses viewports of already allocated targets.

ernestasKupciunas (commented):
Looks good.
Found 1 issue regards the M1 Pro Max chip - Balck screen in the Game view @gmitrano-unity is aware of this already.

Full testing doc: https://docs.google.com/document/d/1l536oqerdT1M-OJcy-ynjDCpf0IMPLGtdO0LZHnYdRI/edit#
Also, 1 issue not related to this PR: Depth of field effect gives screen distortions on Adreno 620GPU

Testing devices:
MacBook Pro (16-inch, 2021), CPU: Apple M1 Max, Memory: 64 GB, macOS: Monterey 12.0
Samsung Galaxy S10+, 10.0.0, Exynos 9 9820, Mali-G76
iPhone 13Pro Max, SoC: A15, iOS: 15.0.1

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -1391,6 +1421,124 @@ void RenderFinalPass(CommandBuffer cmd, ref RenderingData renderingData)
...
+                            case ImageUpscalingFilter.FSR:
+                            {
+                                m_Materials.easu.shaderKeywords = null;
+
+                                var upscaleRtDesc = tempRtDesc;
+                                upscaleRtDesc.width = cameraData.pixelWidth;
+                                upscaleRtDesc.height = cameraData.pixelHeight;
+
+                                // EASU
+                                RenderingUtils.ReAllocateIfNeeded(ref m_UpscaledTarget, upscaleRtDesc, FilterMode.Point, TextureWrapMode.Clamp, name: "_UpscaledTexture");
```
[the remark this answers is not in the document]
gmitrano-unity: Is there a place in the ScriptableRenderPass that only gets called when the backbuffer resizes? This particular render target is always the same size as the backbuffer. The other render target used above this one will change size whenever render scale changes so I'm not sure if it's possible to optimize that one.

ernestasKupciunas (approved):
@gmitrano-unity checked both, iOS and Android. Works fine and renders the Scene and Game views as expected. Good job!

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
```diff
@@ -2,6 +2,7 @@ Shader "Hidden/Universal Render Pipeline/UberPost"
 {
     HLSLINCLUDE
         #pragma exclude_renderers gles
+        #pragma multi_compile_local_fragment _ _GAMMA_20
```
phi-lira, line 5: `_GAMMA_20` and `_LINEAR_TO_SRGB_CONVERSION` are mutually exclusive right? If so, I guess we can make 
`multi_compile_local_fragment _ GAMMA_20 _LINEAR_TO_SRGB_CONVERSION` and save some variants.

phi-lira (approved):

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -673,6 +691,24 @@ public float renderScale
...
+            set { m_UpscalingFilter = value; }
+        }
+
+        public bool fsrOverrideSharpness
+        {
+            get { return m_FsrOverrideSharpness; }
+            set { m_FsrOverrideSharpness = value; }
+        }
+
+        public float fsrSharpness
```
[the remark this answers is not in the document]
gmitrano-unity: I see docs for these in the diff on my end?

```cs
/// <summary>
/// Returns the upscaling filter desired by the user
/// Note: Filter selections differ from actual filters in that they may include "meta-filters" such as
///       "Automatic" which resolve to an actual filter at a later time.
/// </summary>
public UpscalingFilterSelection upscalingFilter
{
    get { return m_UpscalingFilter; }
    set { m_UpscalingFilter = value; }
}

/// <summary>
/// If this property is set to true, the value from the fsrSharpness property will control the intensity of the
/// sharpening filter associated with FidelityFX Super Resolution.
/// </summary>
public bool fsrOverrideSharpness
{
    get { return m_FsrOverrideSharpness; }
    set { m_FsrOverrideSharpness = value; }
}

/// <summary>
/// Controls the intensity of the sharpening filter associated with FidelityFX Super Resolution.
/// A value of 1.0 produces maximum sharpness while a value of 0.0 disables the sharpening filter entirely.
///
/// Note: This value only has an effect when the fsrOverrideSharpness property is set to true.
/// </summary>
public float fsrSharpness
{
    get { return m_FsrSharpness; }
    set { m_FsrSharpness = value; }
}
```

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
gmitrano-unity: I dug into this a bit and realized that FSR didn't play well with the gamma color encoding project setting so I fixed that in a new commit and addressed the unnecessary shader variants as well.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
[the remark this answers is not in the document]
sandy-carter-unity: There is no place that I am aware of

sandy-carter-unity (approved):

phi-lira:
URPUpdate on Win_DX11_playmode_mono_Linear on version trunk - failing on master, we are investigating. 
ShaderGraph on OSX_OpenGLCore_editmode_mono_Linear on version trunk - [failing on master](https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/master/.yamato%252Fshadergraph-osx-openglcore.yml%2523ShaderGraph_OSX_OpenGLCore_editmode_mono_Linear_trunk) as well.

phi-lira:
Yamato URP PR job is green, except for failures mentioned above. Merged master to solve changelog conflicts.

PR #6267, opened 2021-11-10
Expose option for DoF near blend fix

tlaedre:
### Purpose of this PR

This PR exposes an option to enable near dof blur fixing when using manual range, non-physical DoF.

This is a feature that was hard-code disabled 3 years ago in a PR labelled "fixes and performance optimizations", and has just been sitting there doing nothing ever since.

The hard-coded option: https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DepthOfFieldCoC.compute#L28

The commit disabling it almost 3 years ago: https://github.com/Unity-Technologies/Graphics/commit/1b5447fe581e101e6d54d3da15a3bd893a33f7c4

Does it actually make sense to bring this back?

We do have a case in our production that looks more pleasing with this option enabled, and I'd say it generally looks better / closer to physical under certain, quite specific circumstances. (This was also the motivation to create this PR, to lift HDRP customizations out of our project)

If it does not make sense to bring back then I think we should consider removing the codepath instead of just carrying it around disabled until end of time.

I'm including some comparison images from one of the HDRP DoF tests.

Manual near range, physically based DoF:
PhysicallyBased

Manual near range, non-physically based DoF, fixed near blend:
NonPhysicallyFixBlend

Manual near range, non-physically based DoF, no near blend fix (aka current master):
NonPhysicallyNoFixBlend

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
@@ -2174,7 +2174,7 @@ namespace UnityEngine.Rendering.HighDefinition
                 parameters.dofCircleOfConfusionCS.EnableKeyword("USE_MIN_DEPTH");
             }
 
-            if(m_DepthOfField.limitManualRangeNearBlur && m_DepthOfField.focusMode == DepthOfFieldMode.Manual && !m_DepthOfField.physicallyBased && m_DepthOfField.IsNearLayerActive())
+            if (m_DepthOfField.limitManualRangeNearBlur && m_DepthOfField.focusMode == DepthOfFieldMode.Manual && !m_DepthOfField.physicallyBased && m_DepthOfField.IsNearLayerActive())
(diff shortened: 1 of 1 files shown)


pmavridis:
The PR looks good to me, I don't see any problem with conditionally exposing the fix.

One question is if this option could be exposed via scripting only, and hide it from the UI to avoid confusing the users.
If you think it needs to stay in the UI, then I would suggest to change the name of the option a bit, because "Fix" is not very clear. Perhaps something like  "Limit near blur" or "Softer near blur blending" might be more clear to the user (the tooltip is already clear).

pmavridis (approved):

sebastienlagarde:
@tlaedre so regardnig Pavlos comment what's your take? should we move forward with this PR?

tlaedre:
> @tlaedre so regardnig Pavlos comment what's your take? should we move forward with this PR?

After speaking with Pavlos I will rename the option as suggested, but keep it in the UI.

sebastienlagarde (approved):

PR #6271, opened 2021-11-10
[HDRP] Fix custom pass utils in XR

alelievr:
### Purpose of this PR
Fix the custom pass utils Blur and Copy functions in XR.

---
### Testing status
tested in the custom pass sample repository: https://github.com/alelievr/HDRP-custom-passes


fabien-unity (approved):

iM0ve (approved):
Testing should be sufficient for the change

sebastienlagarde (approved):

PR #6275, opened 2021-11-12
Revert Automatic Intermediate Texture handling

pbbastian:
### Purpose of this PR
This PR reverts #5297 due to compatibility issues.


pbbastian:
🟢 `001_SimpleCube_RenderPass` is failing, but is also failing on trunk.

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Multiply blend now keeps DstAlpha as it's RGB only.
 - Particle AlphaModulate() renamed to AlphaModulateAndPremultiply() as it does both. Moved separate AlphaModulate() and AlphaPremultiply() to URP shader library. Fix double alpha multiply for ParticleLit.
 - Improved blending modes trigger a material update which tries to keep the existing look intact. This is not always possible and manual blend mode changes might be required.
+- Reverted intermediate texture behavior.
```
phi-lira, line 48: This might be confusing to users, as it's also not very descriptive as well. 
 Maybe it should also be a good idea to revert the original changelog the intermediate RT texture added?

com.unity.render-pipelines.universal/Runtime/UniversalRendererData.cs
```diff
@@ -236,19 +235,6 @@ internal TileSize tileSize
             }
         }
 
-        /// <summary>
-        /// Controls when URP renders via an intermediate texture.
-        /// </summary>
-        public IntermediateTextureMode intermediateTextureMode
```
phi-lira, line 242: hm, This will probably fail API validation.

phi-lira (approved):
LGTM. Main convern I think this will pass package validation on master but will fail on backport PR.

com.unity.render-pipelines.universal/CHANGELOG.md
pbbastian: I think we should leave those be, as those versions will have the behavior. But yeah, it does sound confusing. I'll make it sound a bit nicer :)

com.unity.render-pipelines.universal/Runtime/UniversalRendererData.cs
pbbastian: Hmm, is there something we can do? Leaving in the API point but unfunctional would be odd.

phi-lira: We have two options, either disable API validation or make it obsolete right away. None of them are good. I vote for making it obsolete as it's less disruptive.

For master / this PR it's fine to remove it.

phi-lira (approved):

thomas-zeng:
Hello Peter,
I was checking the recent master lately and figured the auto intermediate texture handling logic is reverted. This brings the intermediate RT back when the render feature is being used in URP. I am curious about what compatibility issue we were running into with the auto handling logic?

PR #6278, opened 2021-11-15
Fix light flare for pbr sky

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1379487/
https://fogbugz.unity3d.com/f/cases/1381384/

First, slight change to fix from https://github.com/Unity-Technologies/Graphics/pull/6145, it was not handling correclty the case with multiple dircetional lights. Moved the precomputed data to the directional light data struct

Second, updated the template light to have a white flare tint. For some reason it was changed to black, causing the flare to be invisible

---
### Testing status

Tested to change flare size with multiple directinal lights on the pbr sky

---
### Comments to reviewers

The second fix mentionned above is a bit strange, @Unity-Technologies/gfx-qa-hdrp maybe you could investigate a bit to find why the flare color became black in the template, there may be a regression or an update issue ?


FrancescoC-unity (approved):

iM0ve (changes requested):
**Testing Status:** In progress

adrien-de-tocqueville:
You are right, it doesn't have any effect, because there is no flare and the sky is hdri anyway, but in case a user want to try out flare with pbr sky, it is more straightforward this way

I asked pierre and it was an unwanted change so i'll revert to the default flare color which is white

iM0ve (approved):
All concerns/issues were addressed. Good to go

sebastienlagarde (approved):

PR #6284, opened 2021-11-15
Fixed interpolation issue with wind orientation (case 1379841).

anisunity:
https://fogbugz.unity3d.com/f/cases/1379841/
The interpolation for the custom mode, wasn't properly done (taking shortest path). This also removes the possibility of using the multiply mode for wind orientation.

Testing status
Pair tested with @pierre-unity


RemyUnity (approved):
As the bug cas was entered by @pierre-unity , and the description mention the fix was pair-tested with him, I assume testing is sufficient.

sebastienlagarde (approved):

PR #6285, opened 2021-11-15
Fixed range fields for depth of field

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1376609/

Added checks to prevent far being lower than near and the other way around

dof-range

---
### Testing status

see the gif aboce

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.core/CHANGELOG.md
+++ b/com.unity.render-pipelines.core/CHANGELOG.md
@@ -22,9 +22,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added new DebugUI widget types: ProgressBarValue and ValueTuple
 - Added common support code for FSR
 
-### Fixed
-- Fixed issue when changing volume profiles at runtime with a script (case 1364256).
-
 ## [13.1.0] - 2021-09-24
 
 ### Added


RemyUnity (changes requested):
Looking at your video capture, it's nice that the **end** value of a range can't be smaller than the **start** value, but is it expected that the **start** of **_near_** can be bellow the **end** of **_far_** ?

adrien-de-tocqueville:
> Looking at your video capture, it's nice that the **end** value of a range can't be smaller than the **start** value, but is it expected that the **start** of **_near_** can be higher than the **end** (and **start**) of **_far_** ?

i don't know, @pmavridis does it make sense ?

JulienIgnace-Unity:
I guess it does make sense. Probably harmless to do it too.

adrien-de-tocqueville:
yeah so i checked it's done anyway when preparing parameters for rendering so no issue.
Added necessary checks so that all orders are respected but only checking overriden parameters

pmavridis (approved):

pmavridis:
>  is it expected that the start of near can be higher than the end (and start) of far ?
I think that preventing this in the UI makes it clear for the users how to use the feature (overlapping near/far ranges does not make much sense in practice), so it's good that we have added the extra check in the code. The PR looks good to me now.

RemyUnity (approved):
Tested and everything seems to work fine :)

sebastienlagarde (approved):

PR #6286, opened 2021-11-15
[ant-276] copy project with rsync

hazelbit:
## Motivation

Small PR to copy the project using [rsync](https://linux.die.net/man/1/rsync) to remote macOS devices to improve stability and reduce walltime. 

The two main benefits are increased resilience (especially in regards to retries) and skipping `.` files/folders.

Quick note: `rsync` can compress files for transfer (`-z`, trading cpu utilization and network bandwidth) which led to a walltime penalty in our scenarios (see [ticket](https://jira.unity3d.com/browse/ANT-276) for more details) so it is not part of the PR.

Draft PR with more pipelines to be scheduled, for success we should not see regressions in walltime and all further steps should find their required files in the same location (we should not see any pipeline failures related to this change). See [ant-276](https://jira.unity3d.com/browse/ANT-276).

### DIFF
--- a/.yamato/config/__shared.metafile
+++ b/.yamato/config/__shared.metafile
@@ -350,7 +350,7 @@ platforms:
           - command: curl -s https://artifactory.prd.it.unity3d.com/artifactory/unity-tools-local/utr-standalone/utr --output utr
             retries: 2
           - chmod +x utr
-          - command: rsync -uaqP --exclude=".*" --inplace $YAMATO_SOURCE_DIR bokken@$BOKKEN_DEVICE_IP:~/
+          - command: rsync -uaqP -e 'ssh -i ~/.ssh/id_rsa_macmini' --exclude=".*" --inplace $YAMATO_SOURCE_DIR bokken@$BOKKEN_DEVICE_IP:~/
             retries: 2
           - command: scp -i ~/.ssh/id_rsa_macmini ~/.ssh/id_rsa_macmini bokken@$BOKKEN_DEVICE_IP:~/.ssh/id_rsa_macmini
             retries: 2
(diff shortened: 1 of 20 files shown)


.yamato/config/__shared.metafile
```diff
@@ -350,7 +350,7 @@ platforms:
           - command: curl -s https://artifactory.prd.it.unity3d.com/artifactory/unity-tools-local/utr-standalone/utr --output utr
             retries: 2
           - chmod +x utr
-          - command: scp -i ~/.ssh/id_rsa_macmini -r $YAMATO_SOURCE_DIR bokken@$BOKKEN_DEVICE_IP:~/Graphics
+          - command: rsync -uaqP -e 'ssh -i ~/.ssh/id_rsa_macmini' --exclude=".*" --inplace $YAMATO_SOURCE_DIR bokken@$BOKKEN_DEVICE_IP:~/
```
theo-pnv, line 353: The arguments look good to me, looking at the man.
Except I'm wondering why the `exclude=".*"`?

theo-pnv:
It's unclear whether the retry mechanism in case the connection was reset is built-in rsync or not. My understanding is that it's not, so if we want to have one maybe we'll have to have to call it inside some `while` loop? See this SO reply https://superuser.com/a/302843

hazelbit:
That is right, I was relying on the `retries: 2` mechanism (so two attempts). The `-P` implies `--partial` which is what we want for the retries (though arguably I should pass that explicitly, the second implied arg, `--progress`, is unused due to `-q/--quiet`, it's a leftover). So this setup would fail if rsync did not receive any reply within the default 30s after which yamato would retry the command. rsync will figure out which files should be synced (checksums) and only attempt to transfer what is necessary. 

I'll close this PR again and follow up, to my understanding the problem with the previous approach was that we could get partial copies but still a `success` exit code, now we should see a clear success or failure code, though I will gather how resilient the job should become (focus on the resilience of the copy vs resilience of the job).

.yamato/config/__shared.metafile
hazelbit: Closed PR, the motivation was that to skip `.` files/folders (e.g. `.git/` as well as others like `.yamato/`).

PR #6288, opened 2021-11-15
Fix cascade blend transition when using an high normal bias

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1377713/ 

Two line fix but a long investigation :P 

Long story short, given the blend evaluates two cascades the normal bias needs to be recomputed when evaluating the shadows for the second cascade as the world space covered by a texel changes significantly and that needs to be taken into account. 

**Before**

https://user-images.githubusercontent.com/43168857/141818620-3c5ef49b-faee-4786-88ef-665f2b9ded4b.mp4

**After**

https://user-images.githubusercontent.com/43168857/141818636-ebb40b4a-9131-4a40-a299-bc09e1714ac4.mp4

**What did I test** A setup as show in the video above


RemyUnity (approved):
LGTM

sebastienlagarde (approved):

PR #6289, opened 2021-11-15
Fixed broken rendering when duplicating a camera while the Rendering Debugger is opened.

JulienIgnace-Unity:
### Purpose of this PR
Fixed broken rendering when duplicating a camera while the Rendering Debugger is opened.
This was due to uninitialized frame settings registered to the debug system.

The underlying problem is with the Rendering Debugger undo system. The issue is that when duplicating a camera and registering it to the debug system, we directly register the state of all the frame settings to the undo system. The problem is that they are not initialized yet. So to fix this we call AggregateFrameSettings once in OnEnable in order to properly initialize the debug data.

Fixes: 
https://fogbugz.unity3d.com/f/cases/1379200/
https://fogbugz.unity3d.com/f/cases/1378334/

---
### Testing status
Tested the repro case and tried to load/unload scenes and duplicate cameras while rendering debugger is opened.

---
### Comments to reviewers
It might be worth checking a bit more in depth that this fix does not cause additional issues to the frame settings.
(duplication, serialization etc)


TomasKiniulis (approved):
Fix looks good!
- Tested that Rendering Debugger Undo/Redo, Reset of various debug modes do not cause additional PR related issues when camera is Duplicated or Deleted.
- Made sure that tweaking Frame Settings in Global Settings assets or Custom Frame Settings duplicating camera are kept and visible in Game/Scene view.
- Made sure undo/redo is working in Frame Settings and settings a preserved reopening the project.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
```diff
@@ -690,6 +690,11 @@ void OnEnable()
             m_Camera.allowMSAA = false; // We don't use this option in HD (it is legacy MSAA) and it produce a warning in the inspector UI if we let it
             m_Camera.allowHDR = false;
 
+            // By doing that, we force the update of frame settings debug data once. Otherwise, when the Rendering Debugger is opened,
+            // Wrong data is registered to the undo system because it did not get the chance to be updated once.
+            FrameSettings dummy = new FrameSettings();
+            FrameSettingsHistory.AggregateFrameSettings(ref dummy, m_Camera, this, HDRenderPipeline.currentAsset, null);
```
RSlysz, line 696: Then shouldn't it be in `RegisterDebug()`? It would be more clear no?

RSlysz (approved):
Nice catch.
If I get it well though, we don't need to do that in all OnEnable. Only if we are in the editor or if we are in a dev build. Check if this need to be done to ensure no time lost if possible (even if minimal). Also I believe we can put everything into RegisterDebug() at this point.

sebastienlagarde (approved):

PR #6292, opened 2021-11-15
[Fogbugz # 1377986] Fixing DLSS exception when motion vectors are disabled

kecho:
### Purpose of this PR
DLSSPass fails because it casts incoming textures as RenderTexture. This will cause exception when for example we cast a dummy black texture (the missing velocity texture). Thus, just changing the accepting texture type to Texture will be a robust fix for this issue.
https://fogbugz.unity3d.com/f/cases/1380282

---
### Testing status
Tested repro case in fogbugz


RemyUnity:
Is there any side effect of this that you think might require testing ?

remi-chapelain:
I think you got the wrong fogbugz link. 
I believe [this is the one](https://fogbugz.unity3d.com/f/cases/1380282/) you are fixing.

Looking at the code, makes me think that now DLSS won't make the whole render fail but will just not be used silently ?
Maybe it would be better to have a warning somewhere that DLSS indeed need motion vectors to work properly ? 
This could be shown when enabling DLSS when motion vectors is disabled, or when disabling motion vectors when DLSS is already activated in the HDRP asset. 

Also, it may be common knowledge but there's no mention of motion vectors in the [documentation either](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@13.1/manual/deep-learning-super-sampling-in-hdrp.html)

remi-chapelain (changes requested):
Blocking since I'm awaiting answers from previous comment !

kecho:
@remi-chapelain sorry, yes indeed wrong buglink. Updated! This fix should definitely fix this bug though. Let me know what you find out. Id say we just test having motion vectors off. DLSS can still be used as an anti aliasing effect without motion vectors.

remi-chapelain (approved):
Indeed. I agree on this ✔️

sebastienlagarde (approved):

PR #6296, opened 2021-11-16
Fix for shadows disappearing when no dynamic shadow caster is available

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1379186/ 

The issue was weird, essentially following the repro case, what happens is that no shadow caster are there for the shadow maps, as a result the light is considered _not_ to cast shadows. At the same time the contact shadows for _directional only_ were assumed coming only from the shadow casting directional light. 

This PR fixes the issue above by checking all directional lights directly for contact shadows. 

**What did I test**: the repro case and playing around that


iM0ve (changes requested):
**Testing status:** In progress

**Issues:**
1) Its fixed for rasterization, but still occurs in ray tracing

https://user-images.githubusercontent.com/36502659/142167842-e5001e33-9d7e-449e-b18b-e39cf2352daa.mp4

FrancescoC-unity:
I think I fixed it but don't have a DXR setup at hand, could you please check?

iM0ve (approved):
Yes, everything is fixed. Good to go

sebastienlagarde (approved):

PR #6301, opened 2021-11-16
Fixed ray-tracing variant of the checkerboard shader graph node

eturquin:
Implemented a more readable and robust variant for ray tracing.
The previous, slightly ambiguous code was failing with recent NVidia drivers (was always returning the same color).


anisunity:
Good catch, but i wonder why it was working because the test were not failing

remi-chapelain (approved):
Looks good to me. Works the same on the user side ✔️

com.unity.shadergraph/Editor/Data/Nodes/Procedural/CheckerboardNode.cs
```diff
@@ -27,12 +27,12 @@ protected override MethodInfo GetFunctionToConvert()
             return
 @"
 {
-    UV = (UV.xy + 0.5) * Frequency;
-    $precision2 distance3 = 4.0 * abs(frac(UV + 0.25) - 0.5) - 1.0;
 #if defined(SHADER_STAGE_RAY_TRACING)
-    int2 alpha = saturate(distance3 * FLT_MAX);
```
anisunity, line 33: this must be throwing warnings on console

eturquin: Well, it's gone now.

eturquin: Ah, no it doesn't.

sebastienlagarde (approved):

sebastienlagarde (approved):

sebastienlagarde:
For SG team: the behavior of the node don't change for regular rasterization, only for raytracing, merging.

PR #6303, opened 2021-11-16
[HDRP] Improve decal performances when they use different material and the same draw order.

alelievr:
### Purpose of this PR
Optimize decals for this use case: https://forum.unity.com/threads/decal-projector-performance.965822/

Note that this PR changes the behavior of the order in which decals are rendered when they have the same draw order. This is fine because two decals with the same draw order are not guaranteed to be rendered in a specific order (but the sort is still stable between frames).

---
### Testing status
Initial profiling with 500 decals in an empty scene. Decals are all using a unique material spawned with this script: 
```cs
        for (int i = 0; i < count; i++)
        {
            var d = GameObject.Instantiate(decal, new Vector3(Random.Range(-radius, radius), 0.5f, Random.Range(-radius, radius)), Quaternion.identity, parent.transform);
            var proj = d.GetComponent<DecalProjector>();
            proj.material = new Material(proj.material);
        }
```

Decal projector settings are identical for every decal:
image

The decal shader is a ShaderGraph with a simple gradient noise animated based on time.

This is without the fix in CPU profiling:
image

With the fix:
image

As you can see on the CPU side there is a 5x improvement. GPU side is identical since the decal list is the same, only order changes.

### Notes for the future

We may benefit from a proper instancing system for decals, meaning that our decals sets will be stored per shader and not per material. This would remove the cost of having to go through the initialization of jobs for every decal material in a scene (this is currently the next big bottleneck on decals). Though that would involve a lot of work since all the ShaderGraph / decal properties would need to be instanced in the shader.

### Yamato
https://yamato.cds.internal.unity3d.com/jobs/902-Graphics/tree/hd%252Fimprove-decal-material-perfs/.yamato%252Fall-hdrp.yml%2523PR_HDRP_trunk/9969232/job


kecho (approved):

kecho:
Good change!

RemyUnity (approved):
Ship it !

sebastienlagarde (approved):

sebastienlagarde:
Please, add changelog

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -72,6 +72,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed the point distribution for the diffuse denoiser sometimes not being properly intialized.
 - Fixed the bad blending between the sun and the clouds (case 1373282).
 - Fixed and optimize distance shadowmask fade.
+- Fixed compilation errors when using Elipse, Rectangle, Polygon, Checkerboard, RoundedPolygon, RoundedRectangle in a ray tracing shader graph (case 1377610).
```
sebastienlagarde, line 75: please fix changelog

PR #6306, opened 2021-11-16
[RPW] Full deprecation of Density Volume

alex-vazquez:
### Purpose of this PR
Applying obsolete path for Density Volume
Version 12 Deprecation of Density Volume with warnings ( Done when renaming Local Volumetric Fog )
Version 13 Full Deprecation ( Done in this PR )
Version 14 Full delete the classes


theo-pnv (approved):
As far as packman is concerned it looks good now. The packages test is still failing but it's because of the remaining URP issue, so unrelated to this PR. There is no mention of any HDRP class anymore. Thanks Alex! 🚀

sebastienlagarde (approved):

PR #6308, opened 2021-11-16
Fixed the missing debug menus to visualize the ray tracing acceleration structure (case 1371410).

anisunity:
https://fogbugz.unity3d.com/f/cases/1371410/
This PR adds a full screen debug menu that allows the user to visualize what is in the various layers of the acceleration structure (and to understand if an object is missing). This allowed to figure out that a bug was introduced with the c++ culling feature that needs to be fixed in a separate (engine PR).
This also allowed to remove the flagging of the shadow flags when the light had ray traced shadows but disabled shadows.

The user has now two sub options to pick in what mode they want to display (primitiveID, instanceID) and view (reflection, GI, Path tracing, etc)
image

Testing status
Tried to test as much as possible all the different combinations

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/9920_WaterSurface.png
+++ b/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/OSXEditor/Metal/None/9920_WaterSurface.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:8b057497f3026af72ec8ef93599d2066659acccb9328ec34b1018ce20780a48b
+size 248813
--- a/TestProjects/HDRP_Tests/Assets/Tests/TestFilters/TestCaseFilters.asset
+++ b/TestProjects/HDRP_Tests/Assets/Tests/TestFilters/TestCaseFilters.asset
@@ -166,12 +166,3 @@ MonoBehaviour:
     XrSdk: 
     StereoModes: 0
     Reason: unstable with cache server
-  - FilteredScene: {fileID: 0}
-    FilteredScenes:
-    - {fileID: 102900000, guid: fa0d54c5ccbc8b143a8db100374e0f5e, type: 3}
-    ColorSpace: 1
-    BuildPlatform: -2
-    GraphicsDevice: 16
-    XrSdk: 
-    StereoModes: 0
-    Reason: not supported on metal for now
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed auto-exposure mismatch between sky background and scene objects in path tracing (case 1385131).
 - Fixed option to force motion blur off when in XR.
 - Fixed write to VT feedback in debug modes (case 1376874)
+- Fixed the water system not working on metal.
 - Fixed the missing debug menus to visualize the ray tracing acceleration structure (case 1371410).
 
 ### Changed
--- a/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.Skin.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.Skin.cs
@@ -165,7 +165,6 @@ namespace UnityEditor.Rendering.HighDefinition
             public static readonly GUIContent probeVolumeSHBands = EditorGUIUtility.TrTextContent("SH Bands", "Determines up to what SH bands the Probe Volume will use. Choosing L2 will lead to better quality, but also higher memory and runtime cost.");
             public static readonly GUIContent maxLocalVolumetricFogSizeStyle = EditorGUIUtility.TrTextContent("Max Local Fog Size", "Specifies the maximum size for the individual 3D Local Volumetric Fog texture that HDRP uses for Local Volumetric Fog. This settings will affect your memory consumption.");
             public static readonly GUIContent maxLocalVolumetricFogOnScreenStyle = EditorGUIUtility.TrTextContent("Max Local Fog On Screen", "Sets the maximum number of Local Volumetric Fog can handle on screen at once. This settings will affect your memory consumption.");
-            public static readonly GUIContent waterUnsupportedWarning = EditorGUIUtility.TrTextContent("Water surfaces are not supported on metal for now.");
 
             public const string cacheErrorFormat = "This configuration will lead to more than 2 GB reserved for this cache at runtime! ({0} requested) Only {1} element will be reserved instead.";
             public const string cacheInfoFormat = "Reserving {0} in memory at runtime.";
--- a/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.cs
@@ -715,9 +715,6 @@ namespace UnityEditor.Rendering.HighDefinition
             EditorGUILayout.PropertyField(serialized.renderPipelineSettings.supportWater, Styles.supportWaterContent);
             using (new EditorGUI.DisabledScope(!serialized.renderPipelineSettings.supportWater.boolValue))
             {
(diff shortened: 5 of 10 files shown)


com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDShaderPasses.cs
```diff
@@ -1286,6 +1286,61 @@ IncludeCollection GenerateIncludes()
...
 
+        #region Raytracing Debug
+
+        public static PassDescriptor GenerateRaytracingDebug()
+        {
+            return new PassDescriptor
+            {
+                // Definition
+                displayName = "DebugDXR",
+                referenceName = "SHADERPASS_RAYTRACING_Debug",
```
arttu-peltonen, line 1297: Debug -> uppercase?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPass.cs.hlsl
```diff
@@ -31,8 +31,9 @@
 #define SHADERPASS_RAYTRACING_GBUFFER (21)
 #define SHADERPASS_RAYTRACING_SUB_SURFACE (22)
 #define SHADERPASS_PATH_TRACING (23)
-#define SHADERPASS_CONSTANT (24)
-#define SHADERPASS_FULL_SCREEN_DEBUG (25)
+#define SHADERPASS_RAY_TRACING_DEBUG (24)
```
arttu-peltonen, line 34: Does this define work? Elsewhere you're using SHADERPASS_RAYTRACING_DEBUG (without underscore in the middle).

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
```diff
@@ -387,6 +413,12 @@ internal DebugDisplaySettings()
                 .ToArray();
             s_MsaaSamplesDebugValues = (int[])Enum.GetValues(typeof(MSAASamples));
 
+            s_RTASViewDebugStrings = Enum.GetNames(typeof(RTASDebugView)).Select(t => new GUIContent(t)).ToArray();
```
arttu-peltonen, line 416: I think you can get rid of these arrays by using DebugUI.EnumField.autoEnum.

looks like the MSAADebugValues/Strings properties are never even used. Fixing optional :)

arttu-peltonen (changes requested):
Looks like a super useful debug mode if you work with RT! I mostly reviewed the DebugDisplay integration which looks good. I'm less familiar with the shader stuff, but I did notice something that looks suspicious with keywords (or at least doesn't conform to style), so please double-check that.

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDTarget.cs
```diff
@@ -1535,6 +1541,15 @@ static class CoreKeywordDescriptors
...
         };
+
+        public static KeywordDescriptor RayTracingDebug = new KeywordDescriptor()
+        {
+            displayName = "RayTracingDebug",
+            referenceName = "RAY_TRACING_DEBUG",
+            type = KeywordType.Boolean,
+            definition = KeywordDefinition.MultiCompile,
+            scope = KeywordScope.Global
+        };
```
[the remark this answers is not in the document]
anisunity: Yeah, they debug_display keyword may be triggered and is not used in the same context

com.unity.render-pipelines.high-definition/Editor/Material/ShaderGraph/HDShaderPasses.cs
anisunity: Indeed

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
anisunity: It's automatically generated, (RayTracing => RAY_TRACING, raytracing => RAYTRACING). Should be with a T upper case

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPass.cs.hlsl
anisunity: It's automatically generated, (RayTracing => RAY_TRACING, raytracing => RAYTRACING). Should be with a T upper case

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugDisplay.cs
anisunity: image

anisunity: it seems to be used

arttu-peltonen: yeah sorry I meant only written to, never read.

arttu-peltonen: When you use autoEnum, you don't need to provide the enumValues and enumNames when constructing the DebugUI.EnumField. So you only want to manually pass in the values/names if you want a subset of the enum or something. But it's really optional to fix, just FYI.

anisunity: Yeah, I am too lazy to fix this one. I'll fix the other ones

remi-chapelain (approved):
I made public (with added summary) some function to be able to fetch this new debug mode from the test framework which I also modified. 

3 tests have been added to prevent future regression. 
- One to test instanceID with RTR
- One to test primitiveID with RTR
- One to test primitiveID with RTShadows

The only problem is that instanceID is not deterministic between scene launch, so there is only one object in the instanceID scene to prevent randomness instanceID attribution. 

**_2009_Debug_RTAS_ScreenSpaceReflections_InstanceID_**
2009_Debug_RTAS_ScreenSpaceReflections_InstanceID

**_2009_Debug_RTAS_Shadows_PrimitiveID_**
2009_Debug_RTAS_Shadows_PrimitiveID

**_2009_Debug_RTAS_ScreenSpaceReflections_PrimitiveID_**
2009_Debug_RTAS_ScreenSpaceReflections_PrimitiveID

Other than that, tested the feature with the rest of the effects / manual RTAS and light layers. ✔️

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingIntersection.hlsl
```diff
@@ -11,11 +11,37 @@ struct RayCone
     float spreadAngle;
 };
 
+// Structure that defines the current state of the visibility
+struct RayIntersectionDebug
+{
+    // Distance of the intersection
+    float t;
+    // Barycentrics of the intersection
+    float2 barycentrics;
```
sebastienlagarde, line 20: This doesn't seems to be used, is it useful? barycentrics

sebastienlagarde (commented):
Let's discuss. From what I read in the code you don't need to add a SHADER_RAYTRACING_Debug  pass to all shader, you can only have a generic one. Is there any reasons you chose to not do that?

arttu-peltonen (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingIntersection.hlsl
anisunity: Yes, it's for the future, a debug mode that we may want to use

anisunity:
As discussed we cannot have a generic closest hit in the ray traced, so i'll be trying the shaderlab fallback feature see if it helps

anisunity:
@remi-chapelain you forgot to include the materials in your commit :(

anisunity:
Something is currently broken with the test framework, the tests wont run when  launching all of em.
The tests that have been added in this PR are green...
image

cdxntchou (approved):
Looks good.  I think Terrain is defining the same usepass list, but you guys look closer to landing, so we can work with them to adapt their PR.

sebastienlagarde (approved):

PR #6311, opened 2021-11-16
Fixed Nans happening due to volumetric clouds when the pixel color is perfectly black (case 1379185).

anisunity:
https://fogbugz.unity3d.com/f/cases/1379185/
Since the improved transmittance code was merged, the case where the background pixel is perfectly black was not properly handled. This handles it better and fixes the Nan issues.

Testing status
The issue doesn't repro anymore on the QA test dataset


iM0ve (approved):
The feature/code is isolated to volumetric clouds, developer testing on the repro projects should be sufficient.

sebastienlagarde (approved):

PR #6313, opened 2021-11-16
Fixed camera bridge action in release build

adrien-de-tocqueville:
### Purpose of this PR

Fix https://fogbugz.unity3d.com/f/cases/1367866/

Camera bridge was never executed in non debug build because of `PostProcessIsFinalPass()`. Plus it reads from the buffer, and we can't bind the backbuffer as shader input, so I forced creation of another buffer in case camerabridge is used
I am not sure of that fix since i am not too familiar with that code but it seems to solve the issue

---
### Testing status

Tested to run the project in a debug and non debug build


JulienIgnace-Unity (approved):

RemyUnity (approved):
works

PR #6314, opened 2021-11-16
Fix when volume asset is null

adrien-de-tocqueville:
### Purpose of this PR

Bug introduced by #5882


adrien-de-tocqueville:
already fixed

PR #6316, opened 2021-11-16
Improve Screen Space Refraction Default Behavior

johnpars:
### Purpose of this PR

Address https://fogbugz.unity3d.com/f/cases/1379733/

This PR changes the default behavior of screen space refraction when there is no reflection probe present. 

Before, the behavior was to configure an infinite projection on-the-fly. With this PR, it will instead default to the renderer's bounding box to be used as the refraction proxy. 

The result of this change is that a user will now get a meaningful default result for a material configured with a box refraction model (image below). 

Left, Box Refraction Mode + Infinite Projection (Before)
Right, Box Refraction Mode + Bounding Box Projection (After)
image

---
### Testing status
- Yamato + Local Tests. 

---
### Comments to reviewers
I didn't think a backport was necessary as it is more of a change to a working behavior, rather than a bug fix.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added denoising for the path tracer.
 - Added an initial version of under water rendering for the water system.
 - Added Asymmetric projection and Screen Coordinates Override frame settings. Adapted post effects to support Screen Coordinates Override. (Used, for example, to support Cluster Display.)
+- Added default DOTS compatible loading shader (MaterialLoading.shader)
+- Add #pragma editor_sync_compilation directive to MaterialError.shader
 
 ### Changed
 - Render Graph object pools are now cleared with render graph cleanup
@@ -128,6 +130,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed Probe Debug view misbehaving with fog.
 - Fixed issue showing controls for Probe Volumes when Enlighten is enabled and therefore Probe Volumes are not supported.
 - Fixed null reference issue in CollectLightsForRayTracing (case 1398381)
+- Fixed camera motion vector pass reading last frame depth texture
 
 ## [14.0.0] - 2021-11-17
 
--- a/TestProjects/BatchRendererGroup_HDRP/Assets/SampleScenes/BRGGameObjects/RenderBRG.cs
+++ b/TestProjects/BatchRendererGroup_HDRP/Assets/SampleScenes/BRGGameObjects/RenderBRG.cs
@@ -1,3 +1,8 @@
+#if UNITY_EDITOR
+//#define ENABLE_PICKING
+#endif
+//#define ENABLE_ERROR_LOADING_MATERIALS
+
 using System;
 using System.Collections.Generic;
 using Unity.Burst;
@@ -32,16 +37,21 @@ public struct DrawKey : IEquatable<DrawKey>
     public uint submeshIndex;
     public BatchMaterialID material;
     public ShadowCastingMode shadows;
+
+#if ENABLE_PICKING
     public int pickableObjectInstanceID;
+#endif
 
     public bool Equals(DrawKey other)
     {
         return
+#if ENABLE_PICKING
+            pickableObjectInstanceID == other.pickableObjectInstanceID &&
+#endif
             meshID == other.meshID &&
             submeshIndex == other.submeshIndex &&
             material == other.material &&
-            shadows == other.shadows &&
-            pickableObjectInstanceID == other.pickableObjectInstanceID;
+            shadows == other.shadows;
     }
 }
 
@@ -103,6 +113,8 @@ struct SHProperties
 
 public unsafe class RenderBRG : MonoBehaviour
 {
+    public bool SetFallbackMaterialsOnStart = true;
+
     private BatchRendererGroup m_BatchRendererGroup;
     private GraphicsBuffer m_GPUPersistentInstanceData;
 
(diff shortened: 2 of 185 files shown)


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Modified HDRP to use common FSR logic from SRP core.
 - Optimized FSR by merging the RCAS logic into the FinalPass shader.
 - Integrate a help box to inform users of the potential dependency to directional lights when baking.
+- Screen space refraction will now default to a spherical projection at ten meters instead of infinity, when no reflection probe proxy is available (1379733).
```
Vic-Cooper, line 79: Slight phrasing change: 

Screen space refraction now defaults to a spherical projection at ten meters instead of infinity when no reflection probe proxy is available (1379733).

com.unity.render-pipelines.high-definition/Documentation~/Refraction-in-HDRP.md
```diff
@@ -58,7 +58,7 @@ This means that, if screen space refraction does not return information for a pi
 
 The first tier of the refraction hierarchy is a screen space solution. To calculate screen space refraction, the algorithm traces a ray starting from the refractive object. It then refracts the ray according to the properties of the material. To compute the refracted ray, the algorithm assumes that the refractive object can be approximated as a simple shape ([Refraction Model](#RefractionModel)) .
 
-The refracted ray will be then intersected against the closest probe proxy volume to find the pixel in screen space that best approximates the result of the refracted ray. If there is no reflection probe proxy available, HDRP will fallback to a projection at infinite.
+The refracted ray will be then intersected against the closest probe proxy volume to find the pixel in screen space that best approximates the result of the refracted ray. If there is no reflection probe proxy available, HDRP will fallback to a default spherical projection of ten meters.
```
Vic-Cooper, line 61: Removing passive voice: 

The refracted ray intersects against the closest probe proxy volume to find the pixel in screen space that best approximates the result of the refracted ray. If there is no reflection probe proxy available, HDRP uses a default spherical projection of ten meters.

Vic-Cooper (changes requested):
I've suggested tiny phrasing changes.

johnpars:
Putting this PR on temporary hold as we revise our plan for the proxy volume - we plan to add the bounding box for transparent objects which will provide a much better default behavior.

johnpars:
This PR is dependent on this Ono PR and should not land before it: https://ono.unity3d.com/unity/unity/pull-request/138270

sebastienlagarde:
@johnpars can you merge master? there is conflcit. thanks

johnpars:
Done, launching Yamato now too

sebastienlagarde:
So it have impact on few test (3 or 4), https://yamato-artifactviewer.prd.cds.internal.unity3d.com/8d56e204-fc9e-4e00-9aab-a6785f95b8fb%2Flogs%2FTestProjects%2FHDRP_Tests%2Ftest-results/TestReport.html need to check if it is exected behavior, from the shot it is more like we go to infinity box than going to a bounding box one.

AlixMi:
refractionbox
 Hello, I noticed that having a reflection probe without proxy volume reset the refraction at infinity, maybe refraction should be at bounding box size again, and edit the message. I think a tooltip about proxy volume usage on the parameter would be good as well ("proxy volumes are used for screen space refraction" or something like that, are they used for something else ?)

AlixMi:
I'm wondering if we should remove the set proxy volume by default now. (Sorry I know it has been changed back and forth). 
Because now, I have the feeling that the prefered behavior would be to use bounding box instead of a proxy volume. So the more general set up would be to disable them and sometimes only have a proxy volume.

remi-chapelain (changes requested):
Love this change. 👍 
It gives you out of the box box (yes, two times) refraction as you'd expect it to work. 

I personally like having the use influence volume as proxy volume checked by default. 
This will ensure that if a user wants infinite projection it has to do something to have it, uncheck this or select infinite in the proxy set.

As Alix said, the only thing I'd like changed on my side would be to have proper tooltips (they are missing) on the proxy volume to explain a bit that this volume will change the way refraction and reflection is done in screen space.. etc

Also, the PR could be updated with the correct new behavior :)

johnpars:
Will incorporate this feedback (PR updated as well, sorry about that). Right now I am also fighting Yamato, which seems to be imploding for this branch.

remi-chapelain (approved):
Tested again with more complex geometry and comparing with the pathtracer results for refraction as a reference. 

I still think this is a huge improvement to use the bounding box by default and gives relatively good results without having to setup anything. The reflection probe + proxy still gives you the possibility to adjust the refraction shape for more complex scenarios.

**Still need to add the tooltips** but I'm still approving. ✔️

johnpars:
Yamato is still failing, investigating again the cause.

com.unity.render-pipelines.high-definition/Documentation~/Refraction-in-HDRP.md
```diff
@@ -58,8 +58,7 @@ This means that, if screen space refraction does not return information for a pi
 
 The first tier of the refraction hierarchy is a screen space solution. To calculate screen space refraction, the algorithm traces a ray starting from the refractive object. It then refracts the ray according to the properties of the material. To compute the refracted ray, the algorithm assumes that the refractive object can be approximated as a simple shape ([Refraction Model](#RefractionModel)) .
 
-The refracted ray will be then intersected against the closest probe proxy volume to find the pixel in screen space that best approximates the result of the refracted ray. If there is no reflection probe proxy available, HDRP will fallback to a projection at infinite.
-
+The refracted ray intersects against the closest probe proxy volume to find the pixel in screen space that best approximates the result of the refracted ray. If there is no reflection probe proxy available, HDRP will use the renderer's bounding box to determine the proxy extents.
```
Vic-Cooper, line 61: Almos perfect, we just need to remove "will". 

"...HDRP uses the the renderer's bounding box to determine the proxy extents."

Vic-Cooper (changes requested):
Further small language suggestions.

Vic-Cooper (approved):
Docs changed approved.

PR #6321, opened 2021-11-17
Fixed missing information in the tooltip of affects smooth surfaces of the ray traced reflections denoiser (case 1376918). 

anisunity:
https://fogbugz.unity3d.com/f/cases/1376918/
This is not a bug, when we have multiple bounces, we cannot automatically skip the smooth surfaces as we may have rough surfaces to denoise seen in the reflections. This PR precises the tooltip to remove the ambiguity of the behavior.


remi-chapelain (commented):
I agree with the fix, but I would add it to the doc also in addition of the tooltips (in [here](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@13.0/manual/Override-Screen-Space-Reflection.html#ray-traced)). 

Also, since we don't have a test for it, are we sure that the affect smooth surface still works or I misremember its usecase ? In this example, I'd expect the smooth surface (mirror on the left) to blur when checking affect smooth surface but it doesn't ?

remi-chapelain (approved):
had been discused and everything works as expected. ✔️

PR #6322, opened 2021-11-17
Fix color curves changes not applying immediately in editor

FrancescoC-unity:
Fix for : https://fogbugz.unity3d.com/f/cases/1378596/

The reason was that the curve contents are not part of the hash of the volume component, so we cannot know if a change happened in the curves. As a result, we need to recompute the LUT if any of the parameters on the curve component is overridden (indicating intent of changing something) 

**What did I test**
Can see here
https://user-images.githubusercontent.com/43168857/142181141-e93871c0-8510-44f5-9206-abafb542afc8.mp4


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -4279,7 +4279,8 @@ TextureHandle ColorGradingPass(RenderGraph renderGraph)
             var currentGradingHash = ComputeLUTHash();
 
             // The lut we have already is ok.
-            if (currentGradingHash == m_LutHash)
+            if (currentGradingHash == m_LutHash &&
```
JulienIgnace-Unity, line 4282: Does this mean that we recompute the LUT every frame whenever any parameter of the curve is overridden? I suppose this can actually happen quite often :/

FrancescoC-unity: Yes, but the LUT is cheaper to compute than the hash (and up until really recently it was computed every frame anyhow :P )

JulienIgnace-Unity (approved):

iM0ve (approved):
Thanks for adding the test showcase. I also verified with all the curves, was curious. No issues were found.

sebastienlagarde (approved):

PR #6324, opened 2021-11-17
Avoid having screen space overlays being rendered by the camera when HDR is off

FrancescoC-unity:
Partial fix for: https://fogbugz.unity3d.com/f/cases/1380194/

For HDR we _need_ to have it drawn by camera (or well.. we could if we accept something fairly slow, so we'll first wait to see if we get complaints), but we can revert the whatever the pipeline used to do before for non HDR cases.

I will add an entry in the HDR docs that are currently under review to specify this behaviour in HDR. 

**What did I test**: I used renderdoc :P But can check having no camera and UI, and see that it will render now and won't before (or in HDR mode)


alelievr (approved):

sebastienlagarde (approved):

sebastienlagarde:
For Qa: Merging this fix as there is no much to test outside of what Francesco test already

PR #6325, opened 2021-11-17
Fix debug window reset

adrien-de-tocqueville:
### Purpose of this PR

fix https://fogbugz.unity3d.com/f/cases/1381547/

When resetting the debug window and then creating a new camera, the values before the reset were restored.
Clearing all values when resetting fixes the issue

---
### Testing status

tested the repro steps


adrien-de-tocqueville:
revert commit, i pushed to the wrong branch

RSlysz (approved):
📗

iM0ve (approved):
I also tried the repro, no issues.

PR #6327, opened 2021-11-17
Fix nullref in non dev build

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1381556/

For some reason, the conditional was not working during menu registration, so i changed it to a good old ifdef

---
### Testing status

tested the repro steps


RemyUnity (approved):
Makes sense. LGTM

arttu-peltonen (approved):
Awesome, thanks for the fix. I have a feeling this will resolve my "PS4 non-dev black screen" issue as well (https://fogbugz.unity3d.com/f/cases/1380019/), will still verify if it does. 

The reason I changed over to the conditionals to begin with was that `Debug.isDebugBuild` is unreliable and shouldn't be used (may give wrong results after a build). Good to know that these conditionals shouldn't be used either :(

arttu-peltonen:
My best guess as to why ConditionalAttribute doesn't work here: The way ConditionalAttribute works is by stripping out all calls to the method that has this attribute. However, because this method is decorated with [RuntimeInitializeOnLoadMethod] attribute, it gets called through reflection and this call is not visible to the ConditionalAttribute mechanism and doesn't get stripped.

PR #6329, opened 2021-11-17
Fixed an issue causing Render Graph execution errors after a random amount of time.

JulienIgnace-Unity:
### Purpose of this PR
Fixed an issue causing Render Graph execution errors after a random amount of time.

This was caused by the render graph resource validity hash to collide with a reserved value for shared resources.
This caused resources to be considered valid in some piece of code even though they weren't. Causing issues when used later on.

Fixes https://fogbugz.unity3d.com/f/cases/1379350/

---
### Testing status
Tested the repro described in the fogbugz cases.


TomasKiniulis (approved):
Left my PC over night with project using PR branch. Exception does not show up. Doing a quick test with debug modes also does not invoke the error anymore.

FrancescoC-unity (approved):

PR #6334, opened 2021-11-18
[URP] RTHandles Load/Store actions optimizations fix

manuele-bonanno:
---
### Purpose of this PR
The recent RTHandle adoption disabled all the store action optimizations. It also highlighted some URP tech debt, since the load actions logic was not correct.

This PR fixes the store action optimization bug and rewrites the logic that sets up the scriptable render passes load actions.

---
### Testing status
Tested locally in Metal editor, iPhone11

---


sandy-carter-unity (approved):

jonuuukas (approved):

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -987,18 +987,6 @@ protected void SetupRenderPasses(in RenderingData renderingData)
...
                 rendererFeatures[i].SetupRenderPasses(this, in renderingData);
             }
-
-            // Remove any null render pass that might have been added by user by mistake
-            int count = activeRenderPassQueue.Count;
-            for (int i = count - 1; i >= 0; i--)
-            {
-                if (activeRenderPassQueue[i] == null)
-                    activeRenderPassQueue.RemoveAt(i);
-            }
```
phi-lira, line 997: Why do we remove this?

phi-lira (approved):
Missing changelogs.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
manuele-bonanno: this was the copy paste error in the RTHandles PR which disabled by mistake all the store actions optimizations (the same code is already in AddRenderPasses)

com.unity.render-pipelines.universal/Runtime/Passes/ColorGradingLutPass.cs
```diff
@@ -183,8 +183,10 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
 
                 renderingData.cameraData.xr.StopSinglePass(cmd);
 
+                cmd.SetGlobalTexture(ShaderPropertyId.sourceTex, m_InternalLut);
+                CoreUtils.SetRenderTarget(cmd, m_InternalLut, RenderBufferLoadAction.DontCare, RenderBufferStoreAction.Store, ClearFlag.None, Color.clear);
                 // Render the lut
-                cmd.Blit(null, m_InternalLut.nameID, material);
+                cmd.Blit(null, BuiltinRenderTextureType.CurrentActive, material, 0);
```
phi-lira, line 189: I'm confused why we still rely on BuiltinRenderTextureType after RTHandle. Why is that?

phi-lira (approved):

PR #6335, opened 2021-11-18
Remove RuntimeProfiler test to unblock srp2core

arttu-peltonen:
Disable runtimeprofiler test because it fails on Katana (passes on Yamato though).


arttu-peltonen:
Cherrypicked to srp2core, closing this PR. Proper fix is in https://github.com/Unity-Technologies/Graphics/pull/6347.

PR #6347, opened 2021-11-19
Disable RuntimeProfilerTests in batchmode due to unreliability

arttu-peltonen:
### Purpose of the PR

It turns out that while the RealtimeProfiler test passes in Yamato, it fails in Katana causing srp2core problems. This PR hopefully fixes that.

Almost all (if not all) test projects in graphics repo have been specified with "disablebatchmode", so this problem didn't surface there. 

The root of the problem is that skipping to next frame inside a test case using `yield return new WaitForEndOfFrame();` is not supported in batchmode - it hangs until timeout (3 minutes). I thought I fixed the issue by doing `yield return null;` instead but seems like that might not work correctly either. In local testing it turned out the editor is updating at 3000-4000FPS during the Katana tests which results in the backend giving "gpuFrameTime=0ms" which fails the test. Since this test is adequately covered by non-batchmode test projects in graphics repo, I have just disabled it in batchmode to bypass the problem.

Example failures:
https://katana.ds.unity3d.com/projects/Unity/builders/proj0-Test%20PlayModeTests%20-%20MacEditor%20%2810.13%29/builds/75668?unity_branch=graphics/vendoring/2021-11-17

https://katana.ds.unity3d.com/projects/Unity/builders/proj0-Test%20PlaymodeTests%20-%20Windows64Editor%20%28Isolated%20Packages%20Verified%29/builds/86787?unity_branch=graphics/vendoring/2021-11-17

---
### Testing status
Tested manually that the test gets disabled as intended when running in batchmode.

### DIFF
--- a/com.unity.render-pipelines.core/Tests/Runtime/RuntimeProfilerTests.cs
+++ b/com.unity.render-pipelines.core/Tests/Runtime/RuntimeProfilerTests.cs
@@ -24,6 +24,9 @@ namespace UnityEngine.Rendering.Tests
             if (!FrameTimingManager.IsFeatureEnabled())
                 Assert.Ignore("Frame timing stats are disabled in Player Settings, skipping test.");
 
+            if (Application.isBatchMode)
+                Assert.Ignore("Frame timing tests are not supported in batch mode, skipping test.");
+
             // HACK #1 - really shouldn't have to do this here, but previous tests are leaking gameobjects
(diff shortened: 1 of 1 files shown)


alex-vazquez (approved):

theo-pnv (approved):
Thanks! I will keep an eye on it when doing the next srp2core PR and report any issue.

PR #6351, opened 2021-11-22
[URP] Add missing material property wrappers

adrien-de-tocqueville:
### Purpose of this PR

For material variants, property wrappers have been added around material shader properties.
They are added by default when using materialeditor.ShaderProperty, but when doing custom GUI, they need to be added manually.
Note: some wrappers still don't work because they are handled by srp core, but they are fixed by a separate PR.

Also, upgraded default materials, since importer version was bumped

Warning: feature has just been merged to trunk, you'll need a recent version to test. For more info, you can find relevant doc here:
https://docs.google.com/document/d/1UfpSGfzHYa30s6BGrqQQpPrxOhs6VryWzpgOztCLpqs/edit#heading=h.rvidm87zd8ze

---
### Testing status

Manually tested by creating material variants from URP shaders
What needs to be tested is that when changing a property on a material variant with a URP shader, there is the white bar displayed on the left showing the property is overriden


ellioman (approved):

adrien-de-tocqueville:
Given how small the changes are, mabye there's no need for a QA pass on this PR ?
Also it was tested a around 6 month ago, and I don't think things have changed that much in UI code, so what do you think of merging the PR as is ?

simon-engelbrecht-soerensen (approved):
Tested that properties in URP shaders/material variants are highlighed withwhite when changed. Works!

phi-lira (approved):

PR #6357, opened 2021-11-22
[HDRP] Fix CoC size computation when dynamic resolution is enabled

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1382220/

CoC size computation (in pixels) was not reading the scaled image size when dynamic res is enabled, resulting in discrepancies when resolution changes.

---
### Testing status
Tested by enabling dynamic resolution on the game view camera at fixed 50% rate. Toggling dynamic res on and off does not result in popping anymore:

DoF_dynscale


FrancescoC-unity (approved):

TomasKiniulis (approved):
Testing status seems to be enough for this change. Looks good!

PR #6360, opened 2021-11-22
[RP Workflows] UX Improvements for Cross Pipeline Volumes

alex-vazquez:
### Purpose of this PR
- Fix: https://fogbugz.unity3d.com/f/cases/1382805/
- Volume Components from another pipeline on a Volume Profile are hidden now( https://jira.unity3d.com/browse/XPIPELINE-327)
- Volume Component Editor Attribute is deprecated, instead we use the official CustomEditor.
- Avoid a second refresh of the editors ( Init was generating the editors, and on the first OnGUI, they were cleared and build again)
- Help URL for volume components are only fetch for volume components on the list, avoiding a TypeCache look up.
- Only on Volume Profile Inspector we show the info message (https://jira.unity3d.com/browse/XPIPELINE-413) 

image

---
### Testing status
Changing RP, see that the Volume Components are changed.


com.unity.render-pipelines.core/Editor/Volume/VolumeComponentListEditor.cs
```diff
@@ -97,15 +102,7 @@ public void Init(VolumeProfile asset, SerializedObject serializedObject)
...
-            // Gets the list of all available component editors
-            var editorTypes = CoreUtils.GetAllTypesDerivedFrom<VolumeComponentEditor>()
-                .Where(
-                    t => t.IsDefined(typeof(VolumeComponentEditorAttribute), false)
-                    && !t.IsAbstract
-                );
-
-            // Map them to their corresponding component type
-            foreach (var editorType in editorTypes)
+            foreach (var editorType in TypeCache.GetTypesWithAttribute<VolumeComponentEditorAttribute>())
```
arttu-peltonen, line 105: You've removed the !IsAbstract requirement, do you think this attribute is never given to abstract classes?

arttu-peltonen (commented):
Looks good, just minor question.

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentListEditor.cs
alex-vazquez: I think it should never be given to abstract classes, but you are right better be safe. I have changed the VolumeComponentEditor for the official CustomEditor for Unity, in this way users are more used to that pattern and besides this we have all the checks that they are performing and the editor creation from trunk.

arttu-peltonen: Thanks. Now that this is a much bigger change, do you have any context for what the original motivation behind using the custom VolumeComponentEditorAttribute instead of CustomEditorAttribute? 

This should also get a full QA pass to avoid regressions.

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentEditor.cs
```diff
@@ -222,12 +204,13 @@ static void ReloadDecoratorTypes()
...
-            base.Repaint();
-            //if (m_Inspector != null) // Can happen in tests.
-                //m_Inspector.Repaint();
+            inspector?.Repaint();
+
             // Volume Component Editors can be shown in the ProjectSettings window (default volume profile)
             // This will force a repaint of the whole window, otherwise, additional properties highlight animation does not work properly.
             SettingsService.RepaintAllSettingsWindow();
+
+            base.Repaint();
```
arttu-peltonen, line 213: Looks like you've effectively enabled the code that was commented, could you open a little what happened here? Also the order of repaints is now different.

arttu-peltonen (approved):
I'm not familiar with CustomEditors, but to me this looks like a great change. I assume all of the removed code was redundant stuff that was needed because we were using the custom VolumeComponentEditorAttribute, and now that we are using CustomEditor, a lot of those removed things work by default. 

I'd still be interested in some historical info on why CustomEditor wasn't used in the first place. As long as this is well tested to avoid regressions, I'll happily approve :)

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentListEditor.cs
alex-vazquez: Basically, everything on Volumes is custom, because while ago when this was developed we didn't have this classes

com.unity.render-pipelines.core/Editor/Volume/VolumeProfileEditor.cs
```diff
@@ -23,6 +23,11 @@ public override void OnInspectorGUI()
         {
             serializedObject.Update();
             m_ComponentList.OnGUI();
+
+            EditorGUILayout.Space();
+            if (m_ComponentList.hasHiddenVolumeComponents)
+                EditorGUILayout.HelpBox("There are Volume Components that are hidden in this asset because they are incompatible with the current active render pipeline. Change the active Render Pipeline to see them.", MessageType.Info);
```
martint-unity, line 29:
```suggestion
                EditorGUILayout.HelpBox("There are Volume Components that are hidden in this asset because they are incompatible with the current active Render Pipeline. Change the active Render Pipeline to see them.", MessageType.Info);
```

com.unity.render-pipelines.core/Editor/Volume/VolumeComponentListEditor.cs
```diff
@@ -162,14 +142,18 @@ void CreateEditor(VolumeComponent component, SerializedProperty property, int in
                 m_Editors[index] = editor;
         }
 
+        int m_CurrentHashCode;
         void RefreshEditors()
         {
-            // Disable all editors first
-            foreach (var editor in m_Editors)
-                editor.OnDisable();
+            if (m_Editors.Count > 0)
```
martint-unity, line 148:
```suggestion
            if (m_Editors.Any())
```

martint-unity (approved):
Looks good. Please add URP QA

JulienIgnace-Unity (approved):

TomasKiniulis (changes requested):
I'm not sure I agree with some of the changes here:
> Volume Components from another pipeline on a Volume Profile are hidden now( https://jira.unity3d.com/browse/XPIPELINE-327)

This feels like choosing between bloated Volume overrides and completely hiding the data introducing workflow issues in cases when user would like to match post processing settings from one pipeline to another. 
As I understand this change was implemented to be consistent with Camera, Reflection Probe and Light srp specific properties being hidden depending on the active pipeline, but this is thought to be changed, I think it would make more sense to move forward with hiding volume overrides once we have a way to filter out or show properties/overrides for only URP / only HDRP / both (/ none ?)?

> Only on Volume Profile Inspector we show the info message (https://jira.unity3d.com/browse/XPIPELINE-413)

I think Volume Profile is rarely directly selected to modify anything. So the info message can be rarely seen. Is there a reason not to display this on Volume Component as well?

TomasKiniulis (approved):
Listed my concerns above, besides this haven't ran into any technical issues. Feel free to merge.

PR #6361, opened 2021-11-22
Fix preview scene objects marked dirty by migration 

arttu-peltonen:
---
### Purpose of this PR
Fixes case [1367204](https://fogbugz.unity3d.com/f/cases/1367204/).

HDRP migration system marks GameObjects dirty whenever their components (e.g. `HDAdditionalLightData`) have received new versions and have gone through migration. This lets user know that they should save the scene (presumably to avoid executing migration every single time the scene is opened). 

This workflow seems fine, but it's problematic when a prefab requiring migration is used as "Editing Environment" prefab (see [documentation](https://docs.unity3d.com/Manual/EditingInPrefabMode.html)), meaning "a prefab that is used as the background context scene in Prefab Mode". In this situation, the editing environment scene is obviously not saved, and thus it also doesn't make sense to mark it dirty.

---
### Testing status
Retraced repro steps of [1367204](https://fogbugz.unity3d.com/f/cases/1367204/). Before the fix, asterisk appeared (briefly). After the fix, the asterisk no longer appears (and verified in debugger that the gameobjects with migrated components are not marked as dirty).

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed custom shader GUI for material inspector.
 - Fixed custom pass utils Blur and Copy functions in XR.
 - Fixed the ray tracing acceleration structure build marker not being included in the ray tracing stats (case 1379383).
+- Fixed objects belonging to preview scenes being marked as dirty during migration (case 1367204).
 
 ### Changed
 - Optimizations for the physically based depth of field.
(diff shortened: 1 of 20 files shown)


TomasKiniulis (approved):
Thanks for the fix! Also took a brief look if prefab is still dirtied after an actual modification is made and marked clean when saved. Looks good.

RSlysz (commented):
Have you tested that the prefab itself is dirty (meaning when you do save all assets, the serialization of the prefab is updated ?

arttu-peltonen:
> Have you tested that the prefab itself is dirty (meaning when you do save all assets, the serialization of the prefab is updated ?

Yes, the bug prefab was made on an earlier version so it's out of date and requires re-save, after which I saw asset version was bumped. Opening and saving the prefab bypassed the issue even without the fix.

RSlysz (approved):

arttu-peltonen:
Will retarget this to master. To me seems 10.x backport is needed, but 21.2 should be easy.

PR #6363, opened 2021-11-22
New SG integration VFX shaders where not visible in object picker

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1379523/

https://user-images.githubusercontent.com/85612173/142900642-520327bc-a132-4f6b-a17f-0f7734918663.mp4

This bug happens because of the combination of two changes:

1. The object picker now use a new search window (it was done for texture picking to include render textures)
2. The new ShaderGraph integration creates the shader asset a bit differently leading to expected asset not being listed in the search window (it's a bug in the search window that's will be fixed by the owner team)

In the meantime I fixed the issue with a workaround.

---
### Testing status
New SG assets (with the checkbox) are now listed
image
Old SG assets is still listed
image

---
### Comments to reviewers
⚠The issue is related to subassets, so a special attention should be given to cases with single or multiple targets. Because the old SG integration did not create subasset when there was only one target.

I took this opportunity to change a few other things:
- The new SG integration assets are now listed, but they are also sorted so that they appear at the top
- The texture picker now provides a dedicated tab for `TextureXD` and another one for `Render Texture`


VitaSkruibyte (approved):
Looking good! Couldn't find any outstanding issues :) Maybe one minor nitpicky note - our Shader Graphs no longer have proper icon in Search and Outputs (it used to display Scriptable Object icon instead of blank icon). 

Test summary:
1. With only one target (for example URP), multiple targets (URP HDRP), with Support VFX Graph checkbox checked- SGs are always shown in the search
2. Only Built- in target selected or Support VFX Graph unchecked - it is not shown in search
3. Tried hiding ShaderGraphs deeply in folder hierarchy - search found them anyway
4. Both old Shader Graph and New Shader Graph shaders are shown in Search
5. Different Outputs

PaulDemeulenaere (approved):
Looks good to me, the code review can't check the behavior but already manually verified by @VitaSkruibyte 👍 
You forgot to update the changelog (with the public reference issue https://issuetracker.unity3d.com/product/unity/issues/guid/1379523/)

PR #6365, opened 2021-11-23
Disable RadeonProDenoiser test

pigselated:
---
### Purpose of this PR
Disable unstable test. See See https://fogbugz.unity3d.com/f/cases/1383020/


KEngelstoft:
Look good, please create a new bug about this test being disabled.

pigselated:
> Look good, please create a new bug about this test being disabled.

Why? There is a bug on it being unstable is that not enough? https://fogbugz.unity3d.com/f/cases/1383020/

KEngelstoft (approved):

KEngelstoft:
We need a separate jira card or bug to track it according to our playbook, chapter 4.

pigselated:
> We need a separate jira card or bug to track it according to our playbook, chapter 4.

Should not be needed when immeditately disabling.

pigselated:
Investigated the failed builders. They are unrelated to this PR.

PR #6368, opened 2021-11-23
Ensure DebugFrameTiming doesn't allocate

arttu-peltonen:
### Purpose of this PR

Case [1382817](https://fogbugz.unity3d.com/f/cases/1382817/). Fix memory allocations by DebugFrameTiming. This is can fail gfx unit tests if you happen to have the Rendering Debugger window open. With this PR, that should no longer happen.

---
### Testing status
Tested locally according to repro steps in the ticket: In HDRP_Tests project, 1101_Unlit fails to GCAlloc issues before fix if Rendering Debugger is open, and passes afterwards.

**NOTE!** Seems like 1101_Unlit_Smooth_Distortion still fails to a different issue if you have Rendering Debugger open (though not 100% - sometimes only happens after you run multiple tests). This seems to be somehow related to FrameSettingsHistory and is a separate issue.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Debugging/FrameTiming/FrameTimeBottleneck.cs
+++ b/com.unity.render-pipelines.core/Runtime/Debugging/FrameTiming/FrameTimeBottleneck.cs
@@ -44,6 +44,8 @@ namespace UnityEngine.Rendering
 
         internal void DiscardOldSamples(int historySize)
         {
+            Debug.Assert(historySize > 0, "Invalid sampleHistorySize");
+
             while (m_Bottlenecks.Count >= historySize)
                 m_Bottlenecks.RemoveAt(0);
 
--- a/com.unity.render-pipelines.core/Runtime/Debugging/FrameTiming/FrameTimeSample.cs
+++ b/com.unity.render-pipelines.core/Runtime/Debugging/FrameTiming/FrameTimeSample.cs
@@ -117,6 +117,8 @@ namespace UnityEngine.Rendering
 
         internal void DiscardOldSamples(int sampleHistorySize)
         {
+            Debug.Assert(sampleHistorySize > 0, "Invalid sampleHistorySize");
+
             while (m_Samples.Count >= sampleHistorySize)
                 m_Samples.RemoveAt(0);
 


JulienIgnace-Unity (approved):

com.unity.render-pipelines.core/Runtime/Debugging/FrameTiming/FrameTimeBottleneck.cs
```diff
@@ -33,6 +33,11 @@ internal struct BottleneckHistogram
     /// </summary>
     internal class BottleneckHistory
     {
+        public BottleneckHistory(int initialCapacity)
```
alex-vazquez, line 36: It seems to me that `BottleneckHistory` and `FrameTimeSampleHistory` are pretty the same. I would encourage you to do a base generic class.

```
internal class History<T>
{
 List<T> m_Samples = new();
 public History(int initialCapacity):
 public void DiscardOldSamples(int sampleHistorySize);
 public void Add(T sample);
 public void Clear();
}

class BottleneckHistory : History<PerformanceBottleneck> and FrameTimeSampleHistory : History<FrameTimeSample>

File: com.unity.render-pipelines.core/Runtime/Debugging/FrameTiming/FrameTimeSample.cs [hunk 2]
```diff
@@ -114,6 +119,8 @@ internal void DiscardOldSamples(int sampleHistorySize)
         {
             while (m_Samples.Count >= sampleHistorySize)
```
alex-vazquez, line 120: I would check the sampleHistorySize to have a correct value. As I might call it with -1 and get an infinte loop.

alex-vazquez (approved):
Looks good to me, if you want you can apply mi comments in another PR, and get the GC issue fixed as soon as possible.

com.unity.render-pipelines.core/Runtime/Debugging/FrameTiming/FrameTimeBottleneck.cs
arttu-peltonen: Yes, I had the same thought when making the same changes to both files, but decided to not tackle the issue right now :) I do agree this would make sense though. I'll make a note to come back to this later.

com.unity.render-pipelines.core/Runtime/Debugging/FrameTiming/FrameTimeSample.cs
arttu-peltonen: Good catch, I fixed this one right away.

PR #6370, opened 2021-11-23
[HDRP][DXR] Remove Custom Interpolators from SpeedTree8 Graph to support raytracing

remi-chapelain:
### Purpose of this PR
PR fixes the issue of black tree in the pathtracer. See GIF
b87541d8652dcc6055d3f38bd0b7c02e

The issue comes from the fact that the graph uses custom intepolators to pass data (normals, tangeants, bitangenats) through the vertex stage to the fragment stage. Custom intepolators are not supported in raytracing so it fails. Atfer talking with multiple people, no one saw the addition of using custom interpolators so they have been removed to support raytracing by default. 

Here's the old and new graph. 
**_Before_**
Before_Fix

**_After_**
image

---
### Testing status
- Added multiple trees in a scene and now they are rendering fine with the pathtracer. ✔️ 
- Did the same with Ray Traced Reflection. ✔️ 

---
### Comments to reviewers
- They are still some more issue related to LOD with pathtracer with this graph but this is beyond the scope of this PR and it will be discussed after 
- It's not supposed to change anything on the rasterization side. (We have a test in the HDRP_Runtime test with tree (005) so issue will be caught there if anything. (when yamato will be back up)


eturquin (approved):

TomasKiniulis (approved):
Thanks for the change! I tried the fix applied in Xpipeline Village and it worked nicely - as mentioned in the PR, had no additional issues on rasterization side and trees were properly shaded in pathtracing.

sebastienlagarde:
@remi-chapelain reverted the PR as something bad have happen with various error message and something about exceeding variant limit. I will redo one

sebastienlagarde:
https://github.com/Unity-Technologies/Graphics/pull/6397

PR #6373, opened 2021-11-23
[HDRP] Add FSR Sharpness Control

gmitrano-unity:
### Purpose of this PR
This change adds a new GUI control in HDRP's camera properties that
allows users to control the intensity of the sharpness pass used by
FSR. The sharpness intensity is specified as a value from 0.0 (least
sharp) to 1.0 (most sharp).

Adding a 0.0 -> 1.0 interface for sharpness required some special
mapping logic since FSR expects a non-linear sharpness value. This
logic was added to SRP Core as a small wrapper function called
"SetRcasConstantsLinear".

When the sharpness parameter is set to zero, the post processing logic
now disables the RCAS shader code to improve performance since it won't
make a visible difference.

Pipeline Configuration UI:
image

Camera Configuration UI: (FSR indentation level only visible when FSR is the current upscaling filter)
image

---
### Testing status
Manually tested that sharpness visually scales from 0.0 -> 1.0 and RCAS is turned off at 0.0.

---
### Tasks
- [x] Changelog
- [x] Docs


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Rendering.Drawers.cs
```diff
@@ -150,9 +154,14 @@ static void Drawer_Draw_DLSS_Section(SerializedHDCamera p, Editor owner)
...
                 }
             }
-
 #endif
+            static void Drawer_Draw_FSR_Section(SerializedHDCamera p, Editor owner)
+            {
+                if (!p.allowDynamicResolution.boolValue)
+                    return;
 
+                EditorGUILayout.PropertyField(p.fsrSharpness, HDRenderPipelineUI.Styles.fsrSharpnessText);
```
FrancescoC-unity, line 163: We should enable this to be visible only when FSR is the chosen filter, otherwise most of the time would just be visual noise. 

I'd argue this should be part of the dynamic resolution settings altogether.

kecho, line 163: We should add a property as well in the quality assert settings, and then an override in the camera. 
If you want to stick to a property in the camera, i'd say you either hide it if the quality asset has a different FSR method, or put a note in the camera that says its currently disabled.

kecho (changes requested):
* Everything looks nice, only recommendation is to either put a note in the camera setting telling the user FSR is not enabled (or shade it gray). Or hide the setting completely, just like we are doing with DLSS. Or make it the same style of everything else, setting is in quality asset, camera has overrides (this is what I guess the team would probably ask for since its consistent) The problem at hand is to not clutter the camera UI with options, only when needed.

* Run yamato to sanity check
* Ensure the new function you added on core doesn't break URP, seams like its not called and is also I dont see any side effects. Still sanity check URP CI and we should be good.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Camera/HDCameraUI.Rendering.Drawers.cs
gmitrano-unity: After looking into GlobalDynamicResolutionSettings and reading @kecho's comments, I guess I missed the pipeline level setting for DLSS and incorrectly assumed everything was set per camera. I'll rework this change to make it work in the pipeline setting + camera override model.

gmitrano-unity:
The most recent commit that changes back to a slider control requires the fix from #6390 before it can be merged.

JulienIgnace-Unity:
@kecho @FrancescoC-unity is this PR ok in this state?
@gmitrano-unity can you add a line in the changelog file please?

FrancescoC-unity (approved):
LGTM now

RemyUnity (approved):
Tested on my Mac, everything seems to work fine.

PR #6380, opened 2021-11-24
[HDRP] Fix rasterized accumulation motion blur when DoF is enabled

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1378497/
We were getting artifacts with rasterization and accumulation motion blur when DoF was enabled, because we were applying post-process DoF only at the final accumulated frame (using the depth of the last frame).

 This PR changes the code a bit to grab the image that is going to be accumulated after DoF is applied.

---
### Testing status
Used the repro project.

Before:
image

After:
image

---
### Comments to reviewers
It would be nice to have a test for this, but still waiting for the infrastructure changes in our test framework to support accumulation.


TomasKiniulis (approved):
Did a quick check in repro project with multiple cameras. Motion Blur seems to show up properly now when recording.

JulienIgnace-Unity (approved):

eturquin (approved):

PR #6383, opened 2021-11-24
[ShaderGraph][2020.3][Backport] Properly cast float properties in integer mode

jessebarker:
### Purpose of this PR
This PR fixes https://fogbugz.unity3d.com/f/cases/1332564/.  It is a backport of #5119.  Please see that PR for images/videos of before and after the fix is applied.

The property block was fixed in #4382, and this PR fixes the in-graph value.  The user's choice is still preserved, so going from Integer back to Float without changing the value will cause the fractional part to return.

---
### Testing status
Repro steps from the bug using the fractional node to demonstrate the correct preview.   As mentioned above, please see #5119 for illustrations (my testing results are identical to those).

---
### Comments to reviewers


Nightmask3 (approved):
LGTM!

esmelusina (approved):

PR #3040, opened 2021-01-08
Sg/swizzle node UI alignment 2

tracychen1234:
# **Please read the [Contributing guide](CONTRIBUTING.md) before making a PR.**

* Read the [Graphics repository & Yamato FAQ](http://go/graphics-yamato-faq).

### Checklist for PR maker
- [x] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [x] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR. If you do add documentation, make sure to add the relevant Graphics Docs team member as a reviewer of the PR. If you are not sure which person to add, see the [Docs team contacts sheet](https://docs.google.com/spreadsheets/d/1rgUWWgwLFEHIQ3Rz-LnK6PAKmbM49DZZ9al4hvnztOo/edit#gid=1058860420).
- [x] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Why is this PR needed, what hard problem is it solving/fixing?
Update current Swizzle node UI for better alignment with other graph tools.
image

---
### Testing status
Describe what manual/automated tests were performed for this PR
Tested with different vector size input values and mask input edge cases.
Succeed in both local Paly Mode/Edit Mode tests.

---
### Comments to reviewers
Notes for the reviewers you have assigned.
I wrongly reverted a merge commit in the previous sg/swizzle-node-ui-alignment [PR](https://github.com/Unity-Technologies/Graphics/pull/2897), which caused many change files and broke the branch. This is the new branch & PR with the intended changes. Apologies for any inconvenience caused!

PR #6388, opened 2021-11-24
[SG][10.x.x][2020.3] Backport of Swizzle node upgrade

Nightmask3:
Backport of PR: #3040

Adds the new swizzle node UI/UX to ShaderGraph in 2020.3 and the 10.x package.
 
Specifically this work was done to address this ISS backport request: https://fogbugz.unity3d.com/f/cases/1359480/

Validated that the bug behavior cannot be repro-ed anymore as the new node's UI completely changes the old way the node functioned and in doing so, fixes the bug. 

Validated that the new node handles upgrading from the older Swizzle node versions without any errors for some test cases.  Seems robust.


esmelusina (approved):
Looks fine-- would like to make sure this doesn't require an upgrade path for the node.

PR #6389, opened 2021-11-24
Fixed shader graph errors when disabling the bias on texture samplers.

anisunity:
When creating a shader graph node sampler and disabling the second set of macros are used instead of the initial ones which were not properly supported for ray tracing. This fixes the issue.

Code only change, as long as it passes the formatting we're good.


JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Small change. Bugfix verified by developer seems to be enough

PR #6394, opened 2021-11-25
Fix normal/albedo in AOVs list

RemyUnity:
Small doc fix where normal and albedo where swapped in the AOV table


RemyUnity:
Does this need a line in the changelog ?

sebastienlagarde (approved):

PR #6395, opened 2021-11-25
A couple of small fixes for APV

FrancescoC-unity:
@tlaedre Need to momentarily revert this change as it creates problem on new projects trying to set up. We'll look into this, I suspect due to some other changes since yours. 

Also another issue that spew errors when no probe volume is init, but is active.


sebastienlagarde (approved):

sebastienlagarde:
Merging as it is a revert a previous issue

PR #6402, opened 2021-11-25
OnPreprocessMaterialDescription have a priority that override default users implementations [Waiting C++ PR]

bastienunity:
---
### Purpose of this PR
Fixing the OnPreprocessMaterialDescription priority order to avoid overriding users' implementations that uses the default priority of 0.
This PR needs to wait for https://github.cds.internal.unity3d.com/unity/unity/pull/6889 to land on trunk before it can land here or the internal implementation will override this one.

The PR is also fixing the keyword used for glossiness which now seems to be smoothness and is adding the usage of the Shininess value from fbx material descriptions.

This is part of fixing fogbugz 1370665 - 1360618 - 1370645

---
### Testing status
This PR need to be tested on top of this editor revision https://github.cds.internal.unity3d.com/unity/unity/pull/6889 because it need internal changes to some AssetPostprocessor for the graphic repo changes to work.
What has changed and need testing:
1. 
- Using each render pipeline and fbx files with Arnold Surface, Autodesk Interactive, Physical Material (3dsmax) and standard shader.
- Importing such assets in Unity is using the active render pipeline and all materials looks similar to Maya or 3dsMax rendering without changing any import settings option.
2. 
- FBX with standard materials using Shininess in another DCC are imported in each render pipeline with a corresponding smoothness value (see https://fogbugz.unity3d.com/f/cases/1370645/ for an example asset with multiple Shininess values)

---


RSlysz (approved):
Thanks for the change :)

TomasKiniulis (commented):
Hey @bastienunity! Could you give a brief description of what has been tested? Also anything specific for QA to take a look at? As I understand it's mostly about making sure material property values do not change when material is upgraded from built-in or reimporting an FBX file?

ellioman (approved):

TomasKiniulis (approved):
Looks good, validated that fogbugz reports are indeed fixed now.

Tested fbx with standard shader in legacy and upgrading to HDRP/URP - properties are changed correctly now.

Created 3ds max Fbx files which use Autodesk Interactive, 3d Max Physcially Based and Arnold shaders. Properties appear to be shown with correct values when imported or changing pipeline and reimporting asset. Note that checking with Thomas Chollet it appears that not all shader properties can be handled in Unity when imported as there's no equivalent , so some properties are still left out with default values resulting in significant visual differences compared to what I see in 3ds max.

For example in 3ds Max Physically Based material these do not change:
- BumpMapStrength
- Coat Roughness
- Coating IOR
- Transparency Color
- Transparency Weight

Attaching test results.
3ds max view Top sphere - Autodesk Interactive, Middle - 3ds Max Physcially Based, Bottom - Arnold
image

Legacy import
image

Upgrade to URP
image

Upgrade to HDRP
image

sebastienlagarde:
all good, now waiting for C++ Pr to land

sebastienlagarde:
C++ PR have landed. Merging

PR #6408, opened 2021-11-26
[ANT-302] Replace s10s with s21s

theopnv:
### Purpose of this PR
https://jira.unity3d.com/browse/ANT-302

- [x] Compare results with nightly from same base commit + List findings in [this document](https://docs.google.com/document/d/1uyLlcf_0TZ4PEPJUJEaxYwJ9USdH33aDtaRF5UFLOAM/edit)

---

### DIFF
--- a/.yamato/_abv.yml
+++ b/.yamato/_abv.yml
@@ -7,54 +7,6 @@
 
 ---
 
-TMP_ALL_S21_trunk:
-    name: TMP ALL S21 on trunk
-    variables:
-        CUSTOM_REVISION: '{{trunk.changeset.id}}'
-        UTR_VERSION: "current"
-        TEST_FILTER: .*
-        CACHE_ARGS: '{{cache.flags}}'
-    dependencies:
-      -  path: .yamato/urp_terrain-android-vulkan.yml#URP_Terrain_Android_Vulkan_Standalone_XR_il2cpp_Linear_trunk
-         rerun: on_new_revision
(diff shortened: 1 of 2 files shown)


theo-pnv:
Comparison of failing tests with master: https://docs.google.com/document/d/1uyLlcf_0TZ4PEPJUJEaxYwJ9USdH33aDtaRF5UFLOAM/edit?usp=sharing
All failing jobs are matching master (from november 26th), or are not run on master.

.yamato/_abv.yml
```diff
@@ -7,6 +7,54 @@
 
 ---
 
+TMP_ALL_S21_trunk:
```
theo-pnv, line 10: TODO before merging: Remove TMP job

tomzig16 (approved):

sophiaaar (approved):

theo-pnv:
On 10xx and 7xx we are still using hikeys (I think), since we're using `Unity::mobile::android`. Is that intenteded/is it planed to replace these with samsung devices (S21s?) @tomzig16

tomzig16:
> I'm planning the backports for this PR, and realized on 10xx and 7xx we are still using hikeys (I think), since we're using `Unity::mobile::android`. Is that intenteded/is it planed to replace these with samsung devices (S21s?) @tomzig16

I had left hikeys as we used to have small amount of the devices at the time of making the PR. Now if we have enough s21s we could move, but I believe there might be a reference image failure or two. When I was moving from hikeys it was a failure in terrain test on vulkan due to small rasterization difference which was solved by updating reference images so I'd expect something similar on these backports. @theo-pnv

PR #6411, opened 2021-11-26
IET Updates (version bump, 1080p Layout fix, fixed typo)

pierre-unity:
### Purpose of this PR
- Upgrade to IET framework version 2.1.1 (was 2.1)
- Fixed typo in Volume tutorial
- Fixed 1080p layout issue (too narrow panel)
- Improved a link leading to doc (using #)
- Fixed unnecessary indentation for Fog section

1080p:
image

4K:
image

### Comments to reviewers
Test that the layout looks fine at 1080p and 4K

### DIFF
--- a/com.unity.template-hd/Assets/SampleSceneAssets/Tutorials/Settings/Tutorials.asset
+++ b/com.unity.template-hd/Assets/SampleSceneAssets/Tutorials/Settings/Tutorials.asset
@@ -95,7 +95,7 @@ MonoBehaviour:
     Heading:
       m_Untranslated: Local Volumetric Fog
     Text:
-      m_Untranslated: ' Learn how to control the Volumetric Fog in specific areas'
+      m_Untranslated: Learn how to control the Volumetric Fog in specific areas
     Metadata: 
     Url: 
     Image: {fileID: 2800000, guid: 78a3d6c49f1c2e74c95bd052a569cc08, type: 3}


iM0ve (changes requested):
Its good with a few small nit picks:
* Materials **Bush_01_Mat**  and **Fern_01_Mat** have their height map amplitude changed from 0.01 to 0.02. But in the scene thyre not using any height map so I dont see any difference. Was this an intentional change or maybe something with upgrade system? Does it have some effect and I missed it? Do we want to revert these files if its accidental? We could merge as is, just want to understand.
* Its not introduced with this PR, but maybe we could also change the indentation of the tutorial Volumetric Fog tutorial with this PR?

Volumetric fog
Screenshot 2021-11-29 at 14 18 58

General result(good)
Screenshot 2021-11-29 at 13 59 24

We could merge as is but we can consider the optional fixes

pierre-unity:
> Its good with a few small nit picks:
> 
> * Materials **Bush_01_Mat**  and **Fern_01_Mat** have their height map amplitude changed from **0.01 to 0.02**. But in the scene they're not using any height map so I dont see any difference. Was this an intentional change or maybe something with upgrade system? Does it have some effect and I missed it? Do we want to revert these files if its accidental? We could merge as is, just want to understand.
> * Its not introduced with this PR, but maybe we could also change the indentation of the tutorial Volumetric Fog tutorial with this PR?
> 
> Volumetric fog Screenshot 2021-11-29 at 14 18 58
> 
> General result(good) Screenshot 2021-11-29 at 13 59 24
> 
> We could merge as is but we can consider the optional fixes

Thanks, I've now fixed the indentation issue. For the 0.02 amplitude, I'll leave it as it is for the template, as it doesn't alter visuals.

iM0ve (approved):

PR #6413, opened 2021-11-28
Modify the banding test to try to avoid the crashes.

anisunity:
We are often getting crashs related to the volumetric clouds banding test. I am trying to modify the test to see if we can avoid those crashes. Maybe the pow, 100 which is OOF plus the HDRI. not sure what is causing the issue trying stuff.

Also if not crashing this will require updating test screenshots.


anisunity:
Seems to not be enough, i am gonna strip the pow, maybe it helps?

PR #6415, opened 2021-11-29
APV Baking States

adrien-de-tocqueville:
### Purpose of this PR

This PR adds baking states for APV. It allows to bake lighting in different scene configurations that can then be used at runtime when needed.

First here's as gif showing how it works. Note that there's a script changing the lights when the state is changed
apv-states

New UI: Total number of states is fixed to 16, and the names are shared between sets, but the content can be baked separately, and some sets can choose to use only some of the states
image

Added new api in `ProbeReferenceVolume`
* `bakingState` to set/get the current active state
* `onBakingStateChanged`, a delegate so that scripts can register to state change

I also changed a bit the logic when selecting "Generate Lighting > Bake the set", which now doesn't require saving and reloading the scenes if the scenes in the set are already currently loaded.
Also "clear baked data" now delete the asset on disk

Notes:
* To detect which bake is out of date, it relies on a hash compute during baking. It will need to be updated if we add more paraemters in the future

---
### Testing status

Tested baking multiple states, also tested in a setup with multiple sets

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneData.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneData.cs
@@ -179,6 +179,9 @@ namespace UnityEngine.Experimental.Rendering
                 sceneBakingSettings.Add(settingsItem.sceneGUID, settingsItem.settings);
             }
 
+            if (string.IsNullOrEmpty(m_BakingState))
+                m_BakingState = ProbeReferenceVolume.defaultBakingState;
+
             foreach (var set in serializedBakingSets)
                 bakingSets.Add(set);
         }
--- a/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
+++ b/com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeReferenceVolume.cs
@@ -728,14 +728,14 @@ namespace UnityEngine.Experimental.Rendering
 
 #if UNITY_EDITOR
             if (sceneData != null)
-            {
                 UnityEditor.SceneManagement.EditorSceneManager.sceneSaved += sceneData.OnSceneSaved;
-                foreach (var data in ProbeReferenceVolume.instance.perSceneDataList)
-                    data.SetBakingState(sceneData.bakingState);
-            }
             AdditionalGIBakeRequestsManager.instance.Init();
 #endif
             m_EnabledBySRP = true;
+
+            if (sceneData != null)
+                foreach (var data in ProbeReferenceVolume.instance.perSceneDataList)
+                    data.SetBakingState(sceneData.bakingState);
         }
 
         /// <summary>


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -454,7 +452,7 @@ static void OnAdditionalProbesBakeCompleted()
             m_BakedCells.Clear();
 
             // Clear baked data
-            Clear();
+            ProbeReferenceVolume.instance.Clear();
```
FrancescoC-unity, line 455: Why changed to this clear? We nmeed to invalidate data on probe volumes and per scene data too.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -554,16 +552,13 @@ static void OnAdditionalProbesBakeCompleted()
             // Reset index
             UnityEditor.Experimental.Lightmapping.SetAdditionalBakedProbes(m_BakingBatch.index, null);
 
-            // Map from each scene to an existing reference volume
+            // Map from each scene to an existing reference volume, and each reference volume to a new asset
```
FrancescoC-unity, line 555: There is no such thing as multiple reference volumes

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -554,16 +552,13 @@ static void OnAdditionalProbesBakeCompleted()
...
             var scene2Data = new Dictionary<Scene, ProbeVolumePerSceneData>();
-            foreach (var data in GameObject.FindObjectsOfType<ProbeVolumePerSceneData>())
-                scene2Data[data.gameObject.scene] = data;
-
-            // Map from each reference volume to its asset
             var data2Asset = new Dictionary<ProbeVolumePerSceneData, ProbeVolumeAsset>();
-            foreach (var data in scene2Data.Values)
+            foreach (var data in GameObject.FindObjectsOfType<ProbeVolumePerSceneData>())
```
FrancescoC-unity, line 558: This would need to be called once and cached, at the very least for clarity.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -67,10 +58,23 @@ public void OnBeforeSerialize()
...
         }
 
-        internal void StoreAssetForState(ProbeVolumeState state, ProbeVolumeAsset asset)
+#if UNITY_EDITOR
+        internal ProbeVolumeAsset CreateAssetForCurrentState()
+        {
+            var asset = ProbeVolumeAsset.CreateAsset(gameObject.scene, m_CurrentState);
+
+            assets[m_CurrentState] = asset;
+            QueueAssetLoading();
```
FrancescoC-unity, line 67: This is not necessarily something we want. Why change the previous logic that will enqueue asset loading upon state change?

FrancescoC-unity:
There are a bunch of things we need to discuss about this actually before I go on with the review, give me a ping when you around for a call

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -554,16 +552,13 @@ static void OnAdditionalProbesBakeCompleted()
...
             {
-                data2Asset[data] = ProbeVolumeAsset.CreateAsset(data.gameObject.scene);
+                scene2Data[data.gameObject.scene] = data;
+                data2Asset[data] = data.CreateAssetForCurrentState();
```
FrancescoC-unity, line 561: This also has other side effects, why changing from what was there before? Responsability of creating the asset should not be on the data itself.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeBakingWindow.cs
```diff
@@ -464,19 +635,37 @@ void DrawBakeButton()
...
-                return;
-            }
-
             var scenesToRestore = new List<string>();
             for (int i = 0; i < EditorSceneManager.sceneCount; i++)
                 scenesToRestore.Add(EditorSceneManager.GetSceneAt(i).path);
 
-            // First, load all the scenes
-            LoadScenesInBakingSet(set);
+            bool sceneSetChanged = scenesToRestore.Count != set.sceneGUIDs.Count;
```
FrancescoC-unity, line 642: What happens if I am baking a different set with the same count of scenes of the ones that I am trying to unload?

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -126,19 +129,16 @@ void OnDestroy()
...
-            {
-                if (assets.ContainsKey(m_PreviousState) && assets[m_PreviousState] != null)
-                    ProbeReferenceVolume.instance.AddPendingAssetRemoval(assets[m_PreviousState]);
+            if (assets.ContainsKey(m_CurrentState) && assets[m_CurrentState] != null)
+                ProbeReferenceVolume.instance.AddPendingAssetRemoval(assets[m_CurrentState]);
 
-                QueueAssetLoading();
-            }
+            m_CurrentState = state;
+            QueueAssetLoading();
```
FrancescoC-unity, line 141: We def. don't want to queue an asset loading if the state is the same.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -25,16 +17,15 @@ internal class ProbeVolumePerSceneData : MonoBehaviour, ISerializationCallbackRe
         [System.Serializable]
         struct SerializableAssetItem
         {
-            [SerializeField] public ProbeVolumeState state;
+            [SerializeField] public int state;
             [SerializeField] public ProbeVolumeAsset asset;
         }
 
-        internal Dictionary<ProbeVolumeState, ProbeVolumeAsset> assets = new Dictionary<ProbeVolumeState, ProbeVolumeAsset>();
+        internal Dictionary<int, ProbeVolumeAsset> assets = new Dictionary<int, ProbeVolumeAsset>();
```
FrancescoC-unity, line 24: Being just ints, how does the user know what state is what?

FrancescoC-unity: We should have a way to query names a-la layers.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneData.cs
```diff
@@ -76,6 +77,31 @@ internal class BakingSet
...
+        public static int bakingState
+        {
+            get => ProbeReferenceVolume.instance.sceneData.m_BakingState;
+            set
+            {
+                var sceneData = ProbeReferenceVolume.instance.sceneData;
+                if (value == sceneData.m_BakingState)
+                    return;
+                sceneData.m_BakingState = value;
+                foreach (var data in Object.FindObjectsOfType<ProbeVolumePerSceneData>())
```
FrancescoC-unity, line 99: This is really bad to be called at runtime Object.FindObjectsOfType<ProbeVolumePerSceneData>(). 

We should have the per scene data register itself somewhere upon loading and then go through the registered ones.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneData.cs
```diff
@@ -230,6 +266,45 @@ internal BakingSet CreateNewBakingSet(string name)
             return set;
         }
 
+        internal void CreateNewBakingState(string name)
```
FrancescoC-unity, line 269: Note that a problem on having the baking states  globally like this is that it forces a given amount of states all with the same name for all the sets in the project. 

This might be relatively ok, but we need to think about the implications a bit.

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeSceneData.cs
```diff
@@ -230,6 +266,45 @@ internal BakingSet CreateNewBakingSet(string name)
...
+        {
+            bakingStates.Add(name);
+        }
+
+        internal void RemoveBakingState(int index)
```
FrancescoC-unity, line 274: So this reassigns indices to the right states which is fine, but what about lingering references to old states (just ints afterall)?

FrancescoC-unity:
Also adding @alelievr as there are touches to the baking window.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
adrien-de-tocqueville: Indeed there was an error that should be fixed by commit 668245b
My issue is that InvalidateAllAssets was clearing the assets map, so we would loose all references, now i only do AddPendingAssetRemoval.
In Clear though, i still clear the map, and additionally i also deleted the asset file on disk, i find that more logical

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
adrien-de-tocqueville: added a global cache registered with OnEnable/OnDisable

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
adrien-de-tocqueville: If you are just talking about QueueAssetLoading i guess i can remove it since it's called after anyway during baking

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
adrien-de-tocqueville: i thought it would make sense that the data does it since it manages the states, but i don't really mind, i reverted it to how it was before

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeBakingWindow.cs
adrien-de-tocqueville: theres a loop just below that compares the scenes, it just to avoid the loop in case the count is the same

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
adrien-de-tocqueville: there's a `if (state == m_CurrentState)` at the start of the function that early out in that case

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
adrien-de-tocqueville: Do they need to know ? they have a name in the UI that should be enough

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -67,20 +59,42 @@ public void OnBeforeSerialize()
...
+            AssetDatabase.StartAssetEditing();
+            foreach (var asset in assets)
+            {
+                if (asset.Value != null)
+                {
+                    ProbeReferenceVolume.instance.AddPendingAssetRemoval(asset.Value);
+                    AssetDatabase.DeleteAsset(ProbeVolumeAsset.GetPath(gameObject.scene, asset.Key, false));
+                }
+            }
+            AssetDatabase.StopAssetEditing();
```
FrancescoC-unity, line 85: Need ifdef unity editor

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumeAsset.cs
```diff
@@ -89,6 +82,42 @@ internal void StoreProfileData(ProbeReferenceVolumeProfile profile)
...
+            if (a.simplificationLevels != b.simplificationLevels) return false;
+            if (a.cells.Count != b.cells.Count) return false;
+            for (int i = 0; i < a.cells.Count; i++)
+            {
+                if (a.cells[i].position != b.cells[i].position) return false;
+                if (a.cells[i].minSubdiv != b.cells[i].minSubdiv) return false;
+                if (a.cells[i].indexChunkCount != b.cells[i].indexChunkCount) return false;
+                if (a.cells[i].shChunkCount != b.cells[i].shChunkCount) return false;
+                if (a.cells[i].bricks.Count != b.cells[i].bricks.Count) return false;
+                if (a.cells[i].probePositions.Length != b.cells[i].probePositions.Length) return false;
```
FrancescoC-unity, line 106: As we discuss this on slack reminder: hash of array when baking

alelievr (approved):

JMargevics:
Don't forget the changelog 👍

adrien-de-tocqueville:
if we duplicate a scene, the hidden PerSceneData gameobject gets duplicated and keeps references to the volume asset from the other scene. I don't know if it's really a problem, but we should probably fix that when we refactor the ProbeVolumeAsset serialization

FrancescoC-unity:
> if we duplicate a scene, the hidden PerSceneData gameobject gets duplicated and keeps references to the volume asset from the other scene. I don't know if it's really a problem, but we should probably fix that when we refactor the ProbeVolumeAsset serialization

Not really a problem since they will be treated as separate scenes (via GUID) when a new bake happens. In the meantime if old asset is kept it just shows "valid" data.

iM0ve (changes requested):
**Testing status:** In progress

**Issues:**
* Probes no longer work if the debug mode is off
* In general the there is a big visual difference when comparing baking on Master vs Branch


Master (expected result)

https://user-images.githubusercontent.com/36502659/144884034-860cacd1-83ba-4ab8-9fa9-249e3453b179.mov


PR Branch

https://user-images.githubusercontent.com/36502659/144884093-a159466e-c2ff-4763-96fd-aad0406ef61e.mov

adrien-de-tocqueville:
Since i renamed the baked data file, it would not load existing ones. Changed the data file name for the first baking state to be the same as the previous baked data, so that upgrading a project works correctly

alelievr (approved):
Question though, in the APV baking window screenshot is it expected that the table doesn't match the width of the window?

image

adrien-de-tocqueville:
@alelievr yes i hardcoded the width, otherwise it's too big and it's difficult to read

adrien-de-tocqueville:
Also I looked through the c++ code and i noticed they are thinkgs to detect when renderers that contribute to GI + their materials + lights + other things change, so in the future if we make them accessible to c# we could detect when a scene is baked, or if it's out of date

TomasKiniulis (changes requested):
Continuing on Motiejus testing. I seem to be running into lighting issues when switching between two states after I use same states in multiple Baking Sets. Probe volumes themselves remain unchanged but the lighting is way off. Maybe a potential bug to lighting team or not pr related, but the states feature reveal this.

Running into some UX issues where I seem to be not getting "Out Of Date" state usually when having multiple sets. Also loading scenes from another set does not load probes from state, but the state is highlighted. 

I also find it confusing switching states, but having to manually switch through volume lighting setups or different lighting scenes, would be good to have some sort of functionality directly in the ui to disable/enable specific volumes load/unload scenes, I guess that's something planned for the future?

Listed my findings in test doc: https://confluence.unity3d.com/display/HDRP/APV+Baking+State could you have a look? And let me know if more info is needed.

TomasKiniulis (approved):
Looks good! Issues listed in test doc are addressed for this PR.  Status seems to be properly shown after bake and clearing or not saving the scene. Baking States table is hidden when no Scene is assigned.

adrien-de-tocqueville:
UX is back to this point, where number of state is not fixed, and states can be different between two baking sets.
(Note however that they are identified by names, so if you switch between sets that have states with the same name, it will automatically be loaded)
image

adrien-de-tocqueville:
@FrancescoC-unity right now since the state name is appended to the baked data file name on disk, existing project will need to rebake. I didn't do a conversion step but i could do it if you think it's important

FrancescoC-unity:
> @FrancescoC-unity right now since the state name is appended to the baked data file name on disk, existing project will need to rebake. I didn't do a conversion step but i could do it if you think it's important

I think it's fine; we are still in experimental so it's fine breaking. Only thing is remember to ping the slack channel when this lands.

TomasKiniulis (approved):
Gave another look after the changes. Looks good! Haven't ran into any PR specific issues

FrancescoC-unity (approved):

iM0ve (approved):
Tomas tested the PR. Couldnt remove myself, approving.

PR #6182, opened 2021-11-02
Display diffusion profile overrides in volume debug menu

adrien-de-tocqueville:
### Purpose of this PR

Added new debug widgets for the debug menu to show what diffusion profiles are currently in use by the volume system.

At the same time, i made a slight change to the volume debug table to improve readability, that you can see in the first screenshot below. I changed the column header name to be only the profile name, and added a row to link to the gameobjects contaning the profile. (that was an artist feedback from a long time ago)

This is how they show in the volume debug menu in the editor:
_menu_editor

This is the same as above, but in play mode:
_menu_playmode

This is a test to show the look of the two new widgets, but outside of the volume debug menu table
_menu_editor2

---
### Testing status
See the screenshots
Tested in a standalone build

PR #5863, opened 2021-09-30
SRP Workflows - Volume Debug Panel for URP

alex-vazquez:
---
### Purpose of this PR
https://jira.unity3d.com/browse/XPIPELINE-264

This PR is including the Volume Debug Panel into Core and is being used on URP.

---
### Testing status
- Tested the Volume Debug Panel on Editor and on Runtime.
- Test that the volume components are being filtered by each pipeline.
HDRP
image
URP
image
- Parameters are being updated.
- Values are correct.
- Check unit tests are green
image

PR #6422, opened 2021-11-29
Remove volume name and fix formatting

adrien-de-tocqueville:
### Purpose of this PR

In #6182 i removed the volume name from the table header of the volume debug menu
Probably due to a merge issue it appeared back with #5863

Also fixed some formatting. It doesn't cause any error but it's not looking nice

I think the second PR mentionned landed before branch off, so this needs to be backported to 2022.1

### Testing status

tested that it compiled and volume debug menu has no volume name anymore


alex-vazquez (approved):
Looks good, thank you.

sebastienlagarde (approved):

PR #6423, opened 2021-11-29
Fix Scene View + Render Scale Logic

gmitrano-unity:
# Purpose of this PR
This commit reverses an unintentional change to the scene view's
handling of the render scale setting. This change was introduced
in #6274.

This commit also attempts to consolidate some of the render scale logic
in URP's code by modifying the shader constant logic to query the scaled
camera dimensions directly from the camera target descriptor rather than
calculating them again.

# Testing status
### What has been tested
- Tested locally that the scene view no longer scales with Render Scale on URP Viking Village
### What needs testing
- Test in a standalone build that render scale still works as intended. 
- Test both when rendering normally as well as having a camera render to a RenderTexture


ellioman (approved):

phi-lira (approved):

phi-lira:
Talked with @ernestasKupciunas , we will merge this one to staging and then test it there.

PR #6424, opened 2021-11-29
- Fixed the ray tracing fallbacks being broken since an Nvidia Driver Update.

anisunity:
After a recent update the saturate(x * FLT_MAX) is not reliable anymore for a fallback to derivative. A different variant of the same approach replaces saturate(x * 1e7) which make the tests pass locally on GPUs with the latest drivers.

Testing status
All the DXR tests are now passing locally.


sebastienlagarde (approved):

sebastienlagarde:
For SG team, this is only affecting raytrace part.

PR #6434, opened 2021-11-30
Removed Approximation Function Usage in 16-bit FSR

gmitrano-unity:
### Purpose of this PR
Usage of shader function approximations was causing a black screen on
Nvidia DX11 drivers. Even the higher precision variants also led to
visual artifacts so the approximations have been replaced completely in
the 16-bit path.

Strangely enough, these functions do not cause issues on the DX12 and
Vulkan paths, so there may be some shader compilation difference that we
need to address instead of changing this.

---
### Testing status
Tested locally on Nvidia DX11 driver


gmitrano-unity:
Closing this in favor of just disabling 16-bit for DX11 in URP. Devices that typically run DX11 aren't likely to see performance benefits from 16-bit anyways.

PR #6435, opened 2021-11-30
Probe interpolation for RenderBRG

kecho:
### Purpose of this PR
Enabling probe interpolation for the BRG through burst and compute.

---
### Testing status
Tested on Spaceship demo, hdrp template and on a custom level.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/BatchRenderer/BRGTransformUpdater.cs
+++ b/com.unity.render-pipelines.core/Runtime/BatchRenderer/BRGTransformUpdater.cs
@@ -3,6 +3,7 @@ using Unity.Burst;
 using Unity.Collections;
 using Unity.Mathematics;
 using System.Threading;
+using UnityEngine.Assertions;
 using Unity.Collections.LowLevel.Unsafe;
 using Unity.Jobs;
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.core/Runtime/BatchRenderer/BRGTransformUpdater.cs
```diff
@@ -137,63 +210,110 @@ public void Execute(int index, TransformAccess transform)
                     worldToLocal1 = new float4(mi.m11, mi.m21, mi.m02, mi.m12),
                     worldToLocal2 = new float4(mi.m22, mi.m03, mi.m13, mi.m23),
                 };
+
+                if (hasProbes[index])
+                {
+                    var shLp = new SphericalHarmonicsL2();
+                    var occlusionProbe = new Vector4();
+                    lightProbesQuery.CalculateInterpolatedLightAndOcclusionProbe(
+                        transform.position, -1, //TODO: this can be speed up if we cache the resulting tetrahedron index.
```
stramit, line 219: Use the 'batch' version of this command and you can pass in the index ;) ;)

PR #6437, opened 2021-12-01
[2020.3 Backport] Make Android tests use windows images to build android player

ellioman:
# Purpose of this PR
Backport of #5713 & #6023

This PR makes android tests to use windows image to build the android player. Reason for that being Unity using old SDK from android image, which at the moment throws an error. We download SDK together with Unity, so there's no need to use that image. However, we still use the same android image for running tests as it is the easiest way to download ADB and other needed tools.


tomzig16 (approved):

phi-lira (approved):

ellioman:
The Android jobs have all passed so I'm merging this PR.

PR #6439, opened 2021-12-01
[VFX] New Timeline Integration

PaulDemeulenaere:
---
## Purpose of this PR
New timeline integration (with scrubbing🎉), see [JIRA](https://jira.unity3d.com/browse/GVFXG-147)

https://user-images.githubusercontent.com/22494706/144219409-8131b35f-c313-447d-8cf3-c18084499a58.mp4

---
## Testing status
- [Testing document](https://docs.google.com/document/d/1j0_Svr0tklg5D63jMfZT5OfjEy81JmZ3gzVwIepXxdY/edit)
- Add a Graphic Test
GraphicTest
- Yamato ⏳ (which already integrates legacy timeline integration)

Think to merge [internal update](https://github.cds.internal.unity3d.com/unity/ScriptableRenderPipelinePrivate/pull/373) of reference images as well.

---
## Comments to reviewers
▶️ See this whole [thread](https://unity.slack.com/archives/C02GXV7TLF7/p1633619468002500) on slack.

### Assignement of reviewers
@VladNeykov for QA, see [testing document]([Testing document](https://docs.google.com/document/d/1j0_Svr0tklg5D63jMfZT5OfjEy81JmZ3gzVwIepXxdY/edit))
@MarieGuff for artist validation
@JulienDidierjean for UX validation (see original [figma](https://www.figma.com/file/GXleGnWQTrzcZUfymRy3oo/VFX-clip?node-id=0%3A1))
@julienamsellem for IMGUI implementation (see [Editor](https://github.com/Unity-Technologies/Graphics/blob/7bce071428b69d56bfc02836b3ab126a7b63bf13/com.unity.visualeffectgraph/Editor/Utilities/Playables/VisualEffectControlPlayableAssetEditor.cs) and [Inspector](https://github.com/Unity-Technologies/Graphics/pull/6439/files#diff-00cc1e8173764ab4ed1f9915ba3afa7ec159827fd00c0702d62a7aa12c204057R1))
@julienf-unity for behavior implementation (see [Mixer](https://github.com/Unity-Technologies/Graphics/blob/7bce071428b69d56bfc02836b3ab126a7b63bf13/com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs))

### Scrubbing
The ability of scrubbing is the main addition (and the reason of this PR) in timeline integration.

https://user-images.githubusercontent.com/22494706/144306795-491fc428-8798-42cf-9d9f-dc6e64a2df2f.mp4

When scrubbing is enabled on a clip, "Reinit" option is forced to "On Enter And Exit Clip" because the simulation will only start at the beginning of the clip.
The start seed of the VFX will be overidden here and the prewarm option is available.

When scrubbing is disabled, the "Reinit" option becomes available to user.

### Reinit on Clip

https://user-images.githubusercontent.com/22494706/144723592-7de05a39-44ad-43f8-9d3a-dcfe7ba43403.mp4

When the scrubbing is disabled, the reinit entering and exiting the track is actually optionnal, the previous behavior wasn't calling any reinit. These Reinit aren't triggering any event or prewarm, it uses an [internal API](https://ono.unity3d.com/unity/unity/pull-request/133781/_/graphics/vfx/feature/timeline-scrubbing-minimal-api).

### Reinit on Track

https://user-images.githubusercontent.com/22494706/144723463-1899cb96-b798-4a4d-8fac-db0cdc122a73.mp4

The reinit option is available on track inspector (don’t be confused with the previous reinit on clip). It controls how the binding of the director affects the initial state of the VisualEffect.
The most obvious state is the default: Reinit while binding and unbinding. The timeline will reset the VFX before taking control over it and will reset it after modifying it.
The reinit on binding doesn’t send any initial event and doesn’t process the initial prewarming.
Actually, this option has been mainly added for backward compatibility with previous track behavior. We could potentially remove it from the user UX.

### Seed control
_seed_control
When scrubbing is enabled or reinit entering clip active, the seed must be controlled by the clip. It’s especially needed for scrubbing, the reseedOnPlay option is forced to false to avoid random result of everytime the clip is replayed.

### Prewarm

https://user-images.githubusercontent.com/22494706/144723691-83bc0528-9ac7-4bd2-b078-42911d999d8e.mp4

When the reinit entering clip is active or scrubbing enabled, the prewarm option is available, it allows the processing of several frames before the first display of the VFX.
You have to be aware that a great amount of frames to be processed during prewarm can lead to a CPU/GPU peak.
So far, the approach is pretty naïve but it could be improved later (probably with a way to save the VFX state in editor to avoid costly prewarming in runtime)

### Event Clip & Single Event

https://user-images.githubusercontent.com/22494706/144809983-3934565a-ee08-4ebe-86ad-cec688cd715d.mp4

With previous activation clip, there was only two events triggered for each clips respectively at the begin and end of the clip.
The new implementation allows more control over event sent, we are considering two kind of events:
- Clip Event: couple of events, we are expecting turn on/off event of a VFX machine.
- Single Event: unique event, we are expecting single burst system event here.

When scrubbing is enabled, there isn't any difference between clip event and single event, we could produce the same behavior of a clip event with two single events.
When scrubbing is disabled, clip events are important: we are considering them as a state, thus, if we are entering (whatever the direction), the enter is sent, when exiting, the exit event is sent. This is a behavior which was already supported by the previous activation clip.

Side note: When scrubbing is disabled, the single event are only sent while playing forward 🔽

https://user-images.githubusercontent.com/22494706/144818608-9bec630d-ba7c-48c6-addb-dc0c24857af8.mp4

### Event Attributes

https://user-images.githubusercontent.com/22494706/144723788-74030c55-edde-49ef-85c9-066cf19feb61.mp4

The previous activation clip was already providing a set of attributes to be send with enter and exit events. Now, any event is supporting eventAttributes. There are new behaviors specific to this new integration:
- The attribute name field is now a free text edition, it support custom attributes.
- The internal data uses a [SerializeReference](https://github.com/Unity-Technologies/Graphics/blob/vfx/feature/timeline-scrubbing/com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlPlayableBehaviour.cs#L26) to store the real type. Previously, everything was stored with float, it could lead to precision issue trying to assign a large integer.

### Limitations

**Multiple Track, Same VFX**
When Scrubbing is enabled, several timeline tracks can't be bound on the same VisualEffectComponent.
In that case, a warning in displayed in scene overlay.
image
However, this behavior is still allowed without scrubbing (and needed for backward compatibility).

**Maximum Scrubbing Time**
Internally, the scrubbing simulates the VFX with a constant fixed time step (we are using the [maxDeltaTime](https://docs.unity3d.com/2021.2/Documentation/ScriptReference/VFX.VFXManager-maxDeltaTime.html), see [here](https://github.com/Unity-Technologies/Graphics/blob/23e406bc9c43d39e13fd631c848d7acc43cac457/com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs#L322))
To avoid a GPU Hang because Unity will have to process too many frame, we assuming a maximum "exact" scrubbing time. If we are trying to simulate a longer duration, the fixed time step will be artificially increased. It can lead to unexpected results (see [here](https://unity.slack.com/archives/C02GXV7TLF7/p1635844164031100?thread_ts=1635351184.018700&cid=C02GXV7TLF7)).

https://user-images.githubusercontent.com/22494706/144308659-b9c74c35-5e2c-4de7-a737-0b1e14566735.mp4

**Other**
- Speed multiplier: The control of Shuriken supports the speed multiplier in clip, this capabality isn't listed in VFX Control Clip (yet)
- Timeline interaction: the drawing of clipEvent & singleEvent relies on [DrawBackground](https://docs.unity3d.com/Packages/com.unity.timeline@1.0/api/UnityEditor.Timeline.ClipEditor.html) any interactive edition isn't possible, we have to rely on inspector settings so far.
- Playing backward: Event if it technically works for reasonnable duration, our current implementation use a lot of simulate and can lead to high GPU peak, it's recommanded to nover play a VFX backward in runtime.

## Migration
The data structure completely changed between visual activation track and visual control clip.
There isn’t any dedicated migration mechanism in timeline. I firstly hacked it converting all track to new one during compilation but it isn’t recommended.
So, we are relying on a AssetProcessor to sanitize the old activation clip to timeline clip.
It should be silent for the user perspective but it assumes the upgrade of the visual effect graph package correctly reimport the timeline asset.
If something went wrong, an [error](https://github.com/Unity-Technologies/Graphics/blob/vfx/feature/timeline-scrubbing/com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrack.cs#L68) will be logged during timeline compilation.
What the [AssetPreprocessor](https://github.com/Unity-Technologies/Graphics/blob/vfx/feature/timeline-scrubbing/com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs#L126) does on VisualEffectActivation track is:
- Create a mirror control clip, same duration, same position.
- Set the track reinit mode to "None"
- Disable the scrubbing
- Disable any reinit on clip

▶️ The behavior will be the same than the previous activation clip.

See the Timeline asset in VFX_HDRP Graphic Test after migration, it hasn't be modified in the PR to cover the expected conversion.

https://user-images.githubusercontent.com/22494706/144753192-5b195321-9028-4e91-b489-acaef58e66b0.mp4

## Next steps
- **Documentation**: see this [discussion](https://unity.slack.com/archives/C02GXV7TLF7/p1638797628159400?thread_ts=1638797067.159200&cid=C02GXV7TLF7) & [JIRA](https://jira.unity3d.com/browse/DOCF-985)
- **Make public API**: VisualEffect component is now exposing two new internal API, `Reinit(false)` to reset a system with prewarming or sending initial event and an access to internal total time. Both API can be relevant for user and should be exposed with public documentation.
- **Prewarm**: The UX is pretty counterproductive, it requires serious attention (but the inspector code is shared between timeline and asset, I anticipated a future improvement of this)
- **The internal behavior of prewarm** can be really long to processed, it would be nice to have a cache system in editor to avoid processing simulate in runtime. It could also probably be a first step of "saved" state to improve scrubbing in really long timeline clip.
- **Follow timeline update:** We provided the best solution with the current timeline possibilities, this feature should follow potential timeline improvement, for instance, if we are able to use an `ui.toolkit` callback


com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
```diff
@@ -0,0 +1,760 @@
...
+            foreach (var attribute in attributes.content)
+            {
+                if (attribute != null && attribute.ApplyToVFX(vfxAttribute))
+                {
+                    anyRelevantAttribute = true;
+                }
+            }
+
+            if (!anyRelevantAttribute)
+                return null;
```
julienamsellem, line 473:
```csharp
if (!attributes.Any(x => x?.ApplyToVFX(vfxAttribute) == true)
{
  return null;
}
```

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
```diff
@@ -0,0 +1,760 @@
...
+        {
+            var events = VFXTimeSpaceHelper.GetEventNormalizedSpace(PlayableTimeSpace.Absolute, behavior);
+            foreach (var itEvent in events)
+            {
+                //Apply clamping on the fly
+                var absoluteTime = itEvent.time;
+                if (absoluteTime > behavior.clipEnd)
+                    absoluteTime = behavior.clipEnd;
+                if (absoluteTime < behavior.clipStart)
+                    absoluteTime = behavior.clipStart;
```
julienamsellem, line 488:
```csharp
var absoluteTime = Math.Max(behavior.clipStart, Math.Min(behavior.clipEnd, itEvent.time));
```

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
```diff
@@ -0,0 +1,760 @@
...
+
+#if UNITY_EDITOR
+            m_ClipState = null;
+#endif
+        }
+
+        private static double Abs(double a)
+        {
+            return a < 0.0 ? -a : a;
+        }
```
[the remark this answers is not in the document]
PaulDemeulenaere: Should use `Math.Abs`,  I was too focus on using `Mathf` to realize the System was already providing the approriate function 😅

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
```diff
@@ -0,0 +1,760 @@
...
+            if (m_LastChunk != currentChunkIndex)
+            {
+                if (IsTimeInChunk(playableTime, m_LastChunk))
+                    currentChunkIndex = m_LastChunk;
+            }
+
+            if (currentChunkIndex == kErrorIndex)
+            {
+                var startIndex = m_LastChunk != kErrorIndex ? (uint)m_LastEvent : 0u;
+                for (uint i = startIndex; i < startIndex + m_Chunks.Length; i++)
```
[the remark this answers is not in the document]
PaulDemeulenaere: Yes, it's expacted : we are starting at the last chunk but we will also check previous chunk if we have actually played backward.

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
```diff
@@ -0,0 +1,760 @@
...
+                {
+                    var actualIndex = (int)(i % m_Chunks.Length);
+                    if (IsTimeInChunk(playableTime, actualIndex))
+                    {
+                        currentChunkIndex = actualIndex;
+                        break;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
PaulDemeulenaere: It's correct, but we should always start testing the current chunk (and the following one if it fails), the most probable playing case is forward, but any jump in timeline is possible.

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlPlayableBehaviour.cs
```diff
@@ -0,0 +1,269 @@
...
+
+        private static IEnumerable<VisualEffectPlayableSerializedEvent> CollectClipEvents(VisualEffectControlClip source)
+        {
+            if (source.clipEvents != null)
+            {
+                foreach (var clip in source.clipEvents)
+                {
+                    var eventEnter = (VisualEffectPlayableSerializedEvent)clip.enter;
+                    var eventExit = (VisualEffectPlayableSerializedEvent)clip.exit;
+                    eventEnter.editorColor = eventExit.editorColor = clip.editorColor;
```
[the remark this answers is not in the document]
PaulDemeulenaere: What do you mean by `Will not pass formatter rules.` ?
> I would say that modifying properties of objects in a method that is called CollectClipEvents is not expected.

These are structures, we aren't modifying the object but a copy.

com.unity.visualeffectgraph/Editor/Inspector/VFXAssetEditor.cs
```diff
@@ -421,6 +421,96 @@ private string UpdateModeToString(VFXUpdateMode mode)
...
+    public static void PrewarmGenericInspector(SerializedObject resourceObject, SerializedProperty prewarmDeltaTime, SerializedProperty prewarmStepCount)
+    {
+        if (!prewarmDeltaTime.hasMultipleDifferentValues && !prewarmStepCount.hasMultipleDifferentValues)
+        {
+            var currentDeltaTime = prewarmDeltaTime.floatValue;
+            var currentStepCount = prewarmStepCount.intValue;
+            var currentTotalTime = currentDeltaTime * currentStepCount;
+            EditorGUI.BeginChangeCheck();
+            currentTotalTime = EditorGUILayout.FloatField(EditorGUIUtility.TrTextContent("PreWarm Total Time", "Sets the time in seconds to advance the current effect to when it is initially played. "), currentTotalTime);
+            if (EditorGUI.EndChangeCheck())
```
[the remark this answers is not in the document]
PaulDemeulenaere: `PrewarmGenericInspector` is a copy/past of the previous behavior, I won't modify it in this PR (but we should adress a general refactor of this later). Gathering in one section will sligthly change the computation.

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
PaulDemeulenaere: Fix at 536a79b4b1222fbdd419e286a49eded66f9bab6c

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrack.cs
```diff
@@ -0,0 +1,109 @@
...
+            behaviour.Init(this, reinitBinding, reinitUnbinding);
+            return mixer;
+        }
+
+        public override void GatherProperties(PlayableDirector director, IPropertyCollector driver)
+        {
+            VisualEffect trackBinding = director.GetGenericBinding(this) as VisualEffect;
+            if (trackBinding == null)
+                return;
+            base.GatherProperties(director, driver);
```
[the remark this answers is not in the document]
PaulDemeulenaere: Fixed at 244284401b15d8f151e10ab0b3b4ad00b6506311 (by the way, I should double check the need of this override)

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrack.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    Debug.LogErrorFormat("Unexpected clip type : {0} in timeline '{1}'", clip, UnityEditor.AssetDatabase.GetAssetPath(timelineAsset));
+                }
+#endif
+            }
+
+            var mixer = ScriptPlayable<VisualEffectControlTrackMixerBehaviour>.Create(graph, inputCount);
+            var behaviour = mixer.GetBehaviour();
+            var reinitBinding = false;
+            var reinitUnbinding = false;
```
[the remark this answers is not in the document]
PaulDemeulenaere: Simply iterations without clean, good catch 😅 
Fix at 652839c50864534be854d123a7c755ec72db8b44

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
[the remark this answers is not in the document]
PaulDemeulenaere: Fixed at db10bb62a8590e47344d2eeb0eb879b6601a18c3

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrack.cs
```diff
@@ -0,0 +1,109 @@
...
+                m_VFXVersion = kCurrentVersion;
+            }
+        }
+
+        public override Playable CreateTrackMixer(PlayableGraph graph, GameObject go, int inputCount)
+        {
+            foreach (var clip in GetClips())
+            {
+                var customClip = clip.asset as VisualEffectControlClip;
+                if (customClip != null)
```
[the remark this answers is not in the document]
PaulDemeulenaere: Fixed at 19210ab15b3ae2404d4270fed481ffa7ab1dc724

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrack.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    allClipAreControl = false;
+                    break;
+                }
+            }
+
+            if (allClipAreControl)
+            {
+                m_VFXVersion = kCurrentVersion;
+            }
```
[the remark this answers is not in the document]
PaulDemeulenaere: Fixed at ffbad321403a971291a9529317c29b04cca5afb9

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlClip.cs
```diff
@@ -0,0 +1,165 @@
...
+            }
+
+            if (clipEvents == null)
+                clipEvents = new List<ClipEvent>();
+            if (singleEvents == null)
+                singleEvents = new List<VisualEffectPlayableSerializedEvent>();
+
+            behaviour.clipEventsCount = (uint)clipEvents.Count;
+            behaviour.events = new VisualEffectPlayableSerializedEvent[clipEvents.Count * 2 + singleEvents.Count];
+            int cursor = 0;
```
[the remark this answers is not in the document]
PaulDemeulenaere: Fixed at 4301f13484483ecab1f813f89379654cfb6ab3e0

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
```diff
@@ -0,0 +1,188 @@
...
+                            exit = new VisualEffectPlayableSerializedEventNoColor()
+                            {
+                                name = (string)oldAsset.activationBehavior.onClipExit,
+                                time = 0.0,
+                                timeSpace = PlayableTimeSpace.BeforeClipEnd,
+                                eventAttributes = MigrateEventStateToAttributes(oldAsset.activationBehavior.clipExitEventAttributes)
+                            }
+                        }
+                    };
+                    newAsset.singleEvents = new List<VisualEffectPlayableSerializedEvent>();
```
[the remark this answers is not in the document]
PaulDemeulenaere: Fix at 057685cf8b6b64051e50f7100e91f8716b66e3d0 (it can be null after serilization btw, but it's safely handled afterwards)

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
PaulDemeulenaere: `Any ` would have break at the first false element, we have to apply all in this implementation.

julienamsellem: That's true, sorry for bad suggestion.
But this one should work as expected :)
```csharp
if (attributes.Count(x => x?.ApplyToVFX(vfxAttribute) == true) == 0)
{
  return null;
}
```

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlPlayableBehaviour.cs
```diff
@@ -0,0 +1,269 @@
...
+        [SerializeReference]
+        public EventAttribute[] content;
+    }
+
+    [Serializable]
+    abstract class EventAttribute
+    {
+        public ExposedProperty id;
+        public abstract bool ApplyToVFX(VFXEventAttribute eventAttribute);
+    }
```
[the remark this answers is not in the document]
PaulDemeulenaere: Fixed at cdd3a16b419bf934d69c7bc74c777fc3fdfab0e2

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
```diff
@@ -0,0 +1,188 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Playables;
+using UnityEngine.Timeline;
+using UnityEngine.VFX;
+
+namespace UnityEditor.VFX.Migration
+{
+    class ActivationToControlTrack : AssetPostprocessor
```
PaulDemeulenaere, line 12: Side note: We should investigate for a better solution than AssetPostprocessor, maybe an OnInvalidate on VisualEffectControlTrack  ?

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        static IEnumerable<VisualEffectControlTrack> GetOutOfDateControlTrack(TimelineAsset timeline)
+        {
+            foreach (var track in timeline.GetOutputTracks())
```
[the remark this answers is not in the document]
PaulDemeulenaere: Fixed at 466bbaaf70527734bd85723990ad00c2e77de5f7 (but see also this [comment](https://github.com/Unity-Technologies/Graphics/pull/6439/files/4414d8a02ca785d099cfb58139303392d803b604#r765877719), AssetPostprocessor isn't ideal.

com.unity.visualeffectgraph/Editor/Inspector/VFXAssetEditor.cs
```diff
@@ -421,6 +421,96 @@ private string UpdateModeToString(VFXUpdateMode mode)
...
+                }
+
+                if (currentDeltaTime > currentTotalTime)
+                {
+                    currentTotalTime = currentDeltaTime;
+                }
+
+                if (currentTotalTime != 0.0f)
+                {
+                    var candidateStepCount_A = Mathf.FloorToInt(currentTotalTime / currentDeltaTime);
```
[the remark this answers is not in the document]
PaulDemeulenaere: Same as [this](https://github.com/Unity-Technologies/Graphics/pull/6439#discussion_r763116327)
I won't modify this code in this PR (it's simply a move), it deserve to be addressed indepently anyway.

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrackMixerBehaviour.cs
PaulDemeulenaere: Fixed at 0745bd0bd500cd91d7ff2c340fe558c6084480ff

julienamsellem (approved):
All good from my point of view.
Thanks for this huge improvement 👍

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
PaulDemeulenaere: Discussiong with @bastienunity about potential better approach (prototyping [here](https://github.com/Unity-Technologies/Graphics/tree/vfx/feature/timeline-scrubbing-without-assetpreprocessor))

MarieGuff:
First test remarks:
-Typo inconsistency with other tracks
-We should automaticly have a clip when creating the VFX Track to save a few clicks
-Drag and drop on the right part creates the wrong control track

PaulDemeulenaere:
Corner case bug fix at this revision: https://github.com/Unity-Technologies/Graphics/pull/6439/commits/ff2fdcfe670fee1c7c89fa9670d36fdbafb7551c 🔧 

There was a case where we could send twice the same event with no scrubbing & playing backward.
**Before**
_before_fix_no_scrubbing

**After**
_after_fix_no_scrubbing

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
PaulDemeulenaere: All in all, about this ⏫, even if the upgrade solution is pretty naive in this PR, it can be reworked afterwards.

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrack.cs
```diff
@@ -44,6 +50,45 @@ protected override void OnBeforeTrackSerialize()
...
+                    //Copy Prewarm settings
+                    vfxClip.prewarm.eventName = m_CurrentVisualEffect.initialEventName;
+                    if (m_CurrentVisualEffect.visualEffectAsset != null)
+                    {
+                        using var resourceObject = new SerializedObject(m_CurrentVisualEffect.visualEffectAsset);
+                        var prewarmDeltaTime = resourceObject.FindProperty("m_Infos.m_PreWarmDeltaTime");
+                        var prewarmStepCount = resourceObject.FindProperty("m_Infos.m_PreWarmStepCount");
+                        if (prewarmDeltaTime != null && prewarmStepCount != null)
+                        {
+                            if (prewarmStepCount.uintValue != 0u)
```
julienamsellem, line 76: You can merge these two `if`

TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXCustomSpawnerRecordSpawnCount.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace UnityEditor.VFX.Test
+{
+    class VFXCustomSpawnerRecordSpawnCount : VFXSpawnerCallbacks
+    {
+        public class InputProperties
+        {
+        }
+
+        public static List<int> s_ReceivedSpawnCount = new List<int>();
```
julienamsellem, line 15: It seems to be used only in unit tests, but I feel that leaving this public static field writable is dangerous.

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlTrack.cs
PaulDemeulenaere: Fix at bf30d3ae887b9acdb15caa2e1d0377be06df43e0

TestProjects/VisualEffectGraph_HDRP/Assets/AllTests/Editor/Tests/VFXCustomSpawnerRecordSpawnCount.cs
PaulDemeulenaere: Fix (kind of) at f6f92e9ea7791284fc7bcf33fba078dfb52c58f1
I think I could create a common analyze helper for all these custom test spawner (using local delagate)

MarieGuff (approved):
Nice job! 
Tested on 2022.2.0a1 (2dd09014126d)

VladNeykov (approved):
No outstanding points, all findings on the [test doc](https://docs.google.com/document/d/1j0_Svr0tklg5D63jMfZT5OfjEy81JmZ3gzVwIepXxdY/edit) are addressed.

It's also really a game-changer when it comes to making short videos with the VFX Graph, makes timing things really a breeze.

Thanks for the feature! \o/

Hellhound testing demo:
https://user-images.githubusercontent.com/7230872/148532654-d39739d5-3c99-4bbf-8ebf-c6146c7d572a.mp4

com.unity.visualeffectgraph/Editor/Utilities/Playables/VisualEffectControlClipInspector.cs
```diff
@@ -0,0 +1,522 @@
...
+            m_ReoderableClipEvents.DoLayoutList();
+            m_ReoderableSingleEvents.DoLayoutList();
+            if (EditorGUI.EndChangeCheck())
+            {
+                serializedObject.ApplyModifiedProperties();
+
+#if FIX_DUPLICATE
+                //See case 1381005
+                //Special detection of duplicated referenced element due to manual duplicate in reordable list
+                //TODOPAUL, it doesn't work because we are missing this modification: https://unity.slack.com/archives/C06TQ9LFP/p1636737000313000
```
julienf-unity, line 491: Those "TODOPAUL" should not be shipped

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationDeprecated/VisualEffectActivationBehaviour.cs
```diff
@@ -0,0 +1,44 @@
...
+using UnityEngine.VFX;
+using UnityEngine.VFX.Utility;
+
+[Serializable]
+class VisualEffectActivationBehaviour : PlayableBehaviour
+{
+    [Serializable]
+    public enum AttributeType
+    {
+        //Actually enum values are synchronized with VFXValueType
```
julienf-unity, line 15: Why do we need another enum then? Seems potentially error prone to maintain

julienf-unity: Mmh it's in a "deprecated" file so probably no relevant anymore

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
julienf-unity: Well the refactored feature should be shipped along with its upgrade path. So not sure what you mean by reworked afterwards

julienf-unity (commented):
in progress

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationDeprecated/VisualEffectActivationBehaviour.cs
PaulDemeulenaere: I actually moved it in this PR to deprecated, it's needed to be able to deserialize old data before sanitization.

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
PaulDemeulenaere: We should be able to do the same without `AssetPostprocessor` (it's never a good thing to rely on this callback for sanitization).

There isn't any sanitize mecanism in Timeline, the usage of AssetPostprocessor is the simple and naïve solution, but not the best (because the callback can't be discarded afterwards, it can't be a long term solution).

A better solution would be to upgrade data directly in `OnValidate` of the track. The problem is, during the first invocation, all clips aren't loaded yet. It can't work out of the box (while PostProcessor is correctly called after loading):
image
There is potentially a trick forcing a loading (using [ForceReserializeAssets](https://docs.unity3d.com/ScriptReference/AssetDatabase.ForceReserializeAssets.html)) of child clip in track during OnValidate but we aren't completly sure it's legal and it doesn't cause side effect.

Side Node: Initially, we have considered to keep both integration but remove the old one from listing, the problem of this solution comes if we want to do the sanitization later, it's really complex to sanitize director component which are stored in the scene (the bindings are registering a type & guid). Furtheremore, as far as I know, there isn't any way to filter out an existing implemented track from UX.

PaulDemeulenaere: Adding an editor test to check migration at 4b74606e940db9ea34129bfc77579b40c2cb1e2b
image

com.unity.visualeffectgraph/Editor/Inspector/VFXAssetEditor.cs
```diff
@@ -421,6 +421,96 @@ private string UpdateModeToString(VFXUpdateMode mode)
 
     private static readonly float k_MinimalCommonDeltaTime = 1.0f / 800.0f;
 
+    public static void PrewarmGenericInspector(SerializedObject resourceObject, SerializedProperty prewarmDeltaTime, SerializedProperty prewarmStepCount)
```
julienf-unity, line 424: Not sure about the naming here. Wouldnt Display/UpdatePrewarmGUI be more explicit?

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
julienf-unity: If the issue is the additional callback, we can always use the already existing AssetPostProcessor we already have in the package

julienf-unity (commented):
How do we prevent the iteration count during scrubbing to potentially explode? There seem to be no consistency check on maxcrubtime and fixedstep. We should probably ensure users cannot put extreme values that will cause the editor to hang

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
PaulDemeulenaere: The problem isn't really the additional callback but the `AssetDatabase.LoadAssetAtPath<TimelineAsset>` which force the loading of every playable after import.
Additionnaly, it won't work if there isn't any import.

We are looking for two alternatives:
1. https://github.com/Unity-Technologies/Graphics/commit/d0da33a029af0f5caecd6da264962c41536e85bc <= Trying to switch the type from Awake in old VisualEffectActivationClip, it can be problematic from AssetDB due to the rentry from an OnValidate. However, it doesn't work so far and it looks hacky, need validation from scriptable & assetDB ⏳
2. https://github.com/Unity-Technologies/Graphics/commit/4d6a3c4d2410c99e8d4d7d964ed292c1edea5ad7 <= Trying to early detect if there is a VisualEffectActivationClip to prevent the loading of TimelineAsset but it doesn't work because `LoadAssetAtPath` doesn't list hidden asset ⏳

PaulDemeulenaere:
The invistagation for a better approach to handle migration is in progess, see this [conversation](https://github.com/Unity-Technologies/Graphics/pull/6439/files#r784682057)

com.unity.visualeffectgraph/Editor/Inspector/VFXAssetEditor.cs
PaulDemeulenaere: Fix at c42c384f33a0718ffd4a67f69be19be12d8a8083

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationDeprecated/VisualEffectActivationBehaviour.cs
PaulDemeulenaere: By the way, the new implementation doesn't require a serialization of an enum, it uses _SerializableReference_: https://github.com/Unity-Technologies/Graphics/blob/11af1e72761af53f27294b3ee93945156445b105/com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectControl/VisualEffectControlPlayableBehaviour.cs#L26

com.unity.visualeffectgraph/Runtime/Utilities/Playables/VisualEffectActivationTrackMigration.cs
PaulDemeulenaere: These prototypes are a deadend, did the minimal smart change at https://github.com/Unity-Technologies/Graphics/commit/11af1e72761af53f27294b3ee93945156445b105
Ideally, we should have a mecanism in Timeline to handle sanitization & upgrade of tracks.

PaulDemeulenaere:
The failure with `VFX_HDRP on OSX_Metal_playmode_mono_Linear on version trunk` is related to this pending PR https://github.com/Unity-Technologies/Graphics/pull/6972

PaulDemeulenaere:
Ready to be merged !

Manual check at 468235a78caa91c778d65917c2368885827db7df 🟢 

The documentation is still in [progress](https://docs.google.com/document/d/16yW4YHZUyyQn7p9KbRldbvjWQI_s1nLozFVNfsKwe-Q/edit) and will land later.

Known unrelated issue with reordable list (not blocking) : 
- Case [1399803](https://fogbugz.unity3d.com/f/cases/1399803/) : it causes an execption while inspecting a VFXControlClip, the [fix is in landing pattern](https://ono.unity3d.com/unity/unity/pull-request/138642/_/editor/fix-reorderable-list-element-height-callback), see also this [conversation](https://unity.slack.com/archives/CQTQ2ANDB/p1643702012668639). I've been involve in verification.

- Case [1381005](https://fogbugz.unity3d.com/f/cases/1381005/) : it caused inconsistent behavior with duplicate, which doesn't occur with `+` button [fix is in landing pattern](https://ono.unity3d.com/unity/unity/pull-request/136294/_/editor/fix-reorderable-list-context-menu-changed-callback) ⬇️ 

https://user-images.githubusercontent.com/22494706/153247738-71b7fbef-0b56-4bfd-af3f-c4986fd833bf.mp4

- Minor UX issue noticed by @MarieGuff ⬇️ (look at the size of the "Scrubbing Disabled" field in clip background)
VFX_Bug_ScrubTypo (1). I will address this issue later.

PR #6442, opened 2021-12-01
Decal Layers (Rendering Layers)

lukaschod:
#### Changing light layers backend into more general rendering layers
- Adding new texture _CameraRenderingLayersTexture into URP that contains screen space rendering layers of meshes.
- Adding new class DrawObjectsAndRenderingLayersPass that writes into _CameraRenderingLayersTexture as second render target.
- Adding new varation of functionality DepthNormalOnlyPass that allows writing into _CameraRenderingLayersTexture as second render target.
- Changing GBuffer to write into _CameraRenderingLayersTexture instead of light layers now.
- Added new interanl API for requesting rendering layers `ScriptableRendererFeature.RequireRenderingLayers(bool isDeferred, out RenderingLayerUtils.Event atEvent, out RenderingLayerUtils.MaskSize maskSize)`

#### Removing light layers concept from shader backend
Now all logic for comparing layers is done with full rendering layer mask in shaders (C# converts them during the property set). The benefits of this simplifies logic as there is no longer a need to transform rendering layers to light layers when comparing and also avoiding this with decal layers. This should not effect performance as shader always contained them as 32 bit integer. Also this opens a door in the future if we would decide to have rendering layers only in UX side too.

#### Adding internal API for renderer features to request _CameraRenderingLayersTexture
ScriptableRendererFeature will contain new method `internal override bool RequireRenderingLayers(bool isDeferred, out RenderingLayerUtils.Event atEvent, out RenderingLayerUtils.MaskSize maskSize)`. This will allow URP to know if to produce _CameraRenderingLayersTexture and also at what event. Depending on the event it will be produced either with DepthNormalOnlyPass/DrawObjectPass/GBufferPass.

#### Decal Layers
Decal Layers implemented with above features and optimized for tiled-renderers. In case of DBuffer DepthNormalOnlyPass used for rendering layers. For Screen Space Forward ForwardOpaquePass provides it and for deferred GBuffer provides it.

New Decal Layers option to control its support in renderer feature:
image

New property in Decal Projector to specify Decal Layer it belongs to:
image

---
### Testing status
- [x] Tested locally on simple project combinations of DecalLayer/LightLayer/Forward/Deferred/NativeRenderPass/GameView/SceneView/ScreenSpaceDecal/DBuffer/DepthNormalPass.
- [x] Added automated test 250 to test Rendering Layers. Basically render rendering layer at specific events.
image
- [ ] Green ABV.
- [ ] Regression Test.

### Tiny Presentation
https://docs.google.com/presentation/d/12FzT-Kec55c2IdIXXH24nkI7Ksmb7tzDGWLSDdAcflk/edit#slide=id.g11145254bef_0_43

### Technical Design Document
https://docs.google.com/document/d/1Vw6VkHaJkgwwi2_Rv_a98241OaFHsKtm72KuJsIGnJc/edit#

---
### Comments to reviewers
Currently there is idea for removing decal and light layers and keeping only rendering layers for UX too (There is more info in TDD).

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
@@ -731,13 +731,12 @@ namespace UnityEngine.Rendering.Universal
 
                 RenderingUtils.ReAllocateIfNeeded(ref renderingLayersTexture, renderingLayersDescriptor, FilterMode.Point, TextureWrapMode.Clamp, name: renderingLayersTextureName);
 
-                CommandBuffer cmd = CommandBufferPool.Get();
                 cmd.SetGlobalTexture(renderingLayersTexture.name, renderingLayersTexture.nameID);
                 cmd.SetGlobalInt(ShaderPropertyId.renderingLayerMaskSize, RenderingLayerUtils.GetBits(renderingLayersSize));
                 if (this.renderingModeActual == RenderingMode.Deferred) // As this is requested by render pass we still want to set it
                     cmd.SetGlobalTexture("_CameraRenderingLayersTexture", renderingLayersTexture.nameID);
                 context.ExecuteCommandBuffer(cmd);
-                CommandBufferPool.Release(cmd);
+                cmd.Clear();
             }
 
             // Allocate normal texture if used


com.unity.render-pipelines.universal/Editor/Decal/DecalPass.template
```diff
@@ -63,7 +63,9 @@ Pass
     #endif
     #if _RENDER_PASS_ENABLED
     #define GBUFFER3 0
```
lukaschod, line 65: Need better way to specify inputs

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesFunctions.hlsl
```diff
@@ -417,4 +417,21 @@ float2 GetNormalizedScreenSpaceUV(float4 positionCS)
...
 
+uint GetMeshRenderingLayer()
+{
+    return asuint(unity_RenderingLayer.x);
+}
+
+float EncodeMeshRenderingLayer(uint renderingLayer)
+{
+    // TODO: Expose as property for the size
+    return PackInt(renderingLayer, 16);
```
lukaschod, line 428: I am guessing adding propery that controls size is the best approach here

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -153,21 +153,27 @@ internal GraphicsFormat GetGBufferFormat(int index)
             else if (index == GBufferShadowMask) // Optional: shadow mask is outputed in mixed lighting subtractive mode for non-static meshes only
                 return GraphicsFormat.R8G8B8A8_UNorm;
             else if (index == GBufferRenderingLayers) // Optional: rendering layers is outputed when light layers are enabled (subset of rendering layers)
-                return GraphicsFormat.R8_UNorm;
+                return GraphicsFormat.R16_UNorm;
```
lukaschod, line 156: This needs to be controllable

com.unity.testing.urp/Scripts/Runtime/DrawRenderingLayers.shader
```diff
@@ -0,0 +1,40 @@
...
+            Name "DrawRenderingLayers"
+            ZTest Always
+            ZWrite Off
+            Cull Off
+
+            HLSLPROGRAM
+            #pragma vertex FullscreenVert
+            #pragma fragment Fragment
+            #pragma multi_compile _ _USE_DRAW_PROCEDURAL
+            #pragma enable_d3d11_debug_symbols
```
sebastienlagarde, line 19: don't forget to comment this enable_d3d11_debug_symbols

kaychang-unity:
Does ShaderPreprocessor.cs needs to be updated to strip out `_WRITE_RENDERING_LAYERS` ?

lukaschod:
> Does ShaderPreprocessor.cs needs to be updated to strip out `_WRITE_RENDERING_LAYERS` ?

Updated

kaychang-unity (approved):

rbechoe:
Was this ever fully implemented into Unity? I can't find it in the 2021 version(s).

PR #6446, opened 2021-12-01
fixed link to master stack hair

emilybrown1:
In response to https://fogbugz.unity3d.com/f/cases/1384638/

Master Nodes are now called Master Stacks but the link was still trying to point to master node. I have changed it to point to the correct page.


Vic-Cooper (approved):
Link fixed!

sebastienlagarde (approved):

PR #6447, opened 2021-12-01
Fixed error thrown when layered lit material has an invalid material type

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1384960/

ensure material type is an authorized value during the validation step
Tested the repro steps

NOTE: switching back to HDRP/Lit will not restore the original material type, but keep material type standard


sebastienlagarde (approved):

sebastienlagarde:
Qa: merging for 10.8.0 package

PR #6452, opened 2021-12-02
[HDRP][Docs] Fix missing code highlight in dynamic resolution docs

0lento:
### Purpose of this PR

Dynamic Resolution doc page is missing one highlight for the code where you can swap the upscale filter via scripting API. I'm doing this PR because I'm constantly skimming over this specific doc page and not catching that one line there as it now kind of blends with the rest of the text paragraphs. 

Aim of this PR is to make the command pop out more and be formatted like all other code lines in the docs.


FrancescoC-unity (approved):

sebastienlagarde (approved):
Thanks for it :)

PR #6454, opened 2021-12-02
Fixed a null texture error caused by lens flare pass

JulienIgnace-Unity:
### Purpose of this PR
Fixed a null texture error caused by lens flare pass.
The issue came from a missing resource read on the depth buffer, causing it to be released before the lens flare pass when rendering a preview camera.
This highlighted the lack of check in the render graph code when dereferencing a texture after it's been released.
Such a check is added in this PR. This highlighted another issue with custom pass executed after the prepass that could try to access motion vector buffers before it was created.

Fixes: https://fogbugz.unity3d.com/f/cases/1383604/

---
### Testing status
Ran all graphics tests and the fogbugz test case


sebastienlagarde (approved):

RemyUnity (approved):
Fix confirmed

PR #6456, opened 2021-12-02
[Universal] RenderPass RTHandles refactor

jonuuukas:
### Purpose of this PR
---

As we recently introduced RTHandles into URP this allows RenderPass API to get a lot more data from the targets of ScriptableRenderPasses. This PR targets and removes some of the code, that is no longer needed which would setup all the relevant information regarding target dimensions and etc.
Also removing some other unneeded code (like GBufferAttachmentIdentifiers).

Main thing to note - NativeRenderPass will not work if the pass still uses RenderTargetIdentifiers. As the API was (and still is) internal, this shouldn't cause any breakages.

### Testing status

---

Manually ran Foundation/Lighting tests with RenderPass enabled on Windows and MacEditor
Ran BoatAttack on iOS 

### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
```diff
@@ -423,12 +419,6 @@ internal bool IsRuntimeSupportedThisFrame()
                 };
             }
             this.DepthAttachmentHandle = this.DepthAttachment;
-#if ENABLE_VR && ENABLE_XR_MODULE
-            // In XR SinglePassInstance mode, the RTs are texture-array and all slices must be bound.
-            if (renderingData.cameraData.xr.enabled)
-                for (int i = 0; i < this.GbufferAttachmentIdentifiers.Length; ++i)
-                    this.GbufferAttachmentIdentifiers[i] = new RenderTargetIdentifier(this.GbufferAttachmentIdentifiers[i], 0, CubemapFace.Unknown, -1);
-#endif
```
manuele-bonanno, line 431: being this XR do we need any XR QA or specific testing to make sure we don't break anything? Or are Yamato XR tests enough?

com.unity.render-pipelines.universal/Runtime/Decal/DBuffer/DBufferRenderPass.cs
```diff
@@ -135,7 +135,7 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
 
                 if (isDeferred)
                 {
-                    cmd.SetGlobalTexture("_CameraNormalsTexture", deferredLights.GbufferAttachmentIdentifiers[deferredLights.GBufferNormalSmoothnessIndex]);
+                    cmd.SetGlobalTexture("_CameraNormalsTexture", deferredLights.GbufferAttachments[deferredLights.GBufferNormalSmoothnessIndex].nameID);
```
sandy-carter-unity, line 138: It's probably better to use the name directly
```
cmd.SetGlobalTexture(deferredLights.GbufferAttachments[deferredLights.GBufferNormalSmoothnessIndex].name, deferredLights.GbufferAttachments[deferredLights.GBufferNormalSmoothnessIndex].nameID);
```

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -339,41 +307,23 @@ internal void SetNativeRenderPassAttachmentList(ScriptableRenderPass renderPass,
...
                     var usesTargetTexture = cameraData.targetTexture != null;
-                    var depthOnly = renderPass.depthOnly || (usesTargetTexture && IsDepthOnlyRenderTexture(cameraData.targetTexture));
-                    // Offscreen depth-only cameras need this set explicitly
-                    if (depthOnly && usesTargetTexture)
+                    var depthOnly = (pass.colorAttachmentHandle.rt != null && IsDepthOnlyRenderTexture(pass.colorAttachmentHandle.rt)) || (usesTargetTexture && IsDepthOnlyRenderTexture(cameraData.targetTexture));
+
+                    int samples;
+                    RenderTargetIdentifier colorAttachmentTarget;
+                    // We are not rendering to Backbuffer so we have the RT and the information with it
+                    if (passColorAttachment.nameID != BuiltinRenderTextureType.CameraTarget)
```
sandy-carter-unity, line 315: You might have an issue with XR slice being -1 and not being caught here.

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -52,9 +52,9 @@ public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderin
             descriptor.graphicsFormat = isDepth ? GraphicsFormat.D32_SFloat_S8_UInt : GraphicsFormat.R32_SFloat;
             descriptor.msaaSamples = 1;
 #if UNITY_EDITOR
-            ConfigureTarget(destination, destination, GraphicsFormat.R32_SFloat, descriptor.width, descriptor.height, descriptor.msaaSamples);
+            ConfigureTarget(destination, destination);
```
sandy-carter-unity, line 55: Would need to revisit this. Again, unless doing backwards compatibility, destination should be an RTHandle of color XOR depth

sandy-carter-unity (changes requested):
I love seeing all these removed lines. It's getting simpler and better.

My main concern is what looks like code to accommodate RTHandles with Depth in the color attachment.
Just making sure that it's not to accommodate RTHandles with Color + Depth.
This should not happen with RTHandles, especially with the RTID backwards compatibility path disallowed for render pass.
Is it for the CopyDepth pass?

com.unity.render-pipelines.universal/Runtime/DeferredLights.cs
jonuuukas: maybe @sandy-carter-unity would know best, but from what i see - these identifiers are not used anymore, so didn't see point of keeping them. This was basically the one of two places were the gbuffer identifiers were used. I highly doubt we still need XR specific QA for this, i think yamato XR would catch it if this broke anything

com.unity.render-pipelines.universal/Runtime/Decal/DBuffer/DBufferRenderPass.cs
jonuuukas: i think Decals rely on _CameraNormalsTexture being there, so we bind GBuffer2 as that. This would require changes for Decal shaders and probably additional checks whether it's forward or deferred

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
jonuuukas: yeah, this is basically a hack to make dx11 work properly. In this case - depth is actually not used, but if we don't specify it - it will render to backbuffer and mess up all of the depth in further passes

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
jonuuukas: will change this to passColorAttachment != kCameraTarget. Should that take care of the check?

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
```diff
@@ -692,6 +634,16 @@ internal static int GetValidPassIndexCount(int[] array)
             return array.Length - 1;
         }
 
+        internal static RTHandle GetFirstAllocedRTHandle(ScriptableRenderPass pass)
+        {
```
manuele-bonanno, line 638: nitpick: I would use Allocated instead of Alloced because it sounds weird :)

manuele-bonanno (approved):
Nice!

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
bwrsandman: I don't think it will. You're better off using `new RenderTargetIdentifier(passColorAttachment.nameId, 0, slice=0)`

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
bwrsandman: Could you add a comment to that line saying this?

jonuuukas: sure, will do

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
jonuuukas: Done in eb5919e

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
jonuuukas: Done in eb5919e

com.unity.render-pipelines.universal/Runtime/NativeRenderPass.cs
jonuuukas: Done in eb5919e

sandy-carter-unity (approved):
LGTM

jonuuukas:
@Unity-Technologies/gfx-qa-urp - would it be possible to do a quick sanity check? as i tested this on BoatAttack on iOS, maybe running a  project or two on some other platform would be beneficial just to be safe

phi-lira (approved):
I only looked at yamato when approving, as there is enough coverage by other reviewers.

phi-lira:
@Unity-Technologies/gfx-qa-urp merging this without QA approval because it's going to staging PR.

PR #6460, opened 2021-12-02
Added urp-optimization.md from master.

oleks-k:
Added urp-optimization.md from master.

### DIFF
--- a/com.unity.render-pipelines.universal/Documentation~/urp-optimization.md
+++ b/com.unity.render-pipelines.universal/Documentation~/urp-optimization.md
@@ -1,0 +1,7 @@
+# Optimization
+
+This section contains information related to optimization and debugging.
+
+This section contains the following topics:
+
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/Documentation~/urp-optimization.md
```diff
@@ -0,0 +1,7 @@
+# Optimization
```
phi-lira, line 1: Should the section name be Optimization and Debugging? 
 It's weird to me it's just called optimization and there's no optimization material referred in this page yet.

oleks-k: The Rendering Debugger window has more features for optimization, rather than specifically for debugging, but I can add "Debugging" to the title of course.

phi-lira (approved):

PR #6461, opened 2021-12-02
[gpupipeline] deferred materials - diff between gpu deferred materials and gpudriven BRG branch

kecho:
### Purpose of this PR
First rough version of deferred materials + gpu driven rendering on HDRP.
PR Contains
* Full implementation of geometry pool
* Deferred material BRG class that builds draw calls

---
### Testing status

---
### Comments to reviewers


com.unity.render-pipelines.core/ShaderLibrary/UnityInstancing.hlsl
```diff
@@ -239,6 +239,11 @@
 
     #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/UnityDOTSInstancing.hlsl"
 
+    #if defined(UNITY_SETUP_INSTANCE_ID)
+        #undef UNITY_SETUP_INSTANCE_ID
+        #define UNITY_SETUP_INSTANCE_ID(input) { DEFAULT_UNITY_SETUP_INSTANCE_ID(input); SetupDOTSVisibleInstancingData(); }
```
JussiKnuuttila, line 244: This should probably be ifdeffed so that the `SetupDOTSVisibleInstancingData` is only called in DOTS_INSTANCING_ON variants.

com.unity.render-pipelines.core/ShaderLibrary/UnityDOTSInstancing.hlsl
```diff
@@ -150,31 +150,43 @@ static const uint kDOTSInstancingFlagCameraMotion     = (1 << 2); // Object uses
 static const uint kDOTSInstancingFlagHasPrevPosition  = (1 << 3); // Object has a separate previous frame position vertex streams (e.g. for deformed objects)
 static const uint kDOTSInstancingFlagMainLightEnabled = (1 << 4); // Object should receive direct lighting from the main light (e.g. light not baked into lightmap)
 
+static DOTSVisibleData unity_SampledDOTSVisibleData;
+
+void SetupDOTSVisibleInstancingData()
+{
+    unity_SampledDOTSVisibleData = unity_DOTSVisibleInstances[unity_InstanceID];
```
JussiKnuuttila, line 157: Have you checked the shader codegen produced by this approach compared to the previous version?

If the codegen is the same, then I think we can use this approach always. If the codegen is worse for some reason, then we could consider separating the regular `DOTS_INSTANCING_ON` variants from deferred material ones, and add some kind of macro here to use the static variable only in the latter case.

I would imagine the codegen of this approach should be good, but I think it would be a good idea to check since shader compilers can often be surprising. :)

JussiKnuuttila (approved):

theo-pnv:
Hi, is this the correct merge target? A lot of files that should not be updated are impacted by this PR (.yamato/*).

com.unity.render-pipelines.core/ShaderLibrary/UnityInstancing.hlsl
kecho: This is already in a block of code that has UNITY_DOTS_INSTANCING_ENABLED. But Ill add here here just to be more expressive :)

PR #6462, opened 2021-12-02
[Backport] Fixed NRE when debug mode is active and changing opened VFX asset

julienamsellem:
### Purpose of this PR
[Backport] https://fogbugz.unity3d.com/f/cases/1379676/
[Original] https://fogbugz.unity3d.com/f/cases/1347420

Fixed an exception with the following scenario
- Create two VFX assets
- Put one of them in the scene
- Open this VFX in the graph editor
- Attach the GO (if 2021.2 simply select it, it will auto-attach)
- Display the VFX Control
- Choose a debug mode (Efficiency for instance)
- Open the other VFX from the project view
=> Error messages in the console

See video below
uv6OTveKAX

---
### Testing status
Tested initial scenario
c79SdGzcdk
o

### DIFF
--- a/com.unity.visualeffectgraph/CHANGELOG.md
+++ b/com.unity.visualeffectgraph/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [12.1.3] - 2021-11-17
 ### Fixed
 - Automatically offset contexts when a new node is inserted to avoid overlapping
+- Fixed null reference exception when opening another VFX and a debug mode is enabled [Case 1347420](https://issuetracker.unity3d.com/product/unity/issues/guid/1347420/)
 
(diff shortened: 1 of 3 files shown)


PaulDemeulenaere (approved):
Looks good to me.
Can you just check one expected behavior (even if it's a backport):
If there are two components using the same asset in the scene, if we are debugging the particleCount, switching for a component to the other should keep the same debug state.
It's pretty convinient to inspect several component in the scene.

julienamsellem:
> Looks good to me. Can you just check one expected behavior (even if it's a backport): If there are two components using the same asset in the scene, if we are debugging the particleCount, switching for a component to the other should keep the same debug state. It's pretty convinient to inspect several component in the scene.

I checked that changing the attached GO (with same VFX asset of course) keeps the debug mode enabled. So the behavior is the same then before. But if you open an different VFX in the graph editor, then the debug mode is disabled.

PR #6201, opened 2021-11-04
[diedsoon] Transparent QoF (part 2): RemapAlpha, Node Scene Depth Difference, 

skhiat:
Remap Transparent Channel of the BaseColor/Color texture for {Lit, Unlit, LayeredLit, LitTesselation, LayeredLitTesselation}
Lit:
image

Unlit:
image

LayeredLit:
image

LitTesselation:
image

LayeredLitTesselation:
image

Add Scene Depth Difference for 3 format: {Linear01, Raw, Eye}:
image

Approved by UX - Maddelena

UX review notes: changes the order of parameters

https://jira.unity3d.com/browse/HDRP-1600
https://jira.unity3d.com/browse/HDRP-1602
https://jira.unity3d.com/browse/HDRP-1603

PR #6466, opened 2021-12-02
Transparent QoL part 1 - Refract & FresnelEquation Node

skhiat:
Note: already reviewed [PR - splitting ShaderGraph only code]

Transparent Quality Of Life (part 1)

- RefractNode:
Parametrize with eta which is the ratio of both IOR (Index Of Refraction) from incoming_medium (n_i) and outgoing_medium (n_o) eta == n_i/n_i. For instance from Air to Glass that would be 1.0/1.3333.
If it's air-to-air interaction eta == 1:
image
Otherwise:
Air-glass interaction:
image

Note: the parameter incident is a vector which is from "source-to-surface" (for instance from eye-to-pixel):
image

We have 2 mode:
- Safe: Protect against NaN when a sqrt of a negative number which happen when we have glass-air interaction, at a specific angle we have no refraction but pure reflection which should be handle by Fresnel.
- CriticalAngle: Not protected the user should care of the negative case.

Fresnel node:
image
Note: already reviewed [PR - splitting ShaderGraph only code]

Add Scene Depth Difference for 3 format: {Linear01, Raw, Eye}:
image

UX review notes: changes the order of parameters

https://jira.unity3d.com/browse/HDRP-1600
https://jira.unity3d.com/browse/HDRP-1602
https://jira.unity3d.com/browse/HDRP-1603

ref:
https://github.com/Unity-Technologies/Graphics/pull/6201

https://jira.unity3d.com/browse/HDRP-1599

ref:
https://github.com/Unity-Technologies/Graphics/pull/6032


kecho (approved):
Approved: just make sure you move the docs to the hdrp package for hdrp specific nodes

Vic-Cooper (approved):
Modified docs.

esmelusina (commented):
As mentioned in Slack, We are holding off reviewing new feature requests until Jan at the earliest. We'll want a design and product review before we consider the PR.

PR #6472, opened 2021-12-03
[HDRP][Path Tracing] Environment (sky) importance sampling

eturquin:
Added environment importance sampling for the path tracer. When changing the global sky settings, CDF and marginal tables are computed (happens once), and the notion of a sky light has been added to the already existing local and distant lights.

**Before:**
Sky values would only be accessed by material rays that were not intersecting anything. If there were small, bright spots in the envmap, they would result in extremely noisy images.

**Now:**
HDRI Skies are importance-sampled, resulting in much better convergence rate when using environments with hot, bright spots (e.g. a sky with a sun).

**Tests:**
This has already been tested with HDRI images from the dedicated HDRP pack. See examples below, without then with importance sampling, all computed with 256 samples. Note that setting the sky type to anything but HDRI will not use this new importance sampling scheme: although it would work fine, importance-sampling very smooth skies (as is the case with gradient and physically-based skies) is actually counter-productive, so it has been explicitly disabled.

To be tested further with as many sky settings as possible.
Sky Importance Sampling can now be toggled manually. The parameter offers 3 modes:
- HDRI Only: on on HDRI skies, but off on Gradient or Physical skies
- On: Always on
- Off: Always off

**Note:** This PR includes changes made in this other PR https://github.com/Unity-Technologies/Graphics/pull/6525

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
@@ -243,7 +243,7 @@ void ComputeSurfaceScattering(inout PathIntersection pathIntersection : SV_RayPa
                 }
 
                 // Apply material absorption
-                nextPathIntersection.value = ApplyAbsorption(mtlData, surfaceData, nextPathIntersection.t, isSampleBelow, nextPathIntersection.value);        
+                nextPathIntersection.value = ApplyAbsorption(mtlData, surfaceData, nextPathIntersection.t, isSampleBelow, nextPathIntersection.value);
 
                 pathIntersection.value += value * rrFactor * nextPathIntersection.value;
             }
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
@@ -224,15 +224,15 @@ namespace UnityEngine.Rendering.HighDefinition
         {
             switch (m_PathTracingSettings.skyImportanceSampling.value)
             {
-            case SkyImportanceSamplingMode.On:
-                return true;
+                case SkyImportanceSamplingMode.On:
+                    return true;
 
-            case SkyImportanceSamplingMode.Off:
-                return false;
+                case SkyImportanceSamplingMode.Off:
+                    return false;
 
-            default: // HDRI Only
-                var visualEnvironment = hdCamera.volumeStack.GetComponent<VisualEnvironment>();
-                return visualEnvironment.skyType.value == (int)SkyType.HDRI;
+                default: // HDRI Only
+                    var visualEnvironment = hdCamera.volumeStack.GetComponent<VisualEnvironment>();
+                    return visualEnvironment.skyType.value == (int)SkyType.HDRI;
             }
         }
 


eturquin:
without0
with0

eturquin:
without1
with1

eturquin:
without2
with2

eturquin:
without3
with3

eturquin:
As expected, many of the path tracing tests (all on one sample) fail, simply based on the fact that there is an additional light to sample, which in many situations makes the result noisier. One option to alleviate this would be to only activate env importance sampling for HDRIs and maybe physical skies, but not for gradient skies.
Tests 1000 and 5005 results need to be investigated further.

eturquin:
Reference images on path tracing tests that use HDRIs (the default texture) have been updated, tests should now go through successfully.

iM0ve (dismissed):
**Test status:** Completed
The improvement to my test scene is substantial, however found a change that I believe might not be intended. Please verify.

**The improvement:**
My scene uses a Directional light and an HDRI sky which has multiple tube lights captured. The new importance sampling really shines here. Captures are of 256 samples

Before
Before2

After
After2

**The concern:**
On the template however, where HDRI is continuous and does not contain any bright light sources, we see a loss of highlights (the material ball and the ceiling in the 2nd room). Also the picture is maybe a little more grainy.

 For light baking there is a dedicated checkbox to enable and disable importance sampling, maybe having it on all the time can cause negative effects? Captured with 1024 samples and other path tracer settings are default.

Before
Before3

After
After3


**Whats tested:**
* Path tracer on personal project
* Path tracer on template

iM0ve (changes requested):
The last commit provided some improvements but also introduced some like unwanted effects.

The template now matches better, bloom is again visible. Some light is still lost (under the material ball and red wall in the back)

Before PR
BeforeUpdated

After last commit
Round2A3

Here is the more concerning problem. I have changed the HDRI used in the template to [this](https://polyhaven.com/a/photo_studio_01). So it does not contain even lighting, more representing the target of this PR. With latest commit new streaks of light are visible, maybe the cause of selective clamping. **update - after investigation it seems expected result**

Before PR
Before6



After last commit
After6

eturquin:
I added a Path Tracing parameter to control when sky importance sampling is performed. Documentation will be added to this PR if the current controls are deemed satisfactory.
On top of that, intensity clamping has been reverted to the Master branch behaviour, i.e. applied only on indirect lighting: let's see how it performs, and if needed we may revisit it.

iM0ve (approved):
Final round of tests. So the new commit added a new parameter Sky Importance Sampling, allowing to not only turn the feature on/off for HDRI, but also for other types of skies. In both cases there was a reduction in noise.I additionally tested:

* Importance sampling on Physically based sky. Used nasa provided star map but with super hight exposure so the stars could light up the scene.
* Importance sampling on Gradient sky
* Turning off the feature for all sky types
* Rechecked previous tests to see nothing broke
* Undo functionality

Physical sky importance sampling OFF
ZEEWKTvoAs

Physical sky importance sampling ON
8xeFoUftA5

Gradient sky importance sampling OFF
Unity_k84kiF1ltS

Gradient sky importance sampling ON
Unity_Unk3N1LK19

remi-chapelain (approved):
Tested a few cases especially with Cloud layer / Volumetric Clouds since adding them to a PBSky or Gradient Sky adds clouds to the skybox directly resulting in high contrast areas.
 
The default HDRI Only is not perfect since there is some cases when we'd like to use importance sampling for uniform / gradient sky for example when there's no lights but the difference is not substantial enough to counteract the added noise when there's lights. 
So it's a good compromise. ✔️ 

Looks good to me :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
```diff
@@ -246,7 +252,7 @@ void ComputeSurfaceScattering(inout PathIntersection pathIntersection : SV_RayPa
 
 #else // SHADER_UNLIT
 
-    pathIntersection.value = computeDirect ? bsdfData.color * GetInverseCurrentExposureMultiplier() + builtinData.emissiveColor : 0.0;
+    pathIntersection.value = computeDirect ? surfaceData.color * GetInverseCurrentExposureMultiplier() + builtinData.emissiveColor : 0.0;
```
sebastienlagarde, line 255: is it expected? surfaceData.color?

eturquin: Yes

sebastienlagarde (approved):

PR #6473, opened 2021-12-03
[HDRP] Implement conservative voxelization for APV

alelievr:
### Purpose of this PRimage
- This PR improves the behavior of the voxelization used to place probes for APV when dealing with small objects.
The problem was that when an object is smaller than a brick (3 x min probe distance) it could be ignored by the probe placement system, this is caused by the rasterization of objects and can be solved by enabling [conservative rasterization](https://docs.microsoft.com/en-us/windows/win32/direct3d12/conservative-rasterization)

- There is also a small bugfix about cell position not being correct when enabling realtime voxelization

- Replaced the geometry distance offset since it's not necessary anymore with the conservative rasterization and added a new field called "Minimum Renderer Volume Size" to ignore small renderers when placing probes.

You can override this new field per probe volume or set it globally in the baking set:
image
image

When the bounding box volume of a renderer in a probe volume is smaller than the value of this field, then the renderer is discarded from the list of objects to place probes around

Added a new LayerMask field in the APV window and an override on the probe volume component instead of only being able to set it on the probe volume.

image
image

### technical details

The main issue with conservative rasterization is that it's not made to work in 3D and causes extreme artifacts when directly writing to a UAV in a fragment. There is no good solution for this as it's expected that conservative rasterization gives "bad" results for fragment interpolated values, especially when the triangle is close to being degenerate. So the only solution I found was to discard voxels written outside of the bounding box of the current rasterized triangle (this is done with a geometry shader, it's not optimal but still better than writing a software rasterizer in compute).

Note that in some cases this implementation can lead to false positives and generate more probes than before (especially in scenes with very large triangular shapes).

---
### Testing status

Without conservative voxelization

https://user-images.githubusercontent.com/6877923/144602605-6f7971d5-f235-4ee5-967b-17e26cd35ab2.mp4

With conservative voxelization

https://user-images.githubusercontent.com/6877923/144602618-2a5466ce-812f-4d31-a7eb-206ac67fe5da.mp4

Minimum renderers size feature:

https://user-images.githubusercontent.com/6877923/144643291-ec323c4b-3f2f-4cd3-b0fd-59ff6b4a7b47.mp4


FrancescoC-unity:
Because this can lead to more probes we should add this as an option (ON by default probably, but turn offable regardless)

JMargevics:
Don't forget the changelog 👍

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -732,9 +732,18 @@ public static ProbeSubdivisionResult BakeBricks(ProbeSubdivisionContext ctx)
                     {
                         var go = renderer.component.gameObject;
                         int rendererLayerMask = 1 << go.layer;
+                        renderer.volume.CalculateCenterAndSize(out _, out var rendererBoundsSize);
+                        float rendererBoundsVolume = rendererBoundsSize.x * rendererBoundsSize.y * rendererBoundsSize.z;
```
FrancescoC-unity, line 736: To avoid issues with planes I'd say we should min all dimensions to avoid 0, probably to 2/3 cm?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Skin.cs
```diff
@@ -12,7 +12,8 @@ internal static class Styles
...
             internal static readonly GUIContent s_ObjectLayerMask = new GUIContent("Layer Mask", "Control which layers will be used to select the meshes for the probe placement algorithm.");
-            internal static readonly GUIContent s_GeometryDistanceOffset = new GUIContent("Probe Placement Distance Offset", "Affects the minimum distance at which the subdivision system will attempts to place probes near the geometry. This value can be useful in situations where the generated probes don't fully cover an object.");
+            internal static readonly GUIContent s_MinRendererVolumeSize = new GUIContent("Minimum Renderer Volume Size", "Specifies the minimum bounding box volume of renderers to consider placing probes around.");
+            internal static readonly GUIContent s_OverrideMinRendererVolumeSize = new GUIContent("Override Renderer Filters", "Overrides the Minimum Renderer Volume Size specified in the Probe Volume Settings Window.");
```
FrancescoC-unity, line 16: I assume this 

> Override Renderer Filters

Should have been Minimum Renderer Volume Size?

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
alelievr: So, in this renderer size, there is already padding of 1cm in every axis

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeVolumeUI.Skin.cs
alelievr: `Override Minimum Renderer Volume Size`
Was too long in the inspector, needed to find another name

JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Haven't ran into any issues. Changes look good to me!

What I've checked:
- Renderer Filter Settings - Layer Mask
- Renderer Filter Settings - Minimum Renderer Volume Size, smaller objects are correctly ignored
  - Bounding Box Of 0.5x0.5x0.5 - bricks are shown with Min Renderer Volume Size up to 0.13
  - Bounding Box Of 1x1x1 - bricks are shown with Min Renderer Volume Size up to 1.03
  - Bounding Box Of 5x5x5 - bricks are shown with Min Renderer Volume Size up to 125
  - Bounding Box Of 5x10x50 - bricks are shown with Min Renderer Volume Size up to 2500
  - Overlapping bounding boxes - overlap is ignored, setting only checks for individual renderer bounding boxes
- Multiple local volumes with different settings
- Complex shapes
- Different Max Distance Between Probes settings

image (5)

- Probe comparison between master and PR in Classroom project - looks nearly identical

Classroom scene PR
image (6)

Classroom scene master
image (7)

sebastienlagarde:
quesotin: is it possible to create an APV test like this ?

image

sebastienlagarde (approved):

TomasKiniulis:
> quesotin: is it possible to create an APV test like this ?
> 
> image

I could try to see if I can add one in. Although do we have a way to visualize probe volume subdivision levels debug mode in Graphics Test? We have a Debug View Controller script in use for some scenes, but it seems to support only Materials, Lighting and Rendering section, will need to dig into it how to get that extended.

FrancescoC-unity:
> quesotin: is it possible to create an APV test like this ?
> 
> image

It'd be cool, the problem is that it'd also be fairly slow as we'd need to retrigger the baking. 

Although I guess we can have a scene with realtime subdivision enabled, that won't be _too_ slow. With realtime subdiv and no asset assigned in the scene (no baking ever triggered), it should work, no @alelievr ?

PR #6475, opened 2021-12-03
[VFX/SG] Fix missing ObjectToWorld using Transform

PaulDemeulenaere:
---
### Purpose of this PR
Fix this [case 1363279](https://fogbugz.unity3d.com/f/cases/1363279/)

The VFX relies on these require tranforms to populate local elementToWorld matrix (see [here](https://github.com/Unity-Technologies/Graphics/blob/e2cb5e6f60f2ef2047f0fd317ee512162c88360f/com.unity.visualeffectgraph/Editor/ShaderGraph/Templates/VFXConfig.template.hlsl#L198))

---
### Testing status
Tested locally

**Before**

https://user-images.githubusercontent.com/22494706/144620570-5e5a9762-abdb-450e-adab-48cd1a248317.mp4

**After**

https://user-images.githubusercontent.com/22494706/144620599-b9c04d5b-7b11-4bad-8cf0-b47efc418abf.mp4

---
### Comments to reviewers
See also this [conversation](https://unity.slack.com/archives/C3KQ4UD6V/p1638528607165700)

The changelog is actually only on VFX because I don't see another case where this missing declaration can cause an issue (maybe DOTS ?).


VitaSkruibyte (approved):
Thanks, looking solid!

Tested:
- Ran VFX FTP SG Scene tests
- Checked other conversions of Transform node (Vertex Stack and Fragment Stack)

com.unity.shadergraph/Editor/Data/Util/SpaceTransformUtil.cs
```diff
@@ -371,6 +375,17 @@ public static void AbsoluteWorldToWorld(SpaceTransform xform, string inputValue,
             }
         };
 
+        internal static IEnumerable<NeededTransform> ComputeTransformRequirement(SpaceTransform xform)
+        {
+            //Implicit always needed transform
+            yield return new NeededTransform(xform.from.ToNeededCoordinateSpace(), xform.to.ToNeededCoordinateSpace());
```
cdxntchou, line 381: I think technically we only need this transform if it's not Identity or ViaWorld.

com.unity.shadergraph/Editor/Data/Util/SpaceTransformUtil.cs
```diff
@@ -371,6 +375,17 @@ public static void AbsoluteWorldToWorld(SpaceTransform xform, string inputValue,
...
+
+            //ViaWorld & ObjectToAbsoluteWorld are also using the intermediate world space
+            var func = k_TransformFunctions[(int)xform.from, (int)xform.to];
+            if (func == ViaWorld || func == ObjectToAbsoluteWorld)
+                yield return new NeededTransform(xform.from.ToNeededCoordinateSpace(), NeededCoordinateSpace.World);
```
cdxntchou, line 386: should we return both (From --> World) and (World --> To) as required transforms for ViaWorld?

cdxntchou (approved):
Just two things that might be good to add, but not necessary.
Rest looks good!

com.unity.shadergraph/Editor/Data/Util/SpaceTransformUtil.cs
PaulDemeulenaere: I actually moved this code from `public NeededTransform[] RequiresTransform(ShaderStageCapability stageCapability)`.
When xform.from is equals to xform.to, it won't generates any needtransform, no need for an extra test.
However, you are right, ViaWorld will always does "From A to World, from world to B", thus "From A to B" doesn't seems needed.
Let me double check ⏳

PaulDemeulenaere: Fix at 65a242ca48740b12866a86e87fbae5ca0f8b5873

com.unity.shadergraph/Editor/Data/Util/SpaceTransformUtil.cs
PaulDemeulenaere: Fixed at 65a242ca48740b12866a86e87fbae5ca0f8b5873

PaulDemeulenaere:
Double check locally at 65a242ca48740b12866a86e87fbae5ca0f8b5873 (after adjustement)

https://user-images.githubusercontent.com/22494706/145183418-80efd897-eba5-436c-98c8-cd845bcd6391.mp4

PaulDemeulenaere:
Awaiting for Yamato result before merge ⏳

PaulDemeulenaere:
The failure about VFX_URP are expected.
See this [conversion](https://unity.slack.com/archives/GBJ7AU4V8/p1638877943055700?thread_ts=1638432337.045400&cid=GBJ7AU4V)

PaulDemeulenaere:
Backport in blocked , see https://github.com/Unity-Technologies/Graphics/pull/6579

PaulDemeulenaere:
Cancelling the needs-backport-2021.2 because root change isn't available.

PR #6476, opened 2021-12-03
Universal/fix shader paths

eh-unity:
### Purpose of this PR
Fix URP shader resource reload paths.

---
### Testing status
Manually tested URP EditorResources "Reload All". No errors after this fix.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
@@ -38,7 +38,10 @@ namespace UnityEngine.Rendering.Universal
             [Reload("Runtime/Materials/Lit.mat")]
             public Material lit;
 
-            [Reload("Runtime/Materials/ParticlesLit.mat")]
+            // particleLit is the URP default material for new particle systems.
+            // ParticlesUnlit.mat is closest match to the built-in shader.
+            // This is correct (current 22.2) despite the Lit/Unlit naming conflict.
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
```diff
@@ -38,7 +38,10 @@ public sealed class MaterialResources
             [Reload("Runtime/Materials/Lit.mat")]
             public Material lit;
 
-            [Reload("Runtime/Materials/ParticlesLit.mat")]
+            // particleLit is the URP default material for new particle systems.
+            // ParticlesUnlit.mat is closest match to the built-in shader.
+            // This is correct (current 22.2) despite the Lit/Unlit naming conflict.
+            [Reload("Runtime/Materials/ParticlesUnlit.mat")]
```
phi-lira, line 44: What do you mean, the ParticlesUnlit.mat uses particles Lit shader?

eh-unity: The resource code has everything named as ParticlesLit, and we return that as a default particle material. 
However built-in default particle material is Unlit. 
The Unlit material here uses the URP Unlit particle shader correctly.

The confusion here is that we have code named as "Lit" particle material, but it has always been Unlit. 
Unlit appears to be correct from the perspective of matching the built-in renderer. There's never been ParticlesLit.mat

phi-lira (approved):

phi-lira:
Formatting failure not caused by this PR. VFX failures also on master.

PR #6481, opened 2021-12-06
Remove options that are not available with low res transparent. 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1383425/

Now the options are just not there 

Unity_BlZ1yByuXb

**What did I test** What is above :D


FrancescoC-unity:
Forgot the changelog, will do in a sec.

sebastienlagarde (approved):

iM0ve (approved):
Small UI change, testing the visibility is updated should be sufficient.

PR #6483, opened 2021-12-06
Fix errors when APV debug is on and we switch to an asset that has APV disabled 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1385066/

The issue was that the debug was still drawn even if the feature is disabled by the pipeline. 

**What did I test** The repro steps


sebastienlagarde (approved):

JulienIgnace-Unity (approved):

iM0ve (approved):
Small fix, testing the repro should be sufficient

PR #6484, opened 2021-12-06
Fix physical camera UI

adrien-de-tocqueville:
### Purpose of this PR

When enabling physical camera, FOV was recomputed to match physical settings. Changed it to recompute physical settings based on fov, so that the FOV value is not lost in the process

camera-ui

---
### Testing status

compared undo with unticking the box, they now give the same result


pmavridis (approved):

sebastienlagarde (approved):

iM0ve (approved):
I double checked, good addition.

PR #5842, opened 2021-09-29
Improved BatchRendererGroup API (for new Hybrid Renderer) and BatchRendererGroup Test Projects

cinight:
### IMPORTANT
See link in "What should be reviewed" section to find a filter that excludes the non-code files, which are the vast majority of the changed files in the PR

### Purpose of this PR
This PR contains the Graphics changes for the new BatchRendererGroup API which is also used by the new Hybrid Renderer with DOTS 1.0.
It is meant to work with the C++ part. The PR is still in flight here, and expected to land shortly: https://ono.unity3d.com/unity/unity/pull-request/134116/_/graphics/hrnext

Detailed technical design document here: https://docs.google.com/document/d/1WOj_Uaq-h9X0BRHZo1H5wJkYE-JxO8meE6Rv1_a6PjY/edit#

Draft user manual here: https://docs.google.com/document/d/10iAN_JUsTgjqdYUF-mtJzRvVK5IqsdVh8pBq5C25Vlo/edit#heading=h.5eo6s16tlfl3

### What should be reviewed
The PR changes are very large, but this is mainly because all the DOTS related projects that lived in this repo have been removed. Some new test projects that do not depend on DOTS have also been added. The size of the 'real' changes is much more manageable. 
You can find the filtered file query [here](https://github.com/Unity-Technologies/Graphics/pull/5842/files?authenticity_token=IpgI7HHJZvTN9rBr6UnyFAhLs7aGEddyiZ%2F%2BBD9geqnMAeCGDbcd%2BhynkUE5Lhobo1pJelobM5KYneqdda3vxg%3D%3D&file-filters%5B%5D=.cs&file-filters%5B%5D=.hlsl). Basically .cs and .hlsl changes.

### Technical changes
- Hybrid Renderer V1 has been completely removed, and ifdefs related to it have been removed. The name "Hybrid V2" is also removed. Going forwards, there will only be a single Hybrid Renderer (the third version, but not called "V3").
- Shader macros have been refactored to eliminate the workaround for a bug in the fxc preprocessor. The new macros are simpler and easier to use as a result, and will only work with the new caching preprocessor which is nowadays enabled by default.
- A number of built-in shader properties in DOTS instancing variants have been migrated from DOTS instanced properties to efficiently encoded constant buffer constants set up by dedicated SRP Batcher C++ code. This saves a significant amount of GPU memory and should also be faster.
[...]

PR #6490, opened 2021-12-06
Fix HDRP build issues with DOTS_INSTANCING_ON shader variant

vincent-breysse:
### Purpose of this PR
Fix shader compilation errors with `DOTS_INSTANCING_ON` variants. This issue has been introduced a few weeks ago with https://github.com/Unity-Technologies/Graphics/pull/5842. However, no one noticed it because of a C++ side bug causing some shader errors not to be logged anywhere. A trunk PR fixing this recently landed and caused these shader compilation errors to become visible (case 1367936). Also, the pathological shader variants were in fact not used at all, which is why we didn't notice any rendering or picking issue :D

Now regarding the problem itself: 

- The UnityPerDraw cbuffer doesn't exist with the `DOTS_INSTANCING_ON` shader variants, which means that `unity_ObjectToWorld` and friends don't exist as shader constants either. That kind of per-entity data needs to be loaded from the GPU buffer `unity_DOTSInstanceData`. This is done with a few helper functions that are called when using the `UNITY_MATRIX_M` macro. `unity_ObjectToWorld` should never be used directly as a shader constant with `DOTS_INSTANCING_ON` variants. 

- The macros used to fetch per-entity data like `UNITY_MATRIX_M` don't need, and should not be re-defined by the picking header when using the `DOTS_INSTANCING_ON` shader variants. It needs to be like this to match the way BatchRendererGroup objects picking is handled on the C++ side.

---
### Testing status
- Manual testing: Building/Running BatchRendererGroup_HDRP + Run BRGGameObjects in Editor

- Yamato PR HDRP on trunk green except for the following:
    - Fail on master as well:
    `HDRP on Win_DX11_playmode_mono_Linear on version trunk`
    `HDRP on Win_Vulkan_playmode_mono_Linear on version trunk`
    `HDRP on OSX_Metal_playmode_mono_Linear on version trunk`
    `HDRP on Linux_Vulkan_playmode_mono_Linear on version trunk`
    `Find missing docs - hdrp`
    - Instability on master, fail on this PR
    `HDRP_DXR on Win_DX12_playmode_mono_Linear on version trunk`

- Yamato BatchRendererGroup on trunk green

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/PickingSpaceTransforms.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/PickingSpaceTransforms.hlsl
@@ -9,10 +9,11 @@
 
 #undef SHADEROPTIONS_CAMERA_RELATIVE_RENDERING
 
+// Define the correct matrices
 #ifndef UNITY_DOTS_INSTANCING_ENABLED
 
-#undef unity_MatrixPreviousM
 #undef unity_ObjectToWorld
+#undef unity_MatrixPreviousM
 
 #undef UNITY_MATRIX_M
 #define UNITY_MATRIX_M unity_ObjectToWorld
@@ -28,7 +29,6 @@
 
 #endif
 
-// Define the correct matrices
 #undef unity_MatrixVP
 float4x4 unity_MatrixV;
 float4x4 unity_MatrixVP;


joeldevahl (approved):

adrien-de-tocqueville (approved):
I don't know anything about dots, but it looks good HDRP wise

sebastienlagarde (approved):

PR #6491, opened 2021-12-06
Fix fbx material

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/resolve/1370671/
Normal map texture is named `_NormalMap` and not `_BumpMap` in HDRP/Lit

---
### Testing status

tested the repro steps


iM0ve (approved):
Small fix. Looks good.

sebastienlagarde (approved):

PR #6492, opened 2021-12-06
[Fogbugz # 1383093] Fixing leaky state of hardware resolution when turned off from camera

kecho:
### Purpose of this PR
fogbugz:
https://fogbugz.unity3d.com/f/cases/1383093/

Current hardware resolution completely craps out when we disable it from the camera. The reason is that nobody 'turns off' the global call to ResizeBuffers on the Dynamic resolution handler.
The solution is to flush the state always, even in the early return path.

This PR also fixes a problem that the caching state code of depth pyramid is caching the wrong state (only the res). I am now caching the actual hardware texture size and fixing depth pyramid when DRS or hardware drs is turned off.

---
### Testing status
Tested on a scene in dx12, by reproducing the bug.
Suggested test plan to QA:
* Software resolution (make sure everything works, no leaky state there)
* Test with multiple windows open (gameview and inspector open at the same time)


sebastienlagarde (approved):

remi-chapelain (approved):
No more leaky state apparently. ✔️ 
- Tested repro project with software and hardware resolution for DLSS and all DRS upscaling filters. ✔️ 
- Tested that AO was still working fine whatever the DRS settings on the cameras ✔️ 
- Tested with multiple tabs and combination of scene view / game view / inspector open ✔️

PR #6493, opened 2021-12-06
Fixed spot light shadows near plane

adrien-de-tocqueville:
### Purpose of this PR

fix https://fogbugz.unity3d.com/f/cases/1382785/

Changed the min value to 0.01 like, for camera near plane, which should not cause issues

---
### Testing status

Now setting the near plane to the minimal value doesn't make shadows disappear


iM0ve (approved):
Min value of 1 cm should be good, especially if it matches the camera near clipping distance

sebastienlagarde (approved):

PR #6494, opened 2021-12-06
Improve Path tracing docs

Vic-Cooper:
Revised the section "How to enable path tracing on a Material", with approval from docs reviewers.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
@@ -84,7 +84,7 @@ GameObjects with path tracing enabled.
 
 ## Path tracing and double-sided materials
 
-When you use path tracing, the **Double-Sided** property (menu: **Inspector** > **Surface Options** > **Double-Sided**) allows transparent materials to accumulate correctly. If the you disable **Double-sided** property, rays can't exit the GameObject. 
+When you use path tracing, the **Double-Sided** property (menu: **Inspector** > **Surface Options** > **Double-Sided**) allows transparent materials to accumulate correctly. If the you disable **Double-sided** property, rays which exit the GameObject will not behave correctly.
 
 The following images display the same GameObjects with a single-sided Material and a double-sided material:
 


Vic-Cooper:
@sebastienlagarde Resolved merge conflict.

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
```diff
@@ -53,47 +53,69 @@ Path tracing uses the [Volume](Volumes.md) framework, so to enable this feature,
...
+Path tracing in HDRP makes your scene appear more realistic. To do this, path tracing implements more precise light transport simulations than rasterization. The result is more realistic effects, especially for:
 
-Some light phenomena can be expressed more naturally in path tracing than in rasterization, for instance:
 
 - Light refraction in transmissive objects.
 - Light absorption in transmissive objects.
 - Subsurface scattering.
 
-Rasterization uses various methods to approximate complex lighting effects, with each of those methods relying on dedicated Material parameters. Path tracing computes those same effects all at once, without the need for as many parameters.
+Rasterization uses separate methods to approximate lighting effects, which require multiple Material parameters. Path tracing computes all lighting effects and how light interacts with Materials at the same time, which means Materials in your scene need fewer Material parameters.
```
sebastienlagarde, line 66: >how light interacts with Materials
we don't need to say tat, just "Path tracing computes all lighting effects at the same time" is enoguh

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
```diff
@@ -53,47 +53,69 @@ Path tracing uses the [Volume](Volumes.md) framework, so to enable this feature,
 
 **Minimum Depth** set to 2, **Maximum Depth** set to 2: indirect lighting only (1 bounce)
 
-## Materials parameterization
+## Setting path tracing parameters for Materials
```
sebastienlagarde, line 56: I guess the meaning should be slightly different as we don't have specific parameters for path tracing. It is more that some parameters are ignore, or have a slight different behavior.

So maybe we should say: Material parametrization change with Path Tracer.   Or something like that.

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
```diff
@@ -53,47 +53,69 @@ Path tracing uses the [Volume](Volumes.md) framework, so to enable this feature,
...
 - Light absorption in transmissive objects.
 - Subsurface scattering.
 
-Rasterization uses various methods to approximate complex lighting effects, with each of those methods relying on dedicated Material parameters. Path tracing computes those same effects all at once, without the need for as many parameters.
+Rasterization uses separate methods to approximate lighting effects, which require multiple Material parameters. Path tracing computes all lighting effects and how light interacts with Materials at the same time, which means Materials in your scene need fewer Material parameters.
+
+Some parameters have no effect when you use path tracing, and path tracing also changes how a Lit Material’s refraction model behaves. For more information, see [Refraction models](#refraction-models).
+
+## How to enable path tracing on a Material
+
```
sebastienlagarde, line 71: This is incorrect, the re-wording here is wrong. There is no "path tracing" option on a Material.
We used exactly the same Material with or without PathTracer, only the interpretation of the parameters is changing.

And 
>Path tracing only works on [Lit materials](Lit-Shader.md).
this is incorrect, path tracing works for all Materials.

I guess the initial document wasn't clear enough. We want to highlight that "refrcation model" in the lit shader behave differently, but we still support other Material. And ultimatly, the refraction model could be extent to other shader in the future like "stacklit". For the sake of this documentation, I think we can simply focus on lit's refraction model and say that the interpretation of this parameter on the lit shader is different when path tracing is enabled, and then we explain what is different.

sebastienlagarde: currenlty this is what you write in:

"Path tracing changes the way refraction models on a Lit Material behave."

so I guess we just need to removed the section: ## How to enable path tracing on a Material - guess this section is causing confusion.

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
```diff
@@ -61,47 +60,86 @@ You can also enable two additional settings:
...
+<a name="surface-types"></a>
 
-Some light phenomena can be expressed more naturally in path tracing than in rasterization, for instance:
+## Path tracing and double-sided materials
 
-- Light refraction in transmissive objects.
-- Light absorption in transmissive objects.
-- Subsurface scattering.
+When you use path tracing, the **Double-Sided** property (menu: **Inspector** > **Surface Options** > **Double-Sided**) allows transparent materials to accumulate correctly. If the you disable **Double-sided** property, rays can't exit the GameObject.
```
eturquin, line 87: "If the you disable **Double-sided** property, rays can't exit the GameObject."
to be replaced by
"If you disable the **Double-sided** property, rays which exit the GameObject will not be properly considered as such."

eturquin: Or something like that: basically the rays will still exit, but not in the expected way, and the resulting computation will not produce the correct result.

eturquin (changes requested):
Baring a minor change on rays exiting objects, it's all good!

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
sebastienlagarde: I replace by " rays which exit the GameObject will not behave correctly.". I will merge the PR for the cutoff, but feel free to make a new PR with a correction for this particular sentence if it isn't satisfying

sebastienlagarde (approved):

PR #6495, opened 2021-12-06
[HDRP] Fix rendering when adding an incompatible platform to the project

alelievr:
### Purpose of this PR

Fix https://fogbugz.unity3d.com/f/cases/1384634/

In the editor we actually don't need to check all the APIs in the project settings, we can only check the active one. Also, this change makes it easier to work with both HDRP and URP at the same time with the build pipeline.

Note that it's also possible to backport this if needed.

---
### Testing status
Added OpenGL to the list of graphics API on windows (along with Vulkan enabled) and it works.
Note that I also tried the repro on the bug but just adding an unsupported API to the list works too.
I also tested that the build still works correctly when a valid API is used


adrien-de-tocqueville (approved):

TomasKiniulis (approved):
Looks good! Also checked switching from URP to HDRP with HDRP supported and unsupported platform. Did not run into any issues.

sebastienlagarde (approved):

PR #6496, opened 2021-12-06
Enable Transparent Depth Pre/Post Pass for HDRP Hair Shader Graph

johnpars:
### Purpose of this PR
This is a tiny PR that addresses case https://fogbugz.unity3d.com/f/cases/1376667/.

The PR is an asset change. It enables by default the Transparent Depth Pre and Post pass for the built-in HDRP Hair Shader Graph asset. 

---
### Testing status
I manually verified that the hair card setup provided with the bug report now renders correctly. 

---
### Comments to reviewers
~


iM0ve (approved):
Testing repro project should be sufficient

sebastienlagarde (approved):

PR #6498, opened 2021-12-06
Updated Physically Based Sky documentation with more warnings about warmup cost.

JulienIgnace-Unity:
Updated Physically Based Sky documentation with more warnings about warmup cost.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Documentation~/Override-Physically-Based-Sky.md
+++ b/com.unity.render-pipelines.high-definition/Documentation~/Override-Physically-Based-Sky.md
@@ -50,7 +50,7 @@ To change how much the atmosphere attenuates light, you can change the density o
 | **Property**                   | **Description**                                              |
 | ------------------------------ | ------------------------------------------------------------ |
 | **Spherical Mode**             | Enables **Spherical Mode**. When in Spherical Mode, you can specify the location of the planet. Otherwise, the planet is always below the Camera in the world-space x-z plane. |
-| **Planetary Radius** | The radius of the planet in meters. The radius is the distance from the center of the planet to the sea level. Only available in **Spherical Mode**. |
+| **Planetary Radius**           | The radius of the planet in meters. The radius is the distance from the center of the planet to the sea level. Only available in **Spherical Mode**. |
 | **Planet Center Position**     | The world-space position of the planet's center in meters. This does not affect the precomputation. Only available in **Spherical Mode**. |
 | **Sea Level**                  | The world-space y coordinate of the planet's sea level in meters. Not available in **Spherical Mode**. |
 | **Planet Rotation**            | The orientation of the planet.                               |


com.unity.render-pipelines.high-definition/Documentation~/Override-Physically-Based-Sky.md
```diff
@@ -133,6 +133,8 @@ The default values in either mode make it so the planet's surface is at **0** on
 ## Warmup cost
 
 This sky type requires heavy precomputation to be rendered. Because of this, the first few frames (depending on the *Number of bounces* parameter) are going to take much longer. This needs to be taken into consideration when switching from another sky type to the Physically Based Sky for example as it might produce a noticeable drop in framerate.
```
Vic-Cooper, line 135: Improved this bit too. Suggested change: 

When you use a Physically Based Sky it might cause a noticeable drop in framerate. This is because HDRP performs a large amount of precomputations to render a Physically Based Sky, so the first few frames (depending on the **Number of bounces** parameter) takes more time to render than other HDRP sky types.

com.unity.render-pipelines.high-definition/Documentation~/Override-Physically-Based-Sky.md
```diff
@@ -133,6 +133,8 @@ The default values in either mode make it so the planet's surface is at **0** on
...
+This also applies when interpolating between two different Physically Based Skies with different sets of parameters by using the volume system. By doing this, Unity is forced to restart the precomputation every frame as long as interpolation is happening causing noticeable drop in framerate. To avoid that, it is recommended to only rely on a single set of parameters for a particular scene and use the sun light direction and intensity to attain the desired result.
```
Vic-Cooper, line 136: Suggested change:

This also applies when HDRP uses the volume system to interpolate between two different Physically Based Skies with different sets of parameters. To do this, HDRP restarts the precomputation every frame in which it performs interpolation. This causes a noticeable drop in framerate. To avoid this, use a single set of Physically Based Sky parameters for a scene and change the sun light direction and intensity to achieve the result you want.

com.unity.render-pipelines.high-definition/Documentation~/Override-Physically-Based-Sky.md
```diff
@@ -133,6 +133,8 @@ The default values in either mode make it so the planet's surface is at **0** on
...
+Here is the list of parameters that will cause to restart precomputation upon change: Type, Planetary Radius, Ground Tint, Air Maximum Altitude, Air Density, Air Tint, Aerosol Maximum Altitude, Aerosol Density, Aerosol Tint, Aerosol Anisotropy, Number of Bounces.
```
Vic-Cooper, line 137: Suggested change: 

HDRP restarts precomputation when you change the following parameters: 

- **Type**
- **Planetary Radius**
- **Ground Tint**
- **Air Maximum Altitude**
- **Air Density**
- **Air Tint**
- **Aerosol Maximum Altitude**
- **Aerosol Density**
- **Aerosol Tint**
- **Aerosol Anisotropy**
- **Number of Bounces**

Vic-Cooper (changes requested):
Suggested changes on a few sections.

com.unity.render-pipelines.high-definition/Documentation~/Override-Physically-Based-Sky.md
JulienIgnace-Unity: Change sounds good but I think it's missing emphasis on this part:
> This needs to be taken into consideration when switching from another sky type to the Physically Based Sky for >example as it might produce a noticeable drop in framerate.
That stresses out the fact that switching sky type during gameplay is going to show this drop in framerate.

Vic-Cooper: Thanks for your feedback! I've changed my suggestion to the following: 

When you switch to or from a Physically Based Sky, it might cause a noticeable drop in framerate. This is because HDRP performs a large amount of precomputations to render a Physically Based Sky, so the first few frames (depending on the **Number of bounces** parameter) takes more time to render than other HDRP sky types.

sebastienlagarde (approved):

PR #6502, opened 2021-12-06
Small Material UX improvements for Unlit Distortion Field

johnpars:
### Purpose of this PR
This PR addresses various small UX requests from case https://fogbugz.unity3d.com/f/cases/1383398

Here are the requests and I wrote underneath what was done to resolve them:

> 1. The Scale and Bias (the 2 and -1) values after the RGB Vector Map field is not explained properly. I assume It's probably to "remap" the values in the RG channels of the texture between -1 and 1 but, it's unclear, that the first input is multiplied and the second addtionned. 
	Possible solutions: 
		- Maybe the tooltip should mention all that ? 
		- Why use the TexturePropertySingleLine and not a remapping slider if all that does is remapping the values ?
		- Or even a "classic" remap like we do in shader graph, meaning we have a Vector2 in, and a Vector2 out ?
		
I improved the language in the tooltip to make it more clear what these fields do. 

> 2. Using tab on the bias (the default 2), will make the focus go on the distortion scale instead of the offset number (the default -1), which is painful

This one is an editor core bug, not HDRP - Ill have a separate PR or reach out to editor devs to make the small fix (it's basically an ordering issue in the UI draw calls). 

> 3. The distortion blur scale can be set to be negative, to me it does not make sense to allow a negative number here.  

Clamped to zero.

> 4. Finally is it expected that the tiling / offset is not applied to this texture ? If not, why ? If yes, it does not work. 

Added a tiling / offset field if a texture is linked to the field. 

> 5. It should be a bit more clear to how the distortion blend mode does, There's not much info on the tooltip, the add and replace are similar and multiply makes no distortion at all?

I improved the documentation for Surface Options to explain the blend modes. They were already documented for Shader Graph options so I just reused them. 

> 6. Distortion depth test is basically a "force depth write" for the transparent ?

Yes. I updated the tooltip to hopefully make it more clear. 

--

Visually there is little organizational change to the field, aside from the texture transform option:
image

---
### Testing status
I manually used the UI to confirm the usage of the tiling/offset and clamping of blur scale. 

---
### Comments to reviewers
~


sebastienlagarde (approved):

iM0ve (approved):
Small changes, looks good.

johnpars:
Forgot to add a docs team ticket, standby

PR #6505, opened 2021-12-07
Fixed flipped UV for directional light cookie on PBR Sky

adrien-de-tocqueville:
### Purpose of this PR

fix https://fogbugz.unity3d.com/f/cases/1382656/
It doesn't look like this was a design choice, so i think it should be fixed

Once again, I am not sure what backport labels to put

---
### Testing status

tested the repro project, with various sun orientation
Now:
image


RemyUnity:
Maybe some notes should be added to the upgrade guide about this, as it will alter the rendering of scenes that were using directional light cookies.

iM0ve (changes requested):
Indeed, please include a short note in the docs section. Other than that good addition.
Upgrade guides example: https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@13.1/manual/Upgrading-from-2021.1-to-2021.2.html

adrien-de-tocqueville:
waiting on #6516 to add the upgrade guide

sebastienlagarde (approved):

sebastienlagarde:
no backport for this one

iM0ve (approved):
I would write it like this, but I am okay to whatever doc team suggests.

While Physically Based Sky is used, Directional lights can have a surface texture, located in Celestial Body section. The orientation of the texture was incorrect, in HDRP 14 it was fixed by flipping UVs on the x axis. When upgrading a project, suns texture might need to be flipped.

So its clear that it only matters for PB sky from the start of the sentence, at the end what the user is expected to do.

adrien-de-tocqueville:
yeah it's better, it will be a better starting point for doc, thanks

sebastienlagarde:
need to solve conflict

sebastienlagarde:
have you trigger yamato?

PR #6506, opened 2021-12-07
Disable raytraced shadows when in full shadowmask

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1382831/ 

The solution we went for is just disable raytraced shadows when fully shadowmask. This because the other way round requires bigger changes that we can look into if it becomes a problem later on. 

Now raytraced shadows are off when full shadow mask is used, to match this I also added a disable in the UX. 

Updated a bit the docs.  

What was tested: scene similar to repro description. Static object shadowmask and raytraced enabled. Before double shadows, now no double shadows


iM0ve:
Please describe whats tested, or what we could look into

FrancescoC-unity:
> Please describe whats tested, or what we could look into

Done :-)

sebastienlagarde (approved):

remi-chapelain (changes requested):
There's something funky happening with the UI; 
Before, you can't enable the ray traced shadow without screen space shadow, 
https://user-images.githubusercontent.com/57442369/145860337-5bbbfa72-f150-45eb-8917-71ba17d469c8.mp4

After, you can (even though it has no effect because raytraced shadow is disabled because of shadowmask)

https://user-images.githubusercontent.com/57442369/145860571-e6b842ad-9c97-48c3-8927-7b662f354390.mp4

Also, after baking with the fix, you can't have raytraced shadow anymore (which is the point), but when going back to real time and clearing the shadowmaps, you can't have raytraced shadows anymore for some reasons :| 

https://user-images.githubusercontent.com/57442369/145860964-e1bda3f3-277d-46a7-9b28-82a4a289fc52.mp4

FrancescoC-unity:
The UX one is easy to fix and already did, the other is turning out to be much more complex and this is mostly due to the fact that changes of "baking mode" don't trigger a change in the UI and the UI for them is controlled C++ side. 

With the new burst lightloop, there is no update of properties unless a change in UI happens, so we are a bit in a weird spot. 

Trying to find solutions.

FrancescoC-unity:
@remi-chapelain  should be good now?

remi-chapelain (approved):
All good now ! ✔️ 
Tested multiple types of lights combined. When in mixed and shadow mask rt is grey out and even if it was checked, it's not visible now. 

After removing shadowmask or mixed lighting on the light, we get the raytraced shadow back ✔️

sebastienlagarde:
test are failing: error: UnhandledLogMessageException: Unhandled log message: '[Error] C:\build\output\Unity-Technologies\Graphics\com.unity.render-pipelines.high-definition\Runtime\Lighting\Light\HDAdditionalLightData.cs(3632,29): error CS1061: 'Light' does not contain a definition for 'lightmapBakeType' and no accessible extension method 'lightmapBakeType' accepting a first argument of type 'Light' could be found (are you missing a using directive or an assembly reference?)'. Use UnityEngine.TestTools.LogAssert.Expect

PR #6509, opened 2021-12-07
[URP][Backport] make the Vulkan RenderTexture MSAA support fallback work in the same way as the swapchain backend implementation

manuele-bonanno:
backport of: 
https://github.com/Unity-Technologies/Graphics/pull/6180
https://github.com/Unity-Technologies/Graphics/pull/2342

Backport case:
https://fogbugz.unity3d.com/f/cases/1384313/

### Purpose of this PR
The way RenderTextures handle MSAA fallback when an unsupported sample count of 2 is requested (falling back to numSamples = 1), differs fom the way the fallback is handled when setting up the Vulkan swapchain (rounding up numSamples to 4, if supported). This caused an issue on Mali GPUs which don't support 2x MSAA.
This PR makes sure that on Vulkan the MSAA unsupported fallback behaviour is consistent between RenderTextures and Swapchain.

We are also making sure that depth copy is always disabled on GLES, since it doesn't work on that platform.

---
### Testing status
Tested the repro project of [this case](https://fogbugz.unity3d.com/f/cases/1375673).

---
### Comments to reviewers

we should review how all backends handle MSAA fallbacks and move these implementation details in engine code.

check the implementation of SelectSupportedSampleCount(const VkSampleCountFlags supportedSamplesCounts, const VkSampleCountFlagBits requested) in GfxDeviceVK.cpp for more info

### DIFF
--- a/com.unity.render-pipelines.core/package.json
+++ b/com.unity.render-pipelines.core/package.json
@@ -3,7 +3,7 @@
     "description": "SRP Core makes it easier to create or customize a Scriptable Render Pipeline (SRP). SRP Core contains reusable code, including boilerplate code for working with platform-specific graphics APIs, utility functions for common rendering operations, and  shader libraries. The code in SRP Core is use by the High Definition Render Pipeline (HDRP) and Universal Render Pipeline (URP). If you are creating a custom SRP from scratch or customizing a prebuilt SRP, using SRP Core will save you time.",
     "version": "7.8.0",
     "unity": "2019.4",
-    "unityRelease": "28f1",
+    "unityRelease": "34f1",
     "displayName": "Core RP Library",
     "dependencies": {
         "com.unity.ugui" : "1.0.0",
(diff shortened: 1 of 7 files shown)


aleksandrasdzikia (approved):

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -6,12 +6,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [7.8.0] - 2021-07-23
 
+### Added
+- Added a supported MSAA samples count check, so the actual supported MSAA samples count value can be assigned to RenderTexture descriptors.
+
```
phi-lira, line 11: This should be removed as we are not adding any user facing functionality.

phi-lira (approved):

PR #6513, opened 2021-12-07
Avoid typed load on volumetric fog filtering on platforms that don't support it 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1386535/ 

When typed load are not supported (see image below) , we need an extra buffer allocated at least temporarily to have different output texture than the input. 

image

This should be a problem only on very old GPUs so **important for QA**: Please make sure to check on a GPU from the one that have NO in the row for typed loads in the image above.

**What did I test**: I forced both modes in code and looked at the test for volumetric fog filtering (5008) and that it looked good in both paths.


com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/HDRenderPipeline.VolumetricLighting.cs
```diff
@@ -1031,6 +1034,7 @@ TextureHandle VolumetricLightingPass(RenderGraph renderGraph, HDCamera hdCamera,
                     passData.viewCount = hdCamera.viewCount;
                     passData.filterVolume = ((int)fog.denoisingMode.value & (int)FogDenoisingMode.Gaussian) != 0;
                     passData.sliceCount = (int)(cvp.z);
+                    passData.filteringNeedsExtraBuffer = true;// !(SystemInfo.IsFormatSupported(GraphicsFormat.R16G16B16A16_SFloat, FormatUsage.LoadStore));
```
sebastienlagarde, line 1037: ?

FrancescoC-unity: That's me commiting debug code by accident :P

FrancescoC-unity: Will fix up Monday first thing (a matter of bringing back the commented part)

RemyUnity:
I tried as much as I could to test this PR.
On the two potential candidate machines I had at disposal, one (pc, gtx770) didn't log any warning/error before the PR anyway, neither after, but it didn't affect the rendering.
And the second one (mac, gtx 750m) simply didn't want to open unity editor without any information :/

If Seb thinks it's safe to merge, I trust him.

PR #6515, opened 2021-12-07
FSR Upscaling Filter (With Special Merge Target)

gmitrano-unity:
### Purpose of this PR
Add FSR Upscaling Filter to URP

This change adds a new upscaling filter to URP that uses FidelityFX
Super Resolution. It is only available on platforms that support
shader target 4.5 or above. On unsupported platforms, it will behave
as if the user selected the "Auto" filter in the pipeline asset.

Unlike other filters, this filter will execute even when the renderer
is not configured to scale. HDRP has the same behavior. This allows the
FSR shaders to improve visual quality even when scaling is not active.

77% Scaling with Bilinear Filter:
image

77% Scaling with FSR Filter:
image

100% Scaling (Native):
image

All images were rendered with 1080p as their final resolution and include the mip bias changes currently out for review in a different PR.

---
### Testing status
Tested on Windows with Vulkan, Android with Vulkan. Still needs XR testing and more comprehensive testing in general.

---
### Comments to reviewers
This PR is based on top of a temporary branch to make review easier. When it's time to merge, this PR should be retargeted to master or staging instead.

---
### Notes for QA
Please test each filter type at 0.5 resolution scale, and 0.77 resolution scale with FXAA on and off (AA should either be FXAA or MSAA). A fixed viewport size should be used for this (such as 1920x1080) to make sure the integer scaling path is triggered during Auto.

---
### TODO
- [ ] Changelog updates
- [ ] Documentation updates
- [ ] Automated graphics tests
- [x] Re-enable 16-bit mode before merging (Fix black-screen issue in RCAS)

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
+++ b/com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
@@ -231,6 +231,8 @@ namespace UnityEngine.Rendering.Universal
             bool requireFinalPostProcessPass =
                 lastCameraInStack && !ppcUpscaleRT && stackHasPostProcess && cameraData.antialiasing == AntialiasingMode.FastApproximateAntialiasing;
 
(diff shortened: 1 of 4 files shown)


gmitrano-unity:
Closing test PR

PR #6517, opened 2021-12-07
Write to VT feedback in debug modes

adrien-de-tocqueville:
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1376874/

VT feedback texture was not written to when DEBUG_DISPLAY was enabled, so textures would not be updated

---
### Testing status

Tested on the repro project, with scene lighting mode off, and material debug in the rendering debugger with:
* forward no msaa
* forward msaa
* deferred

NOTE: it doesn't work with DX11 in deferred or forward no msaa, i think it's a c++ issue and assigned the fogbugz to the appriopriate team


JulienIgnace-Unity (approved):

TomasKiniulis (approved):
Tested on Dx12 and Dx11 with multiple Scene Views, split Scene/Game views. VTs works fine for me now with various debug modes and lighting disabled.

debaetsd (approved):
The fact that this does not seems to work in dx11 was throwing me off but the code looks good (and my quick tests does indicate that it work as-expected on dx12)

sebastienlagarde (approved):

PR #6519, opened 2021-12-07
Add option to force motion blur off when XR is active

FrancescoC-unity:
As the title says, but worded positively as I know UX doesn't like negative action (Force Motion Blur off was my instict :D) 

This option allow to force MB off when XR is enabled as a setting in the HDRP Asset. 

I tested with MockHMD.

@fabien-unity  lemme know if this is what you meant :-)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/GlobalXRSettings.cs
```diff
@@ -12,7 +12,8 @@ internal static GlobalXRSettings NewDefault() => new GlobalXRSettings()
         {
             singlePass = true,
             occlusionMesh = true,
-            cameraJitter = false
+            cameraJitter = false,
+            allowMotionBlur = true
```
fabien-unity, line 16: I think this should be OFF by default actually.

fabien-unity (approved):
LGTM. Thanks!

PR #6520, opened 2021-12-07
Add Overlay Render Queue to Custom Pass

johnpars:
### Purpose of this PR
Address case https://fogbugz.unity3d.com/f/cases/1379399/

The original request of this bug report was to add support for the Overlay render queue to HDRP. 

Since overlay is an invalid render queue range for HDRP's internal draw calls, I instead explored the route of ensuring that the custom pass volume is capable over drawing renderers in this queue. I found that the Draw Renderers Custom Pass did not include the Overlay range, so I added it. 

image

We should now support shaders that decide to use the Overlay queue, like TextMeshPro, by utilizing a draw renderers custom pass. 

image

---
### Testing status
I manually used the draw renderers custom pass with overlay queue shaders to verify that they are now correctly drawn. 

---
### Comments to reviewers
~


iM0ve (approved):
Waiting for Jans review

sebastienlagarde (approved):

alelievr (approved):

JMargevics:
https://user-images.githubusercontent.com/50581102/145565055-8fb4f0ef-aaed-40ce-858c-f4dc92d8105a.mp4

Verified with Overlay TMP shaders and a custom unlit shader.  The solution is not so obvious for common user who just wants an overlay TMP shader, so covering this somewhere in docs would be a good idea i think :)

JMargevics (approved):

PR #6522, opened 2021-12-08
Universal, Fix Swap Buffer Issue

etienne-p:
While writing custom Scriptable Render Features, 
I stumbled upon a `Dimensions of color surface does not match dimensions of depth surface` error.

After some digging, I noticed that in the `UniversalRenderer`, line 1196 in `SwapColorBuffer`, we have:
`ConfigureCameraTarget(m_ColorBufferSystem.GetBackBuffer(cmd), m_ColorBufferSystem.GetBufferA());`

In some cases, this would result in `m_CameraColorTarget` and `m_CameraDepthTarget` RTs having different dimensions.

The cause is that when invoking `GetBufferA` on `RenderTargetBufferSystem`,
if msaa is not used, the `RTHandle` we get may not have been reallocated with the proper dimensions.

Indeed, in the `RenderTargetBufferSystem.ReAllocate`,
we do not reallocate the msaa rtHandles if msaa is not used (`!(desc.msaaSamples > 1)`).
So previously allocated msaa targets (with different dimensions) survive and later get used,
assigned as depth target through `GetBufferA()`.

While my fix seems to work,
I'm not very familiar with this area and will let you determine if it is indeed the right way to fix it.
I'm not sure I understand the introduction/use/implementation of `GetBufferA()` to begin with.
(It is only used in one place afaik.)

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/RenderTargetBufferSystem.cs
+++ b/com.unity.render-pipelines.universal/Runtime/RenderTargetBufferSystem.cs
@@ -110,7 +110,7 @@ namespace UnityEngine.Rendering.Universal.Internal
 
         public RTHandle GetBufferA()
         {
-            return m_A.rtMSAA;
+            return (m_AllowMSAA && m_A.msaa > 1) ? m_A.rtMSAA : m_A.rtResolve;
         }
 
         public void EnableMSAA(bool enable)


sandy-carter-unity:
Looks like an omission, probably due to a merge from the RTHandle branch and some work @robin-demoor was doing.

sandy-carter-unity (approved):

robin-demoor:
At first sight this looks good to me, I don't have access to my workstation right now so will have a closer look tomorrow to be sure. A note on the GetBufferA. Before RTHandles the depth and color were stored together in one RenderTargetHandle. Because there was no need for a swapbuffer system for depth. The depth was always stored in A and never in B so we don't initialise two depth buffers. Because of that there was a need to specifically be able to access buffer A to get the depth. As far as I am aware RTHandles splits depth and color? So this GetBufferA can probably be removed and be replaced by wherever the intermediate depth is currently stored.

etienne-p:
@robin-demoor thanks, I hope I didn't land too fast (in any case it's an improvement afaik). If we can ditch `GetBufferA` it'd likely be better. It kind of stands out as odd in the `RenderTargetBufferSystem` interface and the unique usage confirms this. Feel free to submit another PR if you'd rather iterate on the fix, I'd be interested so add me as a (non blocking) reviewer please. (I'm on the Virtual Production team, not graphics, I can only allocate so much time to SRP fixes). We could use a comment explaining what you just stated in `SwapColorBuffer` :)

PR #6525, opened 2021-12-08
[HDRP][Path Tracing] Fixed auto-exposure mistmatch between sky background and rendered objects

eturquin:
Addresses this bug: https://fogbugz.unity3d.com/f/cases/1385131

Because we only compute the sky background on the first frame of an accumulation, there was a mismatch between the auto-exposure multiplier applied to the sky, and the one used on objects of the scene (on Unlit materials).

This has been fixed by removing the exposure multiplier from the initial sky render, which makes the process cleaner too, as we don't need the inverse multiplier during accumulation anymore. This is a low-risk change that can be merged quickly.


JulienIgnace-Unity (approved):

eturquin:
One test is failing (5004), looking into it.

eturquin:
Test 5004 issue fixed, all tests are green.

sebastienlagarde (approved):

patrickp-unity3d:
Thanks!

PR #6530, opened 2021-12-08
[ShaderGraph][CI] Add ShaderGraph BuiltIn Target jobs to "all shadergraph" job

jessebarker:
### Purpose of this PR
Some recent breakages in the ShaderGraph Builtin Target were due to the tests not having been run on the changes.  The nightly job runs the BuiltIn Target jobs, but the main PR job does not.  This PR adds them.

I would like to hear that URP, HDRP, and ShaderGraph are all ok with this change before it merges (i.e., if people are happy running these jobs manually, or running them via the nightly job, maybe we don't need it).

---
### Testing status
Kicking off a ShaderGraph PR job on Yamato to confirm this has the desired effect.

---
### Comments to reviewers


jessebarker:
I've pulled the formatting fix from master, so this should be ready to go if everyone agrees to add these jobs.

theo-pnv (approved):
Keep in mind that doing so will make the PR job longer, if one day we reenable its auto trigger on the PRs. But that day is not tomorrow so we can re-assess the change when it's relevant.

PR #6532, opened 2021-12-08
Add option to do Object Space Planar/Triplanar mapping for Lit

FrancescoC-unity:
As mentioned and requested on Slack, this PR adds a mode to chose the space of the position used for triplanar and planar mapping. This only for Lit and nothing else. 

https://user-images.githubusercontent.com/43168857/145231647-562f5acc-29f5-4ba9-a9ec-459a3e66414d.mp4

What I tested can be seen above and I also tried with emissive (not shown in the video)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitData.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitData.hlsl
@@ -142,15 +142,16 @@ void GetLayerTexCoord(float2 texCoord0, float2 texCoord1, float2 texCoord2, floa
 {
     layerTexCoord.vertexNormalWS = vertexNormalWS;
     float objectSpaceMapping = false;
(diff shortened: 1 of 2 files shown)


TomasKiniulis (changes requested):
Tested various maps with object space Planar and Triplanar:
- Base Map
- Mask Map
- Normal Map
- Subsurface Mask Map
- Coat Mask
- Specular Color
- Iridescence Mask

No issues with any specific masks, but when objects are rotated with triplanar projection textures are stretched, same appears to happen on shadergraph if triplanar does not use object space normals. 

https://user-images.githubusercontent.com/50582134/145373355-180975c5-d070-40f5-a309-38b3ee2a07b5.mp4

FrancescoC-unity:
Issue still there, I think I know why. Will fixup as soon as I get in the office after lunch

TomasKiniulis (approved):
Looks good, ObjectSpace projection works like a charm now!

sebastienlagarde (approved):

PR #6536, opened 2021-12-08
[ON HOLD - DO NOT MERGE] Separate multiplier for ambient probe and Sky reflection

FrancescoC-unity:
This PR introduces two new fileds in the indirect lighting controller 

image

Namely to scale ambient probe separately (**note** it will not work when the ambient probe is sampled through old tetra probes as no way to know what is ambient probe and what not there) and sky fallback reflections. 

Discussed this with @pierre-unity  as first step in this direction, more complext separated tuning will require further discussions. 

I updated the screenshot for the already existing test, but will kick yamato now.


sebastienlagarde (approved):

iM0ve (changes requested):
Hello, please update whats tested section. Also what else its worth checking besides the Indirect controller override or it shouldnt affect anything else?

JulienIgnace-Unity (approved):

FrancescoC-unity:
> Hello, please update whats tested section. Also what else its worth checking besides the Indirect controller override or it shouldnt affect anything else?

I changed the test :-) 

I tested looking at water/clouds/fog with ambient probe affecting them and moving the changes. 
I also checked the sky reflection probe being scaled by the factor.

iM0ve (changes requested):
The feature does work as expected, but we are still iterating on the tooltip and maybe the name of the Ambient Probe Multiplier parameter.

https://user-images.githubusercontent.com/36502659/146540625-04879ead-b2e1-481e-abc0-c20ee7eaa5bd.mov

sebastienlagarde:
talk with Pierre Yves, close for now

PR #6547, opened 2021-12-09
[DO NOT MERGE] Universal tests upgrading assets in our test projects

martint-unity:
---
### Purpose of this PR
To upgrade our tests to skip the Upgrade pop-up that we have and to have the latest assets. 
this also makes it faster to open the projects since they do not need to upgrade.


TestProjects/UniversalGfxTestStereo/Assets/ForwardRenderer.asset
```diff
@@ -12,8 +12,12 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: de640fe3d0db1804a85f9fc8f5cadab6, type: 3}
   m_Name: ForwardRenderer
```
phi-lira, line 13: @martint-unity Does this need to be updated?

phi-lira (approved):

ellioman (approved):

martint-unity:
DO NOT MERGE. Looks like there are some files that got in that not supposed to be there,

martint-unity:
out of date

PR #6552, opened 2021-12-09
Sanitize vfx asset before updating model if the asset file has changed.

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1382874/

When a VFX asset file is modified (let's say outside of Unity Editor) and that this same asset is opened in Unity it can lead to an error because of inconsistency between the state of the graph in memory (before file changes) and the new state of the graph (from the file). In the worst case, the graph can even be completely different (overwriting the file with another one).

Same issue can happen when changing the from inside Unity when you create a new VFX asset and you overwrite another asset which is already opened. See video below.
yHlk5YUJMm

---
### Testing status
Tested:
- overwriting the whole file with a different content from a text editor
- only modifying meaningless values in the file (like nodes coordinates) from a text editor
- overwriting the asset within Unity Editor with the default VFX template

---
### Comments to reviewers
This bug revealed that VFX Graph is currently not consistent with other tools like shadergraph regarding file modification.
When you modify a scene file or a shadergraph file you are prompted with a message box giving you the choice between reloading the asset to take changes into account or keep the current state (saving the asset would then discard the changes made on the file).

So here this is just a bug fix, but we should probably go a little farther and implement the same behavior as other tools.
I would prefer not to do it in the scope of this bug because it will involve lot more risky changes.

### DIFF
--- a/com.unity.visualeffectgraph/Editor/GraphView/VFXViewWindow.cs
+++ b/com.unity.visualeffectgraph/Editor/GraphView/VFXViewWindow.cs
@@ -9,7 +9,6 @@ using UnityEngine;
 using UnityEngine.VFX;
 using UnityEngine.UIElements;
 
-
 namespace UnityEditor.VFX.UI
 {
     [Serializable]
@@ -19,6 +18,7 @@ namespace UnityEditor.VFX.UI
 
         ShortcutHandler m_ShortcutHandler;
         VisualEffect m_pendingAttachment;
+        bool m_HasChanged;
 
         void OnEnable()
         {
@@ -120,6 +120,11 @@ namespace UnityEditor.VFX.UI
             }
 
             return false;
+        }
+
+        public void MarkHasChanged()
+        {
+            m_HasChanged = true;
         }
 
         public VFXView graphView { get; private set; }
@@ -342,6 +347,12 @@ namespace UnityEditor.VFX.UI
             var filename = "No Asset";
             if (controller != null)
             {
+                if (m_HasChanged)
+                {
+                    m_HasChanged = false;
+                    controller.graph.SanitizeForImport();
+                }
+
                 controller.NotifyUpdate();
                 if (controller.model != null)
                 {
--- a/com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs
+++ b/com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs
@@ -4,7 +4,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Reflection;
-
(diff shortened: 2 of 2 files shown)


com.unity.visualeffectgraph/Editor/GraphView/VFXViewWindow.cs
```diff
@@ -342,6 +347,12 @@ void Update()
             var filename = "No Asset";
             if (controller != null)
             {
+                if (m_HasChanged)
+                {
+                    m_HasChanged = false;
+                    controller.graph.SanitizeForImport();
```
julienamsellem, line 353: The cost of this call probably depends on the size/complexity of the graph but with the default template it's 0ms.

PaulDemeulenaere: I have to check twice, isn't the SanitizeForImport already called [here](https://github.com/Unity-Technologies/Graphics/blob/171f783523aa4405adc946f18990a0b11a378a62/com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs#L38) ? The sanitized state isn't serialized, so, I would expect to be false while trying to reload but maybe I'm wrong. 🤔 
The problem of calling `SanitizeForImport();` here is it will hide some issue when graph is opened.

julienamsellem: That's correct, I'll see why the call to Sanitize in postproceessor is not enough (or is not called)

VladNeykov (changes requested):
Hi, the issue in the description still reproduces 

Repro steps:
1. Create an effect, delete everything, add _Simple Heads and Sparks_, and save it
2. Open a new VFX window, create a new VFX, and override the effect

_Note: this seems to work fine with simple graphs, graphs with multiple outputs, and graphs containing multiple systems. Could only narrow down GPU events as problematic._

Is this something addressable in this PR or outside of the scope?

julienamsellem:
Thanks @VladNeykov for your insight. Thanks to your comment I identified that the root cause is rather related to multiple systems assets (which makes sense regarding the error message). I'll dig in that direction.

com.unity.visualeffectgraph/Editor/GraphView/VFXViewWindow.cs
PaulDemeulenaere: In the end, should we postponed this PR ? The change here seems really risky (regarding dandling import we have right now).

julienamsellem: Yes, you're right.
That could be a test case for future importer refactorization.

PR #6553, opened 2021-12-09
[2020.3][ShaderGraph][URP Target] Fixup additional lights keywords for the URP Shade…

jessebarker:
### Purpose of this PR
This PR addresses https://fogbugz.unity3d.com/f/cases/1387281/

The core issue in the parent case was an "over-normalization" of the view direction which caused incorrect rendering.  That issue is not present in 2020.3 (the normalization was added later).  However, the secondary issue of the configuration of the additional lights keywords is still an issue.  This backports that change to the URP Target.

---
### Testing status
Testing according to the repro steps in the parent case yields generated shaders that look correct  Additionally, the frame debugger does not show the keyword set as shown here:
Screen Shot 2021-12-09 at 7 07 37 AM

---
### Comments to reviewers


ellioman (approved):

esmelusina (approved):
LGTM

PR #6557, opened 2021-12-09
16-Bit FSR Cleanup & Fixes

gmitrano-unity:
### Purpose of this PR
This change removes the 16-bit option in the FSRCommon.hlsl in favor of
an automatic solution based on REAL_IS_HALF. Users of the FSR code no
longer need to explicitly opt-in to the 16-bit implementation. It will
now be enabled automatically whenever the target platform has support.

Some vendors have issues with 16-bit floating point casts to uint16
which are required for FSR's math approximations. This issue only
occurs on DX11 drivers, so the 16-bit implementation is currently
disabled whenever DX11 is in use until this is fixed.

This change also contains a small modification for the uint to float
conversion logic used inside FSRUtils.cs:SetRcasConstants. The updated
logic avoids an unnecessary memory allocation which was causing
failures in URP's graphics tests.

---
### Testing status
Tested on Windows & Android, + iOS via Yamato


kecho (approved):

kecho:
Thanks Greg, Yamato is close to gree on HDRP. LMK if you struggle getting this through.

phi-lira (approved):
LGTM.

PR #6562, opened 2021-12-10
Urp/command buffer refactor prototype

manuele-bonanno:

### Checklist for PR maker

---
### Purpose of this PR

---
### Testing status

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
@@ -536,7 +536,7 @@ namespace UnityEngine.Rendering.Universal
 
                 // Doesn't create texture for Overlay cameras as they are already overlaying on top of created textures.
                 if (intermediateRenderTexture)
-                    CreateCameraRenderTarget(context, ref cameraTargetDescriptor, useDepthPriming);
+                    CreateCameraRenderTarget(context, ref cameraTargetDescriptor, useDepthPriming, renderingData.commandBuffer);
 
                 m_ActiveCameraColorAttachment = createColorTexture ? m_ColorBufferSystem.PeekBackBuffer() : m_XRTargetHandleAlias;
                 m_ActiveCameraDepthAttachment = createDepthTexture ? m_CameraDepthAttachment : m_XRTargetHandleAlias;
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -29,6 +29,8 @@ public struct RenderingData
         /// True if post-processing effect is enabled while rendering the camera stack.
         /// </summary>
         public bool postProcessingEnabled;
+
+        public CommandBuffer commandBuffer;
```
phi-lira, line 33: Maybe `globalCommandBuffer`? Needs public API docs as well.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -638,7 +642,7 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
 
         if (xrActive)
         {
-            CommandBuffer cmd = CommandBufferPool.Get();
+            CommandBuffer cmd = CommandBufferPool.Get(); // pipelineCommandBuffer; //CommandBufferPool.Get();
```
phi-lira, line 645: Why do we still call get here in this path instead of getting the pipelineCommandBuffer?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -998,6 +1003,10 @@ static void InitializeAdditionalCameraData(Camera camera, UniversalAdditionalCam
             renderingData.perObjectData = GetPerObjectLightFlags(renderingData.lightData.additionalLightsCount);
             renderingData.postProcessingEnabled = anyPostProcessingEnabled;
 
+            //pipelineCommandBuffer = CommandBufferPool.Get();
+            //renderingData.commandBuffer = pipelineCommandBuffer;
+            renderingData.commandBuffer = null; // initialize after this?
```
phi-lira, line 1008: This got me questioning if we should get a command buffer per camera or one per frame. It seems like if you initialize in rendering data, then it should be per-frame?

com.unity.render-pipelines.universal/Runtime/UniversalRenderer.cs
```diff
@@ -624,10 +624,11 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
                 depthDescriptor.msaaSamples = 1;// Depth-Only pass don't use MSAA
                 RenderingUtils.ReAllocateIfNeeded(ref m_DepthTexture, depthDescriptor, FilterMode.Point, wrapMode: TextureWrapMode.Clamp, name: "_CameraDepthTexture");
 
-                CommandBuffer cmd = CommandBufferPool.Get();
+                CommandBuffer cmd = renderingData.commandBuffer; //CommandBufferPool.Get();
                 cmd.SetGlobalTexture(m_DepthTexture.name, m_DepthTexture.nameID);
-                context.ExecuteCommandBuffer(cmd);
-                CommandBufferPool.Release(cmd);
+                context.ExecuteCommandBuffer(cmd); // do we need this here?
+                cmd.Clear(); // do we need this here?
```
phi-lira, line 630: Not entirely sure why it's done here, I would assume it's only needed if there's any ScriptableRenderContext call that would break execution of command buffer in those passes. 
 
 Maybe it's a safer approach to avoid changing the execute / clear logic of command buffers until the RendererList PR lands?

PR #6570, opened 2021-12-10
Fixed a null ref exception when using sky lighting override.

JulienIgnace-Unity:
### Purpose of this PR
Fixed a null ref exception when using sky lighting override.

---
### Testing status
Tested repro from @TomasKiniulis


sebastienlagarde (approved):

TomasKiniulis (approved):
Looks good. No more issues changing Lighting Override Mask in HDRP asset.

PR #6575, opened 2021-12-10
Enable state was not taken into account when pasting a Block

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1315482/

Copy paste or duplicate a block in a context would ignore the disable state.
d6ujZqP904

---
### Testing status
Tested: 
- copy paste disabled block with CTRL+C / CTRL+V
- copy paste disabled block with context menu commands
- duplicate with CTRL+D
- duplicate with context menu commands
- enabled blocks still works like before

---
### Comments to reviewers
This is one liner fix, very low risk 😀


VladNeykov:
Thanks for the fix and for the testing status!
Tried a few other bits including undo/redo of the operations you already listed and it's looking good.

I did however find an issue which does not reproduce in master - creating subgraph blocks sets all blocks in the subgraph to be disabled:
https://user-images.githubusercontent.com/7230872/145676688-c4e96275-e783-43d9-8b7d-2e48df7a449c.mp4

julienamsellem:
Thanks for the finding, it should be fix with this commit https://github.com/Unity-Technologies/Graphics/pull/6575/commits/5c3ad9b9c890b08d63290526d9fb528bdafdaae1

VladNeykov (approved):
Thanks for the fix; looking solid!

Tested:
 Copy/duplicate via right click or keyboard shortcuts on:
- blocks, subgraph blocks
- between different contexts, systems, and VFX assets
- within blocks
- from blocks (copy) to VFX contexts (paste)
- copying/duplicating/creating subgraphs from selections containing disabled and enabled blocks
- undo/redo on all of the above

PaulDemeulenaere (approved):
Looks good to me but you can probably update the change log to track this resolution.
Thanks 🟢

PR #6305, opened 2021-11-16
[RP Workflows] Unification of Shader Variant Log Level

alex-vazquez:
### Purpose of this PR
Another step towards unification of shader stripping across SRP's

https://jira.unity3d.com/browse/XPIPELINE-445

This PR handles the unification of the Shader Variant Log Level.

URP Shader Variant Log Level
image

HDRP Shader Stripping new section
image

So if a Pipeline doesn't extend the interface we will be sure they do not want shader stripping logging/exporting.

---
### Testing status
Test that the values are properly serialized/deserialized.

---
### Comments to reviewers

We are using an extension pattern for global settings, in order to allow from code on core to obtain this shader stripping values. For example the code on core will look like:

```
if (RenderPipelineManager.currentPipeline.defaultSettings is IShaderVariantSettins shaderVariantSettings)
{
// do stuff with shaderVariantSettings.shaderVariantLogLevel and exportShaderVariants
}
```

PR #6546, opened 2021-12-09
 Universal Staging Branch

phi-lira:
- [URP] Native Render Pass depth resolve support (#6217) 
- [Universal][Mobile] Light Banding fixes (halfs to floats) (#6355) 
- Document Mixed Lighting in URP PostPro Tests (#6359) 
- [URP] RTHandles Load/Store actions optimizations fix (#6334) 
- Scriptable Renderer: Dismiss depth slice for non-camera passes check (#… 
- Universal/swapbuffer 2d perf (#6240) 
- DecalRendererFeature: Convert to using RTHandles (#6401) 
- Universal/fix shader paths (#6476) 
- Revert Automatic Intermediate Texture handling (#6275) 
- Fix Scene View + Render Scale Logic (#6423) 
- commited updated materials to fix test issue (#6540) 
- [URP] Upscaling Filters & FXAA Scaling Logic (#6178) 
- [URP] Port HDRP Mip Bias Logic to URP (#6229)
- Disabling depth resolve temporarily on OSX because of gfx driver issu… 
- Fix missing depth on overlay camera DoF (#6549) 
- Shader debug print (#1028)

PR #6582, opened 2021-12-13
[RPW] Fix crash when upgrading URP Global Settings when no render has been done

alex-vazquez:
Null reference exception while migratting an URP asset to global settings.

Fix for issue introduced on: https://github.com/Unity-Technologies/Graphics/pull/6305

By making Ensure, we obtain the correct instance,(as if the instance is null, and no URP global settings have been created is created one)


arttu-peltonen (approved):

ellioman (approved):

phi-lira (approved):
LGTM thank you for jumping on it quickly. @alex-vazquez does your team do RCA on regressions? I feel like it would be incredibly useful to see what could be done in testing to catch regressions like this.

phi-lira:
The null reference exception on VFX_URP on Win_DX11_playmode_mono_Linear on version trunk is fixed on [universal staging branch](https://github.com/Unity-Technologies/Graphics/pull/6546).

PR #6589, opened 2021-12-13
[2022.2] Docs: API changes

ellioman:
# Purpose of this PR
Going through the public API Docs and tweaking the scope and/or documentation


com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.Skin.cs
```diff
@@ -5,7 +5,7 @@ namespace UnityEditor.Rendering.Universal
 {
     partial class UniversalRenderPipelineLightUI
     {
-        static class Styles
+        private static class Styles
```
phi-lira, line 8: This seems issue in the docs tooling as well, when no access modifier is specified then private is used. 
It's ok for this PR of couse but this should be followed up. @oleks-k 

[From C# reference](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers): 
> Class and struct members, including nested classes and structs, have private access by default.

com.unity.render-pipelines.universal/Editor/PostProcessDataEditor.cs
```diff
@@ -4,6 +4,9 @@
 
 namespace UnityEditor.Rendering.Universal
 {
+    /// <summary>
+    /// Editor script for a <c>PostProcessData</c> class.
+    /// </summary>
     [CustomEditor(typeof(PostProcessData), true)]
     public class PostProcessDataEditor : Editor
```
phi-lira, line 11: This should not be public I think, all editors should be private or internal.

com.unity.render-pipelines.universal/Editor/ScriptableRendererDataEditor.cs
```diff
@@ -9,6 +9,9 @@
 
 namespace UnityEditor.Rendering.Universal
 {
+    /// <summary>
+    /// Editor script for a <c>ScriptableRendererData</c> class.
+    /// </summary>
     [CustomEditor(typeof(ScriptableRendererData), true)]
     public class ScriptableRendererDataEditor : Editor
```
phi-lira, line 16: Same here, should be private or internal.

com.unity.render-pipelines.universal/Editor/ScriptableRendererFeatureEditor.cs
```diff
@@ -3,9 +3,13 @@
 
 namespace UnityEditor.Rendering.Universal
 {
+    /// <summary>
+    /// Editor script for a <c>ScriptableRendererFeature</c> class.
+    /// </summary>
     [CustomEditor(typeof(ScriptableRendererFeature), true)]
     public class ScriptableRendererFeatureEditor : Editor
```
phi-lira, line 10: Same here, should be private or internal.

com.unity.render-pipelines.universal/Editor/UniversalAdditionalLightDataEditor.cs
```diff
@@ -3,10 +3,14 @@
 
 namespace UnityEditor.Rendering.Universal
 {
+    /// <summary>
+    /// Editor script for a <c>UniversalAdditionalLightData</c> class.
+    /// </summary>
     [CanEditMultipleObjects]
     [CustomEditor(typeof(UniversalAdditionalLightData))]
     public class UniversalAdditionLightDataEditor : Editor
```
phi-lira, line 11: Same here, should be private or internal.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
```diff
@@ -206,6 +206,7 @@ bool ShouldUpgradeShader(Material material, HashSet<string> shaderNamesToIgnore)
             return !shaderNamesToIgnore.Contains(material.shader.name);
         }
 
+        /// <inheritdoc/>
```
phi-lira, line 209: Not sure why this is needed as well, even though this is public here the class is internal. Maybe another shortcoming in the tooling?

com.unity.render-pipelines.universal/Editor/UniversalRendererDataEditor.cs
```diff
@@ -4,6 +4,9 @@
 
 namespace UnityEditor.Rendering.Universal
 {
+    /// <summary>
+    /// Editor script for a <c>UniversalRendererData</c> class.
+    /// </summary>
     [CustomEditor(typeof(UniversalRendererData), true)]
     public class UniversalRendererDataEditor : ScriptableRendererDataEditor
```
phi-lira, line 11: This should be internal or private.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -15,10 +15,29 @@ namespace UnityEngine.Rendering.Universal
...
+        /// </summary>
         Depth = 1 << 0,
+
+        /// <summary>
+        /// Used when a <c>ScriptableRenderPass</c> requires a normal texture.
+        /// </summary>
         Normal = 1 << 1,
+
+        /// <summary>
+        /// Used when a <c>ScriptableRenderPass</c> requires a normal texture.
```
phi-lira, line 34:
```suggestion
        /// Used when a <c>ScriptableRenderPass</c> requires a color texture.
```

phi-lira (approved):
Approved with some comments.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
ellioman: Fixed!

com.unity.render-pipelines.universal/Editor/Lighting/UniversalRenderPipelineLightUI.Skin.cs
ellioman: I don't think the doctools were complaining about this line. I simply added it to make things clearer as some are internal/public and some private.

com.unity.render-pipelines.universal/Editor/ScriptableRendererFeatureEditor.cs
ellioman: Leaving it as public as it might break users with custom versions.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineMaterialUpgrader.cs
ellioman: It actually wasn't needed. I simply added it in case we make it public.

com.unity.render-pipelines.universal/Editor/Deprecated.cs
```diff
@@ -2,9 +2,13 @@
 
 namespace UnityEditor.Rendering.Universal
 {
+    /// <summary>
+    /// Editor script for a <c>ForwardRendererData</c> class.
+    /// </summary>
     [Obsolete("ForwardRendererDataEditor has been deprecated. Use UniversalRendererDataEditor instead (UnityUpgradable) -> UniversalRendererDataEditor", true)]
```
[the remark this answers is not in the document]
oleks-k: I'm trying to use the filter.yml approach again now, will tell the results later.

PR #6597, opened 2021-12-13
Screen Coordinates Override

etienne-p:
### Purpose of this PR

This PR introduces **Screen Coordinates Override** (**SCO** for the remainder of this document) for both URP and HDRP. SCO
provides a transformation of screen space coordinates when evaluating screen space effects.

We have found it useful in virtual production scenarios.

The first use case for this is **Cluster Display** (formerly known as [Cluster Rendering](https://docs.unity3d.com/560/Documentation/Manual/ClusterRendering.html)), that is, distributed rendering. It allows for post effects such as vignette to be properly rendered at the scale of a grid of displays.

Another use case is the implementation of Lens Distortion when replicating the distortion incurred by a physical lens. Since pixels near the edges may be pushed inwards of the viewport by the distortion, we need to render the scene with overscan. (By overscan, we mean rendering to an expanded pixel surface, later cropped out to its original size before being
presented.) To preserve the distortion's aspect in the cropped image, we need to apply a transform to screen space coordinates when the lens distortion is being computed. 

The first implementation of this technique was part of the first Cluster Display effort, and was then called **Cluster
Display Space** or **Global Screen Space**. A high level description of the technique can be found in
[this document](https://docs.google.com/document/d/1I_mR42Tt119nVhhHwHG9F_n7DwpPin7Tds4yhx59oko/edit?usp=sharing) and [this one](https://docs.google.com/document/d/1qZqUTA5VAeV35PkldjOXXjpKD-JxWfgs2KJC6rh-4mo/edit?usp=sharing).

The switch to the **Screen Coordinates Override** terminology is justified by the fact that the technique is relevant outside of
Cluster Display. As illustrated above by our Lens Distortion example.

In the changes we regularly see the introduction of 2 fields:

- `screenCoordScaleBias` holds the scale and bias used to transform screen space coordinates.
- `screenSizeOverride` holds the screen size (and its reciprocal) to be used in conjunction with transformed screen
  space coordinates.

---
### Changes Overview

* #### SRP Core
    - `ScreenCoordOverrideUtils` -> basic utilities to manage the shader keyword. Main point is to avoid the duplication of
      the shader keyword.
    - `ScreenCoordOverride.hlsl` -> shader code and macros to perform the screen space coordinates conversion.

* #### HDRP
    - `FrameSettings` -> add `ScreenCoordOverride` and `AsymmetricProjection` settings. `AsymmetricProjection` allows the
      activation of the asymmetric projection code path even when XR is not enabled. This is useful for Cluster Display.
    - `HDAdditionalCameraData` -> add the `ScreenSizeOverride` and `ScreenCoordScaleBias` fields.
    - `ShaderVariablesGlobal` -> add the `ScreenSizeOverride` and `ScreenCoordScaleBias` fields.
    - `HDCamera` -> update global shader variables.
    - `HDRenderPipeline` -> activates the shader keyword.
    - `HDRenderPipeline.PostProcess` -> activates the shader keyword based on frame settings.
    - `UberPost.compute` -> add SCO support.

* #### URP
    - `ScriptableRenderer` -> fix a swap buffer issue (first
      fixed [here](https://github.com/Unity-Technologies/Graphics/pull/6522))
    - `PostProcessUtils` -> init random generation using frame count so that it's consistent across machines in a Cluster
      Display scenario.
    - `RenderTargetBufferSystem` -> remove a method not needed anymore. (Was involved in the swap buffer issue.)
    - `UniversalAdditionalCameraData` -> add the `ScreenSizeOverride` and `ScreenCoordScaleBias` fields and feature
      activation.
    - `UniversalRenderPipeline` -> populate additional camera data.
    - `UniversalRenderPipelineCore` -> add fields and associated shader properties.
    - `UniversalRenderer` -> related to swap buffer fix.
    - `Input.hlsl` -> add the `ScreenSizeOverride` and `ScreenCoordScaleBias` fields.
    - `FinalPost.shader` -> add SCO support.
    - `UberPost.shader` -> add SCO support.

---
### Testing status

We added the following graphics tests. We test custom post processes and passes since we want to enable users to use the
feature.

* #### HDRP_Tests
    - **9930-ScreenCoordOverride** -> test with a built-in vignette effect.
    - **9931-ScreenCoordOverrideCustomPostProcess** -> test with a custom post process.
    - **9932-ScreenCoordOverrideCustomPass** -> test with a custom pass.

* #### UniversalGraphicsTest_PostPro
    - **250_ScreenCoordOverrideRenderPass** -> test with a scriptable render pass.
    - **250_ScreenCoordOverridePostProcess** -> test with a built-in vignette post process.

We are also in the process of developing Cluster Display and test this feature as part of our work. We are also adding
graphics test to that package as well.

---
### Comments to reviewers

@sandy-carter-unity regarding the swap buffer bug we discussed in late december.
@phi-lira feel free to assign relevant members from your team.
@sebastienlagarde feel free to assign relevant members from your team.
Let me know how CHANGELOG update works, I assume I shouldn't add entries below past releases.


com.unity.render-pipelines.core/ShaderLibrary/ScreenCoordOverride.hlsl
```diff
@@ -0,0 +1,24 @@
...
+}
+
+#if defined(SCREEN_COORD_OVERRIDE)
+    #define SCREEN_COORD_APPLY_SCALEBIAS(xy)          ScreenCoordApplyScaleBias(xy, _ScreenCoordScaleBias)
+    #define SCREEN_COORD_REMOVE_SCALEBIAS(xy)         ScreenCoordRemoveScaleBias(xy, _ScreenCoordScaleBias)
+    #define SCREEN_SIZE_OVERRIDE                     _ScreenSizeOverride
+#else
+    #define SCREEN_COORD_APPLY_SCALEBIAS(xy)          xy
+    #define SCREEN_COORD_REMOVE_SCALEBIAS(xy)         xy
+    #define SCREEN_SIZE_OVERRIDE                     _ScreenSize
```
wenzhang-unity, line 21: Do you mean `#define SCREEN_SIZE_OVERRIDE _PostProcessScreenSize`?

wenzhang-unity (dismissed):

sebastienlagarde:
note: can you merge master and trigger yamao test (if not done) thanks

sebastienlagarde:
I qdded @RSlysz for the additional API added on HDAddtionalCamera as there is a refactor coming

com.unity.render-pipelines.core/Runtime/Utilities/ScreenCoordOverrideUtils.cs
```diff
@@ -0,0 +1,29 @@
+namespace UnityEngine.Rendering
+{
+    /// <summary>
+    /// Screen Coord Override Utility class.
+    /// </summary>
+    public class ScreenCoordOverrideUtils
```
sebastienlagarde, line 6: Do you really need to create a new class to enable/disable keyword? looks like simply defining 
const string k_ShaderKeyword = "SCREEN_COORD_OVERRIDE"; will be sufficient
(This is a remark, I am not against it but seems extra code we could avoid.

com.unity.render-pipelines.core/ShaderLibrary/ScreenCoordOverride.hlsl
sebastienlagarde: I notice ths comment
// SCREEN_SIZE_OVERRIDE defaults to _ScreenSize when SCREEN_COORD_OVERRIDE is not defined.
#ifdef SCREEN_COORD_OVERRIDE
but it is true qs zenzhang mention that there is this replacement: https://github.com/Unity-Technologies/Graphics/pull/6597/files#diff-d50957ca80ebaf9db3fbd56554efbd4b827572fa4449b5b2245e677c5fa004d9L133

so what's happen?

sebastienlagarde: and here we have a ifdef SCREEN_COORD_OVERRIDE
for _PostProcessScreenSize usage, so there is certainly a mistake
https://github.com/Unity-Technologies/Graphics/pull/6597/files#diff-d50957ca80ebaf9db3fbd56554efbd4b827572fa4449b5b2245e677c5fa004d9R202

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
```diff
@@ -360,6 +360,12 @@ public enum TAAQualityLevel
         /// <summary>Enable to retain history buffers even if the camera is disabled.</summary>
         public bool hasPersistentHistory = false;
 
+        /// <summary>Screen size used when Screen Coordinates Override is active.</summary>
+        public Vector4 screenSizeOverride;
```
sebastienlagarde, line 364: tag for @RSlysz

com.unity.render-pipelines.core/Runtime/Utilities/ScreenCoordOverrideUtils.cs
sebastienlagarde: i.e referring to this code 

               if (m_ScreenCoordOverride)
                {
                    ScreenCoordOverrideUtils.EnableKeyword(passData.uberPostCS);
                }

                if (PostProcessEnableAlpha())
                {
                    passData.uberPostCS.EnableKeyword("ENABLE_ALPHA");
                    
                    
which can simply be rewrite

               if (m_ScreenCoordOverride)
                {
                     passData.uberPostCS.EnableKeyword(k_ScreenOverrideShaderKeyword);
                }

                if (PostProcessEnableAlpha())
                {
                    passData.uberPostCS.EnableKeyword("ENABLE_ALPHA");

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2237,6 +2239,12 @@ void InitializeGlobalResources(ScriptableRenderContext renderContext)
             else
                 FrameSettings.AggregateFrameSettings(ref currentFrameSettings, camera, additionalCameraData, m_Asset);
 
+            // Sanity check.
+            if (xrPass.enabled && currentFrameSettings.IsEnabled(FrameSettingsField.AsymmetricProjection))
+            {
+                Debug.LogError("XR and explicit Asymmetric Projection (Frame Settings) are not expected to be enabled simultaneously.");
```
sebastienlagarde, line 2245: dont do this, this is hard to catch and polute the log. Instead disable (i.e grey in the ux) the framesettings if XR is enabled, and in the agreggate framesettings function do something like this:
FrameSettingsField.AsymmetricProjection = FrameSettingsField.AsymmetricProjection && XREnabled
We do that for many other settings, you can take example on it.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -158,6 +158,12 @@ public enum FrameSettingsField
         /// <summary>When enabled, Cameras using these Frame Settings render water surfaces.</summary>
         [FrameSettingsField(0, autoName: Water, customOrderInGroup: 106, tooltip: "When enabled, Cameras using these Frame Settings render water surfaces.")]
         Water = 99,
+        /// <summary>When enabled, HDRP accounts for asymmetry in the projection matrix.</summary>
```
sebastienlagarde, line 161: I am a bit worried about this statement as it it not totally true, effect like volumetric fog will fail with (strong) asymetric matrix. But I am not sure what to say instead :P

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -158,6 +158,12 @@ public enum FrameSettingsField
...
+        [FrameSettingsField(0, displayedName: "Asymmetric Projection", customOrderInGroup: 107, tooltip: "When enabled HDRP will correctly render using asymmetric projection matrices, even when physical properties are used.")]
+        AsymmetricProjection = 78,
+        /// <summary>When enabled, HDRP evaluates post effects using transformed screen space coordinates.</summary>
+        [FrameSettingsField(0, displayedName: "Screen Coordinates Override", customOrderInGroup: 108, tooltip: "When enabled HDRP will use Screen Coordinate Override for post processing and custom passes.")]
```
sebastienlagarde, line 165: Statement is not self explaining. We should say that this allow cluster display as when reading current message, I still have no idea what is the usage.

Also please add documentation inside the doc folder (i.e the .md file for framesettings)

sebastienlagarde: https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.high-definition/Documentation~/Frame-Settings.md

com.unity.render-pipelines.universal/Shaders/PostProcessing/UberPost.shader
```diff
@@ -12,9 +12,11 @@ Shader "Hidden/Universal Render Pipeline/UberPost"
         #pragma multi_compile_local_fragment _ _USE_FAST_SRGB_LINEAR_CONVERSION
         #pragma multi_compile_vertex _ _USE_DRAW_PROCEDURAL
         #pragma multi_compile_fragment _ DEBUG_DISPLAY
+        #pragma multi_compile _ SCREEN_COORD_OVERRIDE
```
sebastienlagarde, line 15: could be multi_compile_fragment

sebastienlagarde (changes requested):
some small adjustement to do, see comment

also could you add a changelog, thanks

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
RSlysz: This should have no impact on the refactor we are doing

RSlysz (commented):
This will not prevent us to deliver the refactor. 

I am more concerned about the added utility in CoreRP. I am not sure it will be really useful in its current state. And name let little room for extensibility. Is there any other things that we will need to add in ScreenCoordOverrideUtils ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
etienne-p: Actually I can just remove this, it is a somewhat overly defensive piece of code I just added since using XR and Cluster Display at the same time doesn't really make sense. But it doesn't actually break anything afiak. I'll add a check in our Cluster Display package instead.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
etienne-p: made the comment more specific based on this frame setting usage

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
etienne-p: thanks, on a side note it seems to me some checkboxes are not documented, ex: Virtual Texturing, Water, ...

etienne-p:
> This will not prevent us to deliver the refactor.
> 
> I am more concerned about the added utility in CoreRP. I am not sure it will be really useful in its current state. And name let little room for extensibility. Is there any other things that we will need to add in ScreenCoordOverrideUtils ?

I can't foresee anything more to be added currently. I agree that this class is very shallow and should be removed. I only added it to not duplicate the hardcoded shader keyword which is a bit excessive.

com.unity.render-pipelines.core/Runtime/Utilities/ScreenCoordOverrideUtils.cs
etienne-p: This is a very shallow class I initially added not to duplicate the hardcoded shader keyword but I'm removing it :)

etienne-p:
@sebastienlagarde @RSlysz I addressed the feedback. Wanted to wait on some graphics tests I'm working on in our Cluster Display package to be wrapped up as they help catching issues with this PR. The one trickier comment had to do with `_PostProcessScreenSize`. Since the Screen Coord Override shader code lives in `core`, I added a file specific to HDRP to redefine `SCREEN_SIZE_OVERRIDE` (https://github.com/Unity-Technologies/Graphics/blob/a5320db1d913d448265f8221e30926de2c050109/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ScreenCoordOverride.hlsl). Alternatively we could have had one `ScreenCoordOverride.hlsl` per pipeline but it would have led to duplication. Or 3 files instead of 2. Besides with this approach we can control whether we want the redefine to happen. (not that strong of a selling point in itself, but still)  @wenzhang-unity FYI

com.unity.render-pipelines.core/ShaderLibrary/ScreenCoordOverride.hlsl
etienne-p: see https://github.com/Unity-Technologies/Graphics/pull/6597#issuecomment-1014956858

sandy-carter-unity:
Is `_ScreenCoordScaleBias` meant to be a drop-in replacement for `_BlitScaleBias` as it is used int the Blitter Api?

etienne-p:
> Is `_ScreenCoordScaleBias` meant to be a drop-in replacement for `_BlitScaleBias` as it is used int the Blitter Api?

No it is meant to provide a transformation for screen coordinates, to be performed before some screen space post processing shader computations are performed. You can look at the docs linked in the PR description for more explanations of the problem/solution. I'd also be available for a zoom call :) The only commonality with `BlitScaleBias` is that we describe our 2D transformation as a scale and translation

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
sebastienlagarde: thanks for reporting non documented checkbox, will poke people

sebastienlagarde (approved):

sandy-carter-unity (approved):
These changes look sensible and don't conflict with my dynamic scaling work which touch uv but on the vertex shader side.

phi-lira (changes requested):
LGTM overall. Mostly concerned about shader stripping.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
```diff
@@ -6,6 +6,7 @@ Shader "Hidden/HDRP/FinalPass"
         #pragma editor_sync_compilation
         #pragma only_renderers d3d11 playstation xboxone xboxseries vulkan metal switch
 
+        #pragma multi_compile _ SCREEN_COORD_OVERRIDE
```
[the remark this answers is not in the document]
etienne-p: I need to use `multi_compile_fragment` here

phi-lira (approved):
LGTM, thank you for adding the stripping option.

patrickp-unity3d:
@eturquin when using this feature, should users also use https://github.com/Unity-Technologies/Graphics/pull/5953 when using path tracing? My guess is otherwise there would be 'noise' seams at the edges of the tiles, correct?

PR #6599, opened 2021-12-14
Automatically enable Alpha-To-Coverage when alpha clipping is on

adrien-de-tocqueville:
### Purpose of this PR

Removed the alpha to coverage option on material and shadergraphs, it will now be always enabled in forward msaa when alpha clipping is on.
URP will do it like that with #6312 and it's actually a good idea since the performance cost is minimal and it removes a shader variant. Also it's much simpler to setup because not a lot of users know about this option

There was also a frame setting to enable/disable alpha to coverage per camera. I also removed it, but maybe there can be an interest in keeping it.

I removed every mention of alpha to coverage in the docs, and search all occurence of alphatocoverage in code to make sure i removed everything.

NOTE: i'll have to mention the change in what's new if we want to merge

---
### Testing status

There are two tests 9402 that test alpha to coverage with regular shaders and shadergraphs, they don't change with this PR
Tested that in deferred, and forward without msaa there is no issue.

Yamato looks green

---
### Comments to reviewers


remi-chapelain (commented):
Good idea in general and I agree it's also good to align on what URP does. 
I'm for deleting reference of it in the editor, but it can still be interesting to note it in the doc that we use this method / algorithm by default in the MSAA / Alpha Clipping docs just for the sake of it because it was implemented recently-ish (1 year ago)

remi-chapelain (approved):
Looks good to me. 
Feature has been tested and tests are still passing ✔️

sebastienlagarde (approved):

PR #6600, opened 2021-12-14
Fixed potential asymmetrical resource release in the volumetric clouds (case 1388218).

anisunity:
https://fogbugz.unity3d.com/f/cases/1388218/
When toggling very quickly, the asset's previous state could be captured incorrectly this fixes the issue.

Testing status
The issue doesn't repro anymore on the repro dataset.


TomasKiniulis (approved):
Thanks! Checking template no warnings occur anymore when Volumetric Clouds are quickly turned on/off. Clouds are still rendered correctly or disabled when checkbox is on/off.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #6601, opened 2021-12-14
Fixed the intensity of the sky being reduced signficantly even if there is no clouds (case 1388279).

anisunity:
https://fogbugz.unity3d.com/f/cases/1388279/
Due to numerical imprecisions, when normalizing the neighbors we could end up with a slightly lower transmittance, which attenuated significantly the intensity of the sun due to the non linear profile. This considers that there is no transmittance if we are past a certain threshold.

Testing status
Issue doesn't repro anymore on the dataset.

### DIFF
--- a/com.unity.postprocessing/CHANGELOG.md
+++ b/com.unity.postprocessing/CHANGELOG.md
@@ -4,8 +4,10 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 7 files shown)


remi-chapelain (approved):
This has been pair tested and pair validated. 

It's still not perfect because of how the intensity is attenuated when the sun goes behind a cloud, reducing significatnly its intensity and bloom indirectly but at least now, when the sun is not behind a cloud, the intensity is the same (and bloom too) than without clouds at all. 

https://user-images.githubusercontent.com/57442369/146041640-8d87756d-59f8-46dd-80fb-d5f511eb4c1b.mp4

anisunity:
Updated the screenshots and relaunched the tests

sebastienlagarde (approved):

PR #6616, opened 2021-12-15
[2022.2] Docs: API changes - Part 2

ellioman:
# Purpose of this PR
Going through the public API Docs and tweaking the scope and/or documentation


com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -19,46 +19,147 @@ public enum ShadowQuality
...
         _2048 = 2048,
+
+        /// <summary>
+        /// Use this for 4096x4096 shadow resolution.
+        /// </summary>
         _4096 = 4096
     }
 
+    /// <summary>
+    /// This controls the size of the Light Cookie texture.
```
eh-unity, line 54: **atlas** texture for additional lights (point, spot)

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -19,46 +19,147 @@ public enum ShadowQuality
...
         _2048 = 2048,
+
+        /// <summary>
+        /// Use this for 4096x4096 Light Cookie resolution.
+        /// </summary>
         _4096 = 4096
     }
 
+    /// <summary>
+    /// Options for selecting the format for the Light Cookie texture.
```
eh-unity, line 85: Affects whether the additional lightcookies are grayscale only or color. Atlas texture format, so additional lights only. Mainlight uses the cookie texture directly.

eh-unity: Low precision saves memory and bandwidth. Intended for mobile.

com.unity.render-pipelines.universal/Runtime/ShaderUtils.cs
```diff
@@ -3,22 +3,72 @@
...
 namespace UnityEngine.Rendering.Universal
 {
+    /// <summary>
+    /// Options to get a shader path to URP shaders when calling ShaderUtils.GetShaderGUID();
+    /// <see cref="ShaderUtils"/>.
+    /// </summary>
     public enum ShaderPathID
     {
+        /// <summary>
+        /// Use this for URP Lit shader.
```
eh-unity, line 13: Is ComplexLit missing from this list? Should it be here?
Anyway ComplexLit is a superset of Lit for advanced effects that would need additional GBuffer storage. I is always forward rendered and it's used as deferred fallback (for those effects).

eh-unity (commented):
Lgtm. I added a couple improvement suggestions.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -133,10 +284,21 @@ public enum VolumeFrameworkUpdateMode
     /// </summary>
     public enum UpscalingFilterSelection
     {
+        /// <summary>
+        /// Use this to have URP handle upscaling automatically.
```
gmitrano-unity, line 288:
```suggestion
        /// Indicates that URP will select an appropriate upscaling filter automatically.
```

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -133,10 +284,21 @@ public enum VolumeFrameworkUpdateMode
...
+        /// </summary>
         [InspectorName("Automatic")]
         Auto,
+
+        /// <summary>
+        /// Use this to use Bilinear Filtering when performing upscaling.
```
gmitrano-unity, line 294:
```suggestion
        /// Indicates that Bilinear filtering will be used when performing upscaling.
```

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -133,10 +284,21 @@ public enum VolumeFrameworkUpdateMode
...
+        /// </summary>
         [InspectorName("Bilinear")]
         Linear,
+
+        /// <summary>
+        /// Use this to use Nearest-Neighbour filtering when performing upscaling.
```
gmitrano-unity, line 300:
```suggestion
        /// Indicates that Nearest-Neighbour filtering will be used when performing upscaling.
```

gmitrano-unity (changes requested):
Suggested some changes for the upscaling filter comments. It's mostly wording stuff, but the automatic filter comment now tries to make it clear that it will select one of the existing filters depending on which one is most appropriate rather than doing something else that's not available to the user.

ellioman:
@gmitrano-unity Thanks for the help! I've committed your suggestions so please do a re-review for me :)

gmitrano-unity (approved):
Updates look good to me! :)

oleks-k (approved):
Looks good, thanks, Elvar!

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Added.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Added.

PR #6618, opened 2021-12-15
Fix HDRP tests not compiling

alelievr:
### Purpose of this PR
Fix compilation issue on HDRP_Tests.


alex-vazquez (approved):
Double check with @alelievr , that warnings on the HDRP test are errors.

JulienIgnace-Unity (approved):

PR #6626, opened 2021-12-15
Fixing light render properties not updating properly during timeline updates

kecho:
### Purpose of this PR
Because of a previous optimization, light properties were not flushed properly when a timeline modified its properties (properties exclusively in the HDAdditionalLightData).
This PR allows this to work properly.

---
### Testing status
* Added a light, and animated a couple of the properties in a timeline
* Animated Fade Distance, ensured that it worked correctly
* Suggestions for QA would be to test this also in play mode

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
@@ -2693,6 +2693,7 @@ namespace UnityEngine.Rendering.HighDefinition
         void OnDidApplyAnimationProperties()
         {
             UpdateAllLightValues(fromTimeLine: true);
+            UpdateRenderEntity();
         }
 
         /// <summary>
@@ -2806,6 +2807,7 @@ namespace UnityEngine.Rendering.HighDefinition
 #endif
 
             data.UpdateAllLightValues();
+            data.UpdateRenderEntity();
         }
 
         // As we have our own default value, we need to initialize the light intensity correctly
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -89,6 +89,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed rendering in the editor when an incompatible API is added (case 1384634).
 - Fixed issue with typed loads on RGBA16F in Volumetric Lighting Filtering.
 - Fixed Tile/Cluster Debug in the Rendering Debugger for Decal and Local Volumetric Fog
+- Fixed timeline not updating PBR HDAdditionalLightData parameters properly.
 
 ### Changed
 - Converted most TGA textures files to TIF to reduce the size of HDRP material samples.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -2796,6 +2812,7 @@ public void CopyTo(HDAdditionalLightData data)
 #endif
 
             data.UpdateAllLightValues();
+            UpdateRenderEntity();
```
tlaedre, line 2815: Shouldn't this be `data.UpdateRenderEntity()`?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -2678,6 +2682,18 @@ void LateUpdate()
                 timelineWorkaround.oldDisplayAreaLightEmissiveMesh = displayAreaLightEmissiveMesh;
                 timelineWorkaround.oldLightColorTemperature = legacyLight.colorTemperature;
             }
+
+            if (timelineWorkaround.renderEntityDataVersion != m_CurrentRenderEntityVersion)
```
tlaedre, line 2686: It looks to me like none of this will work properly in player builds because of https://github.com/Unity-Technologies/Graphics/pull/6626/files#diff-ce5da627f91b89bc1411c29b4bba111e4197513b924bbb3fa614ed87f2f74c34R2576-R2579

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -2678,6 +2682,18 @@ void LateUpdate()
...
+            }
+        }
+
+
+        void DirtyRenderEntityData()
+        {
+            ++m_CurrentRenderEntityVersion;
         }
 
         void OnDidApplyAnimationProperties()
```
tlaedre, line 2699: A bit nit-picky, but is this "fromTimeline" stuff actually necessarily from timeline? It could be any playable graph, including MecAnim, no?

tlaedre (changes requested):

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
kecho: good catch.

kecho:
@tlaedre ok, next attempt. Simplified the logic. We are now spending the cost of flushing every time only the lights taht get animated. This should be ok. Let me know if this works.

tlaedre (approved):

remi-chapelain (approved):
There's most propably some timeline issue because in some cases, on the first run as discussed with @kecho "OnDidApplyAnimationProperties()" doesn't get called but we're good on the HDRP side. 

Tried to animate the angular diameter of the sun and works fine now. 

https://user-images.githubusercontent.com/57442369/146788063-7f9b144f-ee70-4056-9dc0-3d04a518c663.mp4

PR #6629, opened 2021-12-16
Fixed a crash with render graph viewer

JulienIgnace-Unity:
### Purpose of this PR
Fixed a crash with render graph viewer when render graph is not provided with an execution name
Fogbugz https://fogbugz.unity3d.com/f/cases/1388207/

---
### Testing status
Tested repro case


sebastienlagarde (approved):

sebastienlagarde:
test should be sufficiant givent the change, merging

PR #6634, opened 2021-12-16
Fix shaderdebug compile error + enhancements.

eh-unity:
### Purpose of this PR
Fixes a compile error/warning when new input system package is missing but enabled.
Add docs.
Fix incorrect handling of truncated print data.
Add a profiling sampler, so it's clear in the profile that the debug prints are on.
Add data types back into printed output (lost in previous refactor).
Fix bool shader value type in one of the print functions.

---
### Testing status
Manually tested that printing works with various types. Feature is excluded by default anyway (behind an explicit define).


alex-vazquez (approved):
Thanks

PR #6635, opened 2021-12-16
[HDRP] Fix tile/cluster debug for Decal and Fog

simonb-unity:
### Purpose of this PR
TileClusterCategoryDebug and LightCategory are out of sync (ProbeVolumes are no longer a light category), so in the Rendering Debugger -> Lighting -> Tile/Cluster Debug the visuals for Decal and LocalVolumetricFog are not correct. This PR fixes them and changes the debug enum to use the category values in its definition so they cannot get out of sync.

---
### Testing status
Manually tested in the HDRP template. Result for fog after fix:
image


sebastienlagarde (approved):
THanks a lot for this fix, could you add a changelog ?

PR #6643, opened 2021-12-17
Changed the behavior the max ray length for recursive rendering to match RTR and rasterization.

anisunity:
In order to avoid having objects that are not rendered while flagged for recursive ray tracing we are forcing the ray length of the camera ray to be equal to the rasterization near and far plane and keep the max ray length for the additional bounces.

Testing status
Tested it on various set ups and it does what it says.


remi-chapelain (changes requested):
Agree with this change, makes sense for it to work the same way ray trace reflection works. ✔️ 
 
Would be even cooler if the documentation mentioned that this length is taken now after the first intersection.

remi-chapelain (approved):
✔️

JulienIgnace-Unity (approved):

anisunity:
Tests are green locally except the regression that is not related to this pr

PR #6645, opened 2021-12-17
[URP] workaround for DX XR clear issues causing tests to fail

manuele-bonanno:
---
### Purpose of this PR
Workaround for DX clear issues on XR configurations. Those issues need investigation and fix, so for now as a workaround we do not clear before the draw opaques pass on non TBDR platforms

---
### Testing status
tested URP foundation XR tests locally on DX11 and DX12
full Yamato run

---


ellioman (approved):

ellioman:
CI Failures are already known and this PR is reducing the amount of failures significantly.
I'm therefore merging the PR.

PR #6646, opened 2021-12-17
Fix issue with Probe Volume icon appearing when it shouldn't in baking window

FrancescoC-unity:
Discussed with @TomasKiniulis , the PV icon would always appear even if it shouldn't. 

This fixes the issue.


TomasKiniulis (approved):
Thanks! That works out nicely. Getting icon displayed only when Probe Volume component is present in scene

alelievr (approved):

sebastienlagarde (approved):

PR #6647, opened 2021-12-17
Added shader compatibility information to the SSAO page

JordanL8:
### Purpose of this PR
Updated SSAO pge to mention it works with all URP shaders as well as custom opaque Shader Graphs.


0lento:
Since the SSAO doc page is getting updated, might be worth also updating the outdated images and add mention of AfterOpaque -option which never got any mention on the docs? AfterOpaque has been undocumented since it's introduction with SSAO's deferred update.

ellioman (approved):

ellioman:
@0lento Yes, you're absolutely correct. There is a PR incoming soon with improvements for SSAO where we'll address that as well 👍

PR #6650, opened 2021-12-17
Revert "Automatically enable Alpha-To-Coverage when alpha clipping is on (#6599)"

JulienIgnace-Unity:
### Purpose of this PR
Revert "Automatically enable Alpha-To-Coverage when alpha clipping is on (#6599)"


LaserYGD:
Hi, 
was there an issue with the original PR adding this, why was the commit removed and is it coming back?

JulienIgnace-Unity:
Yes it broke some behaviors with double sided shadow flag on mesh renderers. We'll come back to it once the relevant people are back from vacation in early january.

PR #6651, opened 2021-12-17
[Fogbugz#1388961] Fixing PBR Dof for DLSS

kecho:
### Purpose of this PR

* PBR Dof is broken for DLSS. It is using the wrong resolution to compute the min/max. 
* TAAU stabilization pass is not considering if DLSS / TAAU is on . This adds extra stability (less flickering) because of the subsampled depth buffer.
* Fixing post process history size to correctly sample history.

Fogbugz: https://fogbugz.unity3d.com/f/cases/1388961/

Before:
image

After:
image

---
### Testing status
* Tested with fogbugz scene description with Software DRS. 
* Advice QA to test with DLSS on dx12 with hw DRS, forced res and also runtime DRS

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
@@ -88,33 +88,31 @@ namespace UnityEngine.Rendering.HighDefinition
 
         private Vector2Int postProcessViewportSize { get { return resGroup == ResolutionGroup.AfterDynamicResUpscale ? m_AfterDynamicResUpscaleRes : m_BeforeDynamicResUpscaleRes; } }
 
-        private class PostProcessTextureAllocator
+        private struct PostProcessHistoryTextureAllocator
         {
             private String m_Name;
-            private Vector2 m_Scale;
-            private bool m_EnableMips = false;
-            private bool m_UseDynamicScale = false;
+            private Vector2Int m_Size;
+            private bool m_EnableMips;
+            private bool m_UseDynamicScale;
+            GraphicsFormat m_Format;
 
-            public String name { set { m_Name = value; } }
-            public bool enableMips { set { m_EnableMips = value; } }
-            public bool useDynamicScale { set { m_UseDynamicScale = value; } }
-            public Vector2 scale { set { m_Scale = value; } }
-
-            public Func<string, int, RTHandleSystem, RTHandle> allocatorFunction;
-
-            public PostProcessTextureAllocator()
+            public PostProcessHistoryTextureAllocator(String newName, Vector2Int newSize, GraphicsFormat format = GraphicsFormat.R16_SFloat, bool enableMips = false, bool useDynamicScale = false)
             {
-                allocatorFunction = (string id, int frameIndex, RTHandleSystem rtHandleSystem) =>
-                {
-                    return rtHandleSystem.Alloc(
-                        m_Scale, TextureXR.slices, DepthBits.None, GraphicsFormat.R16_SFloat,
-                        dimension: TextureXR.dimension, enableRandomWrite: true, useMipMap: m_EnableMips, useDynamicScale: m_UseDynamicScale, name: $"{id} {m_Name}"
-                    );
-                };
+                m_Name = newName;
+                m_Size = newSize;
+                m_EnableMips = enableMips;
+                m_UseDynamicScale = useDynamicScale;
+                m_Format = format;
+            }
+
+            public RTHandle Allocator(string id, int frameIndex, RTHandleSystem rtHandleSystem)
+            {
+                return rtHandleSystem.Alloc(
+                    m_Size.x, m_Size.y, TextureXR.slices, DepthBits.None, m_Format,
+                    dimension: TextureXR.dimension, enableRandomWrite: true, useMipMap: m_EnableMips, useDynamicScale: m_UseDynamicScale, name: $"{id} {m_Name} {frameIndex}"
+                );
             }
         }
-
-        private PostProcessTextureAllocator m_PostProcessTextureAllocator = new PostProcessTextureAllocator();
 
         // Prefetched components (updated on every frame)
         Exposure m_Exposure;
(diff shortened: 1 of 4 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -1696,13 +1696,13 @@ class TemporalAntiAliasingData
             }
             passData.destination = builder.WriteTexture(dest);
 
-            bool needToUseCurrFrameSizeForHistory = camera.resetPostProcessingHistory || TAAU != camera.previousFrameWasTAAUpsampled;
+            bool needToUseCurrFrameSizeForHistory = camera.resetPostProcessingHistory || upscalingEnabled != camera.previousFrameWasTAAUpsampled;
```
FrancescoC-unity, line 1699: This condition now reads a bit weird 

upscalingEnabled != camera.previousFrameWasTAAUpsampled  I assume will always be true with DLSS on, no?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -1713,7 +1713,7 @@ class TemporalAntiAliasingData
             Vector4 scales = new Vector4(historyRenderingViewport.x / prevHistory.rt.width, historyRenderingViewport.y / prevHistory.rt.height, mainRTScales.x, mainRTScales.y);
             passData.taaScales = scales;
 
-            passData.finalViewport = (TAAU || runsAfterUpscale) ? camera.finalViewport : new Rect(0, 0, RTHandles.rtHandleProperties.currentViewportSize.x, RTHandles.rtHandleProperties.currentViewportSize.y);
+            passData.finalViewport = (upscalingEnabled || runsAfterUpscale) ? camera.finalViewport : new Rect(0, 0, RTHandles.rtHandleProperties.currentViewportSize.x, RTHandles.rtHandleProperties.currentViewportSize.y);
```
FrancescoC-unity, line 1716: These `upscalingEnabled || runsAfterUpscale` checks are redundant for DLSS, here we want to check TAAU explicitly

FrancescoC-unity:
For QA: Can you please do a pass on TAAU only without considering DoF too on this PR?  It is undirectly affected.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -1618,8 +1618,8 @@ class TemporalAntiAliasingData
                 passData.temporalAAMaterial.EnableKeyword("ENABLE_MV_REJECTION");
             }
 
-            passData.runsTAAU = TAAU;
-            if (TAAU && !postDoF)
+            passData.runsTAAU = upscalingEnabled;
+            if (upscalingEnabled && !postDoF)
```
FrancescoC-unity, line 1622: This (and I suspect other of these changes) is wrong. This will flag TAA Upscale mode not necessarily when TAAU is on. I would keep, at the very least for clarity, the checks that are explicitly needed only for TAAU separate.

FrancescoC-unity:
Point of discussion mostly for @pmavridis , I was already uncomfortable about the prpeare pass data is the same for the main TAA pass (which is the same pass for TAAU) and the postDoF one. 
This PR solidifies this feeling; I think we should at least consider splitting the two and factor common code into functions if needed too.

tlaedre:
Testing seems to confirm that it fixes the case reported in 1388961. (I don't know enough about all the possible variants this might  affect to I won't give any explicit approval, though)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
pmavridis: The upscalingEnabled variable seems redundant, since we already have the runsAfterUpscale one. So in all cases above I would replace upscalingEnabled with runsAfterUpscale. Or perhaps I have missed something?

pmavridis:
> Point of discussion mostly for @pmavridis , I was already uncomfortable about the prpeare pass data is the same for the main TAA pass (which is the same pass for TAAU) and the postDoF one. This PR solidifies this feeling; I think we should at least consider splitting the two and factor common code into functions if needed too.

When we first wrote the code there was no DLSS/TAAU, so it made some sense to re-use the same fucntion. I agree that the code is not clean now, so yes, perhaps it's time to split it and clean-up.

remi-chapelain (changes requested):
I tested extensively TAAU / DLSS in hardware and software mode with PBR DoF. Issue seems to be fixed **BUT**
I found one issue specifically when disabling DRS + DLSS, changing from software to hardware then re-enabling DRS (in play mode). The DoF kinda disappear depending on the camera angle and it seems to be linked with the PBR DoF (doesn't repro when it's off)

https://user-images.githubusercontent.com/57442369/146779226-5d814410-2739-4db2-8c93-717697072838.mp4

https://user-images.githubusercontent.com/57442369/146779228-cc801aa6-dcba-4259-a12f-b046ea36fe78.mp4

[Repro.](https://drive.google.com/file/d/140ZtOOyBhzgC6PcyUKRnEPbhll696HFx/view?usp=sharing)

tlaedre:
I also managed to break our project with these fixes. Investigating to see if it was triggered by the same or something different this time.

kecho:
@tlaedre did you manage to discover what exactly broke after you integrated this fix?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.PostProcess.cs
```diff
@@ -2725,10 +2728,11 @@ static void DoPhysicallyBasedDepthOfField(in DepthOfFieldParameters dofParameter
                 {
                     // The sensor scale is used to convert the CoC size from mm to screen pixels
                     float sensorScale;
+                    bool upsampleBeforePost = dofParameters.camera.UpsampleHappensBeforePost();
                     if (dofParameters.camera.camera.gateFit == Camera.GateFitMode.Horizontal)
-                        sensorScale = (0.5f / dofParameters.camera.camera.sensorSize.x) * dofParameters.camera.actualWidth;
+                        sensorScale = (0.5f / dofParameters.camera.camera.sensorSize.x) * (upsampleBeforePost ? dofParameters.camera.camera.pixelWidth : dofParameters.camera.actualWidth);
```
kecho, line 2733: NOTE: replace with post process viewport

kecho: and use height

FrancescoC-unity (approved):

JulienIgnace-Unity (approved):

remi-chapelain (approved):
Tested software and hardware in editor and player. ✔️ 
Tried disabling DoF, enabling DLSS / disabling DRS ✔️ 
Also changing quality settings at runtime works fine ✔️

PR #6652, opened 2021-12-17
[HDRP] Path tracer denoiser 

pmavridis:
### Purpose of this PR
This PR exposes a set of denoising options for the HDRP path tracer. The implementation is using the "Unity Denoising Plugin" package that needs to be installed with the package manager. If the package is installed, then the new denoising options are available in the path tracer UI:
image

The number of the denoisers that are available in the options depends on the "Unity Denoiser Plugin" . 
Currently we expose:
- Intel Open Image Denoise
- NVIDIA Optix denoiser

Example input image (32 samples):
image

Denoised output:
image

## Temporal stability
When recording video sequences with the Unity Recorder, the "temporal" option should be enabled to maximize the temporal consistency of the output. From the available backends, currently only the NVIDIA Optix denoiser supports this option, with some limitations (using AOVs at the same time as temporal mode produces some artefacts).

## Requirements
The Optix 7.4 backend requires an NVIDIA GPU with driver version R495.89 or newer.

For now  you have to install manually the denoising package from this link [FOR INTERNAL USE ONLY]:
https://drive.google.com/file/d/1dxwHF-3iExDUUIVfYwsNJkiogR4jyZ3c/view?usp=sharing
(Unzip the archive and point the package manager to it)

---
### Testing status
Testing the denoisers requires installing the denoising package that will be release separately (as Editor built-in package or in the Unity Registry)

- Tested denoising in scene view, game view, scene+game view side-by-side
- Tested denoising when recording sequences with the Unity recorder.

(TODO: need to add a test scene after we release the package)

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added more explicit error messages when trying to use HDSceneColor, NormalFromHeight, DDX, DDY or DDXY shader graph nodes in ray tracing.
 - Added public API for Diffusion Profile Override volume Component.
 - Added time slicing support for realtime reflection probes.
+- Added denoising for the path tracer.
 
 ### Changed
 - Render Graph object pools are now cleared with render graph cleanup


remi-chapelain (commented):
Did a quick pass for now and a few things has already been fixed. 
Here's a few concerns / notes to keep in mind for the rest of the work. 
- Once the package is out, it would be great to have a way to install it in editor directly (same way we install Nvidia DLSS package). I suggest to have either directly on the override or in the HDRP + DXR Wizard. In the wizard, it can be an optional step to have access to the denoiser. I fear the poor discoverability of the feature if we don't do any of that. If there's no package installed, there's no mention of path tracer denoiser anywhere.
- I know it's not entirely linked to this PR but when you finished accumulating a large sample image, it would be great to be able to switch denoiser without having to re-accumulate the image again. Right now, it automatically triggers the dirtiness of the scene since an asset changed (the serialized volume). I don't know how we could solve this though.
- Also, once the image has converged, currently the denoiser automatically denoise and freeze the frame for a quick second depending on your resolution / performance. So if you have a low sample count, moving on the scene view could be painful due to those quick recurrent lags. There's multiple ways we can solve this, we can have a checkbox to maybe only denoise on the game view ? Or have a "denoise converged frame automatically" checkbox with a "denoise" button that appear when it's off... etc
- In the documentation, when we know more about each filter, it would be great to have a quick pros/cons list of each to help users decide which one to choose depending on their projects
- Radeon filter will probably be discarded in the end, but right now, it lets you believe that the reason it's not supported is your hardware whereas it's just de-activated. 
- When using Intel denoiser, since it does not supports Temporal, it can be removed from the UI

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/PathTracing/PathTracingEditor.cs
```diff
@@ -2,8 +2,8 @@
 
 using UnityEditor.Rendering;
 using UnityEditor.Rendering.HighDefinition;
```
eturquin, line 4: Not from your changes, but this line appears twice, we might as well fix that now. :)

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/PathTracing/PathTracingEditor.cs
```diff
@@ -57,6 +66,35 @@ public override void OnInspectorGUI()
...
+                                    }
+                                }
+                                else
+                                {
+                                    EditorGUILayout.HelpBox($"The selected denoiser is not supported by this hardware configuration.", MessageType.Error, wide: true);
+                                }
+                            }
+                        }
+#else
+                        CoreEditorUtils.DrawFixMeBox("Denoising is not active in the path tracer. To activate it, install the Unity Denoising Plugin package.", MessageType.Info, () =>
```
eturquin, line 93: Suggestion for the message (to reuse the exact same formulation as with DLSS):
"Path Tracing Denoising is not active in this project. To activate it, install the Unity Denoising Plugin package."

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
```diff
@@ -571,6 +698,8 @@ TextureHandle RenderPathTracing(RenderGraph renderGraph, HDCamera hdCamera, Text
             }
 #endif
 
+            pathTracedAOVs.Clear();
```
eturquin, line 701: Couldn't this be moved at the end of the previous ENABLE_UNITY_DENOISING_PLUGIN block?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
```diff
@@ -281,7 +288,18 @@ void ComputeSurfaceScattering(inout PathIntersection pathIntersection : SV_RayPa
...
     #endif
 
+    AOVData aovData;
+    {
+        aovData.albedo = surfaceData.color;
+        aovData.normal = fragInput.tangentToWorld[2];
+    }
+
 #endif // SHADER_UNLIT
+
```
eturquin, line 298: It looks like the shader AOV code is called no matter what (or is the compiler smart enough to prune it when not used?).

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
```diff
@@ -536,6 +636,33 @@ TextureHandle RenderPathTracing(RenderGraph renderGraph, HDCamera hdCamera, Text
         {
             m_PathTracingSettings = hdCamera.volumeStack.GetComponent<PathTracing>();
 
+            var motionVector = TextureHandle.nullHandle;
+            var albedo = TextureHandle.nullHandle;
+            var normal = TextureHandle.nullHandle;
+
```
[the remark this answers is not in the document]
pmavridis: Hmm, I don't see how (at least without rewriting parts of the code).

The idea is that these variables are always passed to the RenderPathTracingFrame functions, but should remain null if we don't need AOVs or if we don't have the denoising plugin at all.

Is there are more clean (or readable) way to do it?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
pmavridis: I don't think that we can rely on the compiler to strip this out. 

As far as I can see, avoiding this would require to introduce a new global material variant "#pragma shader_feature ENABLE_AOV" that we could enable programmatically with Shader.EnableKeyword. However this would double the shader variants, so I'm not sure if this is a good option. 

Keep in mind that part of this computation (certainly the motion vectors) can be moved to the ray-gen shader after we remove the recursion, and in this case we can easily strip-it out. Right now I explicitly avoid the AOV writes in the ray-gen when AOV is not active, but not the computations in the ray tracing shader pass.

So, should we revisit this after the recursion removal or do you think we should do something now?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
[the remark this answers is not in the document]
eturquin: I mean, right after the "// Check the validity of the state before moving on with the computation" test, which returns right away.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassPathTracing.hlsl
eturquin: Yes fair enough, let's address that later if still needed.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/PathTracing/PathTracingEditor.cs
pmavridis: I also got tricked by this one: it is not the same line, one is UnityEditor.* and the other UnityEngine.*
(I removed initially and then I had to revert the commit!)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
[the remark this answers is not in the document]
pmavridis: Oh yes, you are right, the validity test should be first. I was confused because I though "the following block" was referring to only the 3 lines that github shows in the preview. I fixed it now!

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
```diff
@@ -40,6 +40,7 @@ Path tracing uses the [Volume](Volumes.md) framework, so to enable this feature,
...
 | **Maximum Intensity**       | Set a value to clamp the intensity of the light value each bounce returns. This avoids very bright, isolated pixels in the final result.<br />**Note**: This property can make the final image dimmer, so if the result looks dark, increase the value of this property. |
 | **Sky Importance Sampling** | Set the sky sampling mode. Importance sampling favors the brightest directions, which is beneficial when using a sky model with high contrast and very intense spots (like a sun, or street lights). On the other hand, it can be slightly detrimental when using a smooth, uniform sky. It is active by default for HDRI skies only, but can also be turned On and Off, regardless of the type of sky in use. |
+| **Denoising** | Set the denoising mode for the path tracer. This option is only available when the "Unity Denoiser Package" is installed in your project and the available options might change based on the version of this package. The denoising operation can be adjusted by the following parameters: <br> - **Use AOVs**: improves the detail retention after denoising. <br> - **Temporal**: improves temporal consistency when denoising frame sequences.|
```
remi-chapelain, line 43: Maybe here would be a great time to clean the doc? According to our guidelines
- max depth should use "Controls" instead of "Set"
- max intensity should use "Controls"
- Sky Importance Sampling => "Specifies"
- Denoising should use "Specifies"
- Maybe AOV and temporal should have their own lines starting with "When enabled,.. " as the rest of the checkboxes

Also we are missing the layer mask (which should use "specifies" as the rest of the dropdowns)

remi-chapelain:
I did a second pass and here's a few concerns I'd like to talk about before shipping; We probably can't do much about some of those but worth a discussion :) 
- This is probably still expected but the fix button doesn't do anything (package not published yet?)
- When enabling temporal on the nvidia denoiser, the image become darker for some reasons? Is there anything we can do? https://unity.gyazo.com/8b06c01a80f787c898dd5dff961a4f89
- When enabling temporal there seems to be some sort of duplication of the previous / current image superimposed on the screen https://unity.gyazo.com/0f9d3cb9cd96a59398dd9ed97d2a693a
- Most of the time switching from one denoiser to the other (and to none) seems to work without re-accumulating, but not everytime, (still not able to find the repro when that doesn't work) 
- Same for the checkboxes use AOVs and temporal, sometimes it works but not everytime. 
- Using denoiser with scene and game view next to each other makes scene view glitchy https://unity.gyazo.com/d87d90542f7f22c1b4afb1bf6199a129
- In some situations, (try the inside of the template like in the gif), the denoising will pick what it would seem improper data outside the frame leaving you with some border on top and right of the frame https://unity.gyazo.com/df7762f8878227767e88ea0bcf601241
- Also would be cool to add multiple test for thoses if possible (I believe they are deterministic, but I'm worried about the package that could cause problem here)
- Still think we should complete our documentation with:
--- The goal of this feature which would be to clean small remaining noise when the pathtracer is not able to converge (not to have real-time denoised pathtracing). 
--- Providing an example gif with noisy / denoised picture would be great too (I can help with that)
--- Also providing which filter works best in what conditions (i.e what is the reason we provide two filter instead of one)

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
ValGrimm-U3D, line 43: **Denoising** | Enable this feature to denoise the output of the the path tracer.  This feature is only available if the **Unity Denoiser Package** is installed in your project. The parameters available may differ depending on the package version. <br>
There are three denoising libraries available: <br>
- **Intel Open Image Denoise**<br>
- **NVIDIA OptiX**<br>
- **Radeon Image Filtering** (Experimental)<br>
 The default is **Intel Open Image Denoise**.<br>
You can also enable two additional settings:<br>
- **Use AOVs** (Arbitrary Output Variables): Enable this option to increase detail retention after denoising. <br> 
- **Temporal**: Enable this option to improve temporal consistency for denoised frame sequences.<br> 

-----------------
I thought the Denoiser Package was planned to be included with Unity by default, as a core package? That doesn't mean you shouldn't mention that it needs to be installed, just wanted to check this. Also, later we should add a link to the docs for the denoiser package here.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/PathTracing/PathTracingEditor.cs
```diff
@@ -57,6 +66,35 @@ public override void OnInspectorGUI()
...
+                                {
+                                    PropertyField(m_UseAOV);
+                                    if (m_Denoising.value.intValue == (int)DenoiserType.Optix)
+                                    {
+                                        PropertyField(m_Temporal);
+                                    }
+                                }
+                                else
+                                {
+                                    EditorGUILayout.HelpBox($"The selected denoiser is not supported by this hardware configuration.", MessageType.Error, wide: true);
```
ValGrimm-U3D, line 88: selected denoiser --> denoiser selected

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/PathTracing/PathTracingEditor.cs
```diff
@@ -57,6 +66,35 @@ public override void OnInspectorGUI()
...
+                                }
+                            }
+                        }
+#else
+                        CoreEditorUtils.DrawFixMeBox("Path Tracing Denoising is not active in this project. To activate it, install the Unity Denoising Plugin package.", MessageType.Info, () =>
```
ValGrimm-U3D, line 93: Unity Denoising Plugin or Unity Denoiser Plugin?
I've seen both in this OR. 
The first of these two options sounds more grammatical to me.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCameraFrameHistoryType.cs
```diff
@@ -49,6 +49,14 @@ public enum HDCameraFrameHistoryType
         VolumetricClouds1,
         /// <summary>Screen Space Reflection Accumulation.</summary>
         ScreenSpaceReflectionAccumulation,
+        /// <summary>Path traced Albedo AOV.</summary>
+        AlbedoAOV,
+        /// <summary>Path traced Normal AOV.</summary>
+        NormalAOV,
+        /// <summary>Path traced motion vecotr AOV.</summary>
```
ValGrimm-U3D, line 56: vecotr -->  vector

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCameraFrameHistoryType.cs
```diff
@@ -49,6 +49,14 @@ public enum HDCameraFrameHistoryType
...
+        MotionVectorAOV,
+        /// <summary>Denoised path traced frame history.</summary>
```
ValGrimm-U3D, line 58: All of the above:
path traced ---> path-traced

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCameraFrameHistoryType.cs
ValGrimm-U3D: Actually wait, these are just comments, not API doc, huh?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
```diff
@@ -86,6 +114,33 @@ public sealed class PathTracing : VolumeComponent
         [Tooltip("Defines the maximum, post-exposed luminance computed for indirect path segments. Lower values help against noise and fireflies (very bright pixels), but introduce bias by darkening the overall result. Increase this value if your image looks too dark.")]
```
ValGrimm-U3D, line 114: help against ---> help prevent

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
```diff
@@ -86,6 +114,33 @@ public sealed class PathTracing : VolumeComponent
...
+        /// <summary>
+        /// Enables denoising for the converged path tracer frame
+        /// </summary>
+        [Tooltip("Enables denoising for the converged path tracer frame")]
+        public DenoiserParameter denoising = new DenoiserParameter(HDDenoiserType.None);
+
+        /// <summary>
+        /// Improves the detail retention after denoising by using albedo and normal AOVs.
+        /// </summary>
+        [Tooltip("Improves detail of the denoised image by using albedo and normal AOVs")]
```
ValGrimm-U3D, line 128: detail --> the detail
by using --> with its

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
```diff
@@ -86,6 +114,33 @@ public sealed class PathTracing : VolumeComponent
...
+        [InspectorName("Use AOVs")]
+        public BoolParameter useAOVs = new BoolParameter(true);
+
+        /// <summary>
+        /// Enables temporally stable denoising (not all denosing backends support this option)
+        /// </summary>
+        [Tooltip("Enables temporally stable denoising")]
```
ValGrimm-U3D, line 135: temporally-stable

ValGrimm-U3D (changes requested):
I have noted the changes requested in my comments.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCameraFrameHistoryType.cs
pmavridis: Actually it's published as API doc in the package, so I'm fixing everything. Thanks for the comments.
(This is how we do scripting API docs in packages, hence the confusion in our meeting earlier today ;) )

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
remi-chapelain: Altough, there's no Radeon filter anymore in the latest change so I think it has been removed for good.

pmavridis: @ValGrimm-U3D can you confirm these changes (the ones not related with denoising)? For the denoising I have already applied your changes.

pmavridis: @ValGrimm-U3D I have rephrased this a bit, because as Remi said we expose only the two backends in HDRP (for now) and there is also the None option in the dropdown (which is the default). Can you please check again? thanks.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/PathTracing/PathTracingEditor.cs
pmavridis: I think I have Denoising in all publicly visible places right now. Before release we have the flexibility to change that. 
In the scripting API, we do use "Denoiser" for the class that performs the denoising.

pmavridis:
@remi-chapelain Here are a few quick answers on the things that you have noted:
(for some of these fixes you need a new package)

- The fix button needs to be verified with the Editor that will include the built-in package.
- The bug regarding resetting iterations sometimes when switching the denoisers is now fixed (also fixes a related bug when increasing sample count)
- It's expected to reset the path tracing iterations when toggling AOVs or Temporal mode.
- The bug when game-view and scene-view are side-by-side is fixed (was related to rthandle scale).  
- The bug with the denoiser leaking information on the edges from outside the frame is fixed.
- Regarding your comments on documentation, I agree that we need a general user guide regarding the denoising feature. We have discussed this with @ValGrimm-U3D and we plan to add some documentation in the Unity Manual (outside HDRP) since this is common infrastructure between HDRP and the lightmapper (so both can link to it).

I still need to fix the bug related to the temporal mode (motion vectors when rthandle scale is not 1). But we cannot control the color/exposure shift that happens when enabling this mode.

com.unity.render-pipelines.high-definition/Documentation~/Ray-Tracing-Path-Tracing.md
```diff
@@ -40,7 +40,14 @@ Path tracing uses the [Volume](Volumes.md) framework, so to enable this feature,
...
-| **Denoising** | Set the denoising mode for the path tracer. This option is only available when the "Unity Denoiser Package" is installed in your project and the available options might change based on the version of this package. The denoising operation can be adjusted by the following parameters: <br> - **Use AOVs**: improves the detail retention after denoising. <br> - **Temporal**: improves temporal consistency when denoising frame sequences.|
+| **Denoising** |Enable this feature to denoise the output of the the path tracer. This feature is only available if the **Unity Denoising Package** is installed in your project. There are three available options in HDRP: <br> 
+- **None**: Disables denoising (this is the default option).
+- **Intel Open Image Denoise** : Performs denoising using the Intel Open Image Denoise library
+- **NVIDIA OptiX** : Performs denoising using NVIDIA OptiX.
+
+You can also enable two additional settings:
+- **Use AOVs** (Arbitrary Output Variables): Enable this option to increase detail retention after denoising.
+- **Temporal**: Enable this option to improve temporal consistency for denoised frame sequences.|
```
ValGrimm-U3D, line 50: Looks good.

 Two small punctuation/spacing things:

- Just in case it causes a formatting issue, I'd add a space between the first table border and "Enable" and the other table border and "sequences."
- For consistency, it would be good to add a period to the Open Image Denoise line.

ValGrimm-U3D (approved):
Looks good to me. Thanks!

sebastienlagarde (approved):

PR #6653, opened 2021-12-19
Shader graph spacewarp

sholms:
It work perfectly with lit graph, but unlit graph doesn't work.


sophiaaar:
Thanks for your PR, but this does not meet the guidelines laid out in https://github.com/Unity-Technologies/Graphics/blob/master/CONTRIBUTING.md

PR #6656, opened 2021-12-20
Update vfx window title when asset is renamed

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1376814/

When a VFX asset is renamed while a it's opened but the window is not active, the VFX window title was not updated.
Steps to reproduce:
- Open a VFX asset (make it docked next to scene tab)
- Select the Scene tab (so that the VFX window is not visible anymore)
- In the project browser, rename the opened VFX asset
=> The VFX window title is not updated

---
### Testing status
Only scenario described above has been tested

---
### Comments to reviewers
Before this fix, the VFX window title was correctly updated if it was not active but visible.
So the issue would occur only if it was hidden behind other tabs.

### DIFF
--- a/com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs
+++ b/com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs
@@ -19,7 +19,7 @@ namespace UnityEditor.VFX
     {
         static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths)
         {
-            List<string> assetToReimport = new List<string>();
+            List<string> assetToReimport = null;
 
             foreach (var assetPath in importedAssets)
             {
@@ -39,6 +39,7 @@ namespace UnityEditor.VFX
                             graph.SanitizeForImport();
                             if (!wasGraphSanitized && graph.sanitized)
                             {
+                                assetToReimport ??= new List<string>();
                                 assetToReimport.Add(assetPath);
                             }
                         }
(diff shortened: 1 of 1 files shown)


VitaSkruibyte (approved):
Tested: 

1. Rename in Project Browser of graphs, subgraphs blocks and operators
2. Rename in Edit → Rename does not work at all with assets ;(  (just for visibility that tried checking this as well)
3. Checked if name is updated in both VFX tab, and in Inspector
4. Tried entering/getting back to parent graph to see if the name in the VFX tab is updated

com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs
```diff
@@ -18,7 +19,7 @@ class VFXGraphPreprocessor : AssetPostprocessor
     {
         static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths)
         {
-            List<string> assetToReimport = null;
+            List<string> assetToReimport = new List<string>();
```
PaulDemeulenaere, line 22: OnPostprocessAllAssets is called for every import, we should always prevent any unecessary allocation in this callback. Even if the impact is minimal, the memory profiling will be noisy for anyone trying to debug an asset import.
We should definitely let the lazy assetToReimport invocation here.

com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs
```diff
@@ -38,15 +39,19 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
...
-                                    assetToReimport = new List<string>();
                                 assetToReimport.Add(assetPath);
                             }
                         }
                         catch (Exception exception)
                         {
                             Debug.LogErrorFormat("Exception during sanitization of {0} : {1}", assetPath, exception);
                         }
+
+                        var window = VFXViewWindow.GetWindow(graph, false, false);
```
PaulDemeulenaere, line 50: Just a confirmation: can't we have several window opened on the same graph ?

PaulDemeulenaere (changes requested):

PaulDemeulenaere (changes requested):
The UpdateTitle is fine, I only have a concern about the allocation guard in OnPostprocessAllAssets

com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs
julienamsellem: No, only one window for a graph. If you try to open it a second time it will only bring the currently opened window in front.

com.unity.visualeffectgraph/Editor/Models/VFXGraph.cs
julienamsellem: Ok, reverted

PaulDemeulenaere (approved):
LGTM! The modification of windows title doesn't affect the content of the VisualEffectAsset.

Side Note for later: it's strange to retrieve the filename from `controller.name` [here](https://github.com/Unity-Technologies/Graphics/blob/e6b781c726a0cdc6c85f5b1f53efaf5f38af6d09/com.unity.visualeffectgraph/Editor/GraphView/VFXViewWindow.cs#L357).
There is maybe a lack of clear responsability here, not related to this PR and probably will be refactored with next GTF integration.

PR #6659, opened 2021-12-21
[HDRP] Fix issue with dynamic RendererList culling option getting ignored

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1388854
The RendererList culling option was not being initialized properly in rendergraph.

The fix is trivial, but it shows that we are missing a test for this on the HDRP side. 
Not sure how to test such feature, maybe it's time to enable it by default and have the perf metrics capture the regression? 

---
### Testing status
Open The HDRP template, Open the RenderGraph Viewer and in the filters select to show culled passes:
Now toggle the "Dynamic render pass culling" option in the HDRP global settings.
- Before: the number of culled passes stays the same, regardless of dynamic culling.
- After: the number of culled passes is now higher and changes with the view.


alelievr (approved):

RemyUnity (approved):
LGTM, I don't see what could be tested more.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #6660, opened 2021-12-21
Fixed output add button could become invisible

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1389927/

In a VFX subgraph, the output add button (+) in the blackboard can become invisible when the output category has a visible scrollbar.

[Before]
VFX_UX_SubgraphOutputBoard

[After]
UPLQRsGtFl

---
### Testing status
- Resize the blackboard to its minimum size
- Resize the blackboard so that an horizontal scrollbar is visible and move the scrollbar

---
### Comments to reviewers
The original issue is rather in the scope of GraphView because the same issue happens in ShaderGraph.
But since there's no more development on GraphView I fixed it specifically for VFX (the fix is a hack ...)


MarieGuff (approved):
Tested and approved on 2021.2.0a15 (37767e16dc73)

VitaSkruibyte (approved):
Auto approving as the fix is cosmetic and been already verified by Julien and Marie. Thanks for the fix!

PaulDemeulenaere (approved):
Ok, this is a workaround from the GraphView behavior, offsetting the button position according to the scroller cursor. Only cosmetic, the risk is pretty low. 
LGTM. Thanks for the fix.

PR #6661, opened 2021-12-21
[HDRP] Add time slicing to realtime reflection probes

simonb-unity:
### Purpose of this PR

This PR adds a time slicing feature to realtime reflection probes. This is a new checkbox, only shown when the probe is realtime, and supports all realtime modes.

image

When this setting is enabled:
- Updates to the reflection probe take 7 frames: one for each cubemap face, then one to update the reflection probe cache (which performs convolutions + compression)
  - Since the scene can be different for each cube face, this can result in a seam between cubemap faces, so may not be appropriate for all use cases
  - Each reflection probe handles its updates independently, there is no global coordination of work in scenes with multiple realtime reflection probes.
- If triggered by OnDemand or OnEnable while already mid-update, another update will occur immediately after the current one completes. (So triggers are not lost, but they can be delayed.)

---
### Testing status

Manually tested in a scene with multiple realtime reflection probes, and a planar probe (which should be unchanged since only reflection probes support time slicing). Checked for the expected performance improvement, checked the cache updates every 7 frames with some temporary debug logging.

---
### Comments to reviewers

The changes from #6697 are required for correctness, these are currently included in the branch.

Some notes on the changes:
- The time slicing is handled by the `ProbeRenderSteps` flags enum, handling one set bit per frame.
- Changed the HDRP texture cache to require the caller to pass in the content hash, this was a TODO in the previous code so hopefully looks reasonable
- The content hash is now a custom "render count" instead of the core texture "update count", as this seemed to be the simplest way to keep caching the old version until we hit the frame where we want the update.


EagleGamer2005:
Thanks 👍 I was waiting for this!!! Can we have one more option of time slicing frame period if possible which can help us to set frame period from 7 to any number as more optimization options??? And can we also have the reflection update settings out of the box 😊??? thanks again for bringing time slicing back!!

EagleGamer2005:
Thanks 👍 I was waiting for this!!! Can we have one more option of time slicing frame period if possible which can help us to set frame period from 7 to any number as more optimization options??? And can we also have the reflection update settings out of the box 😊??? thanks again for bringing time slicing back!!

simonb-unity:
> Thanks 👍 I was waiting for this!!! Can we have one more option of time slicing frame period if possible which can help us to set frame period from 7 to any number as more optimization options for reflection probes which remain static??? And can we also have the reflection update settings out of the box 😊??? thanks again for bringing time slicing back!!

Hi, thanks for the comments. For the initial version of this feature I think we should limit the changes to parity with built-in and URP, so only support time slicing with a single cube face per frame. It would definitely be possible to add more time slicing modes in future, but probably best to consider any new features after the initial version is done.

JMargevics (approved):
Things tested: 

- Setup a scene from HDRP template, switched one of the bigger reflection probes to dynamic mode and reflected animated object
- Performance with or without Time Slicing
- DX11, 12, Vulkan
- Switching timeSlicing in Runtime
Everything works as expected and as we see on the video it gives a really substantial performance boost. After covering this PR in automation it's good to merge from my side 👍🟢

https://user-images.githubusercontent.com/50581102/151995259-884c7fe6-01c8-4669-a2fd-25ab7a14d5f0.mp4

sebastienlagarde (approved):

PR #6663, opened 2021-12-22
Fixed NRE exception

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1389076/

How to reproduce:
1. Open the attached VFX Graph and drag it to the Scene
2. Attach it to the GameObject and Select Init context
3. Observe that Gizmos are invisible due to Arc Cone's space being set to World

👍  **Expected result**: Gizmos are visible when changing spaces of the block
👎  **Actual result**: Gizmos are invisible and errors are thrown to the console

Reproduced with: 2021.2.4f1 (12.1.1) , 2022.1.0b2(13.1.3)

Error message:

> NullReferenceException: Object reference not set to an instance of an object
> UnityEditor.VFX.UI.VFXDataAnchorGizmoContext+<>c__DisplayClass11_0.<BuildValue>b__3 (System.Collections.Generic.List`1[T] o) (at Library/PackageCache/com.unity.visualeffectgraph@13.1.3/Editor/GraphView/Elements/Controllers/VFXDataAnchorController.cs:742)

[Before]
Screenshot 2021-12-16 at 15 03 42

[After]
Unity_d8qSfp7hbM

---
### Testing status
Tested the scenario described above

---
### ⚠ Comments to reviewers
This PR does not really fix the issue because two different bugs are behind this issue.
1- There's an exception raised in the VFX code which prevents to run further instructions
2- There's an issue (a warning message is displayed in the console) with quaternion preventing the gizmo from being displayed
Here is the console message:
> Look rotation viewing vector is zero

The current PR fixes the first issue. The second one seems to be fixed in trunk (2022.2-a01).
So with this PR it should be completely fixed in 2022.2.

To backport the fix we need to backport this PR off course, but we also need to identify the PR that fixed the second issue to backport it too.
Currently the lead we found for this is this [bug](https://fogbugz.unity3d.com/f/cases/1284682/). The symptoms are similar but it seems to be related to Slider1D when in our case it's Slider2D.


VitaSkruibyte (changes requested):
NRE is no longer present and gizmos are shown but bumping into issue where gizmos are not updated when alongside with values changed (this happens only when operator like pos or vector is connected to the Arc Shape block)

Also noticed that values displayed in the block are dashes (-), shouldn't it reflect Pos operator values? 

NP5UL1C8Hi

VladNeykov:
> NRE is no longer present and gizmos are shown but bumping into issue where gizmos are not updated when alongside with values changed (this happens only when operator like pos or vector is connected to the Arc Shape block)
> 
> Also noticed that values displayed in the block are dashes (-), shouldn't it reflect Pos operator values?
> 
> NP5UL1C8Hi

I manager to reproduce this issue as well on:
> 2022.2.0a1.176
> Revision: trunk 60f583a20d13
> Built: Thu, 09 Dec 2021 20:16:09 GMT

Curiously however, it seems to work fine in latest trunk:
> 2022.2.0a1.368
> Revision: trunk 2dd09014126d
> Built: Mon, 03 Jan 2022 13:31:39 GMT

So looks like we are good :)

VitaSkruibyte:
Double checked on a newly created VFX. Seems to be still reproducible. Side note when Pos node is connected for the first time, gizmos values from block are not reflected in the Scene View, reconnecting resolved the issue, but I believe this is something to look into as well.


Reproduced using:
2022.2.0a1.379
Revision: trunk 47d511408a9a
Built: Wed, 05 Jan 2022 16:12:58 GMT

image

julienamsellem:
I did another change so that the gizmo is updated when the space (from a connected operator for instance) is changed.
The gizmo is supposed to be shown when the space is set to world, but today it's also shown in local space (and is not updated when position changed).
Preventing the gizmo to be shown in local space is planned to be fixed in the scope of another PR (None space)

com.unity.visualeffectgraph/Editor/GraphView/Elements/Controllers/VFXDataAnchorController.cs
```diff
@@ -123,6 +123,10 @@ protected override void ModelChanged(UnityEngine.Object obj)
             UpdateInfos();
             Profiler.EndSample();
 
+            if (m_GizmoContext != null)
+            {
+                m_GizmoContext.Unprepare();
+            }
```
PaulDemeulenaere, line 129: It should be `RefreshGizmo` here, however, this function is called everytime a value is changed...
Do you have any update for this ? I recall we discussed about caching the expression hash to check if it changed, not sure it's a viable solution 🤷

PaulDemeulenaere (changes requested):
Adding note from previous conversation

com.unity.visualeffectgraph/Editor/GraphView/Elements/Controllers/VFXDataAnchorController.cs
julienamsellem: Yes, works as expected 👍

com.unity.visualeffectgraph/Editor/GraphView/Elements/Controllers/VFXDataAnchorController.cs
```diff
@@ -123,6 +125,18 @@ protected override void ModelChanged(UnityEngine.Object obj)
...
 
+            if (m_GizmoContext != null)
+            {
+                HashSet<VFXExpression> expressions = new HashSet<VFXExpression>();
+                model.GetExpressions(expressions);
+
+                var currentExpressionHashCode = UIUtilities.GetHashCode(expressions);
+                if (currentExpressionHashCode != m_expressionHashCode)
+                {
+                    RefreshGizmo();
```
PaulDemeulenaere, line 136: I would add a comment here (for later purpose, if we are able to get more information from ModelChanged) :
```suggestion
                    //The ModelChanged will be triggered when a parent value is modified, detecting if there is a change in graph (link or change space) using expression hash to prevent the invalidation all cached gizmos
                    RefreshGizmo();
```
Feel free to edit or ignore.

PaulDemeulenaere (approved):
Checked locally the expected behavior:

https://user-images.githubusercontent.com/22494706/152553672-24ed8078-5e5e-4b6e-9a11-71da7a7cff29.mp4

And also double checked when the RefreshGizmo is called :

https://user-images.githubusercontent.com/22494706/152554548-487bec7f-4c44-44ac-b67d-b3a9d91a183e.mp4

Thanks for the fix, it makes the whole system more stable.

com.unity.visualeffectgraph/Editor/GraphView/Elements/Controllers/VFXDataAnchorController.cs
julienamsellem: Done, with slightly different message

PR #6664, opened 2021-12-22
Empty inspector when user clicks in "void"

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1364096/
https://fogbugz.unity3d.com/f/cases/1364097/

## First bug [1364096](https://fogbugz.unity3d.com/f/cases/1364096/)
Steps to reproduce the issue:
- Open Any VFX Graph asset
- Select a node
- Notice its display in inspector
- Click into empty space to remove selection
- Notice it's still displayed in the inspector

While it's could be convenient to keep the displayed node, it somehow creates confusion as the user don't know anymore which node/block/context is being edited.

With this PR, now when clicking in the void, the opened VFX Graph (or subgraph) is shown in the inspector.

[Before]
TnIfqNKSp3

[After]
qelyfTHnJx

---
### Testing status
- Checked un-selecting contexts, blocks and operators
- Checked when entering a subgraph and then going back to parent graph

## Second bug [1364097](https://fogbugz.unity3d.com/f/cases/1364097/)

When having a context selected in the Graph, then selecting a VFX asset, the inspector updates to the VFX asset, but when re-selecting the same output it doesn't actualise the inspector as the output was already selected, so we have to click somewhere else and re-click on it. 

Steps to reproduce: 
- Create a new VFX and select a context
- Without unselecting the context, click on your VFX asset in the project folders
- see inspector change
- re-lick on the same context in your graph, inspector doesn't change
- click on something else in the graph and re-click on the same context as before, inspector updates

👎 Actual result: user needs to unselect and then reselect the block
👍 Expected result: when selecting the block, the inspector updates

[Before]
Xvf9YBaniu

[After]
ZtgKbZnwDl

---
### Comment to reviewers
These two bugs have been grouped together because they are linked, the fix for the second one depends on the fix for the first one.


VitaSkruibyte (approved):
Thank you for the very detailed and informative PR description. The PR looks good 🟢 , familiarized with the issues and double checked reproductions. For both issues additionally checked:
1. Behavior when Inspector is locked. It is indeed locked and windows do not update until unlocked.
2.  (1364096) Multiple windows that are opened side by side, and clicked around the blocks/contexts/operators/blackboard properties/ subgraphs/subgraph blocks/subgraph operators -> empty space in the graph and vice versa. Results are consistent among actions.
3.  (1364097) Various combinations of reproduction using subgraph blocks, operators, graphs and contexts/blocks/operators

MarieGuff:
There's a very fast glitch when opening the output's inspector on 2022.2.0a1_2dd09014126d

MarieGuff (approved):
Great improvment, thank you!
Tested on 2022.2.0a1 (2dd09014126d)

com.unity.visualeffectgraph/Editor/GraphView/Views/VFXView.cs
```diff
@@ -1990,7 +1990,6 @@ public void UpdateGlobalSelection()
 
             if (objectSelected.Length > 0)
             {
-                Selection.objects = objectSelected;
                 Selection.objects = objectSelected;
```
PaulDemeulenaere, line 1993: Good catch, it was probably an incorrect default merge conflict ⬆️

PaulDemeulenaere (approved):
Ok, the code review won't really help to identify all cases, looks good to me.
(I was simply wondering if Selection.objects supports `null` but it seems to be the case)

julienamsellem:
VFXGUI tests run locally 🟢

PR #6668, opened 2021-12-23
Shrink operator if possible (when changing data type)

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1364445/

Repro steps:
- Create a multiply operator
- Click on the gear icon ⚙ and change the type for input 'A' to a 'Vector4'
- Change it back to 'Float'
- Observe the operator remains large

👍 Expected result: the operator shrinks back to the min size required to display a Float value
👎 Actual result: the operator keeps a much larger size than needed

With this fix, the operator will shrink back to the minimum width only when the "edit ⚙" mode is ended.

[Before]
ei6p7BhpBx

[After]
ZgjIugnaVC

---
### Testing status
- Switch from `float` => `vector3` => `float`
- Checked collapse and super-collapse interaction

---
### Comments to reviewers
Seems to happen for all operators (add, branch, noise, etc.)
Note that collapsing the operator and then expanding it was already resizing the operator to its minimum size and also it ends the "edit ⚙" mode.

*This is somehow a regression introduced by the [fix](https://github.com/Unity-Technologies/Graphics/pull/5018) I did to keep the operator width constant, even when it's collapsed.
But I think the behavior is even better now because the operator does not change size every time you change the type of an input (while in edit mode) only when you leave the edit mode.*


VitaSkruibyte:
Checked different types of operators like Switch, Probability sampling. in graphs, subgraphs blocks and subgraph operators

1.  **❌** When switching to ‘larger operators’ ( bool → Vec2 → Vec4 let’s say) size changes immediately. If I switch from larger to smaller operator (Vec4 → float) it doesn’t happen immediately, I have to click gear icon to make it update to the smaller operator.

PTei2WZrKD

2.  **❌** Connecting operator to another operator when operators’ type does no match. There seem to be an issue with this behavior.
<pre>
 Example:
       1. Create Arc Cosine and float operators
       2. Change Arc Cosine to vector4 type and connect float to it
       3. Observe that Arc Cosine type switches from vector to float but operator remains large with xyzw inputs
</pre>
pPkUyh2xRD

3.  **✔️** Collapsing/expanding and changing the type of operator

VitaSkruibyte (approved):
Did a quick verification and confidence check. Couldn't find any outstanding issues. Thank you for taking care of this!

PaulDemeulenaere (approved):
LGTM! It's pretty minor, thanks for the fix.

PR #6669, opened 2021-12-28
[Fogbugz # 380357] Fixing negative overflowing of IES attenuation spline

kecho:
### Purpose of this PR
IES profiles that have splines going slightly below 0 or above 1 will cause some byte overflow artifacts.
This change clamps the evaluated spline to be within 0 and 1 range

https://fogbugz.unity3d.com/f/cases/1380357/

Before: 
image

After:
image

---
### Testing status
Tested with the fogbugz example.


TomasKiniulis (approved):
Looks good. Note that the changes only apply when library is deleted.

sebastienlagarde (approved):

PR #6672, opened 2022-01-03
Updated decal docs

Jonasmortensen:
This PR updates images and text in the URP decal documentation


oleks-k (approved):
Approving with the changes.

lukaschod (approved):
Looks good, also really like decal atlas idea. Maybe we should even consider this as feature and making more automatic

PR #6673, opened 2022-01-03
Re-added menu item converting selected materials

martint-unity:
---
### Purpose of this PR
Adding back a menu item for converting selected Materials from built-in to URP.
We removed it awhile back and we got feedback that we should return it. :) So we are doing that with this PR.

image


pbbastian (changes requested):
Code looks good, but the `Window` menu should only be for opening windows such as the Package Manager or Frame Debugger.

How about in the `Assets` menu? Then it would also be visible when right clicking a material. In that case, it should probably also have a `MenuItem` with `isValidateFunction = true` validating that the asset is a material so that the item cannot be clicked for non-materials.

pbbastian (approved):

phi-lira (approved):

PR #6677, opened 2022-01-04
Remove debug probes from depth prepass 

FrancescoC-unity:
Before this PR the debug probes were rendered in the depth prepass, causing them to cast contact shadows, affect SSAO etc. etc. which is not something we want for a debug view.

**What did I test**: verified that SSAO is not affected anymore, all other effects should be fixed too.


iM0ve (approved):
Small change, small halo effect. Testing should be sufficient.

sebastienlagarde (approved):

PR #6678, opened 2022-01-04
Fix sky jittering

adrien-de-tocqueville:
### Purpose of this PR

Fix https://fogbugz.unity3d.com/f/cases/1388187/

Skies and cloud layer were jittering when TAA is enabled.
GIF with HDRI sky and cloud layer before:
jitter

and after:
jitter_fixed

---
### Testing status

tested the repro steps
launched yamato


FrancescoC-unity (approved):

JMargevics (approved):
Looks like the current test coverage is enough. Approving as is 🟢

PR #6690, opened 2022-01-06
Fix compilation error when VR module is disabled in the built-in Unity modules

maloyer-unity:
### Purpose of this PR
Fix compilation error when VR module is disabled in the built-in Unity modules.

---
### Testing status
Using the repro steps in the bug report: https://fogbugz.unity3d.com/f/cases/1389160/
1. Create a new project
2. Disable VR Built-in package
3. Install Post Processing package
4. Observe the Console output

Tested compilation of the package when disabling built-in VR module in Package Manager (and with VR module re-activated afterward)


com.unity.postprocessing/PostProcessing/Runtime/PostProcessLayer.cs
```diff
@@ -723,6 +724,7 @@ void OnPostRender()
                                 m_Camera.projectionMatrix = m_Camera.GetStereoProjectionMatrix(Camera.StereoscopicEye.Left);
```
mikesfbay, line 724: This issue was introduced by the temporalAA projection matrix fix from @fcoulombe. Fyi, all XRSettings calls should be #if with (ENABLE_VR_MODULE && ENABLE_VR) (like in line 8 of PostProcessLayer).

mikesfbay:
Loop in QA to confirm the fix @ryanhy-unity

fcoulombe:
LGTM. (I can't receive my github two factor auth sms for Unity Technologies
and I don't wanna reset my password)

maloyer-unity:
@ryanhy-unity, I would like to know if you have the time to look at this issue (or if anyone else could have a look). 

This is currently the highest user pain issue in our bucket and I would like us to be able to push that fix as soon as we can.

ryanhy-unity (approved):
I was able to reproduce the issue locally and then confirm the changes here fix the issue. Just tested on Windows but given the nature of changes here I think you got it.  Might want Sam to just double check Xbox?

maloyer-unity:
@mikesfbay, is there anything else needed for this review ?

mikesfbay (approved):
LGTM. Sebastien can add to next merge/release.

PR #6691, opened 2022-01-07
Various features for the water system

anisunity:
Here is the list of the major changes in this PR:
- Reduced the size of the GBuffer to 128 bits
- Signficantly improved the foam rendering
- Change of approach for the patch management and tesselation (GPU pipeline)
- Big clean up of the shader graph
- Various small bug fixes

image
https://youtu.be/3C282nWnFBE

Testing status:
Tried to tess every combination all options, didn't find any bug.
The approach for the debug display needs to be reviewed.
The test scene needs to be modified to add foam and caustics testing.

### DIFF
--- a/TestProjects/HDRP_Tests/Assets/Tests/TestFilters/TestCaseFilters.asset
+++ b/TestProjects/HDRP_Tests/Assets/Tests/TestFilters/TestCaseFilters.asset
@@ -112,3 +112,12 @@ MonoBehaviour:
     XrSdk: 
     StereoModes: 0
     Reason: unstable with cache server
+  - FilteredScene: {fileID: 0}
+    FilteredScenes:
+    - {fileID: 102900000, guid: fa0d54c5ccbc8b143a8db100374e0f5e, type: 3}
+    ColorSpace: 1
+    BuildPlatform: -2
+    GraphicsDevice: 16
+    XrSdk: 
+    StereoModes: 0
+    Reason: API Bug
--- a/com.unity.render-pipelines.high-definition/Runtime/Water/HDRenderPipeline.WaterSystem.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Water/HDRenderPipeline.WaterSystem.cs
@@ -470,7 +470,10 @@ namespace UnityEngine.Rendering.HighDefinition
             parameters.rotation = -currentWater.transform.eulerAngles.y * Mathf.Deg2Rad;
             parameters.foamMaskOffset = currentWater.foamMaskOffset;
             parameters.waterMaskOffset = currentWater.waterMaskOffset;
-            if (currentWater.infinite)
+
+            // For now, we are disabling infinite surfaces for metal as there seems to be a bug in the API that
+            // breaks on the DrawMeshInstancedIndirect function.
+            if (currentWater.infinite && SystemInfo.graphicsDeviceType != GraphicsDeviceType.Metal)
             {
                 parameters.infinite = true;
                 parameters.targetMesh = m_TessellableMesh;


remi-chapelain (commented):
_TESTING STILL IN WIP_

- Foam is looking infinitely better now with the new dragging system ✔️ 
- Shader graph is cleaner too ✔️ 

**Not related to this PR (also happens on master) but would be cool to fix.** 
- ❌ Simulation caustics don't travel in the right direction, they are rotated the other way than the waves. 
https://user-images.githubusercontent.com/57442369/148804719-c2f25247-4ec4-4277-a46a-9e3f30aab01a.mp4

**Related to this PR**
- ❌ By Default Tesselation parameters in the volume aren't visible because they're additionnal parameters. Proposed solution, include section title in the additional parameters to avoid having an empty section ? 
https://user-images.githubusercontent.com/57442369/148804789-a4a5ec13-1b69-4a2b-91d9-6da00526bfb5.mp4

- ❌ When pressing Play / Stop on a simulation there's an errors related to water lighting compute shader
image

- ❌ Foam "intensity/opacity" is great when absoption distance is low but as soon as the water becomes very transparent, the foam follows too and I believe the foam should stay more or less opaque no matter the water absoption distance
https://user-images.githubusercontent.com/57442369/149097339-0d0c697b-1b22-4356-8052-5ad3d9cfd985.mp4

- ❌**FIXED** ~~Can't move infinite water surface vertically~~

- ⚠️**FIXED** ~~Tesselation max on console is 15, so either write it in the doc (like in the LitTesselation shader) or change the max at 15 for all~~

- ⚠️Foam drag tooltips has no real info. Proposed solution (to be refined)
_Controls how much foam accumulates after being spawned. A value of zero means no accumulation. A value of one means foam will stay longer at its spawned position leaving a foam trail._

- ⚠️CONCERN: For tesselation we propose a max tesselation factor, a start fade distance and a range. In the Lit Tesselation we have the same but instead of start and range, we have start and end. While I agree this pattern is probably better (start + range), I don't like the fact that two same feature as presented with different setup / input parameters. 

- ⚠️CONCERN: Right now, with the PR, the only way to turn off Tesselation is to put max Tesselation at zero. Just want to make sure that it's something we don't want to turn off since it's not 100% obvious how to do it (meaning it's not a checkbox or something more clear)

**Also possible added bonus request already discussed**
- Add the possibility to kill the bands separately in the shader graph by having a band multiplier vector4 as an input to the water simulation evaluation nodes. 
- Add a default LOD (to fill the plane up to far plane) after the max LOD to avoid defaulting on "nothing" after the last LOD when having an infinitewater surface. 
https://user-images.githubusercontent.com/57442369/148804626-35e971ca-efd6-4b6d-af34-f26c39bcc50d.mp4

For the rest of the addition (GBuffer packing), I will trust dev's judgement.

remi-chapelain (approved):
Most of the issues mentionned in my previous moment are fixed (see details above). 
I modified the test to include tesselation / foam and caustics. 
Screenshots needs to be updates (yamato will fail right now)

The rest will be taken care of later :).

com.unity.render-pipelines.high-definition/Editor/Material/Water/ShaderGraph/Tessellation.template.hlsl
```diff
@@ -0,0 +1,59 @@
...
+    // evaluate vertex graph
+    VertexDescription vertexDescription = VertexDescriptionFunction(vertexDescriptionInputs);
+
+    // The set of vertex description outputs that are used are fixed
+    positionOS = vertexDescription.Position;
+    normalOS = vertexDescription.Normal;
+    uv0 = vertexDescription.uv0;
+    uv1 = vertexDescription.uv1;
+
+    $splice(CustomInterpolatorVertMeshCustomInterpolation)
```
sebastienlagarde, line 56: is this one expected, should it be like in the tessellation.template.hlsl?

com.unity.render-pipelines.high-definition/Runtime/Material/Water/Water.hlsl
```diff
@@ -124,73 +124,123 @@ BSDFData ConvertSurfaceDataToBSDFData(uint2 positionSS, SurfaceData surfaceData)
 // conversion function for deferred
 
 // Layout of the water gbuffer
-// GBuffer0 R16G16B16A16 (64 bits)
-// - GBuffer0: Bits 64 -> 32: Diffuse color (11 11 10)
-// - GBuffer0: Bits 31 -> 16:  Caustics
-// - GBuffer0: Bits 15 -> 0: foam value
+// GBuffer0 R16G16 (32 bits)
+// - GBuffer0: Bits 31 -> 0: Diffuse color (11 11 10)
```
sebastienlagarde, line 128: why not used 111110f format here rather than R16G16?

sebastienlagarde: will also avoid to do instruction in shader code

com.unity.render-pipelines.high-definition/Runtime/Water/WaterVertexTessellation.hlsl
```diff
@@ -0,0 +1,225 @@
...
+[domain("tri")]
+PackedVaryingsToPS Domain(TessellationFactors tessFactors, const OutputPatch<PackedVaryingsToDS, 3> input, float3 baryCoords : SV_DomainLocation)
+{
+    VaryingsToDS varying0 = UnpackVaryingsToDS(input[0]);
+    VaryingsToDS varying1 = UnpackVaryingsToDS(input[1]);
+    VaryingsToDS varying2 = UnpackVaryingsToDS(input[2]);
+
+    VaryingsToDS varying = InterpolateWithBaryCoordsToDS(varying0, varying1, varying2, baryCoords);
+
+    return VertTesselation(varying);
```
sebastienlagarde, line 223: note for myself. Lot of duplication, can we avoid it?

sebastienlagarde (approved):
left few comment, approving for the sake of iterationm can discuss in next PR

sebastienlagarde:
test arent green, looks like metal render grey image

com.unity.render-pipelines.high-definition/Runtime/Water/WaterVertexTessellation.hlsl
anisunity: I don't think we can without making it a mess

anisunity:
I'll check metal tomorrow

anisunity:
Unfortunately on metal there is a bug on the API side that will need to be fixed https://fogbugz.unity3d.com/f/cases/1397458/

anisunity:
Disabled the test on mac, disabled the infite feature on mac and relaunching the tests

anisunity:
merged master and relaunched the tests

anisunity:
@sebastienlagarde all green except for the hd template on the standalone, merging!

PR #6697, opened 2022-01-10
[HDRP] Fix reflection probe rendering order when visible in multiple cameras

simonb-unity:
### Purpose of this PR
Discovered while working on #6661, it seems that render requests are not processed in an order that respects dependencies when some dependencies are shared.

This occurs in the 002-HDMaterials test in HDRP_RuntimeTests. This test has two "root" cameras each with a planar reflection, and a single shared reflection probe. Logging the dependencies between render requests and the final rendering order (taking into account that code that iterates the list in reverse) on frame 1 produces:

```
Request 0 dependencies: 2, 3, 4, 5, 6, 7, 8
Request 1 dependencies: 2, 3, 4, 5, 6, 7, 9
Render order: 8, 0, 2, 3, 4, 5, 6, 7, 9, 1
```

`2-7` are the 6 faces of the reflection probe, `8` and `9` are the planar reflections. The reflection probe renders are emitted _between_ `0` and `1` even though they both depend on them.

These changes replace the flatten with a depth-first traversal of the request graph. With the changes, the order is:

```
Request 0 dependencies: 2, 3, 4, 5, 6, 7, 8
Request 1 dependencies: 2, 3, 4, 5, 6, 7, 9
Render order: 2, 3, 4, 5, 6, 7, 8, 0, 9, 1
```

The rendering order should match the previous behaviour otherwise.

---
### Testing status
Manually tested on 002-HDMaterials, HDRP PR tests on Yamato are green apart from DXR (which seems unrelated to these changes).


fredericv-unity3d (approved):

sebastienlagarde (approved):

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1892,8 +1901,9 @@ ref _cullingResults
 
 
                         // Execute render request graph, in reverse order
```
sebastienlagarde, line 1903: this comment is obsolete now?

simonb-unity: ah yep, will update the comment

sebastienlagarde (approved):

TomasKiniulis (approved):
Looks good. Tested Planar and multiple Realtime Reflection Probes(On Enable, Every Frame) with changed changed types assigned proxy volumes. Checked with multiple cameras and Scene views. Tried dragging all objects reflected by probe - don't see any frame delay.

https://user-images.githubusercontent.com/50582134/151360080-9f42704f-1984-4a52-83aa-3b68da9cfad2.mp4

Frame Debugger layout before PR:
image (44)

Frame Debugger layout with PR:
image (45)

sebastienlagarde (approved):

PR #6702, opened 2022-01-10
Add attributes to profiler wrappers to hide them in deep profiler

bandures:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1372769/
It adds attributes to profiler wrappers to hide them in deep profiler so that they don't break begin/end block for profiler markers. It uses a newly introduced attribute added in this PR https://ono.unity3d.com/unity/unity/pull-request/137412/_/profiler/bugfix/case1372769-calls-are-not-tracked-in-deep-profiling-mode

### Testing status
Manually tested using HDRP project


arttu-peltonen (commented):
Adding @JulienIgnace-Unity to review and @Unity-Technologies/gfx-qa-hdrp  to verify.
 
I'm not sure I understand why hiding these samplers in deep profiling mode fixes the problem, maybe Anton or Julien can explain? If it does fix it, and we see all the SRP brackets also in deep profiling mode, seems like a good change :)

@bandures please wait for the trunk fix to land, and then re-run the Safety Net yamato playlist, currently it fails to compile because the attribute doesn't exist yet. This job must be green before merging.

arttu-peltonen:
@JulienIgnace-Unity also can you confirm that backporting this is not necessary? To me at least it would seem like unnecessary hassle to do so since it depends on a trunk change.

JulienIgnace-Unity (approved):

JulienIgnace-Unity:
@arttu-peltonen yeah not sure we need to backport this unless requested. We'd need to have the C++ PR backported first anyway.

PR #6704, opened 2022-01-10
[2022.2] Shaders: Fixing two lerp functions that were causing issues in our CI.

ellioman:
# Purpose of this PR
In URP's codebase there were two calls in our Shader library to `lerp(half, half3, half)`. 
This causes issues in some compilers, which for example broke two of our CI jobs.

This PR fixes the issue by changing it to `lerp(half3, half3, half)`.

# Testing status
* Tested to see if there were any errors when opening a project
* Ran Yamato


phi-lira (approved):

ellioman:
Failures in CI already happening on Master and not caused by this PR. This PR is successful in fixing the compilation errors.

PR #6706, opened 2022-01-10
[2022.1 backport] Scriptable Renderer: Dismiss depth slice for non-camera passes check

sandy-carter-unity:
The RTHandles conversion puts the XR depth slice for SPI (-1) on all
targets at an earlier stage. All equals checks against CameraTarget must
dismiss possible depth slice to be accurate.

This fixes a depth non-clear in vfx BatchRenderGroup_URP tests by
preventing `m_FirstTimeCameraDepthTargetIsBound` from being set to `false`
before batch opaque draw call.

 Backport of #6426


ellioman (approved):
Looks good. Do we not need to add to the Changelog ?

phi-lira (approved):

sebastienlagarde (approved):

PR #6707, opened 2022-01-11
[Universal] CopyDepth pass editor ConfigureTarget for DX11-only

jonuuukas:
To fix some errors on Vulkan tests, we need to make this call of ConfigureTarget dx11 exclusive as recently an error that would cause a failure on Vulkan 

Ran URP Foundation tests that were failing - no related errors found anymore (only known instabilities/failures)


ellioman (approved):

ellioman:
CI Failures are already happening on Master and not caused by this PR.
This PR is successful at fixing issues in two CI jobs.

PR #6708, opened 2022-01-11
Read-only TextInput text is invisible

julienamsellem:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1387237/

Some recent changes in UIToolkit made our read-only TextInput unreadable because of the opacity being set to all disabled elements in the UI hierarchy instead of only to the leaf.
[Before]
Unity_U8TGh9Wrai

[After]
Unity_Bw1bsbDBeo

---
### Testing status
Checked for operators connected together and for operator connected to a block.

---
### Comments to reviewers
The risks are really limited since the change should affect only "ports".


VladNeykov (approved):
Hi, looking good!
Tested:
- operators and blocks
- single and multiple connected operators
- with subgraph operator
- scanned though a large VFX containing all operators and blocks, didn't see any regressions

Thanks for the fix! :)

ludovic-theobald (approved):
Looks good :)

PR #6712, opened 2022-01-11
[Universal][Editor] Fixing incorrect light attenuation with mobile build target selected

jonuuukas:
Recently SHADER_API_MOBILE started working on editor as well, depending on the active build target. Attenuation had a single check for mobile platform which wouldn't be set on Editor. This PR includes a check for editor and sets correct values leading to correct rendering

Fix for https://fogbugz.unity3d.com/f/cases/1387142/

Tested manually on editor and ios with the bug case project

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
@@ -793,9 +793,13 @@ namespace UnityEngine.Rendering.Universal
                 float lightRangeSqrOverFadeRangeSqr = -lightRangeSqr / fadeRangeSqr;
                 float oneOverLightRangeSqr = 1.0f / Mathf.Max(0.0001f, lightRange * lightRange);
 
-                // On untethered devices: Use the faster linear smoothing factor (SHADER_HINT_NICE_QUALITY).
+                // On untethered devices and Editor if mobile target (as SHADER_API_MOBILE now affects Editor as well): Use the faster linear smoothing factor (SHADER_HINT_NICE_QUALITY).
                 // On other devices: Use the smoothing factor that matches the GI.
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -793,9 +793,13 @@ static GraphicsFormat MakeRenderTextureGraphicsFormat(bool isHdrEnabled, bool ne
...
 
-                // On untethered devices: Use the faster linear smoothing factor (SHADER_HINT_NICE_QUALITY).
+                // On untethered devices and Editor if mobile target (as SHADER_API_MOBILE now affects Editor as well): Use the faster linear smoothing factor (SHADER_HINT_NICE_QUALITY).
                 // On other devices: Use the smoothing factor that matches the GI.
-                lightAttenuation.x = Application.isMobilePlatform || SystemInfo.graphicsDeviceType == GraphicsDeviceType.Switch ? oneOverFadeRangeSqr : oneOverLightRangeSqr;
+                bool isMobileBuildTarget = false;
+                #if UNITY_EDITOR && (UNITY_IOS || UNITY_ANDROID)
+                    isMobileBuildTarget = true;
+                #endif
+                lightAttenuation.x = Application.isMobilePlatform || isMobileBuildTarget || SystemInfo.graphicsDeviceType == GraphicsDeviceType.Switch ? oneOverFadeRangeSqr : oneOverLightRangeSqr;
```
phi-lira, line 802: I think you can replace` Application.isMobilePlatform `with `UnityEngine.Rendering.GraphicsSettings.HasShaderDefine(BuiltinShaderDefine.SHADER_API_MOBILE)` and it should work on both editor and runtime.
 
 @eh-unity for awareness, as you were looking into usages of Application.isMobilePlatform and SHADER_API_MOBILE.

eh-unity: This code is gone in SHADER_API_MOBILE PR. There's only "oneEverLightRangeSqr" left.

PR #6715, opened 2022-01-11
Added an option for culling objects out of the ray tracing acceleration structure.

anisunity:
This  PR adds the option for users to specify a culling mode for the ray tracing acceleration structure. This can be specified in the Ray Tracing Settings Volume. The user can chose between three options:
- Extended frustum: This mirror the frustum and takes the bounding box of the result (default option)
- Sphere: An additional option becomes available in the volume and allows the user to specify a radius.
- None: Previous state

Testing Status
Three new tests were created to cover this new feature.
image


remi-chapelain (approved):
Has already been pair tested while making the precious automated tests. 
Looking good. ✔️

sebastienlagarde (approved):
Pleas add an entry in What's new for this thanks

anisunity:
What is new updated, all DXR tests are green locally all good!

PR #6716, opened 2022-01-11
Stop creating APV data assets for scenes in baking sets, but without probe volumes in it

FrancescoC-unity:
Before this PR, we could get in a weird situation in which we have an asset created for a scene that has no probe volume (especially if it had one before). 

This PR fixes this.


com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -573,7 +576,7 @@ static void OnAdditionalProbesBakeCompleted()
                 {
                     // This scene has a reference volume authoring component in it?
                     ProbeVolumePerSceneData data = null;
-                    if (scene2Data.TryGetValue(scene, out data))
+                    if (scene2Data.TryGetValue(scene, out data) && data2Asset[data] != null)
```
adrien-de-tocqueville, line 579: Picky comment, but I would replace that by `data2Asset.Contains(data)`, to remove the added if below, and above only add the element to the dictionnary if `hasAsset` is true

com.unity.render-pipelines.core/Runtime/Lighting/ProbeVolume/ProbeVolumePerSceneData.cs
```diff
@@ -77,7 +77,16 @@ internal void InvalidateAllAssets()
             foreach (var asset in assets.Values)
             {
                 if (asset != null)
+                {
                     ProbeReferenceVolume.instance.AddPendingAssetRemoval(asset);
+#if UNITY_EDITOR
+                    if (AssetDatabase.TryGetGUIDAndLocalFileIdentifier(asset, out string guid, out long file))
```
adrien-de-tocqueville, line 83: why not use AssetDatabase.GetAssetPath ?

FrancescoC-unity: I was actually testing something else for which I needed also the GUID, now indeed that one makes more sense; will try tomorrow.

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
FrancescoC-unity: I need to double check if somewhere we rely on having one entry in that array per scene; will check tomorrow, but if all is fine I do agree with you

com.unity.render-pipelines.core/Editor/Lighting/ProbeVolume/ProbeGIBaking.cs
```diff
@@ -308,6 +308,8 @@ internal static void PerformDilation()
 
             foreach (var sceneData in perSceneDataList)
             {
+                if (!ProbeReferenceVolume.instance.sceneData.SceneHasProbeVolumes(sceneData.gameObject.scene)) continue;
```
adrien-de-tocqueville, line 311: isn't it possible to destroy all perSceneData objects of scenes that are invalid at start of baking ? I suspect we will forget to call this `SceneHasProbeVolumes` at some point and end up with the same issue

FrancescoC-unity: Yes, doable after this https://github.com/Unity-Technologies/Graphics/pull/6907 PR lands.

JulienIgnace-Unity (approved):

sebastienlagarde (approved):

PR #6725, opened 2022-01-12
Disabled the "Reflect Sky" feature in the case of transparent screenspace reflections for the water system.

anisunity:
In order to avoid multiple artifacts that can happen with transparent objects and screen space reflections, we've decided to disable screen space reflections's reflect sky option on the transparent screen space reflections. The water surfaces is the most obvious case. This does not affect le ray traced reflections on transparent.

Testing status
Tested with the water system, issue doesn't happen anymore.


remi-chapelain (approved):
I just added some info about reflect sky being not supported for SSR on transparent to avoid future confusion but other than that, change has already been discussed and approved ✔️ 
Hopefully, it won't break any current automated test.

sebastienlagarde (approved):

